PR #10, opened 2022-11-03
Create README.md

clubanderson:
## Related issue(s)

Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,24 @@
+# edge-mc
+Edge Multi-Cluster
+
+Driving Concerns
+
+General Context: hierarchy, infrastructure & platform, roles & responsibilities, integration architecture, security issues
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,24 @@
+# edge-mc
+Edge Multi-Cluster
+
```
pdettori, line 3: I think we should start by saying that this is a subproject of kcp that aims to address the concerns of Edge Multi Cluster scenarios

pdettori: Also, I suggest we follow an outline similar to the main kcp project - something along the lines of 

- Overview
- Quickstart
- Next steps
- Contributing

We may not have yet code to try in Quickstart, but I think what might be of interest for developers browsing this site will be something to try out.

ezrasilvera:
Can we just use Edge instead of Edge-mc ?

clubanderson:
discarding -  superseded by #13

PR #67, opened 2022-12-21
:sparkles: Start drafting mailbox controller

MikeSpreitzer:
## Summary
Begin drafting the mailbox controller for an early simple crude milestone.

### DIFF
--- a/config/crds/edge.kcp.io_edgeplacements.yaml
+++ b/config/crds/edge.kcp.io_edgeplacements.yaml
@@ -13,7 +13,7 @@ spec:
     listKind: EdgePlacementList
     plural: edgeplacements
     singular: edgeplacement
-  scope: Namespaced
+  scope: Cluster
   versions:
   - name: v1alpha1
     schema:
--- a/config/crds/edge.kcp.io_singleplacementslice.yaml
+++ b/config/crds/edge.kcp.io_singleplacementslice.yaml
@@ -13,7 +13,7 @@ spec:
     listKind: SinglePlacementSliceList
     plural: singleplacementslice
     singular: singleplacementslice
-  scope: Namespaced
+  scope: Cluster
   versions:
   - name: v1alpha1
     schema:
--- a/config/exports/apiexport-edge.kcp.io.yaml
+++ b/config/exports/apiexport-edge.kcp.io.yaml
@@ -5,6 +5,6 @@ metadata:
   name: edge.kcp.io
 spec:
   latestResourceSchemas:
-  - v230125-b6052f8.edgeplacements.edge.kcp.io
-  - v230125-b6052f8.singleplacementslice.edge.kcp.io
+  - v230126-0a8bf97.edgeplacements.edge.kcp.io
+  - v230126-0a8bf97.singleplacementslice.edge.kcp.io
 status: {}
--- a/config/exports/apiresourceschema-edgeplacements.edge.kcp.io.yaml
+++ b/config/exports/apiresourceschema-edgeplacements.edge.kcp.io.yaml
@@ -2,7 +2,7 @@ apiVersion: apis.kcp.io/v1alpha1
 kind: APIResourceSchema
 metadata:
   creationTimestamp: null
-  name: v230125-b6052f8.edgeplacements.edge.kcp.io
+  name: v230126-0a8bf97.edgeplacements.edge.kcp.io
 spec:
   group: edge.kcp.io
   names:
@@ -10,7 +10,7 @@ spec:
     listKind: EdgePlacementList
     plural: edgeplacements
     singular: edgeplacement
-  scope: Namespaced
+  scope: Cluster
   versions:
   - name: v1alpha1
     schema:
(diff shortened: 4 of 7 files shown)


MikeSpreitzer:
The lint complaint is because of an oversight in the check for contextual logging, fixed in https://github.com/kubernetes-sigs/logtools/pull/10

cmd/mailbox-controller/main.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	wsInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    onAdd,
+		UpdateFunc: onUpdate,
+		DeleteFunc: onDelete,
+	})
+	kcpSharedInformerFactory.Start(ctx.Done())
+	<-ctx.Done()
+	logger.Info("Time to stop")
+}
```
pdettori, line 93: This looks like a good start to exercise workspace informers. Can we start looking at structuring the code in a way more similar to how it is currently structured in the kcp core project ? E.g.  we could have the core logic in`pkg/mailbox-controller` with `mailbox-controller.go` and `mailbox-reconciler.go`, and have main/setup code/command parsing in cmd/mailbox-controller. For now it's probably ok to have separate mains for the different controllers, but I think we may want to have just one main with a controller manager starting all controllers, and there we can use cobra etc. for parameters parsing etc.

cmd/mailbox-controller/main.go
```diff
@@ -0,0 +1,93 @@
...
+	"k8s.io/klog/v2"
+
+	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	kcpinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+
+	edgeindexers "github.com/kcp-dev/edge-mc/pkg/indexers"
+)
+
+func main() {
+	resyncPeriod := time.Duration(0)
```
pdettori, line 38: Why resyncPeriod is 0 ? usually default is a large value (e.g. 10 * time.Hour )

cmd/mailbox-controller/main.go
MikeSpreitzer: Yes, this current form might not be what we ultimately want.  OTOH, it might be.  OTOH, factoring can't hurt.  We should discuss what we want in a larger forum than this PR.  Can we put off the factoring to a later PR?

cmd/mailbox-controller/main.go
MikeSpreitzer: I write controllers that do not need any resync at all.

cmd/mailbox-controller/README.md
```diff
@@ -0,0 +1,24 @@
+# Mailbox Controller
```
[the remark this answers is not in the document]
MikeSpreitzer: I revised this PR to introduce `docs/` and put the mailbox-controller doc there.

docs/mailbox-controller.md
```diff
@@ -0,0 +1,24 @@
+# Mailbox Controller
+
+This is a component in an architecture not yet shared in detail.  It is an early milestone in edge multicluster, achieving partial functionality and implemented in a simple crude approach that is layered on top of transparent multicluster.  In this approach the edge-mc layer maintains a kcp workspace, called a mailbox, per selected edge destination per EdgePlacement object.  Then TMC syncs between the mailbox workspaces ad the edge destinations.
+
```
ezrasilvera, line 4: typo ?  "between the mailbox workspaces **and** the edge"

MikeSpreitzer: Good catch.  Fixed in latest.

MikeSpreitzer:
The force-push to 2e5fcf2 is a squash plus fixing the typo that @ezrasilvera found.

cmd/mailbox-controller/main.go
pdettori: ok for me, but let's keep this in mind

cmd/mailbox-controller/main.go
pdettori: Does `resyncPeriod = 0` mean no resync ? Is this documented somewhere ?

MikeSpreitzer: Sadly the generated informer code lacks good comments.  It is based on the generic library, which documents the answer to your question.  See https://github.com/kubernetes/client-go/blob/v0.24.9/tools/cache/shared_informer.go#L214-L224

MikeSpreitzer:
The force-push to 7497dae is a rebase onto main followed by `make imports`.

MikeSpreitzer:
The force-push to 8114b75 is a rebase onto `main` and almost complete revision.

Now everything compiles, and the current fragment of the mailbox controller behaves as expected.

But #136 is still a mystery; this PR includes overwrites of the wrong CRD content.

MikeSpreitzer:
The force-push to 6da59eb removes deviations from contextual logging in the mailbox controller.

MikeSpreitzer:
This code has drifted into the scheduler's lane.  I rebased and opened #147 in recognition of that.  I will make another start on the mailbox controller.

PR #68, opened 2022-12-21
✨ Begin drafting edge placement API

MikeSpreitzer:
## Summary
This PR adds a draft of the API object types for EdgePlacement and SinglePlacementSlice.  The latter is a scalable interface between scheduling and syncing.

## Related issue(s)
The infrastructure for generating the related code does not exist yet.
I do not know (yet) whether kubebuilder will be involved.

Fixes #


pkg/apis/v1alpha1/edge-placement.go
```diff
@@ -0,0 +1,130 @@
...
+//   - two directives to propagate the same object to the same destination are
+//     simply redundant (they, by design, can not conflict).
+type EdgePlacement struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// `spec` describes a desired binding between Namespaces and Locations.
```
waltforme, line 52: Namespaces _and non-namespaced objects_?

pkg/apis/v1alpha1/single-placement.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SourcePlacementLabelKey is the key of the label used in a SinglePlacementSlice
+// to reference the EdgePlacement that it is part of the response to.
+// We use a label rather than a field because field selectors do not work
+// on a resource defined by a CRD.
+const SourcePlacementLabelKey string = "edge.research.ibm.com/source-placement"
```
waltforme, line 28: Should we use something else in place of our specific company?

pkg/apis/v1alpha1/edge-placement.go
MikeSpreitzer: Fixed in latest revision.

MikeSpreitzer:
Note that, unlike kcp-dev/kcp, the apis package here is not further divided.  That implies a little bigger difference from the kcp-dev/kcp infrastructure for code generation.  Or maybe we want to divide here too?  What subdivisions would make sense?  (Not clear to me.)

pkg/apis/v1alpha1/edge-placement.go
```diff
@@ -0,0 +1,130 @@
...
+// EdgePlacementSpec holds a desired binding between (a) Namespaces and non-namespaced objects
+// and (b) Locations.
+type EdgePlacementSpec struct {
+	// `locationWorkspaceSelector` identifies the workspaces in which to look for Location
+	// objects, in terms of labels on the Workspace objects.
+	LocationWorkspaceSelector metav1.LabelSelector `json:"locationWorkspaceSelector,omitempty"`
+
+	// `locationSelectors` identifies the relevant Location objects in terms of their labels.
+	// A Location is relevant if and only if it passes any of the LabelSelectors in this field.
+	LocationSelectors []metav1.LabelSelector `json:"locationSelectors,omitempty"`
```
pdettori, line 74: Does this mean that first the LocationWorkspaceSelector is applied and then the LocationSelector is applied for each of the workspaces selected by the first selector ?

MikeSpreitzer: Yes

MikeSpreitzer: I am still learning about APIExport and binding in kcp.  I suspect that has a different way to solve the problem: some _service provider_ does APIExport of `Location` from one of their workspaces, and maintains controllers that read --- indirectly through the one workspace with the APIExport --- all the Locations objects based on APIBindings of that APIExport.

MikeSpreitzer:
The force-push to e516db2 is a squash and rebase, plus a commit that adds the missing doc.go and both missing register.go.

pkg/apis/v1alpha1/edge-placement.go
pdettori: Right, I think that would affect the design of this API.

MikeSpreitzer:
After the `go mod tidy`, everything still compiles.  This is good news, it means our go.mod is not hacky anymore.  I think we can thank #129 for tracking changes to kcp.

waltforme:
> After the `go mod tidy`, everything still compiles. This is good news, it means our go.mod is not hacky anymore. I think we can thank #129 for tracking changes to kcp.

Nice!

clubanderson (approved):

PR #71, opened 2023-01-03
:bug: Differentiate logcheck output

### LINKED ISSUE #70: bug: verify-contextual-logging.sh does not differentiate logcheck output
### Describe the bug

hack/verify-contextual-logging.sh invokes logcheck twice, combining both outputs into one file in a way that does not make it clear which output is coming from which invocation.

### Steps To Reproduce

1. Look at hack/logcheck.out and try to figure out whether the output is from the first or second invocation of logcheck

### Expected Behaviour

I expected to be able to understand the complaints.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
Make hack/verify-contextual-logging.sh keep track of which logcheck output is from which invocation.

Also make it note the location of the new output, to help developers debug problems.

## Related issue(s)

Fixes #70

### DIFF
--- a/hack/logcheck.out
+++ b/hack/logcheck.out
@@ -1,1 +1,1 @@
-logcheck: ./... matched no packages
+APIs: logcheck: ./... matched no packages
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
This is not very important, now that we are no longer getting an error message without a filename.  Closing for now.

PR #72, opened 2023-01-03
🌱 Use latest release of logcheck

MikeSpreitzer:
## Summary
This PR updates the Makefile so that CI uses the latest release of logcheck, picking up the fix to the bug that bit #67 

## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ GOTESTSUM_VER := v1.8.1
 GOTESTSUM_BIN := gotestsum
 GOTESTSUM := $(abspath $(TOOLS_DIR))/$(GOTESTSUM_BIN)-$(GOTESTSUM_VER)
 
-LOGCHECK_VER := v0.2.0
+LOGCHECK_VER := v0.4.1
 LOGCHECK_BIN := logcheck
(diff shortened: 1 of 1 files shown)


hack/logcheck.out
```diff
@@ -1 +1,3 @@
+/pkg/reconciler/scheduling/placement/controller.go:209:3: Key positional arguments "Namespace" are expected to no special characters. Please refer to https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#name-arguments.
+/pkg/reconciler/scheduling/placement/controller.go:237:3: Key positional arguments "Location" are expected to no special characters. Please refer to https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#name-arguments.
```
MikeSpreitzer, line 2: @waltforme : you may be interested in these new complaints from logcheck

MikeSpreitzer: See the Kubernetes norms in https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#name-arguments .  They call for the keys to be lowerCamelCase.

clubanderson (approved):

PR #117, opened 2023-01-10
:bug: updates to makefile and supporting files to allow successful build, build-all, and crds

clubanderson:
updates to makefile and supporting files to allow successful build, build-all, and crds

## Summary
updates to makefile and supporting files to allow successful build, build-all, and crds

## Related issue(s)

Fixes #


MikeSpreitzer:
Why were imports re-ordered?  I think it makes sense to group them by layers of dependence: go stuff first, other kube-independent stuff, regular kube, kcp, then edge-mc.

Why was vendor/ added?  I thought the point of go modules is to render vendoring obsolete.

MikeSpreitzer:
why was logcheck removed?

clubanderson:
vendoring is removed in latest PR that supports codegen.  I rebased the makefile on the current KCP version.  We can certainly re-order imports to suit your preference.

PR #127, opened 2023-01-17
🌱 Improve and use `make imports`

MikeSpreitzer:
## Summary
Make everything under kcp-dev sort together in the last block of imports.

Because openshift-goimports is rather opinionated and does not have a way to separate kcp-dev/kcp from kcp-dev/edge-mc with both coming after kubernetes.

Actually invoked `make imports` --- whose effects have not appeared in earlier PRs and whose effects will appear in future PRs that include the effects of `make codegen`.

## Related issue(s)

Fixes #

### DIFF
--- a/cmd/placement/cmd/placement.go
+++ b/cmd/placement/cmd/placement.go
@@ -30,15 +30,14 @@ import (
 	"k8s.io/component-base/version"
 	"k8s.io/klog/v2"
 
-	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
-	kcpinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
-	kcpfeatures "github.com/kcp-dev/kcp/pkg/features"
-	"github.com/kcp-dev/logicalcluster/v2"
-
 	placementoptions "github.com/kcp-dev/edge-mc/cmd/placement/options"
 	edgeclient "github.com/kcp-dev/edge-mc/pkg/client"
 	edgeindexers "github.com/kcp-dev/edge-mc/pkg/indexers"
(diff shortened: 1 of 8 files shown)


MikeSpreitzer:
Related: https://github.com/openshift-eng/openshift-goimports/pull/15

cmd/placement/cmd/placement.go
```diff
@@ -30,15 +30,14 @@ import (
...
-	"github.com/kcp-dev/logicalcluster/v2"
-
 	placementoptions "github.com/kcp-dev/edge-mc/cmd/placement/options"
 	edgeclient "github.com/kcp-dev/edge-mc/pkg/client"
 	edgeindexers "github.com/kcp-dev/edge-mc/pkg/indexers"
 	edgeplacement "github.com/kcp-dev/edge-mc/pkg/reconciler/scheduling/placement"
+	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	kcpinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+	kcpfeatures "github.com/kcp-dev/kcp/pkg/features"
+	"github.com/kcp-dev/logicalcluster/v2"
```
waltforme, line 40: Should kcp-dev/edge-mc go last?

MikeSpreitzer: Using openshift-goimports before https://github.com/openshift-eng/openshift-goimports/pull/15 merges means that if kcp-dev/edge-mc goes last then kcp-dev/kcp goes _before_ kubernetes, which I think is pretty bad.

waltforme: OK

PR #135, opened 2023-01-19
✨ Use kcp `main`, then update the placement controller accordingly

waltforme:
## Summary
This PR is a collaboration.

The first part comes from @MikeSpreitzer, which updates go.mod and the Makefile to use the latest version of kcp-dev/kcp, and re-generates code from the API.

The second part comes from myself, which brings the separated TMC scheduling-placement-controller back to track.

The controller now behaves in the same way with its origin --- the counterpart living in kcp-dev/kcp. However, it is in the same _buggy_ way. Both controllers don't pick up the creation event of Locations. Assuming a Placement exists before a Location is created, the later creation of that Location doesn't trigger the controllers to update the Placement, even if the Location matches the Placement's `spec.locationSelectors`.

Nonetheless, we can still completely traverse the Placement State Machine described in this [document](https://docs.google.com/document/d/1AzyjuyjNIDVAXEGHslaggltIQ9Cs8pLCzc9Ma_RBmuM/edit#heading=h.vmt32rdidje6), or in this [code block](https://github.com/kcp-dev/kcp/blob/fb4d4a42373ea4da001b8c88396eabaf6f825be1/pkg/apis/scheduling/v1alpha1/types_placement.go#L123-L134), by weaken our test case. Instead of catching the creation events of Locations, we can edit the `spec.locationSelectors` of Placements to match/mismatch the Location. The original controller living in kcp-dev/kcp also passes this weaker test case. The README.md is updated for this weaker test.

I suspect that there are some bugs in kcp-dev/kcp. But from the stand point of edge-mc, we don't need to worry about it. Because (1) our purpose is to mimic the original behavior and we faithfully did it, (2) this TMC controller will soon be replaced by our controller for edge.


waltforme:
Force-push https://github.com/kcp-dev/edge-mc/commit/babd16b3c43e1fca891af7664c1ff5bb9f515d57 tries to be compliant with logcheck (Thanks to the clearer messages!), and documents the tested kcp commit.

go.mod
```diff
@@ -19,6 +16,12 @@ require (
 	k8s.io/kube-openapi v0.0.0-20220328201542-3ee0da9b0b42
 )
 
+require (
```
MikeSpreitzer, line 19: I see no need to maintain two separate `require` sections at the top, let's merge these two.

pkg/indexers/util.go
```diff
@@ -0,0 +1,60 @@
+/*
+Copyright 2022 The KCP Authors.
```
MikeSpreitzer, line 2: If this is a new file then I would date it 2023.

cmd/placement/README.md
```diff
@@ -33,24 +33,33 @@ go run cmd/placement/main.go --kcp-kubeconfig=<path to kcp admin.kubeconfig> -v
...
-In a user's workspace, say `root:my-org:dev`, create Placement `dev`:
+In the `root:compute` workspace, create Location `foo`:
 ```console
-kubectl create -f config/samples/placement_dev.yaml
+kubectl create -f config/samples/location_foo.yaml
 ```
-The status of Placement `dev` should show `phase: Pending`.
 
-Switch to the `root:compute` workspace, create Location `foo`:
+Switch to a user's workspace, say `root:my-org:dev`, create Placement `dev`:
```
MikeSpreitzer, line 41: That's not what a user's workspace looks like in the kcp version used here.  I see the following.

```
(base) mspreitz@mjs12 edge-mc % kubectl ws \~
```

MikeSpreitzer (approved):
I noted a few things in-line, but they can be done in follow-up.

/lgtm

cmd/placement/README.md
waltforme: Improving this in https://github.com/kcp-dev/edge-mc/pull/138, thanks! @MikeSpreitzer

PR #142, opened 2023-01-25
:bug: updated api-gen call to include repo_root

### LINKED ISSUE #140: feature: `make crds` should include APIExport generation
### Feature Description

edge-mc should have the functional equivalent of https://github.com/kcp-dev/kcp/blob/cb8dd3c46470f30414c6789be956162ccb21f669/hack/update-codegen-crds.sh#L55-L58 --- ideally without maintaining a copy of the apigen tool.

### Proposed Solution

Have the Makefile or a script `go install github.com/kcp-dev/kcp/cmd/apigen@$right_version` into some tools binary directory and invoke it from there.  https://github.com/kcp-dev/kcp/pull/2669 has merged and this makes it possible to `go install` apigen.

I just tested the apigen from kcp commit cb8dd3c46470f30414c6789be956162ccb21f669 and it produced exactly the same output as the apigen binary that I built earlier from kcp commit 4506fdc06406 (which is the one that edge-mc currently depends on).

### Alternative Solutions

- Maintain a copy of the apigen source.
- Have the Makefile or script `git clone` kcp to get the source and then build it.

### Want to contribute?

### Additional Context

_No response_

clubanderson:
## Related issue(s)

Fixes #140

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -376,7 +376,6 @@ verify-go-versions:
 modules: ## Run go mod tidy to ensure modules are up to date
 	go mod tidy
 	cd pkg/apis; go mod tidy
-# hack/update-go-modules.sh
 
 .PHONY: verify-modules
 verify-modules: modules  ## Verify go modules are up to date
@@ -384,7 +383,6 @@ verify-modules: modules  ## Verify go modules are up to date
 		git diff; \
 		echo "go module files are out of date"; exit 1; \
 	fi
-# hack/verify-go-modules.sh
 
 .PHONY: help
 help: ## Show this help.


clubanderson:
without REPO_ROOT make crds target fails

clubanderson:
@MikeSpreitzer did you 'make tools' first?

clubanderson:
/retitle :bug: updated api-gen call to include repo_root

MikeSpreitzer:
I did not `make tools` first.  It looks like the Makefile tries to pre-req the relevant tools but overlooks one:
```
crds: $(CONTROLLER_GEN) $(YAML_PATCH)
	./hack/update-codegen-crds.sh
.PHONY: crds
```

PR #148, opened 2023-01-26
:book: Start documenting 2023q1 PoC

MikeSpreitzer:
## Summary
This PR introduces a high-level overview of a proposal for a 2023q1 Proof-of-Concept.

## Related issue(s)

This is a high level outline that is a partial response to #157


MikeSpreitzer:
I have somewhat outdated proposals for the summarization interface at https://docs.google.com/document/d/1yY-XgOtCnOPKpVdxdPVhH5ixRN6IWQQe0s3lip1ID0Y/edit?usp=sharing and the customization interface at https://docs.google.com/document/d/1-YF4X9BI8UtgaV2nFr-R_LIeDU0ygQYuwtDtl4tZzDA/edit?usp=sharing .  I intend to introduce updated versions as PRs in the future.

docs/poc2023q1/outline.md
```diff
@@ -0,0 +1,106 @@
...
+This controller maintains one mailbox workspace per SyncTarget.  Each
+of these mailbox workspaces is used for a distinct TMC problem (e.g.,
+Placement object).  These workspaces are all children of the edge
+service provider workspace.
+
+## Edge Scheduler
+
+This controller monitors the EdgePlacement, Location, and SyncTarget
+objects and maintains the results of matching.  For each EdgePlacement
+object this controller maintains an adjacent collection of
```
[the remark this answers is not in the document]
MikeSpreitzer: Addressed in latest revision.

docs/poc2023q1/outline.md
```diff
@@ -0,0 +1,112 @@
...
+service provider workspace.
+
+## Edge Scheduler
+
+This controller monitors the EdgePlacement, Location, and SyncTarget
+objects and maintains the results of matching.  For each EdgePlacement
+object this controller maintains an associated collection of
+SinglePlacementSlice objects holding the matches for that
+EdgePlacement.  These SinglePlacementSlice objects appear in the same
+namespace as the corresponding EdgePlacement; the remainder of how
```
pdettori, line 94: Should it be "workspace" instead of "namespace" ? Are EdgePlacement and SinglePlacementSlice non-namespaced objects ?

MikeSpreitzer: Yes, it was a typo.  Thanks.  Fixed in the latest rev.

docs/poc2023q1/outline.md
```diff
@@ -0,0 +1,257 @@
...
+workspaces that contain dynamic collections of Location and SyncTarget
+objects as defined in [kcp
+TMC](https://github.com/kcp-dev/kcp/tree/main/pkg/apis), and that one
+view can be used to read those Location objects and one view can be
+used to read those SyncTarget objects.
+
+To enable the workload management layer to manipulate each edge
+cluster, that cluster's SyncTarget has an annotation that refers to a
+Secret that holds a kubeconfig that can be used by the Bootstrapper to
+install the edge-mc machinery in the edge cluster.  These annotations
```
ezrasilvera, line 79: What exactly do you mean "install the edge-mc machinery" - what do I need to have on the edge in order to be able to install this (e.g., existing k8s custer, etc..)?

docs/poc2023q1/outline.md
```diff
@@ -0,0 +1,257 @@
...
+and Secrets are created by the inventory management layer for use by
+the workload management layer.
+
+To complete the plumbing of the syncers, for each SyncTarget the
+mailbox controller creates the following needed accompanying items in
+that SyncTarget's workspace (which must authorize the mailbox
+controller to do this).
+
```
ezrasilvera, line 87: Is that the only controller that access the edge cluster?

docs/poc2023q1/outline.md
```diff
@@ -0,0 +1,257 @@
...
+   APIResourceImports (what are these?).
+3. A ClusterRoleBinding that links that ServiceAccount with that
+   ClusterRole.
+
+FYI, those are the things that `kubectl kcp workload sync` directly
+creates besides the SyncTarget.
+
+## Edge Service Provider workspace
+
+The edge multi-cluster service is provided by one workspace that
```
ezrasilvera, line 99: Is the "one" workspace a point in time statement or needed for this architecture?  I suspect that with large scale + sharding you may have multiple such workspaces?

docs/poc2023q1/outline.md
```diff
@@ -0,0 +1,257 @@
...
+includes the following things.
+
+- An APIExport of the edge API group.
+- The edge controllers: scheduler, placement translator, bootstrapper,
+  mailbox controller, and status sumarizer.
+
+## Workload Management workspaces
+
+The users of edge multi-cluster primarily maintain these.  Each one
```
ezrasilvera, line 108: What do you mean by "users of edge multi-cluster"  ? is it the admins of the edge cluster? users of the edge? admins of the edge-mc platform?  We probably need to define roles in the PoC

docs/poc2023q1/outline.md
```diff
@@ -0,0 +1,257 @@
...
+1. APIBindings (maintained by the mailbox controller) to APIExports of
+   workload object types.
+2. Workload namespaces holding workload objects, post customization.
+3. A TMC Placement object.
+
+## Edge cluster
+
+Also called edge pcluster.
+
+One of these contains the following items, maintained by the
```
ezrasilvera, line 138: maintained by the bootstrapper or only provisioned by the bootstrapper?  Maybe we can maintain/manage them through KCP once they are provisioned?

docs/poc2023q1/outline.md
```diff
@@ -0,0 +1,257 @@
...
+The edge-mc implementation propagates the desired state from center to
+edge and collects the specified information from edge to center.
+
+The edge user monitors status summary objects in their workload
+management workspaces.
+
+The status summaries may include limited-length lists of broken
+objects.
+
+Full status from the edge is available in the mailbox workspaces.
```
ezrasilvera, line 257: So we do maintain the full status of all objects in a centralized place?  Can you elaborate what scalability aspects do we solve with the summaries?   Do we plan to put the summaries & the full statuses on different type of data-bases?  Are they going to be accessed through different apiservers?

docs/poc2023q1/outline.md
```diff
@@ -0,0 +1,257 @@
...
+This controller maintains one mailbox workspace per SyncTarget.  Each
+of these mailbox workspaces is used for a distinct TMC problem (e.g.,
+Placement object).  These workspaces are all children of the edge
+service provider workspace.
+
+## Bootstrapper
+
+The bootstrapper is a controller that maintains the needed workload
+management machinery (the items enumerated above in the description of
+the edge clusters) in each edge cluster.
```
pdettori, line 167: Is it important to have the bootstrapper as part of this PoC ? Could we just use some automation (e.g. bash scripts) to generate the syncer install yaml from the inventory workspace and then apply it to each pCluster ? I think in more realistic scenarios bootstrap will be managed with something like FIDO Device Onboard or similar, but I don't think we are building something along those lines in the scope of the PoC.

docs/poc2023q1/outline.md
```diff
@@ -0,0 +1,257 @@
...
+view can be used to read those Location objects and one view can be
+used to read those SyncTarget objects.
+
+To enable the workload management layer to manipulate each edge
+cluster, that cluster's SyncTarget has an annotation that refers to a
+Secret that holds a kubeconfig that can be used by the Bootstrapper to
+install the edge-mc machinery in the edge cluster.  These annotations
+and Secrets are created by the inventory management layer for use by
+the workload management layer.
+
```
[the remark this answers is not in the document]
ezrasilvera: Reading more, I can see you have a bootstarp controller running (as part of edge-mc?). So does it mean bootsraping of the edge cluster is done through KCP ?  
It means that both the syncer and this controller will interact  with the edge cluster ?

docs/poc2023q1/outline.md
MikeSpreitzer: This workspace does store a data volume that goes up with usage, so I do not see a need to scale out this workspace.
It *does* have controllers whose workload scales in volume and throughput as usage increases, but that is a matter that can be handled by sharding those controllers, not this workspace.

docs/poc2023q1/outline.md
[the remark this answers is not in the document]
MikeSpreitzer: If you look at the TMC quickstart scenario in the kcp home page, you will see that the last step in https://docs.google.com/document/d/1hrnPuMgorw6vx76IOYRyqwxDkKN6MlK3r-CAs3NvrB0 is somebody --- not kcp itself --- creating some objects in the compute cluster.

docs/poc2023q1/outline.md
MikeSpreitzer: I do not understand why there is a question at this point in the doc regarding edge clusters.

docs/poc2023q1/outline.md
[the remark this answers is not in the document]
MikeSpreitzer: Yes, both the syncer and the bootstrapper interact with the edge cluster, as described here and shown in the drawing.

docs/poc2023q1/outline.md
MikeSpreitzer: Yes, in this PoC the full status from the edge gets propagated to the center.  That is not feasible in general, it is a choice made for this PoC.  The whole point of also doing sumarization is to demonstrate part of the general answer to the obvious scale problem here.  This PoC is explicitly limited in scale and will not go beyond what can be done with a single kcp server.

docs/poc2023q1/outline.md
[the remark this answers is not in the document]
ezrasilvera: I'm just wondering if such bootstrapping shouldn't be done by completely external management framework. For example the one that is configuring the edge cluster itself?

docs/poc2023q1/outline.md
ezrasilvera: Doesn't a workspace reflects an API entry point and at the end an apiserver instance?  I was talking about throughput of api calls (but maybe I didn't get exactly what it is used for)

docs/poc2023q1/outline.md
ezrasilvera: Sounds reasonable. Maybe we should explicitly clarify it - i.e., where this summarization would be done eventually and why. Otherwise for people who are not familiar with the Kyst history it may not be clear how this is helping here.

docs/poc2023q1/outline.md
ezrasilvera: +1 for removing the bootstrapping controllers out of PoC1. I think that a simple automation (script based) which will be good enough. In any case for the PoC we also need to create the actual pClusters and configure them , so we  might as well run some script that run the `sync` command, and then apply the resulting yaml on the target pCluster. 
Maybe we can put the bootstrap controller as screeched goal ?

docs/poc2023q1/outline.md
MikeSpreitzer: Perhaps what I meant by "that SyncTarget's workspace" is not clear.  It is intended to refer to the kcp workspace that contains the SyncTarget object; this is decidedly distinct from the edge cluster described by that SyncTarget.  The original comment here is attached to text talking about that workspace, not the edge cluster.

docs/poc2023q1/outline.md
MikeSpreitzer: Ah yes, this workspace has a view whose throughput (and data content volume) goes up with usage.

The scope of this PoC does not include sharding for scale.  In this PoC, there will be a single edge service provider workspace, and we will live with the scale implications of that.

docs/poc2023q1/outline.md
MikeSpreitzer: I do not think this PoC should describe a more ambitious design.
The text already excludes general hierarchy and excludes making the implementation support a large number of edge clusters.

MikeSpreitzer:
Based on recent discussions, I intend to revise this outline to exclude bootstrapping from the focus of this PoC.

MikeSpreitzer:
Based on recent considerations, this design needs some significant revision to support workload that has objects that modify apiserver behavior.  The problem is that, currently, no kcp workspace is a fully de-natured apiservr that can simply serve as a container for a workload description; a workspace also contains objects that modify the behavior of that workspace --- and we positively want that in some cases (e.g., CRDs, because they enable creation of corresponding CRs, and _some_ RBAC) and NOT in other cases (e.g., RBAC intended to modify behavior of the edge apiservers).

docs/poc2023q1/outline.md
MikeSpreitzer: I tried to make this scope limitation clearer in the latest revision.

docs/poc2023q1/outline.md
MikeSpreitzer: In the latest revision I simply excluded bootstrapping from the concepts intended to be proven.  Our demo will do _something_ but it will not be something that we advocate for general usage.

MikeSpreitzer:
Eek, there is another big gap regarding how to "contain" certain kinds of workload objects.  Specifically, those that are --- in the edge cluster --- stored in a custom external apiserver (this is one of the two uses of the term "aggregated", the other being objects whose type is defined by a CRD).  I had hoped we could support workloads that have custom apiservers.  But that begs the question of where in the center the storage happens.

waltforme:
> Based on recent considerations, this design needs some significant revision to support workload that has objects that modify apiserver behavior and objects that are cluster-scoped (AKA non-namespaced) --- recognizing that there is some overlap of these categories. For the first category the problem is that, currently, no kcp workspace is a fully de-natured apiserver that can simply serve as a container for a workload description; a workspace also contains objects that modify the behavior of that workspace --- and we positively want that in some cases (e.g., CRDs, because they enable creation of corresponding CRs, and _some_ RBAC) and NOT in other cases (e.g., RBAC intended to modify behavior of the edge apiservers). For the second category the problem is that we need an implementation (syncer, whatever) that can handle the object types (resources) that are configured dynamically in each EdgePlacement object.

During today's kcp community call I was thinking about another quick work around to this issue, which is to have a sandbox workspace, whenever necessary, as a child of each mailbox workspace.

```
    └── mailbox-1
        └── sandbox
```
Anything we want (or don't care) them to alter the behavior of the control plane, we put them into the mailbox (e.g. CRDs).
Anything that we don't want them to alter the behavior of the control plane, we put them into the sandbox (e.g. cluster-scoped RBACs).

Will that work? @MikeSpreitzer

MikeSpreitzer:
@waltforme : the objects that modify apiserver behavior will modify the behavior of the sandbox workspace if there is such a thing.  We do not want a sandbox workspace with excess RBAC objects and we really do not want a sandbox workspace with excess admission control plugins.

waltforme:
@MikeSpreitzer OK. Thank you!

docs/poc2023q1/outline.md
```diff
@@ -0,0 +1,431 @@
...
+## Edge Scheduler
+
+This controller monitors the EdgePlacement, Location, and SyncTarget
+objects and maintains the results of matching.  For each EdgePlacement
+object this controller maintains an associated collection of
+SinglePlacementSlice objects holding the matches for that
+EdgePlacement.  These SinglePlacementSlice objects appear in the same
+workspace as the corresponding EdgePlacement; the remainder of how
+they are linked is TBD.
+
```
pdettori, line 340: It seems the Edge Scheduler is only maintaining the "Where" but not the "What". I think we need also to maintain the "What" by labeling namespaces selected by the Namespace selector (for namespaced objects) and non-namespaced objects selected by the non-namespaced object selector. Furthermore, a "resource" controller could be used to label objects within a labeled namespace.  Other controllers such as the placement translator should not need to re-evaluate the selector to find the "What".  
cc. @waltforme

MikeSpreitzer: Yes, the EdgeScheduler only evaluates the "where".
I doubt that we need to copy the chatty pattern from TMC; I think we can use fewer controllers with less annotating/labeling.

pdettori: My concern is more about separation of responsibility. The Edge Scheduler is evaluating Edge Placement policies to make decisions about what goes where, therefore it seems logical that it is up to the Edge Scheduler to tag selected objects with identifiers for the "where" that is specified in the SinglePlacementSlices. If not, which component is responsible for that ? The Placement Translator should not re-evaluate Edge Placement policies, it seems an overlap of responsibilities to me.

MikeSpreitzer: > the Edge Scheduler is only maintaining the "Where"
Correct.

> The Edge Scheduler is evaluating Edge Placement policies to make decisions about what goes where
No, as noted above.

We *could* change the division of responsibilities to make the thing called Edge Scheduler do more.  If we tagged workload namespaces and non-namespaced objects with references to the "where" resolutions, what would those tags and references look like?  Remember that the resolution of one "where" predicate can be multiple SinglePlacementSlice objects.

I think it is worth being explicit about _why_ we choose a particular division of responsibilities.  The current design exposes the resolution of the "where" predicate so that users can see that resolution, which is the result of munging together several things.

MikeSpreitzer: We _could_ define an interface between the full "what goes where" decision-making and execution using some sort of tagging along the following lines.  Let a tag be an annotation whose key is `"placement.edge.kcp.io/" + some_string_without_a_slash` and whose value is the stringified form of a label selector that selects the SinglePlacementSlice objects containing the where.

pdettori: we could find a way to use a label instead of an annotation so that we could do label based filtering on list and watch. But if we do not have the interface between the full "what goes where" decision-making and execution, can you add more details in the design about how the placement translator "maintains a copy of the workload" ? It seems like placement translator needs to watch the edge placement, find matching objects for SinglePlacementSlice objects and copy them in the mailboxes, and reconcile changes in the edge placement and objects in the workload workspace.

docs/poc2023q1/outline.md
```diff
@@ -0,0 +1,431 @@
...
+syncer that creates the appropriately independent objects in the edge
+cluster and itself tolerates intermittent connectivity.
+
+As further matters of roadmapping development of this PoC:
+customization may be omitted at first, and summarization may start
+with only a limited subset of the implicit functionality.
+
+This PoC builds on TMC and makes some compromises to accommodate that.
+The implementation involves workload components (syncers) writing
+status information to inventory objects (SyncTargets).
```
stevekuznetsov, line 59: Help me understand what aspect of  that is a compromise - simply that it's incompatible with routinely/intermittently offline edge clusters?

MikeSpreitzer: No, the compromise is the violation of the division into infrastructure and workload layers with the workload layer strictly consuming information from the infrastructure layer.  The SyncTarget objects come from the infrastructure layer, so it is not right for the workload layer (syncers) to update the SyncTarget objects.

docs/poc2023q1/outline.md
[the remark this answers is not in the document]
MikeSpreitzer: As noted elsewhere, I have removed bootstrapping from the list of things that this PoC takes a position on.

docs/poc2023q1/outline.md
MikeSpreitzer: To summarize discussions that have occurred elsewhere:

- we have tentatively decided to continue with the currently proposed division of responsibilities.
- I plan to draft internal interfaces between the parts of the placement translator.
- I have updated this doc to mention that division of the placement translator into four parts.

PR #147, opened 2023-01-26
:sparkles: Start drafting edge scheduler for 2023q1 PoC

MikeSpreitzer:
## Summary
This is the tiniest start on the edge schedule for the 2023q1 PoC.
Issue #136 is a problem for this code, as noted in the doc.

## Related issue(s)

Fixes #

PR #161, opened 2023-02-13
✨ Make edge scheduler informed about edgeplacements

waltforme:
## Summary
This is the first step to refactor the separated kcp placement controller towards the edge scheduler.

Running in the edge service provider workspace, the controller is now got informed about the EdgePlacement objects in an edge service consumer workspace, aka a workload management workspace. A short demo of that is described in `docs/poc2023q1/edge-scheduler.md`.

The edge scheduler incorporates the funcs `configForViewOfExport` and `isAPIExportReady` from https://github.com/kcp-dev/edge-mc/pull/147, credits to @MikeSpreitzer.

The edge scheduler is now in a 'minimal runnable' state, a lot of cleaning-ups, renames and reorgs will be done in next PRs.


waltforme:
Force-push to https://github.com/kcp-dev/edge-mc/commit/af397097d2387e0c9ab2f51a3f93bdc2e9f95e10 emphasizes the commit of kcp-dev/kcp we are currently using.

cmd/placement/cmd/placement.go
```diff
@@ -30,12 +33,17 @@ import (
...
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
+	kcpclientsetnoncluster "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 	kcpclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned/cluster"
 	kcpinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
 	kcpfeatures "github.com/kcp-dev/kcp/pkg/features"
 
 	placementoptions "github.com/kcp-dev/edge-mc/cmd/placement/options"
 	edgeclient "github.com/kcp-dev/edge-mc/pkg/client"
+	edgeclientset "github.com/kcp-dev/edge-mc/pkg/client/clientset/versioned/cluster"
```
MikeSpreitzer, line 45: You might want to put "cluster" in the name of this import, to help keep its meaning explicit.
Or you might not, there is so much verbosity here.

cmd/placement/cmd/placement.go
```diff
@@ -122,7 +130,24 @@ func Run(ctx context.Context, options *placementoptions.Options) error {
...
+	if err != nil {
+		logger.Error(err, "failed to create config for view of edge exports")
+		return err
+	}
+	edgeClusterClientset, err := edgeclientset.NewForConfig(edgeViewConfig)
+	if err != nil {
+		logger.Error(err, "failed to create clientset for view of edge exports")
+		return err
+	}
+	edgeSharedInformerFactory := edgeinformers.NewSharedInformerFactoryWithOptions(edgeClusterClientset, resyncPeriod)
```
MikeSpreitzer, line 148: You might want to put "cluster" in the name of this factory, to make its scope explicit.
Or you might not.

MikeSpreitzer, line 148: What good do the resyncs do?

docs/poc2023q1/edge-scheduler.md
```diff
@@ -0,0 +1,64 @@
...
+It starts from the kcp-scheduling-placement controller separated from kcp-dev/kcp, now being refactored towards the goal above.
+
+It works with [this snapshot](https://github.com/kcp-dev/kcp/tree/4506fdc064060b3fe82e1082533f9798b36ba7a5) of kcp.
+
+#### A short demo
+Start a kcp server with [this snapshot](https://github.com/kcp-dev/kcp/tree/4506fdc064060b3fe82e1082533f9798b36ba7a5).
+
+Point `$KUBECONFIG` to the admin kubeconfig for that kcp server.
+
+Workspace `root` is used as the edge service provider workspace.
```
MikeSpreitzer, line 13: I think we should _not_ use the root workspace as the edge service provider workspace.  See how it is done in https://github.com/kcp-dev/edge-mc/blob/main/docs/poc2023q1/mailbox-controller.md#try-it

pkg/reconciler/scheduling/placement/controller.go
```diff
@@ -84,6 +81,9 @@ func NewController(
 
 		placementLister:  placementInformer.Lister(),
 		placementIndexer: placementInformer.Informer().GetIndexer(),
+
+		edgePlacementLister:  edgePlacementInformer.Lister(),
+		edgePlacementIndexer: edgePlacementInformer.Informer().GetIndexer(),
```
MikeSpreitzer, line 86: As you can tell by the stuttering ("...Informer.Informer()"),"edgePlacementInformer" is _not_ actually an _informer_, it is something prior to that.

cmd/placement/cmd/placement.go
waltforme: Lots of verbosity. So another thing is done to distinguish this, I checked the imports and found that clientset is mostly 'clustered', so I put a suffix 'noncluster' on the single import of the 'nonclustered' clientset.

cmd/placement/cmd/placement.go
waltforme: Same as above.

waltforme: This question is so big and fundamental that I have to cite @sttts 's famous book to answer:
image

docs/poc2023q1/edge-scheduler.md
waltforme: OK will change this.

pkg/reconciler/scheduling/placement/controller.go
waltforme: Right, but let's to keep the style consistent with kcp's (so many) controllers? It works well.

waltforme:
Force-push to https://github.com/kcp-dev/edge-mc/commit/007d28e704a01be7ebf3eef69e54bd0925e8cc7e switches the edge service provider workspace from `root` to `root:edge`.

cmd/placement/cmd/placement.go
MikeSpreitzer: Wow, that's outdated.  It does not recognize that informers recover from comm gaps by notifying about everything that was missed as soon as comm is restored.

We can address this separately, it is not just part of adding EdgePlacement informers.

pdettori: Also, resync period is usually a global setting, it's a common practice to control re-enqueing of events (e.g. for error conditions)

docs/poc2023q1/edge-scheduler.md
```diff
@@ -0,0 +1,65 @@
+### Edge scheduler
+The edge scheduler monitors the EdgePlacement, Location, and SyncTarget objects and maintains the results of matching.
+
+It starts from the kcp-scheduling-placement controller separated from kcp-dev/kcp, now being refactored towards the goal above.
```
pdettori, line 4: Perhaps we could say "forked" instead of "separated"

docs/poc2023q1/edge-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+#### A short demo
+Start a kcp server with [this snapshot](https://github.com/kcp-dev/kcp/tree/4506fdc064060b3fe82e1082533f9798b36ba7a5).
+
+Point `$KUBECONFIG` to the admin kubeconfig for that kcp server.
+
+Workspace `root:edge` is used as the edge service provider workspace.
+```console
+kubectl ws root
+kubectl ws create edge --enter
+```
```
pdettori, line 17: I think we should create a new workspace to use as a service provider workspace.

docs/poc2023q1/edge-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+
+Create CRDs and expose APIs.
```
pdettori, line 19: Perhaps we could say "Install Edge Placement CRDs and API Exports"

docs/poc2023q1/edge-scheduler.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+User home workspace is used as the workload management workspace.
+```console
+kubectl ws \~
+```
+
+Bind APIs and create an EdgePlacement CR.
+```console
+kubectl apply -f config/imports/
```
pdettori, line 31: For this one we could use the command `kubectl kcp bind export`

pdettori (changes requested):
I was able to fetch your branch and run the simple scenario successfully. See my comments in the code.

cmd/placement/cmd/placement.go
waltforme: I somehow have the memory that resync is not related to connection between the informer and the apiserver, in the sense that resync _can_ happen and _should_ happen even indeed disconnected to apiserver. I refreshed my memory by revisiting the implementation of the DeltaFIFO and confirmed my memory. So, IMO, the fundamentals described in the book is still valid.

But! If my understanding is wrong, please do share more pointers with me, I am sincerely more than happy to learn more on the internals. Appreciated!

With these been said, I really don't have any strong opinion on what to do with the `resyncPeriod` in edge-mc.

So which way should we go?

docs/poc2023q1/edge-scheduler.md
MikeSpreitzer: I have a preference for showing exactly what is going on, as this text currently does.  I would rather see a note that the `kubectl apply` is equivalent to `kubectl kcp bind` than a replacement.

docs/poc2023q1/edge-scheduler.md
waltforme: It is changed to `root:edge`.

cmd/placement/cmd/placement.go
MikeSpreitzer: This is one of the style issues that we should have a general conversation on.

"resync" is not the same as "relist", but that cited doc predates the distinction.

"resync" is distinct from the requeuing to retry.

"resync" redoes notifications for objects that exist, not for objects that are absent.  So its ability to cover for bugs is limited.

cmd/placement/cmd/placement.go
```diff
@@ -134,18 +159,64 @@ func Run(ctx context.Context, options *placementoptions.Options) error {
...
+	}
+	apiExportClient := providerClient.ApisV1alpha1().APIExports()
+	logger := klog.FromContext(ctx)
+	var apiExport *apisv1alpha1.APIExport
+	for {
+		apiExport, err = apiExportClient.Get(ctx, exportName, metav1.GetOptions{})
+		if err != nil {
+			if k8sapierrors.IsNotFound(err) {
+				logger.Info("Pause because APIExport not found", "exportName", exportName)
+				time.Sleep(time.Second * 15)
```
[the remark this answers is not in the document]
waltforme: I think it is probably OK here. Because this waiting happens before we start the scheduler.

We need this APIExport as a prerequisite of the scheduler, because it provides us with the view of our custom resources.

Actually this waiting logic is incorporated from #147 , and I think it works well.

docs/poc2023q1/edge-scheduler.md
waltforme: Fixed in latest force-push.

docs/poc2023q1/edge-scheduler.md
waltforme: Hmm, I am not following here. We created `root:edge` as a service provider workspace. It is a new one, right?

docs/poc2023q1/edge-scheduler.md
waltforme: Fixed in latest push. Since we have two CRDs, I just say "Install CRDs".

cmd/placement/cmd/placement.go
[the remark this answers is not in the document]
MikeSpreitzer: Note that this pause is during the setup phase, it is not in a worker thread.

docs/poc2023q1/edge-scheduler.md
MikeSpreitzer: That's exactly what current line 13 says is happening here.

docs/poc2023q1/edge-scheduler.md
waltforme: There is some interesting thing here. Look like there is some bug with this command. Specifically, this command doesn't like the APIExport's name `edge.kcp.io`. I have two APIExports, where the only difference is their name. One is named `edge.kcp.io`, the other is named `edge`. Only the second can be bound.

I have the text from my terminal:
```
👉 edge-mc $ kc ws root:edge
Current workspace is "root:edge".
👉 edge-mc $ kc get apiexport
NAME          AGE
edge          7m12s
edge.kcp.io   16h
👉 edge-mc $ kc get apiexport edge.kcp.io -oyaml | yq .spec
identity:
  secretRef:
    name: edge.kcp.io
    namespace: kcp-system
latestResourceSchemas:
  - v230126-0a8bf97.edgeplacements.edge.kcp.io
  - v230207-d7d166d.singleplacementslice.edge.kcp.io
👉 edge-mc $ kc get apiexport edge -oyaml | yq .spec
identity:
  secretRef:
    name: edge
    namespace: kcp-system
latestResourceSchemas:
  - v230126-0a8bf97.edgeplacements.edge.kcp.io
  - v230207-d7d166d.singleplacementslice.edge.kcp.io
👉 edge-mc $ kc ws \~
👉 edge-mc $ kc kcp bind apiexport root:edge:edge.kcp.io
error: fully qualified reference to workspace where APIExport exists is required. The format is `<logical-cluster-name>:<apiexport>` or `<full>:<path>:<to>:<apiexport>`
👉 edge-mc $ kc kcp bind apiexport root:edge:edge
apibinding edge created. Waiting to successfully bind ...
edge created and bound.
👉 edge-mc $ 
```

But I don't know whether I should file this bug. Because I am using the `kubectl-kcp` binary built from the specific commit that edge-mc depends on. Didn't try the latest `kubectl-kcp` binary.

pkg/reconciler/scheduling/placement/controller.go
MikeSpreitzer: The style in the kcp code is not self-consistent, and does not work well in my opinion --- it does not always make it clear what is an all-cluster object vs what is a cluster-specific object.

I can tolerate not stuffing all relevant facts into variable names.  We do not want each variable name to be an essay.  But since kcp introduces the concept of all-cluster objects but also retains cluster-specific objects, it is an issue that is of particular sensitivity.

cmd/placement/cmd/placement.go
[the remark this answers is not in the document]
pdettori: Even if it is in the setup phase it should be tolerant of the APIExport not being available right away. How should we handle the case where the APIExport is added after the controller is started ? Should we let the controller crash so that the pod where is it running is restarted by Kube ?

docs/poc2023q1/edge-scheduler.md
pdettori: Sorry, I misinterpreted here.

cmd/placement/cmd/placement.go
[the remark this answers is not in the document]
MikeSpreitzer: This retry loop *is* handling the case where the APIExport is added after the controller is started (along with other transient problems).

Kubernetes recognizes that a pod may take some startup time before it enters into regular service.  We could add a `/livez` handler to expose that.  In a later PR.

docs/poc2023q1/edge-scheduler.md
MikeSpreitzer: I would test with the latest code before opening an issue.

You could add a note of equivalency with the command that works, without getting into the issue of the command that does not work.

MikeSpreitzer (approved):
/lgtm

docs/poc2023q1/edge-scheduler.md
waltforme: Thanks for the suggestion @MikeSpreitzer ! Actually can I do the test with the latest code and open an issue if the bug is still there? Wanted to grasp this small opportunity to get more involved with kcp-dev/kcp.

waltforme: Hmm looks there is https://github.com/kcp-dev/kcp/issues/2773 already opened a few days ago.

pkg/reconciler/scheduling/placement/controller.go
waltforme: > We should probably have a general discussion of this and other style issues.


Yes, I agree, it is good to bring everyone to the same page in this early stage.

Issue #174, opened 2023-02-20
Incorrect plural forms in manifests for SinglePlacementSlice


waltforme:
The plural form names are incorrect for SinglePlacementSlice's CRD and APIResourceSchema. They should be `singleplacementslices` but the ending s is missing.

As a result, accessing SinglePlacementSlice resources returns an error, since by design a client uses plural forms in the http requests.

The error exists for both kubectl:
```
👉 edge-mc $ kubectl get singleplacementslices
error: the server doesn't have a resource type "singleplacementslices"
👉 edge-mc $ kubectl get singleplacementslice
No resources found
👉 edge-mc $ kubectl get edgeplacements
NAME     AGE
test-1   28s
👉 edge-mc $ kubectl get edgeplacement
NAME     AGE
test-1   29s
```
and for programmatic access:
```
E0220 08:07:12.197015   38247 controller.go:337] "failed creating singlePlacementSlice" err="the server could not find the requested resource (post singleplacementslices.edge.kcp.io)" reconciler="edge-scheduler" key="kvdk2spgmbix|hardcoded" edgeplacement.workspace="kvdk2spgmbix" edgeplacement.namespace="" edgeplacement.name="hardcoded" edgeplacement.apiVersion=""
```

Note that EdgePlacement doesn't have this issue (see the kubectl output above).

The cause should _not_ be from the English word 'slice' because EndpointSlice has a working plural form name:
```
ubuntu@ip-172-31-30-2:~$ kc get endpointslices
NAME         ADDRESSTYPE   PORTS   ENDPOINTS     AGE
kubernetes   IPv4          6443    172.31.30.2   163d
ubuntu@ip-172-31-30-2:~$ kc get endpointslice
NAME         ADDRESSTYPE   PORTS   ENDPOINTS     AGE
kubernetes   IPv4          6443    172.31.30.2   163d
```

MikeSpreitzer:
FYI, the singular form works because it equals the Kind modulo case.

```
(base) mspreitz@mjs12 edge-mc % kubectl get EdGepLacemEnt
No resources found
```

MikeSpreitzer:
It looks to me like the plural form _is_ specified in the CRD --- it is just specified to have the wrong value.

https://github.com/kcp-dev/edge-mc/blob/b92aecb78a43bddacde8f32904563ab1c5217dd0/config/crds/edge.kcp.io_singleplacementslice.yaml#L14

MikeSpreitzer:
[...]

PR #175, opened 2023-02-20
🐛 Manually add plural form names for SinglePlacementSlice

waltforme:
## Summary
This is a _temporary_ fix to #174.

With the plural form names _manually_ written to SinglePlacementSlice's CRD and APIResourceSchema, access to singleplacementslices are back to normal.

kubectl:
```
👉 edge-mc $ kc get singleplacementslices
NAME        AGE
hardcoded   104s
👉 edge-mc $ kc get singleplacementslice
NAME        AGE
hardcoded   106s
```

Programmatic:
```
I0220 15:15:31.493378   45159 controller.go:339] "created SinglePlacementSlice" reconciler="edge-scheduler" key="kvdk2spgmbix|test-1" edgeplacement.workspace="kvdk2spgmbix" edgeplacement.namespace="" edgeplacement.name="test-1" edgeplacement.apiVersion="" name="hardcoded" workspace="kvdk2spgmbix"
```

This PR is only a temporary fix to #174, because
- `make crds` reverts all changes here in this PR;
-  file**names** (not file contents) for the CRD and the APIResourceSchema still use singular form.

### DIFF
--- a/cmd/placement/cmd/placement.go
+++ b/cmd/placement/cmd/placement.go
@@ -140,12 +140,12 @@ func Run(ctx context.Context, options *placementoptions.Options) error {
 		logger.Error(err, "failed to create config for view of edge exports")
 		return err
 	}
-	edgeClusterClientset, err := edgeclientset.NewForConfig(edgeViewConfig)
+	edgeViewClusterClientset, err := edgeclientset.NewForConfig(edgeViewConfig)
 	if err != nil {
 		logger.Error(err, "failed to create clientset for view of edge exports")
 		return err
 	}
-	edgeSharedInformerFactory := edgeinformers.NewSharedInformerFactoryWithOptions(edgeClusterClientset, resyncPeriod)
+	edgeSharedInformerFactory := edgeinformers.NewSharedInformerFactoryWithOptions(edgeViewClusterClientset, resyncPeriod)
 
 	// create the edge-scheduler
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
I am not sure this is the right fix.
Why is https://github.com/kcp-dev/kcp/blob/c4a476cbd525a59b79bb0235d82ce0c2624702d7/config/crds/apis.kcp.io_apiexportendpointslices.yaml#L16 right?

waltforme:
> I am not sure this is the right fix.
> 
> Having something that is overwritten by regular operations is not good.
> 
> Why is https://github.com/kcp-dev/kcp/blob/c4a476cbd525a59b79bb0235d82ce0c2624702d7/config/crds/apis.kcp.io_apiexportendpointslices.yaml#L16 correct?

I guess everyone knows it is not the right fix and it is not good.

I think the question we should ask it, why our lines are wrong? But I don't know the answer. Otherwise I would fix it in the right way.

MikeSpreitzer:
Looks like you found the right fix in #178

ezrasilvera:
@waltforme So can this be closed after https://github.com/kcp-dev/edge-mc/pull/178 was merged?

waltforme:
> Looks like you found the right fix in #178

Yes, thanks for reviewing #178.

waltforme:
> @waltforme So can this be closed after #178 was merged?

Sure, let me do it.

PR #176, opened 2023-02-21
:sparkles: Add example watch-all commands

MikeSpreitzer:
## Summary
This PR adds three example programs that monitor object creations.

1. `kube-watchall` monitors a given ordinary Kubernetes "cluster" (or anything that acts like one --- e.g., a kcp workspace).  Monitors only object kinds that are defined at start time.
2. `kcp-watchall` monitors a kcp server or workspace.  Monitors only object kinds that are defined at start time.  Demonstrates some deficiencies (see main.go) of some facilities you might think to use.
3. `cluster-watchall` monitors the workspaces in a kcp server.  Uses a special informer on "API resources".  Most of the bulk of this PR comes from the introduction of the meta API for API resources.  To be clear, these meta objects are not actually served by any server --- so the generated client stubs are not needed.  I simply have not tried to factor the code generation so that they can be omitted.

## Related issue(s)

Fixes #


MikeSpreitzer:
The force-pushes pick lint wrt contextual logging.

ezrasilvera:
Are all the changes (especially in pkg) needed only for these examples or we think this is a needed general capability?  
Maybe it would be better to separate into two PRs (1) Add all the needed general pluming (e.g., clientset, etc..) (2) add the specific examples with their cmd and config (and hack)

MikeSpreitzer:
Right now this adds `pkg/apiwatch` and that is used only for one of these three example commands.

The bulk of this PR is not in `pkg/apiwatch` but in the code generated from the API object type declaration from APIResource.

ezrasilvera:
Got it. So you think  APIResource might be used also outside those examples ?

ezrasilvera:
I see several duplication across the examples. Can we maybe move some of theses (maybe the general utility ones) into some shared place.  For example verbsSupportInformers ?

MikeSpreitzer:
These are mostly offered as stand-alone examples, so I think sharing infrastructure is not ideal for such a situation.

`pkg/apiwatch` is different exactly because I have hypothesized that it may be useful to the 23q1 PoC.

cmd/kcp-watchall/main.go
```diff
@@ -0,0 +1,225 @@
...
+/* This program is a kcp client that uses (a) either an all-cluster
+   API discovery client or a cluster-focused one and (b) an
+   all-cluster dynamic client to list all objects of the kinds found
+   by API discovery.
+
+   Note that the all-cluster discovery client does not discover all
+   resources.
+
+   Note that the all-cluster dynamic client does not work for some
+   resources --- the ones found in cluster-focused but not all-client
```
waltforme, line 58: Is 'all-client' a typo?

MikeSpreitzer: Yes, thanks!  Fixed in latest push.

MikeSpreitzer:
The force-push to 19935a4 is a rebase onto `main`.

clubanderson:
@ezrasilvera can you please review this today (when you are back in office)?

MikeSpreitzer:
The force-push to 4314c7b is a rebase and squash.

clubanderson (approved):

Issue #155, opened 2023-01-30
bug: 2023q1 PoC mailbox workspace name problems


MikeSpreitzer:
### Describe the bug

The current construction of mailbox workspace names suffers from potential ambiguities and name length overflows.

### Steps To Reproduce

This should be clear from docs/2023q1/mailbox-controller.md

### Expected Behaviour

Not obvious problems.
Linkage that is not problematic for the controller code.
Linkage that is not problematic for users.

MikeSpreitzer:
The controller needs: (a) when notified about a Workspace object, to be able to put into the workqueue a value that holds effective references to both the Workspace and SyncTarget and (b) when notified about a SyncTarget object, to be able to put into the workqueue a value that holds effective references to both the Workspace and SyncTarget; an effective reference is one that can be used to construct a Get request.

For users, it is best if a `kubectl get Workspace` without any fiddly `-o` produces a listing that the user can follow to the corresponding SyncTarget objects.

The most user-friendly approach may be to impose a limit on the length of the name of an allowed SyncTarget.  Perhaps by adding an admission controller on the SyncTarget objects themselves, filtering/error-messaging downstream.

It would be possible to replace the SyncTarget name with its UID in the construction, and make the reference object in the workqueue hold both name and UID of the SyncTarget, with the Workspace object itself also holding the SyncTarget's UID.  This would remove the ambiguity and name length problems but be less easy for users to consume.

MikeSpreitzer:
@pdettori : from a `kubectl get` listing of mailbox workspaces, a user would like to easily find the corresponding SyncTarget for each mailbox workspace.  Finding that SyncTarget requires knowing two things: (1) the pathname of the workspace containing the SyncTarget, and (2) the name of the SyncTarget.  The current behavior does not address (1), simply using the logical cluster ID as the relevant identifier.  This issue is about (2), specifically the fact that the SyncTarget name can be so long that the existing way of deriving a mailbox workspace name can produce a name that is too long.

I am not aware of a good way to map a logical cluster ID to the corresponding workspace pathname in the cases where there _is_ a corresponding workspace (which I presume is all the cases for a logical cluster that contains a SyncTarget).

MikeSpreitzer:
The currently relevant limit on SyncTarget names is https://github.com/kcp-dev/kcp/blob/4506fdc064060b3fe82e1082533f9798b36ba7a5/config/crds/workload.kcp.io_synctargets.yaml#L57 .

[...]

PR #181, opened 2023-02-23
✨ Start drafting placement translator internal interfaces

MikeSpreitzer:
## Summary
This PR is the first draft of the internal interfaces among the parts of the placement translator process of the 2023q1 PoC.

This draft assumes that #155 will be resolved by switching to using the SyncTarget UID instead of its name in the construction of the mailbox workspace name.

## Related issue(s)

Fixes #

### DIFF
--- a/cmd/placement-translator/main.go
+++ b/cmd/placement-translator/main.go
@@ -46,13 +46,13 @@ import (
 	"k8s.io/klog/v2"
 	utilflag "k8s.io/kubernetes/pkg/util/flag"
 
-	edgeclusterclientset "github.com/kcp-dev/edge-mc/pkg/client/clientset/versioned/cluster"
-	edgeinformers "github.com/kcp-dev/edge-mc/pkg/client/informers/externalversions"
 	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
 	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
 	kcpscopedclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 	kcpinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
 
+	edgeclusterclientset "github.com/kcp-dev/edge-mc/pkg/client/clientset/versioned/cluster"
+	edgeinformers "github.com/kcp-dev/edge-mc/pkg/client/informers/externalversions"
 	"github.com/kcp-dev/edge-mc/pkg/placement"
 )
 
--- a/cmd/placement-translator/main.go
+++ b/cmd/placement-translator/main.go
@@ -1,0 +1,221 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+// Import of k8s.io/client-go/plugin/pkg/client/auth ensures
+// that all in-tree Kubernetes client auth plugins
+// (e.g. Azure, GCP, OIDC, etc.)  are available.
+//
+// Import of k8s.io/component-base/metrics/prometheus/clientgo
+// makes the k8s client library produce Prometheus metrics.
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/spf13/pflag"
+
+	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/server/mux"
+	"k8s.io/apiserver/pkg/server/routes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/component-base/metrics/legacyregistry"
+	_ "k8s.io/component-base/metrics/prometheus/clientgo"
+	"k8s.io/klog/v2"
+	utilflag "k8s.io/kubernetes/pkg/util/flag"
+
+	edgeclusterclientset "github.com/kcp-dev/edge-mc/pkg/client/clientset/versioned/cluster"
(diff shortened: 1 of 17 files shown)


cmd/placement-translator/main.go
```diff
@@ -0,0 +1,221 @@
...
+	inventoryConfig.UserAgent = "mailbox-controller"
+
+	// Get client config for view of SyncTarget objects
+	edgeViewConfig, err := configForViewOfExport(ctx, restConfig, "edge.kcp.io")
+	if err != nil {
+		logger.Error(err, "Failed to create client config for view of edge APIExport")
+		os.Exit(4)
+	}
+
+	edgelusterClientset, err := edgeclusterclientset.NewForConfig(edgeViewConfig)
```
waltforme, line 148: edge**C**lusterClientset?

pkg/placement/declarations.go
```diff
@@ -0,0 +1,387 @@
...
+// The implementation may atomize the resolved "where" using a SinglePlacementSliceSetReducer.
+// The implementation may use a BindingOrganizer to get from the atomized
+// "what" and "where" to the ProjectionMapProducer behavior.
+type SetBinder interface {
+	AsWhatConsumer() DynamicMapConsumer[edgeapi.ExternalName, WorkloadParts]
+	AsWhereConsumer() DynamicMapConsumer[edgeapi.ExternalName, ResolvedWhere]
+	ProjectionMapProducer
+}
+
+// WorkloadProjector is kept appraised of what goes where
```
waltforme, line 83: Is WorkloadProjector kept appraised of 'what goes where' or just 'what'?

pkg/placement/declarations.go
```diff
@@ -0,0 +1,387 @@
...
+// ProjectionKey identifies the topmost level of organization,
+// the combinatin of the destination and the API group and resource.
+type ProjectionKey struct {
+	metav1.GroupResource
+	Destination SinglePlacement
+}
+
+// ProjectionPerCluster is the second level of organization.
+// It identifies the API version to use currently and holds
+// the map provider that gets to the lowest level of organization.
```
waltforme, line 182: I found 'map provider' and 'map producer' are both used here and it seems that two terms refer to the same thing. We may want to be consistent for easier reading.

pkg/placement/declarations.go
```diff
@@ -0,0 +1,387 @@
...
+
+	// For non-namespaced objects, Names can optionally be non-nil to restrict
+	// the objects handled.
+	Names *k8ssets.String
+}
+
+// SinglePlacement extends the API struct with the UID of the SyncTarget.
+type SinglePlacement struct {
+	edgeapi.SinglePlacement
+	SyncTargetUID apimachtypes.UID
```
waltforme, line 209: We may need to change this because now #198 is merged.

It is just a kind reminder in case it is omitted, because it is easier to be forgotten comparing to other explicit usage of `ExternalName`.

pkg/placement/map-types.go
```diff
@@ -0,0 +1,80 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package placement
+
+// DynamicMapProducer holds a mutable map and keeps consumers appraised of it.
+// The zero value of Val signals a missing entry in the map.
+type DynamicMapProducer[Key comparable, Val any] interface {
```
waltforme, line 21: Question: It is 'dynamic' because it can produce values in continuous and incremental manner, right?

pkg/placement/map-types.go
```diff
@@ -0,0 +1,80 @@
...
+	// The producer precedes each consumer in the locking order.
+	AddConsumer(consumer DynamicMapConsumer[Key, Val], notifyCurrent bool)
+
+	// Get invokes the given function on the value corresponding to the key.
+	// This does not count as informing any particular consumer.
+	// The producer precedes the function in the locking order.
+	Get(Key, func(Val))
+}
+
+// DynamicMapProducerGet does a non-CPS Get.
```
waltforme, line 36: Question: What is 'non-CPS'?

pkg/placement/map-types.go
```diff
@@ -0,0 +1,80 @@
...
+	Set(Key, Val)
+}
+
+// DynamicMapConsumerFunc is a func value that implements DynamicMapConsumer.
+// Remember that func values are not comparable.
+type DynamicMapConsumerFunc[Key comparable, Val any] func(Key, Val)
+
+func (cf DynamicMapConsumerFunc[Key, Val]) Set(key Key, val Val) { cf(key, val) }
+
+type Client[T any] interface {
```
waltforme, line 71: How about `Consumer`? To resonate with other names.

waltforme (commented):
I'm learning from this PR which designs a generic event-driven data flow mechanism IMO.

pkg/placement/declarations.go
MikeSpreitzer: 'what goes where'

pkg/placement/declarations.go
MikeSpreitzer: Yes, I will rebase this PR.

pkg/placement/map-types.go
MikeSpreitzer: "continuation-passing style"

pkg/placement/map-types.go
MikeSpreitzer: A "Consumer" is generally invoked multiple times, a client only needs to be given its provider once.

MikeSpreitzer: I adjusted the terminology further in the latest push.

pkg/placement/declarations.go
MikeSpreitzer: In the latest push I tried to rationalize the terminology.

MikeSpreitzer:
The force-push to 5fb9e73 removes an obsolete reference UIDer.

pkg/placement/test_uider.go
```diff
@@ -51,7 +51,7 @@ func NewTestUIDer(rng *rand.Rand, wg *sync.WaitGroup) *testUIDer {
 	}
 }
 
-func (tu *testUIDer) AddConsumer(consumer DynamicMapConsumer[ExternalName, apimachtypes.UID], notifyCurrent bool) {
+func (tu *testUIDer) AddReceiver(consumer MappingReceiver[ExternalName, apimachtypes.UID], notifyCurrent bool) {
```
waltforme, line 54: The first parameter of `AddReceiver` seems to be passive in this situation because its `Set` is called by `tu`, not by itself. Also the doc of `MappingReceiver` says 'MappingReceiver is given map entries by a DynamicMapProvider.'

OTOH `tu` implements `DynamicMapProvider`, so by the updated naming convention, it passively provides the service of a DynamicMap.

So both `tu` and `consumer` are passive? I'm trying hard to follow.

MikeSpreitzer: I made another revision, hopefully it is clearer now.

cmd/placement-translator/main.go
MikeSpreitzer: Fixed in latest push.

PR #191, opened 2023-02-27
✨ Some refactorings of the edge-scheduler

waltforme:
## Summary
This PR introduces some refactorings of the edge scheduler.
- change (file, function, etc.) names that inherited from the TMC controller;
- forget about namespaces, since we decided that the edge scheduler doesn't evaluate the 'what' part of scheduling;
- forget about TMC placements;
- ensure a SinglePlacementSlice object for each EdgePlacement object.


ezrasilvera:
I'm a little confused about the name changes. In the code we switched from placement to scheduler, however in the package we moved from placement to edgeplacement under scheduling ?

pkg/reconciler/scheduling/edgeplacement/controller.go
```diff
@@ -183,68 +133,21 @@ func (c *controller) enqueuePlacement(obj interface{}) {
...
+	// 	return
+	// }
+	// clusterName, _, _, err := kcpcache.SplitMetaClusterNamespaceKey(key)
+	// if err != nil {
+	// 	runtime.HandleError(err)
+	// 	return
+	// }
+
+	// TODO: We may or may not need to enqueue edgeplacements because of changes from locations.
+	// That depends on our future design.
```
MikeSpreitzer, line 150: This process should react to Location changes as well as Placement changes.  Perhaps use two workqueues, one for each kind of object, or use one workqueue with object kind included in the object reference values put into the queue.  Or, as does the old code here, use just one queue that holds only references to placements and bluntly enqueue references to all that could matter.

pkg/reconciler/scheduling/edgeplacement/controller.go
```diff
@@ -305,7 +208,7 @@ func (c *controller) process(ctx context.Context, key string) error {
 	obj, err := c.edgePlacementLister.Cluster(clusterName).Get(name)
 	if err != nil {
 		if errors.IsNotFound(err) {
-			logger.Info("object deleted before handled")
+			logger.V(2).Info("object deleted before handled")
```
MikeSpreitzer, line 211: Is this logger already at an elevated V level or does this statement actually correspond with `-v=2` ?  I think that this particular message is too inconsequential to appear at `-v=2`; I would make it 3 or a higher number.

MikeSpreitzer (commented):
Looks mostly good.  A couple of comments inline.

pkg/reconciler/scheduling/edgeplacement/controller.go
waltforme: Changed the comment to documents this. Thanks for suggesting these approaches!

pkg/reconciler/scheduling/edgeplacement/controller.go
waltforme: Just checked the logger, its `.level` field equals 0 before calling `V`. Changed to 3.

MikeSpreitzer (approved):
/lgtm

PR #192, opened 2023-02-27
✨ added customizer.go and updated keys to reflect edge.kcp.io

### LINKED ISSUE #160: feature: Copy API for customization into edge-mc

clubanderson:
## Related issue(s)

Fixes #160


pkg/apis/edge/v1alpha1/customize.go
```diff
@@ -0,0 +1,80 @@
...
+const CustomizerAnnotationKey string = "edge.kcp.io/customizer"
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Customizer defines modifications to make to the relevant objects as they propagate
+// from center to edge.
+//
+// The relevant objects are those with an annotation whose key is
+// "edge.research.ibm.com/customizer" and whose value is the name of this object.
```
MikeSpreitzer, line 38: Domain name update needed here too.
Also, the "whose value is the name of this object" does not cover the case of non-namespaced workload objects, this should be fixed (not necessarily by copying text, referring to it would probably be better).

MikeSpreitzer (commented):
Mostly looks good.

One minor thing noted inline.

You also need to add the standard header to the `.go` file, and then `make crds codegen`.

MikeSpreitzer:
As I revisit this, a refinement occurs to me.  Currently a `Replacement` holds a string that can be anything in JavaScript Object Notation.  It might be convenient to add an alternative that directly and more simply supports the case where the user wants the replacement value to be a JSON string literal; in this case the `Replacement` would have a field that holds the string value, rather than the JSON for the string value.

pkg/apis/edge/v1alpha1/customize.go
clubanderson: updated comment, inserted compiler directives, included license text, and ran make crds codegen.

pkg/apis/edge/v1alpha1/customize.go
```diff
@@ -0,0 +1,100 @@
...
+// that applies to the annotated object as it propagates from center to edge.
+// When the annotated object is namespaced, the Customizer object to apply is in
+// the same namespace and the value of this annotation is the name of the Customizer object;
+// otherwise the value of this annotation uses the form "namespace/name" to reference
+// the desired Customizer.
+const CustomizerAnnotationKey string = "edge.kcp.io/customizer"
+
+// +crd
+// +genclient
+// +genclient:nonNamespaced
```
MikeSpreitzer, line 50: I had intended these to be namespaced.  The comments above are consistent with that (look at how the references work).

We _could_ make them non-namespaced, if you specifically think that would be better.

pkg/apis/edge/v1alpha1/customize.go
```diff
@@ -0,0 +1,100 @@
...
+// +kubebuilder:resource:scope=Cluster
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Customizer defines modifications to make to the relevant objects as they propagate
+// from center to edge.
+//
+// The relevant objects are those with an annotation whose key is
+// "edge.kcp.io/customizer" and whose value refers to this object as explained above.
+// Another Customizer may not be marked as relevant.
```
MikeSpreitzer, line 59: I do not remember why I put this comment here; I wonder if it is a leftover from an earlier revision?
It is now syntactically impossible for two Customizers to be relevant to one workload object.

clubanderson: removed "// Another Customizer may not be marked as relevant."

pkg/apis/edge/v1alpha1/customize.go
clubanderson: removed "// +genclient:nonNamespaced"

clubanderson:
@MikeSpreitzer and @ezrasilvera I have squashed - ready for your review.

pkg/apis/edge/v1alpha1/customize.go
```diff
@@ -0,0 +1,99 @@
...
+// that applies to the annotated object as it propagates from center to edge.
+// When the annotated object is namespaced, the Customizer object to apply is in
+// the same namespace and the value of this annotation is the name of the Customizer object;
+// otherwise the value of this annotation uses the form "namespace/name" to reference
+// the desired Customizer.
+const CustomizerAnnotationKey string = "edge.kcp.io/customizer"
+
+// +crd
+// +genclient
+// +kubebuilder:resource:scope=Cluster
```
MikeSpreitzer, line 50: For cluster-scoped, AKA non-namespaced, object types there are _two_ comments that are needed; for namespaced object types, omit both of them.  See https://github.com/kcp-dev/edge-mc/blob/57a28077f68d8bab083e465973b129589965aa65/pkg/apis/edge/v1alpha1/edge-placement.go#L43-L44 for an example that includes them.  I think you want to omit both here.

pkg/apis/edge/v1alpha1/customize.go
```diff
@@ -0,0 +1,99 @@
...
+// Such a replacement is conceptually done on the JSON representation of that object.
+type Replacement struct {
+	// `path` is a JSON Path identifying the part of the object to replace/inject.
+	Path string `json:"path"`
+
+	// `value` supplies the new value to put where the path points, in JSON.
+	Value string `json:"value"`
+}
+
+//+kubebuilder:object:root=true
```
MikeSpreitzer, line 88: I am not familiar with the comment.  https://book.kubebuilder.io/reference/markers/object.html sugggests, if I read it correctly, that this is a recommended alternative to line 91 below.  If this is right then we probably do not want both.  We probably want consistency among our sources (although that can be finished in a follow-on PR).  We should make all the comments that direct plugins consecutive.

pkg/apis/edge/v1alpha1/customize.go
clubanderson: done - thank you for explaining.  Removed the '...scope=Cluster' line for kubebuilder.

pkg/apis/edge/v1alpha1/customize.go
clubanderson: removed line with "//+kubebuilder:object:root=true"

clubanderson: will test through linter

clubanderson: lint completed without error

clubanderson:
@MikeSpreitzer corrections made - ready for your review

clubanderson:
@MikeSpreitzer corrections made - ready for your review

hack/logcheck.out
```diff
@@ -1 +1,9 @@
-/cmd/mailbox-controller/main.go:98:3: Key positional arguments are expected to be inlined constant strings. Please replace &{flg Name} provided with string value.
+/cmd/mailbox-controller/main.go:102:3: Key positional arguments are expected to be inlined constant strings. Please replace &{flg Name} provided with string value.
+/pkg/client/listers/edge/v1alpha1/customizer.go:112:2: undeclared name: CustomizerNamespaceListerExpansion
```
MikeSpreitzer, line 2: This looks bad.

pkg/apis/edge/v1alpha1/customize.go
```diff
@@ -0,0 +1,98 @@
...
```
MikeSpreitzer, line 87: It's still here.

hack/logcheck.out
clubanderson: how do I fix this?  its in the mailbox-controller

pkg/apis/edge/v1alpha1/customize.go
clubanderson: I forgot to save before commit.  I fixed it - waiting for comment on mailbox-controller issue in hack/logcheck.out before pushing again.

hack/logcheck.out
clubanderson: and how to fix "/pkg/client/listers/edge/v1alpha1/customizer.go:112:2: undeclared name: CustomizerNamespaceListerExpansion"?

MikeSpreitzer: The complaint about  /cmd/mailbox-controller/main.go is a matter of the linter being too rigid; there was already an exception in logcheck.out for this, this PR is only updating the line number for that.  (Because of the way this linter is automated, differences in line numbers do not cause CI failures but _do_ show up when you update the logcheck.out file.)

The real problem here is the "undeclared name: CustomizerNamespaceListerExpansion".  The "logcheck: analysis skipped due to errors in package" are consequences of the undeclared name.

clubanderson:
The logcheck.out looks like this (with error):
```
/cmd/mailbox-controller/main.go:102:3: Key positional arguments are expected to be inlined constant strings. Please replace &{flg Name} provided with string value.
/pkg/client/listers/edge/v1alpha1/customizer.go:112:2: undeclared name: CustomizerNamespaceListerExpansion
logcheck: analysis skipped due to errors in package
logcheck: analysis skipped due to errors in package
logcheck: analysis skipped due to errors in package
logcheck: analysis skipped due to errors in package
logcheck: analysis skipped due to errors in package
logcheck: analysis skipped due to errors in package
logcheck: analysis skipped due to errors in package

```

The error we need help with is "/pkg/client/listers/edge/v1alpha1/customizer.go:112:2: undeclared name: CustomizerNamespaceListerExpansion"

clubanderson:
per Andy G. - generators will not overwrite the expansion files once they’ve been generated. You need to remove it, then regenerate  deleted:    pkg/client/clientset/versioned/typed/apis/v1alpha1/apis_client.go
        deleted:    pkg/client/clientset/versioned/typed/apis/v1alpha1/doc.go
        deleted:    pkg/client/clientset/versioned/typed/apis/v1alpha1/fake/doc.go
        deleted:    pkg/client/clientset/versioned/typed/apis/v1alpha1/fake/fake_apis_client.go
        deleted:    pkg/client/clientset/versioned/typed/apis/v1alpha1/generated_expansion.go
        modified:   pkg/client/listers/edge/v1alpha1/customizer_expansion.go

and then re-ran 'make crds codegen' and then 'make update-contextual-logging'

pkg/apis/edge/v1alpha1/customize.go
MikeSpreitzer, line 87: Still this line is here.

clubanderson: updated and pushed. I reverted during testing of the genclient process-of-elimination testing.  It is removed now.

hack/logcheck.out
clubanderson: fixed

clubanderson:
ready for review

MikeSpreitzer (approved):
/LGTM

PR #194, opened 2023-02-28
:sparkles: Updating edge-mc kcp dependence

### LINKED ISSUE #183: feature: update edge-mc to depend on v0.11.0 of kcp
### Feature Description

The edge-mc repo currently [depends](../blob/main/go.mod) on [commit 4506fdc06406 of the kcp repo](https://github.com/kcp-dev/kcp/tree/4506fdc06406).  Now that there is a later release, the edge-mc repo should be updated to depend on that.

### Proposed Solution

We can do it.

### Alternative Solutions

We could wait for a later release.

### Want to contribute?

### Additional Context

https://github.com/kcp-dev/kcp/releases/tag/v0.11.0 .  Note that the changes listed there are wrt v0.10.0, not commit 4506fdc06406.

mra-ruiz:
## Summary
Updating the kcp dependence from v0.10.1-0.20230119143945-4506fdc06406 to v0.11.0

## Related issue(s)
Fixes #183


MikeSpreitzer:
Shouldn't the required version of `github.com/kcp-dev/kcp/pkg/apis` be updated too?

mra-ruiz:
> github.com/kcp-dev/kcp/pkg/apis

Hi, yes you're right! I just updated the version for `github.com/kcp-dev/kcp/pkg/apis` to `v0.0.0-00010101000000-000000000000` but I am now seeing this issue when I run `go mod tidy`: 

`github.com/kcp-dev/kcp/pkg/apis@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000`

I can't seem to find v0.0.0 of package `github.com/kcp-dev/kcp/pkg/apis`, not even in the [Go site to see more about packages](https://pkg.go.dev/github.com/kcp-dev/kcp/pkg/apis?tab=versions). The previous version for that package was version `v0.10.1-0.20230123155139-7b3126cbf96c` which I last saw in [this commit](https://github.com/kcp-dev/kcp/blob/d95b92ecf030897b358b93ff0e3608e08a52f456/cmd/apigen/go.mod). I wonder if maybe I just can't see the commit associated with v0.0.0 or if v.0.0.0 was a typo from a previous commit? Although, the latter would be unlikely since the codebase would most likely be broken. Thoughts or ideas?

MikeSpreitzer:
`v0.0.0-00010101000000-000000000000` means "nothing to see here, move along".  That works in kcp's `go.mod` because of the later `replace` directive for the same module.  We can not do the same thing in edge-mc for that module because edge-mc's `go.mod` does not have a `replace` directive for that module.

Look at how the current go.mod uses the same version for kcp and kcp/pkg/apis:

https://github.com/kcp-dev/edge-mc/blob/a1d3beb7e5f847f2c5762c42528801af040a5993/go.mod#L8-L9

You want to keep doing the same thing.

go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	k8s.io/code-generator v0.24.3
 	k8s.io/component-base v0.24.3
 	k8s.io/klog/v2 v2.70.1
-	k8s.io/kube-openapi v0.0.0-20220328201542-3ee0da9b0b42
+	k8s.io/kube-openapi v0.0.0-20220328201542-3ee0da9b0b42 // indirect
```
MikeSpreitzer, line 18: The addition of the `// indirect` notation is correct.  This is distinct from the rebase to v0.11.0; this notation addition is needed even without the rebase.

The convention seems to be that direct requirements are listed in the first r`equire` block and indirect ones are listed in the second `require` block.

MikeSpreitzer:
I tested the mailbox controller and edge scheduler according to their docs in docs/poc2023q1, and they worked as expected.

MikeSpreitzer (approved):
/LGTM
Thanks!

PR #198, opened 2023-03-03
:warning: Revise and extend SinglePlacement

MikeSpreitzer:
## Summary
This PR does two things.

1. Break apart the Location reference, because it is more trouble than it is worth; the SyncTarget reference is more commonly needed.
2. Add the SyncTarget UID, because making consumers find it for themselves would require them to be more privileged.

## Related issue(s)

Fixes #


waltforme (commented):
/lgtm

pkg/apis/edge/v1alpha1/single-placement.go
```diff
@@ -51,17 +52,16 @@ type SinglePlacementSlice struct {
 
 // SinglePlacement describes one Location that matches the relevant EdgePlacement.
 type SinglePlacement struct {
-	// `location` refers to the relevant Location.
-	Location ExternalName `json:"location"`
+	// Cluster is the logicacluster.Name of the logical cluster that contains
+	// both the Location and the SyncTarget.
+	Cluster string `json:"cluster"`
+
```
ezrasilvera, line 58: If Cluster contains the Location and the SyncTarget, can't we retrieve & save them using the Cluster reference? (instead of explicitly hold them in the SinglePlacement   ?

ezrasilvera:
/hold
Only a quick question - Can be merged after that is closed.

pkg/apis/edge/v1alpha1/single-placement.go
MikeSpreitzer: What system component do you mean by "we"?

One logical cluster can contain several Location&SyncTarget pairs, so I do not see how it is unnecessary to identify them here.

clubanderson:
@ezrasilvera did you get your answer?

pkg/apis/edge/v1alpha1/single-placement.go
MikeSpreitzer: The _cluster_ holds the SyncTarget and Location objects; the _cluster name_ does _not_ hold the _names_ of the SyncTarget and Location objects.

ezrasilvera:
> @ezrasilvera did you get your answer?

I'll close and approve when I will.

pkg/apis/edge/v1alpha1/single-placement.go
MikeSpreitzer: A client of a kcp server needs to provide a cluster name whenever accessing an API object.

PR #206, opened 2023-03-06
✨ Reshape the structure of edge scheduler

waltforme:
## Summary
This PR reshapes the structure of edge scheduler.

Initially the attempt is to
- within the edge scheduler, let one controller updates exactly one kind;
- within each controller, let one reconciler uses one queue and listens to exactly one kind.

Now the edge scheduler is reshaped into a 'consolidated' manner that
- there is only one controller;
- in the controller, a single queue is used for events from multiple kinds.


waltforme:
Force-push to https://github.com/kcp-dev/edge-mc/commit/b800e9c5340652b1d4bb0e89e9d8e39fac961828 is a rebase on latest main.

pkg/reconciler/scheduling/singleplacementslice/controller.go
```diff
@@ -0,0 +1,272 @@
...
+		DeleteFunc: c.enqueueForEdgePlacement,
+	})
+
+	return c, nil
+}
+
+// controller.
+type controller struct {
+	queue        workqueue.RateLimitingInterface
+	enqueueAfter func(*corev1.Namespace, time.Duration)
```
MikeSpreitzer, line 122: Why is this field here?  I do not see it used anywhere.  Also, `workqueue.RateLimitingInterface` has an `AddAfter` method.

pkg/reconciler/scheduling/singleplacementslice/controller.go
```diff
@@ -0,0 +1,272 @@
...
+		DeleteFunc: c.enqueueSinglePlacementSlice,
+	})
+
+	locationInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: c.enqueueForLocation,
+			UpdateFunc: func(old, obj interface{}) {
+				oldLoc := old.(*schedulingv1alpha1.Location)
+				newLoc := obj.(*schedulingv1alpha1.Location)
+				if !reflect.DeepEqual(oldLoc.Spec, newLoc.Spec) || !reflect.DeepEqual(oldLoc.Labels, newLoc.Labels) {
```
MikeSpreitzer, line 102: Are you aware of https://github.com/kubernetes/apimachinery/blob/release-1.24/pkg/api/equality/semantic.go#L29 ?  I see some Kubernetes code prefers to use that for comparing API object values.

MikeSpreitzer (commented):
I see two layers of multiplicity and do not see how they help.  There are two packages under `reconciler/scheduling`.  In `singleplacementslice`, `func (c *controller) reconcile` invokes two "reconcilers".

Assume that we continue with the limited aim of exactly one SinglePlacementSlice object per EdgePlacement object, I think the edge scheduler needs only one controller with one shared data structure and a work queue that carries typed object references.  The shared data structure would include the following items, and a `sync.Mutex` controlling access to them.

- a map from SyncTarget identity to set of Location identities (those whose `Spec.InstanceSelector` matches the SyncTarget).
- a map from Location identity to a set of records that include the following fields: (a) a set of SyncTarget identity, listing the ones that match the Location's `Spec.InstanceSelector`, and (b) a set of EdgePlacement identity, those whose `Spec.LocationSelectors` matches the Location.
- a map from EdgePlacement identity to set of Location identities, those that match the EdgePlacement's `Spec.LocationSelectors`.

The enqueue methods would enqueue object references that include an indication of object type.  BTW, remember that an object reference does not have to be a string; it could be a convenient struct, for example.  A queue holds general boxed values.

When processing a particular object of a particular type, it should be clear how to update the above shared data structure incrementally.

cmd/scheduler/cmd/scheduler.go
```diff
@@ -177,7 +177,7 @@ func Run(ctx context.Context, options *scheduleroptions.Options) error {
 	kubeSharedInformerFactory.WaitForCacheSync(ctx.Done())
 	kcpSharedInformerFactory.WaitForCacheSync(ctx.Done())
 	edgeSharedInformerFactory.WaitForCacheSync(ctx.Done())
-	c.Start(ctx, 1)
+	go es.Run(ctx, 1)
```
MikeSpreitzer, line 180: Is there a need to wait here for that `es.Run(..)` to finish, lest the overall `main` finish too soon?

pkg/scheduler/controller.go
```diff
@@ -49,13 +47,24 @@ const (
...
+
+const (
+	eventSourceEdgePlacement eventSource = "EdgePlacement"
+	eventSourceLocation      eventSource = "Location"
+	eventSourceSyncTarget    eventSource = "SyncTarget"
+)
+
+type queueItem struct {
+	key         string
+	eventSource eventSource
```
MikeSpreitzer, line 60: I think the most important thing about this field is that it indicates the kind of object being referenced.

pkg/scheduler/controller.go
```diff
@@ -173,82 +187,55 @@ func (c *controller) startWorker(ctx context.Context) {
...
 func (c *controller) processNextWorkItem(ctx context.Context) bool {
 	// Wait until there is a new item in the working queue
-	k, quit := c.queue.Get()
+	i, quit := c.queue.Get()
 	if quit {
 		return false
 	}
-	key := k.(string)
+	item := i.(queueItem)
+	key := item.key
```
MikeSpreitzer, line 195: As well as adding the rest of `item` to the logger, how about adding the object kind too?

waltforme:
Fixed some of the issues, still WIP.

pkg/scheduler/controller.go
waltforme: Makes sense, added in latest commits.

pkg/scheduler/controller.go
waltforme: Renamed this.

cmd/scheduler/cmd/scheduler.go
waltforme: Added waiting here.

pkg/reconciler/scheduling/edgeplacement/reconcile.go
```diff
@@ -37,7 +37,9 @@ type reconciler interface {
 
 func (c *controller) reconcile(ctx context.Context, edgePlacement *edgev1alpha1.EdgePlacement) error {
 	reconcilers := []reconciler{
-		&edgePlacementReconciler{},
+		&edgePlacementReconciler{
+			client: c.edgeClusterClient,
```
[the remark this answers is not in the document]
waltforme: I just got rid of the `reconciler` type because now we are writing the scheduler in the consolidated manner and don't need the type anymore.

cmd/scheduler/cmd/scheduler.go
```diff
@@ -177,7 +189,8 @@ func Run(ctx context.Context, options *scheduleroptions.Options) error {
 	kubeSharedInformerFactory.WaitForCacheSync(ctx.Done())
 	kcpSharedInformerFactory.WaitForCacheSync(ctx.Done())
 	edgeSharedInformerFactory.WaitForCacheSync(ctx.Done())
-	c.Start(ctx, 1)
+	go spsCtlr.Start(ctx, 1)
+	go epCtlr.Start(ctx, 1)
```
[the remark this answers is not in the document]
waltforme: Fixed.

cmd/scheduler/cmd/scheduler.go
```diff
@@ -159,25 +182,31 @@ func Run(ctx context.Context, options *scheduleroptions.Options) error {
...
+	kubeSharedInformerFactory.Start(doneCh)
+	kcpSharedInformerFactory.Start(doneCh)
+	edgeSharedInformerFactory.Start(doneCh)
+	workloadSharedInformerFactory.Start(doneCh)
+	kubeSharedInformerFactory.WaitForCacheSync(doneCh)
+	kcpSharedInformerFactory.WaitForCacheSync(doneCh)
+	edgeSharedInformerFactory.WaitForCacheSync(doneCh)
+	workloadSharedInformerFactory.WaitForCacheSync(doneCh)
+	es.Run(ctx, 1)
+	<-doneCh
```
MikeSpreitzer, line 208: Why two waits on the same condition?

waltforme: Oh indeed.
Now I would like to hear your opinion, which place is better for the waiting? Here or in the controller?

MikeSpreitzer: Controllers commonly have a "Run" method, although it is sometimes misnamed "Start".  I recommend following the common behavior pattern, call the method that returns when the context is closed.  Then no additional code is needed to wait for that close.

waltforme:
Force-push to https://github.com/kcp-dev/edge-mc/commit/d367603e73891bf02609dd3ee4c41e6c6c623478 is a rebase and squash.

cmd/scheduler/cmd/scheduler.go
waltforme: Followed the common pattern.

pkg/reconciler/scheduling/singleplacementslice/controller.go
waltforme: Cool, used this.

pkg/reconciler/scheduling/singleplacementslice/controller.go
waltforme: Removed.

pkg/reconciler/scheduling/singleplacementslice/controller.go
```diff
@@ -0,0 +1,272 @@
...
+
+const (
+	ControllerName = "scheduling-singleplacementslice-controller"
+)
+
+func NewController(
+	kcpClusterClient kcpclientset.ClusterInterface,
+	edgeClusterClient edgeclientset.ClusterInterface,
+	singlePlacementSliceInformer edgev1alpha1informers.SinglePlacementSliceClusterInformer,
+	edgePlacementInformer edgev1alpha1informers.EdgePlacementClusterInformer,
```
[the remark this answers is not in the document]
waltforme: Tried my best to find a good name for it. Sounds like the interfaces provide a set of different mechanism to access the resources, so I call them XXXAccess.

pkg/reconciler/scheduling/edgeplacement/controller.go
```diff
@@ -128,13 +124,13 @@ func (c *controller) enqueuePlacement(obj interface{}) {
 		return
 	}
 
-	logger := logging.WithQueueKey(logging.WithReconciler(klog.Background(), ControllerName), key)
+	logger := logging.WithQueueKey(klog.Background().WithValues("controller", ControllerName), key)
```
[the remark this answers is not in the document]
waltforme: Fixed.

pkg/scheduler/controller.go
```diff
@@ -0,0 +1,249 @@
...
+
+func NewController(
+	context context.Context,
+	kcpClusterClient kcpclientset.ClusterInterface,
+	edgeClusterClient edgeclientset.ClusterInterface,
+	edgePlacementAccess edgev1alpha1informers.EdgePlacementClusterInformer,
+	locationAccess schedulingv1alpha1informers.LocationClusterInformer,
+	syncTargetAccess workloadv1alpha1informers.SyncTargetClusterInformer,
+) (*controller, error) {
+	context = klog.NewContext(context, klog.Background().WithValues("controller", ControllerName))
```
MikeSpreitzer, line 88: What do you think of using `klog.FromContext(context)` instead of `klog.Background()` here?

pkg/scheduler/controller.go
```diff
@@ -0,0 +1,249 @@
...
+			}
+		},
+		DeleteFunc: c.enqueueSyncTarget,
+	})
+
+	return c, nil
+}
+
+func (c *controller) enqueueEdgePlacement(obj interface{}) {
+	key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(obj)
```
MikeSpreitzer, line 146: If there were an alternative that returns a struct rather than a string then the consumer would not have the expense of parsing and the complexity of handling parsing failures.  But, as far as I know, there is no such alternative.

pkg/scheduler/controller.go
```diff
@@ -0,0 +1,249 @@
...
+	defer logger.Info("shutting down controller")
+
+	for i := 0; i < numThreads; i++ {
+		go wait.UntilWithContext(c.context, c.startWorker, time.Second)
+	}
+
+	<-c.context.Done()
+}
+
+func (c *controller) startWorker(ctx context.Context) {
```
MikeSpreitzer, line 209: This does not "start" a worker, it "runs" a worker.

pkg/scheduler/controller.go
```diff
@@ -0,0 +1,249 @@
...
+func (c *controller) processNextWorkItem(ctx context.Context) bool {
+	// Wait until there is a new item in the working queue
+	i, quit := c.queue.Get()
+	if quit {
+		return false
+	}
+	item := i.(queueItem)
+	key := item.key
+
+	klog.FromContext(ctx).V(1).Info("processing key", "triggeringKind", item.triggeringKind, "key", key)
```
MikeSpreitzer, line 223: What do you think of the following pattern, which I have seen other controllers use?

At this point in the logic, first make a logger that logs more stuff and a corresponding context:
```
	ctx = klog.NewContext(ctx, klog.FromContext(ctx).WithValues("triggeringKind", item.triggeringKind, "key", key))
```

From that point on, the information already added to that logger does not have to be explicitly included in any downstream logging statements.  So, for example, we would see the following.

```
	klog.FromContext(ctx).V(1).Info("processing queueItem")
```

cmd/scheduler/cmd/scheduler.go
```diff
@@ -147,7 +147,30 @@ func Run(ctx context.Context, options *scheduleroptions.Options) error {
...
+	workloadConfig, _ := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		clientcmd.NewDefaultClientConfigLoadingRules(),
+		&clientcmd.ConfigOverrides{
+			Context: clientcmdapi.Context{
+				Cluster:  "root",
+				AuthInfo: "kcp-admin",
+			},
+		},
+	).ClientConfig()
+	workloadViewConfig, err := configForViewOfExport(ctx, workloadConfig, "workload.kcp.io")
```
MikeSpreitzer, line 161: BTW, `edgeclient.ConfigForScheduling` used above is a hardcoded version of `configForViewOfExport`.  I think it is better to consistently use the less hardcoded version.

MikeSpreitzer (commented):
I think the scheduler will need three APIExport views:
1. A view of `edge.kcp.io` from the edge service provider workspace for both EdgePlacement and SinglePlacementSlice objects.
2. A view of `scheduling.kcp.io` from the infrastructure service provider workspace for Location objects.
3. A view of `workload.kcp.io` from the infrastructure service provider workspace for SyncTarget objects.

We may, for now, take `root` as the infrastructure service provider workspace.

MikeSpreitzer:
Note also that there is no longer a promise that you can extract a `logicalcluster.Path` from any object.  There only general promise is `logicalcluster.From(obj) Name`.  And that's all you need, because every Name is a Path.

pkg/scheduler/controller.go
waltforme: That makes sense, I tweaked it by https://github.com/kcp-dev/edge-mc/pull/206/commits/5a5cd2a9e9c74a3fd56d2120f37985877aedd50b.

pkg/scheduler/controller.go
waltforme: Fixed in https://github.com/kcp-dev/edge-mc/pull/206/commits/59fdfc24abc59e788a7523de8ea557f9839dd4cd.

pkg/scheduler/controller.go
waltforme: Thanks for the details! Followed the pattern by https://github.com/kcp-dev/edge-mc/pull/206/commits/5a5cd2a9e9c74a3fd56d2120f37985877aedd50b.

cmd/scheduler/cmd/scheduler.go
waltforme: Agree, fixed by https://github.com/kcp-dev/edge-mc/pull/206/commits/aff3b307a2c7c042d5b2253341551972764158cc.

pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (c *controller) reconcileOnEdgePlacement(ctx context.Context, key string) error {
+	logger := klog.FromContext(ctx)
+
+	ws, _, name, err := kcpcache.SplitMetaClusterNamespaceKey(key)
+	if err != nil {
+		logger.Error(err, "invalid key")
+		return err
+	}
+	logger.Info("reconciling", "name", name, "workspace", ws)
```
MikeSpreitzer, line 38: What do you think of doing the following instead?

```
logger = logger.WithValues("workspace", ws, "edgePlacement", name)
ctx = klog.NewContext(ctx, logger)
logger.Info("Reconciling")
```
(and a lot of other log statements get shorter too, and other ones not in this file get more informative)
?

pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -0,0 +1,76 @@
...
+	// TODO(waltforme): should I use a client to bother the apiserver or use local store?
+	ep, err := c.edgeClusterClient.EdgeV1alpha1().EdgePlacements().Cluster(ws.Path()).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	_, err = c.edgeClusterClient.EdgeV1alpha1().SinglePlacementSlices().Cluster(ws.Path()).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logger.V(2).Info("creating SinglePlacementSlice", "workspace", ws.String(), "name", name)
```
MikeSpreitzer, line 49: You might want to reconsider your V levels.  I think this creation is more interesting than `logger.Info("Reconciling")` and `klog.FromContext(ctx).V(1).Info("processing queueItem")`.

pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -0,0 +1,76 @@
...
+							APIVersion: edgev1alpha1.SchemeGroupVersion.String(),
+							Kind:       "EdgePlacement",
+							Name:       name,
+							UID:        ep.UID,
+						},
+					},
+				},
+				Destinations: []edgev1alpha1.SinglePlacement{},
+			}
+			_, err = c.edgeClusterClient.Cluster(ws.Path()).EdgeV1alpha1().SinglePlacementSlices().Create(ctx, sps, metav1.CreateOptions{})
```
MikeSpreitzer, line 64: I recommend setting the `FieldManager` in the various metav1.XXXOptions.

pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -0,0 +1,76 @@
...
+			if err != nil {
```
MikeSpreitzer, line 65: If `errors.IsAlreadyExists(err)` then you might want to take the win.

pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -0,0 +1,76 @@
...
+				logger.Error(err, "failed creating singlePlacementSlice", "workspace", ws.String(), "name", sps.Name)
+			} else {
+				logger.V(2).Info("created SinglePlacementSlice", "workspace", ws.String(), "name", sps.Name)
+			}
+		} else {
+			logger.Error(err, "failed getting SinglePlacementSlice for EdgePlacement", "workspace", ws.String(), "name", name)
```
MikeSpreitzer, line 71: A controller should not just give up when it gets what could be a transient error.

MikeSpreitzer (approved):
I think this needs more work; I left comments in-line.
I am going to approve this because it represents some progress and the remaining work can be done in follow-on PRs.

/lgtm

PR #217, opened 2023-03-08
🌱 Factor mailbox controller

MikeSpreitzer:
## Summary
This PR does the following things.
- Separate out an identifiable controller, with constructor taking any suitable pre-informers and clients.
- Rationalize some names.
- Also fix some oversights in the doc.

The main motivation is to make a clear boundary so that a behavioral unit test can be written.

## Related issue(s)

Fixes #

### DIFF
--- a/cmd/mailbox-controller/controller.go
+++ b/cmd/mailbox-controller/controller.go
@@ -1,0 +1,263 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	_ "k8s.io/component-base/metrics/prometheus/clientgo"
+	"k8s.io/klog/v2"
+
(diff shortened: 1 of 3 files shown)


cmd/mailbox-controller/controller.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		logger.Error(err, "Failed to lookup SyncTargets by mailbox workspace name", "mbwsName", mbwsName)
+		return false
+	}
+	var syncTarget *workloadv1alpha1.SyncTarget
+	if len(byIndex) == 0 {
+	} else {
+		syncTarget = byIndex[0].(*workloadv1alpha1.SyncTarget)
+		if len(byIndex) > 1 {
+			logger.Error(nil, "Impossible: more than one SyncTarget fetched from index; using the first", "mbwsName", mbwsName, "fetched", byIndex)
```
ezrasilvera, line 201: If this shouldn't happen, why not returning with error? Do we know that in this case the first one is at all valid ?

cmd/mailbox-controller/controller.go
```diff
@@ -0,0 +1,263 @@
...
+			if shutdown {
+				logger.V(2).Info("Queue shutdown")
+				return
+			}
+			ctl.sync1(ctx, ref)
+		}
+	}
+}
+
+func (ctl *mbCtl) sync1(ctx context.Context, ref any) {
```
ezrasilvera, line 167: Why can't we do the retry check and the enqueue in syncLoop ?

cmd/mailbox-controller/controller.go
```diff
@@ -0,0 +1,263 @@
...
+	defer ctl.queue.Done(ref)
+	logger := klog.FromContext(ctx)
+	logger.V(4).Info("Dequeued reference", "ref", ref)
+	retry := ctl.sync(ctx, ref)
+	if retry {
```
ezrasilvera, line 172: Is there a limit on retries ?

cmd/mailbox-controller/controller.go
MikeSpreitzer: There is a distinct function here so that `defer` can be used.

cmd/mailbox-controller/controller.go
MikeSpreitzer: Deliberately no.  Giving up would mean potentially not eventually recovering from a transient problem.

cmd/mailbox-controller/controller.go
MikeSpreitzer: The caller can not handle a general error, it only needs to know whether to retry.
We do not know which is the correct one to use when there are multiple, because there should never be multiple.

PR #224, opened 2023-03-09
📖 Document the roadmap thoughts, as they are now.

MikeSpreitzer:
## Summary
This PR documents the roadmap thoughts, such as they are at the moment, for the 2023q1 PoC.

This PR also removes one exclusion that we decided we no longer like: "Support for workload object types that are not either built into kcp or imported via a kcp APIBinding".

## Related issue(s)

Fixes #


clubanderson:
@MikeSpreitzer is this moving over to the /docs/content/en/docs location now?

MikeSpreitzer:
Yes, once #225 merges, I will then rebase this.

MikeSpreitzer:
The force-push to 60d2bcd is the promised rebase.  It takes the revised file contents originally proposed here and adds header material needed in a website source file.  Note that this overwrites changes that were made to the other copy without review.

PR #225, opened 2023-03-09
📖 Remove old copies of files copied to website source location

MikeSpreitzer:
## Summary
Some files were recently copied (with header addition) from elsewhere into the docs subdirectory that is the source for the website.  This PR removes the old copies.

## Related issue(s)

Fixes #


clubanderson:
@MikeSpreitzer we can add edge-scheduler and mailbox-controller also - they are also hosted in the docs site.

MikeSpreitzer:
I am not familiar with the rules for composing the header material.  I see a non-trivial difference between the two copies for the scheduler doc.  The original says

> The edge scheduler monitors the EdgePlacement, Location, and SyncTarget objects and maintains the results of matching.

while the copy for the website says

> {{% pageinfo %}}
> The edge scheduler monitors the EdgePlacement, Location, and SyncTarget objects and maintains the results of matching.
> It starts from the kcp-scheduling-placement controller forked from kcp-dev/kcp, now being refactored towards the goal above.
> It works with [this snapshot](https://github.com/kcp-dev/kcp/tree/4506fdc064060b3fe82e1082533f9798b36ba7a5) of kcp.
> {{% /pageinfo %}}

I am not sure what is the right way to resolve this.  Which content is newer?  Or is there some other reason for the difference?

/cc @waltforme

clubanderson:
no rules, just aesthetics.  Whatever you decide to do in the 'page info' block will show up inside a box on the page.  You can obviate this altogether if you wish.  On some pages I have included this to draw attention to other content or warn readers of the tentative nature of the content.  Other pages do not need this treatment.  If you decide to remove it from any given page, that is your choice as it's author.  I took some creative license when bringing the docs over originally.

waltforme:
> I am not familiar with the rules for composing the header material. I see a non-trivial difference between the two copies for the scheduler doc. The original says
> 
> > The edge scheduler monitors the EdgePlacement, Location, and SyncTarget objects and maintains the results of matching.
> 
> while the copy for the website says
> 
> > {{% pageinfo %}}
> > The edge scheduler monitors the EdgePlacement, Location, and SyncTarget objects and maintains the results of matching.
> > It starts from the kcp-scheduling-placement controller forked from kcp-dev/kcp, now being refactored towards the goal above.
> > It works with [this snapshot](https://github.com/kcp-dev/kcp/tree/4506fdc064060b3fe82e1082533f9798b36ba7a5) of kcp.
> > {{% /pageinfo %}}
> 
> I am not sure what is the right way to resolve this. Which content is newer? Or is there some other reason for the difference?
> 
> /cc @waltforme

The first one is newer. The second seems to be generated from an old version.

MikeSpreitzer:
The last two commits converge the two copies of each of the two component exercise docs.

ezrasilvera:
Overall it looks OK. One general comment - Do we really need to include log output in these files?  The log includes code line numbers which mean that every update to the files the messages may be different even if the overall behavior and instructions are still the same.

MikeSpreitzer:
I think we need to give the reader an idea of what good log output looks like, at the current stage of development.  I expect readers understand that line numbers can change with every edit, and we do not have to worry about keeping them exactly current.

clubanderson (approved):

PR #232, opened 2023-03-13
📖 EMC syncer developers do not know all the kinds of objects

MikeSpreitzer:
## Summary
This PR extends the list of distinctive features of the EMC syncer to note that the kinds of objects to sync are not all known at development time.

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
+++ b/docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
@@ -607,6 +607,7 @@ ways.
 
 - Create self-sufficient edge clusters.
 - Re-nature objects that edge-mc forcibly denatures at the center.
+- The kinds of objects to be synced are not bounded at development time.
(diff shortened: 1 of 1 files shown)


pdettori:
Doesn't the current TMC syncer already offer the ability to configure the resources to sync at deployment time ? If this is the case, why do we need to make this distinction here ?

MikeSpreitzer:
I realize that the TMC syncer is configured with a set of resources to sync.  The question is whether a user can configure the syncing of a resource that the developers did not prepare the syncer for.  Do you know?

pdettori:
My understanding is that they have mutators for deployment, and perhaps pods. But in general you should be able to configure resources to sync at deployment time for namespaced resources. For non-namespaced resources, it's a different story, as those currently are supported writing code.

MikeSpreitzer:
I got testimony today from people who got the TMC syncer to sync custom resources not known to the TMC developers.

PR #233, opened 2023-03-15
✨ First draft of API for identifying associated objects

MikeSpreitzer:
## Summary
This PR adds a proposed interface for specifying how associated objects are identified.

## Related issue(s)

Fixes #


pkg/apis/edge/v1alpha1/associate.go
```diff
@@ -0,0 +1,113 @@
...
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// AssociatorAnnotationKey is the key of an annotation used to point from a workload
+// object to an Associator that says how associated objects in edge clusters are identified.
+// For a namespaced object, the value of the annotation is the name of the Associator.
+// For a non-namespaced object, the value of the annotation is the namespace/name of the Associator.
+// When the annotation is absent, no associated objects are identified.
+const AssociatorAnnotationKey = "edge.kcp.io/associator"
+
```
pdettori, line 27: Have you considered how the current kcp TMC syncer up-syncing works, so that we can take advantages of features that it already provides? See for example https://github.com/kcp-dev/kcp/issues/1980

MikeSpreitzer: No, I do not understand the thoughts there.  I asked yesterday (https://kubernetes.slack.com/archives/C021U8WSAFK/p1678826090217939) and got no answer.  Thanks for the pointer.  I am also pursuing updates to https://github.com/kcp-dev/kcp/blob/e1353f4e70eecfcab0ddb124563f1ac8e0e0aca1/docs/content/concepts/placement-locations-and-scheduling.md

MikeSpreitzer:
I think #239 is the more likely winner here.  I will close this once I better preserve a pointer to small discussion above.

MikeSpreitzer:
The approach in #239 seems to be more favored.

PR #239, opened 2023-03-16
✨ Extend EdgePlacement with spec of upsync

MikeSpreitzer:
## Summary
This PR adds a proposed API for specifying upsync.
It is an alternative to #233.
The approach here is not based on associating upsynced objects with downsynced ones, but rather on identifying the objects to upsync based only on their own properties.  The approach here extends the `EdgePlacementSpec` with a way of identifying objects to upsync.

## Related issue(s)

Fixes #


yuji-watanabe-jp (approved):
This is great! Thanks! @MikeSpreitzer (cc: @yana1205 )

ezrasilvera:
@MikeSpreitzer I'm not sure I understand why the upSync is coupled with the placement objects. Maybe I'm missing something basic in the design here ..

yuji-watanabe-jp:
@ezrasilvera 

> so we define the upSync criteria per edgeplacement ?

upsync criteria specifies what resources could be coming from edge syncer to mailbox workspaces governed by the EdgePlacement, and so I think the coming resources should be handled according to the upsync criteria... e.g. make the resources available for the creator of EdgePlacement, summarizing the resources, maintaining dynamic indices of coming resources, etc...

So, overall, I think it might be good idea to define upsync criteria somewhere in edgeplacement.

MikeSpreitzer:
You could question several things independently.

One question is whether to simply present one interface in the center, or define a fundamental interface at the edge and invite a variety of ways to get the user's intent out to that interface at the edge.  As far as I can tell, the intent starts with central users.  We *could* define a fundamental interface at the edge, such as a convention for labeling objects there; that leaves the question of what puts those labels there.  You suggested doing this, making it the user's problem to put a controller at the edge to apply those labels.  In some cases that is easy.  For example, the author of a Deployment can specify labels that go on the associated Pod objects.  In other cases it is not built-in.  IIRC, the author of a Deployment does not have control over the labels on the associated ReplicaSet objects.  I chose, for simplicity, to specify an interface at the center and stop there, as far as defining interfaces for users to use is concerned.

Having determined to define *some* central interface, you could then question whether to do that by expanding EdgePlacement or defining something similar but independent.  For example, instead of expanding EdgePlacement to cover both downsync and upsync, we could leave EdgePlacement only concerned with downsync and introduce something else, perhaps called something like EdgeReturn, that is only concerned with specifying upsync.  I chose to expand EdgePlacement because I think the uses for upsync are closely related to the uses for downsync.  In the Compliance-to-Policy work, for example, there is a need both for downsync of policies and upsync of reports.  If we had two separate APIs, users with a use case like this would have to maintain two copies of the same "where" predicate.  Or we could refactor the API, so that there is a free-standing "where" predicate and then two kinds of binding objects, one referencing a "where" predicate and holding a "what to downsync" predicate and the other referencing a "where" predicate and holding a "what to upsync" predicate.  Even so, the user would have to take care to have two binding objects referencing the same "where" predicate object.  To me it seems more simple and direct to have a single binding kind of object that covers both downsync and upsync.  Of course, in use cases where there is not a natural association between downsyncs and upsyncs, the expanded EdgePlacement can have a "what to downsync" predicate that matches nothing or have a "what to upsync" predicate that matches nothing.

pkg/apis/edge/v1alpha1/edge-placement.go
```diff
@@ -102,6 +116,35 @@ type NonNamespacedObjectReferenceSet struct {
...
+	// An entry of `"*"` means that all match.
+	// Empty list means nothing matches (you probably do not want this
+	// for namespaced resources).
+	Namespaces []string `json:"namespaces,omitempty"`
+
+	// `Names` is a list of objects that match by name.
+	// An entry of `"*"` means that all match.
+	// Empty list means nothing matches (you probably never want an empty list).
+	Names []string `json:"names,omitempty"`
+}
```
pdettori, line 146: I think we will need additional criteria to select objects to upsync, such as label selector, and possibly a way to express selection of objects based on presence of ownerReferences to downsynced object.

config/crds/edge.kcp.io_edgeplacements.yaml
```diff
@@ -236,6 +241,47 @@ spec:
                   - resources
                   type: object
                 type: array
+              upsync:
+                description: '`upsync` identifies objects to upsync. An object matches
+                  `upsync` if and only if it matches at least one member of `upsync`.'
+                items:
+                  description: 'UpsyncSet specifies a set of objects, which may be
+                    namespaced or cluster-scoped, from one particular API group. An
```
clubanderson, line 249: we should be consistent wrt replacing 'cluster-scoped' with 'non-namespaced'

config/exports/apiresourceschema-edgeplacements.edge.kcp.io.yaml
```diff
@@ -16,18 +16,23 @@ spec:
...
-        _one_ of the selected Locations. \n Overlap between EdgePlacements is allowed:
-        two different EdgePlacement objects may select intersecting Location sets
-        and/or intersecting Namespace sets. This is not problematic because: - propagation
-        _into_ a destination is additive; - propagation _from_ a source is additive;
-        - two directives to propagate the same object to the same destination are
-        simply redundant (they, by design, can not conflict)."
+      description: "EdgePlacement exists in the center and binds (a) a collection
+        of Locations with (b) both (b1) objects in the center to downsync (propagate
+        desired state from center to edge and return reported state from edge toward
+        center), and (b2) a way identifying objects (in edge clusters) to upsync (propagate
```
clubanderson, line 22: should "a way identifying" be "a way of identifying"?

clubanderson (commented):
looks good to me after a grammar and reference issue are resolved.

pkg/apis/edge/v1alpha1/edge-placement.go
```diff
@@ -102,6 +116,35 @@ type NonNamespacedObjectReferenceSet struct {
...
+	// An entry of `"*"` means that all match.
+	// Empty list means nothing matches.
+	Resources []string `json:"resources"`
+
+	// `namespaces` is a list of acceptable namespaces.
+	// An entry of `"*"` means that all match.
+	// Empty list means nothing matches (you probably do not want this
+	// for namespaced resources).
+	Namespaces []string `json:"namespaces,omitempty"`
+
```
pdettori, line 141: Since we are still using TMC, and the TMC syncer (at least for now), namespaces names are generated by kcp and it may be quite difficult to specify selections based on namespaces. I think that until we create a proper EMC syncer that maintains namespace names from workload workspace to pCluster it will be pretty much impossible to use namespaces to filter.

pdettori: Any thoughts on how this up syncing selection mechanism would be implemented ? It looks like it will require a new controller within the syncer, perhaps in addition to the current upsyncer controller. Such controller will need to get the UpsyncSet spec. But such spec for now exists only in the Workload namespace. So I see couple of challenges here: 1) such controller would have to watch / read EdgePlacement objects in the Workload namespace (while currently only accesses the Mailbox workspace, and 2) such controller need to identify which EdgePlacement object applies to it.

pkg/apis/edge/v1alpha1/edge-placement.go
MikeSpreitzer: It is easy enough to imagine such a thing.  I would like to hear whether the C2P work needs it.

pkg/apis/edge/v1alpha1/edge-placement.go
MikeSpreitzer: My first concern was defining something that makes sense when we get where we are trying to go, having an EMC syncer.  In that situation, the resolution of the upsync selection predicates to the objects to upsync would be done by the EMC syncer, and it would not be confounded by namespace mangling.

In the interim while we use the TMC syncer, can we turn off the namespace mangling?  I thought I heard that there were other users asking for this and @davidfestal willing to make it an option, but I might not be remembering correctly, but it could still be an option anyway.

In the interim while using the TMC syncer, one approach would be to supply our own controller that runs in the edge clusters to apply the labels needed to trigger upsync by TMC.  Our controller would need the relevant configuration information, which could relayed by the placement translator and the TMC syncer.  Is the namespace mangling predictable?  If so then the configuration for our controller could take it into account.

I still have questions relevant here about TMC upsync.  One is at https://kubernetes.slack.com/archives/C021U8WSAFK/p1678998135151379; getting an answer to that would help.

Another possible option would be to not bother implementing the mailbox/edge part of this while using the TMC syncer, instead making it the user's problem to make the objects-to-upsync at the edge have the labels that directly drive TMC upsync.  Evaluating the feasibility of this involves both getting the answer to my question above and understanding whether the C2P work needs upsync of only objects that are easy to get the needed labels on.

config/exports/apiresourceschema-edgeplacements.edge.kcp.io.yaml
MikeSpreitzer: Yes, thanks.

config/crds/edge.kcp.io_edgeplacements.yaml
MikeSpreitzer: I think the term cluster-scoped is the one more commonly used in the Kubernetes world.  I sometimes write "non-namespaced" when I particularly want to emphasize the lack of a namespace, which particularly arises in TMC-adjacent work because TMC gives much more emphasis to namespaced objects.

BTW, this file is derived from edge-placement.go, in the future I think it would be better to comment on the source file.

clubanderson: ok - its just not consistent with the way we talk about the objects day-to-day.

ezrasilvera:
> 
Thanks for the detailed reply 
I agree the definition should be in the center. At this time I can't point out specific use-case in which covering it by upSync definition in the edgeplacement will be a problem (or not work) . So if, as you said, we can have an empty upSync I think it makes sense to put it in the edgeplacement. 

For the upSync we must know the target workspace, are we garneted to have 1:1 mapping from pCluster to target upSync workspace defined in the Edgeplacement ?  Are we planning to upsync into the mailbox workspaces ?

MikeSpreitzer:
@ezrasilvera : yes, upsync starts with propagation from edge cluster to its unique corresponding mailbox workspace.  It should be able to continue with summarization to the workload management workspace.

pkg/apis/edge/v1alpha1/edge-placement.go
MikeSpreitzer: @yuji-watanabe-jp ?  @yana1205 ?

MikeSpreitzer:
The force-push to a197cae fixes the typo that @clubanderson found.

config/exports/apiresourceschema-edgeplacements.edge.kcp.io.yaml
MikeSpreitzer: Fixed in the force-push to a197cae

pkg/apis/edge/v1alpha1/edge-placement.go
yuji-watanabe-jp: good to have flexible selector (namespace selector and resources selector)
- namespace ... label
- resource ... GVK, label, annotation + namespace selector

I agree that selector based on ownerReferences might be useful, but raises the implementation concern of syncer. Chasing owner to the downsynced object might not be always possible since syncer's access is constrained only to downsynced/upsynced objects with RBAC, but ownership link is not always directly among them. So, maybe, we can only support direct-ownership only.

yuji-watanabe-jp:
I am planning to upsync into the mailbox workspaces.


________________________________
From: Ezra Silvera ***@***.***>
Sent: Friday, March 17, 2023 14:48
To: kcp-dev/edge-mc ***@***.***>
Cc: Yuji 1 Watanabe ***@***.***>; Assign ***@***.***>
Subject: [EXTERNAL] Re: [kcp-dev/edge-mc] ✨ Extend EdgePlacement with spec of upsync (PR #239)

Thanks for the detailed reply I agree the definition should be in the center. At this time I can't point out specific use-case in which covering it by upSync definition in the edgeplacement will be a problem (or not work) . So if, as you said,
ZjQcmQRYFpfptBannerStart
This Message Is From an External Sender
This message came from outside your organization.

ZjQcmQRYFpfptBannerEnd

Thanks for the detailed reply
I agree the definition should be in the center. At this time I can't point out specific use-case in which covering it by upSync definition in the edgeplacement will be a problem (or not work) . So if, as you said, we can have an empty upSync I think it makes sense to put it in the edgeplacement.

For the upSync we must know the target workspace, are we garneted to have 1:1 mapping from pCluster to target upSync workspace defined in the Edgeplacement ? Are we planning to upsync into the mailbox workspaces ?

pkg/apis/edge/v1alpha1/edge-placement.go
MikeSpreitzer: In the C2P work, what would go wrong if the objects to upsync could be identified  _only_ by API group and resource (equivalently, Kind)?

pkg/apis/edge/v1alpha1/edge-placement.go
```diff
@@ -20,14 +20,23 @@ import (
...
+// Locations with (b) both (b1) objects in the center to downsync
+// (propagate desired state from center to edge and return reported
+// state from edge toward center), and (b2) a way of identifying objects
+// (in edge clusters) to upsync (propagate from edge toward center).
+// Both downsyncing and upsyncing are with all of the Locations.  This
+// is not entirely unrelated to a TMC Placement, which directs the
+// selected Namespaces to propagate to _one_ of the selected
+// Locations.
+//
+// The objects to downsync are those in selected namespaces plus
```
yuji-watanabe-jp, line 33: @MikeSpreitzer non-namespace objects can be specified by group, kind, name, and label. why namespace objects is not specified like that? should we assume all the non-namespace objects under specified namespaces are downsynced?

pkg/apis/edge/v1alpha1/edge-placement.go
```diff
@@ -20,14 +20,23 @@ import (
...
-// to propagate to _all_ of the selected Locations
-// (one SyncTarget in each such Location).
-// This is not entirely unrelated to a TMC Placement, which directs the selected
-// Namespaces to propagate to _one_ of the selected Locations.
+// EdgePlacement exists in the center and binds (a) a collection of
+// Locations with (b) both (b1) objects in the center to downsync
+// (propagate desired state from center to edge and return reported
+// state from edge toward center), and (b2) a way of identifying objects
+// (in edge clusters) to upsync (propagate from edge toward center).
+// Both downsyncing and upsyncing are with all of the Locations.  This
```
yuji-watanabe-jp, line 28: @MikeSpreitzer Is the set of Locations selected only by label of the Location object? Can we assume someone who defines this edge placement spec has knowledge on how to pick label for selecting destination?

pkg/apis/edge/v1alpha1/edge-placement.go
yuji-watanabe-jp: For example, we have use case of downsyncing a secret from mailbox to p-cluster by edge syncer, and want to specifiy specific secret to be downsynced, but not all secrets on the mailbox.

MikeSpreitzer: Can I please start by asking for a clarification of what you mean by "non-namespace"?  I can think of two meanings: (1) not a namespace, and (2) not _in_ a namespace.  The latter I sometimes call "non-namespaced", and it is also called "cluster-scoped" in the general Kubernetes community.

I based my proposal for EdgePlacement on TMC's Placement.  There (https://github.com/kcp-dev/kcp/blob/v0.11.0/pkg/apis/scheduling/v1alpha1/types_placement.go#L78) you will see that there is only a label selector for `Namespace` objects, and everything in each matching namespace gets downsynced.

Does something go wrong for C2P with the current downsync semantics for EdgePlacement and the proposal here for adding upsync?

I should perhaps expand on my answer to @pdettori 's earlier question (https://github.com/kcp-dev/edge-mc/pull/239#discussion_r1139596912).  I am trying to not make the upsync spec more complicated than necessary because, as observed by him in other conversations, we have to come up with implementations (possibly one connecting to TMC, certainly one in EMC).  I do not want to make our initial implementation job harder than it needs to be.  Nothing prevents adding complexity later if and when a need arises.

pkg/apis/edge/v1alpha1/edge-placement.go
MikeSpreitzer: Yes and Yes.
This should be described more clearly in the outline.
It differs from TMC, I think, although my understanding of TMC is incomplete.  As far as I know, in TMC the Placement has a predicate over SyncTarget labels but the SyncTarget objects are not intended to be exposed to the authors of Placements.

pkg/apis/edge/v1alpha1/edge-placement.go
yuji-watanabe-jp: OK, let's revisit this later, since this might be changed if we have new edge syncer.

pkg/apis/edge/v1alpha1/edge-placement.go
yuji-watanabe-jp: not sure if the current selector is enough for supporting all the PEP/PVPs at this point, but no issue for demonstrating Kyverno scenario. Let's revisit this later when we have edge-syncer and conduct more expanded and complicated integration. I believe this flexibility is the key if we use this as a configuration management platform for generic scenarion rather than just for kyverno.

pkg/apis/edge/v1alpha1/edge-placement.go
davidfestal: > In the interim while using the TMC syncer, one approach would be to supply our own controller that runs in the edge clusters to apply the labels needed to trigger upsync by TMC. Our controller would need the relevant configuration information, which could relayed by the placement translator and the TMC syncer. Is the namespace mangling predictable? If so then the configuration for our controller could take it into account.

If I understand this correctly, that seems to best option, at least when using the TMC Syncer.
Yes, namespace mangling is predictable. Here is an example of code that retrieves it:

https://github.com/kcp-dev/kcp/blob/46f72457e994fabbda0175f57183ecc454b03a64/pkg/tunneler/podsubresourceproxy_handler.go#L204

pkg/apis/edge/v1alpha1/edge-placement.go
pdettori: @MikeSpreitzer I am ok with keeping the specs simpler for now.

pkg/apis/edge/v1alpha1/edge-placement.go
pdettori: @MikeSpreitzer I suppose that would be ok for C2P. There are other scenarios (like kcp + Crossplane) where the objects to upsync are secrets. But we may not want to upsync all secrets for security reasons. In any case, I am fine to leave this for later.

PR #240, opened 2023-03-16
✨ First draft of API for reading objects from mailbox workspaces

MikeSpreitzer:
## Summary
This PR introduces an API for staying informed of objects in mailbox workspaces.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/mailboxwatch/interface.go
+++ b/pkg/mailboxwatch/interface.go
@@ -1,0 +1,77 @@
+/*
+Copyright 2023 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mailboxwatch
+
+import (
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	restclient "k8s.io/client-go/rest"
+	upstreamcache "k8s.io/client-go/tools/cache"
+
+	kcpcache "github.com/kcp-dev/apimachinery/v2/pkg/cache"
+	kcpinformers "github.com/kcp-dev/apimachinery/v2/third_party/informers"
+	kcpscopedclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	kcpapiinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+	tenancyv1a1informers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/tenancy/v1alpha1"
+)
+
+// NewSharedInformer constructs a shared index informer on objects of a given kind in mailbox workspaces.
+// It follows the usual pattern for a constructor of informers except for an additional parameter
+// at the front, which it uses to keep appraised of the mailbox workspaces.
+// The ListerWatcher can be constructed using `upstreamcache.NewListWatchFromClient`.
+func NewSharedInformer(
+	mailboxWorkspacePreInformer tenancyv1a1informers.WorkspaceInformer,
+	listerWatcher upstreamcache.ListerWatcher,
+	exampleObject runtime.Object,
+	defaultEventHandlerResyncPeriod time.Duration,
+	indexers upstreamcache.Indexers,
+) kcpcache.ScopeableSharedIndexInformer {
+	if true {
+		panic("not implemented yet")
+	}
+	// Implementation outline:
+	// Use the informer on mailbox workspaces to stay appraised of the logicalcluster.Name for
+	// each mailbox workspace qua logical cluster.
+	// Wrap the given ListerWatcher with filtering that passes only objects in one of those logical clusters.
+	// At first stage of development, do not take special care when mailbox workspaces arrive or depart,
+	// because objects in them will likely arrive later and depart sooner.
+	// At a later stage of development, do not rely on happy timing.
+	var wrappedListerWatcher upstreamcache.ListerWatcher
+	return kcpinformers.NewSharedIndexInformer(wrappedListerWatcher, exampleObject, defaultEventHandlerResyncPeriod, indexers)
+}
+
+// NewSharedInformerForEdgeConfig is like NewSharedInformer but takes a REST Config for the
+// edge service provider workspace and constructs the pre-informer for the mailbox workspaces.
+// That pre-informer is then used to call NewSharedInformer and is also returned,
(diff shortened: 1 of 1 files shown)


yuji-watanabe-jp (commented):
This looks really useful! Do you have a plan to provide the access to a kind of APIExport View though this API? @MikeSpreitzer

pkg/mailboxwatch/interface.go
```diff
@@ -0,0 +1,77 @@
...
+	listerWatcher upstreamcache.ListerWatcher,
+	exampleObject runtime.Object,
+	defaultEventHandlerResyncPeriod time.Duration,
+	indexers upstreamcache.Indexers,
+) (tenancyv1a1informers.WorkspaceInformer, kcpcache.ScopeableSharedIndexInformer, error) {
+	workspaceScopedClientset, err := kcpscopedclientset.NewForConfig(edgeServiceProviderWorkspaceClientConfig)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create clientset for mailbox workspaces: %w", err)
+	}
+	workspaceScopedInformerFactory := kcpapiinformers.NewSharedScopedInformerFactoryWithOptions(workspaceScopedClientset, 0)
```
ezrasilvera, line 74: Do we need to allow setting Rsync period also for the pre- informer ?

MikeSpreitzer: The default resync is on the factory, not the pre-informer.
As far as I know, resync is an obsolete feature.  I do not understand how a correct controller would benefit from using resync.  At best it could paper over _some_ flaws in an incorrect controller.  Resync does not notify of deletions.

Would the C2P Controller use resync?

MikeSpreitzer:
@yuji-watanabe-jp : An APIExport view comes as a consequence of using APIBindings to an APIExport.  The API here would not be used to implement such a thing, APIExports are already implemented.  See how #242 discusses APIExport view as an _alternative_ to using this API.  It would be possible to implement another kind of view, and that implementation might use this interface.

pkg/mailboxwatch/interface.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// NewSharedInformer constructs a shared index informer on objects of a given kind in mailbox workspaces.
+// It follows the usual pattern for a constructor of informers except for an additional parameter
+// at the front, which it uses to keep appraised of the mailbox workspaces.
+// The ListerWatcher can be constructed using `upstreamcache.NewListWatchFromClient`.
+func NewSharedInformer(
+	mailboxWorkspacePreInformer tenancyv1a1informers.WorkspaceInformer,
+	listerWatcher upstreamcache.ListerWatcher,
+	exampleObject runtime.Object,
```
pdettori, line 41: Can we call this `watchedObject` or something like that ? I think it would make the function signature a bit more clear.

pkg/mailboxwatch/interface.go
```diff
@@ -0,0 +1,77 @@
...
+	defaultEventHandlerResyncPeriod time.Duration,
+	indexers upstreamcache.Indexers,
+) kcpcache.ScopeableSharedIndexInformer {
+	if true {
+		panic("not implemented yet")
+	}
+	// Implementation outline:
+	// Use the informer on mailbox workspaces to stay appraised of the logicalcluster.Name for
+	// each mailbox workspace qua logical cluster.
```
pdettori, line 50: I suppose this is a typo ('`workspace qua logical`) but I cannot infer what this was intended to mean.

pkg/mailboxwatch/interface.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// NewSharedInformerForEdgeConfig is like NewSharedInformer but takes a REST Config for the
+// edge service provider workspace and constructs the pre-informer for the mailbox workspaces.
+// That pre-informer is then used to call NewSharedInformer and is also returned,
+// so that the caller can wait on HasSynced of the mailbox workspace informer.
+func NewSharedInformerForEdgeConfig(
+	edgeServiceProviderWorkspaceClientConfig *restclient.Config,
+	listerWatcher upstreamcache.ListerWatcher,
+	exampleObject runtime.Object,
```
pdettori, line 66: Same comment as above on the name `exampleObject`

pkg/mailboxwatch/interface.go
MikeSpreitzer: But it is not _the_ watched object, the purpose here is to watch several objects, and the one passed in is not any one of them.  It is only there to convey information about the type of object to watch.  That is why "example" is on-point.

pkg/mailboxwatch/interface.go
```diff
@@ -0,0 +1,77 @@
...
+	kcpapiinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+	tenancyv1a1informers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/tenancy/v1alpha1"
+)
+
+// NewSharedInformer constructs a shared index informer on objects of a given kind in mailbox workspaces.
+// It follows the usual pattern for a constructor of informers except for an additional parameter
+// at the front, which it uses to keep appraised of the mailbox workspaces.
+// The ListerWatcher can be constructed using `upstreamcache.NewListWatchFromClient`.
+func NewSharedInformer(
+	mailboxWorkspacePreInformer tenancyv1a1informers.WorkspaceInformer,
```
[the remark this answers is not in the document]
MikeSpreitzer: There is an unfortunate convention in the stub generator to use the word "informer" for _two_ **different** things.  One is what nobody would debate calling an "informer".  The other is what I have lately been calling a "pre-informer" and @waltforme called "access".  This type of thing has two methods: one named "Informer" and one named "Lister".  Because this type of thing _precedes_ getting your hands on the informer, that is why I chose "pre-informer".

pkg/mailboxwatch/interface.go
MikeSpreitzer: Below is an example display of a Workspace object.  You can see two logical clusters mentioned in there.  The containing one, in the annotations, and the one that the Workspace represents, in the spec. 

```console
(base) mspreitz@mjs12 kcp % kubectl get Workspace edge -o yaml
apiVersion: tenancy.kcp.io/v1alpha1
kind: Workspace
metadata:
  annotations:
    experimental.tenancy.kcp.io/owner: '{"username":"kcp-admin"}'
    internal.tenancy.kcp.io/cluster: dto88oazto05t06o
    internal.tenancy.kcp.io/shard: 1pfxsevk
    kcp.io/cluster: root
  creationTimestamp: "2023-03-20T18:49:25Z"
  finalizers:
  - core.kcp.io/logicalcluster
  generation: 2
  labels:
    tenancy.kcp.io/phase: Ready
  name: edge
  resourceVersion: "836"
  uid: 1c72205c-1bb6-4f54-9d82-10e941cbde8a
spec:
  URL: https://${myip}:6443/clusters/dto88oazto05t06o
  cluster: dto88oazto05t06o
  type:
    name: organization
    path: root
status:
  conditions:
  - lastTransitionTime: "2023-03-20T18:49:26Z"
    status: "True"
    type: WorkspaceInitialized
  - lastTransitionTime: "2023-03-20T18:49:25Z"
    status: "True"
    type: WorkspaceScheduled
  phase: Ready
```

PR #238, opened 2023-03-16
🐛 Revise and extend remarks about APIBinding in workload workspace

MikeSpreitzer:
## Summary
This PR replaces an incorrect statement about APIBindings with a correct statement and explicitly notes their limited role in the 2023q1 PoC.

## Related issue(s)

Fixes #

PR #241, opened 2023-03-16
✨ Update thoughts on APIBindings and roadmap options

MikeSpreitzer:
## Summary
This PR updates the 2023q1 PoC outline to be more correct and expansive regarding APIBinding objects and also expand options in the roadmap section.

This includes calling out a scenario in which workload objects in mailbox workspaces (as well as all other workspaces) are visible in an APIExport view.

This PR replaces #238 with a more extensive revision.

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
+++ b/docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
@@ -122,6 +122,34 @@ what should be returned as well as implementing it.  This will
 certainly affect the syncer between mailbox workspace and edge
 cluster, and the summarization part will affect the status summarizer.
 
+### Good handling of workload conflicts
+
+We could start by handling workload conflicts in a very simple way:
+treating each as an error.  Later development can handle them better,
+as outlined [later](#edgeplacement-objects).
+
+### Denaturing/renaturing
+
+We could start by not doing this.  For some resources, the effect of
+leaving these resources natured in the center is only to add
+authorizations in the center that are not needed and are undesired in
+a well-secured environment but tolerable in early demonstrations ---
+provided that there is not a conflict with an object of the same name
+that is positively desired in the center.  In particular, these are:
+`ClusterRole`, `ClusterRoleBinding`, `Role`, `RoleBinding`,
+`ServiceAccount`.
+
+For some kinds of object, lack of denaturing/renaturing means that
+edge-mc will simply not be able to support workloads that contain such
+objects.  These are: `MutatingWebhookConfiguration`,
+`ValidatingWebhookConfiguration`, `LimitRange`, `ResourceQuota`.
+
+For some resources, the need to denature is only a matter of
+anticipation.  `FlowSchema` and `PriorityLevelConfiguration` currently
+are not interpreted by kcp servers and so are effectively already
+denatured there.  Hopefully they will be given interpretations in the
+future, and then those resources will join the previous category.
+
 ## Roles and Responsibilities
 
 ### Developers/deployers/admins/users of the inventory management layer
@@ -245,10 +273,19 @@ dependent objects.
 | ---------- | ---- | ---------- |
 | apis.kcp.io/v1alpha1 | APIBinding | false |
 
-A workload management workspace generally has APIBindings to workload
-APIs.  These bindings cause corresponding CRDs to be created in the
-same workspace.  The CRDs propagate to the edge, the APIBindings do
-not.
+A workload management workspace needs an APIBinding to the APIExport
+of the API group `edge.kcp.io` from the edge service provider
+workspace, in order to be able to contain EdgePlacement and related
+objects.  These objects and that APIBinding are not destined for the
+edge clusters.
+
+The edge clusters are not presumed to be kcp workspaces, so
+APIBindings do not propagate to the edge clusters.  However, it is
+possible that APIBindings for workload APIs may exist in a workload
+management workspace and be selected for downsync to mailbox
+workspaces while the edge clusters have the same resources defined by
+CRDs (as mentioned later in the discussion of built-in resources and
+namespaces).
 
 #### For features not supported
 
@@ -345,7 +382,7 @@ this category.
 | v1 | Secret | true |
 | v1 | Service | true |
 
-### Built-in resources and objects
+### Built-in resources and namespaces
 
 An edge cluster has some built-in resources (i.e, kinds of objects)
 and namespaces.  A resource may be built-in by any of several ways: it
@@ -376,6 +413,31 @@ that are a subset of those built into the edge clusters.  The
 propagation from workload management workspace to mailbox workspace
 does not attempt to manage the resource and namespace definitions that
 are built into the mailbox workspaces.
+
+The above wording is deliberately restrained, for the sake of
+flexibility regarding resources that are defined one way in the edge
+clusters and another way in workload management workspace.  For
+example, the following scenario is allowed.
+
+- Some central team owns an API group and produces some
+  CustomResourceDefinition (CRD) objects that populate that API group.
+- That team derives APIResourceSchemas from those CRDs and a
+  corresponding APIExport of their API group.
+- That team maintains a kcp workspace holding those APIResourceSchemas
+  and that APIExport.
+- Some workload management workspaces have APIBindings to that
+  APIExport, and EdgePlacement objects that (1) select those
+  APIBinding objects for downsync and (2) select objects of kinds
+  defined through those APIBindings for either downsync or upsync.
+- Those resources are built into the edge clusters by pre-deploying
+  the aforementioned CRDs there.
+- Those resources are _not_ built into the mailbox workspaces.  In
+  this case the APIBindings would propagate from workload management
+  workspace to mailbox workspaces but not edge clusters.
+- As a consequence of those propagated APIBindings, the APIExport's
+  view includes all of the objects (in workload management workspaces,
+  in mailbox workspaces, and in any other workspaces where they
+  appear) whose kind is defined through those APIBindings.
 
 ### Control objects
 


docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
```diff
@@ -122,6 +122,34 @@ what should be returned as well as implementing it.  This will
...
+For some kinds of object, lack of denaturing/renaturing means that
+edge-mc will simply not be able to support workloads that contain such
+objects.  These are: `MutatingWebhookConfiguration`,
+`ValidatingWebhookConfiguration`, `LimitRange`, `ResourceQuota`.
+
+For some resources, the need to denature is only a matter of
+anticipation.  `FlowSchema` and `PriorityLevelConfiguration` currently
+are not interpreted by kcp servers and so are effectively already
+denatured there.  Hopefully they will be given interpretations in the
+future, and then those resources will join the previous category.
```
yuji-watanabe-jp, line 151: This is really important feature for edge syncer to support these resources. 
- replace to some alternative kind in workspace side
- edge syncer pull the alternative kind and replace it with native one. 
I think this should be one of the priority item.

MikeSpreitzer: Denaturing and renaturing are coupled.  The denaturing is done by a view, the renaturing is done by (the placement translator while the TMC syncer is used, the EMC syncer once it is developed).

I see now that #242 does not yet clearly enough ask the relevant questions here.  So let me ask them here.  First I should note that I hope that kcp will be rebased onto Kubernetes release 1.26 soon, which will make https://kubernetes.io/docs/reference/kubernetes-api/extend-resources/validating-admission-policy-v1alpha1/ usable; I wonder whether (as suggested at the end of that review PDF you shared) C2P could use `ValidatingAdmissionPolicy` objects.  In that future, does C2P need `MutatingWebhookConfiguration`, `ValidatingWebhookConfiguration`, `LimitRange`, or `ResourceQuota`?  More precisely, could we demo an interesting subset of C2P without using those resources?

yuji-watanabe-jp: > could we demo an interesting subset of C2P without using those resources?

Kyverno does not require syncing those resources for C2P demo. `MutatingWebhookConfiguration` and/or `ValidatingWebhookConfiguration` are generated after deploying Kyverno runtime on p-cluster.

MikeSpreitzer: @yuji-watanabe-jp : thank you, I just want to follow up to make sure I understand correctly.  When the Kyverno controller generates a `MutatingWebhookConfiguration` or `ValidatingWebhookConfiguration` object in the edge cluster, we have to be careful about which apiserver this object is written to.  It needs to be written to the apiserver that it is intended to affect.  My concern is what happens while using the TMC syncer.  If the webhook configuration object is written to the kcp workspace but is intended to configure calls to a container running in the edge cluster then this will be a broken configuration.  Is the Kyverno controller described by a Deployment object that will be downsynced with the TMC syncer rewiring its apiserver client config to point back to the mailbox workspace?

MikeSpreitzer: A Kubernetes `Service` does not create webhook configuration objects, but a `Pod` (created as a consequence of a `Deployment`) might.

When TMC downsyncs a `Deployment` object, the syncer modifies the `PodTemplaceSpec` in that `Deployment` so that as it appears on the pcluster, the `Pod` objects that get created will --- IF they use the regular way of connecting --- connect to the kcp workspace rather than the pcluster's apiserver.

yuji-watanabe-jp: Sorry, I am just using service in general sense, but i do not have much concern on this. When kyverno starts on p-cluster, kyverno configure webhook on the cluster the kyverno is running unless any unexpected mutation happens in the manifest. It is p-cluster. nothing related to kcp workspace anymore.

yuji-watanabe-jp: if the syncer is doing something bad when syncing resources from workspace to p-cluster, something happens, but I have not seen the situation so far in our experiments.

MikeSpreitzer: @yuji-watanabe-jp , are you expecting the manifest mutation that the TMC syncer normally does?  They do not consider it "bad", it is the intended behavior of that thing.

yuji-watanabe-jp: OK, "bad" might be wrong word. I meant unexpected for me. If it is expected from TMC perspective, we need to deal with it and maybe need to find workaround. If we use edge-syncer, it is different story. so we can revisit this later.

pdettori: @yuji-watanabe-jp I think it's important to clarify how Kyverno gets installed on the p-Cluster. Based on how it gets installed the behavior of Kyverno controllers can be very different.  If you install Kyverno directly on the pCluster, without any involvement of kcp, then the Kyverno controllers know nothing of kcp and only talk to the  pCluster API server. `MutatingWebhookConfiguration` and `ValidatingWebhookConfiguration` are created in the pCluster by the Kyverno controllers.  But ... if you install kyverno from on a kcp workspace, and you have TMC syncer configured to sync the Kyverno deployment to the pCluster, then the TMC syncer mutators update the deployment to point back to the kcp Workspace. In this scenario, the Kyverno CRDs (and other auxiliary resources such as secrets etc) are installed in the kcp workspace. The Kyverno controllers will then try to create `MutatingWebhookConfiguration` and `ValidatingWebhookConfiguration` on the Kcp workspace, and also it may not be able to build properly a policy report for the pCluster as they are not watching/listing objects on the pCluster but on the kcp workspace. 

I remember some time ago you showed a chart where you were installing Kyverno with OLM and just syncing the OLM templates. In that case, you may have the benefit of installing on the pCluster without being affected by the syncer mutators (which currently only manipulate Deployments, Pods and ReplicaSets). Is that still the case ?

MikeSpreitzer: I think we have some choices here.  Perhaps installing Kyverno via OLM is the better choice in the near term.  In that case, we would have the issue TMC upsync of the policy reports from the edge cluster to the mailbox workspace.

yuji-watanabe-jp: @pdettori you are right. we have been deploying Kyverno to p-cluster with TMC syncer + OLM. We confirmed that works. 
Without OLM, we need to see how to install cluster-scope (ClusterRoleBinding and ClusterRole) via TMC syncer. 
So, we can keep the choice with OLM for installation via TMC syncer. The issue of TMC upsync of the policy reports still remains.

Edge syncer will do not apply such mutation, and so this will never happen there, and resolve the issue of upsync.

ezrasilvera: @yuji-watanabe-jp Cluster-scope deployment and the webhook issue are different. Just to emphasis (and maybe close :-)  ) this discussion - if you plan to deploy  Kyverno  through KCP (i.e., into a KCP workspace) than due to the mutating that TMC performs on the deployed pods your webhooks probably won't work correctly. If you can (for the first stage) deploy the controllers directly on the pCluster(s) - than you avoid the TMC mutating and your pCluster webhooks will probably work without an issue.

ezrasilvera: The upsync seems not related to this , as for both cases (deploy throgh KCP and externally to KCP) , we can do the same: (1) define the criteria of what to upsync through KCP (2) A controller on the pCLuster (part of the Kyverno  syncer) will need to support this upSync (e.g., set the labels/annotations, etc..)

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
```diff
@@ -377,6 +414,31 @@ propagation from workload management workspace to mailbox workspace
...
+  CustomResourceDefinition (CRD) objects that populate that API group.
+- That team derives APIResourceSchemas from those CRDs and a
+  corresponding APIExport of their API group.
+- That team maintains a kcp workspace holding those APIResourceSchemas
+  and that APIExport.
+- Some workload management workspaces have APIBindings to that
+  APIExport, and EdgePlacement objects that (1) select those
+  APIBinding objects for downsync and (2) select objects of kinds
+  defined through those APIBindings for either downsync or upsync.
+- Those resources are built into the edge clusters by pre-deploying
```
ezrasilvera, line 432: Probably only the resources for upsync need to be pre-deployed, right ?  Won't the ones we downsync be deployed by the downsync itself ?

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
```diff
@@ -377,6 +414,31 @@ propagation from workload management workspace to mailbox workspace
...
+  the aforementioned CRDs there.
+- Those resources are _not_ built into the mailbox workspaces.  In
+  this case the APIBindings would propagate from workload management
+  workspace to mailbox workspaces but not edge clusters.
+- As a consequence of those propagated APIBindings, the APIExport's
+  view includes all of the objects (in workload management workspaces,
+  in mailbox workspaces, and in any other workspaces where they
+  appear) whose kind is defined through those APIBindings.
+
```
ezrasilvera, line 441: Is there a diagram that describe those relations ? It may help to understand the mappings of binding, exports, resources, between the workload WS, mailbox WS, and pCluster.

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
MikeSpreitzer: Downsync cannot propagate APIBindings because the edge cluster is not a kcp workspace.
Also, an APIExport, and thus and APIBinding, is for a whole API group; one cannot use that mechanism to import a subset of what was exported.

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
MikeSpreitzer: I have not drawn a picture yet.
Is there information missing from the textual description?

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
ezrasilvera: I was just referring to the "by pre-deploying"  part. Why do we need to pre-deploy into KCP resources that were coming from KCP in the first place?

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
MikeSpreitzer: Is the controller that creates the policy objects something that we control? Can we make it put needed labels and annotations on the policy objects, at least in the TMC interim?

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
MikeSpreitzer: The point of this particular scenario is for the mailbox workspaces to have APIBindings to workload APIExports, so that the workload APIExport view holds the workload objects as they sit in the mailbox workspaces.  Particularly for upsynced resources, since this is something that the C2P people expected.

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
yuji-watanabe-jp: we control it. we can add them.

MikeSpreitzer:
In the latest kcp community meeting, @davidfestal gave an outline of the answer to my question (https://github.com/kcp-dev/kcp/issues/2905) about the `NamespaceLocator` mentioned wrt upsync in https://github.com/kcp-dev/kcp/issues/1980 and said he would document the answer.

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
MikeSpreitzer: (1) Drawings almost always help.  I hope the lack of one does not hold up merging this.

(2) The "sync" done by our processes is eventually consistent, which explicitly admits there may be interim periods of inconsistency.

(3) I suspect your concern is consistency that our automated processes do not even attempt to maintain (e.g., between a resource's schema built into edge clusters vs. the schema given for the same resource in the center).  That can be lacking from the start as well as crop up later!  This is an issue I tried to explicitly defer for later consideration.

(4) The scenario outlined here is no longer the one that we intend to use for the Compliance-to-Policy work.  For C2P we intend to drive CRDs from workload management workspace through mailbox workspaces to edge clusters.

clubanderson: I am ok with that answer.
/lgtm

PR #244, opened 2023-03-16
🌱 go mod tidy

waltforme:
## Summary
Looks like `k8s.io/apiextensions-apiserver` is now a direct dependency.


MikeSpreitzer (approved):
Sorry, oversight of mine.
@clubanderson : maybe CI should check this?

/lgtm

PR #254, opened 2023-03-21
✨ Reconciliation triggered by SyncTarget

### LINKED ISSUE #246: Define the shared data structure for the edge scheduler
We need some internal data structure inside the edge-scheduler to facilitate the incremental updates of the edge-scheduler's output i.e. singleplacementslices. These data structures are going to be concurrently accessed by worker threads.

waltforme:
## Summary
This PR implements parts of (those triggered by SyncTarget changes) edge-scheduler's reconciliation.

Fixes #246


waltforme:
Force-push to https://github.com/kcp-dev/edge-mc/commit/cc50ce296f209e4a049232bf3d381db70a6b2de5 adds boilerplate headers.

pkg/scheduler/reconcile_on_synctarget.go
```diff
@@ -19,20 +19,220 @@ package scheduler
...
+
+		Need data structure:
+		- map from a synctarget to its locations, say 'locsBySelectedSt'
+		- map from a synctarget to its eps, say 'epsByUsedSt'
+	*/
+
+	st, _ := c.synctargetLister.Cluster(stws).Get(stName) // TODO: handle err
+
+	// 1)
+	locsColocating, _ := c.locationLister.Cluster(stws).List(labels.NewSelector()) // TOOD: handle err
```
MikeSpreitzer, line 61: (1) The term "colocating" could refer to a few different sorts of location here.  The edge cluster and the inventory management workspace are both conceivable.  I think the intent is the latter.  What about using "InSameWS" instead of "Colocating"?

(2) I think the intent would be clearer if this used `labels.Everything()` rather than `labels.NewSelector()`.

pkg/scheduler/reconcile_on_synctarget.go
```diff
@@ -19,20 +19,220 @@ package scheduler
...
+	locsSelectingSt, _ := filterLocsBySt(locsColocating, st)                       // TODO: handle err
+
+	locKeys := extractLocKeys(locsSelectingSt)
+	// locsSelectedSt := store.locsBySelectedSt[stKey]
+	store.locsBySelectedSt[stKey] = locKeys
+
+	// 2a)
+	epsUsingSt := []string{}
+	for _, loc := range locsSelectingSt {
+		locKey := kcpcache.ToClusterAwareKey(string(logicalcluster.From(loc)), loc.GetNamespace(), loc.GetName())
```
MikeSpreitzer, line 71: Why not simply use `kcpcache.MetaClusterNamespaceKeyFunc(loc)` here?  BTW, you are guaranteed that it will never return an error when given a `Location`.

pkg/scheduler/store.go
```diff
@@ -0,0 +1,60 @@
...
+	ep <- loc
+	loc <- st
+	ep <- st
+
+*/
+
+type internalData struct {
+	l sync.Mutex
+
+	// TODO(waltforme): should these map values be slices or maps?
```
MikeSpreitzer, line 43: I think the tediosity of the code would be much reduced if the map values here were https://github.com/kubernetes/apimachinery/blob/v0.24.3/pkg/util/sets/string.go#L27

pkg/scheduler/store.go
```diff
@@ -0,0 +1,60 @@
...
+	epsBySelectedLoc map[string][]string // ep <- loc
```
MikeSpreitzer, line 44: I think you could make this field name shorter, because there are no other Locations relevant to an EdgePlacement besides the ones that it selects.  You could omit "Selected" from this field name.

An analogous comment applies to the other fields here too.

MikeSpreitzer (commented):
This stuff cries out for https://docs.python.org/3/library/stdtypes.html#iterator-types ,  https://docs.python.org/3/library/itertools.html , and https://docs.python.org/3/library/functools.html .

waltforme:
Rebased on latest main, added https://github.com/kcp-dev/edge-mc/pull/254/commits/2866dcbc56016600e27ad5fea09c19145fe31f7c to response to @MikeSpreitzer 's review.

pkg/scheduler/reconcile_on_synctarget.go
waltforme: (1) Agreed and renamed.

(2) Yes, changed it.

pkg/scheduler/reconcile_on_synctarget.go
waltforme: I saw `MetaClusterNamespaceKeyFunc` calls `ToClusterAwareKey` so tried the latter out. Changed it back for simplicity.

May I know why an error is never returned here?

pkg/scheduler/store.go
waltforme: Agreed and fixed.

pkg/scheduler/reconcile_on_synctarget.go
```diff
@@ -19,20 +19,220 @@ package scheduler
...
 	logger.V(2).Info("reconciling")
 
+	/*
+		On synctarget change:
+		1) find all its loc(s), and update store
+
+		2a) find all its ep(s), and update store
+		2b) for each of its obsolete ep, remove all sp(s) that has st
+		2c) for each of its ongoing ep, update sp(s)
+		2d) for each of its new ep, add sp(s)
```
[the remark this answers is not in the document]
waltforme: Refactored the data structure so that there are two relationships and no redundancy.

pkg/scheduler/store.go
waltforme: Technically it is correct that the only relationship is "selection". But I'd like to keep the names here, because they reads more natural and carries more essential info, thus make the thinking easier when I use the names.

waltforme:
> This stuff cries out for https://docs.python.org/3/library/stdtypes.html#iterator-types , https://docs.python.org/3/library/itertools.html , and https://docs.python.org/3/library/functools.html .

Not sure I am following here. May I have more info in addition to the pointers?

MikeSpreitzer:
I saw two copies of code that were iterating over collections of a certain kind of thing (was it SyncTarget?  Location?) and doing the same thing for each member but the code was duplicated because the representation of the collection was different: a slice in one case and a map in the other.  The more abstract style of programming in Python would not have required such duplication.  And in general there is a stronger practice of functional programming there, I am sure at least in part because the lack of generics in go made it impossible until recently.

waltforme:
> I saw two copies of code that were iterating over collections of a certain kind of thing (was it SyncTarget? Location?) and doing the same thing for each member but the code was duplicated because the representation of the collection was different: a slice in one case and a map in the other. The more abstract style of programming in Python would not have required such duplication. And in general there is a stronger practice of functional programming there, I am sure at least in part because the lack of generics in go made it impossible until recently.

Thanks for the info! Maybe we can do this latter? How about getting a working scheduler first?

pkg/scheduler/reconcile_on_synctarget.go
```diff
@@ -19,20 +19,254 @@ package scheduler
...
+	// 1a)
+	epsUsedSt := store.findEpsUsedSt(stKey)
+
+	// 2a)
+	locsInStws, err := c.locationLister.Cluster(stws).List(labels.Everything())
+	if err != nil {
+		logger.Error(err, "failed to list Locations")
+		return err
+	}
+	locsSelectingSt, err := filterLocsBySt(locsInStws, st)
```
MikeSpreitzer, line 77: Here is a suggestion that I think will reduce duplication of code.

At this line, write `locsSelectingSt := SliceFilter(locsInStws, func (loc *schedulingv1alpha1.Location) bool {return c.evalLabelSelector(loc.Spec.InstanceSelector, st.Labels)})`

Also write the following broadly re-useable functions.

```
func SliceFilter[Elt any](slice []Elt, test func(Elt) bool) []Elt {...}

func (c *controller) evalLabelSelector(selector LabelSelector, labels map[string]string) bool {... evaluate, error from metav1.LabelSelectorAsSelector gets logged and treated like a mismatch ...}
```

pkg/scheduler/reconcile_on_synctarget.go
```diff
@@ -19,20 +19,254 @@ package scheduler
...
+
+	store.locsBySelectedSt[stKey] = packLocKeys(locsSelectingSt)
+
+	// 2b)
+	epsUsingSt := map[string]empty{}
+	for _, loc := range locsSelectingSt {
+		locKey, _ := kcpcache.MetaClusterNamespaceKeyFunc(loc)
+		eps := store.epsBySelectedLoc[locKey]
+		epsUsingSt = unionTwo(epsUsingSt, eps)
+	}
```
MikeSpreitzer, line 91: This is map/reduce, for which you could write one generic definition rather than spell it out each time.

pkg/scheduler/store.go
```diff
@@ -0,0 +1,78 @@
...
+	store = internalData{
+		l:                sync.Mutex{},
+		epsBySelectedLoc: map[string]map[string]empty{},
+		locsBySelectedSt: map[string]map[string]empty{},
+	}
+
+	store.manuallyFillIn()
+}
+
+func unionTwo(a, b map[string]empty) map[string]empty {
```
MikeSpreitzer, line 60: If you defined `union(maps ... map[string]empty) map[string]empty` then it would be just as easy to call as unionTwo, unionThree, and so on but you would not have to write multiple definitions.

Also, remembering map/reduce, this is one form of "reduce".

pkg/scheduler/reconcile_on_synctarget.go
```diff
@@ -19,20 +19,254 @@ package scheduler
...
+		3b) for each of its ongoing ep, update sp(s)
+		3c) for each of its new ep, add sp(s)
+
+		Need data structure:
+		- map from a synctarget to its locations, say 'locsBySelectedSt'
+	*/
+
+	st, err := c.synctargetLister.Cluster(stws).Get(stName)
+	if err != nil {
+		logger.Error(err, "syncTarget not found in local cache")
```
MikeSpreitzer, line 61: When `IsNotFound(err)` this is not an error condition, this is where the implementation needs to react to deletion of a SyncTarget.

pkg/scheduler/reconcile_on_synctarget.go
```diff
@@ -19,20 +19,254 @@ package scheduler
...
+
+		3a) for each of its obsolete ep, remove all sp(s) that has st
+		3b) for each of its ongoing ep, update sp(s)
+		3c) for each of its new ep, add sp(s)
+
+		Need data structure:
+		- map from a synctarget to its locations, say 'locsBySelectedSt'
+	*/
+
+	st, err := c.synctargetLister.Cluster(stws).Get(stName)
```
MikeSpreitzer, line 59: Note that this is a call to the apiserver.
There is an alternate approach in which this lookup in done in an infomrer's local cache (called a "lister"), without the runtime costs of a call to the apiserver.
In both approaches, the result may be stale information by the time the following code runs.  However, in some controllers the ordering of the request to the apiserver is helpful.

pkg/scheduler/store.go
```diff
@@ -0,0 +1,78 @@
...
+	loc <- st
+
+*/
+
+type empty struct{}
+
+type internalData struct {
+	l sync.Mutex
+
+	epsBySelectedLoc map[string]map[string]empty // ep <- loc
```
MikeSpreitzer, line 44: If the inner map held the `Key` of https://github.com/kcp-dev/apimachinery/pull/70 (or an anticipatory local copy of that), or the `ExternalName` of pkg/placement, rather than `string` then you could eliminate a lot of code handling parsing errors.

MikeSpreitzer (approved):
I see a lot of logic here that will have to be in common with logic for handling Location and for handling EdgePlacement objects.  You will want to think about how to factor so that you do not repeat yourself a lot.

I will approve this as progress.

/lgtm

pkg/scheduler/reconcile_on_synctarget.go
waltforme: Hmm. Am I using the correct Lister already or there is some other lister?

waltforme:
The comments above are very informative and helpful. I hope I'm not driven by the tight schedule, so that I could have enough bandwidth to improve. Right now I'll have to make something runnable first, but will definitely revisit the comments. Thanks @MikeSpreitzer !

pkg/scheduler/reconcile_on_synctarget.go
MikeSpreitzer: Oh, sorry, I misread the code.  Yes, it looks like you are using a lister.

waltforme: Np! And glad to see your brief intro to listers.

PR #256, opened 2023-03-22
🐛 Attempt to fix broken reference to example1

MikeSpreitzer:
## Summary
This PR attempts to fix the reference from placement-translator.md to example1.md, in a way that works after it runs through the conversion to github pages.

## Related issue(s)

Fixes #


clubanderson (approved):
that should do it - lets see if it works

PR #257, opened 2023-03-22
📖 Update outline again

MikeSpreitzer:
## Summary
This PR builds on #241 by adding some clarifications and questions that have arisen in recent discussions.

If #241 merges first then I will rebase this PR.  Alternatively, we could just merge this PR and that would obviate #241 .

## Related issue(s)

Fixes #


docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
```diff
@@ -122,9 +122,37 @@ what should be returned as well as implementing it.  This will
...
 
+### Good handling of workload conflicts
+
+We could start by handling workload conflicts in a very simple way:
+treating each as an error.  Later development can handle them better,
+as outlined [later](#edgeplacement-objects).
+
+### Denaturing/renaturing
+
+We could start by not doing this.  For some resources, the effect of
```
ezrasilvera, line 133: I think we probably say something like "do it on a per need basis" or something like that. i.e., We are not going to implement an overall Denaturing/renaturing  , but I also think we will not "do nothing" . I thought we plan to include the KIvorno requirements for this PoC and they have mandatory requirements in this area

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
```diff
@@ -22,9 +22,9 @@ progress.
 This is a quick demo of a fragment of what we think is needed for edge
 multi-cluster.  It is intended to demonstrate the following points.
 
-- Separation of inventory and workload management.
+- Separation of infrastructure and workload management.
```
ezrasilvera, line 25: Do we need an "inventory" for the workloads as well? (i.e., separate the workload management into modules) or we define that as part of the overall "workload management"

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
```diff
@@ -135,16 +163,37 @@ cluster, and the summarization part will affect the status summarizer.
 ## Design overview
 
 In very brief: the design is to reduce each edge placement problem to
-many instances of kcp's TMC problem.
+many instances of kcp's TMC problem.  Part of making many instances of
```
ezrasilvera, line 166: Maybe "many instances of a single location/target TMC problem:" ?

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
```diff
@@ -135,16 +163,37 @@ cluster, and the summarization part will affect the status summarizer.
...
+
+However: to fully realize the goals of this PoC we will have to use a
+modified version of TMC, which differs from today's TMC in ways
+summarized below in the [Syncers](#syncers) section.
+
+As in TMC, in this design we have _downsync_ and _upsync_ --- but they
+are a little more complicated here.  Downsync involves propagation of
+desired state from workload management workspace through mailbox
+workspaces to edge and return/summarization of reported state.  Upsync
+involves return/summarization of desired and reported state of objects
```
ezrasilvera, line 178: Doesn't it also include the upsync of the schema itself, before we upsync the resources themselves ?

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
MikeSpreitzer: The API objects in the workload management workspaces are the inventory of workload.

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
MikeSpreitzer: I thought it was obvious that "we could start by not doing this" implies that we can incrementally progress through doing it for the various resources that need it.  The resources that need denaturing/renaturing are a finite set, so I think we should aim for eventually implementing the whole set.

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
MikeSpreitzer: Not necessarily. The resource definitions might exist in the center as consequences of CRDs or APIBindings, for example, as discussed elsewhere.

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
MikeSpreitzer: The latest push revises this wording.

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
ezrasilvera: not very clear, but also not very important as I think we will have some of this already implemented for phase 1

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
```diff
@@ -600,14 +685,21 @@ following four parts.
...
+syncer.
+
+The EMC syncers will differ from the plain unspoiled TMC syncers in
+the following ways.
 
 - Create self-sufficient edge clusters.
 - Re-nature objects that edge-mc forcibly denatures at the center.
 - Return reported state from associated objects.
+- (Maybe) not require the SyncTarget to identify all the resources
+  involved.
```
clubanderson, line 702: is there a requirement for the upsync mechanism to provide a view/list of the CRDs it knows about?  How does the syncer know which CRDs are supposed to be upsynced?  Where are the new CRDs registered/stored?

MikeSpreitzer: Upsync cannot propagate an object if the destination (i.e., mailbox workspace) does not have the definition of the object's type.

You may ask of downsync, and of upsync, whether it implicitly propagates a CRD when it is told to propagate an object whose type is defined by that CRD.  I have been assuming "yes" is the best answer.  I also have been thinking that we could start with "no", with the attendant expectation that EdgePlacement objects are written to select the needed CRD objects (as indeed the analogous work from TRL has been doing).

clubanderson: im not following you on this one.  Do you have an example for what you are referring to?

MikeSpreitzer: Let me try again to answer your original questions more directly.

- Q: is there a requirement for the upsync mechanism to provide a view/list of the CRDs it knows about?  A: I presume you remember that CRD != CR.  We have clearly heard that the C2P controller should be able to read the policy report CRs from the mailbox workspaces.  I expect that the C2P controller will have compiled-in knowledge of the definition of policy reports, so will not need to read the CRDs that define policy reports.  I also think that, for general usage, this PoC needs to provide a way for the user to read whatever workload objects they want out of the mailbox workspaces, including cluster-scoped objects, which includes CRDs.
- Q:  How does the syncer know which CRDs are supposed to be upsynced?  A: I propose that our first answer is that the EdgePlacement selects those CRDs for upsync.
- Q:  Where are the new CRDs registered/stored? A: upsync is from edge cluster to corresponding mailbox workspace.

Remember also that in our latest thinking about how to put C2P on this PoC, we will _not_ be upsyncing any resource definitions; Helm will install all the needed CRDs into the workload management workspace and they will be downsynced from there.

MikeSpreitzer: - Q: how does the syncer know that there is a CR waiting to be transported upstream?  A: The answer comes in several interacting parts.  It starts with the EdgePlacement object holding rules for identifying what should be upsynced.  The placement translator translates the EdgePlacement objects into syncer config objects (the current draft of that is in https://github.com/kcp-dev/edge-mc/blob/main/pkg/syncer/apis/edge/v1alpha1/edge-syncconfig.go , and it will probably move into https://github.com/kcp-dev/edge-mc/tree/main/pkg/apis/edge/v1alpha1 so that all the datatypes for the `edge.kcp.io` API group are in one directory), which have their own way of expressing how to identify the objects to upsync.  The syncer reads those expressions and evaluates them against the objects in the edge cluster to determine which ones should be upsynced.
-  Q: How do we prevent sending the same CR upstream more than once?  A: Since we are doing state-based management, the primary questions should be about state rather than transitions.  It is not incorrect to write the same state to the same place multiple times, although it would have performance downsides.  The initial implementation of the EMC syncer simply does periodic propagation without regard for suppressing duplicate actions.  Hopefully later implementations will act more efficiently.
- Q:  How do we react to a CR that is 'lost' upstream? (can this condition happen at all - I think not, but asking to be complete here)  A: Since we are writing state-based management code this is inherently part of the problem to solve.  Remember also that this stuff works in an eventually consistent style.  Whenever something is not as it should be, the implementation should eventually get it to a good state.

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
```diff
@@ -122,9 +122,37 @@ what should be returned as well as implementing it.  This will
...
+provided that there is not a conflict with an object of the same name
+that is positively desired in the center.  In particular, these are:
+`ClusterRole`, `ClusterRoleBinding`, `Role`, `RoleBinding`,
+`ServiceAccount`.
+
+For some kinds of object, lack of denaturing/renaturing means that
+edge-mc will simply not be able to support workloads that contain such
+objects.  These are: `MutatingWebhookConfiguration`,
+`ValidatingWebhookConfiguration`, `LimitRange`, `ResourceQuota`.
+
```
pdettori, line 146: Are these objects currently affecting the behavior of a kcp workspace? Does kcp core has controllers watching and reconciling these objects?

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
```diff
@@ -134,17 +162,40 @@ cluster, and the summarization part will affect the status summarizer.
 
 ## Design overview
 
-In very brief: the design is to reduce each edge placement problem to
-many instances of kcp's TMC problem.
+In very brief: the design is to reduce the edge placement problem to
+kcp's TMC problem, using as many "anycast" TMC Placement objects (each
```
pdettori, line 166: perhaps better `the design is to reduce` -> `the design goal is to reduce`

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
MikeSpreitzer: Doing `kubectl api-resources` on a kcp workspace will list those, so I expect that the kcp server is giving them interpretations.

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
MikeSpreitzer: Reworded in latest revision.

MikeSpreitzer:
The force-push to 493d4de was a rebase onto `main` to pick up the merge of #241 .

docs/content/en/docs/Coding Milestones/PoC2023q1/outline.md
clubanderson: lgtm - thank you for the explanation

PR #258, opened 2023-03-22
📖  add edge syncer design doc

yuji-watanabe-jp:
## Related issue(s)

Fixes #


docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+#### Registering Edge Syncer on a p-cluster
+
+Edge-syncer can be deployed on p-cluster easily by the following steps.
+1. Create SyncTarget and Location
+  - Mailbox controller creates mailbox workspace automatically. 
+  - Get mailbox workspace id
+2. Use command to register edge-syncer
+   ```console
+   kubectl ws <mb-ws name>
+   kcp workload edge-sync <EM Sync Target name> --image <EM Syncer Image> -o edge-syncer.yaml
```
MikeSpreitzer, line 23: I am not sure what you mean by the `kcp` command here.  kcp-dev/kcp/cmd/kcp does not have a "workspace" subcommand.  I expect this is a typo for `kubectl kcp`?

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+![edge-syncer boot](./images/edge-syncer-boot.png)
+
+#### Deploy workload objects from edge-mc to p-cluster
+
+To deploy resources to p-clusters, create the following in workload management workspace
+- workload objects
+  - Some objects are denatured if needed.
+  - Other objects are as it is
+- APIExport/API Schema corresponding to CRD
+  - TBD: Conversion from CRD to APIExport/APISchema could be automated by using MutatingAdmissionWebhook on workload management workspace.
```
MikeSpreitzer, line 36: FYI we have automation for this in https://github.com/kcp-dev/edge-mc/blob/1cec113d1cfe690a1abbf138361ad3d7232e808b/hack/update-codegen-crds.sh#L57

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+- EdgeSyncConfig (This is going to be replaced with EdgePlacement spec)
```
MikeSpreitzer, line 37: More precisely, the placement translator will create these based on the EdgePlacement objects and what they match.

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+- EdgeSyncConfig (This is going to be replaced with EdgePlacement spec) 
+
+![edge-syncer deploy](./images/edge-syncer-deploy.png)
+
+After this, Edge-mc will put the following in the mailbox workspace.
+- Workload objects (both denatured one and not-denatured one)
+- API Binding (to API Export on the workload management workspace)
+- EdgeSyncConfig CR
+
+**TODO**: This is something we should clarify..e.g. which existing controller(s) in edge-mc will cover, or just create a new controller to handle uncovered one.
```
MikeSpreitzer, line 46: I think it is the placement translator's job to produce put the workload objects and syncer config into the mailbox workspaces.

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+
+After this, Edge-mc will put the following in the mailbox workspace.
+- Workload objects (both denatured one and not-denatured one)
+- API Binding (to API Export on the workload management workspace)
+- EdgeSyncConfig CR
+
+**TODO**: This is something we should clarify..e.g. which existing controller(s) in edge-mc will cover, or just create a new controller to handle uncovered one. 
+
+#### EdgeSyncConfig
+- The example is [here](https://gist.github.com/yuji-watanabe-jp/dfa9cdf35b06f612818446c2c3eb7969).
```
MikeSpreitzer, line 49: A pointer to the datatype would help too.

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+- The CR here is used from edge syncer. 
+- The CR is placed at mb-ws to define
+  - object selector
+  - need of renaturing
```
MikeSpreitzer, line 53: I was expecting that the need for denaturing and renaturing is not something that needs to be configured.

In the case of C2P, if we install Kyverno with an Operator, then the Operator will install the Kyverno CRDs in the edge cluster and so those CRDs will not need to be downsynced, right?  If we follow this approach then there is no need for configurable denaturing/renaturing, right?

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+  - need of status upsync for downsyncing
```
MikeSpreitzer, line 54: Is there a reason that this should be configurable?  Why not make this always on?

BTW, "upsync" seems to be specifically about objects that are born on the edge cluster; it might be clearer here to refer to something like "return of reported state".

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+  - need of delete propagation for downsyncing
```
MikeSpreitzer, line 55: Is there a reason to not have this always on?

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+  - Interval of downsyncing/upsyncing
+- The CR is managed by edge-mc
```
MikeSpreitzer, line 57: Specifically, the placement translator.

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+  - At the initial implementation before edge-mc side controller become ready, we assume EdgeSyncConfig is on workload management workspace (wm-ws), and then which will be copied into mb-ws like other workload objects.
+  - This should be changed to be generated according to EdgePlacement spec. 
+- This CR is a placeholder for defining how edge-syncer behaves, and will be extended/splitted/merged according to further design discussion.
```
MikeSpreitzer, line 60: I am confident that we will want a configuration object for the syncer.  I think you have made a great start on defining that.

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Downsyncing
+
+- Edge syncer does downsyncing, which copy workload objects on mailbox workspace to p-cluster
+- If workload objects are deleted on mailbox workspace, the corresponding objects on the p-cluster will be also deleted according to EdgeSyncConfig. 
+- EdgeSyncConfig specifies which objects should be downsynced.
+  - object selector: group, version, kind, name, namespace (for namespaced objects), label, annotation
```
MikeSpreitzer, line 67: Think about verbosity here.  For example, if you want to downsync all the ConfigMap objects in namespace `kyverno` then there is no need to identify them by name as well.  Do you have actual cases where you want to downsync a proper subsset of namespaced objects of a certain kind in a certain namespace?

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Status Upsyncing
+- Edge syncer does upsyncing status of downsynced objects at p-cluster to the status of objects on the mailbox workspace periodically. 
+  - TODO: Failing status upsync in some resources (e.g. deployment and service). Need more investigation. 
+- Status upsyncing on/off is configurable in EdgeSyncConfig. (default is on)
+
+#### Resource Upsyncing
+- Edge syncer does upsyncing resources at p-cluster to the corresponding mailbox workspace periodically. 
+- EdgeSyncConfig specifies which objects should be upsynced from p-cluster.
+  - object selector: group, version, kind, name, namespace (for namespaced objects), label, annotation (, and more such as ownership reference?)
```
MikeSpreitzer, line 86: How much expressiveness is needed for C2P?  Would it suffice to identify the objects to upsync by API group, kind/resource, and namespace (for the ones that are namespaced)?  If not, what is the minimal additional restriction that is needed?

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+
+![edge-syncer boot](./images/edge-syncer-boot.png)
+
+#### Deploy workload objects from edge-mc to p-cluster
+
+To deploy resources to p-clusters, create the following in workload management workspace
+- workload objects
+  - Some objects are denatured if needed.
+  - Other objects are as it is
+- APIExport/API Schema corresponding to CRD
```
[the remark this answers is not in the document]
MikeSpreitzer: BTW, I think it should be the mailbox controller's job to put into the mailbox workspace an APIBinding that picks up the definition of syncer configuration objects.

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
yuji-watanabe-jp: Yes, this is typo. thank you!

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
yuji-watanabe-jp: @MikeSpreitzer thank you! I updated the description.

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
yuji-watanabe-jp: just for reducing the unnecessary returns and propagation in some cases.

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
yuji-watanabe-jp: > downsync all the ConfigMap objects in namespace kyverno

```
kind=configmap, namespace=kyverno
```

> downsync only a specific secret `mysecret` in namespace `my-ns`
```
kind=secret, name=mysecret, namespace=my-ns
```

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
yuji-watanabe-jp: Thank you! I added the description.

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
yuji-watanabe-jp: Thank you. I changed to "return of reported state".

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### EdgeSyncConfig
+- The example is [here](https://gist.github.com/yuji-watanabe-jp/dfa9cdf35b06f612818446c2c3eb7969).
+- The CR here is used from edge syncer. 
+- The CR is placed at mb-ws to define
+  - object selector
+  - need of renaturing
+  - need of status upsync for downsyncing
+  - need of delete propagation for downsyncing
+  - Interval of downsyncing/upsyncing
```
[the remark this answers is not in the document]
yuji-watanabe-jp: Thank you. we should move to informers.

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
yuji-watanabe-jp: > I was expecting that the need for denaturing and renaturing is not something that needs to be configured. There is a fixed plan outlined in https://docs.kcp-edge.io/docs/coding-milestones/poc2023q1/outline/#data-objects .

Yes, in that case, the fixed plan can be encoded as a default configuration of edge syncer, so need to be configured specifically unless required.

> In the case of C2P, if we install Kyverno with an Operator, then the Operator will install the Kyverno CRDs in the edge cluster and so those CRDs will not need to be downsynced, right? If we follow this approach then there is no need for configurable denaturing/renaturing, right?

In this case, the question is how to install operator. 

> Using an Operator to install CRDs on the edge cluster creates a new category of resources: those that are neither built into the edge cluster nor defined by a downsynced CRD. I think we can start with a hack, making the platform config say that those resources are built into the edge clusters. A more accurate way of handling them would be a way of declaring in the workload management workspace that certain resources will eventually be defined in the edge cluster as a consequence of other things.

Yes, this is the case CRD is created in p-cluster side. We will need to think about this situation later. (maybe post Apr. 27)
- Currently, assuming corresponding API Binding is set on mailbox workspace for upsyncing
- In some case, CRD is created in p-cluster, but CR is created/updated in mailbox workspace for downsync. (typical example is OPA/Gatekeeper Constraint CR/CRD). Currently, assuming the same as above.

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
MikeSpreitzer: Can you give examples?

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
MikeSpreitzer: Do you have an actual namespace in a workload management workspace from which you want to downsync some of the Secret objects but not all of them?

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
MikeSpreitzer: I understand how that file works.  My question is whether something simpler and shorter would work.  For example, does something go wrong in the case of actual C2P usage if that file can not contain object names and instead syncing simply applies to all objects that meet the other criteria?

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
yuji-watanabe-jp: if CRD is delivered to p-cluster, and many CRs are created. Even if the CRD is deleted at mailbox, I may want to keep CRD on p-cluster side, because many CRs cannot live without the CRD.

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
yuji-watanabe-jp: e.g. secret should be downsycned selectively, not all.

docs/content/en/docs/Coding Milestones/PoC2023q1/edge-syncer.md
yuji-watanabe-jp: depending on what else is in the namespace. if no other secrets does not exist there, I can send all. if some secrets such as service account token exists, I do not want to send all.

MikeSpreitzer (approved):
There is more work to be done.  I am approving this as progress.

/lgtm

yuji-watanabe-jp:
@clubanderson Thank you so much!

PR #255, opened 2023-03-21
🐛 Fix bugs in example scenario

MikeSpreitzer:
## Summary
This PR fixes bugs discovered while testing the nascent placement translator (#245).

## Related issue(s)

Fixes #

PR #259, opened 2023-03-22
✨ Initial implementation of Edge syncer

yana1205:
## Summary
This is the initial implementation of Edge syncer required for prototype of C2P on KCP.
 
## Related issue(s)

Fixes #


MikeSpreitzer:
BTW, #255 has merged now, this PR can be rebased to pick that up.

pkg/syncer/README.md
```diff
@@ -0,0 +1,225 @@
...
+    ```
+
+### See policy reports generated at p-cluster via API Export on workload management workspace.
+1. In the previous case, PolicyReport CRD is deployed as a CRD. In order to share the API across workspaces, we define PolicyReport API as APIBiinding
+1. Go to workload management workspace (`edge`)  
+    ```
+    $ kubectl kcp ws root:edge
+    Current workspace is "root:edge".
+    ```
+1. Create APIResourceSchema and APIExpose for PolicyReport CRD
```
MikeSpreitzer, line 165: s/APIExpose/APIExport/

pkg/syncer/README.md
```diff
@@ -0,0 +1,225 @@
...
+      reference:
+        export:
+          path: root:edge
+          name: policy-report
+    EOL
+    ```
+1. Denature PolicyReport CRD in Kyverno Helm chart
+    ```
+    Name of CRD: policyreports.wgpolicyk8s.io => policyreports.wgpolicyk8s.io.denatured
+    Group in CRD: wgpolicyk8s.io.denatured => wgpolicyk8s.io
```
MikeSpreitzer, line 191: I do not understand exactly what is being said here.  Is this an English representation of some editing activity?

pkg/syncer/apis/edge/register.go
```diff
@@ -0,0 +1,21 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package edge
+
+const (
+	GroupName = "edge.kcp.io"
```
MikeSpreitzer, line 20: 1. We keep a 1:1 correspondence between API group and source directory.
2. The API group source directories appear under https://github.com/kcp-dev/edge-mc/tree/main/pkg/apis .

pkg/syncer/apis/edge/v1alpha1/edge-syncconfig.go
```diff
@@ -0,0 +1,78 @@
...
+	Kind string `json:"kind,omitempty"`
+
+	// Group of down/up synced resource
+	Group string `json:"group,omitempty"`
+
+	// Version of down/up synced resource
+	Version string `json:"version,omitempty"`
+
+	// Name of down/up synced resource
+	Name string `json:"name,omitempty"`
```
MikeSpreitzer, line 46: When there is more than one object of a given kind in a given namespace, it would be more compact if a list of names could appear here.  If all objects of a given kind in a given namespace are desired, it would be more compact if `"*"` could appear here.

pkg/syncer/controller/controller.go
```diff
@@ -0,0 +1,186 @@
...
+		runtime.HandleError(err)
+		return
+	}
+
+	shared.WithQueueKey(logger, key).V(2).Info("queueing SyncConfig")
+	c.queue.Add(key)
+}
+
+// Start starts the controller workers.
+func (c *controller) Start(ctx context.Context, numThreads int) {
```
MikeSpreitzer, line 111: A better name for this method is "Run".  The word "start" does not imply waiting until completion, but "run" does.

pkg/syncer/controller/controller.go
```diff
@@ -0,0 +1,186 @@
...
+	defer logger.Info("Shutting down controller")
+
+	for i := 0; i < numThreads; i++ {
+		go wait.UntilWithContext(ctx, c.startWorker, time.Second)
+	}
+
+	<-ctx.Done()
+}
+
+func (c *controller) startWorker(ctx context.Context) {
```
MikeSpreitzer, line 127: This also is a "run" rather than a "start".

pkg/syncer/controller/controller.go
```diff
@@ -0,0 +1,186 @@
...
+			return nil // object deleted before we handled it
+		}
+		return err
+	}
+
+	downsyncerError := c.downSyncer.ReInitializeClients(syncConfig.Spec.DownSyncedResources, syncConfig.Spec.Conversions)
+	upSyncerError := c.upSyncer.ReInitializeClients(syncConfig.Spec.UpSyncedResources, syncConfig.Spec.Conversions)
+
+	if downsyncerError != nil && upSyncerError != nil {
+		return fmt.Errorf("failed to reinitialize downsyncer and upsyncer")
```
MikeSpreitzer, line 179:
```suggestion
		return fmt.Errorf("failed to reinitialize downsyncer (%w) and upsyncer (%w)", downsyncerError, upSyncerError)
```

pkg/syncer/apis/edge/v1alpha1/edge-syncconfig.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Namespace of down/up synced resource if it's not cluster scoped resource
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// Resource to be renatured
+type EdgeSynConversion struct {
```
MikeSpreitzer, line 54: Some sticklers may ask you to put the missing "c" in that name, but I will not.

pkg/syncer/controller/controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+				if _err != nil {
+					if errors.IsNotFound(_err) {
+						return false
+					} else {
+						err = _err
+						return true
+					}
+				}
+				return syncConfig != nil
+			}, wait.ForeverTestTimeout, 1*time.Second)
```
MikeSpreitzer, line 101: Why not do the usual wait for `syncConfigInformer.HasSynced()` here?

pkg/syncer/controller/controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+			assert.NoError(t, err)
+
+			controller.Start(ctx, 1)
```
MikeSpreitzer, line 104: Because this method waits until the context is closed, this test hangs at this point.

pkg/syncer/controller/controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+			require.Eventually(t, func() bool {
```
MikeSpreitzer, line 105: If the previous statement did not hang, this one would.  It looks like the test is not fully written yet.  That's OK.

pkg/syncer/syncer.go
```diff
@@ -0,0 +1,153 @@
...
+
+package syncer
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	edgev1alpha1 "github.com/kcp-dev/edge-mc/pkg/syncer/apis/edge/v1alpha1"
+	kcpclientset "github.com/kcp-dev/edge-mc/pkg/syncer/client/clientset/versioned"
```
MikeSpreitzer, line 25: I think it would be clearer if the name of this import used "syncer" in place of "kcp".  Same for the informers import.

pkg/syncer/apis/edge/v1alpha1/edge-syncconfig.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +crd
+// +genclient
+// +genclient:nonNamespaced
+// +kubebuilder:resource:scope=Cluster,shortName=esc
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type EdgeSyncConfig struct {
```
MikeSpreitzer, line 12: BTW, you do not need "Edge" in the name of this type, it is not going in the same package as a TMC or any other syncer config.

pkg/syncer/syncer.go
```diff
@@ -0,0 +1,153 @@
...
+		return err
+	}
+	syncConfigClient := syncConfigClientSet.EdgeV1alpha1().EdgeSyncConfigs()
+	// syncConfigInformerFactory to watch a certain syncConfig on upstream
+	syncConfigInformerFactory := kcpinformers.NewSharedScopedInformerFactoryWithOptions(syncConfigClientSet, resyncPeriod, kcpinformers.WithTweakListOptions(
+		func(listOptions *metav1.ListOptions) {
+			listOptions.FieldSelector = fields.OneTermEqualSelector("metadata.name", cfg.SyncTargetName).String()
+		},
+	))
+	syncConfigInformer := syncConfigInformerFactory.Edge().V1alpha1().EdgeSyncConfigs()
```
MikeSpreitzer, line 75: https://github.com/kcp-dev/edge-mc/pull/206#discussion_r1128492162

pkg/syncer/syncer.go
```diff
@@ -0,0 +1,153 @@
...
+
+	syncConfigInformer.Lister().List(labels.Everything()) // TODO: Remove (for now, need to invoke List at once)
```
MikeSpreitzer, line 77: Why?  What difference does this make?

pkg/syncer/apis/edge/v1alpha1/edge-syncconfig.go
clubanderson: I would add a "c" here

pkg/syncer/syncer.go
```diff
@@ -0,0 +1,153 @@
...
+
+	syncConfigInformerFactory.Start(ctx.Done())
+	syncConfigInformerFactory.WaitForCacheSync(ctx.Done())
+
+	upstreamConfig := rest.CopyConfig(cfg.UpstreamConfig)
+	rest.AddUserAgent(upstreamConfig, "edge-mc#syncer/"+kcpVersion)
```
MikeSpreitzer, line 83: This makes `upstreamConfig` the same as `bootstrapConfig`, right?  Why duplicate this code?

pkg/syncer/syncer.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	downstreamConfig := rest.CopyConfig(cfg.DownstreamConfig)
+	rest.AddUserAgent(downstreamConfig, "edge-mc#syncer/"+kcpVersion)
+	downstreamDyClient, err := dynamic.NewForConfig(downstreamConfig)
+	if err != nil {
+		return err
+	}
+	downstreamDiscoveryClient := discovery.NewDiscoveryClientForConfigOrDie(downstreamConfig)
+	downstreamClientFactory, err := syncers.NewClientFactory(logger, downstreamDyClient, downstreamDiscoveryClient)
```
MikeSpreitzer, line 101: "Dy" does not make a great contrast with "Discovery".  Replacing "Dy" with "Dyn" or "Dynamic" would be clearer.

pkg/syncer/syncer.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return err
+	}
+
+	go controller.Start(ctx, numSyncerThreads)
+	startSync(ctx, logger, cfg, syncConfigInformer.Lister(), upSyncer, downSyncer)
+	return nil
+}
+
+func startSync(ctx context.Context, logger klog.Logger, cfg *SyncerConfig, syncConfigLister edgev1alpha1listers.EdgeSyncConfigLister, upSyncer *syncers.UpSyncer, downSyncer *syncers.DownSyncer) {
```
MikeSpreitzer, line 125: This also is "run" rather than "start".

pkg/syncer/syncers/client.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (c *Client) IsNamespaced() bool {
+	return c.scope == meta.RESTScopeNamespace
+}
+
+func (c *Client) Create(resource edgev1alpha1.EdgeSyncConfigResource, unstObj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	var createdObj *unstructured.Unstructured
+	var err error
+	if c.IsNamespaced() {
+		createdObj, err = c.ResourceClient.Namespace(resource.Namespace).Create(context.Background(), unstObj, v1.CreateOptions{})
```
MikeSpreitzer, line 43: I recommend filling in the `FieldManager` value for the various meta/v1/XXXOptions

pkg/syncer/syncers/client.go
```diff
@@ -0,0 +1,83 @@
...
+	var updatedObj *unstructured.Unstructured
+	var err error
+	if c.IsNamespaced() {
+		// updatedObj, err = c.ResourceClient.Namespace(resource.Namespace).Apply(context.Background(), unstObj.GetName(), unstObj, v1.ApplyOptions{FieldManager: "application/apply-patch"})
+		updatedObj, err = c.ResourceClient.Namespace(resource.Namespace).UpdateStatus(context.Background(), unstObj, v1.UpdateOptions{})
+	} else {
+		updatedObj, err = c.ResourceClient.UpdateStatus(context.Background(), unstObj, v1.UpdateOptions{})
+	}
+	return updatedObj, err
+}
```
MikeSpreitzer, line 83: No Delete?

pkg/syncer/apis/edge/v1alpha1/edge-syncconfig.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Upsncyed resource list
+	UpSyncedResources []EdgeSyncConfigResource `json:"upSyncedResources,omitempty"`
+
+	// Conversions
+	Conversions []EdgeSynConversion `json:"conversions,omitempty"`
+}
+
+// Resource specifies down/up synced resource with exact GVK and name (and namespace if not cluster scoped resource)
+type EdgeSyncConfigResource struct {
```
MikeSpreitzer, line 35: FYI this is not a typical usage of the word "Resource" in the world of Kubernetes code.  While in the wide world of the web, "resource" is anything that has a URL or URI, in Kubernetes code the word "resource" usually has a more specific meaning: a kind of object.  That is, a collection of objects rather than a single object.  Of course, the code base is not entirely consistent in this regard.  The public-facing documentation is worse, it is definitely mixed-up in this regard.

I recommend using the term "object" in the code when an individual API object is meant, reserving "resource" for a kind of object.

Here is one place where the rubber meets the road.  As you know, "kind" and "resource" are two equivalent but distinct way of identifying a sort of object.  In the XXXResource struct at hand, one would expect that the Kind and Name fields hold the kind and resource, respectively.  For example, as in https://github.com/kubernetes/apimachinery/blob/v0.26.3/pkg/apis/meta/v1/types.go#L1081 .  But that is not what is going on here.

pkg/syncer/apis/edge/v1alpha1/edge-syncconfig.go
```diff
@@ -0,0 +1,78 @@
...
+	// Kind of down/up synced resource
+	Kind string `json:"kind,omitempty"`
```
MikeSpreitzer, line 37: The implementation of the syncer would be simpler if this field held the resource rather than the kind.  For example, "deployments" rather than "Deployment".  That would remove the need to go through "rest mapping" to get from kind to resource.

pkg/syncer/syncers/downsyncer.go
```diff
@@ -0,0 +1,181 @@
...
+func (ds *DownSyncer) ReInitializeClients(syncedResources []edgev1alpha1.EdgeSyncConfigResource, conversions []edgev1alpha1.EdgeSynConversion) error {
+	return initializeClients(ds.logger, syncedResources, ds.upstreamClientFactory, ds.downstreamClientFactory, ds.upstreamClients, ds.downstreamClients, conversions)
+}
+
+func (ds *DownSyncer) getClients(resource edgev1alpha1.EdgeSyncConfigResource, conversions []edgev1alpha1.EdgeSynConversion) (*Client, *Client, error) {
+	return getClients(resource, ds.upstreamClients, ds.downstreamClients, conversions)
+}
+
+func (ds *DownSyncer) SyncOne(resource edgev1alpha1.EdgeSyncConfigResource, conversions []edgev1alpha1.EdgeSynConversion) error {
+	ds.logger.V(3).Info(fmt.Sprintf("downsync '%s'", resourceToString(resource)))
```
MikeSpreitzer, line 70: If you use `%q` instead of `%s` then you do not have to supply the quotes in the source and you also get the usual escaping for special characters.

pkg/syncer/syncers/downsyncer.go
```diff
@@ -0,0 +1,181 @@
...
+		ds.logger.Error(err, fmt.Sprintf("failed to get client '%s'", resourceToString(resource)))
+		return err
+	}
+	resourceForUp := convertToUpstream(resource, conversions)
+	ds.logger.V(3).Info(fmt.Sprintf("  get '%s' from upstream", resourceToString(resourceForUp)))
+	upstreamResource, err := upstreamClient.Get(resourceForUp)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			ds.logger.V(3).Info(fmt.Sprintf("  not found '%s' in upstream", resourceToString(resourceForUp)))
+			ds.logger.V(3).Info(fmt.Sprintf("  skip downsync '%s'", resourceToString(resourceForUp)))
```
MikeSpreitzer, line 82: Should there be a downstream delete here?

pkg/syncer/syncers/downsyncer.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+	}
+
+	resourceForDown := convertToDownstream(resource, conversions)
+	ds.logger.V(3).Info(fmt.Sprintf("  get '%s' from downstream", resourceToString(resourceForDown)))
+	downstreamResource, err := downstreamClient.Get(resourceForDown)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			// create
+			ds.logger.V(3).Info(fmt.Sprintf("  create '%s' in upstream since it's not found", resourceToString(resourceForDown)))
```
MikeSpreitzer, line 96: Isn't the creation being done in the downstream?

pkg/syncer/syncers/downsyncer.go
```diff
@@ -0,0 +1,181 @@
...
+				return err
+			}
+		} else {
+			ds.logger.Error(err, fmt.Sprintf("failed to get resource from downstream '%s'", resourceToString(resourceForDown)))
+			return err
+		}
+	} else {
+		if downstreamResource != nil {
+			// update
+			ds.logger.V(3).Info(fmt.Sprintf("  update '%s' in upstream since it's found", resourceToString(resourceForDown)))
```
MikeSpreitzer, line 111: The update is in the downstream, right?

pkg/syncer/syncers/downsyncer.go
```diff
@@ -0,0 +1,181 @@
...
+			upstreamResource.SetResourceVersion("")
+			upstreamResource.SetUID("")
+			upstreamResource.SetManagedFields(nil)
+			applyConversion(upstreamResource, resourceForDown)
+			if _, err := downstreamClient.Update(resourceForDown, upstreamResource); err != nil {
```
MikeSpreitzer, line 116: You could save significant runtime costs if this were conditional on whether there is actually a difference to be made here.

pkg/syncer/syncers/downsyncer.go
```diff
@@ -0,0 +1,181 @@
...
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			// create
+			ds.logger.V(3).Info(fmt.Sprintf("  create '%s' in upstream since it's not found", resourceToString(resourceForDown)))
+			upstreamResource.SetResourceVersion("")
+			upstreamResource.SetUID("")
+			applyConversion(upstreamResource, resourceForDown)
+			if _, err := downstreamClient.Create(resourceForDown, upstreamResource); err != nil {
+				ds.logger.Error(err, fmt.Sprintf("failed to create resource to downstream '%s'", resourceToString(resourceForDown)))
+				return err
```
MikeSpreitzer, line 102: It would be great to return this information upstream.  In my thinking about summarization, I am hoping to be able to summarize it.  You can see my first draft proposal for summarization in https://docs.google.com/document/d/1yY-XgOtCnOPKpVdxdPVhH5ixRN6IWQQe0s3lip1ID0Y (view it as a member of kcp-dev), and I am in the midst of creating a second draft to propose to edge-mc.  Do not worry about it in this PR.

pkg/syncer/syncers/downsyncer.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	resourceForDown := convertToDownstream(resource, conversions)
+	downstreamResource, err := downstreamClient.Get(resourceForDown)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			ds.logger.V(3).Info(fmt.Sprintf("  not found '%s' in downstream", resourceToString(resourceForDown)))
+			ds.logger.V(3).Info(fmt.Sprintf("  skip status upsync '%s'", resourceToString(resourceForDown)))
+			return nil
+		} else {
+			ds.logger.Error(err, fmt.Sprintf("failed to get resource from upstream '%s'", resourceToString(resourceForDown)))
```
MikeSpreitzer, line 142: you mean from downstream here, right?

pkg/syncer/syncers/downsyncer.go
```diff
@@ -0,0 +1,181 @@
...
+			return err
+		}
+	}
+	status, found, err := unstructured.NestedMap(downstreamResource.Object, "status")
+	if err != nil {
+		ds.logger.Error(err, fmt.Sprintf("failed to extract status from downstream object '%s'", resourceToString(resourceForDown)))
+		return err
```
MikeSpreitzer, line 149: Do you want a retry in this case?

pkg/syncer/syncers/downsyncer.go
```diff
@@ -0,0 +1,181 @@
...
+			return err
+		}
+	}
+	_, found, err = unstructured.NestedMap(upstreamResource.Object, "status")
+	if err != nil {
+		ds.logger.Error(err, fmt.Sprintf("failed to extract status from upstream object '%s'", resourceToString(resourceForUp)))
+		return err
+	}
+	upstreamResource.Object["status"] = status
+	updatedResource := unstructured.Unstructured{
```
MikeSpreitzer, line 172: Why is this copy being made?

pkg/syncer/syncers/downsyncer.go
```diff
@@ -0,0 +1,181 @@
...
+		Object: upstreamResource.Object,
+	}
+	applyConversion(upstreamResource, resourceForUp)
```
MikeSpreitzer, line 175: Did you mean to apply conversion to `updatedResource` ?

pkg/syncer/syncers/upsyncer.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (us *UpSyncer) SyncOne(resource edgev1alpha1.EdgeSyncConfigResource, conversions []edgev1alpha1.EdgeSynConversion) error {
+	us.logger.V(3).Info(fmt.Sprintf("upsync '%s'", resourceToString(resource)))
+	upstreamClient, k8sClient, err := us.getClients(resource, conversions)
+	if err != nil {
+		us.logger.Error(err, fmt.Sprintf("failed to get client '%s'", resourceToString(resource)))
+		return err
+	}
+	resourceForDown := convertToUpstream(resource, conversions)
+	us.logger.V(3).Info(fmt.Sprintf("  get '%s' from downstream", resourceToString(resourceForDown)))
```
MikeSpreitzer, line 76: Structured logging?  If not, `%q`?

pkg/syncer/syncers/upsyncer.go
```diff
@@ -0,0 +1,128 @@
...
+	downstreamResource, err := k8sClient.Get(resourceForDown)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			us.logger.V(3).Info(fmt.Sprintf("  not found '%s' in downstream", resourceToString(resourceForDown)))
+			us.logger.V(3).Info(fmt.Sprintf("  skip upsync '%s'", resourceToString(resourceForDown)))
```
MikeSpreitzer, line 81: Delete?

pkg/syncer/syncers/upsyncer.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (us *UpSyncer) SyncOne(resource edgev1alpha1.EdgeSyncConfigResource, conversions []edgev1alpha1.EdgeSynConversion) error {
+	us.logger.V(3).Info(fmt.Sprintf("upsync '%s'", resourceToString(resource)))
+	upstreamClient, k8sClient, err := us.getClients(resource, conversions)
+	if err != nil {
+		us.logger.Error(err, fmt.Sprintf("failed to get client '%s'", resourceToString(resource)))
+		return err
+	}
+	resourceForDown := convertToUpstream(resource, conversions)
```
MikeSpreitzer, line 75: This is actually for up not down, right?

cmd/syncer/helm/converter.go
```diff
@@ -0,0 +1,192 @@
...
+
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	k8syaml "k8s.io/apimachinery/pkg/runtime/serializer/yaml"
+	"k8s.io/klog/v2"
+	sigyaml "sigs.k8s.io/yaml"
+
+	edgev1alpha1 "github.com/kcp-dev/edge-mc/pkg/syncer/apis/edge/v1alpha1"
+)
+
```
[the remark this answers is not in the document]
yana1205: No, there isn't. I will fill the command description and update README in this PR. This command is not related with Syncer but just an utility to generate EdgeSyncConfig manifest from yaml manifests returned by Helm (`helm template kyverno --dry-run`). I would like to leave this command.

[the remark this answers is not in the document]
MikeSpreitzer: FYI, two smaller PRs are easier to review than one bigger PR.

MikeSpreitzer:
BTW, I agree with the approach taken here for how to denature: change the API group from `some.original.thing` to `some.original.thing.denatured`.

pkg/syncer/README.md
yana1205: My description was poor and wrong. I would express the following replacements in CustomResourceDefinition manifest for policyreport resource definition:
1. Replace `metadata.name: policyreports.wgpolicyk8s.io` with `metadata.name: policyreports.wgpolicyk8s.io.denatured`
2. Replace `spec.group: wgpolicyk8s.io` with `spec.group: wgpolicyk8s.io.denatured`

I will update the README so. Thanks for the review.

pkg/syncer/apis/edge/register.go
yana1205: Thanks. I will move it to `/pkg/apis`. 

> for example, because you want to make a distinction between resources that are part of the interface of edge-mc vs. resources that are internal

Yes, this is also what I was wondering if syncer related APIs are part of edge-mc interface or not. EdgeSyncConfig can be referred by edge-mc controllers (PlacementTranslater?) if my understanding is correct. It makes sense for the API to be in the same api groups as edge-mc.

pkg/syncer/apis/edge/v1alpha1/edge-syncconfig.go
yana1205: Thanks. I will do it when I consolidate syncer apis into `/pkg/apis`.

pkg/syncer/syncer.go
yana1205: Without this line, nothing comes to event handlers even when I add/update/delete EdgeSyncConfig resource in k8s. I am still debugging what is the issue.

pkg/syncer/syncers/client.go
yana1205: Right. Deletion is not cared for the 1st implementation.

pkg/syncer/apis/edge/v1alpha1/edge-syncconfig.go
yana1205: Yes, I agree with. Actually, I had to write the code to go through rest mapping. Though, as a user who creates EdgeSyncConfig, I am wondering if "resource" is always represented as lower case of "kind" with "s" appended (; `strings.ToLower(kind) + "s"`). Actually, I can find "kind" in k8s yaml manifest but not find "resource". Maybe it's responsible of a user who creates EdgeSyncConfig with correct "resource" by somehow finding it from somewhere.

pkg/syncer/syncers/downsyncer.go
yana1205: Yes, I will create a gitissue for deletion support.

yana1205: I will create a gitissue for deletion support.

pkg/syncer/syncers/downsyncer.go
yana1205: Yes, correct. Will fix the comment.

pkg/syncer/syncers/downsyncer.go
yana1205: No... Maybe it would be better to have a flag in member variable of a resource client (/pkg/syncer/syncers/client.go) to indicate whether a resource has valid "status" field or not.

pkg/syncer/syncers/downsyncer.go
yana1205: Maybe this has been left unintentionally. I confirmed that the status is properly upsynced without this copy. Thanks. Will be fixed in this PR.

pkg/syncer/syncers/upsyncer.go
yana1205: Yes, should be. Will be covered in https://github.com/kcp-dev/edge-mc/pull/259#discussion_r1146017961

pkg/syncer/controller/controller_test.go
yana1205: Thanks. I was in a middle of writing the test. I have just fixed it and completed to write simple test cases.

pkg/syncer/controller/controller_test.go
yana1205: Resolved in the above fix.

pkg/syncer/controller/controller.go
yana1205: Thanks. Fixed.

pkg/syncer/README.md
yana1205: Thanks.
Fixed.

cmd/syncer/helm/converter.go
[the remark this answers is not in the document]
yana1205: Yes, I totally agree with that. I will do that from the next.

MikeSpreitzer:
There is a bit of lint to pick here before this can merge.

yana1205:
I fixed the codes following linter message and now PR checks passed. 

> BTW, https://github.com/kcp-dev/edge-mc/pull/255 has merged now, this PR can be rebased to pick that up.

I force-pushed with the PR branch rebased from the top of main.

I added thumsup mark in the comment that I fixed in this PR. For things not included in this PR, I summarized them as followup items.
- Move up syncer apis to /pkg/apis
  - https://github.com/kcp-dev/edge-mc/pull/259#discussion_r1145692371
  - https://github.com/kcp-dev/edge-mc/pull/259#discussion_r1145826529
- Support a list of names in EdgeSyncConfig
  - https://github.com/kcp-dev/edge-mc/pull/259#discussion_r1145592961
- Bug? Need to identify why informer doesn't inform anything without invoking List or Get once
  - https://github.com/kcp-dev/edge-mc/pull/259#discussion_r1145840080
- Support propagation of deletion
  - https://github.com/kcp-dev/edge-mc/pull/259#discussion_r1145921774
  - https://github.com/kcp-dev/edge-mc/pull/259#discussion_r1146017961
  - https://github.com/kcp-dev/edge-mc/pull/259#discussion_r1146058283
- Use Resource rather than Kind to make logic simpler
  - https://github.com/kcp-dev/edge-mc/pull/259#discussion_r1145644407
- When syncer failed to create/update resources in downstream, send it back to upstream 
  - https://github.com/kcp-dev/edge-mc/pull/259#discussion_r1145656244
- In order to reduce computation cost, the information if a Resource type doesn not have status field or not can be cached.
  - https://github.com/kcp-dev/edge-mc/pull/259#discussion_r1146035011
- Reduce computational cost by introducing object comparison between up/down before doing update
  - https://github.com/kcp-dev/edge-mc/pull/259#discussion_r1145652241

cmd/syncer/helm/converter.go
```diff
@@ -55,7 +56,7 @@ func main() {
 	ctx = klog.NewContext(ctx, logger)
 	_ = ctx
 	fs.VisitAll(func(flg *pflag.Flag) {
-		logger.V(1).Info("Command line flag", flg.Name, flg.Value)
+		logger.V(1).Info(fmt.Sprintf("Command line flag %s %s", flg.Name, flg.Value))
```
MikeSpreitzer, line 59: Sadly, the linter is overly picky about this.  You can tell because the linter output in this case explicitly tells you that there is a list of accepted complaints that you can update; see https://github.com/kcp-dev/edge-mc/blob/4219d05658f25bc36f0d01d4b57f8266bd566f2c/hack/verify-contextual-logging.sh#L63.  See the exceptions already listed in https://github.com/kcp-dev/edge-mc/blob/4219d05658f25bc36f0d01d4b57f8266bd566f2c/hack/logcheck.out .

cmd/syncer/helm/converter.go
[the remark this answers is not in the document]
MikeSpreitzer: Today I see no reference to this command.

```console
(venv) mspreitz@mjs12 kubestellar % find * \( -path docs/venv -prune \) -or \( -name "*.go" -exec grep helm \{\} \; -print -exec echo \; \)
	fs := pflag.NewFlagSet("helmconverter", pflag.ExitOnError)
	fs.StringVar(&pathToHelmTemplate, "path-to-helm-template", "", "path to Helm tamplate file")
cmd/syncer/helm/converter.go

(venv) mspreitz@mjs12 kubestellar % 
```

It is a problem to build a variant executable named `helm` and put it in the KubeStellar `bin` directory because that will shadow the regular `helm` for users of KubeStellar.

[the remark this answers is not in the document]
MikeSpreitzer: Today I can not find any documentation of the use of this command.

PR #261, opened 2023-03-23
✨ Reconciling logic triggered by Location

waltforme:
## Summary

This PR implements parts of (those triggered by Location) edge scheduler's reconciliation.

### DIFF
--- a/pkg/scheduler/reconcile_on_edgeplacement.go
+++ b/pkg/scheduler/reconcile_on_edgeplacement.go
@@ -83,7 +83,7 @@ func (c *controller) reconcileOnEdgePlacement(ctx context.Context, epKey string)
 
 	/*
 		On EdgePlacement change:
-		- find all its loc(s)
+		- find all its loc(s), and update store --- how to get its previous loc(s)?
 
 		- for each of the found loc, find all its st(s)
 		- for each of the found loc, compose all sp(s)
--- a/pkg/scheduler/reconcile_on_location.go
+++ b/pkg/scheduler/reconcile_on_location.go
@@ -19,33 +19,239 @@ package scheduler
 import (
 	"context"
 
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/klog/v2"
 
 	kcpcache "github.com/kcp-dev/apimachinery/v2/pkg/cache"
+	schedulingv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/scheduling/v1alpha1"
+	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
+	"github.com/kcp-dev/logicalcluster/v3"
+
+	edgev1alpha1 "github.com/kcp-dev/edge-mc/pkg/apis/edge/v1alpha1"
 )
 
 func (c *controller) reconcileOnLocation(ctx context.Context, locKey string) error {
 	logger := klog.FromContext(ctx)
-	ws, _, name, err := kcpcache.SplitMetaClusterNamespaceKey(locKey)
+	lws, _, lName, err := kcpcache.SplitMetaClusterNamespaceKey(locKey)
 	if err != nil {
 		logger.Error(err, "invalid Location key")
 		return err
 	}
-	logger = logger.WithValues("workspace", ws, "location", name)
+	logger = logger.WithValues("locationWorkspace", lws, "location", lName)
 	logger.V(2).Info("reconciling")
 
 	/*
 		On location change:
-		- find all its st(s)
+		1) find all ep(s) that selected loc
 
-		- find all its ep(s), and update store
-		- for each of its obsolete ep, remove all sp(s)
-		- for each of its ongoing ep, update all sp(s)
-		- for each of its new ep, add all sp(s)
+		2a) find all its st(s), and update store --- how to get its previous st(s)?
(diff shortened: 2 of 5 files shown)


pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -83,7 +83,7 @@ func (c *controller) reconcileOnEdgePlacement(ctx context.Context, epKey string)
 
 	/*
 		On EdgePlacement change:
-		- find all its loc(s)
+		- find all its loc(s), and update store --- how to get its previous loc(s)?
```
MikeSpreitzer, line 86: The store would be the natural place to get the previous set of matching locations

pkg/scheduler/reconcile_on_location.go
```diff
@@ -19,33 +19,239 @@ package scheduler
...
 		Need data structure:
 		- map from a location to its eps, say 'epsBySelectedLoc'
+
+		TODO(waltforme): Maybe I can merge 3b) and 3c)
 	*/
 
+	loc, err := c.locationLister.Cluster(lws).Get(lName)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil
```
MikeSpreitzer, line 65: There is work to be done in this case, right?  Updating the SinglePlacementSlice objects that formerly referenced the Location that is now deleted.

pkg/scheduler/reconcile_on_location.go
```diff
@@ -19,33 +19,239 @@ package scheduler
...
+	for _, sp := range sps.Destinations {
+		if sp.Cluster != lws || sp.LocationName != lName {
+			nextDests = append(nextDests, sp)
+		}
+	}
+	sps.Destinations = nextDests
+	return sps
+}
+
+func makeSinglePlacementsForLoc(locSelectingSts *schedulingv1alpha1.Location, sts []*workloadv1alpha1.SyncTarget) []edgev1alpha1.SinglePlacement {
```
MikeSpreitzer, line 244: In this PoC there is supposed to be only one SyncTarget per Location.  We can not prevent the state from ever violating that restriction, but this function can treat it like the (hopefully transient) error condition that it is.

pkg/scheduler/reconcile_on_location.go
```diff
@@ -19,33 +19,239 @@ package scheduler
...
+	store.epsBySelectedLoc[locKey] = epsSelectingLoc
+
+	for ep := range epsSelectedLoc {
+		if _, ok := epsSelectingLoc[ep]; !ok {
+			// 3a)
+			// an (obsolete) ep doesn't select loc anymore
+			// we need to remove all relevant sp(s) from the corresponding sps where 'relevant' means an sp has loc
+			logger.V(1).Info("stop to select Location", "edgePlacement", ep)
+			ws, _, name, err := kcpcache.SplitMetaClusterNamespaceKey(ep)
+			if err != nil {
```
MikeSpreitzer, line 113: This error condition should never arise, because your code controls all the `ep` values, right?  When something impossible happens, I think it is best to not make it abort handling other parts of the problem; better to just log here and move on.

MikeSpreitzer: Also: if the map keys were `pkg/placement.ExternalName` then there would be no parsing, making the code shorter by both (a) removing the call to the parser and (b) not having to handle the possibility of an error.

pkg/scheduler/reconcile_on_location.go
```diff
@@ -19,33 +19,239 @@ package scheduler
...
+				logger.Error(err, "invalid EdgePlacement key")
+				return err
+			}
+			currentSPS, err := c.singlePlacementSliceLister.Cluster(ws).Get(name)
+			if err != nil {
```
MikeSpreitzer, line 118: Is this really an error?  In the asynchronous world where this is set, this code code be running in a situation where that SinglePlacementSlice has not been created yet or has been deleted and needs to be re-created by this very same process.

pkg/scheduler/reconcile_on_synctarget.go
```diff
@@ -54,18 +55,23 @@ func (c *controller) reconcileOnSyncTarget(ctx context.Context, stKey string) er
...
+
+		TODO(waltforme): Maybe I can merge 3b) and 3c)
 	*/
 
 	st, err := c.synctargetLister.Cluster(stws).Get(stName)
 	if err != nil {
-		logger.Error(err, "syncTarget not found in local cache")
-		return nil
+		if errors.IsNotFound(err) {
+			return nil
```
MikeSpreitzer, line 65: This is where the code should handle deletion of a SyncTarget.

MikeSpreitzer (approved):
This is progress.
I think you are starting to see some of the duplication of logic.

/lgtm

pkg/scheduler/reconcile_on_location.go
waltforme: Yes, completed this in PR #267 .

pkg/scheduler/reconcile_on_location.go
waltforme: It sounds more natural to me to trigger the creation logic from somewhere else. That somewhere else in my mind is an extra piece of reconciliation which is triggered by SinglePlacementSlice changes. (The scheduler is watching four kinds including SPS but is not yet reconciling on SPS changes).

My thoughts is to write another piece of code to reconcile on SPS changes, in that piece of code, just inqueue an EP object to the shared workqueue. So that I can reuse the SPS creation logic living in the reconcilation on EP. Would that work?

MikeSpreitzer: Also, the scheduler is not yet adding an event handler to the SinglePlacementSlice informer.

PR #265, opened 2023-03-24
✨ First draft of implementation of mailbox informers

MikeSpreitzer:
## Summary
This PR provides the first draft of the implementation of informers for workload objects in mailbox workspaces.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/mailboxwatch/impl.go
+++ b/pkg/mailboxwatch/impl.go
@@ -1,0 +1,200 @@
+/*
+Copyright 2023 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mailboxwatch
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+
+	machmeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	k8ssets "k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/watch"
+	upstreamcache "k8s.io/client-go/tools/cache"
+
+	tenancyv1a1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+	tenancyv1a1informers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/tenancy/v1alpha1"
+	"github.com/kcp-dev/logicalcluster/v3"
+
+	"github.com/kcp-dev/edge-mc/pkg/placement"
+)
+
+func newFilteredListerWatcher(
+	mailboxWorkspacePreInformer tenancyv1a1informers.WorkspaceInformer,
+	unfiltered upstreamcache.ListerWatcher,
+) *filteredListerWatcher {
+	flw := &filteredListerWatcher{
+		unfiltered: unfiltered,
+		clusters:   k8ssets.NewString(),
+	}
+	mailboxWorkspacePreInformer.Informer().AddEventHandler(flw)
+	return flw
+}
+
+type filteredListerWatcher struct {
+	sync.Mutex
+	unfiltered upstreamcache.ListerWatcher
+	clusters   k8ssets.String
+}
+
+func (flw *filteredListerWatcher) OnAdd(obj any) {
+	flw.setInclusion(obj, true)
+}
+
+func (flw *filteredListerWatcher) OnUpdate(old, obj any) {
+	flw.setInclusion(obj, true)
+}
+
+func (flw *filteredListerWatcher) OnDelete(obj any) {
+	flw.setInclusion(obj, false)
+}
+
+func (flw *filteredListerWatcher) setInclusion(obj any, include bool) {
+	ws := obj.(*tenancyv1a1.Workspace)
+	mbwsName := ws.Name
+	mbwsNameParts := strings.Split(mbwsName, placement.WSNameSep)
+	if len(mbwsNameParts) != 2 {
+		// Only accept the workspace if its name looks like a mailbox workspace name
+		include = false
+	}
+	cluster := ws.Spec.Cluster
+	flw.Lock()
+	defer flw.Unlock()
+	if include {
+		flw.clusters.Insert(cluster)
+	} else {
+		flw.clusters.Delete(cluster)
+	}
+}
+
+type myList struct {
+	metav1.TypeMeta
+	metav1.ListMeta
+	items []runtime.Object
+}
+
+var _ upstreamcache.ListerWatcher = &filteredListerWatcher{}
+
+func (flw *filteredListerWatcher) List(options metav1.ListOptions) (runtime.Object, error) {
+	unfilteredListObj, err := flw.unfiltered.List(options)
+	if err != nil {
+		return nil, err
+	}
+	listMeta, err := machmeta.ListAccessor(unfilteredListObj)
+	if err != nil {
+		return nil, fmt.Errorf("inner Lister produced a %T (%+v), which machmeta.ListAccessor rejects: %w", unfilteredListObj, unfilteredListObj, err)
+	}
+	unfilteredList, err := machmeta.ExtractList(unfilteredListObj)
+	if err != nil {
+		return nil, fmt.Errorf("inner Lister produced a %T (%+v), which machmeta.ExtractList rejects: %w", unfilteredListObj, unfilteredListObj, err)
+	}
+	flw.Lock()
+	defer flw.Unlock()
+	filteredSlice := []runtime.Object{}
+	for _, item := range unfilteredList {
+		objm := item.(metav1.Object)
+		cluster := logicalcluster.From(objm)
+		if flw.clusters.Has(cluster.String()) {
+			filteredSlice = append(filteredSlice, item)
+		}
+	}
+	return &myList{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: unfilteredListObj.GetObjectKind().GroupVersionKind().GroupVersion().String(),
+			Kind:       unfilteredListObj.GetObjectKind().GroupVersionKind().Kind,
+		},
+		ListMeta: metav1.ListMeta{
+			ResourceVersion:    listMeta.GetResourceVersion(),
+			Continue:           listMeta.GetContinue(),
+			RemainingItemCount: listMeta.GetRemainingItemCount(),
+		},
+		items: filteredSlice,
+	}, nil
+}
+
+func (ml *myList) DeepCopyObject() runtime.Object {
+	ans := myList{
+		TypeMeta: ml.TypeMeta,
+		ListMeta: *ml.ListMeta.DeepCopy(),
+		items:    make([]runtime.Object, len(ml.items)),
+	}
+	for index, obj := range ml.items {
+		ans.items[index] = obj.DeepCopyObject()
+	}
+	return &ans
+}
+
+func (flw *filteredListerWatcher) Watch(options metav1.ListOptions) (watch.Interface, error) {
+	unfilteredWatch, err := flw.unfiltered.Watch(options)
+	if err != nil {
+		return nil, fmt.Errorf("inner Watcher failed: %w", err)
+	}
+	ctx := context.Background()
+	ctx, close := context.WithCancel(ctx)
+	ans := &myWatch{
+		unfilteredIfc:  unfilteredWatch,
+		unfilteredChan: unfilteredWatch.ResultChan(),
+		close:          close,
+		doneChan:       ctx.Done(),
+		filtered:       make(chan watch.Event),
+	}
+	go func() {
+		for {
+			select {
+			case <-ans.doneChan:
+				ans.unfilteredIfc.Stop()
+				return
+			case event, ok := <-ans.unfilteredChan:
+				if !ok {
+					ans.unfilteredIfc.Stop() // do we need this in this case?  https://kubernetes.slack.com/archives/C0EG7JC6T/p1679684882556809
+					ans.close()
+					return
+				}
+				objm := event.Object.(metav1.Object)
+				cluster := logicalcluster.From(objm)
+				pass := func() bool {
+					flw.Lock()
+					defer flw.Unlock()
+					return flw.clusters.Has(cluster.String())
+				}()
+				if pass {
+					ans.filtered <- event
+				}
+			}
+		}
+	}()
+	return ans, nil
+}
+
+type myWatch struct {
+	unfilteredIfc  watch.Interface
+	unfilteredChan <-chan watch.Event
+	close          func()
+	doneChan       <-chan struct{}
+	filtered       chan watch.Event
+}
+
+func (mw *myWatch) Stop() {
+	mw.close()
+}
+
+func (mw *myWatch) ResultChan() <-chan watch.Event {
+	return mw.filtered
+}
--- a/pkg/mailboxwatch/impl.go
+++ b/pkg/mailboxwatch/impl.go
@@ -87,7 +87,7 @@ func (flw *filteredListerWatcher) setInclusion(obj any, include bool) {
 type myList struct {
 	metav1.TypeMeta
 	metav1.ListMeta
-	items []runtime.Object
+	Items []runtime.Object
 }
 
 var _ upstreamcache.ListerWatcher = &filteredListerWatcher{}
@@ -125,7 +125,7 @@ func (flw *filteredListerWatcher) List(options metav1.ListOptions) (runtime.Obje
 			Continue:           listMeta.GetContinue(),
 			RemainingItemCount: listMeta.GetRemainingItemCount(),
 		},
-		items: filteredSlice,
+		Items: filteredSlice,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


MikeSpreitzer:
/cc @yana1205 
/cc @yuji-watanabe-jp 
@waltforme

yana1205:
Is `NewSharedInformerForEdgeConfig` the informer that C2P controller calls to access each PolicyReport from mailbox workspaces by using kubeconfig for EdgeServiceProvider Workspace?
Like this
```
_, informer, _ := mailboxwatch.NewSharedInformerForEdgeConfig(edgeServiceProviderWorkspaceClientConfig, ...)
informer.AddEventHandler(
// something to handle Add/Update/Delete event
)
```

MikeSpreitzer:
@yana1205 : yes that is the most likely choice, I think.  The C2P Controller could use `NewSharedInformer` instead, if the C2P Controller had other uses for the clientset or `workspaceScopedInformerFactory` buried inside of `NewSharedInformerForEdgeConfig`.

BTW, the C2P controller may want to also use the returned `tenancyv1a1informers.WorkspaceInformer` to wait for its `HasSynced()` before starting to process the notifications of policy reports.  That is a common pattern.  It may not be important in the case of the C2P Controller; the virtue of that pattern is that it ensures that the informer's local cache has been filled before the controller does any processing, lest the controller react to a transient situation of a policy report that is coming but has not arrived yet.

clubanderson:
@ezrasilvera can you take a look at this PR and see if you agree with @yana1205 to approve?

MikeSpreitzer:
I tried to write a unit test.  That made me realize there was a missing piece for creating the ListerWatcher.  So I added that missing piece, and updated the interface to reference it.

The unit test fails unexpected with a complaint that "no kind is registered for the type apps.ReplicaSet in scheme", despite lots of evidence to the contrary.  This failure is probably specific to the test, not relevant to normal usage.

yana1205:
> "no kind is registered for the type apps.ReplicaSet in scheme"

Might be `k8sapps.ReplicaSet` not `apps.ReplicaSet`?
That error message is gone if I use k8sapps but the test still fails with a different message...

I have a question about NewListerWatcher. What kubeconfig (more exactly saying rest.Config) should a caller (C2P or whatever client) use for kcpClientset of NewListerWatcher? Is it OK by instantiating the clientset by kubeconfig of EdgeServiceProvider workspace?

clubanderson:
@yuji-watanabe-jp @yana1205 can you implement this in c2p?  @ezrasilvera @effi-ofer can you write a 'main' to exercise this in the meantime?

effi-ofer:
> @yuji-watanabe-jp @yana1205 can you implement this in c2p? @ezrasilvera @effi-ofer can you write a 'main' to exercise this in the meantime?

I'll add the main in a separate PR.

MikeSpreitzer:
@yana1205 : For the call on `NewListerWatcher` you do not pass a clientset, you pass a `TypedContextualListWatcher`.  You can get one by starting from the kube client config in the context named "system:admin" in the kubeconfig created by `kcp start` and then proceeding as in the unit test:

```
		k8sClientset := kcpk8sfake.NewSimpleClientset()
		...
		rslw := NewListerWatcher[*k8sapps.ReplicaSetList](ctx, k8sClientset.AppsV1().ReplicaSets())
```

but instead of `kcpk8sfake.NewSimpleClientset` call your analogue of https://github.com/kcp-dev/client-go/blob/aeff170a288b8e135b5db640d08e6e43da0f46ec/kubernetes/clientset.go#L430

and instead of `AppsV1().ReplicaSets()` you call the corresponding methods to get the relevant interface for policy reports.

MikeSpreitzer:
Notice how the server URL in the cluster referenced by the "system:admin" context has a shorter path than the server URLs in the clusters that refer to specific workspaces.

Some of the additional authorizations enjoyed by the user referenced by the "system:admin" context are also critical.

pkg/mailboxwatch/impl.go
```diff
@@ -0,0 +1,200 @@
...
+
+func (flw *filteredListerWatcher) OnDelete(obj any) {
+	flw.setInclusion(obj, false)
+}
+
+func (flw *filteredListerWatcher) setInclusion(obj any, include bool) {
+	ws := obj.(*tenancyv1a1.Workspace)
+	mbwsName := ws.Name
+	mbwsNameParts := strings.Split(mbwsName, placement.WSNameSep)
+	if len(mbwsNameParts) != 2 {
```
ezrasilvera, line 73: Is that enough condition for mb WS? Don't we need to compare the name to the actual MB name pattern ?

ezrasilvera:
@MikeSpreitzer Maybe I'm missing something. Why do you need ReplicaSet in the unittest for MB informer? ReplicaSet is not a native KCP (or KCP edge) resource, right ?

yana1205:
> @yana1205 : For the call on `NewListerWatcher` you do not pass a clientset, you pass a `TypedContextualListWatcher`. You can get one by starting from the kube client config in the context named "system:admin" in the kubeconfig created by `kcp start` (or any other kube client config that is suitable for all-cluster operations) and then proceeding as in the unit test:
> 
> ```
> 		k8sClientset := kcpk8sfake.NewSimpleClientset()
> 		...
> 		rslw := NewListerWatcher[*k8sapps.ReplicaSetList](ctx, k8sClientset.AppsV1().ReplicaSets())
> ```
> 
> but instead of `kcpk8sfake.NewSimpleClientset` call your analogue of https://github.com/kcp-dev/client-go/blob/aeff170a288b8e135b5db640d08e6e43da0f46ec/kubernetes/clientset.go#L430
> 
> and instead of `AppsV1().ReplicaSets()` you call the corresponding methods to get the relevant interface for policy reports.

I see the following errors that looks relating with wsPreInformer. 
```
k8s.io/client-go@v0.0.0-20230210192259-aaa28aa88b2d/tools/cache/reflector.go:215: failed to list *v1alpha1.Workspace: the server could not find the requested resource (get workspaces.tenancy.kcp.io)
```
Is my configuration correct? The KUBECONFIG is with system:admin context.
```
	clusterClient, err := clusterclientset.NewForConfig(config)
	k8sClientset, err := kubernetes.NewForConfig(config)

	espwCluster := logicalcluster.Name("root")

	informerFactory := kcpinformers.NewSharedInformerFactory(clusterClient, 0)
	wsPreInformer := informerFactory.Tenancy().V1alpha1().Workspaces()
	go wsPreInformer.Informer().Run(ctx.Done())

	rslw := mailboxwatch.NewListerWatcher[*corev1.ConfigMapList](ctx, k8sClientset.CoreV1().ConfigMaps())
	rsInformer := mailboxwatch.NewSharedInformer(wsPreInformer.Cluster(espwCluster), rslw, &corev1.ConfigMap{}, 0, upstreamcache.Indexers{})

	handle := func(event string, obj interface{}) {
		logger.Info(fmt.Sprintf("event %s: %v", event, obj))
	}
	rsInformer.AddEventHandler(upstreamcache.ResourceEventHandlerFuncs{
		AddFunc:    func(obj interface{}) { handle("add", obj) },
		UpdateFunc: func(oldObj, newObj interface{}) { handle("update", newObj) },
		DeleteFunc: func(obj interface{}) { handle("delete", obj) },
	})
	<-ctx.Done()
```

I confirmed that I can get the list of workspaces by clusterClient. (e.g. `clusterClient.Cluster(espwCluster.Path()).TenancyV1alpha1().Workspaces().List(ctx, v1.ListOptions{})`)

@effi-ofer Can you successfully run the informer?

MikeSpreitzer:
1. FYI, in https://docs.kcp-edge.io/docs/coding-milestones/poc2023q1/example1/ the edge service provider workspace is NOT `root`, it is `root:edge`.  But you could make it something else if you want.
2. I can not tell, because you did not show the imports, whether `k8sClientset` is an all-cluster clientset or is focused on one cluster.  It is easier for me to see everything I need if you reference something that I can read from github.
3. It is not obvious to me what is going wrong.  You can get more information by turning up the `-v` level on both the controller and the kcp server.  On the kcp server, `-v=3` or numerically higher gets you a log line for every request that can help debug authorizattion mysteries.  On the client side you need a higher number, maybe 6 or 8, but at some point you get debug logging for every HTTP request/response to/from the apiserver.

ezrasilvera:
> @ezrasilvera : in the unit test for the mailbox workload object informer we need to use _some_ workload objects. I picked one kind.
I know it's not related to the actual unittest bug, but wanted to understand - 
Why must it be a workload object?  I thought it should work with any object that synced to the edge-cluster. Maybe I'm missing something on the flow...

MikeSpreitzer:
@ezrasilvera : I thought it would be less confusing to use a workload resource than a control resource.

yana1205:
Thanks for the advice. @MikeSpreitzer 
I got the detail log with -v=8.
```
I0329 08:41:13.396766   40434 round_trippers.go:463] GET https://192.168.10.105:6443/clusters/%2A/apis/tenancy.kcp.io/v1alpha1/workspaces?limit=500&resourceVersion=0
```
Is wildcard query (`/clusters/*/apis/...`) expected? Maybe I am missing something. I uploaded the entire main code. 
https://gist.github.com/yana1205/b9303ea12f3e0230e17677d93fc1ab78

pkg/mailboxwatch/impl.go
```diff
@@ -0,0 +1,200 @@
...
+		flw.clusters.Insert(cluster)
+	} else {
+		flw.clusters.Delete(cluster)
+	}
+}
+
+type myList struct {
+	metav1.TypeMeta
+	metav1.ListMeta
+	items []runtime.Object
```
yana1205, line 90: Is it OK with `items`?
I get the following error (`*mailboxwatch.myList is not a list: no Items field in this object`).
```
E0329 09:40:59.750244   51001 reflector.go:138] k8s.io/client-go@v0.0.0-20230210192259-aaa28aa88b2d/tools/cache/reflector.go:215: Failed to watch *v1.ConfigMap: unable to understand list result &mailboxwatch.myList{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"4001", Continue:"", RemainingItemCount:(*int64)(nil)}, items:[]runtime.Object{(*v1.ConfigMap)(0x140003aaf00)}} (*mailboxwatch.myList is not a list: no Items field in this object)
```

If I change it to `Items`, the error is gone and the informer successfully notifies the handler of the Add/Update/Delete event.

yana1205: This is the case when I use `"k8s.io/client-go/kubernetes"` for `k8sClientset`. Originally, I am used to use `"github.com/kcp-dev/client-go/kubernetes"`. I am still doing try-and-error for https://github.com/kcp-dev/edge-mc/pull/265#issuecomment-1487763749. I am trying non-cluster client.

The full code is https://gist.github.com/yana1205/b9303ea12f3e0230e17677d93fc1ab78#file-main-with-non-cluster-cliientset-go

MikeSpreitzer: Oh, dang, I meant to correct the capitalization of `items`, I must have forgotten it.

MikeSpreitzer: For authorization problems you also need to look at the user involved.  I do not remember whether it shows up on the client, shows up on the server, but do look for it.

For the informer on mailbox workspaces, it should be focused on one specific logical cluster --- the edge service provider workspace.

The clientset-from-config function that you call has to match the server URL in the client config.  Call the all-cluster clientset constructor when the server URL is short (does not include `/clusters/$something_even_star`), call the single-cluster-focused clientset constructor when the server URL is long (includes `/clusters/$something_not_star`).  As far as I know, it should be valid to either make an all-cluster clientset and then use a `Cluster(.)` method to scope it down or start with a single-cluster clientset.

MikeSpreitzer: What is the relationship between the gist you referenced and the github content you referenced?

MikeSpreitzer: I think that the C2P Controller will need at least two kube client configs, just as the placement translator does.  See https://docs.kcp-edge.io/docs/coding-milestones/poc2023q1/placement-translator/#usage .  One kube client config points specifically at the edge service provider workspace and authorizes reading the Workspace objects from there. The other kube client config points at the server base and authorizes everything, including reading policy reports from all clusters.

You might want to re-use the code at https://github.com/kcp-dev/edge-mc/blob/a2f67fc36838e6052043896995a16441ba30527a/cmd/placement-translator/main.go#L65-L92 for command line flags for those two kube client configs.

MikeSpreitzer:
The force-push to ec951b0 is a rebase onto the latest `main` and addition of one commit.

pkg/mailboxwatch/impl.go
MikeSpreitzer: The latest revision fixes the capitalization of Items.

MikeSpreitzer:
@ezrasilvera : yes, the unit test was trying to put the wrong type of object in the tracker.

effi-ofer:
Hey @MikeSpreitzer I am glad you resolved the unit test.  I am still getting a failure due to the objectsEqualCond failing.  You still need help with this, or are you good?
```
=== RUN   TestMailboxInformer
    impl_test.go:188: Added to tracker: Workspace named "c0"
    impl_test.go:222: Added to tracker: ReplicaSet named mailboxwatch.objectName{cluster:"c0", namespace:"ns7946", name:"rs4519"}
    impl_test.go:226: Workspaces did not settle in time: map[] != map[{cluster:espw namespace: name:c0}:0xc000513040]
--- FAIL: TestMailboxInformer (5.00s)
```

MikeSpreitzer:
@effi-ofer : yes, that is what the last commit is referring to when it talks about being stuck on lack of watchers.  The reason that condition is not met is that the tracker does not think it has any watchers for the new object.

MikeSpreitzer:
/approve
(Not because this is done, but just because it is progress.)

yana1205:
Hi @effi-ofer 

Regarding to https://github.com/kcp-dev/edge-mc/pull/265#issuecomment-1486993741
> 
> I'll add the main in a separate PR.

Did you get a chance to have a sample code to run the informer in main function?

effi-ofer:
@yana1205 sorry I haven't done it yet. I am ramping up and I am slow :) I assume you're hoping to use a main so if this is holding you back, please me know.

yana1205:
@effi-ofer No problem. I was just finding something to compare my usage. Thanks!

yana1205:
@MikeSpreitzer @effi-ofer 

Do you know a good way to collect custom resources whose CRD is only defined in mailbox workspaces?
Here is the snippet about what I am doing with the informer. This way creates an informer and the handler per mailbox workspace rather than using single informer with a cluster wide client. I'm not sure whether this approach is expected usage of this informer or not.
```
...
        // wait for workspace pre-informer has been synced.
        wait.PollImmediateInfinite(5*time.Second, func() (bool, error) {
		return wsPreInformer.Informer().HasSynced(), nil
	})

        // custom resource (would be PolicyReport in real use case)
	gvr := schema.GroupVersionResource{
		Group:    "my.domain",
		Version:  "v1alpha1",
		Resource: "samples",
	}
        // client for the custom resource (this resource type is non-namespace resource type)
        // dynamicClusterClient was created from system:admin Kubeconfig
	crClusterClient := dynamicClusterClient.Resource(gvr)

	mbwsList, err := wsPreInformer.Lister().List(labels.Everything())
	if err != nil {
		logger.Error(err, "Failed to list mailbox workspaces")
		os.Exit(30)
	}

	handlers := map[string]*handler{}
	for _, mbws := range mbwsList {
		cluster := mbws.Spec.Cluster
		h := newHandler(logger, gvr, cluster)
		handlers[mbws.Name] = h
		// select cluster instead of doing clusterwide
		clw := crClusterClient.Cluster(logicalcluster.NewPath(cluster))
		rslw := mailboxwatch.NewListerWatcher[*unstructured.UnstructuredList](ctx, clw)
		rsInformer := mailboxwatch.NewSharedInformer(wsPreInformer, rslw, &unstructured.Unstructured{}, 0, upstreamcache.Indexers{})
		rsInformer.AddEventHandler(h)
		go rsInformer.Run(ctx.Done())
	}

	<-ctx.Done()
``` 
(full code: https://gist.github.com/yana1205/e1328af871afd4374796da8b7d8d9cb9)

If I use clusterClient (`crClusterClient`) as `clw` instead of `crClusterClient(logicalcluster.NewPath(cluster))`, the program failed to collect the resources due to
```
failed to list *unstructured.Unstructured: the server could not find the requested resource
```

`/clusters/*` does not have custom resource type. That's why I guess this error happens.
```
KUBECONFIG=$workdir/.kcp/system-admin.kubeconfig k --server="https://9.116.80.31:6443/clusters/*" api-resources

NAME                              SHORTNAMES   APIVERSION                        NAMESPACED   KIND
configmaps                        cm           v1                                true         ConfigMap
events                            ev           v1                                true         Event
limitranges                       limits       v1                                true         LimitRange
namespaces                        ns           v1                                false        Namespace
resourcequotas                    quota        v1                                true         ResourceQuota
secrets                                        v1                                true         Secret
serviceaccounts                   sa           v1                                true         ServiceAccount
mutatingwebhookconfigurations                  admissionregistration.k8s.io/v1   false        MutatingWebhookConfiguration
validatingwebhookconfigurations                admissionregistration.k8s.io/v1   false        ValidatingWebhookConfiguration
customresourcedefinitions         crd,crds     apiextensions.k8s.io/v1           false        CustomResourceDefinition
apibindings                                    apis.kcp.io/v1alpha1              false        APIBinding
apiconversions                                 apis.kcp.io/v1alpha1              false        APIConversion
apiexportendpointslices                        apis.kcp.io/v1alpha1              false        APIExportEndpointSlice
apiexports                                     apis.kcp.io/v1alpha1              false        APIExport
apiresourceschemas                             apis.kcp.io/v1alpha1              false        APIResourceSchema
tokenreviews                                   authentication.k8s.io/v1          false        TokenReview
localsubjectaccessreviews                      authorization.k8s.io/v1           true         LocalSubjectAccessReview
selfsubjectaccessreviews                       authorization.k8s.io/v1           false        SelfSubjectAccessReview
selfsubjectrulesreviews                        authorization.k8s.io/v1           false        SelfSubjectRulesReview
subjectaccessreviews                           authorization.k8s.io/v1           false        SubjectAccessReview
certificatesigningrequests        csr          certificates.k8s.io/v1            false        CertificateSigningRequest
leases                                         coordination.k8s.io/v1            true         Lease
logicalclusters                                core.kcp.io/v1alpha1              false        LogicalCluster
events                            ev           events.k8s.io/v1                  true         Event
clusterrolebindings                            rbac.authorization.k8s.io/v1      false        ClusterRoleBinding
clusterroles                                   rbac.authorization.k8s.io/v1      false        ClusterRole
rolebindings                                   rbac.authorization.k8s.io/v1      true         RoleBinding
roles                                          rbac.authorization.k8s.io/v1      true         Role
```

PR #267, opened 2023-03-27
✨ First working edge scheduler

waltforme:
## Summary
This PR
- implements the reconciliation triggered by EdgePlacement changes
- makes improvements in response to review of PR #261

so that now we have a working edge scheduler.


pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -21,79 +21,140 @@ import (
...
+		On EdgePlacement 'ep' change:
+		1) from cache, find loc(s) that being selected by ep
+
+		2) from cache, for each of the found loc, find st(s) that being selected by the loc
+
+		3) update store, with loc(s) that being selected by ep
+
+		4) update apiserver
+
+		Need data structure: none.
```
MikeSpreitzer, line 54: What does this mean?  This file depends on data structures in the store.

pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -21,79 +21,140 @@ import (
...
+
+	store.l.Lock()
+	defer store.l.Unlock() // TODO(waltforme): Is it safe to shorten the critical section?
+
+	ep, err := c.edgePlacementLister.Cluster(epws).Get(epName)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			logger.V(1).Info("EdgePlacement not found")
+			logger.V(3).Info("dropping EdgePlacement from store")
+			store.dropEp(epKey)
```
MikeSpreitzer, line 65: There is also the SinglePlacementSlice to consider.

pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -21,79 +21,140 @@ import (
...
+		Need data structure: none.
+	*/
+
+	store.l.Lock()
+	defer store.l.Unlock() // TODO(waltforme): Is it safe to shorten the critical section?
+
+	ep, err := c.edgePlacementLister.Cluster(epws).Get(epName)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			logger.V(1).Info("EdgePlacement not found")
```
MikeSpreitzer, line 63: BTW, "not found" connotes that it normally would be found.  But normally deletion leads here.  Wording like "is absent" is less suggestive that there is a problem here.

pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -21,79 +21,140 @@ import (
...
+			store.dropEp(epKey)
 			return nil
+		} else {
+			logger.Error(err, "failed to get EdgePlacement")
+			return err
 		}
+	}
+
+	// 1)
+	locsAll, err := c.locationLister.List(labels.Everything())
```
MikeSpreitzer, line 74: This also does not need to be done with the store locked.

pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -21,79 +21,140 @@ import (
...
+	if err != nil {
+		logger.Error(err, "failed to find Locations for EdgePlacement")
+	}
+	locsSelecting := packLocKeys(locsFilteredByEp)
+
+	singles := []edgev1alpha1.SinglePlacement{}
+	for _, loc := range locsFilteredByEp {
+		// 2)
+		lws := logicalcluster.From(loc)
+		stsInLws, err := c.synctargetLister.Cluster(lws).List(labels.Everything())
```
MikeSpreitzer, line 89: If the store maintained a map from Location[ExternalName] to set of SyncTarget[NameAndUID] then it would not be necessary to enumerate every SyncTarget and evaluate the Location's "where predicate" here.  In the current revision, the number of "where predicate" evaluations is O( (number of Locations selected by this EdgePlacement) X (number of SyncTargets in the world) ).

pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -21,79 +21,140 @@ import (
...
+	locsAll, err := c.locationLister.List(labels.Everything())
+	if err != nil {
+		logger.Error(err, "failed to list Locations in all workspaces")
 		return err
 	}
+	locsFilteredByEp, err := filterLocsByEp(locsAll, ep)
+	if err != nil {
+		logger.Error(err, "failed to find Locations for EdgePlacement")
+	}
+	locsSelecting := packLocKeys(locsFilteredByEp)
```
MikeSpreitzer, line 83: I suggest `locsSelected`.

pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -21,79 +21,140 @@ import (
...
+
+	singles := []edgev1alpha1.SinglePlacement{}
+	for _, loc := range locsFilteredByEp {
```
MikeSpreitzer, line 86: This is where the store starts being used, so just before this loop is where the store needs to be locked.

pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -21,79 +21,140 @@ import (
...
+							Name:       epName,
 							UID:        ep.UID,
 						},
 					},
 				},
-				Destinations: []edgev1alpha1.SinglePlacement{},
+				Destinations: singles,
 			}
-			_, err = c.edgeClusterClient.Cluster(ws.Path()).EdgeV1alpha1().SinglePlacementSlices().Create(ctx, sps, metav1.CreateOptions{})
+			_, err = c.edgeClusterClient.Cluster(epws.Path()).EdgeV1alpha1().SinglePlacementSlices().Create(ctx, sps, metav1.CreateOptions{})
```
MikeSpreitzer, line 135: I would fill in the `metav1.CreateOptions.FieldManager`.

pkg/scheduler/reconcile_on_edgeplacement.go
```diff
@@ -21,79 +21,140 @@ import (
...
 			} else {
 				logger.V(1).Info("created SinglePlacementSlice")
 			}
 		} else {
-			logger.Error(err, "failed getting SinglePlacementSlice for EdgePlacement")
+			logger.Error(err, "failed getting SinglePlacementSlice")
 			return err
 		}
+	} else { // update
+		currentSPS.Destinations = singles
```
MikeSpreitzer, line 149: If every change notification provokes a write, and remember that every write will provoke a change notification unless there is a communication problem, won't this produce an infinite loop?

MikeSpreitzer (approved):
This is progress.
/lgtm

PR #273, opened 2023-03-29
✨ More work on placement translator

MikeSpreitzer:
## Summary
This PR makes further progress on the placement translator, implementing the WhereResolver (really a reader), the expected assembler of SetBindings, the set differencers, and the BindingOrganizer.

## Related issue(s)

Fixes #


MikeSpreitzer:
The force-push to 6eb1275 is a rebase onto `main`

MikeSpreitzer:
The force-push to 24bd54ee fixes a bug in the BindingOrganizer where a transaction would hang due to nested attempts to lock the same mutex.

MikeSpreitzer:
The force-push to d27e25ab fixes another locking bug in the BindingOrganizer.

MikeSpreitzer:
The force-push to 33bc255d makes some debug printing better.  Now I can see that I got a successful test through to results from the SetBinder.

MikeSpreitzer:
The force-push to 96862528 is a rebase onto `main`

pkg/placement/triple-transforms.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// VisitableEmptyable is a collection is both Visitable and Emptyable.
+type VisitableEmptyable[Elt any] interface {
+	Emptyable
+	Visitable[Elt]
+}
+
+// NewIndex123by13to2s maintains an index into a three-column table.
+// The index is keyed by the first and second columns and each entry holds
```
waltforme, line 35: Should it be 'keyed by the first and _third_ columns'?

pkg/placement/triple-transforms.go
```diff
@@ -0,0 +1,145 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package placement
+
+// MappingToEmptyableReceiver is a mapping receiver specialized to the case
```
waltforme, line 19: Should it be 'MappingTo**Visitable**EmptyableReceiver'?

pkg/placement/set-types.go
```diff
@@ -0,0 +1,220 @@
...
+		})
+		return finish(accum)
+	}
+}
+
+// Identity1 is useful in reduers where the accumulator has the same type as the result
+func Identity1[Val any](val Val) Val { return val }
+
+// StatefulReducer makes a Reducer that works with a stateful accumulator
+func StatefulReducer[Elt any, Accum any, Ans any](initialize func() Accum, add func(Accum, Elt), finish func(Accum) Ans) Reducer[Elt, Ans] {
```
waltforme, line 171: Looks like the only difference between `ValueReducer` and `StatefulReducer` is the signature of `add`. May I have more info in what sense that difference makes it 'stateful' here?

pkg/placement/map-set.go
```diff
@@ -0,0 +1,121 @@
...
+		return true
+	}
+	return false
+}
+
+func MapSetAddNoResult[Elt comparable](set MapSet[Elt], elt Elt) {
+	set.Add(elt)
+}
+
+func MapSetSymmetricDifference[Elt comparable](wantLeftMinusRight, wantIntersection, wantRightMinusLeft bool, left, right Set[Elt]) (MapSet[Elt], MapSet[Elt], MapSet[Elt]) {
```
waltforme, line 85: If my understanding is correct, someone `wantIntersection` but doesn't `wantLeftMinusRight` or `wantRightMinusLeft` need to to `Visit` twice --- both from left and from right. Maybe there is some room to optimize?

pkg/placement/map-relation.go
```diff
@@ -0,0 +1,113 @@
...
+		by1: map[First]MapSet[Second]{},
+		by2: map[Second]MapSet[First]{},
+	}
+}
+
+func (mr MapRelation2[First, Second]) IsEmpty() bool {
+	return len(mr.by1) == 0
+}
+
+func (mr MapRelation2[First, Second]) LenIsCheap() bool { return false }
```
waltforme, line 40: Is it necessary (or worth it) to make `LenIsCheap` to be `true` ? I can imagine maintaining a counter to track the incremental change of length whenever a `MapRelation2` is updated, in place of `Relation2LenFromVisit`. But whether we need this might depends on how frequently `Len()` is called.

waltforme (commented):
Exquisitely designed machinery! Learning from it.

pkg/placement/triple-transforms.go
MikeSpreitzer: yes

pkg/placement/triple-transforms.go
MikeSpreitzer: yes

pkg/placement/map-relation.go
MikeSpreitzer: Yeah, we can consider further development in the future.

pkg/placement/map-set.go
MikeSpreitzer: Yes, I considered that case.  But I see now that I did my logic minimization wrong.  So there is a bug to fix here, thanks.

pkg/placement/set-types.go
waltforme: Thanks!

PR #277, opened 2023-03-31
🐛 mailbox name should be composed of clustername-mb-UID

effi-ofer:
## Summary
Hey @MikeSpreitzer shouldn't the mailbox name be composed of two parts separated by "-mb-"?  The current unit test fails to add the mailbox because it fails the `if len(mbwsNameParts) != 2 {` check in setInclusion().

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/mailboxwatch/impl_test.go
+++ b/pkg/mailboxwatch/impl_test.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	upstreamcache "k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


effi-ofer:
I am new to kcp/edge so my apologies in advance if I am totally talking nonsense here.

MikeSpreitzer:
You did notice a problem, but it is bigger.  One string is being used for two purposes, two different strings should be used.  I will try a fix.

MikeSpreitzer:
I think #279 is the correct fix.
The unit test still fails, with the same symptom.

MikeSpreitzer:
Please have a look at #279

effi-ofer:
Closing this PR in light of #279 .

PR #282, opened 2023-04-03
✨ Placement translator refactor

MikeSpreitzer:
## Summary
This PR refactors the internal expression of configuration for the last step of the placement translator, to (a) make it less confusing by treating the namespaced and cluster-scoped resources separately and (b) making that configuration follow relaitonal algebra patterns (because that looks like the way to implement it).

## Related issue(s)

Fixes #


pkg/placement/binding-organizer.go
```diff
@@ -20,160 +20,222 @@ import (
...
+	logger        klog.Logger
+	discovery     APIMapProvider
+	resourceModes ResourceModes
+	eventHandler  EventHandler
+	// projectionMapProvider TransformingRelayMap[ProjectionKey, *projectionPerClusterImpl, *ProjectionPerCluster]
+	workloadProjector    WorkloadProjector
+	clusterWhatWhereFull MappingReceiver[ClusterWhatWhereFullKey, ProjectionModeVal]
 	sync.Mutex
-}
+	workloadProjectionSections WorkloadProjectionSections // non-nil only during a transaction!
```
waltforme, line 178: Question: Does this inline comment refer to line 195 (`sbo.workloadProjectionSections = WorkloadProjectionSections{}`)?

MikeSpreitzer: yes

waltforme: Is `WorkloadProjectionSections{}` itself nil when it has several nil fields? I'm not sure.

MikeSpreitzer: Sorry, typo, the comment should say non-zero rather than non-nil

MikeSpreitzer: Fixed in latest revision.

MikeSpreitzer:
The force-push to 5d253fff is a rebase onto the latest `main` plus more commits working on the real APIMapProvider.

MikeSpreitzer:
@yana1205 
@ezrasilvera

PR #285, opened 2023-04-06
✨ Make client options code in pkg and let the scheduler use it

waltforme:
## Summary
This PR
- puts (and tweaks) code for preparing client options into pkg (as suggested [here](https://github.com/kcp-dev/edge-mc/pull/271#issuecomment-1487603900), thanks to @MikeSpreitzer ) so that controllers can share them;
- use that part of code to allow more configurable flags for the scheduler @dumb0002 .

This PR also improves the English language in logs, thanks to @KPRoche the native speaker.

### DIFF
--- a/cmd/scheduler/options/options.go
+++ b/cmd/scheduler/options/options.go
@@ -21,11 +21,15 @@ import (
 
 	"k8s.io/component-base/config"
 	"k8s.io/component-base/logs"
+
+	"github.com/kcp-dev/edge-mc/pkg/client"
 )
 
 type Options struct {
-	KcpKubeconfig string
-	Logs          *logs.Options
+	EspwClientOpts   client.ClientOpts
+	RootClientOpts   client.ClientOpts
+	SysAdmClientOpts client.ClientOpts
+	Logs             *logs.Options
 }
 
 func NewOptions() *Options {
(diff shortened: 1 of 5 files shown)


waltforme:
Forth-push to e414bb6 adds license boilerplate to new file.

pkg/client/options.go
```diff
@@ -0,0 +1,65 @@
...
+func NewClientOpts(which, description string) *ClientOpts {
+	return &ClientOpts{
+		which:        which,
+		description:  description,
+		loadingRules: clientcmd.NewDefaultClientConfigLoadingRules(),
+		overrides:    clientcmd.ConfigOverrides{},
+	}
+}
+
+func (opts *ClientOpts) OverrideCurrentContext(currCtx string) {
```
MikeSpreitzer, line 43: If this returned the receiver then these calls could be chained

cmd/scheduler/options/options.go
```diff
@@ -34,12 +38,19 @@ func NewOptions() *Options {
...
+		RootClientOpts:   *client.NewClientOpts("root", "access to all clusters"),
+		SysAdmClientOpts: *client.NewClientOpts("sysadm", "access to all clusters as system:admin"),
+		Logs:             logs,
 	}
 }
 
 func (options *Options) AddFlags(fs *pflag.FlagSet) {
-	fs.StringVar(&options.KcpKubeconfig, "kcp-kubeconfig", options.KcpKubeconfig, "Kubeconfig file for kcp")
+	options.EspwClientOpts.AddFlags(fs)
+	options.RootClientOpts.OverrideUserAndCluster("kcp-admin", "root")
```
MikeSpreitzer, line 50: Why not set the defaults in NewOptions?

pkg/client/options.go
waltforme: Done in the latest force-push. Thanks!

MikeSpreitzer (approved):
/lgtm

MikeSpreitzer:
Thank you!

PR #288, opened 2023-04-09
✨ Placement translator development

MikeSpreitzer:
## Summary
The PR continues development of the placement translator.
It has gotten as far as maintaining the syncer config objects (new style).  It seems to create configs OK, but there is some bug in reacting to deletion of an EdgePlacement.

## Related issue(s)

Fixes #

/cc @yana1205 
/cc @waltforme 
@yuji-watanabe-jp

### DIFF
--- a/cmd/placement-translator/main.go
+++ b/cmd/placement-translator/main.go
@@ -160,6 +160,7 @@ func main() {
 	edgeInformerFactory := emcinformers.NewSharedInformerFactoryWithOptions(edgeViewClusterClientset, resyncPeriod)
 	epClusterPreInformer := edgeInformerFactory.Edge().V1alpha1().EdgePlacements()
 	spsClusterPreInformer := edgeInformerFactory.Edge().V1alpha1().SinglePlacementSlices()
+	syncfgClusterPreInformer := edgeInformerFactory.Edge().V1alpha1().SyncerConfigs()
 	var _ edgev1a1informers.SinglePlacementSliceClusterInformer = spsClusterPreInformer
 
 	espwClientset, err := kcpscopedclientset.NewForConfig(espwRestConfig)
@@ -198,7 +199,9 @@ func main() {
 
 	doneCh := ctx.Done()
 	// TODO: more
-	pt := placement.NewPlacementTranslator(concurrency, ctx, epClusterPreInformer, spsClusterPreInformer, mbwsPreInformer, kcpClusterClientset, discoveryClusterClient, crdClusterPreInformer, bindingClusterPreInformer, dynamicClusterClient, edgeClusterClientset)
+	pt := placement.NewPlacementTranslator(concurrency, ctx, epClusterPreInformer, spsClusterPreInformer, syncfgClusterPreInformer,
+		mbwsPreInformer, kcpClusterClientset, discoveryClusterClient, crdClusterPreInformer, bindingClusterPreInformer,
+		dynamicClusterClient, edgeClusterClientset)
 	edgeInformerFactory.Start(doneCh)
 	espwInformerFactory.Start(doneCh)
 	dynamicClusterInformerFactory.Start(doneCh)
--- a/pkg/placement/workload-projector.go
+++ b/pkg/placement/workload-projector.go
@@ -70,10 +70,10 @@ func NewWorkloadProjector(
 	}
 	noteModeWrite := MapChangeReceiverFuncs[edgeapi.SinglePlacement, MutableMap[metav1.GroupResource, ProjectionModeVal]]{
 		OnCreate: func(destination edgeapi.SinglePlacement, _ MutableMap[metav1.GroupResource, ProjectionModeVal]) {
-			wp.changedDestinations.Add(destination)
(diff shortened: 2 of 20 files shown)


cmd/placement-translator/main.go
```diff
@@ -169,29 +176,32 @@ func main() {
...
 
 	doneCh := ctx.Done()
 	// TODO: more
-	pt := placement.NewPlacementTranslator(concurrency, ctx, epClusterPreInformer, spsClusterPreInformer, mbwsPreInformer, kcpClusterClientset, discoveryClusterClient, crdClusterInformer, bindingClusterInformer, dynamicClusterClient)
+	pt := placement.NewPlacementTranslator(concurrency, ctx, epClusterPreInformer, spsClusterPreInformer, syncfgClusterPreInformer,
+		mbwsPreInformer, kcpClusterClientset, discoveryClusterClient, crdClusterPreInformer, bindingClusterPreInformer,
+		dynamicClusterClient, edgeClusterClientset)
 	edgeInformerFactory.Start(doneCh)
 	espwInformerFactory.Start(doneCh)
 	dynamicClusterInformerFactory.Start(doneCh)
```
yana1205, line 207: NewPlacementTranslator() waits returning until Informer.HasSynced is true. I might not correctly understand the behavior of informer but can the informer be a synced status (;Informer.HasSynced is true) before the informer factory starts?

MikeSpreitzer: Yes, this change was a mistake.  I have verified that in testing, and am about to push a fix.

MikeSpreitzer: Fix pushed.

MikeSpreitzer:
The force-push to d84d9cea is a rebase onto the latest `main`

yana1205:
Can I ask when is Syncer Config CRD applied? Is it at mailbox workspace creation by mailbox controller or at enabling Edge Syncer by command `kubectl kcp workload edge-syncer ...`?

yana1205:
>  Users should not have to issue any command line commands. 

How does the user generate bootstrap manifests for Edge Syncer?
Currently, we run `kubectl kcp workload edge-syncer <name>` in mailbox workspace, and then we apply the generated manifests in the p-cluster.

pkg/placement/workload-projector.go
```diff
@@ -141,89 +205,175 @@ func (wp *workloadProjector) configSyncLoop(ctx context.Context, worker int) {
...
+	logger.V(3).Info("Begin transaction")
 	var s1 SetChangeReceiver[Triple[edgeapi.SinglePlacement, NamespaceName, logicalcluster.Name]] = wp.nsDistributions
 	var s2 SetChangeReceiver[Triple[edgeapi.SinglePlacement, metav1.GroupResource, logicalcluster.Name]] = wp.nsrDistributions
 	var s3 SetChangeReceiver[Triple[edgeapi.SinglePlacement, GroupResourceInstance, logicalcluster.Name]] = wp.nnsDistributions
-	changedDestinations := NewMapSet[edgeapi.SinglePlacement]()
-	wp.changedDestinations = &changedDestinations
-	s1 = SetChangeReceiverFork(false, s1, recordFirst[edgeapi.SinglePlacement, NamespaceName, logicalcluster.Name](changedDestinations))
-	s2 = SetChangeReceiverFork(false, s2, recordFirst[edgeapi.SinglePlacement, metav1.GroupResource, logicalcluster.Name](changedDestinations))
-	s3 = SetChangeReceiverFork(false, s3, recordFirst[edgeapi.SinglePlacement, GroupResourceInstance, logicalcluster.Name](changedDestinations))
+	changedDestinations := func() *MutableSet[edgeapi.SinglePlacement] {
```
waltforme, line 308: Looks like there're quite a few usages of anonymous functions in recent PRs. Here is the 1st example, and there are a [2nd one](https://github.com/kcp-dev/edge-mc/blob/05b7f6c0f0a9e3398fc4b8a26dc14a28e8338597/pkg/placement/apiwatch-map-provider.go#L231-L244) a [3rd one](https://github.com/kcp-dev/edge-mc/blob/05b7f6c0f0a9e3398fc4b8a26dc14a28e8338597/pkg/placement/apiwatch-map-provider.go#L271-L276). I have two questions on them:
1. For the 2nd one and the 3rd one, I guess the anonymous functions are used to specify the location for the `defer` statements, am I right?
2. For the 1st anonymous function here, I'm trying to understand, what are the benefits or it is just some personal style?

MikeSpreitzer: The 1st one here is the scars from a battle over mysterious loss of information in the MutableSet here.  I tried making an incontrovertibly shared pointer, thinking maybe different parts of the code were manipulating different copies of the set.  The battle was finally won by wrapping the set with a Mutex, and I have not gotten around to clearing out the junk.  Will attend to in the future.

The other two examples are, as you observed, cases where I want the mutex to apply to just a small section of the code.

waltforme: I see. Thanks for explaining this!

PR #292, opened 2023-04-12
✨ Further placement translator development

MikeSpreitzer:
## Summary
First apparently working draft that includes upsync in the maintained SyncerConfig objects.

This PR includes an expansion of `example1.md` to include upsync.

Needed to add user-level hash maps because UpsyncSet is not "comparable" in go but we want to treat them as effectively comparable values.  (Later we may do more sophisticated logic on them, since they are predicates, but not now.)

## Related issue(s)

Fixes #

/cc @waltforme 
@yana1205

### DIFF
--- a/pkg/placement/hash-domain.go
+++ b/pkg/placement/hash-domain.go
@@ -1,0 +1,170 @@
+/*
+Copyright 2023 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package placement
+
+import (
+	"hash/crc64"
+
+	"github.com/kcp-dev/logicalcluster/v3"
+
+	edgeapi "github.com/kcp-dev/edge-mc/pkg/apis/edge/v1alpha1"
+)
+
+type HashDomain[Elt any] interface {
+	Equal(Elt, Elt) bool
+	Hash(Elt) HashValue
+}
+
+type HashValue = uint64
+
+func NewHashDomainFuncs[Elt any](equal func(Elt, Elt) bool, hash func(Elt) HashValue) HashDomainFuncs[Elt] {
+	return HashDomainFuncs[Elt]{equal, hash}
+}
+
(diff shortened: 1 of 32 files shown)


pkg/placement/hash-domain.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+}
+
+func TripleHashDomain[First, Second, Third any](first HashDomain[First], second HashDomain[Second], third HashDomain[Third]) HashDomain[Triple[First, Second, Third]] {
+	return HashDomainFuncs[Triple[First, Second, Third]]{
+		DoEqual: func(left, right Triple[First, Second, Third]) bool {
+			return first.Equal(left.First, right.First) && second.Equal(left.Second, right.Second) && third.Equal(left.Third, right.Third)
+		},
+		DoHash: func(tup Triple[First, Second, Third]) HashValue {
+			return first.Hash(tup.First)*65539 + second.Hash(tup.Second)*257 + third.Hash(tup.Third)
```
waltforme, line 61: Not sure my understanding is correct or not. Given `const numHashBuckets = 8`, is this line effectively equivalent to `return first.Hash(tup.First)*3 + second.Hash(tup.Second)*1 + third.Hash(tup.Third)`?

MikeSpreitzer: Yes.  I do not want to spend time making great hash functions right now, that is not the priority.  I suspect there must be a decent hash map package that we could import, but I also do not want to spend time looking for it now.  We just need something that works well enough for testing this week.

waltforme: I see. Thanks!

MikeSpreitzer:
The force-push to 68aea418 is a rebase onto `main`.

MikeSpreitzer:
The force-push to 5c9e38bc is a rebase onto `main`.

pkg/placement/slice-differencer.go
```diff
@@ -0,0 +1,41 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package placement
+
+func NewSliceDifferencerParametric[Elt any](isEqual func(Elt, Elt) bool, changeReceiver SetChangeReceiver[Elt], initial []Elt) Receiver[ /*immutable*/ []Elt] {
```
waltforme, line 19: Question: What does the `/*immutable*/` comment refer to?

MikeSpreitzer: It refers to the `[]Elt`.

waltforme: Thanks!

MikeSpreitzer:
I have found a bug in this PR, and the bug is fixed in #307

PR #297, opened 2023-04-12
🐛 Stay away from generated code

### LINKED ISSUE #296: bug: pkg/client/options.go is mixed in with generated code
### Describe the bug

pkg/client is entirely generated --- except for pkg/client/options.go

### Steps To Reproduce

Look at the repo

### Expected Behaviour

Independent source is not mixed in with generated code.

`pkg/client-options/options.go` would be a better choice, for example.

### Additional Context

_No response_

waltforme:
## Related issue(s)

Fixes #296


MikeSpreitzer (approved):
/lgtm
Thanks!

PR #301, opened 2023-04-13
🌱 Better names for ClientOpts methods

waltforme:
## Summary
This PR gives better names to ClientOpts' methods. Thanks to @MikeSpreitzer 's suggestions.


MikeSpreitzer (approved):
/lgtm
Thanks.

FYI, for a more extensive thing like this, see https://github.com/kubernetes/cli-runtime/blob/master/pkg/genericclioptions/config_flags.go .  Note that there are many other settings that a `main` might want to provide default values for.

waltforme:
> FYI, for a more extensive thing like this, see https://github.com/kubernetes/cli-runtime/blob/master/pkg/genericclioptions/config_flags.go . Note that there are many other settings that a `main` might want to provide default values for.

I see. Thanks for the info!

waltforme:
Oh sorry, fixed in force-push to 2d3853e. @MikeSpreitzer

PR #323, opened 2023-04-21
🐛 More cover for IsNotFound(err) and mystery nil

MikeSpreitzer:
## Summary
This PR tries to avoid a mysterious condition sometimes seen where an object returned from a Get-ish function along with an error appears to be non-nil and yet consequent calls cause a nil pointer fault inside the `unstructured` package.

## Related issue(s)

Fixes #

/cc @waltforme 
/cc @francostellari

### DIFF
--- a/pkg/placement/what-resolver.go
+++ b/pkg/placement/what-resolver.go
@@ -324,9 +324,12 @@ func (wr *whatResolver) processObject(ctx context.Context, cluster logicalcluste
 		return true
 	}
 	rObj, err := rr.lister.Get(objName)
-	if err != nil && !k8sapierrors.IsNotFound(err) {
-		logger.Error(err, "Failed to fetch generic object from lister")
-		return true
+	if err != nil {
+		if !k8sapierrors.IsNotFound(err) {
+			logger.Error(err, "Failed to fetch generic object from lister")
+			return true
+		}
+		rObj = nil
 	}
 	oldDetails := rr.byObjName[objName]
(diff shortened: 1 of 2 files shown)


pkg/placement/what-resolver.go
```diff
@@ -324,9 +324,12 @@ func (wr *whatResolver) processObject(ctx context.Context, cluster logicalcluste
 		return true
 	}
 	rObj, err := rr.lister.Get(objName)
```
waltforme, line 326: Just curious, does this `Get` call
https://github.com/kcp-dev/kubernetes/blob/aaa28aa88b2dab8fdaa99deb945ad26d9b4baeb0/staging/src/k8s.io/client-go/tools/cache/listers.go#L150-L159?

MikeSpreitzer: Possibly.

waltforme: Turns out it traces to https://github.com/kcp-dev/kubernetes/blob/aaa28aa88b2dab8fdaa99deb945ad26d9b4baeb0/staging/src/k8s.io/client-go/dynamic/dynamiclister/lister.go#L50-L59 but I don't know why a non-nil pointer could be returned with a non-nil error.

/lgtm

MikeSpreitzer: I choose to not rely on implementation details that could change.
The real mystery is why there is a nil pointer fault on a value that is known to not equal `nil`.

/approve

PR #328, opened 2023-04-23
🌱 Make some example objects more useful

waltforme:
## Summary
This PR updates two example edgeplacements to facilitate quick dev-time tests.
This PR also deletes an example TMC placement which we don't use any more.


MikeSpreitzer (approved):
/lgtm

PR #347, opened 2023-05-03
✨ KubeStellar Contributor Development Experience

dumb0002:
This PR implements the KCP-Edge Contributor Development Environment

## Related issue(s)

Additional features: https://github.com/kcp-dev/edge-mc/issues/289
Fixes # https://github.com/kcp-dev/edge-mc/issues/169


dumb0002:
cc: @MikeSpreitzer @clubanderson

MikeSpreitzer:
Isn't there a typo in "environments/dev-env/docs/wls_ubuntu_platform.md" --- do you want "wsl" rather than "wls"?

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+
+## Required Packages:
+   - ko: https://ko.build/install/ 
+   - gcc
+   - docker 
+   - jq
+   - make
+   - go (version expected 1.19)
+   - kind
+   - kubectl
```
MikeSpreitzer, line 11: Shouldn't we say something about version here, for consistency with the fact that kcp is based on Kubernetes version 1.24.3?

I think that exact match on version is not required; ideally we would document here the range of usable versions for each.

MikeSpreitzer:
I am not convinced that we want two different contributor docs, `environments/dev-env/CONTRIBUTING.md` and `CONTRIBUTING.md`.  Wouldn't it be better to have just the one, at the regular place?

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+
+
+## Supported OS Platforms 
+  - Linux
+  - MacOS
+  - Windows WSL/Ubuntu
+
+
+Run the following script to install the required package (Linux or MacOS ):
+
```
MikeSpreitzer, line 21: I would explicitly mention the `git clone` of this repo before any usage of the local copy of this repo.

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+For Windows WSL/Ubuntu platform, follow the instructions [here](docs/wls_ubuntu_platform.md)
+
+
+## Quickstart
+
+### 1. Clone the kcp-edge repo:
+
+```shell
+git clone https://github.com/kcp-dev/edge-mc.git  kcp-edge
```
MikeSpreitzer, line 40: (1) This has to come before the local usage of `install_req.sh` above.

(2) Must we mandate the renaming of the directory?  I think not.  This just adds extra dissonance.  Developers know that the git repo is named "edge-mc".  You might wish that the repo were named something different, and in the future it might even have a kcp-free name that we have not settled on yet, but at present the least dissonant thing to do is `git clone` without renaming.

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+```
+
+### 2. Build the kcp-edge binaries:
+
+```shell
+cd kcp-edge
+make build
+```
+
+The kcp-edge binaries is created inside the `/bin` directory. Add this directory to your $PATH:
```
MikeSpreitzer, line 50: `/bin` is an absolute pathname, which is not what you mean.  Better to just say `bin`.  Or `./bin` if you really want to emphasize the locality.

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+
+```shell
+export PATH="$PATH:$(pwd)/bin"
+```
+
+
+### 3. Create your own edge infrastructure (edge clusters) - Bring Your Own Cluster (BYOC)
```
MikeSpreitzer, line 57: This header mixes plural and singular; let's not.
"BYOC" is our own internal jargon.  I do not think we need to highlight this piece of jargon.  It is the normal usage pattern.  We should just show users doing what we expect users to do.

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+
+Create your edge cluster or bring your own k8s edge cluster. In this example, we use [kind](https://kind.sigs.k8s.io/) to create an edge cluster that we name “florin”:
```
MikeSpreitzer, line 59: The heading of this section starts --- quite appropriately, IMO --- with a plural clusters.  I think the content should too.  Or at least acknowledge that the general pattern is multiple clusters and say something about starting out with just one.

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+
+```shell
+kind create cluster --name florin
+```  
+
+### 4. Deploy the kcp-edge platform:
```
MikeSpreitzer, line 65: Steps a and b are kcp rather than kcp-edge (edge-mc).  See my earlier discussion with @francostellari about keeping clear on the fact that edge-mc is embedded in a context --- currently kcp but something more general in the future --- that provides a runtime-variable fleet of logical clusters.  That larger context is not part of edge-mc.

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+
+  a) Download kcp binaries for your platform:
```
MikeSpreitzer, line 67: We should probably mention the alternative of `git clone` the kcp repo and locally build and install.  This is particularly relevant for contributors.  At least for myself, I find it useful to have a local copy of the kcp code.

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+
+  ```shell
```
MikeSpreitzer, line 69: https://github.com/kcp-dev/edge-mc/pull/344 , now merged, contributes a script for this; maybe just say to use the script?

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+  ```
+  OR
+  
+  If you need to connect to a remote k8s edge cluster, you can bind kcp to the public interface of your virtual machine:
+
+  ```shell
+  kcp start --bind-address <ip-address> >& kcp_log.txt &
+  ```
+  
+  This uses default 6443 port, just bind to a specific address
```
MikeSpreitzer, line 100: This is confusing as written.  "If you need to connect to a remote k8s edge cluster": why would the server-side address depend on whether the edge cluster is remote?  BTW, the kcp server does not open connections to the edge cluster.  I suspect the issue is about the syncer in the edge cluster being able to open connections to the kcp server.  But I do not understand why specifying a particular server address helps --- the default is to bind to 0.0.0.0, which is the least restrictive thing you can do.

"the public interface of your virtual machine" --- what virtual machine?  No virtual machine has been mentioned, at least not yet.

"This uses default 6443 port, just bind to a specific address" --- the port is potentially important regardless of whether a specific server address is bound.

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+  ```shell
+  ps aux | grep -e mailbox-controller -e placement-translator -e scheduler
+  user    11344   0.1  0.1 34906508  28044 s001  S     8:36PM   0:00.43 placement-translator --allclusters-context system:admin -v=2
+  user    11333   0.1  0.1 34885544  19536 s001  S     8:36PM   0:00.38 scheduler -v 2 --root-user kcp-admin --root-cluster root --sysadm-context system:admin --sysadm-user shard-admin
+  user    11323   0.0  0.1 34892244  20780 s001  S     8:36PM   0:00.13 mailbox-controller --inventory-context=root --mbws-context=base -v=2
+  ```
+
+### 5. Connect your edge cluster to the kcp-edge platform:
+
+  a) Create inventory management workspace (`imw`):
```
MikeSpreitzer, line 139: "Create _an_ inventory management workspace"

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+
+  ```console
+  kubectl ws root
```
MikeSpreitzer, line 142: In a "console" block, we show the shell's prompts; this distinguishes input from output.

```console
$ kubectl ws root
Current context is "root"

$ kubectl ws create imw-1 --enter
Current context is "root:imw-1"
```

clubanderson: my only reservation on that suggestion - adding a '$' renders the copy button useless - does it not?

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+  ```console
+  ensure-location.sh florin  env=prod
+
+  synctarget.workload.kcp.io/florin created
+  location.scheduling.kcp.io/florin created
+  synctarget.workload.kcp.io/florin labeled
+  location.scheduling.kcp.io/florin labeled
+  ```
+
+  A location and synctarget objects are created:
```
MikeSpreitzer, line 161: I find the "are created" a bit vague.  I think it would be clearer to say that the preceding commands did this and the following commands show the effects.  Something like the following.

> The following commands list the objects that were created.

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+  kubectl get locations,synctargets
+  NAME                                RESOURCE      AVAILABLE   INSTANCES   LABELS   AGE
+  location.scheduling.kcp.io/florin   synctargets   0           1                    57s
+
+  NAME                                AGE
+  synctarget.workload.kcp.io/florin   58s
+  ```
+
+
+  The [mailbox-controller](https://docs.kcp-edge.io/docs/coding-milestones/poc2023q1/mailbox-controller/) creates a mailbox workspace for the newly created SyncTarget: `florin`:
```
MikeSpreitzer, line 173: I think setting "`florin`" off with a colon says that `florin` is the mailbox workspace for the newly created SyncTarget.  I think it would be clearer to put `florin` in parentheses right after "SyncTarget".

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+  kubectl ws root
+  Current workspace is "root".
+
+  kubectl ws tree
+  .
+  └── root
+      ├── compute
+      ├── espw
+      │   └── 19igldm1mmolruzr-mb-6b0309f0-84f3-4926-9344-81df2f989f69
+      └── imw-1
```
MikeSpreitzer, line 185: This listing is not useful in general, when there is more than one mailbox workspace.  I think it would be better to do a listing like the one shown at the end of https://docs.kcp-edge.io/docs/coding-milestones/poc2023q1/example1/#the-mailbox-controller :

```console
$ kubectl ws root:espw
$ kubectl get Workspace -o "custom-columns=NAME:.metadata.name,SYNCTARGET:.metadata.annotations['edge\.kcp\.io/sync-target-name'],CLUSTER:.spec.cluster"
NAME                                                       SYNCTARGET   CLUSTER
apmziqj9p9fqlflm-mb-b8c64c64-070c-435b-b3bd-9c0f0c040a54   florin       12299slctppnhjnn
apmziqj9p9fqlflm-mb-bf452e1f-45a0-4d5d-b35c-ef1ece2879ba   guilder      yk9a66vjms1pi8hu
```

environments/dev-env/CONTRIBUTING.md
MikeSpreitzer: Perhaps the issue is the IP address that kcp writes into the generated kubeconfig.  That is never 0.0.0.0, it picks something more specific.

environments/dev-env/CONTRIBUTING.md
francostellari: @MikeSpreitzer +/- 0.1 in the version is allowed. I had to force install kubectl v1.25.3 and not latest

environments/dev-env/CONTRIBUTING.md
francostellari: I agree

environments/dev-env/CONTRIBUTING.md
MikeSpreitzer: including the output, which we are not debating, makes the copy button useless

francostellari:
I think a lot of comments on the MD are open to personal preferences in some cases and to different level of people reading.

In my opinion, I would try to get the script from Braulio's to kcp-edge main ASAP so that we can start referencing them correctly. They may not be perfect but they seem to work, at least in most cases, we can improve on them later.

To be honest I think that the script should go into `hack` not enviroment... and eventualy into the tar.gz

For the documentation, I think we need two nice an clear step by step pages into the kcp-edge.io more than in the git enviroment... This is not a dev-env anymore but the new quickstart for a) users that want to deploy b) users that want to develop and play with the details

environments/dev-env/CONTRIBUTING.md
MikeSpreitzer: What sort of force are you talking about?  Installing kubectl is its own step, I am not aware of any force required to get whatever version you choose.

environments/dev-env/CONTRIBUTING.md
MikeSpreitzer, line 100: I suggest the following revision, for section (b) regarding starting kcp.

b) Start kcp:

```shell
kcp start --bind-address <ip-address> >& kcp_log.txt &
```

In that command, `<ip-address>` is an IP address of the machine running kcp that can be reached from the edge clusters.  This is the address that will appear in the generated kubeconfig file (mentioned next) and the kcp server will listen only at that address.  If all of the host's non-loopback addresses can be reached from the edge clusters then this is a non-issue and the `--bind-address <ip-address>` can be omitted from the command line; in this case the server will listen at all addresses and pick a non-loopback one to put in the generated kubeconfig file.

environments/dev-env/build-edge-syncer.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
```
MikeSpreitzer, line 1: (1) Naming this script "build-<something>" is confusing, because what it does is not usually called "building".

(2) We already have `bin/mailbox-prep.sh` that gets the same job done.  I do not think we want to two scripts.  If the existing script lacks needed features (such as a `--verbose` flag) then let's fix that rather than make a substantially duplicate script.

environments/dev-env/build-edge-syncer.sh
```diff
@@ -0,0 +1,56 @@
...
+done
+
+kubectl ws root:espw > /dev/null 2>&1
+workspace=$(kubectl get Workspace -o json | jq  --arg v "$syncTarget_name" -r '.items | .[] | .metadata | select(.annotations ["edge.kcp.io/sync-target-name"] == $v) | .name')
+
+# Create edge-syncer manifest
+if [ $verbosity == 1 ]; then
+    kubectl ws root:espw:$workspace
+    #kubectl kcp workload edge-sync $syncer_name --syncer-image $img -o $syncTarget_name-syncer.yaml    
+    ./kcp/bin/kubectl-kcp workload edge-sync $syncer_name --syncer-image $img -o $syncTarget_name-syncer.yaml
```
MikeSpreitzer, line 46: This script does not do anything to make sure the current working directory is anything in particular, so is not justified in assuming a particular meaning for the `.` at the start of this command line.

environments/dev-env/CONTRIBUTING.md
MikeSpreitzer: We do not own the definition of the "console" tag.

environments/dev-env/CONTRIBUTING.md
francostellari: Perhaps `kcp start [--bind-address <ip-address>] >& kcp_log.txt &`

MikeSpreitzer: That works for me.

MikeSpreitzer:
I think we are agreed that there should be files in github that get used in the introductory examples.  This raises a basic structural problem: currently we have envisaged the first user experience to consist of downloading and installing releases of kcp and edge-mc, both containing only `bin` content, and then using those.  What about those example files?

It would be great if the example files could be left on github and just referenced by URL.  IIRC `kubectl` can read from a URL as well as from a file.  However, I suspect that not all of our example files are `kubectl` input.  For one example, if you look at example1.md you will see kind cluster config files (that write-up uses a cluster config file so that it can expose ports so that the deployed apps can be exercised by `curl` or a web browser on the developer's machine).  I suspect there are other cases.

To get the example files local in the user experience, I see a few plausible options.  One is that we abandon telling (either in detail or in a bash script) users to fetch and unpack a release archive, instead tell users as well as contributors to `git clone` the edge-mc repo.  But the problem with that is that we do not want to require users to build what they just `git clone`d.  But `git clone` does not yield built binaries.  One way around that would be to maintain the example files in a distinct repo.  The user experience would continue to include fetching and unpacking archives of binaries, and be extended with `git clone` of the edge-mc-examples repo.  Alternatively, we could also pack the example files into archives.  That could be either together with the binaries in one unified archive, or in distinct archives.  The former would be a bit weird, including examples in a release, but not totally unacceptable (IMO); it would have the advantage of clearly keeping consistency.  Using distinct archives makes more steps for the user, although in the quickstart this can be under the covers of a big wrapper script.

dumb0002:
@MikeSpreitzer, yes indeed. Let's discuss and propose how we should merge the content of both CONTRIBUTING.md files

dumb0002:
Some samples repo for developer experience:

Etcd: 
   - https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md

kyverno:
  - https://github.com/kyverno/kyverno/blob/main/DEVELOPMENT.md

Should we maybe call our file DEVELOPMENT.md? Most of the CONTRIBUTING.md file that I found on public repos were focused on providing the guidelines to write code and create pull requests, etc.

environments/dev-env/CONTRIBUTING.md
dumb0002: fixed

environments/dev-env/CONTRIBUTING.md
```diff
@@ -0,0 +1,510 @@
...
+  - Windows WSL/Ubuntu
+
+
+Run the following script to install the required package (Linux or MacOS ):
+
+```bash
+./install_req.sh
+```
+
+Alternatively, run the following command for MacOS platform:
```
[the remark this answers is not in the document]
dumb0002: fixed

environments/dev-env/CONTRIBUTING.md
dumb0002: (1) added a comment stating that the script needs to be run after cloning the repo. I am avoiding to add a step to run the `install_req.sh` script in our instructions -  most of the developers will already have those packages installed in their environment or they will know how to install themselves. Our `install_req.sh` is just an option that we are providing to install the required packages

(2) Fixed

environments/dev-env/CONTRIBUTING.md
dumb0002: In our `demo`  environment using the `kcp-playground`, we are providing a `kind` for users with a sample workload deployed, the term BYOC clearly makes this distinction here.

environments/dev-env/CONTRIBUTING.md
dumb0002: fixed

environments/dev-env/CONTRIBUTING.md
dumb0002: fixed

environments/dev-env/CONTRIBUTING.md
dumb0002: Fixed

environments/dev-env/CONTRIBUTING.md
dumb0002: fixed

environments/dev-env/CONTRIBUTING.md
dumb0002: fixed

environments/dev-env/CONTRIBUTING.md
dumb0002: Added. I looked at examples in other open source repos and they add the shell's prompts to  distinguishes input from output indeed.

environments/dev-env/CONTRIBUTING.md
dumb0002: fixed

environments/dev-env/CONTRIBUTING.md
dumb0002: In this example for simplicity we are only deploying a single workload (`florin`). Then, we have only showing a single workspace. I like the tree structure because it clearly shows the workspace hierarchy used by kcp-edge

environments/dev-env/build-edge-syncer.sh
dumb0002: This script was created to be a helper functions of another main script only.

dumb0002:
@MikeSpreitzer 


> Shouldn't this PR be updating the existing CONTRIBUTING.md rather than creating an alternate file?


I created a new file DEVELOPMENT.md in the main directory of this repo, as I have seen in other open source repos, for example: https://github.com/knative/serving

dumb0002:
> I think we are agreed that there should be files in github that get used in the introductory examples. This raises a basic structural problem: currently we have envisaged the first user experience to consist of downloading and installing releases of kcp and edge-mc, both containing only `bin` content, and then using those. What about those example files?
> 
> It would be great if the example files could be left on github and just referenced by URL. IIRC `kubectl` can read from a URL as well as from a file. However, I suspect that not all of our example files are `kubectl` input. For one example, if you look at example1.md you will see kind cluster config files (that write-up uses a cluster config file so that it can expose ports so that the deployed apps can be exercised by `curl` or a web browser on the developer's machine). I suspect there are other cases.
> 
> To get the example files local in the user experience, I see a few plausible options. One is that we abandon telling (either in detail or in a bash script) users to fetch and unpack a release archive, instead tell users as well as contributors to `git clone` the edge-mc repo. But the problem with that is that we do not want to require users to build what they just `git clone`d. But `git clone` does not yield built binaries. One way around that would be to maintain the example files in a distinct repo. The user experience would continue to include fetching and unpacking archives of binaries, and be extended with `git clone` of the edge-mc-examples repo. Alternatively, we could also pack the example files into archives. That could be either together with the binaries in one unified archive, or in distinct archives. The former would be a bit weird, including examples in a release, but not totally unacceptable (IMO); it would have the advantage of clearly keeping consistency. Using distinct archives makes more steps for the user, although in the quickstart this can be under the covers of a big wrapper script.


- I removed the files related to example1.md  in this PR. I may need to ask the users to clone the repo in order to use the stage scenarios in our automation using the `kcp-playground` tool (aka demo environment)

dumb0002:
FYI, I will create another PR for the following scripts cited in this instruction: `kcp-edge.sh`, `install_req.sh`, `clean_up.sh`

environments/dev-env/CONTRIBUTING.md
MikeSpreitzer: We could show both.

MikeSpreitzer:
As another example of documentation for developers, here is what the Kubernetes project has: https://github.com/kubernetes/community/blob/master/contributors/devel/README.md .  It is much more elaborate than we should have at this point.

DEVELOPMENT.md
```diff
@@ -0,0 +1,496 @@
+# KCP-Edge Contributor Development Environment
```
MikeSpreitzer, line 1: I think the title needs adjusting; we are no longer talking about a distinct "environment" for developers, right?

DEVELOPMENT.md
```diff
@@ -0,0 +1,496 @@
...
+   - kubectl (version range expected: 1.23-1.25)
+
+
+## Supported OS Platforms 
+  - Linux
+  - MacOS
+  - Windows WSL/Ubuntu
+
+
+After cloning this repo (see quickstart), run the following script to install the required package (Linux or MacOS):
```
MikeSpreitzer, line 20: I suppose the intended referent is the "### 1. Clone the kcp-edge repo" section below; the existing reference to "quickstart" does not obviously lead there.

DEVELOPMENT.md
```diff
@@ -0,0 +1,496 @@
...
+
+### 3. Create your own edge infrastructure (edge clusters) - Bring Your Own Clusters (BYOC)
+
+Create your edge clusters or bring your own k8s edge clusters. In this example, we deploy only a single cluster using [kind](https://kind.sigs.k8s.io/). Our edge cluster is named “florin”:
+
+```shell
+kind create cluster --name florin
+```  
+
+### 4. Install KCP
```
MikeSpreitzer, line 66: I think that at this point the reader should be referred to the user documentation.
Attempting to maintain a copy here adds work for us and opportunities to make mistakes.
Having a copy of a large volume of material here also confuses the developer: she wonders why she is reading something different from users --- is there a reason?  Is there an important difference that she is overlooking?

DEVELOPMENT.md
```diff
@@ -0,0 +1,496 @@
...
+```
+
+The kcp-edge binaries is created inside the `./bin` directory. Add this directory to your $PATH:
+
+```shell
+export PATH="$PATH:$(pwd)/bin"
+```
+
+
+### 3. Create your own edge infrastructure (edge clusters) - Bring Your Own Clusters (BYOC)
```
MikeSpreitzer, line 58: I think that at this point the reader should be referred to the user documentation.
Attempting to maintain a copy here adds work for us and opportunities to make mistakes.
Having a copy of a large volume of material here also confuses the developer: she wonders why she is reading something different from users --- is there a reason?  Is there an important difference that she is overlooking?

DEVELOPMENT.md
dumb0002: yes, we'll still have separated instructions for users versus developers, aren't we?  please, clarify. thanks!

DEVELOPMENT.md
dumb0002: What steps would be included in the developer page? just the steps to clone a repo and run `make build`?

I was thinking that the developer documentation would be self-contained in a single page.

DEVELOPMENT.md
MikeSpreitzer: How about simply "Contributor Instructions"?

MikeSpreitzer: "Users" may do 'development' too --- development of higher level code.  That is why I use the terms "contributor" for developers of this package and "users".

dumb0002:
@MikeSpreitzer

> When it comes to "cleanup": the reader wonders what exactly is being removed. Please at least say that clearly. There is a big difference between removing an example usage and uninstalling the platform. Perhaps give the scripts more specific names?
> 

I updated the instructions to clearly explain what's being cleaned or removed. 

> Also, think about the granularity desired --- making tear-down less granular than set-up means potentially un-doing and re-doing work. I had the unhappy experience of doing a lot of that last year.

I added more explanation and a new instruction describing the granularity of the tear-down

DEVELOPMENT.md
dumb0002: Maybe we don't need the same level of detail in the user instructions. For example, we don't expose in the user's instructions the concepts of `mailbox workspace`, `SinglePlacementSlice` and `syncerConfig`.

DEVELOPMENT.md
dumb0002: Fixed

dumb0002:
@MikeSpreitzer,  here is the PR for scripts described in these instructions: https://github.com/kcp-dev/edge-mc/pull/354

DEVELOPMENT.md
```diff
@@ -0,0 +1,94 @@
...
+   - kubectl (version range expected: 1.23-1.25)
+
+
+## Supported OS Platforms 
+  - Linux
+  - MacOS
+  - Windows WSL/Ubuntu
+
+
+After cloning this repo (see [instructions](#1-clone-the-kcp-edge-repo)), run the following script to install the required package (Linux or MacOS):
```
MikeSpreitzer, line 20: I think "package" should be plural here.

DEVELOPMENT.md
```diff
@@ -0,0 +1,94 @@
...
+
+```bash
+hack/install_req.sh
```
MikeSpreitzer, line 23: (1) I do not see this file.
(2) Does it install Docker?  I suspect not.  If I am right then the user should not be told that this script will simply "install the required packages".

DEVELOPMENT.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+On MacOS you could use the following command as an alternative to the above:
+
+```bash
+brew install ko gcc jq make go@1.19 kind kubectl
```
MikeSpreitzer, line 29: I see no Docker here.

DEVELOPMENT.md
```diff
@@ -0,0 +1,94 @@
...
+
+```shell
+git clone https://github.com/kcp-dev/edge-mc.git
+```
+
+
+### 2. Build the kcp-edge binaries:
+
+```shell
+cd edge-mc
```
MikeSpreitzer, line 47: I would move the `cd` into step 1; it is a bit of context-setting that should _not_ be done as the contributor iterates through cycles of edit-build-test.

DEVELOPMENT.md
```diff
@@ -0,0 +1,94 @@
...
+make build
+```
+
+The kcp-edge binaries is created inside the `./bin` directory. Add this directory to your $PATH:
```
MikeSpreitzer, line 51: Not everything created by `make build` is a "binary".  Remember that scripts also go into `bin`.  That is why I say "executable" instead.

DEVELOPMENT.md
```diff
@@ -0,0 +1,94 @@
...
+export PATH="$PATH:$(pwd)/bin"
+```
+
+### 3. Install KCP
+
+  a) Download kcp binaries for your platform:
+
+  ```shell
+  hack/install-kcp-with-plugins.sh --version v0.11.0 --folder $(pwd)/kcp --create-folder
+  ```
```
MikeSpreitzer, line 63: I would point out that this script is just an automation of the instructions at https://docs.kcp.io/kcp/v0.11/#download-kcp so that the reader knows that KubeStellar does not require anything special and gets good description of what the script does without having to fetch and read the script.

DEVELOPMENT.md
```diff
@@ -0,0 +1,94 @@
...
+
+
+  b) Start kcp:
+
+  ```shell
+  kcp start >& kcp_log.txt &
+  ```
+  OR
+  
+  If you need to connect to a remote k8s edge cluster, you can bind kcp to the public interface of your virtual machine:
```
MikeSpreitzer, line 75: Sometimes when I am reading instructions in a hurry, I do things in the order I am told.  If the instructions then say "but wait, you should have done something different if FOO" then I am already in trouble --- I have done the wrong thing before I read that amendment.  I recommend rewording this so as to not lead with something that we quickly retract.

MikeSpreitzer:
The curl-to-bash technique that the quickstart says to use is convenient but wholly inadequate both for contributors and for users who are building KubeStellar into their own bigger system.  We should have documentation of how to install and launch kcp and KubeStellar, and I think that we should have one document that is usable by both of the categories of people listed above (because most of the content is the same).  The only difference between them would be between downloading and unpacking a KubeStellar release distribution vs git clone and build.  

What I see in DEVELOPMENT.md now is almost exactly what I just asked for.  Maybe we should make it that?

PR #355, opened 2023-05-09
🌱 Rely more on type inference.

MikeSpreitzer:
## Summary
This PR removes some explicit type parameterizations.  This makes lint stop telling me about unnecessary type arguments in those cases.

Unfortunately lint is so stupid that it complains about unnecessary type arguments when only some of the type arguments can be inferred, but go does not have an option to specify only some of the type parameters.

## Related issue(s)

Fixes #

/cc @waltforme

### DIFF
--- a/pkg/placement/binding-organizer.go
+++ b/pkg/placement/binding-organizer.go
@@ -121,12 +121,12 @@ func SimpleBindingOrganizer(logger klog.Logger) BindingOrganizer {
 			},
 		)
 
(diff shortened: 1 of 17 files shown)


waltforme:
/lgtm

I tried to truncate the type parameter lists wherever the linter complains about unnecessary type argument in a partially inferred type parameter list. The placement-translator still compiles (will the compiled binary work as expected?)

But I think the truncation doesn't make the code much cleaner, actually it makes the code harder to follow because the length of the type parameter list is different between a generic function's definition and instantiation.

PR #356, opened 2023-05-09
:seedling: Move the code for syncer plugin from kcp to edge-mc 

yana1205:
## Summary
- Move the code for [syncer plugin](https://github.com/yana1205/kcp/blob/emc/pkg/cliplugins/workload/plugin/edgesync.go) and the related codes to /cmd/kubectl-kcpforedgesyncer
- Modify Makefile to use kubectl-kcpforedgesyncer built from edge-mc project sources instead of using https://github.com/yana1205/kcp

Manual tests (kyverno) and E2E tests passed.

## Related issue(s)

Fixes #


cmd/kubectl-kcpforedgesyncer/cliplugins/base/options.go
```diff
@@ -0,0 +1,90 @@
+/*
```
MikeSpreitzer, line 1: This file has the same contents as the corresponding one in kcp.

(1) Would it be better to import the existing package from kcp than to make a copy?

(2) If we do want to make a copy, would it be better to put it in `pkg/`, analogous to where the original sits in kcp and following the usual pattern?

cmd/kubectl-kcpforedgesyncer/cliplugins/workload/cmd/cmd.go
```diff
@@ -0,0 +1,82 @@
+/*
```
MikeSpreitzer, line 1: Would it be better to put all this `/cliplugins/` stuff in `pkg/cliplugins/`, like is done in kcp and following the usual pattern?

go.mod
```diff
@@ -138,7 +139,7 @@ require (
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
 	github.com/google/gnostic v0.5.7-v3refs // indirect
-	github.com/google/go-cmp v0.5.8 // indirect
+	github.com/google/go-cmp v0.5.8
```
MikeSpreitzer, line 142: Shouldn't this move to the first block of require with all the other non-indirect ones?

cmd/kubectl-kcpforedgesyncer/cliplugins/base/options.go
yana1205: I didn't notice that the contents are same.
Fixed to import the package from kcp instead of copy&paste.

go.mod
yana1205: Thanks for informing it. I rerun `go mod tidy` and this package is now in the first required section.

yana1205:
FYI: I rebased the branch to be based on the top of main branch.

MikeSpreitzer (approved):
/lgtm
Thanks!

PR #364, opened 2023-05-12
🐛 Remove installation of edge CRDs in ESPW

MikeSpreitzer:
## Summary
This PR updates the various instructions to stop installing the edge CRD objects in the ESPW.  The contents of `config/exports/` are all that is needed, and adding the CRD objects might actually be causing a redundancy problem.

## Related issue(s)

Fixes #

/cc @yana1205 
/cc @waltforme 
/cc @dumb0002 
/cc @francostellari


waltforme:
Confirmed this by trying the scheduler so
/lgtm

PR #378, opened 2023-05-16
❇️ Update QUICKSTART.md

clubanderson:
added instructions to prevent error with kind cluster pre-existence, curl not working, and guilder prep not running due to wrong workspace location.

## Related issue(s)

Fixes #

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -72,11 +72,15 @@ Create the first edge cluster:
 kind create cluster --name florin --config  kubestellar/examples/florin-config.yaml
 ```  
 
+Note: if you already have a cluster named 'florin' from a previous installation of KubeStellar, please delete the florin cluster ('kind cluster delete florin') and create it using the instruction above.
+
 Create the second edge cluster:
 
 ```shell
 kind create cluster --name guilder --config  kubestellar/examples/guilder-config.yaml
 ```  
+
+Note: if you already have a cluster named 'guilder' from a previous installation of KubeStellar, please delete the guilder cluster ('kind cluster delete guilder') and create it using the instruction above.
 
 ### b. Create a KubeStellar Inventory Management Workspace (IMW) and Workload Management Workspace (WMW)
 
@@ -103,6 +107,7 @@ A WMW does not have to be created before the edge cluster is on-boarded; the WMW
 Let's begin by onboarding the `florin` cluster:
 
 ```shell
+kubectl ws root
 kubectl kubestellar prep-for-cluster --imw root:example-imw florin env=prod
 ```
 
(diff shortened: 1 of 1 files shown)


clubanderson:
- when I went to create the kind clusters I found that I already had them from a previous install test.  Can we state… “if you already have florin or guilder clusters, please stop and delete them now, and then create them using these instructions”
kcp-edge term is in the namespace, serviceaccount, secret, clusterrole, syncer….etc.  Is there a plan to replace these with KubeStellar?
- when I went to ‘onboard’ the guilder cluster I got the error
andan02@Andys-MacBook-Pro  ~/temp  kubectl kubestellar prep-for-cluster --imw root:example-imw guilder env=prod                                              :heavy_check_mark:  8975  19:12:19
Error: current URL "https://127.0.0.1:64837/" does not point to a workspace
Note: 'kubectl ws' now matches 'cd' semantics: go to home workspace. 'kubectl ws -' to go back. 'kubectl ws .' to print current workspace.
Error: the server could not find the requested resource (get [workspaces.tenancy.kcp.io](http://workspaces.tenancy.kcp.io/) ~)"
I issued: ‘kubectl ws root’ to proceed.  Can we please add this to the instructions before onboarding guilder?
6. Finally, when I applied the apache stuff, it took a minute before I got anything more than ‘curl: (52) Empty reply from server’.  Please add a note to ‘wait a minute if you receive a curl error’ (edited)

QUICKSTART.md
```diff
@@ -72,12 +72,16 @@ Create the first edge cluster:
 kind create cluster --name florin --config  kubestellar/examples/florin-config.yaml
 ```  
 
+Note: if you already have a cluster named 'florin' from a previous installation of KubeStellar, please delete the florin cluster ('kind cluster delete florin') and create it using the instruction above.
```
MikeSpreitzer, line 75: Those kind clusters are from the example usage, not the KubeStellar installation.  How about s/installation of KubeStellear/exercise of KubeStellar/ ?

QUICKSTART.md
```diff
@@ -334,6 +340,7 @@ which should yield:
   </body>
 </html>
 ```
+NOTE: if you receive the error: 'curl: (52) Empty reply from server', wait 2 minutes and attempt curl again.  It takes a minute for the Apache HTTP Server to synchronize and start.
```
MikeSpreitzer, line 343: I wonder if "2 minutes" is too specific.  We can not promise that 2 minutes is enough.  If we promise one specific obviously large enough time, that will be wasteful and embarrasing in almost all cases. Things seem to run more slowly for me than for @francostellari .  I can attest that I had to wait a while, but did not measure how long.  Maybe say something more vague, like keep retrying, it should work in a small number of minutes?

clubanderson: see line 361 and 362 below

MikeSpreitzer (approved):
I do not know why the two occurrences of the same issue should be handled differently, but we can continue this in follow-on.
/lgtm
Thanks!

PR #387, opened 2023-05-17
🐛 Remove overspecific and unnecessary args to bootsrap in quickstart

MikeSpreitzer:
## Summary
This PR removes the problematic literal version "0.2.0" from the quickstart, because it will soon be wrong, is always impossible to keep correct when using proper release discipline, and the default does what we want.

## Related issue(s)

Fixes #

/cc @francostellari 
/cc @dumb0002 
/cc @clubanderson


clubanderson (approved):
/lgtm

PR #403, opened 2023-05-17
🐛 Factor ESPW setup out of `kubestellar start`

### LINKED ISSUE #401: bug: ESPW create/delete should be bundled with install/uninstall not start/stop
### Describe the bug

In common Uniix/Linux packages, starting and stopping the daemons does not create/delete the supporting state (e.g., files in /etc).  We should follow the usual pattern.

### Steps To Reproduce

read the doc or experiment

### Expected Behaviour

Unsurprising bundling of behavior

### Additional Context

_No response_

### LINKED ISSUE #402: bug: bootstrap decides whether to `kubestellar start` based ONLY on controllers
### Describe the bug

bootstrap-kubestellar decides whether to invoke `kubestellar start` based ONLY on whether the controllers are all running --- not considering whether the ESPW exists.  Yet `kubestellar start` is what establishes the ESPW (which is a bug, see #401).

Consequently, a user who has the controllers running but no ESPW will get a broken install from bootstrap-kubestellar.

### Steps To Reproduce

See above.

### Expected Behaviour

No inappropriate bundling.  No confusion from bundling.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR separates ESPW setup from `kubestellar start`, into a new subcommand `kubestellar init`.

This PR also removes some inappropriate precondition checking: for docker and for go version.

## Related issue(s)

Fixes #401
Fixes #402 

/cc @dumb0002 
/cc @francostellari 
/cc @clubanderson 
@eaepstein

### DIFF
--- a/bootstrap/bootstrap-kubestellar.sh
+++ b/bootstrap/bootstrap-kubestellar.sh
@@ -238,6 +238,8 @@ if [ "$(kubestellar_installed)" == "false" ]; then
     fi
 fi
 
+kubestellar init  $verbose
+
 # Ensure KubeStellar is running
 if [ "$(kubestellar_running)" == "false" ]; then
     if [ "$verbose" != "" ]; then
--- a/docs/content/en/docs/Coding Milestones/PoC2023q1/commands.md
+++ b/docs/content/en/docs/Coding Milestones/PoC2023q1/commands.md
@@ -17,23 +17,12 @@ be invoked directly using any pathname (not in your `$PATH`).
 
 ## Platform control
 
-The `kubestellar` command has two subcommands, one for setup and one
-for teardown.
+The `kubestellar` command has three subcommands, one to finish setup
+and two for process control.
 
-### Kubestellar start
-
-This subcommand is used after installation.
-This subcommand does two things:
-1. It creates and populates the edge service provider workspace (ESPW) if it does not already exist; and
-2. It stops any running kubesteallar controllers and then starts them all.
-
-```shell
-kubestellar start [flags]
-```
-
-The flags can appear before and/or after the subcommand.
-
-The available flags are as follows.
+This command accepts the following command line flags, which can
+appear before and/or after the subcommand.  The `--log-folder` flag is
+only used in the `start` subcommand.
 
 - `-V` or `--verbose`: calls for more verbose output.  This is a
   binary choice, not a matter of degree.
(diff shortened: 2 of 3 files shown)


dumb0002:
@MikeSpreitzer, to solve this issue I proposed the following:

  1 - Update  `kubestellar`  with the following flags (similar to  Linux `systemctl`  command): start | stop | restart
          a) start - flag checks if the kubestellar controllers are already running and does nothing if they are already running

  2 - `bootstrap-kubestellar.sh`: does not need to check if the kubestellar are running but it does the following:
           a) create espw
          b)  delete espw
          c) check if espw already exists
          d) check if kcp is running and so ...

In the way, we clearly separate the functionality for both scripts `kubestellar`  and ` bootstrap-kubestellar.sh` - each of them only checks and deletes the components that they are responsible for.

MikeSpreitzer:
@dumb0002 : I think we do not want to put more functionality than necessary into the script that we are telling people to curl directly to bash.  Many people simply refuse to do that, and will at least read the script before doing that.  And regular usage will not do that, we are only talking about a special thing for the quickstart.  So I think the ESPW creation belongs in an executable in `bin`.  I am OK with bundling ESPW deletion with a broader uninstall command (also in an executable in `bin`).

dumb0002:
@MikeSpreitzer, Yes, we can keep the ESPW setup into in an executable in bin indeed.  However, scripts like `kubestellar` will be used by both user and developer persona.  The developer may not need to run `bootstrap-kubestellar.sh` to build his development environment. The developer may start the kcp server from a cloned repo and jump directly to use the `kubestellar start` script only. Others functionalities like `kubestellar start | stop| restart` will be useful for that persona. Otherwise, we may need to write another  kubestellar like script for the developer persona. I am ok with that as long as we decide that this is the design that we are going to target.

MikeSpreitzer:
@dumb0002 : for the contributor, indeed we do not expect them to run bootstrap-kubestellar --- that is specifically designed for the assumption that a `git clone` is NOT being used.

I am OK with refactoring the current start (which is really stop-then-start) and stop into start, restart, and stop.  That would align better with expectations for a systemd service, which is a familiar management paradigm in Linux.

MikeSpreitzer:
I am closing this PR in favor of #408 and #409, for the sake of evolutionary discipline.
These do not attempt to refactor `restart` out of `start`, that can be additional PR(s).

PR #408, opened 2023-05-18
🐛 Fix bootstrap-kubestellar condition for calling start

### LINKED ISSUE #402: bug: bootstrap decides whether to `kubestellar start` based ONLY on controllers

MikeSpreitzer:
## Summary
This PR expands the conditions underwhich bootstrap-kubestellar.sh calls `kubestellar start`, to include absence of the ESPW.

I am factoring #403 into two pieces, and this is one of them, to obey the necessary evolutionary discipline.  This change will get to users before a new release is made, and so has to work with the current release.

## Related issue(s)

Fixes #402

### DIFF
--- a/bootstrap/bootstrap-kubestellar.sh
+++ b/bootstrap/bootstrap-kubestellar.sh
@@ -238,10 +238,12 @@ if [ "$(kubestellar_installed)" == "false" ]; then
     fi
 fi
 
+kubectl ws root
+
 # Ensure KubeStellar is running
-if [ "$(kubestellar_running)" == "false" ]; then
+if [ "$(kubestellar_running)" == "false" ] || ! kubectl get workspaces.tenancy.kcp.io espw &> /dev/null ; then
     if [ "$verbose" != "" ]; then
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
I have successfully tested the bootstrap script from this PR, by copying it to an empty directory and running it there.  I then tore partway down and bootstrapped again, to mimic the pathological scenario in #402 .  The revision here seemed to work correctly.

MikeSpreitzer:
@eaepstein

clubanderson (approved):

PR #409, opened 2023-05-18
✨ Introduce kubestellar init

MikeSpreitzer:
## Summary
This PR starts the process of factoring the ESPW control out of the process control.  In particular:
- `kubestellar init` is a new command that just ensures that the ESPW is setup;
- `kubestellar start` still implicitly ensures that the ESPW is setup, so that the bootstrap script does not have to change simultaneously;
- `kubestellar stop` no longer destroys the ESPW (this change is OK to make here because we have no script that invokes `kubestellar stop` and the only recipe, the quickstart, follows it with such obliteration that this detail matters not).

This is part of the project of factoring #403 to respect evolutionary discipline; the other part is #408.

## Related issue(s)

This is the first permissible steps to addressing #401

### DIFF
--- a/scripts/kubestellar
+++ b/scripts/kubestellar
@@ -28,8 +28,9 @@
 
 set -e
 
-install=""
+subcommand=""
 verbosity=0
+verbdir="&> /dev/null"
 remove=0
 cleanup=0
 espw_name="espw"
@@ -39,17 +40,16 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
 while (( $# > 0 )); do
     case "$1" in
-    (start)
-        install=1;; 
-    (stop) 
-        install=0;; 
+    (start|stop|init)
+        subcommand=$1;;
     (--log-folder)
         if (( $# > 1 ));
         then { log_folder="$2"; shift; }
         else { echo "$0: missing log folder" >&2; exit 1; }
         fi;;
     (--verbose|-V)
-        verbosity=1;;
+        verbosity=1
+	verbdir="";;
     (-h|--help)
         echo "Usage: $0 [start| stop][--log-folder log_folder] [-V|--verbose]"
         exit 0;;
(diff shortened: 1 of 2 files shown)


scripts/kubestellar
```diff
@@ -97,6 +86,24 @@ then
...
 fi
 
+function ensure_espw() {
+    if kubectl get Workspace "$espw_name" &> /dev/null; then
+	echo "espw workspace already exists -- using it:"
+	kubectl ws "$espw_name"
+    else 
+        kubectl ws create "$espw_name" --enter
+    fi
+    kubectl apply -f $SCRIPT_DIR/../config/exports $verbdir
```
dumb0002, line 96: It looks great, thanks for making these changes. Just a question: do we need to apply the crds here to. For example: ``kubectl apply -f   $SCRIPT_DIR/../config/crds ``

MikeSpreitzer: No.  It was always unneeded, and possibly a mistake, to put the CRDs into the ESPW.

dumb0002: It install CRDs and APIExport in the ESPW was in our old instructions. For example see instruction to deploy the KubeStellar scheduler: https://docs.kubestellar.io/docs/coding-milestones/poc2023q1/kubestellar-scheduler/. Although the command was updated but the text still says: 

`` Create the Edge Service Provider Workspace (ESPW) and populate it with CRDs and APIs``

dumb0002: We will need to revise our docs to remove any reference to that old instruction in the docs

MikeSpreitzer: We have been making some mistakes for a long time.  We should stop making them.  Yes, we should update the doc too.

MikeSpreitzer: @dumb0002 , regarding https://github.com/kcp-dev/edge-mc/pull/409/files#r1198971583 : those revisions should be made, but need not be coordinated with this PR.

dumb0002: Sounds good.

MikeSpreitzer:
The force-push to ee6ca5ae is a rebase.

Issue #413, opened 2023-05-18
bug: bootstrap-kubestellar crashes when kcp running, not good KUBECONFIG


MikeSpreitzer:
### Describe the bug

If bootstrap-kubestallar is invoked in a situation where kcp is already running but, in the shell running bootstrap-kubestellar, kubectl is not configured to talk to kcp then the script fails with an error message that is meaningless to the user.

### Steps To Reproduce

See above

### Expected Behaviour

Error message that is meaningful to the user.

MikeSpreitzer:
@dumb0002

MikeSpreitzer:
Resolved by #416

Issue #406, opened 2023-05-18
bug: bootstrap-kubestellar.sh fetches latest other scripts from web


MikeSpreitzer:
### Describe the bug

bootstrap-kubestellar.sh has two places where it fetches another script from github, and it fetches the latest version of that script.  This can create problems with evolution.  The latest version of those lower-level scripts have to keep working with all the old versions of bootstrap-kubestellar.sh (that use those scripts).

In built software we use static binding so that caller and callee can evolve together.

### Steps To Reproduce

Read the code

### Expected Behaviour

We can evolve the higher and lower level functionality jointly.

MikeSpreitzer:
@francostellari 
@dumb0002 
@clubanderson

MikeSpreitzer:
For now I think the most pragmatic solution is to just stop calling out to lower level scripts; copy their content into bootstrap-kubestellar.sh (and this will enable simplification regarding command line parsing by the lower level function).

MikeSpreitzer:
Resolved by #416

Issue #400, opened 2023-05-17
bug: bootstrap-kubestellar should report which version is running


MikeSpreitzer:
### Describe the bug

It would be helpful to get version information in the output.

### Steps To Reproduce

See doc

### Expected Behaviour

More information

MikeSpreitzer:
@francostellari 
@dumb0002 
@clubanderson

clubanderson:
@MikeSpreitzer how should this be done?

MikeSpreitzer:
Moot now.

Issue #395, opened 2023-05-17
feature: need kubestellar command that reports release semver


MikeSpreitzer:
### Feature Description

Users need a way to inquire which release of KubeStellar they have.

scripts/kubectl-kubestellar-prep_for_syncer needs to know the right tag to use in the default syncer image reference.

### Proposed Solution

Not sure.

### Alternative Solutions

Muddle on with confused users and double releases.

### Want to contribute?

- [ ] I would like to work on this issue.

MikeSpreitzer:
@clubanderson 
@francostellari 
@yana1205 
@dumb0002 
@waltforme

MikeSpreitzer:
This is half done and I am closing it with the remainder in #422 .

PR #416, opened 2023-05-18
:sparkles: More reliable bootstrap

francostellari:
## Summary

1. do not use external scrips anymore
2. introduces additional checks on kcp (including version) and kubestellar
3. more messages even without -V
4. suggest solutions on errors

## Related issue(s)
#400 ? could be better if #395 is solved i.e. `kubestellar version`
#406 
#413

Fixes #

### DIFF
--- a/bootstrap/bootstrap-kubestellar.sh
+++ b/bootstrap/bootstrap-kubestellar.sh
@@ -40,6 +40,21 @@ kcp_installed() {
     fi
 }
 
+kcp_download() {
+    if [ $verbose == "true" ]; then
+        curl -SL -o kcp.tar.gz "https://github.com/kcp-dev/kcp/releases/download/${kcp_version}/kcp_${kcp_version//v}_${os_type}_${arch_type}.tar.gz"
+        curl -SL -o kcp-plugins.tar.gz "https://github.com/kcp-dev/kcp/releases/download/${kcp_version}/kubectl-kcp-plugin_${kcp_version//v}_${os_type}_${arch_type}.tar.gz"
+    else
+        curl -sSL -o kcp.tar.gz "https://github.com/kcp-dev/kcp/releases/download/${kcp_version}/kcp_${kcp_version//v}_${os_type}_${arch_type}.tar.gz"
+        curl -sSL -o kcp-plugins.tar.gz "https://github.com/kcp-dev/kcp/releases/download/${kcp_version}/kubectl-kcp-plugin_${kcp_version//v}_${os_type}_${arch_type}.tar.gz"
+    fi
+}
+
+kcp_install() {
+    tar -C $kcp_folder -zxf kcp-plugins.tar.gz
+    tar -C $kcp_folder -zxf kcp.tar.gz
+}
+
 kcp_running() {
     if [ "$(pgrep -f 'kcp start')" == "" ]; then
         echo "false"
@@ -56,8 +71,22 @@ kcp_ready() {
     fi
 }
 
+kcp_version() {
+    echo "$(kubectl version --short 2> /dev/null | grep kcp | sed 's/.*kcp-//')"
+}
+
 kcp_get_latest_version() {
     curl -sL https://github.com/kcp-dev/kcp/releases/latest | grep "</h1>" | head -n 1 | sed -e 's/<[^>]*>//g' | xargs
+}
+
+kubeconfig_valid() {
+    if [ "$KUBECONFIG" == "" ]; then
+        echo "false"
+    elif [ -f "$KUBECONFIG" ]; then
+        echo "true"
+    else
+        echo "false"
+    fi
 }
 
 kubestellar_installed() {
@@ -66,6 +95,18 @@ kubestellar_installed() {
     else
         echo "true"
     fi
+}
+
+kubestellar_download() {
+    if [ $verbose == "true" ]; then
+        curl -SL -o kubestellar.tar.gz "https://github.com/kcp-dev/edge-mc/releases/download/${kubestellar_version}/kubestellar_${kubestellar_version}_${os_type}_$arch_type.tar.gz"
+    else
+        curl -sSL -o kubestellar.tar.gz "https://github.com/kcp-dev/edge-mc/releases/download/${kubestellar_version}/kubestellar_${kubestellar_version}_${os_type}_$arch_type.tar.gz"
+    fi
+}
+
+kubestellar_install() {
+    tar -C $kubestellar_folder -zxf kubestellar.tar.gz
 }
 
 kubestellar_running() {
@@ -97,9 +138,19 @@ get_arch_type() {
 }
 
 get_full_path() {
+    # echo "check folder '$1'"
     echo "$(cd "$1"; pwd)"
 }
 
+ensure_folder() {
+    # echo "ensure folder '$1'"
+    if [ -d "$1" ]; then :
+    else
+        mkdir -p "$1"
+    fi
+}
+
+KCP_REQUIRED_VERSION="v0.11.0"
 kcp_version=""
 kubestellar_version=""
 os_type=""
@@ -108,10 +159,13 @@ folder=""
 kcp_address=""
 kubestellar_imw=""
 kubestellar_wmw=""
-verbose=""
+verbose="false"
 flagx=""
+user_exports=""
 
-echo "KubeStellar bootstrap started..."
+echo "************************************************"
+echo "KubeStellar bootstrap started"
+echo "************************************************"
 
 while (( $# > 0 )); do
     case "$1" in
@@ -156,7 +210,7 @@ while (( $# > 0 )); do
         else { echo "$0: missing installation folder" >&2; exit 1; }
         fi;;
     (--verbose|-V)
-        verbose="-V";;
+        verbose="true";;
     (-X)
 	set -x
 	flagx="-X";;
@@ -174,7 +228,7 @@ while (( $# > 0 )); do
 done
 
 if [ "$kcp_version" == "" ]; then
-    kcp_version=v0.11.0
+    kcp_version=$KCP_REQUIRED_VERSION
 fi
 
 if [ "$kubestellar_version" == "" ] || [ "$kubestellar_version" == latest ]; then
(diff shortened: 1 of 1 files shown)


francostellari:
@MikeSpreitzer 
@dumb0002 
cc @clubanderson

bootstrap/bootstrap-kubestellar.sh
```diff
@@ -194,67 +248,131 @@ if [ "$folder" == "" ]; then
...
+        exit 3
+    fi
+    echo "Using 'KUBECONFIG=$KUBECONFIG'"
+    echo "Waiting for kcp to be ready... it may take a while"
+    until $(kcp_ready)
+    do
+        sleep 1
+    done
+    if [ "$(kcp_version)" != "$kcp_version" ]; then
+        echo "kcp installed version $(kcp_version) does not match the desired version $kcp_version ... exiting!"
```
MikeSpreitzer, line 293: How about s/installed/running/ here?

MikeSpreitzer:
Regarding verbosity: I think that in the low-verbosity setting the output should focus on user-visible actions or the deliberate lack thereof.  Stuff like the following, which is really just a minor note about progress of the script, deserves at most one line rather than three, and I think zero is the best choice.

```bash
echo "************************************************"
echo "Ensure kcp is installed"
echo "************************************************"
```

bootstrap/bootstrap-kubestellar.sh
francostellari: ok thank you

francostellari:
@MikeSpreitzer somehow I cannot reply to your comment on the comments.
I changed to this style to match Braulio's in `kubestellar`

bootstrap/bootstrap-kubestellar.sh
```diff
@@ -194,67 +248,131 @@ if [ "$folder" == "" ]; then
...
 fi
 
 # Ensure kcp is running
-if [ "$(kcp_running)" == "false" ]; then
+echo "************************************************"
+echo "Ensure kcp is running"
+echo "************************************************"
+if [ "$(kcp_running)" == "true" ]; then
+    echo "kcp process is running already: pid=$(pgrep kcp) ... skip running."
+    if [ "$(kubeconfig_valid)" == "false" ]; then
```
MikeSpreitzer, line 282: The KUBECONFIG envar is not the only relevant factor in kubectl configuration.  There is also the `$HOME/.kube/config` file.  Perhaps a functional test would be better?

bootstrap/bootstrap-kubestellar.sh
```diff
@@ -194,67 +248,131 @@ if [ "$folder" == "" ]; then
 fi
 
 # Ensure kcp is installed
-if [ "$(kcp_installed)" == "false" ]; then
-    if [ "$verbose" != "" ]; then
-        echo "Installing kcp..."
+echo "************************************************"
+echo "Ensure kcp is installed"
+echo "************************************************"
+if [ "$(kcp_installed)" == "true" ]; then
```
MikeSpreitzer, line 254: I see that does not test whether the `kubectl-ws` plugin is on `$PATH`.  Not sure whether you want to bother.

bootstrap/bootstrap-kubestellar.sh
```diff
@@ -194,67 +248,131 @@ if [ "$folder" == "" ]; then
...
     else
-        if [ "$verbose" != "" ]; then
-            echo "Starting kcp bound to address $kcp_address..."
-        fi
+        echo "Running kcp bound to address $kcp_address... logfile=$PWD/kcp_log.txt"
         kcp start --bind-address $kcp_address >& kcp_log.txt &
     fi
-    export KUBECONFIG="$(pwd)/.kcp/admin.kubeconfig"
+    export KUBECONFIG="$PWD/.kcp/admin.kubeconfig"
+    echo "Waiting for kcp to be ready... it may take a while"
```
MikeSpreitzer, line 307: I see two copies of the same wait.  Can this be re-organized so that the different paths converge before the waiting?

bootstrap/bootstrap-kubestellar.sh
```diff
@@ -194,67 +248,131 @@ if [ "$folder" == "" ]; then
...
     sleep 10
-    echo "Export KUBECONFIG with the command: export KUBECONFIG='$(pwd)/.kcp/admin.kubeconfig'"
+    if [ "$(kcp_version)" != "$KCP_REQUIRED_VERSION" ]; then
+        echo "kcp version $(kcp_version) is not supported, KubeStellar requires kcp $KCP_REQUIRED_VERSION ... exiting!"
+        exit 4
+    else
+        echo "kcp version $(kcp_version) ... ok"
+    fi
+    echo "Export KUBECONFIG environment variable: export KUBECONFIG=\"$KUBECONFIG\""
+    user_exports="$user_exports"$'\n'"export KUBECONFIG=\"$KUBECONFIG\""
```
MikeSpreitzer, line 321: So the user is going to be told twice to do this `export`?

bootstrap/bootstrap-kubestellar.sh
```diff
@@ -194,67 +248,131 @@ if [ "$folder" == "" ]; then
...
+        exit 4
+    else
+        echo "kcp version $(kcp_version) ... ok"
+    fi
+else
     if [ "$kcp_address" == "" ]; then
-        if [ "$verbose" != "" ]; then
-            echo "Starting kcp..."
-        fi
+        echo "Running kcp... logfile=$PWD/kcp_log.txt"
```
MikeSpreitzer, line 300: Shouldn't the kcp log go under the `$kcp_folder`?  (Yes, that would have to be defined even in the non-download case.)

bootstrap/bootstrap-kubestellar.sh
```diff
@@ -194,67 +248,131 @@ if [ "$folder" == "" ]; then
...
-    if [ "$verbose" != "" ]; then
-        echo "Starting or restarting Kubestellar..."
+echo "************************************************"
+echo "Ensure KubeStellar is running"
+echo "************************************************"
+kubectl ws root > /dev/null
+if [ "$(kubestellar_running)" == "true" ]; then
+    echo "KubeStellar processes are running ... ok"
+    if ! kubectl get workspaces.tenancy.kcp.io espw &> /dev/null ; then
+        echo "KubeStellar ESPW does not exists ... run 'kubestellar stop' first ... exiting!"
```
MikeSpreitzer, line 358: Was it a considered choice to not do `kubestellar start` in this case?  `kubestellar start` would stop all the running controllers, ensure the ESPW exists, and start all the controllers.

MikeSpreitzer:
This _is_ progress.

MikeSpreitzer:
I still think that it is premature to have this script attempt to create an IMW or WMW, since it can only create them as direct children of root and the usage scenario does not need this script to create either.

bootstrap/bootstrap-kubestellar.sh
MikeSpreitzer: Erm, ugh.  In the already-on-PATH case, the next question is whether to put the log alongside whatever `bin` holds `kcp`.  So, yeah, it's a bit irregular but perhaps the most pragmatic to just put the kcp log right in the CWD.

bootstrap/bootstrap-kubestellar.sh
francostellari: currently is testing that KUBECONFIG is not empty and points to an existing file.
Additionally, if it is pointing to a non valid file, I expect that kcp_ready and kcp_version would not pass

bootstrap/bootstrap-kubestellar.sh
francostellari: I will add it

bootstrap/bootstrap-kubestellar.sh
francostellari: probably

bootstrap/bootstrap-kubestellar.sh
francostellari: I also was about to change that I was not sue if it was ok to do so.
What if kcp is pre-installed and I cannot write to that location? (/usr/local/bin)
So extatically looks better but it may have more risks or require more complex checks and foldbacks.

bootstrap/bootstrap-kubestellar.sh
francostellari: I decided to exit with an error to let the user decide what to do.
I think the cleaner solution could be a `kubestellar stop` && `kubesterllar start`.

bootstrap/bootstrap-kubestellar.sh
francostellari: ok, I had a conflicting thought about this... so I'm ok to change it to what make more sense
Having it summarize at the end is nice and evident and convenient for the user.
However, if the script fails, it may not get to the end and it may be useful to have the info available.

Technically it is 3 times because we explicitly put in also in the quickstart for copy&paste convenence.

francostellari:
@MikeSpreitzer what about completely removing the `--kcp-version` flag and fix the required dependency (may be a range in the future)?

francostellari:
@MikeSpreitzer regarding the `--imw` and `--wmw` my view is that it's a non-advertised feature (i.e. we don't show any usage example) so most likely nobody will use it. However, it may be convenient for some automation I would like to use it myself, if possible.
It does not seem to hurt anything really.
So my preference would be to keep it.

francostellari:
@MikeSpreitzer @dumb0002 what about
from:
```shell
echo "************************************************"
echo "Ensure kcp is installed"
echo "************************************************"
```
to
```shell
echo "<Ensure kcp is installed>----------------------"
```

bootstrap/bootstrap-kubestellar.sh
MikeSpreitzer: My point is that kubectl does NOT _require_ `$KUBECONFIG` to be set.  If there is a kubeconfig file at `$HOME/.kube/config` then that is enough to make kubectl happy.  If we were being totally good Kubernetes citizens, we would not require more than kubectl requires.

bootstrap/bootstrap-kubestellar.sh
MikeSpreitzer: You know about the `trap` builtin command of bash?  It can be used to ensure something happens even if the script crashes.  It is like `try .. finally` in Python.

MikeSpreitzer:
I think that removing `--kcp-version` for now is perfectly OK, there is no choice to convey there.

Regarding the options to create IMW or WMW: if you want them the perhaps they should be done right, allowing a workspace pathname rather than force them to be direct children of root.  It would not be hard to split a workspace pathname by colons and iterate over it like `mkdir -p`.  But that would be more functionality than we should put in curl-to-bash, it should go in an executable.

Smaller progress notes are better than huge ones.  I still think that in low verbosity they should just be omitted.

francostellari:
regarding the `k ws create <name> -p` I'm unsure how to determine the tree structure.
Should I also detect `root:` at te beginning  as `c:\` or '/' ?... I guess so.

Assuming this exist `root:a:b`.
Currently `k ws "root:a:b"` bring you to `root:a:b` independently from where you start
Also `k ws "b"` only works from 'root:a` and `k ws "a:b"` never works
Should `k ws "a:b"` be recognized as relative path and work only from `root`?
Also do we need special `.` ad `..`?

francostellari:
I also noticed and I forgot to remove the cleanup

dumb0002:
> @MikeSpreitzer @dumb0002 what about from:
> 
> ```shell
> echo "************************************************"
> echo "Ensure kcp is installed"
> echo "************************************************"
> ```
> 
> to
> 
> ```shell
> echo "<Ensure kcp is installed>----------------------"
> ```

@MikeSpreitzer,  yes I agree!

bootstrap/bootstrap-kubestellar.sh
MikeSpreitzer: So it is a considered choice.  Good enough for me.

MikeSpreitzer:
@francostellari , regarding the `mkdir -p` -ish way to create a workspace:

(1) In kcp v0.11.0 the workspaces are arranged in a forest.  Each path is formatted as names with slash as a _separator_ **between** names.  There is no initial slash.  The first thing that appears is the name of the root of the tree.  No special weirdness for Windows, these have nothing to do with filenames.

(2) As far as I know, nobody has ever bothered to test whether our stuff works in any tree other that the one whose root is named "root".

(3) Code this complex should not be in curl-to-bash.

MikeSpreitzer:
Following is a demo of some of the weirdness that we have.  Apparently `~` is special, as is "root".

```console
$ kubectl ws .
Current workspace is "root".

$ kubectl ws tree
.
└── root
    ├── compute
    ├── espw
    │   ├── 1u2rb6sxiexmu65c-mb-28f91235-6ed6-4a4f-b89b-858fd41fc9a6
    │   └── 1u2rb6sxiexmu65c-mb-87616bb0-df12-464a-a9e0-ea757870be7c
    └── imw-1

$ kubectl ws \~

$ kubectl ws tree
.
└── kvdk2spgmbix

$ kubectl ws create fubar
Workspace "fubar" (type root:universal) created. Waiting for it to be ready...
Workspace "fubar" (type root:universal) is ready to use.

$ kubectl ws fubar
Current workspace is "kvdk2spgmbix:fubar" (type root:universal).

$ kubectl ws root
Current workspace is "root".

$ kubectl ws kvdk2spgmbix
Error: workspaces.tenancy.kcp.io "kvdk2spgmbix" not found

$ kubectl ws root:compute
Current workspace is "root:compute".

$ kubectl ws \~:fubar
Error: invalid workspace name format: ~:fubar

$ kubectl ws \~

$ kubectl ws fubar
Current workspace is "kvdk2spgmbix:fubar" (type root:universal).

$ kubectl ws tree
.
└── fubar

$ kubectl ws ..

$ kubectl ws tree
.
└── kvdk2spgmbix
    └── fubar

```

MikeSpreitzer:
This is progress, with the possible exception of insisting that KUBECONFIG be set.
I will test it now...

MikeSpreitzer:
https://github.com/kcp-dev/edge-mc/issues/441 is a bug in the current release.

MikeSpreitzer:
https://github.com/kcp-dev/edge-mc/issues/442 too.
These are just FYIs.

MikeSpreitzer:
Successful test.
There is more work to be done, but I think this can merge.
/lgtm
/approve

PR #433, opened 2023-05-20
✨ Update boilerplate.sh.txt

clubanderson:
update kcp with kubestellar

## Related issue(s)

Fixes #


clubanderson:
@MikeSpreitzer I think I got all the shell scripts in here.

PR #456, opened 2023-05-24
✨ Update florin-config.yaml to use not-so-common port

clubanderson:
using a not-so-common hostport to avoid port conflicts we have seen when using other projects like SPIFFE.  User: Clint (NetFoundry)

## Related issue(s)
QS instructions will need to be updated for curl command to view web server output

Fixes #

### DIFF
--- a/docs/content/Coding Milestones/PoC2023q1/example1.md
+++ b/docs/content/Coding Milestones/PoC2023q1/example1.md
@@ -446,7 +446,7 @@ spec:
         ports:
         - name: http
           containerPort: 80
-          hostPort: 8094
+          hostPort: 8081
           protocol: TCP
         volumeMounts:
         - name: htdocs
--- a/docs/content/Coding Milestones/PoC2023q1/example1.md
+++ b/docs/content/Coding Milestones/PoC2023q1/example1.md
@@ -46,12 +46,12 @@ nodes:
 - role: control-plane
   extraPortMappings:
   - containerPort: 8081
-    hostPort: 8081
+    hostPort: 8094
 ```
 
 For the guilder cluster, which will get two workloads, create a file
 named `guilder-config.yaml` with the following contents.  The workload
-that uses hostPort 8081 goes in both clusters, while the workload that
+that uses hostPort 8094 goes in both clusters, while the workload that
 uses hostPort 8082 goes only in the guilder cluster.
 
 ```yaml
@@ -446,7 +446,7 @@ spec:
         ports:
         - name: http
           containerPort: 80
-          hostPort: 8081
+          hostPort: 8094
           protocol: TCP
         volumeMounts:
         - name: htdocs
@@ -922,7 +922,7 @@ specialstuff                       speciald                           1/1     1 
 Check that the common workload on the florin cluster is working.
 
 ```console
-$ curl http://localhost:8081
+$ curl http://localhost:8094
 <!DOCTYPE html>
 <html>
   <body>
--- a/docs/content/Coding Milestones/PoC2023q1/example1.md
+++ b/docs/content/Coding Milestones/PoC2023q1/example1.md
@@ -51,7 +51,7 @@ nodes:
 
 For the guilder cluster, which will get two workloads, create a file
 named `guilder-config.yaml` with the following contents.  The workload
-that uses hostPort 8094 goes in both clusters, while the workload that
+that uses hostPort 8081 goes in both clusters, while the workload that
 uses hostPort 8082 goes only in the guilder cluster.
 
 ```yaml
--- a/docs/content/Coding Milestones/PoC2023q3/outline.md
+++ b/docs/content/Coding Milestones/PoC2023q3/outline.md
@@ -1,6 +1,0 @@
----
-title: "Details"
-linkTitle: "Details"
----
-
-# Coming Soon
--- a/docs/content/Coding Milestones/PoC2023q3/outline.md
+++ b/docs/content/Coding Milestones/PoC2023q3/outline.md
@@ -1,0 +1,6 @@
+---
+title: "Details"
+linkTitle: "Details"
+---
+
+# Coming Soon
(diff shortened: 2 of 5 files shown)


MikeSpreitzer:
This choice appears in multiple files.  You can find them all with the following command.

```shell
find * -type f -exec grep -w 8081 \{\} \; -print -exec echo \;
```

clubanderson:
@MikeSpreitzer I think I got the other affected files

clubanderson:
looks like I have to open another PR for port 8082 next

clubanderson:
this PR is ready for another review please

docs/content/Coding Milestones/PoC2023q1/example1.md
```diff
@@ -46,12 +46,12 @@ nodes:
 - role: control-plane
   extraPortMappings:
   - containerPort: 8081
-    hostPort: 8081
+    hostPort: 8094
 ```
 
 For the guilder cluster, which will get two workloads, create a file
 named `guilder-config.yaml` with the following contents.  The workload
-that uses hostPort 8081 goes in both clusters, while the workload that
```
MikeSpreitzer, line 54: The one does not change.  `hostPort` is kube networking jargon, referring to the host of the pod.  That is buried inside the kind cluster, it is not talking about the host of the kind cluster.

docs/content/Getting-Started/quickstart.md
```diff
@@ -294,7 +294,7 @@ spec:
         ports:
         - name: http
           containerPort: 80
-          hostPort: 8081
+          hostPort: 8094
```
MikeSpreitzer, line 297: This again is the kube `hostPort` that is staying 8081.

examples/common-workload.yaml
```diff
@@ -35,7 +35,7 @@ spec:
         ports:
         - name: http
           containerPort: 80
-          hostPort: 8081
+          hostPort: 8094
```
MikeSpreitzer, line 38: The kube `hostPort` stays 8081.

MikeSpreitzer:
Note that in the original there is a coincidence: 8081 is used for TWO _different_ things: the port on the kube node inside the kind cluster, and the port on the host of the florin cluster.

If you mean to change both then that is OK.  If you do that then it might be best to remove the coincidences.

examples/common-workload.yaml
clubanderson: cool - was just going to ask you about this.  thank you

docs/content/Getting-Started/quickstart.md
clubanderson: gotcha - will update - thank you

clubanderson:
@MikeSpreitzer thanks for the corrections - ready for another review

docs/content/Coding Milestones/PoC2023q3/outline.md
```diff
@@ -0,0 +1,6 @@
+---
+title: "Details"
+linkTitle: "Details"
+---
+
+# Coming Soon
```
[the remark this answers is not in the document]
clubanderson: will do - multi-tasking mistake

docs/content/Coding Milestones/PoC2023q1/example1.md
```diff
@@ -446,7 +446,7 @@ spec:
         ports:
         - name: http
           containerPort: 80
-          hostPort: 8081
```
MikeSpreitzer, line 449: In addition to the coincidence on port number, there is a coincidence on field name.  In a kind Cluster object, `hostPort` refers to a port on the host of the kind cluster, not the Kube node inside the cluster.  OTOH, in a Kubernetes Pod spec template, `hostPort` refers to a port on the Kube node hosting that pod.

In short, I think you do not want to make a change here (since you are not changing the `containerPort` in the kind Cluster).

MikeSpreitzer:
Do please get a clarification of whether the issue is with SPIFFE or SPIRE.  It matters because the scope of those two is different.  SPIFFE is an interface that workloads use; SPIRE is a server that does not run in every kube Pod (unless they have changed something).

clubanderson:
it was a SPIFFE getting-started if I recall correctly.  Point is, most getting-started examples just default to 8080, 8081.  If we can avoid conflicts, we should

docs/content/Coding Milestones/PoC2023q1/example1.md
clubanderson: fixed

clubanderson:
alright, one more review @MikeSpreitzer please

clubanderson:
@MikeSpreitzer how do we 'rebase safely' for this PR?

MikeSpreitzer:
I ask because we have three nested layers of network namespace --- three nested places to care about port numbers.  I am trying to determine which one(s) are an issue. The innermost is inside the Apache httpd container, and it is not using SPIFFE or SPIRE, so there will be no issue there unless there is some surprise.  Maybe they are thinking of adding a mutating admission control webhook that injects something inside workload containers; that could cause an issue there.  The middle layer is the Kubernetes node inside the kind cluster, this node is the host of the workload container but a guest of the outermost.  The outermost layer is the host where you issue the `kind` command.

docs/content/Getting-Started/quickstart.md
```diff
@@ -345,7 +345,7 @@ Lastly, let's check that the workload is working in both clusters:
 For `florin`:
 
 ```shell
-curl http://localhost:8081
+curl http://localhost:8094
```
MikeSpreitzer, line 348: Remember that the quickstart tells the user to use the bootstrap strip to download and unpack the most recent release, which is v0.2.2 at the moment, and use the kind Cluster objects in that release's `examples` directory.  So the changes in this PR to those `examples` files will not affect users yet.  So changing only the `curl` in the quickstart will create a broken quickstart.

I think the most likely fix is to expand this PR to change the quickstart so that it does not tell the user to use the kind Cluster objects in the `examples` directory but instead exhibits inline the `Cluster` YAMLs to use.

PR #505, opened 2023-05-31
✨ Start local JSONPath implementation

MikeSpreitzer:
## Summary
This PR introduces a more efficient implementation of JSONPath.  It is more efficient because the updates transform umarshaled data structures rather than strings.  The path language is a subset of JSONPath, extended to accept `.Integer` for backwards compatability.  The subset excludes all expressions and use of parentheses, also whitespace.  The wildcards and non-trivial ranges allow compact statement of many replacements.

## Related issue(s)

Fixes #


MikeSpreitzer:
Successfully tested with the MVI example.

clubanderson:
@MikeSpreitzer can this be contributed back to the JSONPath project upstream?  I assume one exists

MikeSpreitzer:
https://github.com/json-path/JsonPath is in Java rather than go, and I do not see an appetite for expanding beyond Java (did I miss something)?

clubanderson:
I assumed this might be a contribution that could stand on its own two feet outside of our repo. Kubestella/jsonpath  

This might be useful to other communities and projects that are looking for something similar.

MikeSpreitzer:
I am not yet sure whether this is the way to go.  I noted a few possible alternatives in https://github.com/kcp-dev/edge-mc/discussions/453 .

waltforme (commented):
All the components (the lexer, the parser, and the evaluation) work great.

I just did [some tweaks](https://github.com/waltforme/edge-mc/commit/574b3718d1049caee8bfc531f75726dec4d56904) to make terminal cleaner when exercising with `cmd/json-path-replace`. But I guess that cmd will not be used in our deployment. So
/lgtm

BTW I found https://github.com/tidwall/sjson seems to be popular (2.1k stars). I put it here just in case it is useful to us.

PR #523, opened 2023-06-01
🐛 Make `kubestellar init` support syncer updating status

### LINKED ISSUE #477: bug: syncer is not updating reported state in downsynced objects
### Describe the bug

When the syncer attempts to write to the `status` subobject of a downsynced object, of a kind defined in the APIExport of kubernetes from the `root:compute` workspace, in a mailbox workspace, the write is rejected due to an authorization failure.

### Steps To Reproduce

See above.

### Expected Behaviour

For downsynced objects of all kinds, reported state should propagate from the managed cluster to the mailbox workspace.

### Additional Context

This can be fixed by adding RBAC objects to `root:compute`.  @yana1205 , can you please show us exactly what needs to be added?

### LINKED ISSUE #546: bug: kubestellar script has botched control over verbosity
### Describe the bug

The script line that says

```bash
    kubectl apply -f "$SCRIPT_DIR/../config/exports" $verbdir
```

produces the error messages

```console
error: Unexpected args: [&> /dev/null]
```

That is because the parameter expansion of `$verbdir` is not followed by re-parsing for redirection directives, leaving the expansion to be treated as arguments to kubectl.

### Steps To Reproduce

Use `kubestellar` _without_ `-V`.

### Expected Behavior

No footgun.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR extends what `kubestellar init` does, making it also add the RBAC objects to the `root:compute` workspace that enable the syncer to update the `status` sub-objects of the objects whose Kind is defined by the kube APIExport from `root:compute`.

## Related issue(s)

Fixes #477 
Fixes #546

/cc @yana1205 
/cc @dumb0002

### DIFF
--- a/scripts/kubestellar
+++ b/scripts/kubestellar
@@ -50,8 +50,10 @@ while (( $# > 0 )); do
     (--verbose|-V)
         verbosity=1
 	verbdir="";;
+    (-X)
+	set -x;;
     (-h|--help)
-        echo "Usage: $0 [start| stop][--log-folder log_folder] [-V|--verbose]"
+        echo "Usage: $0 [init | start | stop][--log-folder log_folder] [-V|--verbose] [-X]"
         exit 0;;
     (-*)
         echo "$0: unknown flag" >&2 ; exit 1;
@@ -141,7 +143,7 @@ function ensure_espw() {
     else 
         kubectl ws create "$espw_name" --enter
     fi
-    kubectl apply -f $SCRIPT_DIR/../config/exports $verbdir
+    eval kubectl apply -f "$SCRIPT_DIR/../config/exports" $verbdir
     echo "Finished populating the espw with kubestellar apiexports"   
 }
 
--- a/scripts/kubestellar
+++ b/scripts/kubestellar
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Usage: $0 (start or stop | -v)
+# Usage: $0 (init or start or stop | -v)
 
 # Purpose: control whether the controllers are running, setting up the ESPW if necessary.. The following components are created:
 #           (a) 1 kcp workspace: edge service provider workspace (espw)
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
This PR is currently a work-in-progress, although I fear to formally mark it as such for fear of suppressing automated checking.  I am waiting on answers to questions in #477 and testing whether the different YAML used here works.

clubanderson:
bug in gh action is fixed.  unfortunately there is no way to update the commit for the workflow file that is used by the action to use the fix.  You can continue with this PR and ignore that test when it is time to merge.  New PRs will not fail on this action.  Looks like dependabot is unable to access secrets when it does a PR.  This is a known issue.  So I hope that this does not happen again unless dependabot steps in again.

scripts/kubestellar
```diff
@@ -86,7 +86,55 @@ then
...
+metadata:
+  name: compute:authenticated:apiexport:kubernetes:maximal-permission-policy-extended
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: compute:apiexport:kubernetes:maximal-permission-policy-extended
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: User
+  name: kcp-admin
```
yana1205, line 131: Could you please correct the subject to the following one? 
```
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: apis.kcp.io:binding:system:authenticated
```

MikeSpreitzer:
The force-push to c4078ef6 is a rebase plus the change requested in review above.
I think this relatively broad permission is OK because it is used in just one of two tests that both have to Allow, the other one being the regular check in the mailbox workspace.

scripts/kubestellar
MikeSpreitzer: done

MikeSpreitzer:
@clubanderson : the rebase seems to have picked up the fixed GH Action.

MikeSpreitzer:
Tested successfully.

```console
(base) mspreitz@mjs12 edge-mc % kubectl ws .
Current workspace is "root:espw:c4xthhgidi9hnsx9-mb-f4015fdc-c2c6-405c-822f-a5e25755ba04".
(base) mspreitz@mjs12 edge-mc % 
(base) mspreitz@mjs12 edge-mc % kubectl get deployments -A
NAMESPACE      NAME       READY   UP-TO-DATE   AVAILABLE   AGE
commonstuff    commond    1/0     1            1           8m46s
specialstuff   speciald   1/0     1            1           8m43s
```

Note: those zeros are there because `spec.replicas` is unset.

MikeSpreitzer:
BTW, I discovered during testing that the verbosity control in the `kubestellar` script is broken, so I fixed that.  I also added a `-X` option and fixed up the options descriptions.

yana1205:
/lgtm

Thanks!

PR #540, opened 2023-06-01
:sparkles: Create SECURITY_CONTACTS and SECURITY.md

clubanderson:
required for CNCF sandbox and for IBM Dev Advocates office

## Related issue(s)

Fixes #

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,36 @@
+
+## Security Announcements
+
+Join the [kubestellar-security-announce](https://groups.google.com/u/1/g/kubestellar-security-announce) group for emails about security and major API announcements.
+
+## Report a Vulnerability
+
+We're extremely grateful for security researchers and users that report vulnerabilities to the KubeStellar Open Source Community. All reports are thoroughly investigated by a set of community volunteers.
+
+You can also email the private [kubestellar-security-announce@googlegroups.com](mailto:kubestellar-security-announce@googlegroups.com) list with the security details and the details expected for [all KubeStellar bug reports](https://github.com/kcp-dev/edge-mc/blob/main/.github/ISSUE_TEMPLATE/bug_report.yaml).
+
+### When Should I Report a Vulnerability?
+
+- You think you discovered a potential security vulnerability in KubeStellar
+- You are unsure how a vulnerability affects KubeStellar
+- You think you discovered a vulnerability in another project that KubeStellar depends on
+  - For projects with their own vulnerability reporting and disclosure process, please report it directly there
+
+
+### When Should I NOT Report a Vulnerability?
+
+- You need help tuning KubeStellar components for security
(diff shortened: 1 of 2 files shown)


SECURITY.md
```diff
@@ -0,0 +1,36 @@
+
+## Security Announcements
+
+Join the [kubestellar-security-announce](https://groups.google.com/u/1/g/kubestellar-security-announce) group for emails about security and major API announcements.
+
+## Report a Vulnerability
+
+We're extremely grateful for security researchers and users that report vulnerabilities to the KubeStellar Open Source Community. All reports are thoroughly investigated by a set of community volunteers.
+
+You can also email the private [kubestellar-security-announce@googlegroups.com](mailto:kubestellar-security-announce@googlegroups.com) list with the security details and the details expected for [all KubeStellar bug reports](https://github.com/kcp-dev/edge-mc/blob/main/.github/ISSUE_TEMPLATE/bug_report.yaml).
```
MikeSpreitzer, line 10: "announce" sounds wrong for the ingress of problem reports, "announce" suggests this is for egress from the maintainers to the public.

SECURITY_CONTACTS
```diff
@@ -0,0 +1,14 @@
+# Defined below are the security contacts for this repo.
+#
+# They are the contact point for the Product Security Committee to reach out
+# to for triaging and handling of incoming issues.
+#
+# The below names agree to address security concerns if and when they arise.
+#
+# DO NOT REPORT SECURITY VULNERABILITIES DIRECTLY TO THESE NAMES, FOLLOW THE
+# INSTRUCTIONS AT https://kubernetes.io/security/
```
MikeSpreitzer, line 9: Too eager copy-and-paste here.

SECURITY.md
clubanderson: I used the same naming convention as Kubernetes

SECURITY_CONTACTS
clubanderson: Thank you

pdettori: /lgtm

SECURITY.md
MikeSpreitzer: I think not.  Compare https://kubernetes.io/docs/reference/issues-security/security/#security-announcements with https://kubernetes.io/docs/reference/issues-security/security/#report-a-vulnerability

clubanderson: yes - so here is my thinking on that one - there was a whole bunch of structure around a security response committee (SRC) - which I was not ready to commit to at this time.  So I made it simple, for now.  Aspiration is to get there, when we get a bit more resourced.

MikeSpreitzer: Sure.  I got the simplification.  My only point is that the term "announce" is a confusing choice for an ingress point.

PR #548, opened 2023-06-02
:book: changing extended example and quickstart docs

amanroa:
## Summary
Added new instructions for users jumping to extended example from QuickStart.

### DIFF
--- a/docs/content/Coding Milestones/PoC2023q1/example1.md
+++ b/docs/content/Coding Milestones/PoC2023q1/example1.md
@@ -9,6 +9,39 @@ This example is presented in stages.  The controllers involved are
 always maintaining relationships.  This document focuses on changes as
 they appear in this example.
 
+## Continuing from the QuickStart
+
+If you navigated here from the QuickStart, be sure that the environment that was created 
+in the QuickStart is completely removed. 
+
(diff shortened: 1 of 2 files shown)


clubanderson:
lets get another set of eyes on this to do a check before we go live.  I think it's a great improvement, but I have not had time to test myself.  Lets bring some others into the fold here.

clubanderson:
@amanroa @waltforme, I was inspired this weekend to write a gh action that automates the testing of all of our docs.  The mailbox-controller and quickstart work and pass the test.  The placement-translator and example1 (in this PR) do not work.  Any code in a ```shell codeblock is collected and executed inline, whereas information in the ``` {.bash .no-copy} codeblock is used for comments only, and any other plane text in the instructions, and is not executed.  This makes the docs on our site an exact match to our work and are therefore guaranteed to work.  Based on this, rebase your PR and then you can include modular code blocks to do things like removing the environment, downloading and starting kcp and kubestellar, creating the espw, and etc.  This automated testing runs in 3-4 minutes for each of our examples.  You can also use the makefile to run these tests locally which helps with testing before checking in a doc that does not 'execute' - and would never work for our readers.

clubanderson:
closing this as the documents have diverged.  @amanroa will open another PR for each document she revises separate from each other.

PR #561, opened 2023-06-05
WIP: Temp logical clusters support

MikeSpreitzer:
## Summary
This is not a real PR, it is just a thing that makes it easy to identify the work in the `temp-logical-clusters-support` branch.

## Related issue(s)

Fixes #

/cc @ezrasilvera 
/cc @effi-ofer 
/cc @andreyod


pkg/apis/edge/v1alpha1/logical-cluster.go
```diff
@@ -0,0 +1,106 @@
...
+type LogicalClusterSpec struct {
+	// ProviderType is the type of the provider of the cluster.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="cluster is immutable"
+	ProviderType ClusterProviderType `json:"ProviderType"`
+
+	// ProviderName is the name of the provider of the cluster.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="cluster is immutable"
+	ProviderName string `json:"ProviderName"`
+
+	// ClusterName is the name of the logical cluster this workspace is stored under.
```
MikeSpreitzer, line 62: Is "workspace" here a typo?

pkg/apis/edge/v1alpha1/logical-cluster.go
```diff
@@ -0,0 +1,106 @@
...
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ClusterProviderType string
+
+const (
+	KindProviderType ClusterProviderType = "kind"
```
MikeSpreitzer, line 26: I think the set of provider types should not be fixed in this repo.  External repos should be able to supply additional provider types.

pkg/apis/edge/v1alpha1/logical-cluster.go
```diff
@@ -0,0 +1,106 @@
...
+
+// LogicalClusterSpec describes a cluster.
+type LogicalClusterSpec struct {
+	// ProviderType is the type of the provider of the cluster.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="cluster is immutable"
+	ProviderType ClusterProviderType `json:"ProviderType"`
+
+	// ProviderName is the name of the provider of the cluster.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="cluster is immutable"
+	ProviderName string `json:"ProviderName"`
```
MikeSpreitzer, line 60: Surely additional information is generally needed beyond type and name, such as client credentials and network endpoint.  How do those come in?

pkg/logical-cluster-manager/cluster-manager-controller.go
```diff
@@ -0,0 +1,244 @@
...
+	kubeconfig       *string
+	ctx              context.Context
+	clusterclientset edgeclient.Interface
+	clusterInformer  v1alpha1.LogicalClusterInformer
+	queue            workqueue.RateLimitingInterface
+}
+
+func NewController(
+	kubeconfig *string,
+	ctx context.Context,
```
MikeSpreitzer, line 49: If I understand correctly, in the Kube world, if not the go world, there is a strong preference to make the context be the first parameter.

pkg/logical-cluster-manager/cluster-manager-controller.go
```diff
@@ -0,0 +1,244 @@
+package clustermanager
```
MikeSpreitzer, line 1: You will want the usual boilerplate in these files, as soon as they appear in public.

pkg/logical-cluster-manager/cluster-manager-controller.go
```diff
@@ -0,0 +1,244 @@
...
+
```
MikeSpreitzer, line 2: I think it would be helpful to have a brief comment here outlinine what this controller does.

pkg/logical-cluster-manager/cluster-manager-controller.go
```diff
@@ -0,0 +1,244 @@
...
+const (
+	opAdd    op = "Add"
+	opUpdate op = "Update"
+	opDelete op = "Delete"
+)
+
+type queueItem struct {
+	op  op
+	key any
+}
```
MikeSpreitzer, line 37: This is a very unusual controller design.  The usual controller design puts into the work queue only object references.  And I do not mean pointers, I mean namespace (if applicable) and name.  That was the workqueue guarantees that only one worker is working on a given object at a time.  If you put more fine grained information in the work queue entries then (a) it may not be the latest by the time it is dequeued and processed and (b) it can cause concurrent work to be done on the same object.

pkg/logical-cluster-manager/cluster-manager-controller.go
```diff
@@ -0,0 +1,244 @@
...
+	var err error
+
+	newClusterConfig := key.(*v1alpha1apis.LogicalCluster)
+	clusterName := newClusterConfig.Spec.ClusterName
+
+	provider := clusterproviderclient.GetProviderClient(newClusterConfig.Spec.ProviderType, newClusterConfig.Spec.ProviderName)
+
+	// Update status to NotReady
+	newClusterConfig.Status.Phase = v1alpha1apis.LogicalClusterPhaseNotReady
+	_, err = c.clusterclientset.EdgeV1alpha1().LogicalClusters().Update(ctx, newClusterConfig, v1.UpdateOptions{})
```
MikeSpreitzer, line 160: What does a client that queries the status before this line runs get?  How/why is that semantically different from LogicalClusterPhaseNotReady?

MikeSpreitzer:
A design outline would be helpful.

I see that for the `kind` type of provider, the "provider name" is not used.  Is there a vision for how it would be used for other types of provicers?

pkg/apis/edge/v1alpha1/logical-cluster.go
MikeSpreitzer: I also think that we should not be trying to build a big binary that supports many types of providers.  Could we instead say that each provider type comes with its own controller image?

MikeSpreitzer:
I see that in sig-cluster-lifecycle, they acknowledge two styles of management: those that are created and deleted through their API, and those that are primarily managed through another API and visible through their API.  I think this is a sensible approach.

pkg/logical-cluster-manager/cluster-manager-controller.go
```diff
@@ -0,0 +1,244 @@
...
+		return false
+	}
+
+	// No matter what, tell the queue we're done with this key, to unblock
+	// other workers.
+	defer c.queue.Done(i)
+
+	//ES: not sure we should treat all errors in the same way (queue back)
+	if err := c.process(ctx, i.(queueItem)); err != nil {
+		runtime.HandleError(fmt.Errorf("%q controller didn't sync, err: %w", controllerName, err))
```
MikeSpreitzer, line 141: A retry is not an un-handled error.

Also: https://kubernetes.slack.com/archives/C0EG7JC6T/p1686033859701419

pkg/mcclient/mcclient.go
```diff
@@ -0,0 +1,158 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
+	log "k8s.io/klog/v2"
+
+	"github.com/kcp-dev/edge-mc/pkg/apis/edge/v1alpha1"
+	ksclientset "github.com/kcp-dev/edge-mc/pkg/client/clientset/versioned"
+	ksinformers "github.com/kcp-dev/edge-mc/pkg/client/informers/externalversions"
+	mcclient "github.com/kcp-dev/edge-mc/pkg/mcclient/clientset"
```
MikeSpreitzer, line 22: While this is well-defined in golang, in my opinion it is confusing to define an import package to have the same name as appears in the `package` statement.

pkg/mcclient/mcclient.go
```diff
@@ -0,0 +1,158 @@
...
+	clientsets map[string]*mcclient.Clientset
+	lock       sync.Mutex
+}
+
+func (mcc *multiClusterClient) Cluster(name string) mcclient.Interface {
+	mcc.lock.Lock()
+	defer mcc.lock.Unlock()
+	clientset, ok := mcc.clientsets[name]
+	if !ok {
+		//TODO change to ClusterOrDie and panic? return an error?
```
MikeSpreitzer, line 46: Return an error here.  Handling should be the client's decision.

pkg/mcclient/mcclient.go
```diff
@@ -0,0 +1,158 @@
...
+		clusterLW[cluster] = listwatch.ClusterListWatch(config, gv, resource, namespace, optionsModifier)
+	}
+	return listwatch.NewCrossClusterListerWatcher(clusterLW)
+}
+
+var client *multiClusterClient
+
+// NewMultiCluster creates new multi-cluster client and starts collecting cluster configs
+func NewMultiCluster(ctx context.Context, managerConfig *rest.Config) (KubestellarClusterInterface, error) {
+	if client != nil {
```
MikeSpreitzer, line 78: You probably want to make this to be safe for concurrent use.

pkg/mcclient/mcclient.go
```diff
@@ -0,0 +1,158 @@
...
+	clusterInformer := clusterInformerFactory.Edge().V1alpha1().LogicalClusters().Informer()
+
+	clusterInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			clusterInfo, ok := obj.(*v1alpha1.LogicalCluster)
+			if !ok {
+				log.Error("unexpected object type. expected LogicalCluster")
+				return
+			}
+			go mcc.handleAdd(clusterInfo)
```
MikeSpreitzer, line 106: Because this directly uses goroutines rather than running the object references through a work queue, this can cause concurrent goroutines working on the same cluster.

Also, by the time the forked handleAdd starts working on a given cluster, it might already be deleted.  Similarly, by the time a forked handleDelete starts working on the cluster, it might have been added again.

pkg/mcclient/listwatch/listwatch.go
```diff
@@ -0,0 +1,108 @@
...
+		log.Infof("--- Whatch for cluster: %s", clusterName)
+		clusterWatch, err := lwForCluster.Watch(options)
+		if err != nil {
+			return nil, fmt.Errorf("Watch for cluster %s failed: %w", clusterName, err)
+		}
+
+		go func(clusterName string) {
+			for {
+				select {
+				case <-context.Background().Done(): //TODO: pass the ctx as param?
```
MikeSpreitzer, line 42: (1) https://github.com/kubernetes/client-go/blob/v0.27.2/tools/cache/listwatch.go#L39 takes no context, so I think you want to follow suit.

(2) `context.Background()` is never done.

(3) You might want an early-out for other reasons, such as https://github.com/kcp-dev/edge-mc/pull/561/files#r1220068089

pkg/mcclient/listwatch/listwatch.go
```diff
@@ -0,0 +1,108 @@
...
+	return clw
+}
+
+func (clw *CrossClusterListerWatcher) Watch(options metav1.ListOptions) (watch.Interface, error) {
+	combinedCh := make(chan watch.Event)
+	for clusterName, lwForCluster := range clw.clusterListWatch {
+		log.Infof("--- Whatch for cluster: %s", clusterName)
+		clusterWatch, err := lwForCluster.Watch(options)
+		if err != nil {
+			return nil, fmt.Errorf("Watch for cluster %s failed: %w", clusterName, err)
```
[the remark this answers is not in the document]
ezrasilvera: this is just an initial place holder / overall flow for the multi-cluster operations (List() & Watch()) , we are planning to refactor the whole error handling of this 
We also plan to update the Watch() and not use only the clusters that exist  when Watch() was called.

effi-ofer:
General comment - this branch is still a bit of a work in progress and hence the separate branch.  We delivered early to be agile and so everybody can have a look at what we are working on but please don't expect perfection just yet :)  @MikeSpreitzer thanks for your the insightful comments.  We'll be going over them in the next day or two (or spill into next week if we are slow) and will answer each in turn as we get to it.

pkg/apis/edge/v1alpha1/logical-cluster.go
effi-ofer: We are creating a separate resource for the provider.  The first drop is in https://github.com/kcp-dev/edge-mc/pull/566 and the full provider information is yet to be finalized.

pkg/mcclient/mcclient.go
andreyod: If we return an error here then something like `mcclient.Cluster("cluster1").KS().EdgeV1alpha1()...` will not compile and the client will have to first check the error from `mcclient.Cluster`
Also the current clustered clientset code [panics on error](https://github.com/kcp-dev/edge-mc/blob/main/pkg/client/clientset/versioned/cluster/clientset.go#L74)

pkg/mcclient/mcclient.go
andreyod: Done

PR #563, opened 2023-06-05
🌱 Wait for a few seconds before getting sps

waltforme:
## Summary
This PR adds some waits after creation of the edgeplacements before checking the singleplacementslices, in the scheduler doc, trying to see whether the doc test CI feature is good with it.


waltforme:
It turns out there were asynchrony problems when we let the CI run the commands in docs which are initially intended for a human to read and exercise. A human can easily tell whether it is the right moment to execute the next command but it is harder for CI to do so. There will be some balancing play between 'always ensure the correctness of the doc' and 'make the docs clean for a human' - but that is out of the scope of this PR. This PR just makes the 'run doc shell' work first.

PR #567, opened 2023-06-07
✨ Start drafting interface for summarization

MikeSpreitzer:
## Summary
This PR contributes a draft design for the API for summarization.

## Related issue(s)

Fixes #

/cc @yana1205 
/cc @yuji-watanabe-jp

### DIFF
--- a/pkg/apis/edge/v1alpha1/edge-placement.go
+++ b/pkg/apis/edge/v1alpha1/edge-placement.go
@@ -90,6 +90,11 @@ type EdgePlacementSpec struct {
 	// An object matches `upsync` if and only if it matches at least one member of `upsync`.
 	// +optional
 	Upsync []UpsyncSet `json:"upsync,omitempty"`
+
+	// SummarizerName, if non-empty, holds the name of a Summarizer object that prescribes
+	// how to create corresponding summary objects.
+	// +optional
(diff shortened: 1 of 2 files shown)


yana1205:
Thanks for drafting the interface concretely.

I am still going through the spec. If a summarizer is defined like
```
  groupers:
  - name: grouper1
    groupBy:
    - name: replicas
      def: $.spec.replicas
    aggregators:
    - name: TotalReplicas
      type: SUM
      subject: $.spec.replicas
``

The summary status look like if I have 2 deployments whose replicas are 3 individually.
```
summaryStatus:
  groupings:
  - name: grouper1
    groups:
    - groupedBy: 
      - name: replicas
        value: 3
    - aggregates:
      - name: replicas
        value: 6
```
Is my understanding correct?

If it's correct, I have another question about how a user can set GVR. For example in this case, I would aggregate apps/v1/deployments. Where can I define it?

MikeSpreitzer:
@yana1205 : yes, your example matches what I am thinking.

I am not sure I understand your question about apps/v1/deployments.  Do you mean to ask how to aggregate reported state from downsynced objects?  If so then the answer is that this PR adds a field to EdgePlacementSpec named "SummarizerName"; put into that field the name of a Summarizer object in the same namespace that defines the summarization that you want.

PR #571, opened 2023-06-07
:sparkles: Update README.md testing ci

clubanderson:
## Related issue(s)

Fixes #


clubanderson:
@ezrasilvera @MikeSpreitzer @pdettori @waltforme @fileppb @dumb0002 @francostellari @KPRoche here is an example of our new PROW output.  The kind folks at kcp have allowed us to share their environment with us.

PR #591, opened 2023-06-11
:sparkles:Add a work queue for logicalclusters informer.

andreyod:
## Summary
Add a work queue for multi-cluster client logicalclusters informer.
 
## Related issue(s)

Fixes #


pkg/mcclient/controller.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (c *controller) handleAdd(cluster interface{}) {
+	clusterInfo, ok := cluster.(*v1alpha1.LogicalCluster)
+	if !ok {
+		runtime.HandleError(errors.New("unexpected object type. expected LogicalCluster"))
+		return
+	}
+	// only clusters in ready state are cached
+	if clusterInfo.Status.Phase != v1alpha1.LogicalClusterPhaseReady {
+		c.handleDelete(clusterInfo.Name)
```
ezrasilvera, line 150: Why delete ?  Don't we need to just wait till Ready, e.g., ignore this and wait for additional status change

pkg/mcclient/mcclient.go
```diff
@@ -21,46 +21,47 @@ package mcclient
...
-	// CrossClusterListWatch returns cross-cluster ListWatch
-	CrossClusterListWatch(gv schema.GroupVersion, resource string, namespace string, fieldSelector fields.Selector) *listwatch.CrossClusterListerWatcher
 }
 
 type multiClusterClient struct {
-	configs    map[string]*rest.Config
-	clientsets map[string]*mcclientset.Clientset
-	lock       sync.Mutex
+	ctx              context.Context
+	configs          map[string]*rest.Config
```
ezrasilvera, line 48: Can't we get the config from the clientset in case someone needs it?

pkg/mcclient/mcclient.go
```diff
@@ -21,46 +21,47 @@ package mcclient
...
 
 func (mcc *multiClusterClient) Cluster(name string) mcclientset.Interface {
 	mcc.lock.Lock()
 	defer mcc.lock.Unlock()
 	clientset, ok := mcc.clientsets[name]
 	if !ok {
-		//TODO change to ClusterOrDie and panic? return an error?
-		panic("invalid cluster name")
+		// try to get LogicalCluster from API server
+		clientset, err := mcc.getFromServer(name)
```
ezrasilvera, line 60: Why don't we cache the config & cs when doing getFromServer ?  If we missed somehow the event you will come here every time someone use this cluster, right ?

pkg/mcclient/controller.go
andreyod: It only deletes from the cache maps. If `Status.Phase` is not ready then the cluster is not accessible and config/cs is not valid. Client code which try to use this config/cs will fail. We will get another event when the cluster is Ready and then we cache it.

pkg/mcclient/mcclient.go
andreyod: We do cache it in `getFromServer ` function if it exists and valid.

pkg/mcclient/mcclient.go
andreyod: I don't see an easy way to do this.

pkg/mcclient/controller.go
ezrasilvera: Thanks. I forgot you use the same function for Update as well. Can you please add a comment (copy from your reply :-) ) ?

pkg/mcclient/mcclient.go
ezrasilvera: right, the gitdiff confused me with  the ondelete change :(

pkg/mcclient/mcclient.go
```diff
@@ -21,46 +21,47 @@ package mcclient
...
+		if err == nil {
+			return clientset
```
ezrasilvera, line 62: nit - if you negate the condition and panik in it, you can remove the return clientset from here

pkg/mcclient/mcclient.go
```diff
@@ -21,46 +21,47 @@ package mcclient
...
-	lock       sync.Mutex
+	ctx              context.Context
+	configs          map[string]*rest.Config
+	clientsets       map[string]*mcclientset.Clientset
+	managerClientset *ksclientset.Clientset
+	lock             sync.Mutex
 }
 
 func (mcc *multiClusterClient) Cluster(name string) mcclientset.Interface {
 	mcc.lock.Lock()
```
ezrasilvera, line 55: I'm little not at ease calling getFromServer from within a lock.  Also it's too wide. 
If we change indeed there might be races between the explicit Get and the events. 
Can you please at minimal add a comment on getFromServer saying it must be called within an mcc lock ?

pkg/mcclient/controller.go
andreyod: Done.

pkg/mcclient/mcclient.go
andreyod: Changed.

pkg/mcclient/mcclient.go
andreyod: Comment added.

PR #590, opened 2023-06-11
:bug: Rename cliplugin directory from kcp-edge to kubestellar

yana1205:
## Summary

I needed to replace directory name "kcp-edge" in pkg/cliplugins. Currently the cliplugins contain only syncer related command. This change does no impact on the other programs than syncer. I also verified that syncer is working with this change (ut, e2e, doc e2e passed).

The same PR will be created for release-0.2 along with #589 soon or later.

## Related issue(s)

Fixes #584

PR #589, opened 2023-06-10
:bug: Rename kcp-edge-syncer and edge-syncer to kubestellar-syncer

yana1205:
## Summary

Rename kcp-edge-syncer and edge-syncer to kubestellar-syncer in all docs and syncer related codes.
I confirmed no regression.

#### Verification
UT passed.
```
make test-syncer
?       github.com/kcp-dev/edge-mc/pkg/syncer   [no test files]
?       github.com/kcp-dev/edge-mc/pkg/syncer/clientfactory     [no test files]
ok      github.com/kcp-dev/edge-mc/pkg/cliplugins/kcp-edge/syncer-gen   (cached)
?       github.com/kcp-dev/edge-mc/pkg/syncer/shared    [no test files]
?       github.com/kcp-dev/edge-mc/pkg/syncer/syncers   [no test files]
ok      github.com/kcp-dev/edge-mc/pkg/syncer/controller        14.336s
```
E2E passed
```
make IGNORE_GO_VERSION=1 e2e-test-kubestellar-syncer
ok      github.com/kcp-dev/edge-mc/test/e2e/kubestellar-syncer  151.148s
```

Syncer command now displays kubestellar-syncer.
```
sed -i.bak 's|repo_url: https://github.com/kcp-dev/edge-mc|repo_url: https://github.com/yana1205/edge-mc|g' ./docs/mkdocs.yml
make IGNORE_GO_VERSION=1 MANIFEST="'docs/content/common-subs/pre-req.md','docs/content/Coding Milestones/PoC2023q1/kubestellar-syncer.md'" docs-ecutable
```
```
...
Current workspace is "root:espw:2di2trhpcb671ydr-mb-0edca87f-3ae3-46c5-8121-59dd0aeeb16d".
Creating service account "kubestellar-syncer-guilder-1h45r386"
Creating cluster role "kubestellar-syncer-guilder-1h45r386" to give service account "kubestellar-syncer-guilder-1h45r386"

 1. write and sync access to the synctarget "kubestellar-syncer-guilder-1h45r386"
 2. write access to apiresourceimports.

Creating or updating cluster role binding "kubestellar-syncer-guilder-1h45r386" to bind service account "kubestellar-syncer-guilder-1h45r386" to cluster role "kubestellar-syncer-guilder-1h45r386".
Wrote physical cluster manifest to guilder-syncer.yaml for namespace "kubestellar-syncer-guilder-1h45r386". Use

[...]

PR #598, opened 2023-06-13
:bug: (release-0.2) Rename kcp-edge-syncer and edge-syncer to kubestellar-syncer

yana1205:
## Summary

This is the same PRs as #589, #590, and #597 for release-0.2.
Confirmed no regression.

Findings: Perhaps, we need to remove line 22 in [hack/verify-go-versions.sh](https://github.com/kcp-dev/edge-mc/blob/release-0.2/hack/verify-go-versions.sh#L22) since [the latest version](https://github.com/kcp-dev/edge-mc/blob/main/hack/verify-go-versions.sh#L22) in the main branch already doesn't have line 22.

## Related issue(s)

Fixes #


clubanderson (commented):
/lgtm
/approve

PR #607, opened 2023-06-14
:sparkles: API resources for logical clusters

ezrasilvera:
## Summary
This PR add the API resources needed for the logical cluster work. It is pushed as a separate PR because this is the only code that need to be merged with existing KubeStellar code (due to the code generation). 
For the reviewer: Please review only the first commit, this commit is the actual code, the second commit includes the generated code.


pkg/apis/logicalcluster/v1alpha1/cluster-provider-desc.go
```diff
@@ -0,0 +1,86 @@
...
+	// ProviderType is the type of the provider of the cluster.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="cluster is immutable"
+	ProviderType ClusterProviderType `json:"ProviderType"`
+
+	// TODO: this should be stored as a secret!
+	// Config is the provider config
+	// +kubebuilder
+	Config string `json:"Config,omitempty"`
+
+	// ClusterPrefixForDiscovery contains the prefix used during cluster discovery.
```
MikeSpreitzer, line 65: How exactly does the provider use this prefix?

Is this properly part of the generic provider config?  Couldn't different providers have different ways of doing (whatever) besides prefixing?

pkg/apis/logicalcluster/v1alpha1/cluster-provider-desc.go
```diff
@@ -0,0 +1,86 @@
...
+
+// ClusterProviderDescSpec describes a cluster.
+// TODO: We are currently only listing the type and config.
+// There will be additional fields in the future.
+type ClusterProviderDescSpec struct {
+	// ProviderType is the type of the provider of the cluster.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="cluster is immutable"
+	ProviderType ClusterProviderType `json:"ProviderType"`
+
+	// TODO: this should be stored as a secret!
```
MikeSpreitzer, line 60: Should the whole config be in a Secret or can some of it be in the open?

pkg/apis/logicalcluster/v1alpha1/logical-cluster.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// LogicalCluster represents a cluster.
+//
+// +crd
+// +genclient
+// +genclient:nonNamespaced
```
MikeSpreitzer, line 27: I am not sure this is the right choice.
I am not sure it is not.

pkg/apis/logicalcluster/v1alpha1/logical-cluster.go
```diff
@@ -0,0 +1,98 @@
...
+	// `status` describes the status of the cluster object.
+	// +optional
+	Status LogicalClusterStatus `json:"status"`
+}
+
+// LogicalClusterSpec describes a cluster.
+type LogicalClusterSpec struct {
+	// ClusterProviderDesc is a reference to a ClusterProviderDesc resource
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="ProviderName is immutable"
+	ClusterProviderDesc string `json:"ClusterProviderDesc"`
```
MikeSpreitzer, line 50: In k/k, if I understand correctly, there is a preference for a field name that holds a reference to an object to include in the field's name an indication of this: the field name would end in "Name" if the reference is just a name, something like "Reference" if the reference is more elaborate.

pkg/apis/logicalcluster/v1alpha1/logical-cluster.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// ClusterName is the name of the logical cluster this workspace is stored under.
```
MikeSpreitzer, line 52: What, here, is a "workspace"?
How is the name here different from the `ObjectMeta.Name` ?

pkg/apis/logicalcluster/v1alpha1/logical-cluster.go
```diff
@@ -0,0 +1,98 @@
...
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="clusterName is immutable"
+	ClusterName string `json:"ClusterName,omitempty"`
+
+	// Managed identifies whether a cluster is managed (true) or unmanaged (false).
+	// Currently this is immutable.
+	// A logical cluster can be created through the ClusterManager (managed) or
+	// discovered/imported (unmanaged).
```
MikeSpreitzer, line 60: When the logical cluster controller is restarted, and it learns about clusters that already exist (because it created them in a previous run), does this count as "discovery"?

pkg/apis/logicalcluster/v1alpha1/cluster-provider-desc.go
ezrasilvera: The prefix just try to limit the scope of the discovered clusters. It allows the provider a simple way to distinguish the clusters that he wants to be discovered and other clusters that might be seen through the same provider endpoint but shouldn't be discovered/imported. It's not mandatory and if empty all clusters will be discovered. In container runtime PoC they had something similar (although hard coded) in which they only discovered `fleet-xxx`  clusters

pkg/apis/logicalcluster/v1alpha1/logical-cluster.go
ezrasilvera: I think it's mainly automatic writing :-)  I don't have strong opinion here - what do you think?

pkg/apis/logicalcluster/v1alpha1/logical-cluster.go
ezrasilvera: typo. This is the logical cluster name. Will fix the comment

ezrasilvera: Fixed. https://github.com/kcp-dev/edge-mc/pull/607/commits/febc2d00536c20cf80e0c173a586f26d5bec0f18

pkg/apis/logicalcluster/v1alpha1/logical-cluster.go
ezrasilvera: Do you have any suggestion for changing the comment ?

pkg/apis/logicalcluster/v1alpha1/logical-cluster.go
MikeSpreitzer: I think that in general a resource should NOT be namespaced iff there is a specific reason why it logically applies to the whole of the cluster wherein it appears.  Otherwise, why take away the ability of users to organize the resource into namespaces?

pkg/apis/logicalcluster/v1alpha1/cluster-provider-desc.go
MikeSpreitzer: Yes, I assumed that different providers will need different info.  For example, a provider of kind clusters on the same machine needs no info at all.  Indeed some providers will likely need information that needs to be kept private.  So the simplest adequate solution would be to keep all of the needed information (if there is any at all), for a given provider, in one Secret object.  A possible further step would be to remove from that Secret object information that does not need to be kept private and instead hold it in a ConfigMap or something.

pkg/apis/logicalcluster/v1alpha1/logical-cluster.go
MikeSpreitzer: I am not sure right now.

MikeSpreitzer:
The logical cluster abstraction layer should reside in one or more go modules of its own, distinct from the go module for the multi-cluster configuration management module.  Remember that we can have multiple go modules in one github repository.

pkg/apis/logicalcluster/v1alpha1/cluster-provider-desc.go
ezrasilvera: Yes, This is stage one in which the provider behavior is controlled through definitions on the cluster-manager side, so we came up with some common behavior. For phase 2,  we plan to support plugin and configuration of external providers not through the cluster manager. The provider client is implemented by the provider and can implement whatever technique they want for selection/filtering of the discovered clusters.

pkg/apis/logicalcluster/v1alpha1/logical-cluster.go
ezrasilvera: Changed logicalcluster to be namespaced. We plan to hot logicalcluster objects in a NS corresponding to provider instance.

ezrasilvera: pushed in new commiy

pkg/apis/logicalcluster/v1alpha1/cluster-provider-desc.go
ezrasilvera: @MikeSpreitzer I opened a new issue to address logical-clusters stage 2.  The above will be part of that.

pkg/apis/logicalcluster/v1alpha1/cluster-provider-desc.go
ezrasilvera: As mentioned above, in stage 2, we will switch to REST based providers where the client connects through regular TLS based connection and the type of info needed to connect to a provider will be identical for all providers (e.g., crt, key, IP)

pkg/apis/logicalcluster/v1alpha1/logical-cluster.go
ezrasilvera: I will add a section to the design doc and issue to reflect the need for better description

ezrasilvera:
> The logical cluster abstraction layer should reside in one or more go modules of its own, distinct from the go module for the multi-cluster configuration management module. Remember that we can have multiple go modules in one github repository.

The plan is to use different repository in KubeStellar. At this point we don't want to spend time on separating to modules (as we would need to change the generation scripts as well). We will probably do it once this repo is moved to KubeSTellar

MikeSpreitzer:
I am not sure that a new repo will happen soon.  In any case, the go module separation work that would be done in the same repo is mostly the same as the go module separation work that would come with a new repo, the only difference is the regular difference in package prefixes.

pkg/apis/logicalcluster/v1alpha1/logical-cluster.go
```diff
@@ -0,0 +1,97 @@
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="ProviderName is immutable"
+	ClusterProviderDescName string `json:"ClusterProviderDescName"`
+
+	// ClusterName is the name of the logical cluster created by the provider.
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="clusterName is immutable"
+	ClusterName string `json:"ClusterName,omitempty"`
+
+	// Managed identifies whether a cluster is managed (true) or unmanaged (false).
+	// Currently this is immutable.
```
MikeSpreitzer, line 57: Saying that something is "immutable" begs questions about finalizers.  Without finalizers, a client that wants to change something could simply delete the object and create a replacement without any update restrictions.

MikeSpreitzer (approved):
There is more work to do but we can do it in follow-up PRs.
/lgtm
/approve

Issue #522, opened 2023-06-01
feature: support the other containerized workload resources


MikeSpreitzer:
### Feature Description

As you can see in https://github.com/kcp-dev/kcp/tree/v0.11.0/config/rootcompute/kube-1.24 , currently kcp's APIExport of Kubernetes resources from the `root:compute` workspace only has four resources (kinds of objects): deployments, pods, ingresses, and services.  KubeStellar should also support use of the other containerized workload resources from Kubernetes.

### Proposed Solution

Do like kcp did for the first four, but for all the others.  I think that involved using https://github.com/kcp-dev/kcp/tree/v0.11.0/cmd/crd-puller .

### Want to contribute?

- [x] I would like to work on this issue.

MikeSpreitzer:
Considering https://github.com/kcp-dev/edge-mc/discussions/582 , we might want to segregate these resources into two groups: namespaced and not.

clubanderson:
pushing back 1 sprint.  @MikeSpreitzer is this referenced by the q4 roadmap?

clubanderson:
namespaces API is supported by KubeStellar today.  Cluster-scoped (non-namespaced) Is under investigation.

clubanderson:
pending issue https://github.com/kube-bind/kube-bind/pull/175

MikeSpreitzer:
We added the others.
And this is moot now.

PR #612, opened 2023-06-15
🐛 Make root:compute export the rest of namespaced kube API

MikeSpreitzer:
## Summary
This PR does the following.
1. Makes `kubestellar init` extend the `root:compute` workspace with additional APIExports that cover the namespaced resources not already exported by kcp-dev/kcp and additional RBAC contents to make those additional APIExports usable.
3. Makes `kubectl kubestellar ensure wmw --with-kube` install corresponding APIBinding objects.

This PR also updates example1 to exercise the new functionality, using a ReplicaSet object instead of a Deployment object in the common workload.  This PR also updates example1 to use `kubestellar init` because that now does a lot more than just create one workspace and do one `kubectl apply`.

Not exporting the cluster-scoped part yet, not sure about ability to support it in the future.

The CRDs and APIExports were produced as follows.

The resources were pulled from a `kind` cluster running Kubernetes release 1.25.2.

These were produced by the following bashery.

The following function converts a `kubectl api-resources` listing into a listing of arguments to the kcp crd-puller.

```bash
function rejigger() {
    if [[ $# -eq 4 ]]
    then gv="$2"
    else gv="$3"
    fi

    case "$gv" in
	(*/*) group=.$(echo "$gv" | cut -f1 -d/) ;;
	(*)   group=""
    esac

    echo "${1}$group"
}
```

With `kubectl` configured to manipulate a kcp workspace, the following command captures the listing of resources built into that kcp workspace.

```bash
kubectl api-resources | grep -v APIVERSION | while read line; do rejigger $line; done > /tmp/kcp-rgs.txt
```

With `kubectl` configured to manipulate a kind cluster, the following commands capture the resource listing split into namespaced and cluster-scoped.

```bash
kubectl api-resources | grep -v APIVERSION | grep -w true | while read line; do rejigger $line; done > /tmp/kind-ns-rgs.txt
kubectl api-resources | grep -v APIVERSION | grep -w false | while read line; do rejigger $line; done > /tmp/kind-cs-rgs.txt
```

With CWD=config/kube/exports/namespaced,

```bash
crd-puller --kubeconfig $KUBECONFIG $(grep -v -f /tmp/kcp-rgs.txt /tmp/kind-ns-rgs.txt)
```

With CWD=config/kube/exports/cluster-scoped,

```bash
crd-puller --kubeconfig $KUBECONFIG $(grep -v -f /tmp/kcp-rgs.txt /tmp/kind-cs-rgs.txt)
```

I manually deleted the four CRDs from https://github.com/kcp-dev/kcp/tree/v0.11.0/config/rootcompute/kube-1.24 .

Sadly, https://github.com/kubernetes/kubernetes/issues/118698 is a thing.
So I manually hacked the CRD for jobs.

Sadly, the filenames produced by the crd-puller are not loved by apigen.  The following function renames one file as needed.

```bash
function fixname() {
    rg=${1%%.yaml}
    case $rg in
	(*.*)
	    g=$(echo $rg | cut -d. -f2-)
	    r=$(echo $rg | cut -d. -f1);;
	(*)
	    g=core.k8s.io
	    r=$rg;;
    esac
    mv ${rg}.yaml ${g}_${r}.yaml
}
```

In each of those CRD directories,

```bash
for fn in *.yaml; do fixname $fn; done
```

Penultimately, with CWD=config/kube,

```bash
../../hack/tools/apigen --input-dir crds/namespaced --output-dir exports/namespaced
../../hack/tools/apigen --input-dir crds/cluster-scoped --output-dir exports/cluster-scoped
```
Finally, https://github.com/kubernetes/enhancements/pull/1111 applies
to APIExport/APIBinding as well as to CRDs.  And the CRD puller does
not know anything about this (not that it would help?).  I manually
hacked the namespaced APIResource files that needed it to have an
`api-approved.kubernetes.io` annotation.  It turns out that the
checking in the apiserver only requires that the annotation's value
parse as a URL (any URL will do).

## Related issue(s)

This addresses part of #522 --- the part for namespaced resources, but not cluster-scoped ones.

/cc @waltforme 
/cc @ezrasilvera

### DIFF
--- a/config/kube/crds/cluster-scoped/core.k8s.io_persistentvolumes.yaml
+++ b/config/kube/crds/cluster-scoped/core.k8s.io_persistentvolumes.yaml
@@ -1,0 +1,1093 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: persistentvolumes.core
+spec:
+  conversion:
+    strategy: None
+  group: ""
+  names:
+    kind: PersistentVolume
+    listKind: PersistentVolumeList
+    plural: persistentvolumes
+    shortNames:
+    - pv
+    singular: persistentvolume
+  scope: Cluster
+  versions:
+  - name: v1
+    schema:
+      openAPIV3Schema:
+        description: 'PersistentVolume (PV) is a storage resource provisioned by an
+          administrator. It is analogous to a node. More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes'
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: 'spec defines a specification of a persistent volume owned
+              by the cluster. Provisioned by an administrator. More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#persistent-volumes'
+            properties:
+              accessModes:
+                description: 'accessModes contains all ways the volume can be mounted.
+                  More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#access-modes'
+                items:
+                  type: string
+                type: array
+              awsElasticBlockStore:
+                description: 'awsElasticBlockStore represents an AWS Disk resource
+                  that is attached to a kubelet''s host machine and then exposed to
+                  the pod. More info: https://kubernetes.io/docs/concepts/storage/volumes#awselasticblockstore'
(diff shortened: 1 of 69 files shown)


ezrasilvera:
@MikeSpreitzer Is the production of the CRDs a one time thing that is done suing the manual instructions you listed in the PR description?   Don't we need to automate this and add that script as part of the PR as well ?

MikeSpreitzer:
I initially planned to script all the stuff done to create the APIExports.  First, on the principle that source code control should capture all the inputs; the vision is that there would be a script and/or Makefile target that gets the job done and some documentation would describe when, why, and how to use that.  Second, note that this extraction was done for one particular release of Kubernetes.  We probably need to get to the point of supporting an evolving set of Kubernetes releases.

If you read the description here of what was done, you will see that it would be non-trivial to script.  I ran into some problems that I solved with manual editing.  This could be scripted, and I even think should be scripted.  But this PR does not include that.

MikeSpreitzer:
This PR has a deficiency that should be addressed in follow-up.  It is a consequence of a corner case in kcp's APIExport/APIBinding.  Before the first object in it is created, an APIExport view behaves oddly: attempts to list and watch it are rebuffed with authentication errors (the correct behavior would be an empty list or watch response), even though `kubectl api-resources` says that those resources exist.  In KubeStellar there is no special response or avoidance of this, so the log is continually spammed with logging these error responses.

clubanderson:
please rebase this PR to get the latest workflow checks

MikeSpreitzer:
The force-push to b75c1e3d adds a newline to the commit comment, trying to prod CI to run the docs executions.

PR #626, opened 2023-06-16
:bug: Fix to set correct UID and resource version when syncing updated objects and do not downsync upsynced objects

### LINKED ISSUE #625: bug: upsyncying issue
### Describe the bug

I'm having trouble with upsyncing objects from edge cluster to mailbox workspace. In edgeplacement, I listed the name of CRs that are needed to be upsynced from edge cluster. I observed that after applying the edgeplacement,  CRs are appropriately created in mailbox workspace; but they are not updated if the CRs  in edge cluster change later.  I think these CRs  are just created in mailbox and will not be updated automatically over time.  

### Steps To Reproduce

My goal is to upsync greencluster objects from edge cluster to mailbox cluster
- Here is the CRD for greencluster:
```
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
 name: greenclusters.stable.example.com
spec:
 group: stable.example.com
 versions:
   - name: v1
     served: true
     storage: true
     schema:
       openAPIV3Schema:
         type: object
         properties:
           spec:
             type: object
             properties:
               cluster:
                 type: string
               cpu:
                 type: integer
               carbon:
                 type: number
               target:
                 type: string
               futureCarbon:
                 type: array
                 items:
                   type: object
                   properties:
                     timeStamp:
                       type: number
                     carbon:
                       type: number     
              
 scope: Cluster
 names:
   plural: greenclusters
   singular: greencluster
   kind: GreenCluster
   shortNames:
   - gc
```
- I updated greencluster object in edge cluster by applying the following yaml file
```
apiVersion: "stable.example.com/v1"
kind: GreenCluster
metadata:
 name: cluster3
spec:
 target: "target"
 cpu: 4040
 carbon: 100
```
 
- In the edge cluster, the value is updated but in the mailbox it is not.
Attached please find the log file for syncer pod; I see an error saying "failed to update"
[pod.log](https://github.com/kcp-dev/edge-mc/files/11767591/pod.log)
 

### Expected Behavior

The generated resources in mailbox workspace must be updated when the corresponding resources in edge cluster are updated.

### Additional Context

_No response_

yana1205:
## Summary

As reported in #625, there is a bug that upsynced objects are not updated. I fixed the bug and made sure that it addressed the issue with the issue opener (Tayebeh).   

What 's the changes
- Fix to set UID and Resource Version of the target k8s cluster's one when downsyncer/upsyncer push updated objects from upstream/downstream 
- Fix not to downsync upsynced objects (exactly saying not downsync objects having `edge.kcp.io/upsynced` annotation) and similarly not to upsync downsynced objects (`edge.kcp.io/downsynced` annotation) 
- Update tests

## Related issue(s)

Fixes #625

### DIFF
--- a/pkg/syncer/syncers/common.go
+++ b/pkg/syncer/syncers/common.go
@@ -178,13 +178,15 @@ func diff(logger klog.Logger, srcResourceList *unstructured.UnstructuredList, de
 	for _, srcResource := range srcResourceList.Items {
 		destResource, ok := findWithObject(srcResource, destResourceList)
 		if ok {
-			srcResource.SetResourceVersion("")
-			srcResource.SetUID("")
+			srcResource.SetResourceVersion(destResource.GetResourceVersion())
+			srcResource.SetUID(destResource.GetUID())
 			srcResource.SetManagedFields(nil)
 			if hasAnnotation(destResource) {
 				setAnnotation(&srcResource)
+				updatedResources = append(updatedResources, srcResource)
+			} else {
+				logger.V(2).Info(fmt.Sprintf("  ignore adding %s to updatedResources since annotation is not set.", destResource.GetName()))
 			}
-			updatedResources = append(updatedResources, srcResource)
 		} else {
 			srcResource.SetResourceVersion("")
 			srcResource.SetUID("")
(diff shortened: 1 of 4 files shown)


MikeSpreitzer:
The UID and ResourceVersion of an object are things that the apiserver writes, I am surprised to see words here about setting them.

yana1205:
Yes, UID and Resource Version are determined at a server side. However, the UID and Resource Version of an object are different between upstream (mbws) and downstream (workload execution cluster). That's why I fixed the code so that when upsyncer syncs a downstream object to an upstream object, upsyncer replaces the UID and Resource Version of the downstream object with those of the upstream object and then pushes it to the upstream.

I have rewritten the title and description.

MikeSpreitzer:
@yana1205 : thanks for the clarification.  Now I understand and it makes sense.
/lgtm

MikeSpreitzer (approved):

PR #666, opened 2023-06-23
🌱 rename prowjobs to match new github org

xrstf:
## Summary
This brings the Prow names in-sync with the new repo name. As there are now 2 orgs handled by the same Prow instance, the org name is included in the job names (pull-kubestellar-kubestellar-....)


clubanderson (commented):
/lgtm
/approve

PR #695, opened 2023-06-28
:bug:  refactor lc provider

ezrasilvera:
## Summary
Changing the provider interface. 

## Related issue(s)

Fixes #

### DIFF
--- a/clusterprovider/kind/kind.go
+++ b/clusterprovider/kind/kind.go
@@ -160,9 +160,12 @@ func (k *KindWatcher) ResultChan() <-chan clusterprovider.WatchEvent {
 					// Check for new clusters.
 					for _, name := range newSetClusters.Difference(setClusters).UnsortedList() {
 						logger.Info("Detected a new cluster")
-						lcInfo, _ := k.provider.Get(ctx, name)
-						// TODO: handle error
-
+						lcInfo, err := k.provider.Get(ctx, name)
+						if err != nil {
+							// Can't get the cluster info, so let's discover it again
+							newSetClusters.Delete(name)
+							continue
+						}
 						k.ch <- clusterprovider.WatchEvent{
 							Type:   watch.Added,
 							Name:   name,
--- a/clusterprovider/kind/kind.go
+++ b/clusterprovider/kind/kind.go
@@ -18,7 +18,6 @@ package kindprovider
 
 import (
 	"context"
-	"strings"
 	"sync"
 	"time"
 
@@ -46,26 +45,16 @@ func New(providerName string) KindClusterProvider {
 	}
 }
 
-func (k KindClusterProvider) Create(ctx context.Context,
-	name string,
-	opts clusterprovider.Options) (clusterprovider.LogicalClusterInfo, error) {
-	var resCluster clusterprovider.LogicalClusterInfo
-
+func (k KindClusterProvider) Create(ctx context.Context, name string, opts clusterprovider.Options) error {
+	logger := klog.FromContext(ctx)
 	err := k.kindProvider.Create(name)
+
 	if err != nil {
-		if strings.HasPrefix(err.Error(), "node(s) already exist for a cluster with the name") {
-			// TODO: check whether it's the same cluster and return success if true
-		} else {
-			return resCluster, err
-		}
+		// TODO:  Need to differentiate between "already exists" and an error
+		logger.Error(err, "Provider.Create error")
 	}
 
-	cfg, err := k.kindProvider.KubeConfig(name, false)
-	if err != nil {
-		return resCluster, err
-	}
-	resCluster = *clusterprovider.New(cfg, opts)
-	return resCluster, err
+	return err
 }
 
 func (k KindClusterProvider) Delete(ctx context.Context,
@@ -75,7 +64,7 @@ func (k KindClusterProvider) Delete(ctx context.Context,
 	return k.kindProvider.Delete(name, opts.KubeconfigPath)
 }
 
-func (k KindClusterProvider) List() ([]string, error) {
+func (k KindClusterProvider) ListNames(ctx context.Context) ([]string, error) {
 	list, err := k.kindProvider.List()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


pkg/clustermanager/providerclient/client_interface.go
```diff
@@ -91,13 +93,23 @@ type WatchEvent struct {
...
 	// and to refresh the list of logical clusters periodically.
-	List() ([]string, error)
+	List(ctx context.Context) ([]LogicalClusterInfo, error)
+
+	// List returns a list of logical cluster names.
+	// This method is used to discover the initial set of logical clusters
+	// and to refresh the list of logical clusters periodically.
+	ListNames(ctx context.Context) ([]string, error)
+
+	// List returns a list of logical clusters.
```
effi-ofer, line 109: Does it return a list or a single LogicalClusterInfo?

pkg/clustermanager/reconcile_logicalcluster.go
```diff
@@ -88,15 +88,14 @@ func (c *controller) processAddLC(newCluster *lcv1alpha1.LogicalCluster) error {
...
 	//ES: what exactly is this kubeconfig
-	createdCluster, err := provider.Create(c.context, clusterName, opts)
+	err = provider.Create(c.context, clusterName, opts)
 	if err != nil {
 		logger.Error(err, "failed to create cluster")
 		return err
 	}
 	logger.Info("Done creating cluster", "clusterName", clusterName)
 
 	// Update the new cluster's status - specifically the config string and the phase
```
effi-ofer, line 98: Please update the comment. We are not updating the config string.

pkg/clustermanager/discovery.go
```diff
@@ -139,6 +140,8 @@ func processProviderWatchEvents(ctx context.Context, w clusterprovider.Watcher,
 				// TODO: when finalizers added - cheeck the logicalcluster delete timestamp
 				// Logical cluster exists , just update its status
 				reflcluster.Status.Phase = lcv1alpha1apis.LogicalClusterPhaseReady
+				// TODO: Should we really update the config ?
```
effi-ofer, line 143: That's a good question. For managed clusters would we always have all the config info at the time of creation of the LogicalCluster object?  I suspect the answer is negative, but I might be wrong?

pkg/clustermanager/providerclient/client_interface.go
ezrasilvera: list of all names

pkg/clustermanager/discovery.go
ezrasilvera: For now let's update it. We should have a logic for "is it the same cluster", and "what to do if not"

pkg/clustermanager/reconcile_logicalcluster.go
```diff
@@ -88,15 +88,14 @@ func (c *controller) processAddLC(newCluster *lcv1alpha1.LogicalCluster) error {
...
-	newCluster.Status.ClusterConfig = createdCluster.Config
 	newCluster.Status.Phase = lcv1alpha1.LogicalClusterPhaseReady
```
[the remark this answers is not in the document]
ezrasilvera: No. Good catch , but it's not in my code :-)

clusterprovider/kind/kind.go
```diff
@@ -75,7 +64,7 @@ func (k KindClusterProvider) Delete(ctx context.Context,
 	return k.kindProvider.Delete(name, opts.KubeconfigPath)
 }
 
-func (k KindClusterProvider) List() ([]string, error) {
+func (k KindClusterProvider) ListNames(ctx context.Context) ([]string, error) {
```
effi-ofer, line 67: I would rename to ListClusterNames to avoid use of the overloaded Name.

pkg/clustermanager/reconcile_logicalcluster.go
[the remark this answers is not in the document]
ezrasilvera: Changed to initializing

pkg/clustermanager/reconcile_logicalcluster.go
ezrasilvera: done

pkg/clustermanager/discovery.go
```diff
@@ -128,6 +128,7 @@ func processProviderWatchEvents(ctx context.Context, w clusterprovider.Watcher,
 				lcluster.Name = lcName
 				lcluster.Spec.ClusterProviderDescName = providerName
 				lcluster.Spec.Managed = false
+				lcluster.Status.ClusterConfig = event.LCInfo.Config
```
andreyod, line 131: Maybe check that config is not an empty string here. We should not update LC to ready if we don't have config.

clusterprovider/kind/kind.go
ezrasilvera: Changed to ListClusters & ListClusterNames

pkg/clustermanager/discovery.go
ezrasilvera: Maybe Watch should send Add only if cluster is ready

clusterprovider/kind/kind.go
```diff
@@ -85,6 +74,44 @@ func (k KindClusterProvider) List() ([]string, error) {
 	return logicalNameList, err
 }
 
+func (k KindClusterProvider) Get(ctx context.Context, lcName string) (clusterprovider.LogicalClusterInfo, error) {
+	logger := klog.FromContext(ctx)
+	cfg, err := k.kindProvider.KubeConfig(lcName, false)
+	if err != nil {
+		logger.Error(err, "couldn't fetch config for cluster")
```
andreyod, line 81: It's better to return the error. I don't think we can send an event if we don't have config. Return `nil` error in last `return`.

clusterprovider/kind/kind.go
```diff
@@ -123,19 +150,23 @@ func (k *KindWatcher) ResultChan() <-chan clusterprovider.WatchEvent {
...
 						// TODO add logging
 						logger.Error(err, "Getting provider list.")
 						continue
 					}
 					newSetClusters := sets.NewString(list...)
 					// Check for new clusters.
-					for _, cl := range newSetClusters.Difference(setClusters).UnsortedList() {
+					for _, name := range newSetClusters.Difference(setClusters).UnsortedList() {
 						logger.Info("Detected a new cluster")
+						lcInfo, _ := k.provider.Get(ctx, name)
```
andreyod, line 163: I don't think we can send an event if we don't have config.

pkg/clustermanager/providerclient/client_interface.go
effi-ofer: The comment refers to the Get() function. I think line 109 needs to be removed. @ezrasilvera .

PR #719, opened 2023-06-30
✨ Rename scheduler to where resolver

### LINKED ISSUE #690: bug: "schduler" is the wrong name
### Describe the bug

Using the word "scheduler" for the controller that resolves "where predicates" to SinglePlacementSlice object collections is confusing because we are partnering with other projects that also have something they call "scheduling" and is more deserving of the name.  The KubeStellar thing should be called something like "where resolver".

### Steps To Reproduce

See above.

### Expected Behavior

No arguments over which project is doing scheduling.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR renames the "scheduler" to the "where resolver".

This is incomplete, the renaming needs to include changes to GitHub actions.

## Related issue(s)

Fixes #690


MikeSpreitzer:
I successfully ran though example1 manually.

MikeSpreitzer:
@clubanderson : The SonarCloud failure is complaining about problems that are not actually being introduced here.  How do we move forward?

clubanderson:
You have 11.5% dupe code in your commits for this PR. We can simply override prow if this is an anomaly and not a correct analysis. We can also increase the 3% threshold for dupe code for all tests but I would rather not raise this to 12% and simply give an exception for this single PR. Let make it a simple thing to avoid deadlock here. Let me know if you are ok with overriding prow for this single PR

ezrasilvera:
Sorry to come late to the party.  I'm not sure "where resolver"  is the best choice here. It sounds to me too abstracted and too sophisticated. 
I agree scheduler is probably not good.  But why not just "placement manager" ?  This module takes placement decisions, it's completely fine that there are higher level modules that take placement decisions as well  - multi-layer placement is a well known architecture.

MikeSpreitzer:
@ezrasilvera: saying that the thing currently called scheduler "makes placement decisions" is, to my mind, confusing --- particularly in the context of higher level managers that are making placement decisions.  The thing being debated has no authority and no wiggle room.  It does not have any choices to make, it does not have any discretion.  It just evaluates predicates.  We considered this issue at the last community meeting and made a decision.

@clubanderson : the things that SonarCloud is objecting to are not work that I have done in this PR.  SonarCloud is apparently looking at deltas and has been triggered because I renamed some files and made some minor edits in some files that were not touched in PRs since the SonarCloud Action was added.  In short SonarCloud is just now noticing things that have been there all along.  I would like to not bundle fixing those problems into this PR.  Is there a way for you to disable just SonarCloud and not the other Actions nor Prow for this PR?

Also, just to be sure I understand correctly: I see several GitHub Actions that are currently "skipped".  Is that a consequence of the SonarCloud failure?  If it is, is there a way and would it make sense to make them run anyway?  More fun: I expect some will fail or be irrelevant because of #720 .

clubanderson:
@MikeSpreitzer rebase (or push a small change like a carriage return).  I changed the quality gate to ignore dupes for now to get around this.  I will put the dupe check back in place after you pass the test.

clubanderson:
@MikeSpreitzer I have dupe code checking turned off to allow this PR through - can you make it a priority to commit a small change to get this PR running through CI and passing?  Then I can turn dupe-checking back on.

clubanderson:
@MikeSpreitzer if you just do a rebase here you will a) pickup the latest changes and b) get past the dupe check, and then I can c) turn the dupe checker back on.

ezrasilvera:
@MikeSpreitzer I don't understand what you mean by "it doesn't do any decisions, It just evaluates predicates"  evaluate predicates is a type of placement. What's the diff with the k8s scheduler in case we just use predicates (taints, labels, etc..) ? 
Also the fact that there are several layers, doesn't change the fact that each layer is doing placement on its own. For me the definition of "placement module" is: something that produces a list (one or more) of targets (clusters, machines, objects, etc.) , that the object in question should "go" to.  It doesn't perform the actual "transfer". 
In your terminology all placement or schedulers are WhereTo resolvers. The k8s whereTo resolver (which is called scheduler) gets a Pod, a list of targets (nodes) and produces a list of candidate targets that that pod can be "sent" to. The last k8s whereto resolver module pick filter that list to one, and set that as the NodeID of the Pod.  So I don't see the difference between our "whereto resolver" to any placement module out there.

MikeSpreitzer:
@ezrasilvera : the TMC scheduler is doing anycast and so has a choice to make; the EMC "scheduler" is doing multicast and so has no choice to make.  I think it is important to have distinct terminology so that people clearly understand that our stack will not have different managers competing to make the same decision.

docs/content/Coding Milestones/PoC2023q1/coding-milestone-invite-q1.md
```diff
@@ -11,7 +11,7 @@ This coding sprint will provide a great opportunity for you to showcase your ski
 
 The coding sprint will be dedicated to completing the following workload management elements:
 
-- Implementing an edge scheduler and placement translator, including customization options,
+- Implementing an Where Resolver and Placement Translator, including customization options,
```
waltforme, line 14:
```suggestion
- Implementing a Where Resolver and a Placement Translator, including customization options,
```

PR #723, opened 2023-07-02
 :bug: Deal with missed deleted cluster events 

ezrasilvera:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/clustermanager/reconcile_logicalcluster.go
+++ b/pkg/clustermanager/reconcile_logicalcluster.go
@@ -64,22 +64,22 @@ func (c *controller) processAddOrUpdateLC(logicalCluster *lcv1alpha1.LogicalClus
 			LogicalClusters(GetNamespace(logicalCluster.Spec.ClusterProviderDescName)).
 			Delete(c.context, logicalCluster.Name, v1.DeleteOptions{})
 	}
-	if logicalCluster.Status.Phase == "" && logicalCluster.Spec.Managed {
-		// The client created a new logical cluster object and we need to
-		// create the corresponding physical cluster.
-		logger := klog.FromContext(c.context)
-		providerInfo, err := c.clientset.LogicalclusterV1alpha1().ClusterProviderDescs().Get(
-			c.context, logicalCluster.Spec.ClusterProviderDescName, v1.GetOptions{})
-		if err != nil {
-			logger.Error(err, "failed to get the provider resource")
-			return err
-		}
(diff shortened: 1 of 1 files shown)


pkg/clustermanager/reconcile_logicalcluster.go
```diff
@@ -64,22 +64,22 @@ func (c *controller) processAddOrUpdateLC(logicalCluster *lcv1alpha1.LogicalClus
 			LogicalClusters(GetNamespace(logicalCluster.Spec.ClusterProviderDescName)).
 			Delete(c.context, logicalCluster.Name, v1.DeleteOptions{})
 	}
+	providerInfo, err := c.clientset.LogicalclusterV1alpha1().ClusterProviderDescs().Get(
```
andreyod, line 67: Get the provider from cache here.

pkg/clustermanager/reconcile_logicalcluster.go
```diff
@@ -97,18 +97,22 @@ func (c *controller) processAddOrUpdateLC(logicalCluster *lcv1alpha1.LogicalClus
...
-	// case lcv1alpha1.LogicalClusterPhaseReady:
-	// The cluster is ready - we have nothing to do but celebrate!
-	// Likely we got here after the provider finished creating a managed
-	// cluster and dicovery moved it into the Ready state.
-	//
-	// If a logical cluster has been created for an unmanaged physical
-	// cluster, then wait for discovery to move the phase to Ready.
+
+	// The code from here deals with the case of provider deleted cluster events that might be missed
+	_, err = provider.Get(c.context, logicalCluster.Name)
```
andreyod, line 102: This is done on each event? What is the performance hit?

effi-ofer: General question - how frequent are LogicalCluster events? I would have thought that they occur only on cluster changes (which is infrequent) and on startup.  And what's the expected number of clusters?

effi-ofer:
Is it possible to lose a delete event if we have a finalizer?

pkg/clustermanager/reconcile_logicalcluster.go
ezrasilvera: This will be taken from a cache I'm building

clubanderson:
please review and close as snowflake if no longer valid

effi-ofer:
> please review and close as snowflake if no longer valid

I think this is still needed but the branch needs to be changed to main and the code rebased.

ezrasilvera:
not  relevant for ks-0.20

PR #724, opened 2023-07-02
:bug:Fixes in discovery and reconcilers.

andreyod:
## Related issue(s)

Fixes #

### DIFF
--- a/clusterprovider/kind/kind.go
+++ b/clusterprovider/kind/kind.go
@@ -75,10 +75,8 @@ func (k KindClusterProvider) ListClustersNames(ctx context.Context) ([]string, e
 }
 
 func (k KindClusterProvider) Get(ctx context.Context, lcName string) (clusterprovider.LogicalClusterInfo, error) {
-	logger := klog.FromContext(ctx)
 	cfg, err := k.kindProvider.KubeConfig(lcName, false)
 	if err != nil {
-		logger.Error(err, "couldn't fetch config for cluster")
 		return clusterprovider.LogicalClusterInfo{}, err
 	}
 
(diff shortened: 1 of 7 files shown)


cmd/logical-cluster-manager/main.go
```diff
@@ -34,7 +34,7 @@ import (
 )
 
 var (
-	resyncPeriod = 4 * time.Second
```
effi-ofer, line 37: Perhaps 30 is the right value for production, but during development/testing a shorter period means we won't have to wait as long.

clusterprovider/kind/kind.go
```diff
@@ -163,6 +161,7 @@ func (k *KindWatcher) ResultChan() <-chan clusterprovider.WatchEvent {
 						logger.Info("Detected a new cluster")
 						lcInfo, err := k.provider.Get(ctx, name)
 						if err != nil {
+							logger.Info("Kind cluster is not ready. Retrying.", "cluster", name)
```
effi-ofer, line 164: We need to decide on one style of logging.  I started using the style you previously used in the reconcilers `c.logger.V(2).Error(err, "failed to get the provider resource", newCluster.Name)` but I am up to whatever you prefer as long as we're consistent :)

pkg/clustermanager/reconcile_clusterprovider.go
```diff
@@ -51,16 +51,22 @@ func (c *controller) reconcileClusterProviderDesc(key string) error {
...
+	if provider.Status.Phase == lcv1alpha1.ClusterProviderDescPhaseInitializing {
 		return nil
 	}
+	name := provider.Name
+	if provider.Status.Phase == lcv1alpha1.ClusterProviderDescPhaseReady {
+		// check if we have this provider in cache
+		if _, ok := c.providers[name]; ok {
+			return nil
+		}
+	}
```
effi-ofer, line 63: It's possible to see a Ready provider before it's in the cache?

PR #725, opened 2023-07-02
:bug: restructure provider

ezrasilvera:
## Related issue(s)

Fixes #


pkg/clustermanager/provider.go
```diff
@@ -20,29 +20,152 @@ import (
...
 const prefixNamespace = "lcprovider-"
 
-func GetNamespace(providerName string) string {
-	return prefixNamespace + providerName
+type provider struct {
+	name            string
+	providerClient  clusterprovider.ProviderClient
+	c               *controller
+	providerWatcher clusterprovider.Watcher
+	nameSpace       string
```
effi-ofer, line 41: nitpicking: should it be capitalized this way? Isn't the convention `namespace` one word?

pkg/clustermanager/provider.go
```diff
@@ -20,29 +20,152 @@ import (
...
 
 // Each provider gets its own namespace named prefixNamespace+providerName
 const prefixNamespace = "lcprovider-"
 
-func GetNamespace(providerName string) string {
-	return prefixNamespace + providerName
+type provider struct {
+	name            string
+	providerClient  clusterprovider.ProviderClient
+	c               *controller
```
effi-ofer, line 39: are we using the controller only to get to the clientset?

pkg/clustermanager/reconcile_clusterprovider.go
```diff
@@ -99,14 +99,14 @@ func (c *controller) handleAdd(provider *lcv1alpha1.ClusterProviderDesc) error {
 }
 
 func (c *controller) handleDelete(providerName string) error {
-	err := c.StopDiscovery(providerName)
+	ns := c.providers[providerName].nameSpace
```
andreyod, line 102: This will panic if the key is not in the map.

ezrasilvera: fixed

pkg/clustermanager/reconcile_clusterprovider.go
```diff
@@ -131,6 +134,7 @@ func (c *controller) handleDelete(providerName string) error {
 		}
 	}
 	if isNsEmpty {
+		delete(c.providers, providerName)
```
andreyod, line 137: We need to delete from cache unconditionally, otherwise we may have cached provider which object is gone. Also we should delete under lock.

PR #727, opened 2023-07-03
✨ Add example1 variant with single git clone

### LINKED ISSUE #720: bug: self-reference problem in executable docs
### Describe the bug

I am working on renaming the "Scheduler" to the "Where Resolver".  I have a PR against the `main` branch, #719 .  I would like to test the executable documents before the PR merges.  But I cannot.  For example, https://github.com/MikeSpreitzer/kcp-edge-mc/blob/38f05ef7ef8c705971cf79b93705f78bb0b57b4e/docs/content/Coding%20Milestones/PoC2023q1/example1-subs/example1-start-kcp.md?plain=1#L20 together with config files in the same commit, contribute the following bash command to be executed.

```bash
git clone -b main https://github.com/kubestellar/kubestellar
```

and from that point on it is hopeless, the commit that I want to test is ignored because the current `main` is what will be used.

### Steps To Reproduce

See above.

### Expected Behavior

I think it would be better if an executable document could exercise the contents of the same commit as the document resides in, rather than something else fetched from afar.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR introduces `example1-local.md` that, when executed, uses the local content rather than fetching from github.  Thus, this executable doc can be used for CI and can be used by a developer to test before pushing and even before committing.

## Related issue(s)

Making all the executable docs work like this would resolve #720

### DIFF
--- a/docs/content/Coding Milestones/PoC2023q1/example1-local.md
+++ b/docs/content/Coding Milestones/PoC2023q1/example1-local.md
@@ -1,0 +1,83 @@
+---
+short_name: example1-local
+manifest_name: 'docs/content/Coding Milestones/PoC2023q1/example1-local.md'
+pre_req_name: 'docs/content/common-subs/pre-req.md'
+---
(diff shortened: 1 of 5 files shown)


MikeSpreitzer:
Agreed with @clubanderson that the better approach is to change the existing documents to not do nested fetches.

PR #729, opened 2023-07-03
:sparkles: add prefix filtering

ezrasilvera:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/clustermanager/controller.go
+++ b/pkg/clustermanager/controller.go
@@ -45,13 +45,6 @@ const (
 	controllerName                                   = "logicalcluster-manager"
 )
 
-func LCKeyFunc(ns string, name string) string {
-	if ns != "" {
-		return ns + "/" + name
-	}
-	return name
-}
-
 type queueItem struct {
 	triggeringKind triggeringKind
 	key            string
--- a/pkg/clustermanager/controller.go
+++ b/pkg/clustermanager/controller.go
@@ -45,6 +45,13 @@ const (
 	controllerName                                   = "logicalcluster-manager"
 )
 
+func LCKeyFunc(ns string, name string) string {
+	if ns != "" {
+		return ns + "/" + name
+	}
+	return name
+}
+
 type queueItem struct {
 	triggeringKind triggeringKind
 	key            string
--- a/pkg/clustermanager/provider.go
+++ b/pkg/clustermanager/provider.go
@@ -19,10 +19,12 @@ package clustermanager
 import (
 	"context"
 	"errors"
+	"strings"
 
 	"github.com/go-logr/logr"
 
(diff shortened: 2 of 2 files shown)


pkg/clustermanager/provider.go
```diff
@@ -38,11 +40,12 @@ func ProviderNS(name string) string {
...
-	providerClient  clusterprovider.ProviderClient
-	c               *controller
-	providerWatcher clusterprovider.Watcher
-	nameSpace       string
+	name             string
+	providerClient   clusterprovider.ProviderClient
+	c                *controller
+	providerWatcher  clusterprovider.Watcher
+	nameSpace        string
+	discoveryPresfix string
```
andreyod, line 48: Typo `discoveryPresfix` -> `discoveryPrefix`

pkg/clustermanager/controller.go
```diff
@@ -45,6 +45,13 @@ const (
 	controllerName                                   = "logicalcluster-manager"
 )
 
+func LCKeyFunc(ns string, name string) string {
```
andreyod, line 48: Looks like this is a local helper function so it shouldn't be exported and maybe should be moved to provider.go where it is called.

pkg/clustermanager/provider.go
```diff
@@ -75,17 +80,33 @@ func CreateProvider(c *controller, providerName string,
 		return nil, errors.New("unknown provider type")
 	}
 
+	providerDescObj, found, _ := c.clusterProviderInformer.GetIndexer().GetByKey(LCKeyFunc("", providerName))
+	if found {
+		providerDesc, ok := providerDescObj.(*lcv1alpha1apis.ClusterProviderDesc)
+		if ok {
+			discoveryPresfix = providerDesc.Spec.ClusterPrefixForDiscovery
```
andreyod, line 87: What happens if it's an empty string?

pkg/clustermanager/provider.go
```diff
@@ -113,21 +134,34 @@ func (p *provider) StopDiscovery() error {
...
 
+		if found {
+			reflcluster, ok = reflclusterObj.(*lcv1alpha1apis.LogicalCluster)
+			if !ok {
+				runtime.HandleError(errors.New("unexpected object type. expected ClusterProviderDesc"))
+				continue
+			}
+		}
+
+		if !found || (found && !reflcluster.Spec.Managed) {
```
andreyod, line 155: This logic probably deserves a comment.

pkg/clustermanager/provider.go
ezrasilvera: fixed

pkg/clustermanager/controller.go
ezrasilvera: fixed

PR #738, opened 2023-07-04
✨ No nested fetch

### LINKED ISSUE #720: bug: self-reference problem in executable docs

MikeSpreitzer:
## Summary
This PR changes the executable docs to not fetch the KubeStellar repo but rather use the copy of the repo that they find themselves executing in.

This makes the executable documents useful as tests, to developers and to CI.

## Related issue(s)

Fixes #720

### DIFF
--- a/docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
+++ b/docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
@@ -8,27 +8,28 @@ kubectl ws root
 kubectl ws create imw-1 --enter
 ```
 
-### Get KubeStellar
+### Get and build or install KubeStellar
 
-Download and build or install
-<a href="{{ config.repo_url }}">KubeStellar</a>, according to your
-preference.  That is, either (a) `git clone` the repo and then `make
-build` to populate its `bin` directory, or (b) fetch the binary
-archive appropriate for your machine from a release and unpack it
-(creating a `bin` directory).  In the following exhibited command
-lines, the commands described as "KubeStellar commands" and the commands
-that start with `kubectl kubestellar` rely on the KubeStellar `bin` directory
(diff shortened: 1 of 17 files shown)


clubanderson:
@MikeSpreitzer I checked out this PR and did mkdocs serve.  When viewing each of the examples I noticed you are doing a "make build" sequence twice in each.  Is this intentional?

MikeSpreitzer:
@clubanderson: I just reviewed `http://127.0.0.1:8000/Coding%20Milestones/PoC2023q1/example1/` and found `make build` twice, as expected --- once for kcp and once for kubestellar.  Are you seeing something different?

clubanderson:
Ok. That's what I thought. Just checking to make sure that was your intent.

clubanderson (commented):
/lgtm
/approve

clubanderson:
@MikeSpreitzer the reason why the docs-ecutable were skipped is because you currently need 'ok-to-test' label to allow those tests to run.  With changes to the workflows for pull and push, this may no longer be necessary

PR #741, opened 2023-07-04
:bug:Fix logical clusters issues.

andreyod:
## Summary
1. Do not return from provider watch loop.
2. Enqueue for all object changes on update event.
3. Do not use controller context and logger outside the controller.
4. Change some log messages.

## Related issue(s)

Fixes #

### DIFF
--- a/clusterprovider/kind/kind.go
+++ b/clusterprovider/kind/kind.go
@@ -45,26 +45,26 @@ func New(providerName string) KindClusterProvider {
 	}
 }
 
-func (k KindClusterProvider) Create(ctx context.Context, name string, opts clusterprovider.Options) error {
-	logger := klog.FromContext(ctx)
+func (k KindClusterProvider) Create(name string, opts clusterprovider.Options) error {
+	logger := klog.Background()
+	logger.Info("Creating Kind cluster", "name", name)
 	err := k.kindProvider.Create(name)
 
 	if err != nil {
 		// TODO:  Need to differentiate between "already exists" and an error
-		logger.Error(err, "Provider.Create error")
+		logger.Error(err, "Failed to create cluster", "name", name)
 	}
 
 	return err
 }
 
-func (k KindClusterProvider) Delete(ctx context.Context,
-	name string,
-	opts clusterprovider.Options) error {
-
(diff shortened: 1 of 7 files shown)


clusterprovider/kind/kind.go
```diff
@@ -174,7 +173,7 @@ func (k *KindWatcher) ResultChan() <-chan clusterprovider.WatchEvent {
 					}
 					// Check for deleted clusters.
 					for _, cl := range setClusters.Difference(newSetClusters).UnsortedList() {
-						logger.Info("Detected cluster was deleted.")
+						logger.Info("Processing Kind cluster delete", "name", cl)
```
ezrasilvera, line 176: What was wrong with the Detected?  why Processing ?

config/samples/clusterproviderdesc.yaml
```diff
@@ -4,3 +4,4 @@ metadata:
   name: default
 spec:
   ProviderType: "kind"
+  ClusterPrefixForDiscovery: "kubestellar"
```
ezrasilvera, line 7: too long, and we want disconnect from kubestellar.  You can put "ks-"  Please add "-"

config/samples/logicalcluster_dev.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: logicalcluster.kubestellar.io/v1alpha1
 kind: LogicalCluster
 metadata:
-  name: dev
+  name: kubestellar-dev
```
ezrasilvera, line 4: We want the managed cluster to be without the prefix !

clusterprovider/kind/kind.go
```diff
@@ -148,20 +148,19 @@ func (k *KindWatcher) ResultChan() <-chan clusterprovider.WatchEvent {
 			for {
 				select {
 				// TODO replace the 2 with a param at the cluster-provider-client level
-				case <-time.After(5 * time.Second):
-					list, err := k.provider.ListClustersNames(ctx)
+				case <-time.After(2 * time.Second):
```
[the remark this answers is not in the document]
effi-ofer: @ezrasilvera Handy for testing?  I suggest creating a const and opening an issue to update it when when we are more production worthy.

Issue #711, opened 2023-06-29
feature: AAA support for KubeStellar


clubanderson:
### Feature Description

Authentication, Authorization, and Accounting are not currently implemented in KubeStellar.  MCAD is requesting support for this.

### Proposed Solution

- [ ] Introduce AAA support in KubeStellar
- [ ] create a tutorial on how to use it.

### Want to contribute?

- [ ] I would like to work on this issue.

MikeSpreitzer:
KubeStellar uses standard Kubernetes mechanisms for AAA.

z103cb:
Yes, we need this !

clubanderson:
@MikeSpreitzer can you expand on your comment to see if this is an acceptable answer for @z103cb ?

MikeSpreitzer:
https://ibm-research.slack.com/archives/C05EK28MYJ1/p1688059961691769?thread_ts=1688054427.563179&cid=C05EK28MYJ1

clubanderson:
Ok @MikeSpreitzer i understand now. Will wait for more info from Hai!

MikeSpreitzer:
Also: some details about authorization appear in the following places.

(I am citing specific commits so that the content does not change out from under me, but to see these nicely rendered you would go to the corresponding page on the website --- which we edit a lot.)

- https://github.com/kubestellar/kubestellar/blob/v0.3.0/docs/content/Coding%20Milestones/PoC2023q1/mailbox-controller.md#usage
- https://github.com/kubestellar/kubestellar/blob/v0.3.0/docs/content/Coding%20Milestones/PoC2023q1/placement-translator.md#usage
- https://github.com/kubestellar/kubestellar/blob/v0.3.0/docs/content/Coding%20Milestones/PoC2023q1/kubestellar-scheduler.md should have an anlogous section but does not (I think?  @waltforme ?)
- https://github.com/kubestellar/kubestellar/blob/v0.3.0/docs/content/Coding%20Milestones/PoC2023q1/commands.md

huang195:
[...]

PR #757, opened 2023-07-06
✨ Make the where-resolver less privileged

waltforme:
## Summary
This PR removes unnecessary privileges from the where-resolver, by switching from the `shard-admin` user to the `kcp-admin` user.

Below are the two relevant contexts in kcp's admin-kubeconfig:
```yaml
- context:
    cluster: base
    user: kcp-admin
  name: base
- context:
    cluster: base
    user: shard-admin
  name: system:admin
```

/cc @MikeSpreitzer 

## Related issue(s)
#711

### DIFF
--- a/cmd/kubestellar-where-resolver/options/options.go
+++ b/cmd/kubestellar-where-resolver/options/options.go
@@ -26,10 +26,10 @@ import (
 )
 
 type Options struct {
-	EspwClientOpts   clientoptions.ClientOpts
-	RootClientOpts   clientoptions.ClientOpts
-	SysAdmClientOpts clientoptions.ClientOpts
-	Logs             *logs.Options
+	EspwClientOpts clientoptions.ClientOpts
+	RootClientOpts clientoptions.ClientOpts
+	BaseClientOpts clientoptions.ClientOpts
+	Logs           *logs.Options
 }
 
 func NewOptions() *Options {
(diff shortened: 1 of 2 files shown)


cmd/kubestellar-where-resolver/options/options.go
```diff
@@ -38,19 +38,19 @@ func NewOptions() *Options {
 	logs.Config.Verbosity = config.VerbosityLevel(2)
 
 	return &Options{
-		EspwClientOpts:   *clientoptions.NewClientOpts("espw", "access to the edge service provider workspace"),
-		RootClientOpts:   *clientoptions.NewClientOpts("root", "access to all clusters"),
-		SysAdmClientOpts: *clientoptions.NewClientOpts("sysadm", "access to all clusters as system:admin"),
-		Logs:             logs,
+		EspwClientOpts: *clientoptions.NewClientOpts("espw", "access to the edge service provider workspace"),
+		RootClientOpts: *clientoptions.NewClientOpts("root", "access to the root workspace"),
+		BaseClientOpts: *clientoptions.NewClientOpts("base", "access to the plain apiserver address which is the base of all workspaces"),
```
MikeSpreitzer, line 43: "access" to an "address" is not clear.  The concern is access to API objects in workspaces.  Perhaps a clearer description would be something like the following.

> access to all logical clusters as kcp-admin

MikeSpreitzer (approved):
/lgtm
Thanks!

Issue #760, opened 2023-07-06
bug:  Functions `kcp_get_latest_version` and `kubestellar_get_latest_version` in `bootstrap-kubestellar.sh` do not work anymore


francostellari:
### Describe the bug

Due to same change in GitHub pages, the two functions fail

### Steps To Reproduce

`curl -sL https://github.com/kubestellar/kubestellar/releases/latest | grep "</h1>" | head -n 1 | sed -e 's/<[^>]*>//g' | xargs`

returns garbage instead of `v0.3.1`

### Expected Behavior

`v0.3.1`

### Additional Context

Fix it by changing `head -n 1` to `tail -n 1`

francostellari:
@dumb0002 this is fixed by pull req. #761

clubanderson:
fixed - Thank you @francostellari

PR #761, opened 2023-07-06
✨ Update bootstrap-kubestellar.sh

francostellari:
## Summary

Fix the issue retrieving the latest release of kcp and kubestellar

Fixes #
#760
#767

### DIFF
--- a/bootstrap/bootstrap-kubestellar.sh
+++ b/bootstrap/bootstrap-kubestellar.sh
@@ -76,7 +76,7 @@ kcp_version() {
 }
 
 kcp_get_latest_version() {
-    curl -sL https://github.com/kcp-dev/kcp/releases/latest | grep "</h1>" | head -n 1 | sed -e 's/<[^>]*>//g' | xargs
+    curl -sL https://github.com/kcp-dev/kcp/releases/latest | grep "</h1>" | tail -n 1 | sed -e 's/<[^>]*>//g' | xargs
 }
 
 kubeconfig_valid() {
@@ -118,7 +118,7 @@ kubestellar_running() {
 }
 
 kubestellar_get_latest_version() {
-    curl -sL https://github.com/kubestellar/kubestellar/releases/latest | grep "</h1>" | head -n 1 | sed -e 's/<[^>]*>//g' | xargs
+    curl -sL https://github.com/kubestellar/kubestellar/releases/latest | grep "</h1>" | tail -n 1 | sed -e 's/<[^>]*>//g' | xargs
 }
 
 get_os_type() {


francostellari:
@dumb0002

bootstrap/bootstrap-kubestellar.sh
```diff
@@ -76,7 +76,7 @@ kcp_version() {
 }
 
 kcp_get_latest_version() {
-    curl -sL https://github.com/kcp-dev/kcp/releases/latest | grep "</h1>" | head -n 1 | sed -e 's/<[^>]*>//g' | xargs
+    curl -sL https://github.com/kcp-dev/kcp/releases/latest | grep "</h1>" | tail -n 1 | sed -e 's/<[^>]*>//g' | xargs
```
MikeSpreitzer, line 79: This is not wrong, but I wonder if you would rather have a more robust search based on more details of the content (e.g., `fgrep '<h1 data-view-component="true"'`).  Following is what I see right now.

```console
(base) mspreitz@mjs12 kubestellar % curl -sL https://github.com/kcp-dev/kcp/releases/latest | grep "</h1>"
      <h1 id="search-suggestions-dialog-header" class="sr-only">Search code, repositories, users, issues, pull requests...</h1>
      </h1>
      </h1>
        <h1 data-view-component="true" class="d-inline mr-3">v0.20.0</h1>
```

francostellari: @MikeSpreitzer ideally, I would like to have a file `latest.txt` and `stable.txt` in the repo pointing to the stable and latest release.

See what go does: `curl 'https://go.dev/VERSION?m=text'`

MikeSpreitzer: OTOH, including in a commit a file that references the commit is problematic.  That is why I think that a GitHub API would be the best solution.

MikeSpreitzer: If we had a different repo in which we could put the pointer to "stable" and "latest", that would be a resolution.

MikeSpreitzer: How do the go developer control what `curl 'https://go.dev/VERSION?m=text'` produces?

MikeSpreitzer: Also, one of the answers in https://stackoverflow.com/questions/51933636/url-for-latest-stable-go-release says that `https://go.dev/dl/?mode=json` should be read instead.

francostellari: > OTOH, including in a commit a file that references that commit (or a tag that will be created after the commit exists) is problematic. That is why I think that a GitHub API would be the best solution.

@MikeSpreitzer I don't share your concern. The main branch will hold the stable/latest files, not the tag releases. The stable could be rolled back anytime if  an issue is discovered.
After the binary release is done, the main branch will decide to update the files as appropriate to advance the users.
The url to stable/latest.txt should be immutable, so it should be in the main branch.

francostellari:
I have opened https://github.com/kubestellar/kubestellar/pull/763

francostellari:
@clubanderson This PR should fix your issue #767, which is the same issue reported in my #760

clubanderson:
ok, so we are waiting on you and @MikeSpreitzer to hash this out.  Thank you

francostellari:
> ok, so we are waiting on you and @MikeSpreitzer to hash this out. Thank you

francostellari:
Sorry I closed by mistake

francostellari:
Since @MikeSpreitzer agrees that this PR fixes the problem, I recommend pushing the fix ASAP.
I'm using the VERSION file in PR #763 to have a long term fix for KubeStellar, alas not kcp

clubanderson:
our QS is broken, this should fix it, convo can continue in another PR at a later time.

/lgtm
/approve

PR #764, opened 2023-07-06
✨ Docker container for KubeStellar

francostellari:
## Summary

Provide:
- `Dockerfile` to create a **KubeStellar** container image
- `Makefile` to automate the processes of creating the image, running the container, cleaning up
- `README.md` detailing the steps and options

I'm currently hosting the following images... that will hopefully move to the KubeStellar registry:
- quay.io/stellari/kubestellar:v0.3.0
- quay.io/stellari/kubestellar:v0.3.1 (==latest)

## Related issue(s)
https://kubernetes.slack.com/archives/C058SUSL5AA/p1688543882794829
https://kubernetes.slack.com/archives/C058SUSL5AA/p1688657230224689

Fixes #

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -20,29 +20,26 @@ FROM redhat/ubi9 AS builder
 ARG TARGETOS
 ARG TARGETARCH
 ARG TARGETPLATFORM
-ARG KUBESTELLAR_VERSION=v0.3.1
+ARG KUBESTELLAR_VERSION
 
 WORKDIR /
 
-# install requirements
-RUN yum install -y wget
-
 # download kubectl binaries
-RUN wget -nv "https://dl.k8s.io/release/v1.25.3/bin/${TARGETPLATFORM}/kubectl" -O kubectl && \
+RUN curl -SL -o kubectl "https://dl.k8s.io/release/v1.25.3/bin/${TARGETPLATFORM}/kubectl" && \
     chmod 0755 kubectl
 
 # download kcp binaries
-RUN curl -SL -o kcp.tar.gz https://github.com/kcp-dev/kcp/releases/download/v0.11.0/kcp_0.11.0_${TARGETOS}_${TARGETARCH}.tar.gz && \
+RUN curl -SL -o kcp.tar.gz "https://github.com/kcp-dev/kcp/releases/download/v0.11.0/kcp_0.11.0_${TARGETOS}_${TARGETARCH}.tar.gz" && \
     mkdir kcp && \
     tar -C kcp -zxf kcp.tar.gz
 
 # download kcp plugins
-RUN curl -SL -o kcp-plugins.tar.gz https://github.com/kcp-dev/kcp/releases/download/v0.11.0/kubectl-kcp-plugin_0.11.0_${TARGETOS}_${TARGETARCH}.tar.gz && \
+RUN curl -SL -o kcp-plugins.tar.gz "https://github.com/kcp-dev/kcp/releases/download/v0.11.0/kubectl-kcp-plugin_0.11.0_${TARGETOS}_${TARGETARCH}.tar.gz" && \
     mkdir kcp-plugins && \
     tar -C kcp-plugins -zxf kcp-plugins.tar.gz
 
 # download KubeStellar binaries
-RUN curl -SL -o kubestellar.tar.gz https://github.com/kubestellar/kubestellar/releases/download/${KUBESTELLAR_VERSION}/kubestellar_${KUBESTELLAR_VERSION}_${TARGETOS}_${TARGETARCH}.tar.gz && \
+RUN curl -SL -o kubestellar.tar.gz "https://github.com/kubestellar/kubestellar/releases/download/${KUBESTELLAR_VERSION}/kubestellar_${KUBESTELLAR_VERSION}_${TARGETOS}_${TARGETARCH}.tar.gz" && \
     mkdir kubestellar && \
     tar -C kubestellar -zxf kubestellar.tar.gz
 
@@ -76,6 +73,7 @@ USER kubestellar
 # setup the environment variables
 ENV PATH=$PATH:/kcp:/kcp-plugins:/kubestellar/bin
 ENV KUBECONFIG=/.kcp/admin.kubeconfig
+ENV EXTERNAL_HOSTNAME=""
 
 # start KubeStellar
 CMD [ "/entry.sh" ]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -15,7 +15,7 @@
 ###############################################################################
 # Builder image
 ###############################################################################
-FROM ubuntu AS builder
+FROM redhat/ubi9 AS builder
 
 ARG TARGETOS
 ARG TARGETARCH
@@ -25,8 +25,7 @@ ARG KUBESTELLAR_VERSION=v0.3.1
 WORKDIR /
 
 # install requirements
-RUN apt-get update && \
-    apt-get install -y curl wget
+RUN yum install -y wget
 
 # download kubectl binaries
 RUN wget -nv "https://dl.k8s.io/release/v1.25.3/bin/${TARGETPLATFORM}/kubectl" -O kubectl && \
@@ -53,11 +52,10 @@ ADD . /
 ###############################################################################
 # KubeStellar image
 ###############################################################################
-FROM ubuntu
+FROM redhat/ubi9
 
 # install requirements (for running inside the container)
-RUN apt-get update && \
-    apt-get install -y jq
+RUN yum install -y jq procps
 
 WORKDIR /
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,76 @@
+# Copyright 2023 The KubeStellar Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+###############################################################################
+# Builder image
+###############################################################################
+FROM ubuntu AS builder
+
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETPLATFORM
+ARG KUBESTELLAR_VERSION=v0.3.1
+
+WORKDIR /
+
+# install requirements
+RUN apt-get update && \
+    apt-get install -y curl wget
+
+# download kubectl binaries
+RUN wget -nv "https://dl.k8s.io/release/v1.25.3/bin/${TARGETPLATFORM}/kubectl" -O kubectl && \
+    chmod 0755 kubectl
+
+# download kcp binaries
+RUN curl -SL -o kcp.tar.gz https://github.com/kcp-dev/kcp/releases/download/v0.11.0/kcp_0.11.0_${TARGETOS}_${TARGETARCH}.tar.gz && \
+    mkdir kcp && \
+    tar -C kcp -zxf kcp.tar.gz
+
+# download kcp plugins
+RUN curl -SL -o kcp-plugins.tar.gz https://github.com/kcp-dev/kcp/releases/download/v0.11.0/kubectl-kcp-plugin_0.11.0_${TARGETOS}_${TARGETARCH}.tar.gz && \
+    mkdir kcp-plugins && \
+    tar -C kcp-plugins -zxf kcp-plugins.tar.gz
+
+# download KubeStellar binaries
+RUN curl -SL -o kubestellar.tar.gz https://github.com/kubestellar/kubestellar/releases/download/${KUBESTELLAR_VERSION}/kubestellar_${KUBESTELLAR_VERSION}_${TARGETOS}_${TARGETARCH}.tar.gz && \
+    mkdir kubestellar && \
+    tar -C kubestellar -zxf kubestellar.tar.gz
+
+# add entry script
+ADD . /
+
+###############################################################################
+# KubeStellar image
+###############################################################################
+FROM ubuntu
+
+# install requirements (for running inside the container)
+RUN apt-get update && \
+    apt-get install -y jq
+
+WORKDIR /
+
+# copy binaries from the builder image
+COPY --from=builder /kubectl         /usr/local/bin/kubectl
+COPY --from=builder /kcp/bin         /kcp
+COPY --from=builder /kcp-plugins/bin /kcp-plugins
+COPY --from=builder /kubestellar     /kubestellar
+COPY --from=builder /entry.sh        /entry.sh
+
+# setup the environment variables
+ENV PATH=$PATH:/kcp:/kcp-plugins:/kubestellar/bin
+ENV KUBECONFIG=/.kcp/admin.kubeconfig
+
+# start KubeStellar
+CMD [ "/entry.sh" ]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -46,16 +46,17 @@ RUN curl -SL -o kubestellar.tar.gz https://github.com/kubestellar/kubestellar/re
     mkdir kubestellar && \
     tar -C kubestellar -zxf kubestellar.tar.gz
 
-# add entry script
-ADD . /
-
 ###############################################################################
 # KubeStellar image
 ###############################################################################
 FROM redhat/ubi9
 
 # install requirements (for running inside the container)
-RUN yum install -y jq procps
+RUN yum install -y jq procps && \
+    groupadd kubestellar && \
+    adduser -g kubestellar kubestellar && \
+    mkdir -p /.kcp /kubestellar-logs && \
+    chown -R kubestellar:kubestellar /.kcp /kubestellar-logs
 
 WORKDIR /
 
@@ -64,7 +65,13 @@ COPY --from=builder /kubectl         /usr/local/bin/kubectl
 COPY --from=builder /kcp/bin         /kcp
 COPY --from=builder /kcp-plugins/bin /kcp-plugins
 COPY --from=builder /kubestellar     /kubestellar
-COPY --from=builder /entry.sh        /entry.sh
+
+# add entry script
+ADD entry.sh /entry.sh
+
+VOLUME [ "/.kcp", "/kubestellar-logs", "/kcp", "/kcp-plugins", "/kubestellar"]
+
+USER kubestellar
 
 # setup the environment variables
 ENV PATH=$PATH:/kcp:/kcp-plugins:/kubestellar/bin
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -78,7 +78,7 @@ exec:
 
 # Kill the container, remove the container image and files
 .PHONY: stop
-clean:
+stop:
 	$(info Arguments: NAME=${NAME})
 	- docker rm --force ${NAME}
 
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -21,6 +21,7 @@ KUBESTELLAR_LATEST := $(shell curl -sL https://raw.githubusercontent.com/kubeste
 KUBESTELLAR_VERSION ?= ${KUBESTELLAR_STABLE}
 TAG ?= ${KUBESTELLAR_VERSION}
 BASEPATH ?= ${HOME}
+EXTERNAL_HOSTNAME ?= "" # $(shell ifconfig | grep -A 1 "enp0s8" | tail -1 | awk '{print $$2}')
 
 
 # Build a local container image for the host architecture
@@ -55,11 +56,13 @@ run:
 	$(info Arguments: TAG=${TAG})
 	$(info Arguments: NAME=${NAME})
 	$(info Arguments: BASEPATH=${BASEPATH})
+	$(info Arguments: EXTERNAL_HOSTNAME=${EXTERNAL_HOSTNAME})
 	mkdir -p ${BASEPATH}/.kcp ${BASEPATH}/kubestellar-logs
-		# -p 6443:6443 \
 	docker run -d \
+		-p 6443:6443 \
 		-v ${BASEPATH}/.kcp:/.kcp \
 		-v ${BASEPATH}/kubestellar-logs:/kubestellar-logs \
+		-e EXTERNAL_HOSTNAME=${EXTERNAL_HOSTNAME} \
 		--user $(shell id -u) \
 		--name ${NAME} ${IMG}:${TAG}
 	docker cp ${NAME}:/kcp-plugins ${BASEPATH}/kcp
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -1,5 +1,3 @@
-#!/bin/env make
-
 # Copyright 2023 The KubeStellar Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -13,31 +13,37 @@
 # limitations under the License.
 
 # Global arguments
-IMG := kubestellar
-NAME := kubestellar
-PLATFORMS := linux/amd64,linux/arm64,linux/ppc64le
-KUBESTELLAR_VERSION := $(shell curl -sL https://github.com/kubestellar/kubestellar/releases/latest | grep "</h1>" | tail -n 1 | sed -e 's/<[^>]*>//g' | xargs)
+IMG ?= kubestellar
+NAME ?= kubestellar
+PLATFORMS ?= linux/amd64,linux/arm64,linux/ppc64le # kcp does not support linux/s390x
+KUBESTELLAR_STABLE := $(shell curl -sL https://raw.githubusercontent.com/kubestellar/kubestellar/main/VERSION | grep stable | sed -n -e 's/^.*=//p')
+KUBESTELLAR_LATEST := $(shell curl -sL https://raw.githubusercontent.com/kubestellar/kubestellar/main/VERSION | grep latest | sed -n -e 's/^.*=//p')
+KUBESTELLAR_VERSION ?= ${KUBESTELLAR_STABLE}
+TAG ?= ${KUBESTELLAR_VERSION}
+BASEPATH ?= ${HOME}
 
 
 # Build a local container image for the host architecture
 .PHONY: build
 build:
-	$(info Building the KubeStellar container using Dockerfile...)
-	$(info Arguments: IMG=${IMG})
+	$(info Building the KubeStellar container using the Dockerfile...)
 	$(info Arguments: KUBESTELLAR_VERSION=${KUBESTELLAR_VERSION})
-	# docker buildx build --build-arg="KUBESTELLAR_VERSION=${KUBESTELLAR_VERSION}" --tag ${IMG}:${KUBESTELLAR_VERSION} .
-	docker build --build-arg="KUBESTELLAR_VERSION=${KUBESTELLAR_VERSION}" --tag ${IMG}:${KUBESTELLAR_VERSION} .
+	$(info Arguments: IMG=${IMG})
+	$(info Arguments: TAG=${TAG})
+	docker build --build-arg="KUBESTELLAR_VERSION=${KUBESTELLAR_VERSION}" --tag ${IMG}:${TAG} .
 
 
 # Build and push a multi-arch container image (docker buildx is required)
 .PHONY: buildx
 buildx:
 	$(info Building the KubeStellar container using Dockerfile...)
-	$(info Arguments: IMG=${IMG})
 	$(info Arguments: KUBESTELLAR_VERSION=${KUBESTELLAR_VERSION})
+	$(info Arguments: IMG=${IMG})
+	$(info Arguments: TAG=${TAG})
 	$(info Arguments: PLATFORMS=${PLATFORMS})
-	- docker buildx create --name kubestellar-builder --use
-	docker buildx build --push --platform ${PLATFORMS} --build-arg="KUBESTELLAR_VERSION=${KUBESTELLAR_VERSION}" --tag ${IMG}:${KUBESTELLAR_VERSION} .
+	- docker buildx rm kubestellar-builder
+	docker buildx create --name kubestellar-builder --use
+	docker buildx build --push --platform ${PLATFORMS} --build-arg="KUBESTELLAR_VERSION=${KUBESTELLAR_VERSION}" --tag ${IMG}:${TAG} .
 	- docker buildx rm kubestellar-builder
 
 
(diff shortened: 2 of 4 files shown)


docker/README.md
```diff
@@ -0,0 +1,149 @@
+# KubeStellar in a container
```
ronenkat, line 1:
```suggestion
# Building KubeStellar deployment as a container
```

ronenkat (commented):
Thank you @francostellari, this is awesome. 
Included some feedback.
I will try it out, and will also explore how to run it on k8s.

clubanderson:
@francostellari check to make sure boilerplate is correct - pull-kubestellar-verify is complaining - https://github.com/kubestellar/kubestellar/blob/main/hack/boilerplate/boilerplate.Dockerfile.txt

clubanderson:
@xrstf what is DCO sign-off?

docker/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+    mkdir kubestellar && \
+    tar -C kubestellar -zxf kubestellar.tar.gz
+
+# add entry script
+ADD . /
+
+###############################################################################
+# KubeStellar image
+###############################################################################
+FROM ubuntu
```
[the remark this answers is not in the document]
francostellari: Thank you for the suggestion, I have moved the Dockerfile to ubi9

francostellari:
`kubestellar/kubestellar` images are now available on quay
image

I will delete my personal images

francostellari:
I pushed a new commit to clear the 2 Security Hotpots from SonarCloud

docker/Makefile
```diff
@@ -0,0 +1,86 @@
...
+	mkdir -p ~/.kcp
+	docker run -d \
+		-p 6443:6443 \
+		-v ~/.kcp:/.kcp \
+		--user $(shell id -u) \
+		--name ${NAME} ${IMG}:${KUBESTELLAR_VERSION}
+	docker cp ${IMG}:/kcp-plugins ~/kcp
+	docker cp ${IMG}:/kubestellar ~/kubestellar
+	$(info export KUBECONFIG=${HOME}/.kcp/admin.kubeconfig)
+	$(info export PATH=$$PATH:${HOME}/kcp:${HOME}/kubestellar/bin)
```
ronenkat, line 60: is `${HOME}/kcp` really needed? as kcp is running in the container.

docker/Makefile
```diff
@@ -0,0 +1,86 @@
...
+
+
+# Run a new container image
+.PHONY: run
+run:
+	$(info Running a new container container...)
+	$(info Arguments: IMG=${IMG})
+	$(info Arguments: KUBESTELLAR_VERSION=${KUBESTELLAR_VERSION})
+	$(info Arguments: NAME=${NAME})
+	mkdir -p ~/.kcp
```
ronenkat, line 51: is it a strict requirement that the `.kcp` directory is shared between the machine that runs kubectl and the container?
What can I do to avoid it?

ronenkat (commented):
Thank you. I am trying to run on k8s instead of Docker and would like to decouple dependencies.

xrstf:
@clubanderson 

For the DCO check, IIRC, you either need to be pat of the kubestellar org or have _all_ commits in the PR be signed-off (`git commit -s`). Alternatively you can disable the dco plugin https://github.com/kcp-dev/infra/blob/main/prow/plugins.yaml entirely.

clubanderson:
Yeah. That's what I thought. Franco is part of the contributors team. He has triage on kubestellar/kubestellar. This is the same as most of our contributors. Scratching my head on this one. I'll take a closer look tomorrow.

docker/Makefile
francostellari: You don't need to mount the kcp folder to run kubestellar. However, you need access to it's KUBECONFIG somehow to do anything with it like `kubectl ws .` At the time I was thinking at the host user use case so I thought that mounting it as a volume was convenient. You can also mount it as a secret or configmap in your cluster.

Once KubeStellar is running, you need to way to intereact with it.
1, syncer yaml need to point to the correct route to access it back
2. who/how do you plan to create objects  in KubeStellar? ... will need the KUBECONFIG access

docker/Makefile
francostellari: See above

docker/Makefile
francostellari: I'm sorry I misunderstood.
I think it depends again on the use case.
The folders `${HOME}/kcp` and `${HOME}/kubestellar/bin` are used to make available to the host user the binaries for manipulating the ws, creating the syncer, etc.  Currently, I included them in the kubesterllar container image so that they can be used from within the container or by the host without the separate need of downloading them.
Perhaps I could be more clear if I understood how you plan to create access kubestellar to create a simple deployment.

For example, if you want to create a "root:abc" workspace you would need to do `kubectl ws roor` and `kuebectl ws create abc`... those two command need access to the KUBECONFIG and the workspace plugin in the kcp folder.

Perhaps @MikeSpreitzer can clarify my fumbling...

docker/Makefile
```diff
@@ -0,0 +1,86 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Global arguments
+IMG := kubestellar
+NAME := kubestellar
+PLATFORMS := linux/amd64,linux/arm64,linux/ppc64le
```
MikeSpreitzer, line 18: What about our 390?

docker/entry.sh
```diff
@@ -0,0 +1,50 @@
...
+mkdir -p /kubestellar-logs
+# chown -R $USER:$USER .kcp logs
+
+# Start kcp
+echo "< Starting kcp >-------------------------------------------"
+
+echo -n "Running kcp... "
+kcp start >& /kubestellar-logs/kcp.log &
+# kcp start --external-hostname "127.0.0.1" --bind-address "127.0.0.1" >& /kubestellar-logs/kcp.log &
+echo "pid=$(pgrep kcp) logfile=/kubestellar-logs/kcp.log"
```
MikeSpreitzer, line 30: Do you know about https://www.gnu.org/software/bash/manual/bash.html#index-_0021-1 ?

docker/entry.sh
```diff
@@ -0,0 +1,50 @@
...
+
+
+# Starting KubeStellar
+echo "< Starting KubeStellar >-----------------------------------"
+
+kubestellar start
+
+# Sleep forerver
+echo "Ready!"
+sleep infinity
```
MikeSpreitzer, line 50: Please include a trailing linebreak.

docker/Makefile
MikeSpreitzer: When a user on the host issues a command like `kubectl ws create imw` this requires kcp's kubectl plugin for `ws` to be on that user's `$PATH`.  The plugin is named `kubectl-ws` and is found in `${HOME}/kcp` (after that `cp` above).

MikeSpreitzer:
None of the targets in the Makefile benefit from being targets in a Makefile.  The main function of `make` --- conditionally invoking targets based on file presence --- is not engaged anywhere in that Makefile.  IMHO it would be better if those were shell scripts rather than a targets in a Makefile.  Shell scripts are less dependent on the flavor of `make` installed on the user's system, can more easily be invoked (both be found and be coded to work correctly) from whatever working directory the user desires (thus giving some more flexibility to the user), and have less obscure syntax & semantics.

docker/README.md
```diff
@@ -0,0 +1,149 @@
...
+  - [Build a local container image](#build-a-local-container-image)
+  - [Build and push a multi-architecture image](#build-and-push-a-multi-architecture-image)
+  - [Run the container and access KubeStellar from the host OS](#run-the-container-and-access-kubestellar-from-the-host-os)
+  - [Checking the container logs](#checking-the-container-logs)
+  - [Login into the running container to use **KubeStellar** from inside the container](#login-into-the-running-container-to-use-kubestellar-from-inside-the-container)
+  - [Teardown and cleanup](#teardown-and-cleanup)
+
+## Install the pre-requisites
+
+Install `nake`:
```
MikeSpreitzer, line 15: typo: nake instead of make

docker/README.md
```diff
@@ -0,0 +1,149 @@
...
+- [KubeStellar in a container](#kubestellar-in-a-container)
+  - [Install the pre-requisites](#install-the-pre-requisites)
+  - [Build a local container image](#build-a-local-container-image)
+  - [Build and push a multi-architecture image](#build-and-push-a-multi-architecture-image)
+  - [Run the container and access KubeStellar from the host OS](#run-the-container-and-access-kubestellar-from-the-host-os)
+  - [Checking the container logs](#checking-the-container-logs)
+  - [Login into the running container to use **KubeStellar** from inside the container](#login-into-the-running-container-to-use-kubestellar-from-inside-the-container)
+  - [Teardown and cleanup](#teardown-and-cleanup)
+
+## Install the pre-requisites
```
MikeSpreitzer, line 13: The instructions here look specific to the Debian family of Linux distros.  This specificity should be stated up front.

It would be good to separate building and pushing the image from using the image.

docker/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+Install `nake`:
+
+```bash
+sudo apt install -y make
+```
+
+Install `docker`:
```
MikeSpreitzer, line 21: Don't say "Do X" and then later tell me "you could have done Y instead".  Tell me "Do X or Y".

docker/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+```bash
+sudo mkdir -p /etc/apt/keyrings
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
+echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+sudo apt update
+sudo apt install -y docker-ce docker-ce-cli containerd.io docker-compose-plugin
+sudo usermod -aG docker $USER
+```
```
MikeSpreitzer, line 30: I am not a fan of maintaining a copy of someone else's instructions.  I think a pointer to them would suffice.

docker/README.md
```diff
@@ -0,0 +1,149 @@
+# Building KubeStellar deployment as a container
```
MikeSpreitzer, line 1: Somewhere right up front there should be a big picture explanation of that this image is for.  I think that is running the kcp server and the kubestellar controllers.

docker/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+Build a local container image `kubestellar` for the host architecture using the latest release of **KubeStellar** with the command:
+
+```bash
+make build
+```
+
+A desired version of **KubeStellar** can be specified by using the optional `KUBESTELLAR_VERSION` argument, such as `KUBESTELLAR_VERSION=v0.3.1`.
+
+The **KubeStellar** version is used to tag the container image, as shown below:
```
MikeSpreitzer, line 49: This wording leaves me uncertain whether setting the KUBESTELLAR_VERSION variable only controls the image tag or whether it also controls version of KubeStellar that goes into the image.  I think the truth is that it does both.  I think it would be better to introduce this variable with a sentence that says that it does both.

docker/README.md
```diff
@@ -0,0 +1,149 @@
...
+```bash
+make run
+export KUBECONFIG=${HOME}/.kcp/admin.kubeconfig
+export PATH=$PATH:${HOME}/kcp:${HOME}/kubestellar/bin
+```
+
+The following arguments can be used:
+
+- `IMG=[<registry>/<organization>/]<image>` select the container image, the default vakue is the local `kubestellar` image
+- `KUBESTELLAR_VERSION=[<version>|latest]`, select the image tag, the default value is the latest version of **KubeStellar**
```
MikeSpreitzer, line 91: Does `latest` really work here?  I do not see the tag `latest` being applied in the builds.

docker/README.md
```diff
@@ -0,0 +1,149 @@
...
+- `NAME=<container-name>` sets the name of the container, the default value is `kubestellar`
+
+The command above will
+
+- make port `6443` accessibe on the host OS
+- share a `~/.kcp` folder containing the `KUBECONFIG` file
+- create a `~/kcp` folder containing a copy of **kcp** plugins for the host OS architecture
+- create a `~/kubestellar` folder containing a copy of **KubeStellar** binaries for the host OS architecture
```
MikeSpreitzer, line 99: If `run` were done with a shell script rather than a Makefile target then you would not have to fix these directories in the user's home, they could be relative to the current working directory.

MikeSpreitzer, line 99: I would say "executables" rather than "binaries".

MikeSpreitzer, line 99: I would say "executable" rather than "binary".

MikeSpreitzer:
This brings in Issue #788 , since the syncers will not be able to open a connection to the kcp server's IP address.

MikeSpreitzer:
If we really intend to keep this then the information in the READE should be moved into the regular documentation.

docker/README.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+At this point, one can follow the instructions and examples in **KubeStellar** [Quickstart](https://docs.kubestellar.io/release-0.3/Getting-Started/quickstart/).
+
+## Checking the container logs
+
+The running container logs can be easily displayed using the command:
+
+```bash
+make logs
```
MikeSpreitzer, line 124: That just does `docker logs` on the container, but the interesting logs are in files in the container.

docker/Makefile
francostellari: @MikeSpreitzer I'll add the linux/s390x in the defaults

docker/entry.sh
francostellari: ok, removed

docker/entry.sh
francostellari: ok

francostellari:
> None of the targets in the Makefile benefit from being targets in a Makefile. The main function of `make` --- conditionally invoking targets based on file presence --- is not engaged anywhere in that Makefile. IMHO it would be better if those were shell scripts, or subcommands of one shell script, rather than a targets in a Makefile. Shell scripts are not dependent on the flavor of `make` installed on the user's system, can more easily be invoked (both be found and be coded to work correctly) from whatever working directory the user desires (thus giving some more flexibility to the user), and have less obscure syntax & semantics.

It's a fair comment, I started using the make/Makefile instead of a bash script for convenience of handling arguments without much coding. I will switch to bash at a later time.

docker/README.md
francostellari: ty

docker/README.md
francostellari: I worked a bit on the text

docker/README.md
francostellari: ok

docker/README.md
francostellari: granted, this is a convenience

docker/README.md
francostellari: Here, we discuss the creation and use of a **KubeStellar** container image for running the **kcp** server and the **KubeStellar** controllers, and delivering the **kcp** and **KubeStellar** executables, and kcp client kube config to the host. This provides an alternative to using **KubeStellar** bootstrap script. Additionally, we are planning to use this images in the future to run **KubeStellar** as a service in a cluster.

docker/README.md
francostellari: ok reworded

docker/README.md
francostellari: ok

docker/README.md
francostellari: I have added the `kubestellar-logs` volume

docker/Makefile
francostellari: @MikeSpreitzer sorry, I just realized that there is not a kcp tarball for s390x so I cannot create an image for it

docker/README.md
francostellari: not really cleanly, I have removed it

MikeSpreitzer:
The way that the Makefile handles arguments is to take no arguments and look in environment variables instead.  You could immediately switch to doing the same thing in a bash script.

docker/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+ARG TARGETPLATFORM
+ARG KUBESTELLAR_VERSION=v0.3.1
+
+WORKDIR /
+
+# install requirements
+RUN yum install -y wget
+
+# download kubectl binaries
+RUN wget -nv "https://dl.k8s.io/release/v1.25.3/bin/${TARGETPLATFORM}/kubectl" -O kubectl && \
```
MikeSpreitzer, line 31: Why `wget` here and `curl` later?

It looks to me like `curl` could be used:

```console
mspreitz@mjs12 docker % docker run --rm -it redhat/ubi9 sh

sh-5.1# curl -L --no-progress-meter -o /tmp/k https://dl.k8s.io/release/v1.25.3/bin/linux/amd64/kubectl

sh-5.1# ls -l /tmp/k
-rw-r--r--. 1 root root 45015040 Jul 14 20:11 /tmp/k

sh-5.1# chmod a+x /tmp/k

sh-5.1# /tmp/k version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.3", GitCommit:"434bfd82814af038ad94d62ebe59b133fcb50506", GitTreeState:"clean", BuildDate:"2022-10-12T10:57:26Z", GoVersion:"go1.19.2", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
The connection to the server localhost:8080 was refused - did you specify the right host or port?
```

docker/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+
+###############################################################################
+# Builder image
+###############################################################################
+FROM redhat/ubi9 AS builder
+
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETPLATFORM
+ARG KUBESTELLAR_VERSION=v0.3.1
```
MikeSpreitzer, line 23: The latest is v0.4.0 now

docker/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+```bash
+make exec [NAME=<name>]
+```
+
+Since `kubectl` is available inside the container, at this point **KubeStellar** can be used as before.
+
+The `NAME` argument above is only necessary if a container name was specified with a similar argument in the `make run` command.
+
+## Teardown and cleanup
```
MikeSpreitzer, line 154: I think it would be better if killing the container were separate from deleting the image.  That separation is the normal pattern in the container world.

docker/README.md
```diff
@@ -0,0 +1,168 @@
...
+The following arguments can be used:
+
+- `IMG=[<registry>/<organization>/]<image>` select the container image, the default value is the local `kubestellar` image. Use `IMG=quay.io/kubestellar/kubestellar` to pull the **stable** container image from our [quay.io registry](https://quay.io/repository/kubestellar/kubestellar?tab=tags).
+- `TAG=<tag>`, select the image tag, the default value is the **stable** version of **KubeStellar** as specified in https://github.com/kubestellar/kubestellar/blob/main/VERSION.
+- `NAME=<container-name>` sets the name of the container, the default value is `kubestellar`
+- `BASEPATH=<path>` set the base path where the **KUBECONFIG** and executable will be made available. By default, `~` is used. Note that this parameter will impact the exact path of the `export` commands above, which are provided for the default case. The `make run` command will print out the correct export commands to be used.
+
+The command above will:
+
+- share a `${BASEPATH}/.kcp` folder containing the `KUBECONFIG` file
```
MikeSpreitzer, line 109: (1) "share" is a little strange from a user's point of view, this `make run` does not cause the user to share that directory with anyone else.  I suppose that you are talking about sharing with the stuff inside the container, but from the user's point of view that is not important (unless they are also a contributor who cares about what goes on inside the container, but that is of secondary importance here).  What a user needs to know is that this command creates the directory if it does not already exist and that the kubeconfig file will appear in it.

(2) The typography of `KUBECONFIG` is potentially confusing, that could be read as a literal filename.  I would suggest writing `$KUBECONFIG` or `admin.kubeconfig` instead.

docker/entry.sh
```diff
@@ -0,0 +1,49 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+echo -e
+
+echo "< Starting Kubestellar container >-------------------------"
+
+mkdir -p /kubestellar-logs
```
MikeSpreitzer, line 21: Isn't this a job for the Dockerfile?

docker/README.md
```diff
@@ -0,0 +1,168 @@
...
+└── root
+    ├── compute
+    └── espw
+```
+
+At this point, one can follow the instructions and examples in **KubeStellar** [Quickstart](https://docs.kubestellar.io/release-0.3/Getting-Started/quickstart/).
+
+## Checking the container logs
+
+The running container logs can be easily displayed using the command:
```
MikeSpreitzer, line 134: It is true that the following command displays _some_ log from the container, but it is none of the most interesting ones.  The kcp server log and the KubeStellar controller logs are _not_ included.  But the wording here does not make that clear.  Perhaps a remark about the logs directory mentioned above is in order here.

MikeSpreitzer:
This stuff adopts the point of view of something outside of the kubestellar repo.  That is surprising, since it is actually inside the kubestellar repo.  Wouldn't it be more natural to adopt the stance of something that is part of the repo?  For example, that would mean that there is no fetching of the repo involved here, the repo has already been fetched by the time the stuff here is being used (to build, at least).

clubanderson:
at some point we can consider splitting this into it's own repo, for now we are a small team and are not staffed to take on the added responsibilities that come with splitting out.

docker/Dockerfile
MikeSpreitzer: The only reason there is a two-stage build here --- and thus the only reason that podman is not sufficient --- is the choice to use wget in place of curl here, right?  I think it would better if the build were single stage.

MikeSpreitzer:
Please do not merge `main` into development branches.  That violates our discipline for how we use branches.

MikeSpreitzer:
As @clubanderson said elsewhere, I think that for now we can keep the external stuff in a subdirectory rather than a distinct repo.

MikeSpreitzer:
I think that the container image building is something that is logically internal to this repo; it should be covered by the main Makefile and instructions for contributors.

If @francostellari wants to give users a Makefile to replace the bootstrap script then that is a design direction that he can explore, but I think it is a different Makefile than the one that contributors use to build KubeStellar.

clubanderson:
@francostellari is there anything you need to address here from @MikeSpreitzer's comments?

francostellari:
@clubanderson I think I handled @MikeSpreitzer comments older than 2 days ago.

I'm unclear what I need to do for https://github.com/kubestellar/kubestellar/pull/764#issuecomment-1649175220

Regarding https://github.com/kubestellar/kubestellar/pull/764#issuecomment-1649179465 I would say that 
1) I agree that in the long term the making of the container should be automated in the release process and the documentation moved to the docs but I would rather get this moving and work on improving it before fully advertising it since we do not a full circle with cluster deployment working
2) the make file can provide an alternative to use kubestellar locally in the container in the short term if one does not want to install it but I think this is not a long term proposition, we will move to the cluster.

clubanderson:
> Please do not merge `main` into development branches. That violates our discipline for how we use branches.

@MikeSpreitzer can you explain what you are seeing on this comment?  I see a fork from a collaborator being PR'ed against main.  The branch is name "docker". - what am I missing here?

docker/Dockerfile
clubanderson: I don't think this is something that we should hold up the PR for.  I am ok with using 2 different methods in the same file if that is what the author is comfortable with.  Lets let this pass.

docker/README.md
clubanderson: I see that "Several examples are shown below. On Debian systems, execute this command:" is included on line 17.  I think this satisfies the requirement.

docker/README.md
clubanderson: @francostellari I think I agree with @MikeSpreitzer here.  Let's be specific that the logs being served are from the container and not from kcp or kubestellar.  I think that is a good idea.

docker/README.md
clubanderson: @francostellari I agree with @MikeSpreitzer on this comment.  Let's make teardown a 2-command process.  One to kill (force stop and remove the running container) and the other to delete (remove container image from local registry and delete created folders).  That seems fair.

docker/entry.sh
clubanderson: I think this is a matter of preference.  I am ok with this living in entry and not in docker file.  It's a toss up honestly.  It can be done in either place.  Not as important 'where' it is done, but that it IS done.

docker/README.md
clubanderson: we can evolve this in the future to be a pointer to external doc.  Let's get this out first and rev the documentation later.  I am ok with this being a copy for now since it is specific to Debian.

docker/README.md
clubanderson: the text is indeed in the README on line 13.  I think that is satisfactory.  Let's move on for now.

docker/README.md
clubanderson: @francostellari you can simply replace 'share' with 'mount on the local filesystem' and I would also use "admin.kubeconfig" to remain consistent with other instruction sets we have.  Just make these 2 changes to move past this item.

docker/Dockerfile
clubanderson: @francostellari to be safe I would change this to take an external var during the "docker build" so that no matter what we do in the future, you are protected from a hard reference.  The input to the docker build command can include a hardcoded value or a reference to a local env var.

either:
docker build --build-arg KUBESTELLAR_VERSION=0.3.1

or: 
docker build --build-arg KUBESTELLAR_VERSION=${KUBESTELLAR_VERSION}

and line 23 would be 
"ARG KUBESTELLAR_VERSION=KUBESTELLAR_VERSION"

clubanderson:
@francostellari I had a deeper dive on the comments.  I see 4 items you can address now and leave the others for later.  I gave direct guidance on the items to offer you a path to adjust your files.  I think once you knock out these 4, your work here is done.  Let me know what you think.

docker/Dockerfile
MikeSpreitzer: @clubanderson : the Makefile is already doing `--build-arg KUBESTELLAR_VERSION`.  So I do not think we need a default value inside the Dockefile at all.

More importantly, however, I do not think this Dockerfile should be fetching KubeStellar.  Building this image should be like building anything else in KubeStellar.  It works with whatever the contributor already has locally, whether that was a release, a branch, or a work-in-progress not yet pushed.

MikeSpreitzer:
@clubanderson , regarding https://github.com/kubestellar/kubestellar/pull/764#issuecomment-1652729106 : 
It is a branch, not a fork, that a PR requests to be merged.
The complaint about this branch is due to the last commit in the picture here.
Screen Shot 2023-07-26 at 9 30 04 PM
That merge _from_ main _into_ this development branch ruins linearizability.

docker/Dockerfile
clubanderson: That's fair too. Just use what is local. That's another good option.

MikeSpreitzer:
@francostellari : adding a new target to the main Makefile for building this container image does not require anyone to use that target until we are ready to promote that.  If the image building were in the main Makefile rather than your external-stance one, then it would be easy and natural for the image building to use what the contributor already has and does not require your external-stance Makefile to also have some internal-stance target(s).

MikeSpreitzer:
@francostellari : I am not sure I understand your comment

> the make file can provide an alternative to use kubestellar locally in the container in the short term if one does not want to install it but I think this is not a long term proposition, we will move to the cluster.

Yes, I understand the expectation that KubeStellar will require deployment into a Kubernetes cluster.  My question is, are we going straight there or are we going to support users using just a container?  I thought the intent here is to support users using just a container.

If the intent is to support users using just a container then we will have to document how a user does that.  What they download/fetch, what commands they issue, stuff like that.  I am not saying that documentation has to be in this PR.  But I am saying that this PR implies either creating that documentation or moving straight on to requiring a hosting Kube cluster.

francostellari:
@MikeSpreitzer are you referring to the 7/19 commit "Merge branch 'kubestellar:main' into docker"?
If so, I don't know what it means since I believe I only edited the files to implement the changes.
I also don't know how to undo or solve the issue.

docker/Dockerfile
MikeSpreitzer: Remember, as long as the Dockerfile insists on fetching a release, a contributor can not test without making a release.  That is backwards.

docker/README.md
MikeSpreitzer: "mount" is a dyadic operator: it makes one thing appear in another location.  Being explicit about both is important because not everyone will know what you are assuming about the thing not mentioned.

MikeSpreitzer: Also, it is jarring that the last two entries in this list lead with creating a folder and the first two do not --- but in fact all four items in this list create a folder (if it does not already exist) on the host and this is more important to a user than anything that happens inside the container.

docker/entry.sh
```diff
@@ -0,0 +1,49 @@
...
+# chown -R $USER:$USER .kcp logs
+
+# Start kcp
+echo "< Starting kcp >-------------------------------------------"
+
+echo -n "Running kcp... "
+kcp start >& /kubestellar-logs/kcp.log &
```
MikeSpreitzer, line 28: You might want to support the scenario where the kcp server crashes and restarts, without losing the logs from before the restart.  I.e., append to the logfile here (and earlier ensure that it exissts) rather than write it from the beginning.

francostellari:
> @clubanderson , regarding [#764 (comment)](https://github.com/kubestellar/kubestellar/pull/764#issuecomment-1652729106) : It is a branch, not a fork, that a PR requests to be merged.
> 
> @francostellari: The complaint about this branch is due to the last commit in the picture here. Screen Shot 2023-07-26 at 9 30 04 PM That merge _from_ main _into_ this development branch ruins linearizability.

@MikeSpreitzer I think I was able to remove the last commit of 7/19, I think it should solve your concern as I understand it. Please confirm. Thank you.

docker/entry.sh
francostellari: ok, I changed to 
`kcp start &>> /kubestellar-logs/kcp.log &`

docker/README.md
francostellari: @clubanderson perhaps it's my shortcoming with English but the quoted text above from README.md seems to say exactly what you are saying. To avoid issues I have removed the section from the makefile and README.md

docker/README.md
francostellari: > @francostellari you can simply replace 'share' with 'mount on the local filesystem' and I would also use "admin.kubeconfig" to remain consistent with other instruction sets we have. Just make these 2 changes to move past this item.

1) I have changed `share` to `mount on the host filesystem`
2) I have changed `KUBECONFIG` to `admin-kubeconfig`

docker/Dockerfile
francostellari: @MikeSpreitzer @clubanderson I have removed the default value from the `Dockerfile`

docker/README.md
clubanderson: Thank you

PR #773, opened 2023-07-07
✨ Fix up some naming and comments in workload projector

MikeSpreitzer:
## Summary
This PR does some cleanup in the workload projector source.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/placement/workload-projector.go
+++ b/pkg/placement/workload-projector.go
@@ -797,6 +797,7 @@ func (wp *workloadProjector) syncDestinationObject(ctx context.Context, doRef de
 	return finish()
 }
 
+// Returns `retry bool`.
 func (wp *workloadProjector) syncSourceObject(ctx context.Context, soRef sourceObjectRef) bool {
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
@clubanderson : the failure here from SonarCloud is from complaints about things that this PR is not touching.  SonarCloud seems to be annoyed about the code that is already merged.  Can we please get the improvements in this PR merged without being held-up by pre-existing issues?  I would be OK with a formal Issue being opened to address what SonarCloud is complaining about, but fixing that should be separate from this PR and should not hold up this PR.

clubanderson:
@MikeSpreitzer - ok - open the issue(s) to care for the sonarcloud complaints found herein, add links to the issue(s) here in this PR, and then lets skip the sonarcloud test by overriding tide.  Let me know if this plan works for you for this PR.

clubanderson:
ok - skipping tide now
/override tide

clubanderson:
skipping tide did not work, but going into sonarcloud ui and closing the 2 issues did work.

Issue #780, opened 2023-07-09
Feature: Redesign the Space provider interface to a uniform REST interface


ezrasilvera:
For phase I , the LC manager is using a specific local instance of the provider client in order to communicate with the provider. This make the LC manager aware of the concrete implementation of the LC provider. This creates dependency between the provider implementation and the LC manager. Also this require the LC manager to configure the provider.
The idea is to convert the provider interface into a pre-define set of REST APIs, so that connecting into a provider just require regular authentication without instantiating a local instance of the specific provider client code.

clubanderson:
placing in sprint 3

clubanderson:
streaming of watch events is a current design consideration

effi-ofer:
Update Aug 24: Still WIP. This is not something that's critical to sprint 3 and we may push it to 4.

ezrasilvera:
not  relevant for ks-0.20

Issue #779, opened 2023-07-09
Feature: The multi-space (MS) client should support cross-cluster operation


ezrasilvera:
Need to support cross cluster List() and Watch() . 
Including:
#577

clubanderson:
placing in sprint 3

clubanderson:
morning to sprint 4 based on priorities.

clubanderson:
@ezrasilvera moving this into the kcp-decouple column.  Change the sprint to your liking.

clubanderson:
moving to sprint 5 per @ezrasilvera - this issue depends on inclusion or exclusion of kube-bind - will be re-evaluated in sprint 5.

clubanderson:
started but not completed.  currently not needed.  moving to backlog

ezrasilvera:
not  relevant for ks-0.20

Issue #778, opened 2023-07-09
Feature: Add Auth and user access management to the space framework 


ezrasilvera:
The initial version of the LC cluster framework and the LC client didn't address auth and user access requirements.

clubanderson:
placed in sprint 3

effi-ofer:
No update.

clubanderson:
moving to backlog per @ezrasilvera

ezrasilvera:
not  relevant for ks-0.20

PR #783, opened 2023-07-10
:sparkles: Logical cluster framework phase1

ezrasilvera:
## Summary
This PR add the logical cluster framework and the MC aware client. 
Please note, this is phase1 of the LC framework which includes the basic functionality   working + e2e testing
There are open issues that cover the additional work expected to be done after this PR is merged, and will be included in phase 2 of the framework. For example:
#778 , #779 #780 

## Related issue(s)

Fixes #


MikeSpreitzer:
It would be great to add design documentation.  That does not have to go in this PR, but the sooner the better.

cmd/logical-cluster-manager/main.go
```diff
@@ -0,0 +1,97 @@
...
+
+const managerClusterContext string = "kind-mgt"
+
+func main() {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	logger := klog.FromContext(ctx)
+
+	var kubeconfig *string
+	if home := homedir.HomeDir(); home != "" {
```
MikeSpreitzer, line 49: Not a blocker.  It would be more DRY if the `if` statement only covered construction of the default value.

cmd/logical-cluster-manager/main.go
```diff
@@ -0,0 +1,97 @@
...
+	flag.Parse()
+
+	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: *kubeconfig},
+		&clientcmd.ConfigOverrides{CurrentContext: managerClusterContext}).ClientConfig()
+	if err != nil {
+		logger.Error(err, "Error")
+		klog.FlushAndExit(klog.ExitFlushTimeout, 1)
+	}
+
```
MikeSpreitzer, line 63: FYI: it is customary to put something informative in the config's UserAgent.  See https://github.com/kubestellar/kubestellar/blob/v0.4.0/pkg/syncer/syncer.go#L63 for an example.

cmd/logical-cluster-manager/main.go
```diff
@@ -0,0 +1,97 @@
...
+	informerProvider := informerFactory.Logicalcluster().V1alpha1().ClusterProviderDescs().Informer()
+
+	controller := clustermanager.NewController(
+		ctx,
+		clientset,
+		k8sClientset,
+		informerLC,
+		informerProvider,
+	)
+	if err != nil {
```
MikeSpreitzer, line 87: This `err` handling is out of place, the previous statement does not produce an `err`.

cmd/logical-cluster-manager/main.go
```diff
@@ -0,0 +1,97 @@
...
+		logger.Error(err, "failed to create controller")
+		klog.FlushAndExit(klog.ExitFlushTimeout, 1)
+	}
+
+	doneCh := ctx.Done()
+	informerFactory.Start(doneCh)
+	informerFactory.WaitForCacheSync(doneCh)
+
+	controller.Run(numThreads)
```
MikeSpreitzer, line 96: You probably want to pass `doneCh` to the controller too.

cmd/mc-client-test/lcdemo.sh
```diff
@@ -0,0 +1,142 @@
...
+clear
+
+TYPE_SPEED=20
+#DEMO_PROMPT="\[\033[01;32m\]\u@\h\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ "
+DEMO_PROMPT="\[\033[01;32m\]\u@WSL2\[\033[00m\]:\[\033[01;34m\]\[\033[00m\]\$"
+
+MGT_NAME="mgt"
+MGT_CTX="kind-$MGT_NAME"
+
+# For live demo use 'pe'  for recording or auto demo use 'pei'
```
MikeSpreitzer, line 27: Not a blocker: a more evolved script would take a parameter indicating which mode to operate in, rather than require editing the script.

pkg/clustermanager/controller.go
```diff
@@ -0,0 +1,188 @@
...
+	queue                   workqueue.RateLimitingInterface
+	logicalClusterInformer  cache.SharedIndexInformer
+	clusterProviderInformer cache.SharedIndexInformer
+	providers               map[string]*provider
+	lock                    sync.Mutex
+}
+
+// NewController returns logicalcluster-manager controller
+func NewController(
+	context context.Context,
```
MikeSpreitzer, line 67: Not a blocker: naming a parameter `context` shadows the imported package `context`.  I commonly see the parameter named `ctx` instead.

pkg/clustermanager/controller.go
```diff
@@ -0,0 +1,188 @@
...
+	if quit {
+		return false
+	}
+	item := i.(queueItem)
+
+	// Done with this key, unblock other workers.
+	defer c.queue.Done(i)
+
+	if err := c.process(item); err != nil {
+		runtime.HandleError(err)
```
MikeSpreitzer, line 170: `runtime.HandleError` is for un-handled errors.  I see this error being handled in the next line.

pkg/clustermanager/provider.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	return pClient
+}
+
+// CreateProvider returns new provider client
+func CreateProvider(c *controller, providerName string, providerType lcv1alpha1apis.ClusterProviderType) (*provider, error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	discoveryPrefix := "*"
```
MikeSpreitzer, line 76: Wouldn't it be simpler to use the empty string as the default?

pkg/clustermanager/provider.go
```diff
@@ -0,0 +1,231 @@
...
+				// No corresponding Logicalcluster, let's create it
+				logger.V(2).Info("Creating new LogicalCluster object", "cluster", event.Name)
+				lcluster := lcv1alpha1apis.LogicalCluster{}
+				lcluster.Name = lcName
+				lcluster.Spec.ClusterProviderDescName = p.name
+				lcluster.Spec.Managed = false
+				lcluster.Status.ClusterConfig = event.LCInfo.Config
+				lcluster.Status.Phase = lcv1alpha1apis.LogicalClusterPhaseReady
+				_, err := p.c.clientset.LogicalclusterV1alpha1().LogicalClusters(p.nameSpace).Create(ctx, &lcluster, v1.CreateOptions{})
+				chkErrAndReturn(logger, err, "Detected New cluster. Couldn't create the corresponding LogicalCluster", "cluster name", lcName)
```
MikeSpreitzer, line 184: Rather than forget about the error, would it be better to use the controller pattern here, so that the cluster reference could be requeued for a later retry?

pkg/mcclient/controller.go
```diff
@@ -0,0 +1,182 @@
...
+	// Only clusters in ready state are cached.
+	// We will get another event when the cluster is Ready and then we cache it.
+	if clusterInfo.Status.Phase != lcv1alpha1.LogicalClusterPhaseReady {
+		c.handleDelete(clusterKey)
+		return
+	}
+
+	config, err := clientcmd.RESTConfigFromKubeConfig([]byte(clusterInfo.Status.ClusterConfig))
+	if err != nil {
+		runtime.HandleError(err)
```
MikeSpreitzer, line 158: FYI: `runtime.HandleError` may or may not panic, depending on server configuration.  You should only handle an error this way if you are willing for that error to crash the server.

MikeSpreitzer (commented):
Since this is new work with no current dependents, I am going to accept it.  I left some comments inline, they can be addressed in follow-up.  I was only able to do limited review, I may have further comments later.
/lgtm
/approve

cmd/mc-client-test/lcdemo.sh
effi-ofer: We are in the process of changing mc-client-test to be more generic (as well as create a more general test script).  This will be pushed via a dedicated PR.

pkg/mcclient/controller.go
effi-ofer: @MikeSpreitzer I see we are using HandleError() in the syncer and where-resolver as well.  In addition, the default behaviour of HandleError is to log the error.  Ping me please if you think we should dig deeper here.

cmd/logical-cluster-manager/main.go
effi-ofer: Handled by PR #816

pkg/clustermanager/controller.go
effi-ofer: Handled by PR #816

pkg/clustermanager/controller.go
effi-ofer: Handled by PR #816

pkg/clustermanager/provider.go
effi-ofer: Handled by PR #816

Issue #524, opened 2023-06-01
EPIC:  Use KubeStellar without KCP core  Milestone1


ezrasilvera:
This epic tracks all tasks needed for milestone1 of decoupling KCP and KS.

# Milestone1 goals
- Demonstrate fully functional KS while the KCP server is not running, 
- KS code doesn't use the KCP k8s fork (directly or indirectly).  - TBD if optional or mandatory
  - Stop using KCP's code generators, 
  - Don't use KCP runtime 

## Open issues
- Do we want to continue supporting KCP export/bind after the decoupling (using some abstraction that will allow supporting both kube-bind & KCP bind)
  - Only impact the bind/export performance. We can have a Space provided by a KCP provider while using kube-bind
  - Adding a complexity level + additional (significant?) work 

## Main areas:
- Use alternatives to KCP workspaces 
- Use alternatives to KCP's implementation of APIExport/APIBind, 
  - Use alternatives for KCP's APIExport Views 
- Be able to denature base k8s resources for providers that don't denature resources (e.g., for KIND space provider)
- Don't use code based on the k8s fork of KCP (e.g., generated go-client code)

#### Additional tasks
- Update KS architecture documentation to include the Space framework, new bind functions, etc..
- Update the KS user documentation 
- Update KS setup and configuration
- Update KS testing

# List of Issues & PRs for milestone1
## Stop using KCP's workspaces
- [x] #849 
- [x]  Merge the space framework and multi-space client library into KS
- [x] #931
- [x] #977
  - [x] #989
  - [x] place-holder:  Convert the space-framework into a go module
- [x] #1012
  - [x] #1013 
- [x] #976 
  - [x] #1015 
- [x] #852 
  - [ ] #1218 
- [x] #853 
  - [ ] #904 
  - [ ] #923 
#### Delivery description & Impact on KubeStellar
The items listed above don't add new functionality to KS. KS will still need to use KCP server in order to support bind & export. The main reasons to integrate the above changes are:
1. Avoid maintaining a side branch till all the needed code for milestone1 is ready (which can take couple of months)
2. Minimize the potential impact on KS by gradually integrating the needed components (instead of a single huge transition phase) 

## Use kube-bind 
- [x] #525 
  - [x] https://github.com/kube-bind/kube-bind/pull/165
  - [x] https://github.com/kube-bind/kube-bind/pull/166
  - [x] https://github.com/kube-bind/kube-bind/pull/174
[...]

PR #787, opened 2023-07-10
✨ Make where-resolver work for both kcp and plain Kubernetes

waltforme:
This PR enables the where-resolver to run on top of both kcp (by default) and plain Kubernetes, controlled by a switch.

It is verified in this PR, that the kcp fork of client-go is a superset of the original client-go, in the sense that a subset of the fork can interact with a regular Kubernetes apiserver, to watch and CRUD KubeStellar's API objects.

This PR also removes the obsolete code which watches the Locations and SyncTargets views that are exposed by APIExports "scheduling.kcp.io" and "workload.kcp.io". We are currently using APIExport "edge.kubestellar.io" for Locations and SyncTargets.

## Related issue(s)

#524

### DIFF
--- a/cmd/kubestellar-where-resolver/cmd/where-resolver-cmd.go
+++ b/cmd/kubestellar-where-resolver/cmd/where-resolver-cmd.go
@@ -31,9 +31,7 @@ import (
 
 	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
 	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
-	kcpclientsetnoncluster "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
-	kcpclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned/cluster"
-	kcpinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+	kcpclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 	kcpfeatures "github.com/kcp-dev/kcp/pkg/features"
 
 	resolveroptions "github.com/kubestellar/kubestellar/cmd/kubestellar-where-resolver/options"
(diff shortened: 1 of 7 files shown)


MikeSpreitzer:
1. In the plain kube case, did you create the CRDs in the kube cluster before launching the where resolver?  If not then that would explain the error message from the reflector code.
2. I see no difference in this PR in how the `options.BaseClientOpts` is used.  All those calls like `c.edgeClusterClient.EdgeV1alpha1().SinglePlacementSlices().Cluster(ws.Path())` will make URLs that do not work for a plain kube cluster.

waltforme:
Yes, the CRDs are there. Can you @MikeSpreitzer  talk about more about what RBAC objects are needed?

I still think the root cause is from the kcp-dev's fork of client-go. Let's look at the more verbose log (maybe I could paste it earlier):
```
I0724 11:11:33.683713   48366 round_trippers.go:553] GET https://<hidden>:6443/clusters/%2A/apis/edge.kcp.io/v1alpha1/edgeplacements?limit=500&resourceVersion=0 404 Not Found in 23 milliseconds
W0724 11:11:33.683983   48366 reflector.go:376] k8s.io/client-go@v0.0.0-20230210192259-aaa28aa88b2d/tools/cache/reflector.go:215: failed to list *v1alpha1.EdgePlacement: the server could not find the requested resource (get edgeplacements.edge.kcp.io)
```
We can see that 1) the pseudo-version number of client-go matches kcp v0.11.0, 2) `clusters` in the path of the URL.

Now we want to switch to the plain client-go, but there is problem which is the collision of module path, documented here a few weeks ago: https://github.com/kubestellar/kubestellar/discussions/776

Because of this collision, I realized that it is tricky (if not impossible) to support plain kube and kcp simultaneously as two logical cluster providers. **And there should be a decision, on how we proceed with the client-go libraries - which I don't think I can make by myself.** So, I didn't change too much of the code in this PR. This PR is created to open discussions and solicit suggestions.

MikeSpreitzer:
@waltforme : item 2 in https://github.com/kubestellar/kubestellar/pull/787#issuecomment-1647052538 points out where the current code is getting into trouble.

I actually think that it is not impossible to use kcp client stubs with plain kubernetes, but the code doing so has to not use the kcp client stubs that are designed for all-cluster access.  Recall that the kcp stub generator generates two sets of stubs: one for all-cluster access and one for accessing one particular cluster.  The latter are equivalent to what the plain kube generator generates.

clubanderson:
just note the parallel work on issue #853

waltforme:
Force-push to https://github.com/kubestellar/kubestellar/commit/3845f7469194c2e2c1a42107c9dea5af45c5c7b1 was a rebase on main.

waltforme:
Force-push to https://github.com/kubestellar/kubestellar/commit/73797b1e4183d8cc03632ec95aeeda0d2912fb34 was a rebase on main (and squashes) after the recently merged API group change.

waltforme:
Keeping this branch for future reference since work pioneered here on using the single-cluster clients should be adopted into the other PRs.

PR #797, opened 2023-07-13
✨ Bump latest version in VERSION to 0.4.0

francostellari:
## Summary

Bump `latest` version in VERSION to v0.4.0
Stay `stable` version at v0.3.1

## Related issue(s)

Fixes #

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,2 +1,2 @@
 stable=v0.3.1
-latest=v0.3.1
+latest=v0.4.0


clubanderson:
and, btw, what is this file used for?

MikeSpreitzer:
@clubanderson: we have discussed using this file in place of parsing the HTML that GitHub chooses to serve up when someone asks it about the latest release.  This usage is starting in #764

MikeSpreitzer (approved):
/lgtm
/approve

clubanderson:
@MikeSpreitzer @francostellari why is there 'stable' and 'latest'?  Our 'main' branch is used as 'latest' but there is no GH release that points at 'main'.  The 'latest' and 'stable' are always going to be the most recent release.  So why do we need 2 pointers in this file?

Issue #632, opened 2023-06-19
Feedback on website - Documentation Improvements


ronenkat:
1. The website can have more motivation and use-cases. For example, the goals in the website are collaborative goals (how we are building Kubnestellar), and not Kubestellar (as a software) goals.
1. There is an assumption of understanding and knowing prior terms and definitions such as KCP, workspaces, etc. Would. be good to either have a reference or documentation.
1. The quick-start is very technical and "do that, do this" with no explanation of what is being done, how, and how it reflects in Kubestellar. The extended example provides more context to each stage.
   * more information on the context of the YAMLs would be good, for example `EdgePlacement.yaml` 
   * In step (d) the reference to `edge-placement-c` in the text seems wrong or not explained well.
1. A primer on the underlaying concepts of Kubestellar would be nice to read.

clubanderson:
@ronenkat thank you for your feedback.  We had similar feedback from @aslom re: having a glossary.  Maybe it is time to take this as a priority.

clubanderson:
@amanroa can you please take a closer look at this issue and see if you can come up with plan to tackle the items Ronen has listed?

MikeSpreitzer:
Regarding the current home page, https://github.com/kcp-dev/edge-mc/blob/f4d99eee6d9030b5b7a6951da0c3504657abb6c7/docs/content/index.md :

- I think modularity is critical, and since we seem to be intending the term "KubeStellar" to cover multiple user-distinguishable modules of functionality, that should be crystal clear right up front.
- I currently see two modules: the original multi-cluster configuration management module, and the logical cluster abstraction layer.
- The current home page starts with a listing of features that has features drawn from both modules without distinction.  I think this needs to be revised so that the top-level structure is a list of modules, and inside the section for each module can be a list of most exciting features and other important features.

amanroa:
@ronenkat could you elaborate further on part 1 of your feedback? Specifically, what is the difference between software goals vs collaborative goals?

amanroa:
Because I'm not sure about part 1, I'll provide feedback on parts 2-4. I also have some of my own feedback regarding the webpage. 

2.  Adding a glossary page with terms + succinct definitions would be great for new users to get background knowledge (I really support this idea). This could also be a way to solve part 4, by adding a section before/after the terms and definitions describing the concepts that Kubestellar is based on.

[...]

PR #807, opened 2023-07-20
📖 : make base glossary

amanroa:
## Summary
Add a quick glossary to the user guide. Some definitions are missing, and some need more elaboration (because I think others can provide a much better definition than I can).
## Related issue(s)
#632


MikeSpreitzer (approved):
Thanks for making a start on this!
/lgtm
/approve

PR #814, opened 2023-07-21
:sparkles: feature Remove dependency on KCP SyncTarget and Location resources

fileppb:
Create Kubestellar SyncTarget and Location resources and remove references to KCP's SyncTarget and Location resources

## Related issue(s)
Issue 559
Issue 560

Fixes #


cmd/mailbox-controller/options/options.go
```diff
@@ -0,0 +1,63 @@
...
+	"k8s.io/component-base/logs"
+
+	clientoptions "github.com/kubestellar/kubestellar/pkg/client-options"
+)
+
+type Options struct {
+	EspwClientOpts clientoptions.ClientOpts
+	RootClientOpts clientoptions.ClientOpts
+	BaseClientOpts clientoptions.ClientOpts
+	Logs           *logs.Options
```
MikeSpreitzer, line 32: These are not used, right?

cmd/mailbox-controller/main.go
```diff
@@ -108,23 +111,28 @@ func main() {
...
+	edgeViewClusterClientset, err := edgeclientset.NewForConfig(edgeViewConfig)
 	if err != nil {
-		logger.Error(err, "Failed to create clientset for view of SyncTarget exports")
+		logger.Error(err, "failed to create clientset for view of edge exports")
 		os.Exit(6)
 	}
-
-	stViewInformerFactory := kcpinformers.NewSharedInformerFactoryWithOptions(stViewClusterClientset, resyncPeriod)
-	syncTargetClusterPreInformer := stViewInformerFactory.Workload().V1alpha1().SyncTargets()
+	edgeSharedInformerFactory := edgeinformers.NewSharedInformerFactoryWithOptions(edgeViewClusterClientset, resyncPeriod)
```
MikeSpreitzer, line 131: Not a blocker: It wouldn't hurt to keep "View" in the name of this variable.

cmd/mailbox-controller/main.go
```diff
@@ -186,6 +195,7 @@ func configForViewOfExport(ctx context.Context, providerConfig *rest.Config, exp
 			return nil, fmt.Errorf("error reading APIExport %s: %w", exportName, err)
 		}
 		if isAPIExportReady(logger, apiExport) {
+			logger.Info(" ### export is ready, exportName: ", exportName)
```
MikeSpreitzer, line 198: It would be better if this were at a numerically higher `V` level: at least 2, to match the complaint above.

scripts/kubectl-kubestellar-ensure-location
```diff
@@ -108,13 +108,25 @@ fi
 
 set -e
 
+echo "--- current directory is $PWD"
+cwsi=$(kubectl ws "${kubectl_flags[@]}" .)
+cwsi=${cwsi#*'"'}; cwsi=${cwsi%'"'*}
```
MikeSpreitzer, line 113: Isn't there a `--short` flag that makes this unnecessary?

scripts/kubectl-kubestellar-ensure-location
```diff
@@ -108,13 +108,25 @@ fi
...
+echo "current ws is $cwsi"
+echo "imw is $imw"
 if [ "$imw" != "." ]
 then kubectl ws "${kubectl_flags[@]}" "$imw"
+     cwsi=$imw
+fi
+echo "current cwsi is $cwsi"
+if ! kubectl get apibinding "edge.kcp.io" &> /dev/null; then
+   kubectl kcp bind apiexport root:espw:edge.kcp.io
+   echo "bound apiexport root:espw:edge.kcp.io to $cwsi:edge.kcp.io"
```
MikeSpreitzer, line 123: I would say "in" rather than "to" here.

scripts/kubectl-kubestellar-ensure-location
```diff
@@ -108,13 +108,25 @@ fi
...
+cwsi=${cwsi#*'"'}; cwsi=${cwsi%'"'*}
+echo "current ws is $cwsi"
+echo "imw is $imw"
 if [ "$imw" != "." ]
 then kubectl ws "${kubectl_flags[@]}" "$imw"
+     cwsi=$imw
+fi
+echo "current cwsi is $cwsi"
+if ! kubectl get apibinding "edge.kcp.io" &> /dev/null; then
+   kubectl kcp bind apiexport root:espw:edge.kcp.io
```
MikeSpreitzer, line 122: FYI: In the other scripts that depend on the ESPW, the workspace pathname of the ESPW is an optional command-line flag.

scripts/kubectl-kubestellar-ensure-location
```diff
@@ -159,16 +171,16 @@ while (( index < ${#labels[*]} )); do
     let index=index+2
     stval=$(jq -r <<<"$stlabs" ".[\"${key}\"]")
     if [ "$stval" != "$val" ]; then
-	if ! kubectl "${kubectl_flags[@]}" label --overwrite SyncTarget "$objname" "${key}=${val}" ; then
+	   if ! kubectl "${kubectl_flags[@]}" label --overwrite synctargets.edge.kcp.io "$objname" "${key}=${val}" ; then
```
MikeSpreitzer, line 174: Let's stick with four spaces per level of indentation.

scripts/kubectl-kubestellar-ensure-location
MikeSpreitzer: Is `cwsi` only used in `echo "current cwsi is $cwsi"`?  If so then wouldn't it be simpler to do `kubectl ws "${kubectl_flags[@]}" .` there and delete all the messing with `cwsi`?

scripts/kubectl-kubestellar-ensure-location
```diff
@@ -108,13 +108,25 @@ fi
 
 set -e
 
+echo "--- current directory is $PWD"
+cwsi=$(kubectl ws "${kubectl_flags[@]}" .)
```
MikeSpreitzer, line 112: What does "cwsi" stand for?

fileppb: current workspace "cws" I don't remember what "i" was for, if anything.

scripts/kubectl-kubestellar-ensure-location
fileppb: I guess "bind into" would make sense :--)

scripts/kubectl-kubestellar-ensure-location
fileppb: I don't know. Will research.

scripts/kubectl-kubestellar-prep_for_syncer
```diff
@@ -118,7 +118,20 @@ trap 'kubectl ws "${kubectl_flags[@]}" "$original_ws"' EXIT
...
+cwsi=$(kubectl ws "${kubectl_flags[@]}" .)
+cwsi=${cwsi#*'"'}; cwsi=${cwsi%'"'*}
+echo "current ws is $cwsi"
+echo "imw is $imw"
+
+echo "current cwsi is $cwsi"
+if ! kubectl get apibinding "edge.kcp.io" &> /dev/null; then
+   kubectl kcp bind apiexport root:espw:edge.kcp.io
+   echo "bound apiexport root:espw:edge.kcp.io to $cwsi:edge.kcp.io"
+else echo "edge.kcp.io apibinding exists in workspace $cwsi"
```
MikeSpreitzer, line 131: This is not something we want.  This script should be used after the IS (née IMW) and KCS (née ESPW) have already been constructed, and it is positively good that there is something that can be checked for in those spaces --- see the check for an APIExport below as an example.

cmd/mailbox-controller/main.go
fileppb: I don't know the go logger, and am not familiar with the loglevel conventions. Will change it.

cmd/mailbox-controller/options/options.go
fileppb: Will look

scripts/kubectl-kubestellar-remove-location
```diff
@@ -69,14 +69,26 @@ fi
...
+cwsi=${cwsi#*'"'}; cwsi=${cwsi%'"'*}
+echo "current ws is $cwsi"
+echo "imw is $imw"
 if [ "$imw" != "." ]
 then kubectl ws "${kubectl_flags[@]}" "$imw"
+     cwsi=$imw
+fi
+echo "current cwsi is $cwsi"
+if ! kubectl get apibinding "edge.kcp.io" &> /dev/null; then
+   kubectl kcp bind apiexport root:espw:edge.kcp.io
```
MikeSpreitzer, line 83: Since this script is for ensuring the absence of an object, I do not think we want to ensure the presence of an APIBinding here.

MikeSpreitzer (approved):
This is important progress.
I left some comments inline that can be addressed in follow-up.
Thanks!

/lgtm
/approve

scripts/kubectl-kubestellar-ensure-location
```diff
@@ -108,13 +108,25 @@ fi
...
+cwsi=${cwsi#*'"'}; cwsi=${cwsi%'"'*}
+echo "current ws is $cwsi"
+echo "imw is $imw"
 if [ "$imw" != "." ]
 then kubectl ws "${kubectl_flags[@]}" "$imw"
+     cwsi=$imw
+fi
+echo "current cwsi is $cwsi"
+if ! kubectl get apibinding "edge.kcp.io" &> /dev/null; then
```
MikeSpreitzer, line 121: The `"${kubectl_flags[@]}"` was omitted in this and the following command.

MikeSpreitzer:
@fileppb : as you noted earlier, moving the _relevant_ SyncTarget and Location exports from the `root` workspace to the `root:espw` workspace adds an ordering constraint in the usage.  https://github.com/fileppb/edge-mc/blob/0e6e3f90c12020bcba9cc409c2278f8a4e8c6108/docs/content/Coding%20Milestones/PoC2023q1/example1-subs/example1-post-kcp.md?plain=1#L8 is now out of order because it depends on the later line https://github.com/fileppb/edge-mc/blob/0e6e3f90c12020bcba9cc409c2278f8a4e8c6108/docs/content/Coding%20Milestones/PoC2023q1/example1-subs/example1-post-kcp.md?plain=1#L110 .  And there are other mis-orderings too.

MikeSpreitzer:
I do not understand how https://github.com/kubestellar/kubestellar/actions/runs/5626329356 succeeded.  When I test the same commit (0e6e3f90c12020bcba9cc409c2278f8a4e8c6108) myself, it fails due to that mis-ordering.  Here is what I see after the attempt to use the not-yet-existing APIExport fails:
```
(venv) %n@%m %1~ %# kubectl ws root
Current workspace is "root".
(venv) %n@%m %1~ %# kubectl ws tree
.
└── root
    ├── compute
    └── imw-1

```

MikeSpreitzer:
If I manually issue the `kubestellar init` command and then switch back to the IMW, then the `ensure location` succeeds.

fileppb:
@MikeSpreitzer Hi Mike, is there something that I need to revise in this code?

MikeSpreitzer:
@fileppb : what needs revisions is the documents that tell users to `kubectl kubestellar ensure location` before `kubestellar init`.

fileppb:
Silly documents. How could you not do init before start and everything else, especially since that's in the quick start? But ok this is a new issue: bugfix 817, so needs new branch etc. Can this wait a day?

MikeSpreitzer:
Before your move of the definitions of SyncTarget and Location, some of the commands did not rely on the ESPW because the relevant APIExport was put in the `root` workspace by `kcp start`.

Also: #817 is not about any bugs in what you did, it is about how the testing did not reveal the problems.

scripts/kubectl-kubestellar-ensure-location
MikeSpreitzer: This happened in other scripts too.

scripts/kubectl-kubestellar-remove-location
fileppb: @MikeSpreitzer Is it possible to create synctargets.edge.kcp.io in a workspace without have created the APIBinding? And isn't possible to delete the APIBinding without deleting synctargets.edge.kcp.io? So then if you don't perform the binding you might have orphaned synctargets?

MikeSpreitzer: It is impossible to create an object of a kind that is not defined.  I do not know whether it is possible to delete an APIBinding while objects of kinds defined by that APIBinding exist.  If someone did manage to remove the APIBinding while leaving such objects in storage, those objects would still not be accessible through to apiserver --- so that are effectively gone.  I do not think we need to worry here about provider bit tweezers for someone who manages to cause exceptional damage to their space.

fileppb: well I know it's possible to delete an APIBinding and still have orphaned objects in storage...because I did it while testing :--/ So the tweezer was for myself. But I'll remove it.

PR #830, opened 2023-07-27
✨ Draft scripts to support external processes

MikeSpreitzer:
## Summary
This PR adds a first draft of scripts designed to support more complex deployment scenarios, as outlined in #788.  These scripts are concerned with the following two things.

1. Support generation of server credentials to pass to kcp or plain kube server via `--tls-sni-cert-key`.
2. Create alternate kubeconfig files by replacing the protocol://host:port prefix of cluster server URLs and the CA certificate data.

These scripts are factored so that the user can launch the kcp server in their own way and support multiple sets of clients needing to connect to the kcp server using different domain names.

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/Coding Milestones/PoC2023q1/commands.md
+++ b/docs/content/Coding Milestones/PoC2023q1/commands.md
@@ -34,6 +34,22 @@ name.
 
 ### kubestellar-ensure-kcp-server-creds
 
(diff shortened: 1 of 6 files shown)


francostellari:
@MikeSpreitzer thank you very much, great job!

PR #841, opened 2023-07-30
:sparkles:Use provider config field for provider client

andreyod:
## Summary
This will allow provider client to access the provider by using config from provider description object.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/clustermanager/provider.go
+++ b/pkg/clustermanager/provider.go
@@ -58,13 +58,13 @@ type provider struct {
 }
 
 // TODO: this is termporary for stage 1. For stage 2 we expect to have a uniform interface for all informers.
-func newProviderClient(providerName string, providerType lcv1alpha1apis.ClusterProviderType, config string) clusterprovider.ProviderClient {
+func newProviderClient(providerDesc *lcv1alpha1apis.ClusterProviderDesc) clusterprovider.ProviderClient {
 	var pClient clusterprovider.ProviderClient = nil
-	switch providerType {
+	switch providerDesc.Spec.ProviderType {
 	case lcv1alpha1apis.KindProviderType:
-		pClient = kindprovider.New(providerName)
+		pClient = kindprovider.New(providerDesc.Name)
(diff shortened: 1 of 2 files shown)


pkg/clustermanager/provider.go
```diff
@@ -58,7 +58,7 @@ type provider struct {
 }
 
 // TODO: this is termporary for stage 1. For stage 2 we expect to have a uniform interface for all informers.
-func newProviderClient(providerName string, providerType lcv1alpha1apis.ClusterProviderType) clusterprovider.ProviderClient {
+func newProviderClient(providerName string, providerType lcv1alpha1apis.ClusterProviderType, config string) clusterprovider.ProviderClient {
```
ezrasilvera, line 61: why not passing just the desc, it includes all other fields, right ?

ezrasilvera:
Why don't you pass the Desc into the New() of all providers ?  This will allow you to use it in the providers that needs it.

pkg/clustermanager/provider.go
andreyod: Changed

pkg/clustermanager/provider.go
```diff
@@ -58,46 +58,39 @@ type provider struct {
...
 
 // TODO: this is termporary for stage 1. For stage 2 we expect to have a uniform interface for all informers.
-func newProviderClient(providerName string, providerType lcv1alpha1apis.ClusterProviderType) clusterprovider.ProviderClient {
+func newProviderClient(providerDesc *lcv1alpha1apis.ClusterProviderDesc) clusterprovider.ProviderClient {
 	var pClient clusterprovider.ProviderClient = nil
-	switch providerType {
+	switch providerDesc.Spec.ProviderType {
 	case lcv1alpha1apis.KindProviderType:
-		pClient = kindprovider.New(providerName)
+		pClient = kindprovider.New(providerDesc.Name)
```
ezrasilvera, line 65: but you still pass only the name. You need also the config, don't you ?

PR #849, opened 2023-08-01
:sparkles:Add space provider for KCP

andreyod:
## Summary
Space provider for KCP runtime

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/space-manager/provider.go
+++ b/pkg/space-manager/provider.go
@@ -19,6 +19,7 @@ package spacemanager
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/go-logr/logr"
@@ -30,6 +31,7 @@ import (
 
 	spacev1alpha1apis "github.com/kubestellar/kubestellar/pkg/apis/space/v1alpha1"
 	spaceprovider "github.com/kubestellar/kubestellar/pkg/space-manager/providerclient"
+	providerkcp "github.com/kubestellar/kubestellar/space-provider/kcp"
 	kindprovider "github.com/kubestellar/kubestellar/space-provider/kind"
 	kflexprovider "github.com/kubestellar/kubestellar/space-provider/kubeflex"
 )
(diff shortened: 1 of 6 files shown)


pkg/space-manager/provider.go
```diff
@@ -65,6 +67,13 @@ func newProviderClient(providerDesc *spacev1alpha1apis.SpaceProviderDesc) spacep
 		pClient = kindprovider.New(providerDesc.Name)
 	case spacev1alpha1apis.KubeflexProviderType:
 		pClient = kflexprovider.New(providerDesc.Name)
+	case spacev1alpha1apis.KcpProviderType:
+		pClient, err := providerkcp.New(providerDesc.Spec.Config)
```
ezrasilvera, line 71: this is strange that only one returns error. Let's change all providers to return err and move the error handling outside the case?

space-provider/kcp/kcp.go
```diff
@@ -0,0 +1,244 @@
...
+	k.logger.V(2).Info("Created KCP workspace for space", "name", name)
+	return nil
+}
+
+func (k *KcpClusterProvider) Delete(name string, opts clusterprovider.Options) error {
+	k.logger.V(2).Info("Deleting KCP workspace", "name", name)
+	return k.clientset.Delete(k.ctx, name, metav1.DeleteOptions{})
+}
+
+// ListSpacesNames is N/A for KCP
```
effi-ofer, line 93: I think it's redundant for the other providers as well?

pkg/space-manager/provider.go
andreyod: But other providers doesn't have any error case. New(...) is not part of an interface. The parameters are different and the return values are different.

space-provider/kcp/kcp.go
andreyod: I agree. Let's remove this func from the interface in separate PR.

pkg/space-manager/provider.go
ezrasilvera: Let's add an issue to try and unify this. It may be part of the REST change

PR #851, opened 2023-08-01
🐛 Make server cert filename depend on domain name

MikeSpreitzer:
## Summary
This PR updates `scripts/kubestellar-make-kcp-server-cert` to make the name of the file that holds the server certificate depend on the domain name in the certificate, so that certs for different domain names can coexist at the same time.

## Related issue(s)

Fixes #


MikeSpreitzer:
/cc @francostellari 
@dumb0002

PR #856, opened 2023-08-03
:sparkles: Add client install script for kubestellar binaries.

ronenkat:
## Summary
When running Kubestellar non-locally you need to install kubestellar client binaries for `kubectl` to work.
Getting the binaries from the running Kubestellar may not always work. For instance when the OS and architecture of the client machine is different than the Kubestellar running environment.

The script is modeled after the kubestellar boostrap script and uses the Kubestellar distribution to get the relevant binaries.
## Related issue(s)

Fixes #


clubanderson:
@ronenkat can we close this in favor of #860 ?

ronenkat:
Replaced by #860

PR #862, opened 2023-08-04
🐛 Fix file name issues in kubestellar-ensure-kcp-server-creds

MikeSpreitzer:
## Summary
This PR changes the `kubestellar-ensure-kcp-server-creds` and `kubestellar-make-kcp-server-cert` scripts to work around the fact that EasyRSA does not want to work with long filenames.  It also fixes a bug in the idempotency logic in `kubestellar-ensure-kcp-server-creds`.

## Related issue(s)

Fixes #

### DIFF
--- a/scripts/kubestellar-ensure-kcp-server-creds
+++ b/scripts/kubestellar-ensure-kcp-server-creds
@@ -22,6 +22,7 @@ if [ $# != 1 ]; then
 fi
 
 set -e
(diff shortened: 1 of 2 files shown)


francostellari:
@MikeSpreitzer thank you for the quick fix. I tested on my IBM cloud deployment and it seem to work.
/lgtm

PR #864, opened 2023-08-04
✨ Change switch-domain to take output pathname

MikeSpreitzer:
## Summary
This PR changes the `switch-domain` and `wait-and-switch-domain` commands to simply take an output pathname rather than compute it from other parts.  This one usage we have does not want to use the output path that was being computed.

## Related issue(s)

Fixes #


MikeSpreitzer:
/cc @francostellari 
/cc @dumb0002 
@waltforme

clubanderson (commented):
/approve

Issue #866, opened 2023-08-07
feature: Adding support to select a subset of resources in a namespace to be synced to a workload execution cluster


dumb0002:
### Feature Description

KubeStellar needs to support the use of EdgePlacement object to specify a subset of resources in a namespace to be synced to a particular workload execution cluster. 

This feature is required for some applications such as the following:

-  KubeFlow:  the entire pipeline needs to run in one namespace at the same time, but we may want to run parts of the namespace in different clusters

- MCAD: in some use-cases MCAD needs to cherry pick objects from a namespace to be shipped to a particular location.

### Proposed Solution

KubeStellar EdgePlacement object allows selection of a subset of resources in a namespace to be synced to a particular workload execution cluster.  

### Want to contribute?

- [ ] I would like to work on this issue.

clubanderson:
@MikeSpreitzer unless you object, I think this is our next priority.

clubanderson:
@roytman please review along with @cwiklik  - Thank you

ezrasilvera:
Few questions: 
1. Why should this be done through EdgePlacement  ?  
2. Why can a user use multiple NS and put all resources needed to be synced in special NS ?

roytman:
> Why can a user use multiple NS and put all resources needed to be synced in special NS ?

KFP puts all its objects in the same NS.

cwiklik:
Current workaround for this issue requires creation of WDS for each target cluster.

dumb0002:
Fixed in this PR: https://github.com/kubestellar/kubestellar/pull/884

Issue #772, opened 2023-07-07
Special case for binding to exactly one workload execution cluster


clubanderson:
### Discussed in https://github.com/kubestellar/kubestellar/discussions/703

<div type='discussions-op-text'>

<sup>Originally posted by **MikeSpreitzer** June 28, 2023</sup>
While an EdgePlacement generally binds to an arbitrary number of workload execution clusters, there is a significant body of usage in which that number is intended to be exactly 1.  In this case there is no summarization problem; the source object's status section is properly designed to handle this case and arguably the reported state should propagate back to there.</div>

clubanderson:
pending #884

ezrasilvera:
@MikeSpreitzer I think we need to have a common design for any N. For example we can have the sumerizer configurable so it can just copy the full copy of the statuses into the summery object if the user choose to. It will be up to the user to decide if the number of WECs is too large so a real summary function need to be apply or that it's good enough to embed the full status of all related resources. 
In the above approach the case of single WEC deployment will be just a private case of "embed the full statuses of objects across all WECs" 
There are two advantages here:
1. Users get the status on the same space no matter what's the number of deployed WECs (don't need to go to MB spaces)
2. Users get the status from the same "Summery" object , whether it's embedding a direct copy of the statuses or the result of a summery operation. 
I don't see any special case with the number "1" for deployed  WEC

MikeSpreitzer:
I have been talking about a common design for any N, that is summarization.  I have been taking the approach that summarization should be programmable.  See #567 for my current proposal.  I think you could reasonably ask that the user be able to request a very simple form of summarization: a list of complete copies of the object in question.  As we discussed earlier, this impractical for large N, and we would point that out to users.

N=1 is special because the Kubernetes community has a standard way to represent the status from N=1 executing copy and it is so commonly expected that many people do not even think of the possibility of N other than 1.

ezrasilvera:
I think we are a special case. The whole purpose of KS is to transparently do CM over multiple WECs.  We can dictate the user that for KS you deploy an object X and you get the status in a different object (as a vector or outcome of summarization operation). 
I think we will get many many cases in which during normal operation N will increase from 1 to a larger (very small number) , especially when WECs are scarce.  It will be pity if users that just add another WEC suddenly need to change their code that process their status as it now passed in a completely different way.

MikeSpreitzer:
[...]

PR #884, opened 2023-08-11
✨ Let downsync object selection discriminate within a namespace

MikeSpreitzer:
## Summary
This PR expands the expressiveness of the "what predicate" in an EdgePlacement object, because users want to be able to select a subset of the objects for a given namespace and resource.

This PR includes both the API change and the implementation change.

For the syncer, #885 outlines an incremental roadmap and the milestone that needs to be reached to support this.

## Related issue(s)

This partially addresses #866; completion of that issue also requires the syncer change.


pkg/apis/edge/v1alpha1/edge-placement.go
```diff
@@ -73,32 +73,29 @@ type EdgePlacement struct {
...
 
-	// `namespaceSelector` identifies the relevant Namespace objects in terms of their labels.
-	NamespaceSelector metav1.LabelSelector `json:"namespaceSelector,omitempty"`
-
-	// `nonNamespacedObjects` defines the non-namespaced objects to bind with the selected Locations.
-	// +optional
-	NonNamespacedObjects []NonNamespacedObjectReferenceSet `json:"nonNamespacedObjects,omitempty"`
+	// `downsync` selects the objects to bind with the selected Locations for downsync.
+	// An object is selected if it matches at least one member of this list.
+	Downsync []DownsyncObjectTest `json:"downsync,omitempty"`
```
ronenkat, line 84:
```suggestion
	Downsync []DownsyncObjectCriteria `json:"downsync,omitempty"`
```

pkg/apis/edge/v1alpha1/edge-placement.go
```diff
@@ -73,32 +73,29 @@ type EdgePlacement struct {
...
 }
 
-// NonNamespacedObjectReferenceSet specifies a set of non-namespaced objects
-// from one particular API group.
-// An object is in this set if:
-// - its API group is the one listed;
-// - its resource (lowercase plural form of object type) is one of those listed; and
-// - EITHER its name is listed OR its labels match one of the label selectors.
-type NonNamespacedObjectReferenceSet struct {
+// DownsyncObjectTest is a test that can be applied to an object to tell whether it matches.
```
ronenkat, line 92:
```suggestion
// DownsyncObjectCriteria is the criteria that is used to tell whether objects should be down synced.
```

pkg/apis/edge/v1alpha1/edge-placement.go
```diff
@@ -73,32 +73,29 @@ type EdgePlacement struct {
...
+// An object matches if:
+// - the `apiGroup` criterion is satisfied;
+// - the `resources` criterion is satisfied;
+// - the `namespaceSelectors` criterion is satisfied; and
+// - EITHER the `objectNames` criterion or the `labelSelectors` criterion matches.
+type DownsyncObjectTest struct {
```
ronenkat, line 98:
```suggestion
type DownsyncObjectCriteria struct {
```

pkg/apis/edge/v1alpha1/edge-placement.go
```diff
@@ -107,10 +104,16 @@ type NonNamespacedObjectReferenceSet struct {
...
+	// at least one member of this list.
+	NamespaceSelectors []metav1.LabelSelector `json:"namespaceSelectors"`
+
+	// `objectNames` is a list of object names that match.
 	// An entry of `"*"` means that all match.
 	// Empty list means nothing matches.
-	ResourceNames []string `json:"resourceNames,omitempty"`
+	ObjectNames []string `json:"objectNames,omitempty"`
 
 	// `labelSelectors` allows matching objects by a rule rather than listing individuals.
```
ronenkat, line 118: Does all the labels represent a single predicate that matches all the labels? i.e., logical AND
or a subset? i.e., logical OR ?
Is it possible to have more than one predicate if all labels needs to match?

MikeSpreitzer: I added a clarification here.

dmatch01:
I'm not sure this is the place to put my comment so feel free to suggest an alternative location.

I believe this PR is intended to resolve the current workaround for allowing KubeStellar (KS) clients creating K8s objects with the same namespace to be placed on different clusters (sync targets?) without requiring the KS client to switch workspaces. The current workaround requires KS clients to switch workspaces in order to create K8s objects with the same namespace in different clusters.  It would be great to also have some documentation on how to use the KS client differently once this PR is merged.  Thanks!

MikeSpreitzer:
@dmatch01 : yes, you are right about the purpose here.  Yes, this line of work should include updating the documentation to show how to exploit the added expressiveness.

dmatch01:
> @dmatch01 : yes, you are right about the purpose here. Yes, this line of work should include updating the documentation to show how to exploit the added expressiveness.

Awesome!  Thanks @MikeSpreitzer 
cc:@cwiklik

pkg/apis/edge/v1alpha1/edge-placement.go
MikeSpreitzer: What is the reason for suggesting "Criteria" instead of "Test"?

ronenkat: I made the connection to QA (testing) rather than a test for a condition.
when you replicate something you do so based on a criteria. The term test is correct, but I don't think it is helpful to understand the API

clubanderson:
@cwiklik need your review on this PR to help us satisfy the need for MCAD to sync objects from different edgeplacements inside the same namespace.  This change is required before we tackle #772

MikeSpreitzer:
The force-push to 6a4edda7 is a rebase onto `main`.

clubanderson:
@yana1205 and @waltforme please review at your nearest convenience

pkg/apis/edge/v1alpha1/edge-placement.go
MikeSpreitzer: I have updated the comment on that type, hopefully it is clearer now.  Is this good enough for you?

MikeSpreitzer:
The force-push to e1e3aa41 adds the clarification to "DownsyncObjectTest" and squashes two other commits together.

clubanderson:
moving to sprint 4, did not make sprint 3

fileppb:
@MikeSpreitzer what's happenign with the lint step?

MikeSpreitzer:
@fileppb : the most important thing is that the current source does not compile.  I am in the midst of working on it.

The contextual-logging check is issuing the first error message that you see, but that is not something to focus on.  I can not work on the contextual logging checking until after the code compiles.

MikeSpreitzer:
The force-push to acd9bed1 is a rebase.

yana1205:
@MikeSpreitzer 
I am creating a test case and want to make sure whether I understand the specification of the proposed SyncerConfig correctly. Could you review my understanding?

Here is the SyncerConfig for the test case (https://github.com/yana1205/kubestellar/blob/discriminate-namespaced-syncer/test/e2e/kubestellar-syncer/testdata/namespaced-objects/syncer-config.yaml).
```
apiVersion: edge.kubestellar.io/v1alpha1
kind: SyncerConfig
metadata:
  name: syncer-config
spec:
  namespaceScope:
    namespaces:
    - test1
    resources:
    - apiVersion: v1
      group: ""
      resource: configmaps
    - apiVersion: v1
      group: ""
      resource: secrets
      
  namespacedObjects:
  - apiVersion: v1
    group: ""
    resource: configmaps
    objectsByNamespace:
    - namespace: test2
      names:
      - cm1
    - namespace: test3
      names: []
    - namespace: should-not-synced
      names: []
  - apiVersion: v1
    group: ""
    resource: secrets
    objectsByNamespace:
    - namespace: test2
      names: []
    - namespace: test3
      names:
      - s1
    - namespace: should-not-synced
      names: []
```

In `namespaceScope` field, `test1` namespace is set in `namespaces` child field and `v1/configmaps` and `v1/secrets` are set in `resources` chile field.
So, all configmaps and secrets in `test1` namespace are downsynced. That's nothing chaniging as Syncer does today.
In `namespacedObjects` field, for `v1/configmaps`, `cm1` object name  is set in `test2` namespace while empty array is set in both `test3` namespace and `should-not-synced` namespace. As for `v1/secrets`, `s1` object name is set in `test3` namespace while empty array is set in both `test2` namespace and `should-not-synced` namespace.

Therefore, what Syncer downsyncs are as follows:
- Namespace: `test1`, `test2`, `test3`
- Configmaps: all objects in `test1` namespace, `cm1` in `test2` namespace
- Secrets: all objects in `test1` namespace, `s1` in `test3` namespace

What Syncer does not dowynsync are as follows:
- Namespace: `should-not-synced`
- Configmaps: others than `cm1` in `test2` namespace, all objects in both `test3` namespace
- Secrets: others than `s1` in `test3` namespace, all objects in `test2` namespace
- Any namespaced objects in `should-not-synced` namespace

MikeSpreitzer:
@yana1205 : yes, I agree with what you wrote above.

MikeSpreitzer:
The force-push to 03b43de4 does a little re-organization and squashing.
This PR now starts with the one and only commit of #973, which is smaller and hopefully can get merged quickly.  Once that is done, I will rebase this PR onto `main`.

MikeSpreitzer:
The example1 docs-ecutable succeeded, with `kubectl-kubestellar-prep_for_syncer` calling for an image built from the current version of #968 , which is commit 8731b55053a056a7f3211279ef6cbd3e502c2fce .

MikeSpreitzer:
The force-pushes just above are squashing and updating the syncer image reference to the just-built tag for #968 .

MikeSpreitzer:
The force-push to ca5b2233 is a rebase onto `main`.

MikeSpreitzer:
More implementation code cleanup could be done, but I think that is not as high a priority as getting the interface change done.

MikeSpreitzer:
This is passing the example1 docs-ecutable with the latest syncer.

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-stage-2.md
```diff
@@ -150,12 +150,21 @@ metadata:
...
+    resources: [ configmaps ]
+    namespaceSelectors:
+    - matchLabels: {"common":"si"}
+    objectNames: [ httpd-htdocs ]
+  - apiGroup: apps
+    resources: [ replicasets ]
+    namespaceSelectors:
+    - matchLabels: {"common":"si"}
+    objectNames: [ "*" ]
   - apiGroup: apis.kcp.io
```
pdettori, line 164: If I want to downsync all objects in a namespace (as it was previously) can I just provide something like:
```
downsync
 - namespaceSelectors:
    -  matchLabels: {"common":"si"}
```
or do I need to provide something like:
```
 downsync:
  - apiGroup: "*"
    resources: [ "*" ]
    namespaceSelectors:
    - matchLabels: {"common":"si"}
    objectNames: ["*"]
```

pkg/apis/edge/v1alpha1/edge-placement.go
```diff
@@ -107,12 +104,19 @@ type NonNamespacedObjectReferenceSet struct {
...
+	// An object maches if and only if its labels match at least one member of this list.
 	LabelSelectors []metav1.LabelSelector `json:"labelSelectors,omitempty"`
 }
```
pdettori, line 122: Now that we can select specific API groups and versions, how do we ensure that the resources groups and versions are included in the list of resource groups and versions that the syncer is configured to sync ?

pdettori (commented):
Overall looks good to me, would be great to clarify on couple of questions I asked.

pkg/apis/edge/v1alpha1/edge-placement.go
MikeSpreitzer: The EdgePlacementSpec does not allow the author to be specific about API version.
The placement translator remains responsible for maintaining the translation of the collection of EdgePlacement objects into the collection of SyncerConfig objects.

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-stage-2.md
MikeSpreitzer: This is an interface _change_.  The old way of specifying namespaced objects to downsync is no longer available.  The comment on the new type, DownsyncObjectTest, gives its precise semantics.

PR #885, opened 2023-08-11
✨ Generalize SyncerConfig for anticipated EdgePlacement change

MikeSpreitzer:
## Summary
This PR generalizes the SyncerConfig object type in anticipation of changes to allow the EdgePlacement object to be more discriminating among namespaced downsynced objects.

This change is deliberately backward-compatible so that we can work on introducing it incrementally.  This API change does not cause build failures.  Then the syncer can be updated to implement the more general behavior in follow-on PRs (one or more, it does not have to be done until the placement translator starts using the added option).  This particular API change was designed to move SyncerConfig to be more like EdgeSyncConfig, expecting that this will be relatively easy for the syncer to follow.  Once the syncer implements the new option, we can change the placement translator to use it instead of the old way; this much is needed to support #884.  Once that is done, we can then simplify the syncer to not support the older way of specifying namespaced downsynced objects, and then we can remove that from the SyncerConfig definition leaving a simpler system all around.

## Related issue(s)

This will be needed before we can complete #884.


clubanderson:
@waltforme can you review this?

MikeSpreitzer:
@ronenkat : this particular change was designed to make SyncerConfig more like EdgeSyncConfig, anticipating that this similarity would both (a) make this change easier to implement and (b) enable eventual removal of some complexity.  EdgeSyncConfig uses lists of individuals.

PR #890, opened 2023-08-14
📖 Make example1 show both ways of deploying

MikeSpreitzer:
## Summary
This PR updates example1 to walk through both ways of deploying kcp and KubeStallar: the old way (as bare processes) and the new way (as workload in a Kubernetes cluster).

## Related issue(s)

Fixes #

/cc @clubanderson 
/cc @ronenkat 
/cc @dumb0002 
@francostellari


clubanderson (commented):
/lgtm
/approve

PR #891, opened 2023-08-14
📖 Make port number more clearly an example

MikeSpreitzer:
## Summary
This PR changes an example port number to make it more obvious that this particular bit of documentation has zero normative force regarding the port number but rather is trying to talk about how to cope with whatever port number is coming from the actual usage context.

## Related issue(s)

Fixes #

/cc @ronenkat 
@clubanderson 
/cc @dumb0002


clubanderson (commented):
/lgtm
/approve

PR #893, opened 2023-08-14
📖 Factor pre-reqs in quickstart

MikeSpreitzer:
## Summary
This PR factors the pre-reqs in the quickstart into three groups:
(1) Those needed for KubeStellar, period.
(2) The additional things needed to run the example in the quickstart.
(3) The additional things needed by contributors (which this PR moves to the CONTRIBUTING document).

## Related issue(s)

Fixes #

/cc @clubanderson 
/cc @dumb0002 
/cc @ronenkat 
@francostellari

### DIFF
--- a/docs/content/Contribution guidelines/CONTRIBUTING.md
+++ b/docs/content/Contribution guidelines/CONTRIBUTING.md
@@ -9,7 +9,9 @@ We can't wait to collaborate with you!
 
 ### Prerequisites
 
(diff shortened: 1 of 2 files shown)


clubanderson (commented):
had a look on my local laptop.  It's a bit much, but I think it works.  We will see if clients feedback.  I am hopeful that these instructions will reduce, in terms of size and pre-reqs, as we mature and lose some dependencies.  This is good for now.
/lgtm
/approve

PR #892, opened 2023-08-14
🐛 Add helm chart to platform archive

MikeSpreitzer:
## Summary
This PR adds the Helm chart to the stuff put into a platform archive, because the `kubectl kubestellar deploy` command uses it.

## Related issue(s)

Fixes #

/cc @ronenkat 
/cc @dumb0002 
@clubanderson 
@francostellari

PR #894, opened 2023-08-15
✨ Make bootstrap able to deploy to kubernetes cluster

MikeSpreitzer:
## Summary
This PR updates the bootstrap script so that it has the option to deploy to a Kubernetes cluster rather than as bare processes, and updates the quickstart so that it describes how to use both techniques.

## Related issue(s)

Fixes #

/cc @ronenkat 
/cc @dumb0002 
/cc @clubanderson 
@francostellari

### DIFF
--- a/bootstrap/bootstrap-kubestellar.sh
+++ b/bootstrap/bootstrap-kubestellar.sh
@@ -16,10 +16,13 @@
 
 # Usage: $0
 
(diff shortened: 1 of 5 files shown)


MikeSpreitzer:
Because I forgot to open #892 before the v0.5.0 release was made, this will not work with v0.5.0 and earlier.

clubanderson (commented):
/lgtm
/approve

Issue #852, opened 2023-08-01
feature: Change  the KS setup create Spaces instead of workspaces


ezrasilvera:
Use bootstrap space+ the space-manager to create the KS spaces ) inventory, WL, and core

andreyod:
PR https://github.com/kubestellar/kubestellar/pull/897 . Need to address review comments.

effi-ofer:
At this point this is still WIP because there (1) are additional scripts in the scripts directory that need to be converted from workspaces to spaces and (2) the documentation needs to be updated.

ezrasilvera:
Replaced with https://github.com/kubestellar/kubestellar/issues/1215

PR #897, opened 2023-08-15
:sparkles:Change the setup scripts to create Spaces instead of workspaces

andreyod:
## Summary
1. Create workspace for space management and deploy needed CRDs.
2. Start space-manager controller.
3. Use kubectl to create spaces instead of workspaces.

## Related issue(s)
https://github.com/kubestellar/kubestellar/issues/852

Fixes #

### DIFF
--- a/config/space/provider_kcp.yaml
+++ b/config/space/provider_kcp.yaml
@@ -1,0 +1,8 @@
+apiVersion: space.kubestellar.io/v1alpha1
+kind: SpaceProviderDesc
+metadata:
+  name: default
+spec:
+  ProviderType: "kcp"
+  SpacePrefixForDiscovery: "espw"
+  Config: |
--- a/docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
+++ b/docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
@@ -92,11 +92,66 @@ to pass the new file with `--kubeconfig` on the command lines when
 using kcp or KubeStellar.
 
 
+### Space Manager
+
+The Space Manager will use space provider to create and manage underlying cluster
+for each Space object. Space objects are created in `root:space-mgt` context
+under `spaceprovider-default` namespace.
+
+```shell
+space-manager --context root:space-mgt &
+```
+``` { .bash .no-copy }
+...
+I0829 13:35:53.848054  379407 controller.go:143] "starting manager space controller"
+controller="space-manager"
+...
+I0829 13:35:53.865272  379407 reconcile_spaceprovider.go:87] "Provider namespace cre
+ated" controller="space-manager" provider="default" namespace="spaceprovider-default"
+...
+I0829 13:35:53.870486  379407 provider.go:176] "New space was detected" space="espw"
+provider="default"
+```
+
+You can get a listing of Spaces as follows.
+
+```shell
+kubectl config use-context root:space-mgt
+kubectl get spaces -A
+```
+``` { .bash .no-copy }
+NAMESPACE               NAME   AGE
+spaceprovider-default   espw   90s
+```
+
+
 ### Create an inventory management workspace.
 ```shell
-kubectl ws root
-kubectl ws create imw-1 
+kubectl config use-context root:space-mgt
+kubectl apply -f - <<EOF
+apiVersion: space.kubestellar.io/v1alpha1
+kind: Space
+metadata:
+  name: imw-1
+  namespace: spaceprovider-default
+spec:
+  SpaceProviderDescName: default
+  Managed: true
+EOF
 ```
+``` {.bash .hide-me}
+sleep 5
+```
+
+Check out the Space status as follows.
+
+```shell
+kubectl get space imw-1 -n spaceprovider-default -o yaml | grep Phase
+```
+``` { .bash .no-copy }
+  Phase: Ready
+```
+
 ### Create SyncTarget and Location objects to represent the florin and guilder clusters
 
 Use the following two commands. They label both florin and guilder
--- a/scripts/kubectl-kubestellar-ensure-wmw
+++ b/scripts/kubectl-kubestellar-ensure-wmw
@@ -17,10 +17,12 @@
 # Purpose: ensure that a workload management workspace exists and has
 # the needed APIBindings.
 
-# Usage: kubectl ws parent; $0 ($kubectl_flag | --with-kube boolean)* wmw_name
+# Usage: $0 ($kubectl_flag | --with-kube boolean)* wmw_path
 
 want_kube=true
 wmw_name=""
+space_mgt="root:space-mgt"
+spaceprovider_ns="spaceprovider-default"
 
 while (( $# > 0 )); do
     case "$1" in
@@ -35,7 +37,7 @@ while (( $# > 0 )); do
 	(--*=*|-?=*)
 	    kubectl_flags[${#kubectl_flags[*]}]="$1";;
 	(-h)
-	    echo "Usage: kubectl ws parent; kubectl kubestellar ensure wmw (\$kubectl_flag | --with-kube boolean)* wm_workspace_name"
+	    echo "Usage: kubectl kubestellar ensure wmw (\$kubectl_flag | --with-kube boolean)* wm_workspace_path"
 	    exit 0;;
 	(--*|-?)
 	    if (( $# > 1 ))
@@ -58,7 +60,7 @@ while (( $# > 0 )); do
 done
 
 if [ "$wmw_name" == "" ]; then
-    echo "$0: workload management workspace name not specified" >&2
+    echo "$0: workload management workspace path not specified" >&2
     exit 1
 fi
 
@@ -70,20 +72,57 @@ esac
 
 set -e
 
-original_ws=$(kubectl ws "${kubectl_flags[@]}" . --short)
+IFS=':' read -ra ws_array <<< "$wmw_name"
+first_item="true"
+parent=""
+kubectl config use-context $space_mgt
 
-if [ "$original_ws" != root ]; then
-    kubectl ws "${kubectl_flags[@]}" ..
-    if kubectl "${kubectl_flags[@]}" get workspaces.tenancy.kcp.io "$wmw_name" &> /dev/null; then
-	echo "Warning: parent has a child with the given WMW name; are you sure the right workspace is current?" >&2
-	sleep 15 # give a chance for ^C
-    fi
-    kubectl ws "${kubectl_flags[@]}" -
-fi
+create_space() {
+    kubectl "${kubectl_flags[@]}" apply -f - <<EOF
+apiVersion: space.kubestellar.io/v1alpha1
+kind: Space
+metadata:
+  name: $item
+  namespace: $spaceprovider_ns
+  annotations:
+    kubestellar.io/space-path: $parent
+spec:
+  SpaceProviderDescName: default
+  Managed: true
+EOF
+}
 
-if kubectl "${kubectl_flags[@]}" get workspaces.tenancy.kcp.io "$wmw_name" &> /dev/null
-then kubectl ws "${kubectl_flags[@]}" "$wmw_name"
-else kubectl ws "${kubectl_flags[@]}" create "$wmw_name" --enter
+for item in "${ws_array[@]}"; do
+    case "$item" in
+    (root)
+        if [ "$first_item" == "true" ]; then
+            parent=$item
+        else
+            echo "\"root\" workspace can only be used at the beginning"
+            exit 2
+        fi;;
+    (*)
+        if [ "$verbose" == "true" ]; then
+            if [[ -z "$(kubectl get space $item -n $spaceprovider_ns --ignore-not-found)" ]] ; then
+                echo "Space "$item" does not exist. Creating"
+                create_space
+                sleep 3
+            fi
+        else
+            if [[ -z "$(kubectl get space $item -n $spaceprovider_ns --ignore-not-found)" ]] ; then
+                create_space 1> /dev/null
+                sleep 3
+            fi
+        fi
+        parent=$parent":"$item;;
+    esac
+    first_item="false"
+done
+
+if ! kubectl ws "${kubectl_flags[@]}" "$wmw_name"
+then
+    echo "Can't access workspace ""$wmw_name"".Check space exist and ready"
+    exit 3
 fi
 
 if ! kubectl "${kubectl_flags[@]}" get APIBinding bind-espw &> /dev/null; then
--- a/scripts/kubestellar
+++ b/scripts/kubestellar
@@ -255,10 +255,10 @@ function ensure_space_mgt() {
     fi
     kubectl ws create-context --overwrite
     # create space CRDs
-    eval kubectl create -f "$SCRIPT_DIR/../config/crds/space.kubestellar.io_spaces.yaml" $verbdir
-    eval kubectl create -f "$SCRIPT_DIR/../config/crds/space.kubestellar.io_spaceproviderdescs.yaml" $verbdir
+    eval kubectl apply -f "$SCRIPT_DIR/../config/crds/space.kubestellar.io_spaces.yaml" $verbdir
+    eval kubectl apply -f "$SCRIPT_DIR/../config/crds/space.kubestellar.io_spaceproviderdescs.yaml" $verbdir
     # deploy KCP provider
-    eval kubectl create -f "$SCRIPT_DIR/../config/space/provider_kcp.yaml" $verbdir
+    eval kubectl apply -f "$SCRIPT_DIR/../config/space/provider_kcp.yaml" $verbdir
     echo "Finished populating the space-mgt with kubestellar apiexports"
 }
 
--- a/scripts/kubestellar
+++ b/scripts/kubestellar
@@ -34,6 +34,7 @@ verbdir="&> /dev/null"
 remove=0
 cleanup=0
 espw_name="espw"
+space_mgt_name="space-mgt"
 log_folder=$(pwd)/kubestellar-logs
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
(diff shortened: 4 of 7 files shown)


scripts/kubestellar
```diff
@@ -240,6 +241,24 @@ EOF
     echo "Finished augmenting root:compute for KubeStellar"
 }
 
+function ensure_space_mgt() {
+    kubectl ws root &> /dev/null
+    $SCRIPT_DIR/space-provider-config "$KUBECONFIG" "$SCRIPT_DIR/../config/space/provider_kcp.yaml"
+    if kubectl get Workspace "$space_mgt_name" &> /dev/null; then
+	echo "space-mgt workspace already exists -- using it:"
+	kubectl ws "$space_mgt_name"
```
effi-ofer, line 249: If the space mgr already exists, do we need to set the context to it?

scripts/kubestellar
```diff
@@ -240,6 +241,24 @@ EOF
...
+    else
+        kubectl ws create "$space_mgt_name" --enter
+    fi
+    kubectl ws create-context --overwrite
```
effi-ofer, line 253: Sorry for being ignorant - what does this command actually do?

config/space/provider_kcp.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: space.kubestellar.io/v1alpha1
+kind: SpaceProviderDesc
+metadata:
+  name: default
+spec:
+  ProviderType: "kcp"
+  SpacePrefixForDiscovery: "espw"
+  Config: |
```
MikeSpreitzer, line 8: Because

```go
	Config string `json:"Config,omitempty"`
```

specifying the empty string as the value for this field is the same as not mentioning this field.

At least, I am sure that would be true if it had been marked `// +optional` (as it should); I think the `json` tag rules in this matter.

MikeSpreitzer: BTW, the convention in the design of Kubernetes API objects is to use lowerCamelCase for field names in JSON.

scripts/kubestellar
MikeSpreitzer: .. and why is it needed here?

scripts/kubestellar
```diff
@@ -240,6 +241,24 @@ EOF
...
+    # create space CRDs
+    eval kubectl create -f "$SCRIPT_DIR/../config/crds/space.kubestellar.io_spaces.yaml" $verbdir
```
MikeSpreitzer, line 255: Better to write this in idempotent style; use `kubectl apply` rather than `kubectl create`.

MikeSpreitzer: What about all the other spaces where these resources need to be defined?  How do these resources get defined there?

scripts/kubestellar
```diff
@@ -320,6 +348,17 @@ if [[ ! -d $log_folder ]]; then
     mkdir -p $log_folder
 fi
 
+space-manager -kubeconfig $KUBECONFIG -context "root:"$space_mgt_name -v=2 >& $log_folder/space-manager-log.txt &
```
MikeSpreitzer, line 351: Why is `-kubeconfig $KUBECONFIG` needed here?  The `space-manager` command should be sensitive to the `KUBECONFIG` envar like any other command.

scripts/space-provider-config
```diff
@@ -0,0 +1,57 @@
...
+if [ $# -ne 2 ]; then
+    echo "Usage: $0 <config file> <provider yaml>"
+    exit 1
+fi
+
+config="$1"
+provider="$2"
+
+# Check if config file exists
+if [ ! -f "$config" ]; then
```
MikeSpreitzer, line 32: https://www.gnu.org/software/bash/manual/bash.html#Bash-Conditional-Expressions says you could test `-r` here

scripts/space-provider-config
```diff
@@ -0,0 +1,57 @@
...
+    echo "Error: $config does not exist."
+    exit 1
+fi
+
+# Check if provider file exists
+if [ ! -f "$provider" ]; then
```
MikeSpreitzer, line 38: https://www.gnu.org/software/bash/manual/bash.html#Bash-Conditional-Expressions says you could test `-w` here

scripts/space-provider-config
```diff
@@ -0,0 +1,57 @@
...
+    echo "Error: $provider does not exist."
+    exit 1
+fi
+
+# Define the specConfig string
+specConfig="  Config: |"
+
+# Check if config field exists in the provider yaml file and remove text after it
+if grep -q "$specConfig" "$provider"; then
+    sed -i "/$specConfig/,\$d" "$provider"
```
MikeSpreitzer, line 48: You should check out `sed -i` on both Mac and Linux.

scripts/space-provider-config
```diff
@@ -0,0 +1,57 @@
...
+    echo "Error: $config does not exist."
+    exit 1
+fi
+
+# Check if provider file exists
+if [ ! -f "$provider" ]; then
+    echo "Error: $provider does not exist."
+    exit 1
+fi
+
```
MikeSpreitzer, line 42: We already spec `jq` and `yq` as pre-reqs; better to use those tools that faff about with `sed`.

scripts/kubestellar
andreyod: 1. It is not necessary. Can be removed.

2. Will change to append

scripts/kubestellar
MikeSpreitzer: OK, I see that we have not considered #913 .

scripts/space-provider-config
MikeSpreitzer: In particular, the `-i` flag with no actual suffix is handled differently in the two.

scripts/kubestellar
andreyod: Done.

scripts/space-provider-config
MikeSpreitzer: Mustfix: this will not work on a Mac, where `-i` MUST be followed by an extension.

```console
mspreitz@mjs12 ~ % man sed | head -10
SED(1)                       General Commands Manual                      SED(1)

NAME
     sed – stream editor

SYNOPSIS
     sed [-Ealnru] command [-I extension] [-i extension] [file ...]
     sed [-Ealnru] [-e command] [-f command_file] [-I extension] [-i extension]
         [file ...]
```

scripts/space-provider-config
MikeSpreitzer: Also: wouldn't it be simpler to output a file that is a function of two distinct inputs (the config lacking entirely lacking what will be injected, and the config to inject) rather than try to modify whatever happens to be in the readwrite file at the moment?

scripts/kubestellar
MikeSpreitzer: Since creating the context is a separate step from creating the Workspace, yes, it is possible that one happened in the past but not the other.

scripts/kubestellar
MikeSpreitzer: Better to write this in idempotent style; use `kubectl apply` rather than `kubectl create`.

MikeSpreitzer, line 255: Better to write this in idempotent style; use `kubectl apply` rather than `kubectl create`.

scripts/kubestellar
MikeSpreitzer: Now I suspect that this context is used in more places.  If it is not just a private detail of one file then it needs to be documented.

scripts/space-provider-config
effi-ofer: I believe changing the line to sed -i "" "/$specConfig/,\$d" "$provider" should work on both gnu sed and bsd sed. but I need to find a machine to test this on.

scripts/space-provider-config
andreyod: Changed

scripts/space-provider-config
andreyod: Changed

scripts/space-provider-config
andreyod: Replaced `sed` with `awk`. Tested on both Mac and Linux.

config/space/provider_kcp.yaml
andreyod: Agree, will be fixed in dedicated PR.

scripts/kubestellar
andreyod: We create context in all cases. The `--overwrite` flag will overwrite the context if it's already exist.

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
```diff
@@ -92,11 +92,66 @@ to pass the new file with `--kubeconfig` on the command lines when
 using kcp or KubeStellar.
 
 
+### Space Manager
+
+The Space Manager will use space provider to create and manage underlying cluster
```
MikeSpreitzer, line 97: I thought we were making a distinction between:
- a "space provider", an underlying technology that implements multiple spaces and is independent of the space framework; examples include kcp, kind, kubeflex, rancher; vs.
- a "provider adapter", the code that makes the connection between the space framework (Space Manager) and a particular space provider, thus depending on both.

MikeSpreitzer, line 97: but the underlying thing is not necessarily a _cluster_, right?

MikeSpreitzer, line 97: In English we would write "a space provider" or "the space provider" or "space providers".

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
```diff
@@ -92,11 +92,66 @@ to pass the new file with `--kubeconfig` on the command lines when
...
+for each Space object. Space objects are created in `root:space-mgt` context
```
MikeSpreitzer, line 98: (A) In English this parses as saying that the context is under the namespace.  I think that is the opposite of what you mean.  I suspect that what you mean would be better said as "Space objects are created in the `spaceprovider-default` namespace in the ... context".

(B) I assume that by "context" here you mean the concept in a kubeconfig file.  A context does not hold API objects or namespaces.  A context is a particular kube client configuration.  So I suspect that what you mean would be better written as "Space objects are created in the `spaceprovider-default` namespace using the context named `root:space-mgt` in your chosen kubeconfig file".

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
```diff
@@ -92,11 +92,66 @@ to pass the new file with `--kubeconfig` on the command lines when
...
+under `spaceprovider-default` namespace.
+
+```shell
+space-manager --context root:space-mgt &
```
MikeSpreitzer, line 102: Is this configuring the space-manager to use a pre-existing context or telling it to create one?  I suspect the former.  If so, the reader of this document has not seen what causes that context to exist (unless that appears earlier due to a file that GitHub is listing later in this PR).

scripts/kubectl-kubestellar-ensure-wmw
```diff
@@ -70,20 +72,57 @@ esac
...
-fi
+create_space() {
+    kubectl "${kubectl_flags[@]}" apply -f - <<EOF
+apiVersion: space.kubestellar.io/v1alpha1
+kind: Space
+metadata:
+  name: $item
+  namespace: $spaceprovider_ns
+  annotations:
+    kubestellar.io/space-path: $parent
```
MikeSpreitzer, line 88: This is not interpreted by all space providers, right?  What does this script mean when the space provider does not interpret this?

scripts/kubectl-kubestellar-ensure-wmw
```diff
@@ -70,20 +72,57 @@ esac
...
 
-if [ "$original_ws" != root ]; then
-    kubectl ws "${kubectl_flags[@]}" ..
-    if kubectl "${kubectl_flags[@]}" get workspaces.tenancy.kcp.io "$wmw_name" &> /dev/null; then
-	echo "Warning: parent has a child with the given WMW name; are you sure the right workspace is current?" >&2
-	sleep 15 # give a chance for ^C
-    fi
-    kubectl ws "${kubectl_flags[@]}" -
-fi
+create_space() {
```
MikeSpreitzer, line 80: I do not understand the logic here.  If users ask for paths `root:foo:baz` and `root:bar:baz` then they will be thinking of a tree with five vertices, including two leaves --- but that is not what they will get.

scripts/kubectl-kubestellar-ensure-wmw
```diff
@@ -70,20 +72,57 @@ esac
...
+    esac
+    first_item="false"
+done
 
-if kubectl "${kubectl_flags[@]}" get workspaces.tenancy.kcp.io "$wmw_name" &> /dev/null
-then kubectl ws "${kubectl_flags[@]}" "$wmw_name"
-else kubectl ws "${kubectl_flags[@]}" create "$wmw_name" --enter
+if ! kubectl ws "${kubectl_flags[@]}" "$wmw_name"
+then
+    echo "Can't access workspace ""$wmw_name"".Check space exist and ready"
```
MikeSpreitzer, line 124: I doubt that the quotes here are doing what you intend.

```console
bash-3.2$ wmw_name=root:foo:baz

bash-3.2$     echo "Can't access workspace ""$wmw_name"".Check space exist and ready"
Can't access workspace root:foo:baz.Check space exist and ready
```

Do you know about `${wmw_name@Q}` ?  Did you know that MacOS ships with a version of bash so old that it does not support `@Q`?

scripts/kubectl-kubestellar-ensure-wmw
```diff
@@ -17,10 +17,12 @@
 # Purpose: ensure that a workload management workspace exists and has
 # the needed APIBindings.
 
-# Usage: kubectl ws parent; $0 ($kubectl_flag | --with-kube boolean)* wmw_name
+# Usage: $0 ($kubectl_flag | --with-kube boolean)* wmw_path
```
MikeSpreitzer, line 20: This script is documented.  You have to change the documentation along with the behavior of the script.

scripts/kubestellar
MikeSpreitzer: Ah, right.

scripts/kubestellar
MikeSpreitzer: If it is not just a private detail of one file then it needs to be documented.

scripts/kubestellar
```diff
@@ -243,6 +244,24 @@ EOF
...
+    $SCRIPT_DIR/space-provider-config "$KUBECONFIG" "$SCRIPT_DIR/../config/space/provider_kcp.yaml"
+    if kubectl get Workspace "$space_mgt_name" &> /dev/null; then
+	echo "space-mgt workspace already exists -- using it:"
+	kubectl ws "$space_mgt_name"
+    else
+        kubectl ws create "$space_mgt_name" --enter
+    fi
+    kubectl ws create-context --overwrite
+    # create space CRDs
+    eval kubectl apply -f "$SCRIPT_DIR/../config/crds/space.kubestellar.io_spaces.yaml" $verbdir
```
MikeSpreitzer, line 258: Look at the picture of software components in https://docs.google.com/drawings/d/1U5nwzHeSdpEufzOPmYX_8cIvNOUzRa3oOFuKqNS4smM .

If we put multiple components in one GitHub _repository_ then we should cleanly separate them at a high level in the directory structure.  We are slouching towards a monorepo.  We can no longer afford to slouch, this needs to be done with some recognition and discipline.

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
MikeSpreitzer, line 102: Since `kubestellar init` is doing this, why is it also being done separately?

scripts/space-provider-config
```diff
@@ -0,0 +1,58 @@
...
+# Check if provider file exists
+if [ ! -w "$provider" ]; then
+    echo "Error: $provider does not exist or not writable."
+    exit 1
+fi
+
+# Define the specConfig string
+specConfig="  Config: |"
+
+# Check if config field exists in the provider yaml file and remove text after it
```
MikeSpreitzer, line 46: This editing logic is hard to review.
It would be much easier to review logic that simply output the provider file.

MikeSpreitzer:
The current title of this PR is misleading.  I think this PR _adds_ creation of Spaces.  The "instead of" is confusing because workspaces are still being created too.  This is important: it needs to be clear how this PR fits into an incremental development plan.

The current title is also narrower than the full scope of the PR.  This PR is intended to be about all of the setup logic, if I understand correctly.

In support of incrementality, this PR _could_ be broken into a series of two or more: the first one makes the setup deploy the space framework but does not change any scripts to use the space framework to manipulate spaces, while later PRs would change scripts to use the now-extant space framework to manipulate spaces (and thus workspaces indirectly).

MikeSpreitzer:
Please also bear in mind that the multi-cluster configuration management code now supports two ways of being deployed: as bare processes, or as a Kubernetes workload.  See, e.g., the discussion early in https://docs.kubestellar.io/main/Coding%20Milestones/PoC2023q1/commands/

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
andreyod: On `kubestellar init` we only create space-mgt workspace and a context for this workspace. The `space-manager` controller is started on `kubestellar start`. Currently no controllers are started on init. Should we start `space-manager` on init?

scripts/kubectl-kubestellar-ensure-wmw
andreyod: Can you give me a pointer to this document?

scripts/kubectl-kubestellar-ensure-wmw
andreyod: The `create_space` func is called from the `for` loop for each part of the path except `root`. So if user ask for `root:my-org:wmw-c` , the script will create Space for `my-org` and for `wmw-c`. Than `wmw-c` workspace will be created under `my-org` workspace.

scripts/kubectl-kubestellar-ensure-wmw
andreyod: Will change

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
andreyod: This kubeconfig context is created on `kubestellar init`. Should we document this in `Initialize the KubeStellar platform` section of this document or/and somewhere else?

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
effi-ofer: That's correct.  In the case of kubeflex for example they are really a control plane or an api server.  I guess the question is how do we define 'space'.  And the answer perhaps is that a space, similar to a kcp workspace that it replaces, provides a mechanism for multi cluster and multi tenancy by providing isolated api resources and endpoint?

scripts/kubectl-kubestellar-ensure-wmw
effi-ofer: #issue 951 is opened for this.

scripts/kubectl-kubestellar-ensure-wmw
andreyod: Correct, we currently don't support this because the space name is unique. We can return an error to the user if the leave space already exit, and we will open an issue to support full hierarchy.
We don't have to deal with paths. We can create the parent first and then run the script to create a wmw.
WDYT?

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
pdettori: why the name of the context in the kubeconfig file is so important that it needs to be mentioned? Is it because there is some convention that for context naming that the reader need to be aware of?

pdettori: well, the answer seems to be in the script below, where you need to set that context - ignore my comment

ezrasilvera:
I suggest we wait till we have the restructuring
/hold

PR #907, opened 2023-08-16
✨ update logging to append instead of replace

### LINKED ISSUE #825: feature: containerization: `kubestellar start` should append to, rather than overwrite, logs
### Feature Description

`kubestellar start` currently launches processes that overwrite their log file if it already exists.  It would be better if the older log contents were preserved.

### Proposed Solution

Make `kubestellar start` append to logs that already exist.

### Want to contribute?

### Additional Context

_No response_

clubanderson:
## Related issue(s)

Fixes #825

### DIFF
--- a/scripts/kubestellar
+++ b/scripts/kubestellar
@@ -320,7 +320,7 @@ if [[ ! -d $log_folder ]]; then
     mkdir -p $log_folder
 fi
 
-mailbox-controller -v=2 >>& $log_folder/mailbox-controller-log.txt &
+mailbox-controller -v=2 &>> $log_folder/mailbox-controller-log.txt &
 
 run_status=$(wait_for_process mailbox-controller)
 if [ $run_status -eq 0 ]; then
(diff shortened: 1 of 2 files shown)


scripts/kubestellar
```diff
@@ -320,7 +320,7 @@ if [[ ! -d $log_folder ]]; then
     mkdir -p $log_folder
 fi
 
-mailbox-controller -v=2 >& $log_folder/mailbox-controller-log.txt &
+mailbox-controller -v=2 >>& $log_folder/mailbox-controller-log.txt &
```
MikeSpreitzer, line 323: Are you sure about these?  https://www.gnu.org/software/bash/manual/bash.html#Appending-Standard-Output-and-Standard-Error says it should be `&>>`.  And that is what I get when testing:

```console
mspreitz@mjs12 ~ % bash
bash-5.2$ ls -ld fubar
ls: fubar: No such file or directory
bash-5.2$ date >>& fubar
bash: syntax error near unexpected token `&'
bash-5.2$ date &>> fubar
bash-5.2$ ls -ld fubar
-rw-r--r--  1 mspreitz  staff  29 Aug 16 17:44 fubar
```

clubanderson:
I assumed it was just an addition of a '>' - but yes, your right - updated.

scripts/kubestellar
clubanderson: changed all 3 to match your example.

clubanderson:
closing, accidentally checked in something I was working on that is not part of this issue.

PR #910, opened 2023-08-17
✨ update logging to append instead of replace

### LINKED ISSUE #825: feature: containerization: `kubestellar start` should append to, rather than overwrite, logs

clubanderson:
## Related issue(s)

Fixes #825

### DIFF
--- a/scripts/kubestellar
+++ b/scripts/kubestellar
@@ -320,7 +320,7 @@ if [[ ! -d $log_folder ]]; then
     mkdir -p $log_folder
 fi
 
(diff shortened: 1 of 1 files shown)


MikeSpreitzer (commented):
/lgtm
/approve
Thanks!

Issue #414, opened 2023-05-18
feature: the kubectl plugins should be based on proper arg parsing


MikeSpreitzer:
### Feature Description

The kubectl plugins, aside from the syncer-gen one, are currently bash scripts that do some generic parsing of kubectl flags.  This is more complex programming than should be done in bash and, more importantly, is ignorant of what kubectl flags really are.  

### Proposed Solution

These should be rewritten to rely on a standard library that actually understands kubectl flags, such as https://github.com/kubernetes/cli-runtime .

### Alternative Solutions

🤷 

### Want to contribute?

- [ ] I would like to work on this issue.

clubanderson:
@MikeSpreitzer can you point me at an example plugin that uses https://github.com/kubernetes/cli-runtime?  I need to learn by example before trying to modify any of our plugins.

clubanderson:
I am working on converting the get_log script first.  I hope to have something working shortly.

clubanderson:
the scripts I am targeting for conversion, in no particular order, are:

kubectl-kubestellar-deploy
kubectl-kubestellar-ensure-location
kubectl-kubestellar-ensure_wmw
kubectl-kubestellar-get_external_kubeconfig
kubectl-kubestellar-get_internal_kubeconfig
kubectl-kubestellar-get_log
kubectl-kubestellar-prep_for_cluster
kubectl-kubestellar-prep_for_syncer
kubectl-kubestellar-remove-location
kubectl-kubestellar-remove-wmw

I am not sure if there is a need to support:
kubectl-workspace-ensure

MikeSpreitzer:
For example, the KubeStellar scripts do not know which flags take an argument vs. which do not.

clubanderson:
@namasl is covering this starting with #1294

clubanderson:
we will pick this up in the new year

namasl:
With the recent architecture change to use the space framework, PR #1294 has been closed, and a new PR will contain similar Bash script to Go conversions for the latest kubectl-kubestellar* commands.

As this issue's description is generic enough to cover the requisite changes under the new architecture, I figured I'd add this comment here rather than make a new issue.

PR #912, opened 2023-08-17
✨ replacing kubectl-kubestellar-get_log script with clientgo version

clubanderson:
## Related issue(s)
#414 

Fixes #


clubanderson:
I got the get_log to compile and it now retrieves logs correctly.  1 (major) issue that I can’t figure out right now, when I add get_log as a subcommand ‘kubectl kubestellar’ yields the get_log output (usage).  And, I can issue ‘kubectl kubestellar -f mailbox-controller’ and get the log for the mailbox-controller - it should not do this.  When I issue ‘kubectl kubestellar get_log -f mailbox-controller -n 2’ it works properly.  So I think I am just having an issue with having honoring get_log as a subcommand and not a replacement for ‘kubestellar’

not sure why this is happening, any hints are appreciated.

cmd/kubectl-kubestellar/cmd/kubectlKubestellar.go
```diff
@@ -0,0 +1,62 @@
+/*
```
MikeSpreitzer, line 1: We conventionally eschew uppercase in filenames because developers work on a variety of OSes, which have an unfortunate variety in the ways they handle upper case in filenames.  "kubectl-kubestellar.go" would be commonly used in place of "kubectlKubestellar".

cmd/kubectl-kubestellar/main.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+
+	"github.com/spf13/pflag"
+
+	"github.com/kubestellar/kubestellar/cmd/kubectl-kubestellar/cmd"
+)
+
+func main() {
+	flags := pflag.NewFlagSet("kubectl-kubestellar", pflag.ExitOnError)
+	pflag.CommandLine = flags
```
MikeSpreitzer, line 30: What is the point of this replacement?  I see no use of this flag set.  This is not the pattern shown in https://github.com/spf13/pflag#supporting-go-flags-when-using-pflag .  KubectlKubestellarCommand makes and uses a different FlagSet.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	podName, err := getServerPodName(clientset, namespace)
+	if err != nil {
+		fmt.Println("Error getting server pod name:", err)
+		os.Exit(1)
+	}
+
+	var serverPod string
+	for {
+		if podName != "" && len(podName) > 0 {
```
MikeSpreitzer, line 115: The two factors here are always equal, you only need one.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,220 @@
...
+		os.Exit(1)
+	}
+	podName, err := getServerPodName(clientset, namespace)
+	if err != nil {
+		fmt.Println("Error getting server pod name:", err)
+		os.Exit(1)
+	}
+
+	var serverPod string
+	for {
```
MikeSpreitzer, line 114: `podName` is not varying as this loop repeats.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,220 @@
...
+		if podName != "" && len(podName) > 0 {
+			serverPod = strings.Trim(string(podName), "'")
+			break
+		}
+		time.Sleep(10 * time.Second)
+	}
+
+	err = executeCommandInPod(clientset, namespace, config, podName, []string{"ls", "/home/kubestellar/ready"}, os.Stdin, os.Stdout, os.Stderr)
```
MikeSpreitzer, line 122: You want this to be a test inside the loop.  The idea is to loop until the ready signal appears.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,220 @@
...
+	if err != nil {
+		fmt.Println("Error executing command in pod:", err)
+		os.Exit(1)
+	}
+
+	kubectlLogArgs := []string{"tail", "kubestellar-logs/" + logfile}
+	if followFlag {
+		kubectlLogArgs = append(kubectlLogArgs, "-f")
+	}
+	if linesFlag != "+0" {
```
MikeSpreitzer, line 132: I assume you are trying to match the behavior of the existing script.  In that case, passing along the "lines" value should be unconditional.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,220 @@
...
+		kubectlLogArgs = append(kubectlLogArgs, linesFlag)
```
MikeSpreitzer, line 133: The `tail` command takes `-n` before the "lines" value.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,220 @@
...
+	if err != nil {
+		fmt.Println("Error building Kubernetes config:", err)
+		os.Exit(1)
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		fmt.Println("Error creating Kubernetes clientset:", err)
+		os.Exit(1)
+	}
+	namespace, err := getNamespaceNames(clientset, ksNamespace)
```
MikeSpreitzer, line 102: I do not understand why this is here.  At development time you know the namespace that you want to use.

clubanderson: I had the wrong kubeconfig when I tried to use this for the first time.  without checking for ns, I would have never known what the problem was.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
clubanderson: removed serverPod

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
clubanderson: I thought this was the case, I think that line was in the bash.  don't remember.  removed the latter.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
clubanderson: no longer conditional, always appended now

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
clubanderson: I am not sure how to tell the difference between a failed command run and 'not ready'.  How would we tell the difference programmatically?

clubanderson:
@MikeSpreitzer how do I resolve the verify error in CI here.  It is saying something about 'make codegen'. - I ran that and committed the results in the last commit.  I don't know what this job requires to pass.

cmd/kubectl-kubestellar/cmd/kubectlKubestellar.go
clubanderson: updated

cmd/kubectl-kubestellar/main.go
clubanderson: removed

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,220 @@
...
+		Run:   getLog,
+	}
+
+	flags := cmd.Flags()
+	flags.StringVarP(&kubeconfig, "kubeconfig", "c", "", "Path to the kubeconfig file")
+	flags.BoolVarP(&followFlag, "follow", "f", false, "Follow the log output")
+	flags.StringVarP(&linesFlag, "lines", "n", "+0", "Number of lines to display")
+	flags.StringArrayVar(&kubectlFlags, "kubectl-flag", []string{}, "Additional kubectl flags")
+
+	if err := cmd.Execute(); err != nil {
```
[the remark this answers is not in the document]
clubanderson: this was my mistake in the copy and paste I performed last night - fixed..

This was what was causing my problem with running the get_log as part of the main command invocation.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
clubanderson: fixed

MikeSpreitzer:
About codegen, are you referring to https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kubestellar_kubestellar/912/pull-kubestellar-verify/1692342888744095744#1:build-log.txt%3A12 ?  That is not an error message; that is just part of what was done in the course of that checking.  The error message I see is https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kubestellar_kubestellar/912/pull-kubestellar-verify/1692342888744095744#1:build-log.txt%3A1377

MikeSpreitzer:
Following is what I see about `go.mod` and `go.sum`.  It looks like you _do_ need to `go mod tidy`.

```console
mspreitz@mjs12 kubestellar % git checkout plugins
branch 'plugins' set up to track 'aa/plugins'.
Switched to a new branch 'plugins'

mspreitz@mjs12 kubestellar % git status
On branch plugins
Your branch is up to date with 'aa/plugins'.

nothing to commit, working tree clean

mspreitz@mjs12 kubestellar % git rev-parse HEAD
3ed3702f9ae750147f5ef537e77f2e54a37013a9

mspreitz@mjs12 kubestellar % go mod tidy

mspreitz@mjs12 kubestellar % git status
On branch plugins
Your branch is up to date with 'aa/plugins'.

	modified:   go.mod
	modified:   go.sum

no changes added to commit (use "git add" and/or "git commit -a")

mspreitz@mjs12 kubestellar % git diff
diff --git a/go.mod b/go.mod
index 07a795e1..3a71a4d1 100644
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,6 @@ require (
        github.com/kcp-dev/kcp/pkg/apis v0.11.0
        github.com/kcp-dev/logicalcluster/v3 v3.0.2
        github.com/martinlindhe/base36 v1.1.1
-       github.com/muesli/reflow v0.1.0
        github.com/spf13/pflag v1.0.6-0.20210604193023-d5e0c0615ace
        github.com/stretchr/testify v1.7.1
        github.com/tidwall/sjson v1.2.5
@@ -30,7 +29,6 @@ require (
 )
 
 require (
-       github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
        github.com/Azure/go-autorest v14.2.0+incompatible // indirect
        github.com/Azure/go-autorest/autorest v0.11.28 // indirect
        github.com/Azure/go-autorest/autorest/adal v0.9.18 // indirect
@@ -75,11 +73,9 @@ require (
        github.com/jonboulle/clockwork v0.2.2 // indirect
        github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de // indirect
        github.com/mattn/go-isatty v0.0.14 // indirect
-       github.com/mattn/go-runewidth v0.0.7 // indirect
        github.com/matttproud/golang_protobuf_extensions v1.0.2-0.20181231171920-c182affec369 // indirect
        github.com/mitchellh/mapstructure v1.4.1 // indirect
        github.com/moby/spdystream v0.2.0 // indirect
-       github.com/moby/term v0.0.0-20210619224110-3f7ff695adc6 // indirect
        github.com/monochromegane/go-gitignore v0.0.0-20200626010858-205db1a8cc00 // indirect
        github.com/opencontainers/go-digest v1.0.0 // indirect
        github.com/opencontainers/runc v1.1.1 // indirect
diff --git a/go.sum b/go.sum
index 686c279a..a3eb9a4c 100644
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,3 @@
-bazil.org/fuse v0.0.0-20160811212531-371fbbdaa898/go.mod h1:Xbm+BRKSBEpa4q4hTSxohYNQpsxXPbPry4JJWOB3LB8=
-bitbucket.org/bertimus9/systemstat v0.0.0-20180207000608-0eeff89b0690/go.mod h1:Ulb78X89vxKYgdL24HMTiXYHlyHEvruOj1ZPlqeNEZM=
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
...
```

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
MikeSpreitzer: I am not sure whether there _is_ a way to tell the difference.  Note that the script being replaced here does not attempt to treat those two differently.  That means that if something else is going wrong then the command will wait forever for the ready signal that never comes.  The user will have catch on an investigate.

The better approach is to recognize that Kubernetes includes a framework by which a Pod designer can specify how readiness of that Pod is tested, and use that framework.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
MikeSpreitzer: But the problem is that `podName` is not varying.  If `podName == ""` when this loop is entered then this loop will never exit.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
MikeSpreitzer: I do not understand.  If `err == nil` after this line then you are guaranteed that `namespace == ksNamespace`.

If `err != nil` after this line because of using the wrong kubeconfig then if you were to delete the usage of `getNamespaceNames` then the invocation of `getServerPodName` would fail because of the wrong kubeconfig.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,209 @@
...
+	if kubeconfig == "" {
+		if home := homedir.HomeDir(); home != "" {
+			kubeconfig = home + "/.kube/config"
+		}
+	}
+	return clientcmd.BuildConfigFromFlags("", kubeconfig)
+}
+
+func getNamespaceNames(clientset *kubernetes.Clientset, namespace string) (string, error) {
+	nsList, err := clientset.CoreV1().Namespaces().List(context.TODO(), metav1.ListOptions{})
```
MikeSpreitzer, line 145: Getting the list of all the namespaces and looking through it for one with a given name is functionally equivalent to just doing a `Get` on the name you already know you want.

MikeSpreitzer:
Perhaps I did not explain github.com/spf13/cobra clearly enough; I am certainly unsure whether I understand exactly what you are trying to produce at this first stage of development.  After `go mod tidy`, and correcting "get_log" to "get-log" in the source (at https://github.com/clubanderson/kubestellar/blob/3ed3702f9ae750147f5ef537e77f2e54a37013a9/pkg/cliplugins/kubestellar/get_log/cmd/cmd.go#L51), and getting the `kubectl-kubestellar-get_log` bash script out of my `$PATH`, the behavior that I see is normal for cobra.  That package is about building one executable that has sub-commands and potentially sub-sub-commands and so on.  An incomplete invocation on the command line gets you an explanation of the next things you can say.

Let's compare this `kubectl kubestellar` plugin with the `kubectl kcp` plugin.

```console
mspreitz@mjs12 kubestellar % kubectl kcp
KCP is <good stuff>.

This command provides KCP specific sub-command for kubectl.

Usage:
  kcp [command]

Available Commands:
  bind        Bind different types into current workspace.
  claims      Operations related to viewing or updating permission claims
  completion  Generate the autocompletion script for the specified shell
  crd         CRD related operations
  help        Help about any command
  workload    Manages KCP sync targets
  workspace   Manages KCP workspaces

Flags:
...
```

I see parallel structure in the following.

```console
mspreitz@mjs12 kubestellar % kubectl kubestellar
KubeStellar is <good stuff>.

This command provides the kubestellar sub-command for kubectl.

Usage:
  kubestellar [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  get-log     Get the log output from a central KubeStellar or kcp process
  help        Help about any command

Flags:
...
```

Going to the next level, here is an example from the kcp side:

```console
mspreitz@mjs12 kubestellar % kubectl kcp bind
Bind different types into current workspace.

Usage:
  kcp bind [flags]
  kcp bind [command]

Available Commands:
  apiexport   Bind to an APIExport
  compute     Bind to a location workspace

Flags:
...
```

That command line was incomplete, so I got told what I can say next on the command line.

On the KS side we get to a leaf, as we should.

```console
mspreitz@mjs12 kubestellar % kubectl kubestellar get-log
Error: accepts 1 arg(s), received 0
Usage:
  kubestellar get-log [flags] which

Flags:
  -f, --follow                     Follow the log output
  -h, --help                       help for get-log
  -c, --kubeconfig string          Path to the kubeconfig file
      --kubectl-flag stringArray   Additional kubectl flags
  -n, --lines string               Number of lines to display (default "+0")

Global Flags:
      --add_dir_header                   If true, adds the file directory to the header of the log messages
      --alsologtostderr                  log to standard error as well as files (no effect when -logtostderr=true)
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory (no effect when -logtostderr=true)
      --log_file string                  If non-empty, use this log file (no effect when -logtostderr=true)
      --log_file_max_size uint           Defines the maximum size a log file can grow to (no effect when -logtostderr=true). Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
      --logtostderr                      log to standard error instead of files (default true)
      --one_output                       If true, only write logs to their native severity level (vs also writing to each lower severity level; no effect when -logtostderr=true)
      --skip_headers                     If true, avoid header prefixes in the log messages
      --skip_log_headers                 If true, avoid headers when opening log files (no effect when -logtostderr=true)
      --stderrthreshold severity         logs at or above this threshold go to stderr when writing to files and stderr (no effect when -logtostderr=true or -alsologtostderr=false) (default 2)
  -v, --v Level                          number for the log level verbosity
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging

error: accepts 1 arg(s), received 0
```

.. or, with a different error (this looks like a bug, which I will address in another comment) ...

```console
mspreitz@mjs12 kubestellar % kubectl kubestellar get-log mailbox-controller
Error building Kubernetes config: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,209 @@
...
+		Short: "Get the log output from a central KubeStellar or kcp process",
+		Args:  cobra.ExactArgs(1),
+		Run:   getLog,
+	}
+
+	flags := cmd.Flags()
+	flags.StringVarP(&kubeconfig, "kubeconfig", "c", "", "Path to the kubeconfig file")
+	flags.BoolVarP(&followFlag, "follow", "f", false, "Follow the log output")
+	flags.StringVarP(&linesFlag, "lines", "n", "+0", "Number of lines to display")
+	flags.StringArrayVar(&kubectlFlags, "kubectl-flag", []string{}, "Additional kubectl flags")
```
MikeSpreitzer, line 61: These are not connected to anything.

A user of a kubectl plugin should not have to wrap the ordinary kubectl flags.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func buildConfig() (*rest.Config, error) {
+	if kubeconfig == "" {
+		if home := homedir.HomeDir(); home != "" {
+			kubeconfig = home + "/.kube/config"
+		}
+	}
+	return clientcmd.BuildConfigFromFlags("", kubeconfig)
+}
```
MikeSpreitzer, line 142: I think we want to be invoking standard library code to do stuff like this, rather than replicate this small subset of that functionality here.

I was hoping that we could ride on the main functionality of `k8s.io/cli-runtime/pkg/genericclioptions`, not just its streams object.  See https://github.com/kubestellar/kubestellar/blob/v0.6.0/cmd/kube-watchall/main.go#L55-L56 and https://github.com/kubestellar/kubestellar/blob/v0.6.0/cmd/kube-watchall/main.go#L79 for the powerful stuff.  This is where the standard set of kubectl flags is defined and used.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
clubanderson: true - going to remove this line

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
clubanderson: replaced with Get.  just learned the difference.. thanks

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
clubanderson: ok - so nothing to change here until we update the pod design.  I can create a loop in a follow-on pr when the design changes.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
clubanderson: I would prefer to remove this loop altogether.  I don't know why it is here.  I changed this to just be an else condition

clubanderson:
@MikeSpreitzer I tried to cut over to the genericclioptions.  I created the cliopts in New, but have no idea how to reference from the get_log func.  I was looking to see if golfing had the concept of 'self' but was unable to determine if something like this exists or not.  If it did, then I could self-reference the new object that was created.  I am new to golang.  Just need some help crossing between the functions in a way other than pure arg passing.  If you pull the latest copy of this PR and try to build, you will see the dilemma here.

clubanderson:
I did 'go mod tidy' and still getting lint issue.  no idea what to try next to resolve it.

cmd/kubectl-kubestellar/cmd/kubectl-kubestellar.go
```diff
@@ -0,0 +1,63 @@
...
+			This command provides the kubestellar sub-command for kubectl.
+		`),
+		SilenceUsage:  false,
+		SilenceErrors: false,
+	}
+
+	// setup klog
+	fs := goflags.NewFlagSet("klog", goflags.PanicOnError)
+	klog.InitFlags(fs)
+	// do we want to support the GoFlagSet in addition to our own?
```
MikeSpreitzer, line 50: This kubectl plugin should support, at its base, the standard Go command line flags and the standard kubectl flags and the klog flags.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,217 @@
...
+func New(streams genericclioptions.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "get-log [flags] which",
+		Short: "Get the log output from a central KubeStellar or kcp process",
+		Args:  cobra.ExactArgs(1),
+		Run:   getLog,
+	}
+
+	fs := pflag.NewFlagSet("get-log", pflag.ExitOnError)
+	klog.InitFlags(flag.CommandLine)
```
MikeSpreitzer, line 59: If I understand cobra correctly, you do not need to define the klog flags, for example, on both the root Command and on each of its descendants.  It *might* make a difference, allowing those flags to appear either before or after the subcommand on the command line.  But I think it is conventional to define each flag at just one vertex along each leaf's path from the root, and expect the user to put each flag at the relevant position on the command line.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,217 @@
...
+	fs.AddGoFlagSet(flag.CommandLine)
+	cliOpts := genericclioptions.NewConfigFlags(true)
+	cliOpts.AddFlags(fs)
+	fs.StringVarP(&kubeconfig, "kubeconfig", "c", "", "Path to the kubeconfig file")
+	fs.BoolVarP(&followFlag, "follow", "f", false, "Follow the log output")
+	fs.StringVarP(&linesFlag, "lines", "n", "+0", "Number of lines to display")
+	fs.Parse(os.Args[1:])
```
MikeSpreitzer, line 66: The command line parsing should not be done as part of Command construction.
I suspect that a client of cobra is expected to hook into the flag framework in cobra, and cobra will invoke the Parse function at the right time.

MikeSpreitzer:
Regarding how to convey added variables to func values in Go, there are a few ways.

The simplest is to not even do that, which is plausible in this use case.  In this case you are making a command that is run just once, so there will be no concurrent use or stale use of that variable.  Put the added variable at the top (i.e., outermost) level of the file.

In general you do not have such a usage restriction and would want to pair the added variable with the func value in question.  There are a few ways to accomplish this.  One uses the fact that in Go a func value carries its closure with it.  So for example:

```go
func getLogFn(extra MyType) func (]string) {
    return func(args []string) {
        return ... stuff using args and extra ...
    }
}

func xyz(...) {...
    extra := ...
    getLogCmd := &cobra.Command{
        Run: getLogFn(extra),
        ...
    }
    ...
}
```

Another way to do it is to recognize that a method is a func and you can extract this func from the receiver like so: `receiver.methodName`.  So for example:

```go
type MyTy struct { ... }

func (mt MyTy) getLog(args []string) { ... }

func xyz(...) {...
    myTy := MyTy{...}
    getLogCmd := &cobra.Command{
        Run: myTy.getLog,
    ...
    }
}
```

MikeSpreitzer:
More importantly, it is important to understand how cobra deals with flags.  It is also important to understand how Go packages handle flags in general.  There is this concept of a "flag set", which is the definitions of a collection of flags.  Importantly, you can pour the flags from one flag set into another.  Annoyingly, I find the name and calling convention of most methods for doing so counter-intuitive.  For example, in

```go
	cliOpts := genericclioptions.NewConfigFlags(true)
	cliOpts.AddFlags(fs)
```

the second line is pouring the flag definitions from `cliOpts` into the flag set `fs` (even though, to my eye, it reads like it goes the opposite direction).

I went and looked at the cobra documentation.  It is comprehensible.  There is a discussion of flags at https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#working-with-flags .  There is an example usage of cobra --- sadly, without flags --- at https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#example .  Also, I exercised the cobra CLI generator and saved the result in a github repo: https://github.com/MikeSpreitzer/cobra-example .  If you look at the generated code there are commented-out statements working with flags.

I think what we want is for the standard kubectl flags to be added to the persistent flag set of the root command of this kubectl plugin.

clubanderson:
figured out how to use genericclioptions.  up for review.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
MikeSpreitzer: The behavior currently in this PR is different from the behavior in the bash script.  The bash script will loop until the `ls` succeeds.  I think that is the better behavior to do here too.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		klog.Error("Error getting server namespace:", err)
+		os.Exit(15)
+	}
+	podName, err := getServerPodName(clientset, namespace, ksSelector)
+	if err != nil {
+		klog.Error("Error getting server pod name:", err)
+		os.Exit(20)
+	} else {
+		podName = strings.Trim(string(podName), "'")
```
MikeSpreitzer, line 128: Do you have evidence that this Trim is necessary?  My understanding of the API used in `getServerPodName` is that it returns names without quotes around them.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
MikeSpreitzer: The parsing should not be done as part of the Command construction.
The baseline design of cobra is that cobra invokes the parsing before invoking the Command's Run or RunE method.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,219 @@
...
+
+	// Create a new pflag.FlagSet for custom flags
+	fs := pflag.NewFlagSet("get-log", pflag.ExitOnError)
+	fs.StringVarP(&kubeconfig, "kubeconfig", "c", "", "Path to the kubeconfig file")
+	fs.BoolVarP(&followFlag, "follow", "f", false, "Follow the log output")
+	fs.StringVar(&linesFlag, "lines", "+0", "Number of lines to display")
+	fs.Parse(os.Args[1:])
+
+	// Add custom pflag.FlagSet to the Cobra command
+	cmd.Flags().AddFlagSet(fs)
```
MikeSpreitzer, line 65: Nope!  That's what I meant about the name and calling convention being backwards.  This statement copies the flag definitions from `cmd.Flags()` into `fs`.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,219 @@
...
+
+	// Create a genericclioptions.ConfigFlags instance
+	cliOpts := genericclioptions.NewConfigFlags(true)
+	cliOpts.AddFlags(cmd.PersistentFlags())
```
MikeSpreitzer, line 69: The point of creating `cliOpts` is to call its `ToRESTConfig` method after the command line flags are parsed, rather than maintaining our own less functional `buildConfig`.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,212 @@
...
+	fs.StringVar(&linesFlag, "lines", "+0", "Number of lines to display")
+
+	// Add custom pflag.FlagSet to the Cobra command
+	getLogCmd.Flags().AddFlagSet(fs)
+
+	ctx := context.Background()
+	logger := klog.FromContext(ctx)
+	ctx = klog.NewContext(ctx, logger)
+	fs.VisitAll(func(flg *pflag.Flag) {
+		logger.V(1).Info("Command line flag", flg.Name, flg.Value)
```
MikeSpreitzer, line 70: This is more valuable after the command line has been parsed, which is after the Command is constructed.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,212 @@
...
+		logfile = "mailbox-controller-log.txt"
+	case "placement-translator":
+		logfile = "placement-translator-log.txt"
+	default:
+		fmt.Println("Argument 1 must be one of: kcp, where-resolver, mailbox-controller, placement-translator")
+		os.Exit(1)
+	}
+
+	cliOpts := genericclioptions.NewConfigFlags(false)
+	cliOpts.AddFlags(getLogCmd.PersistentFlags())
```
MikeSpreitzer, line 97: (1) This needs to be done before cobra parses the command line, which is before the Command's Run/RunE method is invoked.

(2) I think we want to add these flags to the cobra "persistent" flag set of the root command, because they are needed by all the subcommands of `kubectl kubestellar`.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,216 @@
...
+	fs.StringVar(&linesFlag, "lines", "+0", "Number of lines to display")
+
+	// Add custom pflag.FlagSet to the Cobra command
+	getLogCmd.Flags().AddFlagSet(fs)
+
+	cliOpts.AddFlags(getLogCmd.PersistentFlags())
+
+	ctx := context.Background()
+	logger := klog.FromContext(ctx)
+	ctx = klog.NewContext(ctx, logger)
```
MikeSpreitzer, line 74: Nothing uses the context and logger constructed here.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,216 @@
...
+var (
+	kubeconfig string
+	followFlag bool
+	linesFlag  string
+	whichArg   string
+	logger     klog.Logger
+)
+
+func New(streams genericclioptions.IOStreams) *cobra.Command {
+	cliOpts := genericclioptions.NewConfigFlags(false)
```
MikeSpreitzer, line 50: You do not want to make another one of these, you want to use the one already made in `func KubectlKubestellarCommand()`.

clubanderson:
@MikeSpreitzer can I get another review here to get this done before sprint ends on Thursday?

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,212 @@
...
+		os.Exit(30)
+	}
+
+	getLogCmd.Flags().VisitAll(func(flg *pflag.Flag) {
+		logger.V(1).Info(fmt.Sprintf("Command line flag %s %s", flg.Name, flg.Value))
+	})
+
+}
+
+func buildConfig() (*rest.Config, error) {
```
MikeSpreitzer, line 144: This frunc is not used, right?  If so then I would delete it.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,212 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/remotecommand"
+	"k8s.io/client-go/util/homedir"
+	"k8s.io/klog/v2"
+)
+
+var (
+	kubeconfig string
```
MikeSpreitzer, line 42: Nothing consumes this variable, right?  If so then I would remove it.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		logger.Error(err, "Error getting server pod name:")
+		os.Exit(20)
+	} else {
+		podName = strings.Trim(string(podName), "'")
+	}
+
+	err = executeCommandInPod(clientset, namespace, config, podName, []string{"ls", "/home/kubestellar/ready"}, os.Stdin, os.Stdout, os.Stderr)
+	if err != nil {
+		logger.Error(err, "Error executing command in pod:")
```
MikeSpreitzer, line 122: As far as the user is concerned, the important point here is that the pod is not ready.  I would make that the primary message here.

pkg/cliplugins/kubestellar/get_log/cmd/cmd.go
```diff
@@ -0,0 +1,212 @@
...
+
+	err = executeCommandInPod(clientset, namespace, config, podName, kubectlLogArgs, os.Stdin, os.Stdout, os.Stderr)
+	if err != nil {
+		logger.Error(err, "Error executing command in pod:")
+		os.Exit(30)
+	}
+
+	getLogCmd.Flags().VisitAll(func(flg *pflag.Flag) {
+		logger.V(1).Info(fmt.Sprintf("Command line flag %s %s", flg.Name, flg.Value))
+	})
```
MikeSpreitzer, line 140: I would put this logging early in this func, before any `return` or `os.Exit`, so that the user has a chance to get more clues about what is going wrong (in the cases where something goes wrong).

MikeSpreitzer:
I think the command line flag handling is good now.  There remain two issues.

This behaves differently than the script it aims to replace, erring out when the pod is not ready rather than waiting for the pod to be ready.

I assume this is just the start of a campaign of development to replace all the `kubectl kubestellar` plugin scripts with go code.  I think that would be great.  But there is a problem with releasing a partially completed campaign: a user who explores the plugin command structure by saying just `kubectl kubestellar` will be misinformed about all the sub-(sub-...)commands they can issue.  For example, at the current level of development, this makes a user who says `kubectl kubestellar` get told that `kubectl kubestellar get-log` is available and no other sub-commands exist.  I recommend that at interim points in this campaign we not do that.  Two approaches immediately occur to me.  One is to have `make build` not put `kubectl-kubestellar` in `bin/`.  The other is to temporarily use a different name until the campaign is complete, perhaps `kubectl-ks` or `kubectl-kubestellar2` or `kubectl-kubestellargo`.

MikeSpreitzer:
If you want to see an example of a plugin that is a leaf executable, see https://github.com/kubestellar/kubestellar/tree/main/cmd/kubectl-kubestellar-syncer_gen , which implements `kubectl kubestellar syncer-gen`.

clubanderson:
closing due to #1097

PR #918, opened 2023-08-18
✨ brew formulas for kubestellar, kcp, and kcp_kubectl

### LINKED ISSUE #908: feature: use Brew to install KubeStellar plugins
### Feature Description

I would like to be able to install KubeStellar as a container (perhaps using helm), and then install plugins for kubectl easily

### Proposed Solution

create a package for krew that automates the installation of our kubestellar plugins

### Want to contribute?

### Additional Context

_No response_

clubanderson:
## Related issue(s)

Fixes #


brew/kubestellar-0.6.0.rb
```diff
@@ -0,0 +1,107 @@
...
+      when :arm
+        home_directory = File.join(ENV["HOMEBREW_PREFIX"])
+        target_directory = File.join(home_directory, 'kubestellar')
+        # puts "HELLO THERE #{target_directory}"
+
+        bin.install Dir["bin/*"]
+
+        system "osascript", "-e", <<-EOS
+          do shell script "mkdir -p '#{target_directory}'" with administrator privileges
+          do shell script "cp -r config user LICENSE README.md '#{target_directory}'" with administrator privileges
```
MikeSpreitzer, line 69: This looks identical to the other case?  Do we need a case statement here?

MikeSpreitzer:
I am not familiar with authoring brew.

In fact, I am a little confused about using it.  There are casks, bottles, kegs, and more.  And when I successfully install something in the easy case, it does not tell me that what I have done involves any of those words.  `brew list` output for me has a huge list of "Formulae" and one Cask.

I know that brew has a normal case and also things that are more hesitant or install in alternate places to avoid overwriting stuff from the user's OS vendor.

I am not sure of the words to use when authoring something for the normal case.

I think we should be using the normal case for installing KubeStellar.  kcp is more debatable, but I think it is practically OK to use the normal case for that too.

The next thing to be clear about is where things will go.  For brew on my Mac, the foobar package normally goes in /usr/local/Cellar/foobar/ and brew also adds symlinks to /usr/local/bin pointing to executables in /usr/local/Cellar/foobar/2.3.5/bin (if 2.3.5 is my current version of foobar).  A KubeStellar distribution has not only executables in bin/ but also related stuff in config/ and user/ that the executables expect to find adjacent to them.

MikeSpreitzer:
I demonstrated how symlinks interact with current directory in exectables.  I made a symlink and then invoked it.  The script looked where I expected.

```console
mspreitz@mjs12 ~ % ls -l /usr/local/bin/kubectl-kubestellar-deploy
lrwxr-xr-x  1 mspreitz  admin  88 Aug 18 17:05 /usr/local/bin/kubectl-kubestellar-deploy -> /Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin/kubectl-kubestellar-deploy

mspreitz@mjs12 ~ % kubectl-kubestellar-deploy -X --external-endpoint foo.bar:1234
...
+ helm upgrade -i kubestellar /usr/local/user/helm-chart --set clusterType=Kubernetes --set-string EXTERNAL_HOSTNAME=foo.bar --set EXTERNAL_PORT=1234 --set image.tag=built-2023-08-08-14-55-25
```

KPRoche:
I'm not a MacOS user, so can't comment intelligently on the commands there.
Couple items:

I realize this is the first rough attempt, but I'd suggest your readme should mention/link to the install instructions for homebrew itself. (Only bringing it up because one of the instructors in the Cloud SGI presumed everyone already used brew, and that session got sidetracked significantly as a result.)

I see that in the section of the formula preceding the def install block you used the same kind of conditionals as you used inside def install, rather than the on_xxx blocks. Was this just for simplicity in your first run at a formula?
(I'm tearing through the brew documentation to understand better)

brew/kubestellar-0.6.0.rb
clubanderson: removed

clubanderson:
> I'm not a MacOS user, so can't comment intelligently on the commands there. Couple items:
> 
> I realize this is the first rough attempt, but I'd suggest your readme should mention/link to the install instructions for homebrew itself. (Only bringing it up because one of the instructors in the Cloud SGI presumed everyone already used brew, and that session got sidetracked significantly as a result.)
> 
> I see that in the section of the formula preceding the def install block you used the same kind of conditionals as you used inside def install, rather than the on_xxx blocks. Was this just for simplicity in your first run at a formula? (I'm tearing through the brew documentation to understand better)

what are 'on_xxx blocks?

clubanderson:
> I am not familiar with authoring brew.
> 
> In fact, I am a little confused about using it. There are casks, bottles, kegs, and more. And when I successfully install something in the easy case, it does not tell me that what I have done involves any of those words. `brew list` output for me has a huge list of "Formulae" and one Cask.
> 
> I know that brew has a normal case and also things that are more hesitant or install in alternate places to avoid overwriting stuff from the user's OS vendor.
> 
> I am not sure of the words to use when authoring something for the normal case.
> 
> I think we should be using the normal case for installing KubeStellar. kcp is more debatable, but I think it is practically OK to use the normal case for that too.
> 
> The next thing to be clear about is where things will go. For brew on my Mac, the foobar package normally goes in /usr/local/Cellar/foobar/ and brew also adds symlinks to /usr/local/bin pointing to executables in /usr/local/Cellar/foobar/2.3.5/bin (if 2.3.5 is my current version of foobar). A KubeStellar distribution has not only executables in bin/ but also related stuff in config/ and user/ that the executables expect to find adjacent to them.

if I knew how to have that 'default' location to install to, I would use it.  The default behavior for brew is to "build" something that comes from a tar.gz - like our source tarball.  We are not compiling anything, we have binaries and shells in our tarball.  So we have to have a destination for them.  I don't know a way to get the 'cellar' location that comes from a brew install build.  I will research more.  The likelihood is to put files in some obscure location and then put a logical link to it to simulate what a "build" would do.

clubanderson:
I see something called a bottle which is for binary-only installs.  I will experiment with it.

KPRoche:
> what are 'on_xxx blocks?

They are another way, in the section of the formula outside the def install block, to specify actions based on system configuration: https://docs.brew.sh/Formula-Cookbook#handling-different-system-configurations

`Often, formulae need different dependencies, resources, patches, conflicts, deprecations or keg_only statuses on different OSes and architectures. In these cases, the components can be nested inside on_macos, on_linux, on_arm or on_intel blocks. For example, here’s how to add gcc as a Linux-only dependency:

    on_linux do
      depends_on "gcc"
    end

`

clubanderson:
@KPRoche read further on on_xxx - its for config, not install - https://docs.brew.sh/Formula-Cookbook#inside-def-install-and-test-do.

need to use the if, then, else in the install blocks.  So I am going to keep it consistent and use what I have now in both install and config.  Thank you for having a look and suggesting this approach though.

MikeSpreitzer:
It would help me a lot to see the intended user documentation for this.  I see there is now a README with command lines in it.  It would help a lot to also see an explanation of what the installation command is intended to accomplish.  What it will put where, what it will run using systemd or Kubernetes or raw processes or whatever.

clubanderson:
@MikeSpreitzer included some preliminary docs to get started.  Some of the KubeStellar commands make assumptions about where they are being executed.  "KubeStellar init" assumes it is being run in the directory along side the config path.  Also, "kubectl kubestellar deploy" assumes the helm chart is located alongside it in the helm path.  These should be updated to allow for the KubeStellar or kubectl KubeStellar to be run anywhere on the file system - taking the path for config or helm chart as a parameter

clubanderson:
closing in favor of #1140

PR #940, opened 2023-08-28
:bug: Fix Syncer not to update status field

### LINKED ISSUE #938: bug: periodically removes status of a downsync propagated AppWrapper instance
### Describe the bug

We have a deployment when KS downsyncs `AppWrapper` objects from a hub cluster to spokes.
The status field of the AppWrapper objects is periodically removed. 
I suspect that MCAD controller sets the status, and Kubestellar removes it. (When I stopped KS, I did not observe the issue; after restarting KS, the issue returned.).

### Steps To Reproduce

go to the WEDC2 cluster and several times check the status AppWrappers objects in the `test-kfp` ns.  

### Expected Behavior

The status entry should not be deleted. 

### Additional Context

_No response_

yana1205:
## Summary

I added a code to explicitly set `status` field and `meta.managedField` to the same value of the existing object at the destination when downsyncing objects in order to avoid overwriting the status field just in case.

## Related issue(s)

Fixes #938

### DIFF
--- a/pkg/syncer/syncers/common.go
+++ b/pkg/syncer/syncers/common.go
@@ -180,7 +180,7 @@ func diff(logger klog.Logger, srcResourceList *unstructured.UnstructuredList, de
 		if ok {
 			srcResource.SetResourceVersion(destResource.GetResourceVersion())
 			srcResource.SetUID(destResource.GetUID())
-			srcResource.SetManagedFields(nil)
+			srcResource.SetManagedFields(destResource.GetManagedFields())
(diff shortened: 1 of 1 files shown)


clubanderson (commented):
/lgtm
/approve

MikeSpreitzer:
@yana1205 , can you please go to https://github.com/yana1205/kubestellar/actions/workflows/docs-ecutable-example1.yml and ask for that workflow to run on the branch in this PR?

Issue #945, opened 2023-08-28
bug: incorrect way of updating status of an object


MikeSpreitzer:
### Describe the bug

The correct way to update the `status` field of an object depends on whether that kind of object has a "status subresource".  Neither the placement-translator nor the syncer has a general way of knowing which is which.

### Steps To Reproduce

Try propagating status upward from a kind of object that the syncer does not handle correctly.

### Expected Behavior

Status propagates upward for all kinds of objects.

### Additional Context

#944 is nearby but distinct.

clubanderson:
can we close this since it just a PSA?

PR #946, opened 2023-08-28
🐛 release-0.4: Choose update status method according to API group

MikeSpreitzer:
## Summary
This PR updates the syncer to make it choose how to update the status of an upstream object based on the API group.  This is better than the previous code but not fully correct; the correct code would somehow depend on whether the resource actually has a status subresource.

## Related issue(s)

Partially addresses #945


MikeSpreitzer:
/cc @waltforme 
@dumb0002 
@cwiklik

MikeSpreitzer:
This is the back-level version of #947

PR #947, opened 2023-08-28
🐛 Choose update status method according to API group

MikeSpreitzer:
## Summary
This PR updates the syncer to make it choose how to update the status of an upstream object based on the API group. This is better than the previous code but not fully correct; the correct code would somehow depend on whether the resource actually has a status subresource.

## Related issue(s)

This partially addresses #945 .

### DIFF
--- a/pkg/syncer/syncers/downsyncer.go
+++ b/pkg/syncer/syncers/downsyncer.go
@@ -19,6 +19,7 @@ package syncers
 import (
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
/cc @waltforme 
@dumb0002 
@cwiklik

yana1205:
Thanks
/lgtm

MikeSpreitzer:
BTW, successful test in https://github.com/MikeSpreitzer/kcp-edge-mc/actions/runs/6003086992

Issue #931, opened 2023-08-22
feature: Document space framework for users of it


MikeSpreitzer:
### Feature Description

Users of the space framework need a document addressed to them.  I, as a reviewer of PRs changing space functionality consumers to use the space framework, need to understand how the space framework is intended to be used.

While the Go compiler does a certain amount of checking for consistency with the APIs already checked in, an author of code using the space framework needs to know much more than what is written in the Go code and checked by the compiler.

Clues can be found in my comments in my reviews of the PRs currently in flight: #897, #904, and #923.  Also the high level issues such as #905, #906, #913, #914, #928 .

Such a document will address the following issues, among others.

- Division of roles and responsibilities; who makes what decisions when, how are those decisions recorded by the decision-maker and conveyed to decision consumers.
- What needs to be configured in code that uses the space framework.
- How space names correspond to kcp workspaces paths, when the kcp provider is used.
- An architectural outline, with clear and distinct identification of process boundaries and library boundaries, boundaries between who owns/authors/maintains what.

Given the current extreme time crunch, I do not expect everything will be done to our complete satisfaction at first.  The needed document will at least be a roadmap.  It might call out forks in the road ahead, where we have decisions still to make.

### Proposed Solution

This document gets written.

### Want to contribute?

- [ ] I would like to work on this issue.

MikeSpreitzer:
@effi-ofer 
@andreyod 
@clubanderson

clubanderson:
pushing to backlog, not committed for milestone1 - per @ezrasilvera

Issue #840, opened 2023-07-30
Epic: Convert KubeStellar to work with space provider instead of directly with KCP workspaces


ezrasilvera:
This feature describes the steps needed in order to switch from using KCP workspaces directly to using spaces through the space-manager. The following plan doesn't depend on new kube-bind support for apibind, apiexport, and apiexport view. KS can continue using KCP's export/bind. 
This transition doesn't impact the interaction with KubeStellar - **no changes are needed to anyone using KS**
There are two dependent phases:

1.  Changes to the KS setup process and scripts: bring up  the space manager, create and configure a KCP space provider
2. Gradually convert the KS controllers that are interacting with Workspaces through the cluster scoped KCP go-client into using the multi-space aware client (msclien)

The plan includes the following tasks and changes to KS:
## prerequisites 
- [x] Working KCP space provider  #849 
- [x]  Merge the space framework and multi-space client library into KS

### Related features and PRs:
- [ ] #978
- [ ] #977
- [ ] #976 
- [ ] #975 
- [ ] #852 
  - [ ] #897 
- [ ] #853 
  - [ ] #904 
  - [ ] #923

effi-ofer:
What is WS workspace?  Did you mean KS Workspace?

clubanderson:
placing in sprint 3

MikeSpreitzer:
I assume that in "Update the KS setup: setup the space framework", in the first bullet, the intent is to talk about the KubeStellar Core Space.  BTW, we _do_ have a glossary now, in https://docs.kubestellar.io/main/Getting-Started/user-guide/#glossary ; further improvements are solicited.

MikeSpreitzer:
In "Update the KS setup: setup the space framework", in the last bullet, about creating the other spaces: I expect that we can start with i; obviously we need to eventually get to ii.

ezrasilvera:
@MikeSpreitzer I have updated the flow. We now have a bootstrapping space that is used only by the space-manager and is not crated through the space-manager. (Originally the flow used the KS core space for that). 
I suggest we actually start directly with creating the spaces through the space resources  rather than letting the space manager import KCP workspaces. This way we won't need to change anything when we switch to another provider.

ezrasilvera:
The epic moved into the following epic https://github.com/kubestellar/kubestellar/issues/524

PR #952, opened 2023-08-29
:sparkles: Add a space-framework.md file

ezrasilvera:
This file should eventually become the design doc of the space-framework. For now it contains the text from the google doc maintaining the design doc. 

## Related issue(s)

Fixes #

### DIFF
--- a/space-provider/space-framework.md
+++ b/space-provider/space-framework.md
@@ -1,0 +1,117 @@
+This document includes the text from the space-framework design document
+
+The full design document can be found here: https://docs.google.com/document/d/1WZCxlGoaCZVymqMhLCa5JZeURdCTzju8C5ABS8V4lW0/edit?usp=sharing
+
+
+# Space support for KubeStellar
+
+## Background
+
+KubeStellar architecture leverages spaces as the container building block for various purposes. For example, the separation between the KubeStellar service provider and the inventory is done by using different spaces for the inventory and the service provider. 
+
+The KubeStellar design currently 
+
+1. Separate different user workloads into different spaces (workload management spaces)
+2. Synchronise each EdgeCluster into a dedicated space (called the mailbox space ). 
+
+In the future some of the above design may change (e.g., the need for  1:1 mapping  of MBLC:EdgeCluster or the need for separate WMLC for each workload), however we believe some usage of spaces will still be needed.
+
+Today KubeStellar is using KCP workspaces as spaces and is practically built on top of KCP. 
+
+This proposal tries to disconnect KubeStellar architecture and internal design and the underlying support for spaces. 
+
+The main concept of the proposal is to define a space provider interface that the KubeStellar will be able to consume without being exposed to the internal implementation details of the Space provider itself. 
+
+In addition, in order to allow an easy usage of the spaces by clients (e.g., controllers, etc..) we introduce a cluster-aware client that will be able to retrieve the needed cluster information by simply specifying the cluster ID (basically the cluster name). 
+
+
+## Space provider framework
+
+ill be created/deleted by the
+
+The diagram illustrates the high level architecture of the space framework. The framework supports two types of spaces:
+
+1. Managed Space : This cluster is created through an explicit request to the management api server to create a new Space. As a result the Space  manager will initiate a request to the Space  provider to create the actual cluster. The desired state of the managed Space  cluster is defined through the Space object. For example, in order to delete this cluster you need to simply delete the corresponding Space object. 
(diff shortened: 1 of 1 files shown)


space-provider/space-framework.md
```diff
@@ -0,0 +1,117 @@
+This document includes the text from the space-framework design document
+
+The full design document can be found here: https://docs.google.com/document/d/1WZCxlGoaCZVymqMhLCa5JZeURdCTzju8C5ABS8V4lW0/edit?usp=sharing
+
+
+# Space support for KubeStellar
+
+## Background
+
+KubeStellar architecture leverages spaces as the container building block for various purposes. For example, the separation between the KubeStellar service provider and the inventory is done by using different spaces for the inventory and the service provider.
```
MikeSpreitzer, line 10: "container" is potentially confusing here, since the sort involving Linux namespaces and cgroups is so close nearby.

ezrasilvera: I still prefer at this point to work through comments on the google doc until we just move the design doc to here as MD file.

space-provider/space-framework.md
```diff
@@ -0,0 +1,117 @@
...
+
+# Space support for KubeStellar
+
+## Background
+
+KubeStellar architecture leverages spaces as the container building block for various purposes. For example, the separation between the KubeStellar service provider and the inventory is done by using different spaces for the inventory and the service provider. 
+
+The KubeStellar design currently 
+
+1. Separate different user workloads into different spaces (workload management spaces)
```
MikeSpreitzer, line 14: Since this is a new document, I would prefer that we not create technical debt by using old terminology or create confusion by using mixed terminology.  We have a glossary now, see https://docs.kubestellar.io/main/Getting-Started/user-guide/#glossary .  If you mean to suggest a change in terminology please explicitly raise that point, as a distinct issue.

space-provider/space-framework.md
```diff
@@ -0,0 +1,117 @@
...
+2. Synchronise each EdgeCluster into a dedicated space (called the mailbox space ). 
+
+In the future some of the above design may change (e.g., the need for  1:1 mapping  of MBLC:EdgeCluster or the need for separate WMLC for each workload), however we believe some usage of spaces will still be needed.
+
+Today KubeStellar is using KCP workspaces as spaces and is practically built on top of KCP. 
+
+This proposal tries to disconnect KubeStellar architecture and internal design and the underlying support for spaces.
```
MikeSpreitzer, line 21: Everything is connected to everything, more or less directly.  I find talking about "disconnect" here to be confusing.  I am more comfortable with words about adding an intermediary that allows the multi-cluster configuration management code to consume space functionality from a variety of providers of that functionality.

Remember also the issue of the scope of the name "KubeStellar".  As long as it is a large scope that encompasses the space framework then it is not a narrow thing refers only to the multi-cluster configuration management.

MikeSpreitzer:
I am not quite sure I understand the purpose of this document now.  At the moment it points to the design document and has a copy of a subset of that document?  I am not sure I understand the point of having a copy of a subset.  Is this something that will change very shortly, or are we talking about a long-term evolution?  When there are two copies of text this creates the problems of keeping them identical and which one do I review and where to readers go and why.

MikeSpreitzer:
The existing document describes the space framework and that is great; we also need documents that address bigger pictures.  We need something that has the abstraction layer, space provider(s), provider adapter(s), kube-bind, and multi-cluster configuration management.  We need a document that outlines an incremental development plan, in more precision and detail than I currently see in #840 .  Or perhaps the issue is that I expect different substance.  I think we need a plan that has modest sized steps that can be merged into the `main` branch while maintaining the fact that said branch works for our current use cases.  This will involve first generalizations rather than "instead of" changes; old functionality can only start to be removed after we are satisfied with the new functionality.

MikeSpreitzer:
The title of the linked Google doc looks mangled; it currently reads "cluster providerThe".

MikeSpreitzer:
I think we need a high-level picture of how the software components relate, such as https://docs.google.com/drawings/d/1U5nwzHeSdpEufzOPmYX_8cIvNOUzRa3oOFuKqNS4smM .  That shows two things we might aim for.  The current reality is different, that needs to get the same sort of drawing.

MikeSpreitzer:
Please also consider #931

space-provider/space-framework.md
ezrasilvera: Will verify

effi-ofer: @MikeSpreitzer reading the glossary, I am not clear on whether a workload descriptor space is the same as a space or is something different?

clubanderson:
@ezrasilvera to converge

ezrasilvera:
Replaced with #1251

PR #963, opened 2023-09-04
🐛 Cope with simplicity of SimpleClientset

MikeSpreitzer:
## Summary
This PR makes a fix to the unit test of the mailboxwatch patckage.  The fix updates the checking of the Workspace informer behavior to cope with the fact that the fake clientset constructed by NewSimpleClientset has a dumb implementation of WATCH: it does not deliver the pre-existing objects.

This test still fails.  A full fix would require making a fake clientset whose LIST and WATCH behavior have full fidelity to ResourceVersion semantics.  The use of those semantics is an important part of what we want to test.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/mailboxwatch/impl_test.go
+++ b/pkg/mailboxwatch/impl_test.go
@@ -22,11 +22,12 @@ import (
 	"math"
 	"math/rand"
 	"os"
+	"strconv"
 	"sync"
 	"testing"
 	"time"
 
-	k8sapps "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


waltforme:
@MikeSpreitzer Should this PR be reviewed now or be reviewed when the 'full fix' is done?

MikeSpreitzer:
@waltforme: I do not think that the full fix is a high priority for us, so I would like to get this reviewed now.

waltforme:
@MikeSpreitzer OK, will do.

Issue #964, opened 2023-09-04
bug: KCP provider stops getting events after some time


andreyod:
### Describe the bug

KCP provider stops getting events from KCP server after approximately 1 hour.

### Steps To Reproduce

1. Strat KCP and space-manager
2. Create provider object for KCP
3. Run for more than 1 hour
4. Create Space object

### Expected Behavior

Provider should get create event and update the space-manager about it.

clubanderson:
@andreyod is there an issue open in KCP (assuming this is a bug for them)?

andreyod:
This is for our space-framework KCP provider.
Fixed by https://github.com/kubestellar/kubestellar/pull/966
Closing.

PR #966, opened 2023-09-05
:bug: Use informer in KCP space provider

andreyod:
## Summary
This PR will change KCP space provider to use LogicalCluster informer instead of `watch` API call which is unstable. 
## Related issue(s)

Fixes #
https://github.com/kubestellar/kubestellar/issues/964

### DIFF
--- a/space-provider/kcp/kcp.go
+++ b/space-provider/kcp/kcp.go
@@ -18,24 +18,32 @@ package providerkcp
 
 import (
 	"context"
+	"errors"
+	"reflect"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/go-logr/logr"
 
 	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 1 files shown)


space-provider/kcp/kcp.go
```diff
@@ -193,69 +202,194 @@ func (k *KcpWatcher) ResultChan() <-chan clusterprovider.WatchEvent {
...
+			Name:      owner.Name,
+			SpaceInfo: spaceInfo,
+		}
+	}
+	if ok && lc.Status.Phase != kcpcorev1alpha1.LogicalClusterPhaseReady {
+		c.logger.V(2).Info("KCP workspace is not ready")
+		if ok {
+			delete(c.watcher.provider.workspaces, path)
+			c.watcher.ch <- clusterprovider.WatchEvent{
+				Type: watch.Deleted,
```
effi-ofer, line 374: Can you clarify please why we are deleting the workspace and sending the watcher a delete event if the kcp logical cluster phase is not ready?

andreyod: This is the case when ready workspace became not ready for some reason. In this case we want to remove this workspace from the list of ready workspaces and send an event to the space-manager to set the space object phase to not ready.

PR #968, opened 2023-09-06
:sparkles: Syncer supports NamespacedObjects in syncerconfig

### LINKED ISSUE #898: feature: support for new selection criteria for downsynced objects
### Feature Description

There have been updates to the selection criteria for objects that can be found in https://github.com/kubestellar/kubestellar/pull/885 and https://github.com/kubestellar/kubestellar/pull/884/files.

### Proposed Solution

update the syncer to take namespaced selection of downsynced objects and also specific objects within a namespace.

@yana1205 please review and update the syncer code accordingly.

### Want to contribute?

### Additional Context

_No response_

yana1205:
## Summary

- Extend Syncer to support newly introduced namespacedObject field in SyncerConfig
  - It keeps backward compatibility and also supports the new format.
- Add test cases
  - [Existing SyncerConfig](https://github.com/yana1205/kubestellar/blob/cd9892c5abfaf910cc9a0316cc0852e65ee017d3/test/e2e/kubestellar-syncer/testdata/kyverno/syncer-config.yaml) (used namespaceScoped field) for backward compatibility
  - [New SyncerConfig](https://github.com/yana1205/kubestellar/blob/cd9892c5abfaf910cc9a0316cc0852e65ee017d3/test/e2e/kubestellar-syncer/testdata/kyverno/syncer-config-namespaced-objects.yaml) (used namespacedObject field)
  - [Union](https://github.com/yana1205/kubestellar/blob/cd9892c5abfaf910cc9a0316cc0852e65ee017d3/test/e2e/kubestellar-syncer/testdata/kyverno/syncer-config-namespaced-objects-coexist.yaml) (used both namespaceScoped and namespacedObject) just in case if both are given. Though not sure this is a real use case.
  - [Another complex case](https://github.com/yana1205/kubestellar/blob/cd9892c5abfaf910cc9a0316cc0852e65ee017d3/test/e2e/kubestellar-syncer/testdata/namespaced-objects/syncer-config.yaml) (used both namespaceScoped and namespacedObject)

#### Verification

UT passed:
```
$ make test-syncer
go test -count 1 `go list ./... | grep "/pkg/cliplugins\|/pkg/syncer"`
?       github.com/kubestellar/kubestellar/pkg/syncer   [no test files]
?       github.com/kubestellar/kubestellar/pkg/syncer/clientfactory     [no test files]
?       github.com/kubestellar/kubestellar/pkg/syncer/shared    [no test files]
?       github.com/kubestellar/kubestellar/pkg/syncer/syncers   [no test files]
ok      github.com/kubestellar/kubestellar/pkg/cliplugins/kubestellar/syncer-gen        0.394s
ok      github.com/kubestellar/kubestellar/pkg/syncer/controller        17.585s
```

E2E passed:
```
$ make e2e-test-kubestellar-syncer
...
ok      github.com/kubestellar/kubestellar/test/e2e/kubestellar-syncer  255.863s
```

Verified manual test (example1) too.

## Related issue(s)

Fixes #898

### DIFF
--- a/pkg/syncer/controller/syncerconfigcontroller_test.go
+++ b/pkg/syncer/controller/syncerconfigcontroller_test.go
@@ -73,63 +73,82 @@ var testAPIResourceList = []*metav1.APIResourceList{
 	},
 }
 
-func TestSyncerConfig(t *testing.T) {
-	tests := []struct {
-		description      string
-		op               string
-		syncerConfig     *edgev1alpha1.SyncerConfig
-		syncerConfigSpec edgev1alpha1.SyncerConfigSpec
-		expected         Expected
-	}{
-		{
-			description:  "Syncer updates downsyncer/upsyncer and synced resources following to syncConfig",
-			syncerConfig: syncerConfig("test-sync-config", types.UID("uid")),
-			syncerConfigSpec: edgev1alpha1.SyncerConfigSpec{
-				NamespaceScope: edgev1alpha1.NamespaceScopeDownsyncs{
-					Namespaces: []string{"default"},
-					Resources: []edgev1alpha1.NamespaceScopeDownsyncResource{
-						{
-							GroupResource: metav1.GroupResource{Group: "", Resource: "configmaps"},
-							APIVersion:    "v1",
-						},
-					},
+type typeSyncerConfigTest struct {
+	description      string
+	syncerConfig     *edgev1alpha1.SyncerConfig
+	syncerConfigSpec edgev1alpha1.SyncerConfigSpec
+	expected         Expected
+}
+
+var test1 typeSyncerConfigTest = typeSyncerConfigTest{
+	description:  "Syncer updates downsyncer/upsyncer and synced resources following to syncConfig",
+	syncerConfig: syncerConfig("test-sync-config", types.UID("uid")),
+	syncerConfigSpec: edgev1alpha1.SyncerConfigSpec{
+		NamespaceScope: edgev1alpha1.NamespaceScopeDownsyncs{
+			Namespaces: []string{"default"},
+			Resources: []edgev1alpha1.NamespaceScopeDownsyncResource{
+				{
+					GroupResource: metav1.GroupResource{Group: "", Resource: "configmaps"},
+					APIVersion:    "v1",
 				},
-				ClusterScope: []edgev1alpha1.ClusterScopeDownsyncResource{
-					{
-						GroupResource: metav1.GroupResource{Group: "cheese.testing.k8s.io", Resource: "cheddars"},
-						APIVersion:    "v1",
-						Objects:       nil,
-					},
-				},
-				Upsync: []edgev1alpha1.UpsyncSet{
-					{
-						APIGroup:   "cheese.testing.k8s.io",
-						Resources:  []string{"*"},
-						Namespaces: []string{"*"},
-						Names:      []string{"*"},
-					},
-				},
-			},
-			expected: Expected{
-				downSyncedResources: []edgev1alpha1.EdgeSyncConfigResource{
-					{Group: "", Version: "v1", Kind: "ConfigMap", Namespace: "default", Name: "*"},
-					{Group: "", Version: "v1", Kind: "Namespace", Name: "default"},
-					{Group: "cheese.testing.k8s.io", Version: "v1", Kind: "Cheddar", Name: "*"},
-				},
-				upSyncedResources: []edgev1alpha1.EdgeSyncConfigResource{
-					{Group: "cheese.testing.k8s.io", Version: "v1", Kind: "Cheddar", Namespace: "*", Name: "*"},
-					{Group: "cheese.testing.k8s.io", Version: "v1", Kind: "Gouda", Name: "*"},
-					{Group: "cheese.testing.k8s.io", Version: "v27alpha15", Kind: "Cheddar", Namespace: "*", Name: "*"},
-					{Group: "cheese.testing.k8s.io", Version: "v27alpha15", Kind: "Gouda", Name: "*"},
-					{Group: "", Version: "v1", Kind: "Namespace", Name: "*"},
-				},
-				conversions: []edgev1alpha1.EdgeSynConversion{{
-					Upstream:   upSyncedResource,
-					Downstream: downSyncedResource,
-				}},
 			},
 		},
-	}
+		ClusterScope: []edgev1alpha1.ClusterScopeDownsyncResource{
+			{
+				GroupResource: metav1.GroupResource{Group: "cheese.testing.k8s.io", Resource: "cheddars"},
+				APIVersion:    "v1",
+				Objects:       nil,
+			},
+		},
+		Upsync: []edgev1alpha1.UpsyncSet{
+			{
+				APIGroup:   "cheese.testing.k8s.io",
+				Resources:  []string{"*"},
+				Namespaces: []string{"*"},
+				Names:      []string{"*"},
+			},
+		},
+	},
+	expected: Expected{
+		downSyncedResources: []edgev1alpha1.EdgeSyncConfigResource{
+			{Group: "", Version: "v1", Kind: "ConfigMap", Namespace: "default", Name: "*"},
+			{Group: "", Version: "v1", Kind: "Namespace", Name: "default"},
+			{Group: "cheese.testing.k8s.io", Version: "v1", Kind: "Cheddar", Name: "*"},
+		},
+		upSyncedResources: []edgev1alpha1.EdgeSyncConfigResource{
+			{Group: "cheese.testing.k8s.io", Version: "v1", Kind: "Cheddar", Namespace: "*", Name: "*"},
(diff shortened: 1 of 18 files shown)


pkg/syncer/controller/syncerconfigmanager.go
```diff
@@ -129,6 +132,99 @@ func (s *SyncerConfigManager) upsertNamespaceScoped(syncerConfig edgev1alpha1.Sy
 	s.syncConfigManager.upsert(edgeSyncConfig)
 }
 
+type typeFlatNamespacedObject struct {
+	APIVersion string
```
MikeSpreitzer, line 136: Since "APIVersion" is commonly used elsewhere in the Kubernetes milieu to mean the combination of API group and version, it is worth a comment here saying that only the version part appears in this field.

pkg/syncer/controller/syncerconfigmanager.go
```diff
@@ -129,6 +132,99 @@ func (s *SyncerConfigManager) upsertNamespaceScoped(syncerConfig edgev1alpha1.Sy
 	s.syncConfigManager.upsert(edgeSyncConfig)
 }
 
+type typeFlatNamespacedObject struct {
```
MikeSpreitzer, line 135: As a general rule, I do not think it is helpful in the Go programming language to put the word "type" in the name of a type.  Go is syntactically stratified, a type name can only appear where the name must name a type.  Is there a reason for an exception here?

pkg/syncer/controller/syncerconfigmanager.go
```diff
@@ -129,6 +132,99 @@ func (s *SyncerConfigManager) upsertNamespaceScoped(syncerConfig edgev1alpha1.Sy
...
+				Namespace:     obj.Namespace,
+				Names:         obj.Names,
+			})
+		}
+	}
+	namespacedObjectsTable := typeTableOfFlatNamespacedObject{FlatNamespacedObjects: flatNamespacedObjects}
+
+	requiredNamespaces := namespacedObjectsTable.filter(func(row typeFlatNamespacedObject) bool { return len(row.Names) > 0 }).getAllNamespaces()
+
+	s.logger.V(3).Info("upsert namespaced objects as syncerConfig to syncConfigManager stores", "syncerConfigName", syncerConfig.Name, "numNamespaces", len(requiredNamespaces))
```
MikeSpreitzer, line 180: Does this logger already have a name-value pair holding the logicalcluster.Name or Path where the SyncerConfig appears?  Normally `syncerConfig.Name` is always "the-one".

pkg/syncer/controller/syncerconfigmanager.go
```diff
@@ -129,6 +132,99 @@ func (s *SyncerConfigManager) upsertNamespaceScoped(syncerConfig edgev1alpha1.Sy
...
+	if lgr := s.logger.V(4); lgr.Enabled() {
+		for _, agrs := range upstreamGroupResourcesList {
+			lgr.Info("APIGroupResources", "group", agrs.Group, "bar", agrs.VersionedResources)
```
MikeSpreitzer, line 183: Is there a bit of unfinished editing here?

pkg/syncer/controller/syncerconfigmanager.go
```diff
@@ -129,6 +132,99 @@ func (s *SyncerConfigManager) upsertNamespaceScoped(syncerConfig edgev1alpha1.Sy
...
+		}
+		edgeSyncConfigResources = append(edgeSyncConfigResources, edgeSyncConfigResourceForNamespace)
+	}
+
+	for _, object := range namespacedObjectsTable.FlatNamespacedObjects {
+		group := object.Group
+		version := object.APIVersion
+		resource := object.Resource
+		gvr := schema.GroupVersionResource{Group: group, Version: version, Resource: resource}
+		versionedResources := findVersionedResourcesByGVR(group, version, resource, upstreamGroupResourcesList, s.logger)
```
MikeSpreitzer, line 202: Why does this func return a slice instead of a single match?  `resource` is never `"*"`.  Also, it might be worth logging a warning message when there is no match, something higher level might have expected one.

pkg/syncer/controller/syncerconfigcontroller_test.go
```diff
@@ -73,63 +73,82 @@ var testAPIResourceList = []*metav1.APIResourceList{
...
+			}},
+		}},
+		ClusterScope: test1.syncerConfigSpec.ClusterScope,
+		Upsync:       test1.syncerConfigSpec.Upsync,
+	},
+	expected: test1.expected,
+}
+
+func TestSyncerConfig(t *testing.T) {
+	tests := []typeSyncerConfigTest{test1, test2}
```
MikeSpreitzer, line 151: FYI, a `testing.T` has a deadline, and the construction of a `context.Context` just below could be extended to take that deadline into account (using `context.WithDeadline`).  This is not something that has to be addressed in this PR.

MikeSpreitzer:
Thank you very much, this is good enough to merge.  I will give you a chance to consider the comments that I made inline first.

pkg/syncer/controller/syncerconfigmanager.go
yana1205: There isn't a significant reason... I wanted to make this type name a private scope in this module. To do that, I defined its name as starting small letter. The reason why I add a prefix `type` rather than just naming `flatNamespacedObject` is that I just wanted to use `typeFlatNamespacedObject` as a local variable name.

pkg/syncer/controller/syncerconfigmanager.go
yana1205: I think no.

pkg/syncer/controller/syncerconfigmanager.go
yana1205: Since `apiGroupResources.VersionedResources[version]` is array of `v1.APIResource` at https://github.com/kubestellar/kubestellar/blob/8731b55053a056a7f3211279ef6cbd3e502c2fce/pkg/syncer/controller/syncerconfigmanager.go#L355, I made a code to collect all elements of the array matching with `resource`.

>  it might be worth logging a warning message when there is no match, something higher level might have expected one.

I agree with it.

pkg/syncer/controller/syncerconfigmanager.go
MikeSpreitzer: Why the key "bar"?

pkg/syncer/controller/syncerconfigmanager.go
MikeSpreitzer: My question is whether the findVersionedResourcesByGVR would ever return a slice holding more than one member.

I asked about the underlying question at https://kubernetes.slack.com/archives/C0EG7JC6T/p1694053099115029 .  As best I can tell, the answer there implies that findVersionedResourcesByGVR will never return a slice longer than 1.

pkg/syncer/controller/syncerconfigmanager.go
MikeSpreitzer: I am not sure I understand that answer.  I see `typeFlatNamespacedObject` used as a package-scoped type name, not a more narrowly-scoped variable name.  I do not see `flatNamespacedObject` used at all.

yana1205:
@MikeSpreitzer 
Thank you for your kind reviews. 

> This does not have to be addressed in this PR.

Thanks. I will address the review comments for the future commits in another PR or issues.

pkg/syncer/controller/syncerconfigmanager.go
yana1205: "bar" would come from https://github.com/yana1205/kubestellar/blob/syncer-support-namespacedobject/pkg/syncer/controller/syncerconfigmanager.go#L94... I just copied and pasted it. It's probably typo of "ver". I will fix it.

pkg/syncer/controller/syncerconfigmanager.go
yana1205: > I do not see flatNamespacedObject used at all.

I thought I used `flatNamespacedObject` as a name of a local variable for `typeFlatNamespacedObject`. There is no reason to put `type` in the name of type. I will remove it.

yana1205:
I pushed some minor changes for 
- [comment](https://github.com/kubestellar/kubestellar/pull/968#discussion_r1320985568)
> "bar" would come from https://github.com/yana1205/kubestellar/blob/syncer-support-namespacedobject/pkg/syncer/controller/syncerconfigmanager.go#L94... I just copied and pasted it. It's probably typo of "ver". I will fix it.
- [comment](https://github.com/kubestellar/kubestellar/pull/968#discussion_r1320987810)
> I thought I used flatNamespacedObject as a name of a local variable for typeFlatNamespacedObject. There is no reason to put type in the name of type. I will remove it.
- [comment](https://github.com/kubestellar/kubestellar/pull/968#discussion_r1318796953)
> it might be worth logging a warning message when there is no match, something higher level might have expected one.

MikeSpreitzer (approved):
/lgtm
/approve
Thank you!

PR #980, opened 2023-09-11
📖 Improve pointers to getting kcp in example1

### LINKED ISSUE #974: bug: example1 too mum about fetch and install kcp
### Describe the bug

example1 is AKA the "extended example".

(1) This doc has two alternate paths for how to "deploy kcp and KubeStellar" --- either as bare processes or as workload in a Kubernetes cluster.  In the first path there are words about how to get kcp.  In the second path there are no words at all.

(2) In the first path, the pointer given for kcp is not good.  It does not point to instructions for what the user needs to do.  A better pointer would be to the begging of https://docs.kcp.io/kcp/v0.11/#quickstart .

### Steps To Reproduce

See above.

### Expected Behavior

Users know how to get what they need.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR improves the material in example1 about getting kcp.

## Related issue(s)

This addresses the first two points in #974.
This PR does not address the need to recognize that not all users are deployers.

### DIFF
--- a/docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
+++ b/docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-post-kcp.md
@@ -65,7 +65,11 @@ kubestellar init
 
 ### Deploy kcp and KubeStellar as a workload in a Kubernetes cluster
 
-First you will need to get a build of KubeStellar.  See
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
/cc @dumb0002 
/cc @francostellari 
/cc @clubanderson 
@eaepstein

clubanderson (commented):
/lgtm
/approve

PR #981, opened 2023-09-11
:sparkles: Makefile cleanup

ezrasilvera:
Removing some leftovers from KCP. In KS we don't use different go module for pkg/api

## Related issue(s)

Fixes #


MikeSpreitzer (commented):
/lgtm
/approve
Thanks!

PR #987, opened 2023-09-12
✨ cleanup cluster-watchall and apiwatch

MikeSpreitzer:
Also fix doc nits and remove unused client.

## Summary
This PR does a little cleanup in preparation for bigger things. The one does the following.

1. Switch the cluster-watchall command to use the library for kubectl options.
2. Add missing fenced code block tags in the README for cluster-watchall.
3. Remove an unused internal field in pkg/apiwatch.

## Related issue(s)

Believe it or not, this is part of working on #772 .  The placement-translator is going to need to know whether to update a status "subresource" or a plain field (#945).

### DIFF
--- a/cmd/cluster-watchall/README.md
+++ b/cmd/cluster-watchall/README.md
@@ -1,18 +1,20 @@
 # cluster-watchall
 
 This is a demonstration of a way to monitor all objects in all
-workspaces in a kcp server.  A little more carefully: the objects
-monitored are those of kinds that support informers --- that is, they
-support the `list` and `watch` verbs.  This demonstration only reports
-object creations.  Note also that it is workspaces, not other kinds of
-logical clusters, that matter here.
+workspaces in a kcp server.  This is: a demonstration of how to use
+`pkg/apiwatch`, a demonstration of how to monitor objects, and a
+utility to investigate what API resources are actually defined in a
+kcp server (more precisely than `kubectl api-resources`).
+
+This demonstration only reports object creations.  Note also that it
+is workspaces, not other kinds of logical clusters, that matter here.
 
 This demonstrates, among other things, use of the informer on "API
 resources" in [pkg/apiwatch](../../pkg/apiwatch).
 
-The notable command line flags are as follows.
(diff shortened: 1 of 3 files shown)


cmd/cluster-watchall/README.md
```diff
@@ -10,9 +10,9 @@ logical clusters, that matter here.
 This demonstrates, among other things, use of the informer on "API
 resources" in [pkg/apiwatch](../../pkg/apiwatch).
 
-The notable command line flags are as follows.
+The command line flags include all the usual ones for `kubectl`, of which a few prominent ones are shown below, plus one for configuring the metrics & debug endpoint.
```
ezrasilvera, line 13: Is there different meaning or usage for some of the flags ? If not why do you want to explicitly list some of them? Why don't we just point to the kubectl flags doc ?

cmd/cluster-watchall/main.go
```diff
@@ -61,34 +60,13 @@ import (
...
-	return clientConfig.ClientConfig()
-}
-
 func main() {
 	serverBindAddress := ":10203"
 	fs := pflag.NewFlagSet("watchall", pflag.ExitOnError)
 	klog.InitFlags(flag.CommandLine)
 	fs.AddGoFlagSet(flag.CommandLine)
-	cliOpts := NewClientOpts()
+	cliOpts := genericclioptions.NewConfigFlags(false)
```
ezrasilvera, line 68: Maybe I'm missing something here , genericclioptions.NewConfigFlags() returns ConfigFlags. I guess that if it passes compilation it implements all the functions you had (and still using) for ClientOpts ...
nit: maybe change the name to cliFlags ?

ezrasilvera:
Can you add short explanation what it is used for? I thought that originally you developed this just as a test/demo

pdettori (commented):
I have similar questions to @ezrasilvera, but other than that it looks good to me.

MikeSpreitzer:
The force-push to 73b447c2 responds to the reviews above.

cmd/cluster-watchall/README.md
MikeSpreitzer: No unusual meaning.  I added references.

I chose to highlight the ones I expect are most commonly interesting.

cmd/cluster-watchall/main.go
MikeSpreitzer: Not coincidentally, the old `NewClientOpts` func produced something that behaves much like what `genericclioptions.NewConfigFlags` returns.

I changed the variable name as suggested.

PR #988, opened 2023-09-12
✨ Extend apiwatch

### LINKED ISSUE #911: bug: meta API group is exported from ESPW
### Describe the bug

https://github.com/kubestellar/kubestellar/blob/v0.6.0/scripts/kubestellar#L252 creates an APIExport for the `meta.kcp.io` API group, but that is not intended.  There are no objects of that API group in any apiserver.  (That is an incidental side-effect of defining the in-process API for API introspection in the same place/way as the apiserver object oriented APIs.)

### Steps To Reproduce

See above.

### Expected Behavior

No BS.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR extends the apiwatch package to optionally be able to report about subresources.

This also extends the cluster-watchall command to have that option.

This PR is structured as three commits, to make reviewing easy.  The first is the change to the meta API that makes room for reporting subresources in an easy-to-consume way.  The second is the code generation.  The third is the client and server side updates.

## Related issue(s)

This PR contributes to #772: the placement translator will need to know how to update the reported state in the WDSes (see #945).

This PR fixes #911 by removing the all the generation of unwanted stuff for the meta.kubestellar.io API group.


cmd/cluster-watchall/main.go
```diff
@@ -91,7 +91,10 @@ func main() {
 	cliOpts := NewClientOpts()
 	cliOpts.AddFlags(fs)
 	fs.Var(&utilflag.IPPortVar{Val: &serverBindAddress}, "server-bind-address", "The IP address with port at which to serve /metrics and /debug/pprof/")
-	fs.Parse(os.Args[1:])
+	includeSubresources := false
+	fs.BoolVar(&includeSubresources, "include-subresources", includeSubresources, "list subresources too")
+	args := os.Args[1:]
+	fs.Parse(args)
```
ezrasilvera, line 98: why adding the args tmp variable?

cmd/cluster-watchall/main.go
```diff
@@ -251,6 +256,14 @@ func (acw *allClustersWatcher) NewClusterWatcher(clusterName logicalcluster.Name
 	return cw
 }
 
+func (acw *allClustersWatcher) writeSubresources(clusterName logicalcluster.Name, gv schema.GroupVersion, subs []*urmetav1a1.APIResourceSpec, prefix string) {
+	for _, sub := range subs {
+		informable := verbsSupportInformers(sub.Verbs)
+		acw.WriteCSV([]string{"SUBRESOURCE", clusterName.String(), gv.Group, gv.Version, sub.Kind, prefix + sub.Name, fmt.Sprintf("%v", informable)})
+		acw.writeSubresources(clusterName, gv, sub.SubResources, prefix+sub.Name+"/")
```
ezrasilvera, line 263: do we assume includeSubresources means recursive over all ?  maybe this should be an additional option and single level (at any level) should be achieved by specifying the subresource as the resource and setting the  includeSubresources to true.

pdettori: Does Kubernetes has sub resources under sub resources ? What is the use case for that ?

cmd/cluster-watchall/main.go
MikeSpreitzer: To help debugging a mystery that I am having about command line parsing.

cmd/cluster-watchall/main.go
MikeSpreitzer: @pdettori : I do not know whether it actually happens, but the API seems to say that it can.  See https://github.com/kubernetes/client-go/blob/v0.28.1/dynamic/interface.go#L34, for example.

@ezrasilvera : I am not sure I follow what you are asking and suggesting.  See my response to Paolo.  The API change here follows that possibility of more than one level of "sub".  I expect that the API here will serve #772 .

MikeSpreitzer:
However, maybe kube API machinery does not admit recursive datatypes?

I do not understand the error message below.

```console
bash-3.2$ kubectl apply -f config/exports
apiexport.apis.kcp.io/edge.kubestellar.io unchanged
apiexport.apis.kcp.io/meta.kubestellar.io unchanged
apiexport.apis.kcp.io/space.kubestellar.io unchanged
apiresourceschema.apis.kcp.io/v230810-2d48e9f7.customizers.edge.kubestellar.io unchanged
apiresourceschema.apis.kcp.io/v230810-2d48e9f7.edgeplacements.edge.kubestellar.io unchanged
apiresourceschema.apis.kcp.io/v230810-2d48e9f7.edgesyncconfigs.edge.kubestellar.io unchanged
apiresourceschema.apis.kcp.io/v230810-2d48e9f7.locations.edge.kubestellar.io unchanged
apiresourceschema.apis.kcp.io/v230810-2d48e9f7.singleplacementslices.edge.kubestellar.io unchanged
apiresourceschema.apis.kcp.io/v230729-f70497310.spaceproviderdescs.space.kubestellar.io unchanged
apiresourceschema.apis.kcp.io/v230813-0036646a.spaces.space.kubestellar.io unchanged
apiresourceschema.apis.kcp.io/v230817-d752f622.syncerconfigs.edge.kubestellar.io unchanged
apiresourceschema.apis.kcp.io/v230810-2d48e9f7.synctargets.edge.kubestellar.io unchanged
Error from server (Forbidden): error when creating "config/exports/apiresourceschema-apiresources.meta.kubestellar.io.yaml": apiresourceschemas.apis.kcp.io "v230911-02826cb7.apiresources.meta.kubestellar.io" is forbidden: [spec.versions[0].schema.openAPIV3Schema.properties[spec].properties[subResources].items.type: Invalid value: "": must be object if parent array's x-kubernetes-list-type is map spec.versions[0].schema.openAPIV3Schema.properties[spec].properties[subResources].items.type: Required value: must not be empty for specified array items]
```

ezrasilvera:
Other than the recursive thing it looks good to me. I don't have strong opinion on that though. I see that you posts some errors - so please comment  If you are OK with leaving the code as is and I can approve it.

MikeSpreitzer:
The commit added after the posted errors resolves them, as the comment in the commit is trying to say.

MikeSpreitzer:
The force-push to 36fba495 is a rebase onto `main`.

MikeSpreitzer:
The force-push to 8d92803a6 is a rebase onto `main`.

PR #989, opened 2023-09-12
:sparkles: Space restructure phase1

ezrasilvera:
1. Move all space-framework code to a separate directory
2. Use separate Makefile to build the space-manager
3. Separate the space related objects and the space generated code into dedicated directories under the space-framework 
4. Create a separate go module for the space-framework

Note to reviewer:  There are no changes to existing KS files. The only changes in KS existing directories are removing the space files. 

## Related issue(s)

Fixes #


Makefile
```diff
@@ -99,6 +99,10 @@ LDFLAGS := \
 	-X k8s.io/component-base/version.gitMinor=${KUBE_MINOR_VERSION} \
 	-X k8s.io/component-base/version.buildDate=${BUILD_DATE} \
 	-extldflags '-static'
+
+# So the recursive Space MAkefile can use the same flags
```
MikeSpreitzer, line 103: The wording of this comment suggests a dependency that runs in the wrong direction (the Space Makefile relying on a service provided by this Makefile).  I think an accurate and less alarming explanation is that this is a shorthand for explicitly setting the values of these variables in the nested `make` invocations.

Makefile
```diff
@@ -483,4 +486,20 @@ verify-modules: modules  ## Verify go modules are up to date
...
+	$(MAKE) -C space-framework crds
+.PHONY: space-crds
+
+space-codegen: 
+	$(MAKE) -C space-framework codegen
+.PHONY: space-codegen
+
+space-modules: 
+	$(MAKE) -C space-framework modules
+.PHONY: space-modules
```
MikeSpreitzer, line 503: Why do we want targets in this Makefile that are just aliases for targets in another Makefile?  Why not tell contributors to use the underlying Makefile and targets directly?

MikeSpreitzer:
The multi-space client is logically part of the space framework.  The fact that it has a `KS() ksclient.Interface` method precludes that.  I have always thought that method was out of place, and I have now realized that this is why.

go.mod
```diff
@@ -190,6 +184,7 @@ require (
 // replace directive is aligned with kcp v0.11.0 (https://github.com/kcp-dev/kcp/blob/v0.11.0/go.mod)
 replace (
 	github.com/google/cel-go => github.com/google/cel-go v0.12.6
+	github.com/kubestellar/kubestellar/space-framework => ./space-framework
```
MikeSpreitzer, line 187: If I recall correctly, such a `replace` statement makes this `go.mod` impossible to reference in a `go.work`.

space-framework/Makefile
```diff
@@ -0,0 +1,144 @@
...
+TMPDIR := $(shell mktemp -d)
+
+
+.PHONY: all
+all: build
+
+.PHONY: build
+build: WHAT ?= ./cmd/space-manager
+build: bin-dir
+	GOOS=$(OS) GOARCH=$(ARCH) CGO_ENABLED=0 go build $(BUILDFLAGS) -ldflags="$(LDFLAGS)" -o bin $(WHAT)
```
MikeSpreitzer, line 32: This Makefile should be convenient and unsurprising to use on its own.

space-framework/hack/update-codegen-clients.sh
```diff
@@ -0,0 +1,47 @@
...
+
+${CODE_GENERATOR} \
+  "client:outputPackagePath=github.com/kubestellar/kubestellar/space-framework/pkg/client,apiPackagePath=github.com/kubestellar/kubestellar/space-framework/pkg/apis,singleClusterClientPackagePath=github.com/kubestellar/kubestellar/space-framework/pkg/client/clientset/versioned,headerFile=${BOILERPLATE_HEADER}" \
+  "lister:apiPackagePath=github.com/kubestellar/kubestellar/space-framework/pkg/apis,headerFile=${BOILERPLATE_HEADER}" \
+  "informer:outputPackagePath=github.com/kubestellar/kubestellar/space-framework/pkg/client,apiPackagePath=github.com/kubestellar/kubestellar/space-framework/pkg/apis,singleClusterClientPackagePath=github.com/kubestellar/kubestellar/space-framework/pkg/client/clientset/versioned,headerFile=${BOILERPLATE_HEADER}" \
+  "paths=./..." \
+  "output:dir=./../client"
+popd
+
+go install "${CODEGEN_PKG}"/cmd/openapi-gen
```
MikeSpreitzer, line 47: This is not an absolute requirement, but: (a) I would prefer to not have stuff that makes GitHub complain like this, and (b) many (but not all) text editors tend to add a terminal linebreak if it is not already there, which would introduce a spurious difference if someone views a file that lacks a terminal linebreak.

Screenshot 2023-09-12 at 3 48 18 PM

MikeSpreitzer:
I think you should produce a working go.work for this GitHub repo.  You should not include cmd/go-template-execute/go.mod and cmd/kube-watchall/go.mod, because they are just utilities and their go.mods are explicitly and deliberately not coordinated with the rest of the code here.

When there is a working `go.work`, `go vet ./...` will not fail.

I found the solution at https://github.com/golang/go/issues/60313#issuecomment-1612370659 helpful.

go.mod
ezrasilvera: I think this replace is needed for the initial merge + local work when you want to test changes on the local module. 
Once we push it to the repository, I believe we can remove that line completely.

space-framework/hack/update-codegen-clients.sh
ezrasilvera: Will change

ezrasilvera:
> I think you should produce a go.work file that works for this GitHub repo (I am not saying --- yet --- to put it in GitHub). You should not include cmd/go-template-execute/go.mod and cmd/kube-watchall/go.mod, because they are just utilities and their go.mods are explicitly and deliberately not coordinated with the rest of the code here.
> 
> When there is a working `go.work`, `go vet ./...` will not fail.
> 
> I found the solution at [golang/go#60313 (comment)](https://github.com/golang/go/issues/60313#issuecomment-1612370659) helpful: using go1.21.1, `go get cloud.google.com/go/compute@v1.7.0` edited `go.mod` and `go.sum` so that `go vet ./...` works in the presence of the obvious `go.work`.

Thanks. Will look, but as far as I saw the command practice is to use go.work only for local development and never upload as part of the project.

Makefile
ezrasilvera: Agree. Will be moving to independent MAkefile

Makefile
ezrasilvera: Agree

ezrasilvera:
> The multi-space client is logically part of the space framework. The fact that it has a `KS() ksclient.Interface` method precludes that. I have always thought that method was out of place, and I have now realized that this is why.

I agree, my plan is to have space-client-library in the space framework , and leave only a KS specific space client on the KS side which is using the space-client-library.   The KS space client won't need to import any space go-client functions , only the space-client-library (that will implement the space's configs cache, spaces watcher, etc..)
My plan was to push that as a separate PR after this one is merged, focusing on the client restructure. Is that OK ?

MikeSpreitzer:
> I agree, ... My plan was to push that as a separate PR after this one is merged, focusing on the client restructure. Is that OK ?

I will not say "no", just realize that you are planning on further directory and go.mod restructuring in another PR.

MikeSpreitzer:
My point about go.work is, as I said, not to put it in GitHub but to make sure that it is possible to have a go.work that works.  Demonstrate that to yourself by creating one.  Sharing it informally would be a nice addition.

go.mod
MikeSpreitzer: Also, I am unable to reconstruct the problem that I was vaguely remembering.  Maybe go tooling has gotten less picky since then.

MikeSpreitzer:
After the merge conflicts are handled I expect this will be good enough to merge with the expectation of immediate follow-on work.  The Makefiles should be independent: the space Makefile should do its own derivation of BUILDFLAGS and LDFLAGS, and the MCCM Makefile should not recurse into the space Makefile.  The space client library should lose its MCCM-specific method and move into the space framework.

ezrasilvera:
@MikeSpreitzer I rebased + changed the Makefiles (of KS & space-framework) so the space Makefile is now self-sufficient and not called from within the KS Makefile. 
After this PR is merged I plan to push the PR that move the space-client code into the space-framework so that the space-client doesn't depend on KS.

hack/logcheck.out
```diff
@@ -1,9 +1,23 @@
...
+/pkg/placement/workload-projector.go:703:2: the result of logger.V should be stored in a variable and then be used multiple times: if logger := logger.V(); logger.Enabled() { ... logger.Info ... }
+/pkg/placement/workload-projector.go:880:1: A function should accept either a context or a logger, but not both. Having both makes calling the function harder because it must be defined whether the context must contain the logger and callers have to follow that.
+/pkg/placement/workload-projector.go:962:4: the result of logger.V should be stored in a variable and then be used multiple times: if logger := logger.V(); logger.Enabled() { ... logger.Info ... }
+logcheck: analysis skipped due to errors in package
+logcheck: analysis skipped due to errors in package
+logcheck: analysis skipped due to errors in package
+logcheck: analysis skipped due to errors in package
+logcheck: analysis skipped due to errors in package
+logcheck: analysis skipped due to errors in package
+logcheck: analysis skipped due to errors in package
```
MikeSpreitzer, line 23: If I recall correctly, these lines appear only when stuff is not compiling.  That sounds bad.

pkg/client/clientset/versioned/cluster/typed/meta/v1alpha1/apiresource.go
```diff
@@ -0,0 +1,72 @@
+//go:build !ignore_autogenerated
```
MikeSpreitzer, line 1: Please double-check your rebase, this file was deleted in #988 and should not be coming back.

ezrasilvera: fixed

MikeSpreitzer (approved):
/lgtm
/approve

PR #999, opened 2023-09-14
:seedling: Remove the dependency of core.kcp.io/v1alpha1/logicalclusters from Syncer

### LINKED ISSUE #874: Remove kcp dependencies from Kubestellar Syncer
### Feature Description

Remove dependencies on KCP from Kubestellar syncer. As per Yana, "The code doesn't work if we just replace it with kubestellar logicalcluster clientset. Syncer codes depend on kcp logical cluster (core.kcp.io/v1alpha1/logicalclusters) that's created by kcp when a workspace is created. The kcp logical cluster is different from kubestellar logical cluster both syntactically and semantically. We need to remove the dependency from codes relating with Syncer/Syncer manifest generation."

### Proposed Solution

Remove dependencies on KCP from Kubestellar syncer. 

### Want to contribute?

### Additional Context

See https://github.com/kubestellar/kubestellar/issues/848#issuecomment-1669254839

yana1205:
## Summary

This is the PR for #874.

#### What's the changed

- Remove the dependency of kcp logical cluster function from Syncer

I reviewed Syncer code whether Syncer really depends on logical cluster id or not and concluded it's not required. I have removed the dependent codes on the logical cluster stuff from Syncer. As a result, Syncer does no longer require `--from-cluster` option.

#### Verification

UT passed
```
$ make test-syncer COUNT=1
go test -count 1 `go list ./... | grep "/pkg/cliplugins\|/pkg/syncer"`
?       github.com/kubestellar/kubestellar/pkg/syncer   [no test files]
?       github.com/kubestellar/kubestellar/pkg/syncer/clientfactory     [no test files]
?       github.com/kubestellar/kubestellar/pkg/syncer/shared    [no test files]
?       github.com/kubestellar/kubestellar/pkg/syncer/syncers   [no test files]
ok      github.com/kubestellar/kubestellar/pkg/cliplugins/kubestellar/syncer-gen        0.952s
ok      github.com/kubestellar/kubestellar/pkg/syncer/controller        18.689s
```

E2E passed
```
$ make e2e-test-kubestellar-syncer                                                                                       
rm -rf /Users/yana/git/trl/kcp/kcp-edge-mc/.kcp/etcd-server
rm -rf /Users/yana/git/trl/kcp/kcp-edge-mc/.kcp/.admin-token-store /Users/yana/git/trl/kcp/kcp-edge-mc/.kcp/admin.kubeconfig
rm -rf /Users/yana/git/trl/kcp/kcp-edge-mc/.kcp/apiserver.* /Users/yana/git/trl/kcp/kcp-edge-mc/.kcp/sa.key
export PATH=/Users/yana/git/trl/kcp/kcp-edge-mc/kcp/bin:$PATH && \
        kcp start --root-directory=/Users/yana/git/trl/kcp/kcp-edge-mc/.kcp > /Users/yana/git/trl/kcp/kcp-edge-mc/.kcp/kcp.log 2>&1 & PID=$! && echo "PID $PID" && \
        trap 'kill -TERM $PID' TERM INT EXIT && \
        while [ ! -f "/Users/yana/git/trl/kcp/kcp-edge-mc/.kcp/admin.kubeconfig" ]; do sleep 1; echo "kcp is not ready. wait for 1s...";done && \
        echo 'Starting test(s)' && \
        NO_GORUN=1 GOOS=darwin GOARCH=arm64 \
                go test -race  ./test/e2e/kubestellar-syncer/...  \
                --kcp-kubeconfig /Users/yana/git/trl/kcp/kcp-edge-mc/.kcp/admin.kubeconfig --suites kubestellar-syncer \

PID 6248
kcp is not ready. wait for 1s...
kcp is not ready. wait for 1s...
kcp is not ready. wait for 1s...
Starting test(s)
ok      github.com/kubestellar/kubestellar/test/e2e/kubestellar-syncer  258.144s
```

I also verified that a manual test (example1.md) passed with the Syncer image built by this branch

## Related issue(s)

Fixes #874

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "go.sum|docs/package-lock.json|cmd/kube-watchall/go.sum|docs/config.toml|docs/resources/_gen|docs/static|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2023-06-11T03:51:35Z",
+  "generated_at": "2023-09-14T15:02:41Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -92,7 +92,7 @@
         "hashed_secret": "4f151c44a6783667af6cedaa3d12560f8af2f38a",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 499,
+        "line_number": 481,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -100,7 +100,7 @@
         "hashed_secret": "eba2ae817b95fdb713f3af658a6ae1821e452264",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 605,
+        "line_number": 585,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/cmd/syncer/main.go
+++ b/cmd/syncer/main.go
@@ -28,8 +28,6 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 
-	"github.com/kcp-dev/logicalcluster/v3"
-
 	synceroptions "github.com/kubestellar/kubestellar/cmd/syncer/options"
 	"github.com/kubestellar/kubestellar/pkg/syncer"
 )
@@ -76,7 +74,6 @@ func main() {
 	syncerConfig := &syncer.SyncerConfig{
 		UpstreamConfig:   upstreamConfig,
 		DownstreamConfig: downstreamConfig,
-		SyncTargetPath:   logicalcluster.NewPath(options.FromClusterPath),
 		SyncTargetName:   options.SyncTargetName,
 		SyncTargetUID:    options.SyncTargetUID,
 	}
--- a/cmd/syncer/options/options.go
+++ b/cmd/syncer/options/options.go
@@ -24,15 +24,14 @@ import (
 )
 
 type Options struct {
-	QPS             float32
-	Burst           int
-	FromKubeconfig  string
-	FromContext     string
-	FromClusterPath string
-	ToKubeconfig    string
-	ToContext       string
-	SyncTargetName  string
-	SyncTargetUID   string
+	QPS            float32
+	Burst          int
+	FromKubeconfig string
+	FromContext    string
+	ToKubeconfig   string
+	ToContext      string
+	SyncTargetName string
+	SyncTargetUID  string
 }
 
 func NewOptions() *Options {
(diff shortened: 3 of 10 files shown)


MikeSpreitzer:
This PR in its current form breaks the example1 test.  That is because this PR both: (1) changes the syncer from _requiring_ the `--from-cluster` command line argument to _forbidding_ it, and (2) changes the `kubectl kubestellar syncer-gen` plugin from putting that command line flag in the generated PodSpec for the syncer to omitting that flag.  But this PR does _not_ change the `kubectl kubestellar prep-for-syncer` plugin to default to a syncer image built from (the effects of) this PR.  So a test fails because the old syncer image is used and it just crashes due to the missing command line flag.

To fix I suggest either (A) update `kubectl-kubestellar-prep_for_syncer` to specify `quay.io/kubestellar/syncer:test-999` (which I built from this PR) as the default syncer image, as a temporary expedient (after this merges we should build a new syncer image from `main` and update the prep-for-syncer plugin to default to that one) or (B) break this PR into two, with the first one making the syncer allow but not require the `--from-cluster` command line flag.

yana1205:
Thanks for the investigation. I didn't notice that removing this option was breaking the backward compatibility. To keep the compatibility, the suggestion (B) might be safer for those who use syncer-gen plugin with older Syncer images. I added another commit that syncer-gen adds `--from-cluster` option with a dummy value.

yana1205:
Maybe, syncer-gen plugin has to keep `--from-cluster` option permanently for those who use older version of Syncer (`kubectl kubestellar syncer-gen  --syncer-image <older version of Syncer>`).

MikeSpreitzer:
The added commit bfe7371912ccc4acfa685d87883d5fa0802794dd does not take the approach I labeled "B".  This PR now passes testing with the example1 docs-ecutable but still requires a coordinated change in the future: anyone running the new syncer code will also have to run revised syncer-gen code that omits the `--from-cluster` flag.  The way to enable incremental transition is as follows.
- The first PR that merges makes the `--from-cluster` _optional_ on the syncer command line, and does not change the syncer-gen code from what is currently merged (i.e., syncer-gen continues to emit `--from-cluster`).
- After that merges we can build a syncer image from `main` and start using that.  We can change prep-for-syncer to default to the new syncer image.
- At that same time or later, we change syncer-gen to _not_ emit the `--from-cluster` flag.

yana1205:
I got it. The phasing approach is needed. I am concerning the last step.

> At that same time or later, we change syncer-gen to not emit the --from-cluster flag.

If syncer-gen stops emitting `--from-cluster` flag, syncer image that's equal or less than v0.7.0 won't be supported.
How about adding a logic that if specified image version in `--syncer-image <= v0.7.`0, syncer-gen emits `--from-cluster` in generated yaml while if the version > v0.7.0, syncer-gen doesn't emit the flag?
Or, simply syncer-gen returns an error if a user specifies `--syncer-image` with the version <= v0.7.0

MikeSpreitzer:
I think that is more complicated than necessary.  We do not really have a use case for very uncoordinated usage.  In fact, I think my previous remark was more demanding than necessary.  The current state of this PR is good enough.  It passes tests.  After it merges we can build a new syncer image from `main`.  Then in just one more PR we can change syncer-gen to omit the `--from-cluster` flag and change `prep-for-syncer` to default to using the new syncer image.

yana1205:
> I think that is more complicated than necessary

Yes, I agree with it. I feel that is too much.

PR #1000, opened 2023-09-14
✨ Make syncer image not depend on commit ID

MikeSpreitzer:
## Summary
This PR changes the command that builds the syncer image so that the image depends only on the code and other functional stuff that goes into it, to support managing the image references in a functional style without self-reference problems.

This PR also updates the Makefile variables relevant to building the syncer image, so that their default values match what we normally want.  This also includes renaming the variable "ARCHS" to "SYNCER_PLATFORMS" because (a) its usage is specific to building the syncer image and (b) "platform" is the word used in in the documentation about this sort of building.

## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,8 @@ SHELL := /usr/bin/env bash -e
 CENTER_PLATFORMS ?= linux/amd64,linux/arm64,linux/ppc64le # kcp does not support linux/s390x
 CENTER_IMAGE ?= quay.io/kubestellar/kubestellar:$(shell date -u +b%y-%m-%d-%H-%M-%S)
 
+SYNCER_PLATFORMS ?= linux/amd64,linux/arm64,linux/s390x
+
 GO_INSTALL = ./hack/go-install.sh
 
 TOOLS_DIR=hack/tools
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
I find that this makes the image digest not depend on irrelevant ordinary filesystem contents, but doing a `git commit` makes a difference.

yana1205:
/lgtm

> to support managing the image references in a functional style without self-reference problems.

Though, I'm not sure whether I correctly understand what problem is resolved but this change is to add git commit id to the image tag instead of embedding the commit id into the image metadata in order to make the hash of created image not depending on something relating with git commit.

MikeSpreitzer:
The problem is that our current image building technique does not make the image be a function of its inputs, other stuff creeps in.  My plan is that once the image is a pure function of just the relevant inputs --- that is, the image digest will only change when relevant code or such changes --- that we change our practice of how the image is referenced to be by digest rather than release semver or other interim tag.  This will remove a self-reference problem that we have in releases.

MikeSpreitzer:
The force-push to c7c3a5e67 also takes the git metadata out of the go build (did you know about https://tip.golang.org/doc/go1.18#go-version ?) used in `ko build`.  Now the image digest _is_ just a function of the relevant code and such; changing the git commit ID does NOT affect the image digest.

This force-push also changes the default values of the variables used in `make build-kubestellar-syncer-image` to match what is normally used, and renames "ARCHS" to "SYNCER_PLATFORMS" because this variable's use is that specific.

yana1205:
I have just understood it. I agreed with making the syncer image not depend on irrelevant stuff and with using the image digest instead of semver. It helps when we update scripts or documents to point already the existing image digest at releasing since the image digest doesn't change at the release. Thanks for giving the detail.

/lgtm

PR #1002, opened 2023-09-14
✨ Use pr 999

MikeSpreitzer:
## Summary
This PR builds on #999 by changing the prep-for-syncer script to use an image built from that PR.

This is for testing.  DO NOT MERGE, we will want to actually use an image built from `main` after #999 merges.

## Related issue(s)

Fixes #

/cc @yana1205 
/cc @clubanderson


MikeSpreitzer:
This PR has served its testing purpose.

PR #1003, opened 2023-09-15
:seedling: Remove --from-cluster flag and temporally use syncer:pr-999 for transitioning to the new syncer-gen

yana1205:
## Summary

This is followup change for https://github.com/kubestellar/kubestellar/pull/999#issuecomment-1720389062

> * After that merges we can build a syncer image from main and start using that. We can change prep-for-syncer to default to the new syncer image.
> * At that same time or later, we change syncer-gen to not emit the --from-cluster flag.

Mike has made the latest image (syncer:pr-999) built after https://github.com/kubestellar/kubestellar/pull/999 is merged. 
The PR contains 
- remove --from-cluster flag
- switch the default image to this image

## Related issue(s)

Fixes #


MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #1007, opened 2023-09-18
:book: QuickStart corrections

namasl:
## Summary

This PR is mostly command fixes found while following the QuickStart guide on a fresh install of Ubuntu Server 20.04.6.

A lot of the changes involve adding `sudo` before `kind` or certain `kubectl` commands; maybe there is a cleaner way of dealing with this in the documentation to keep it specific to installations that actually require `sudo`.

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/Getting-Started/quickstart-subs/quickstart-2-apache-example-deployment-a-kind-clusters-florin.md
+++ b/docs/content/Getting-Started/quickstart-subs/quickstart-2-apache-example-deployment-a-kind-clusters-florin.md
@@ -2,7 +2,7 @@
 Create the first edge cluster:
 
 ```shell
-sudo kind create cluster --name florin --config - <<EOF
+kind create cluster --name florin --config - <<EOF
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
--- a/docs/content/Getting-Started/quickstart-subs/quickstart-2-apache-example-deployment-a-kind-clusters-florin.md
+++ b/docs/content/Getting-Started/quickstart-subs/quickstart-2-apache-example-deployment-a-kind-clusters-florin.md
@@ -2,7 +2,7 @@
 Create the first edge cluster:
 
 ```shell
-kind create cluster --name florin --config - <<EOF
+sudo kind create cluster --name florin --config - <<EOF
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
--- a/docs/content/Getting-Started/quickstart-subs/quickstart-1-install-and-run-kubestellar.md
+++ b/docs/content/Getting-Started/quickstart-subs/quickstart-1-install-and-run-kubestellar.md
@@ -41,7 +41,6 @@ kubectl ws tree
 which should yield:
 
 ``` { .sh .no-copy }
-kubectl ws tree
 .
 └── root
     ├── compute
@@ -111,7 +110,6 @@ kubectl ws tree
 which should yield:
 
 ``` { .sh .no-copy }
-kubectl ws tree
 .
 └── root
     ├── compute
--- a/docs/content/Getting-Started/quickstart-subs/quickstart-2-apache-example-deployment-a-kind-clusters-guilder.md
+++ b/docs/content/Getting-Started/quickstart-subs/quickstart-2-apache-example-deployment-a-kind-clusters-guilder.md
@@ -2,7 +2,7 @@
 Create the second edge cluster:
 
 ```shell
-sudo kind create cluster --name guilder --config - <<EOF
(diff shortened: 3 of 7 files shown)


docs/content/Getting-Started/quickstart-subs/quickstart-2-apache-example-deployment-a-kind-clusters-florin.md
```diff
@@ -14,4 +14,4 @@ EOF
 ```  
 
 Note: if you already have a cluster named 'florin' from a previous exercise of KubeStellar, please delete the florin cluster ('kind delete cluster --name florin') and create it using the instruction above.
-<!--quickstart-2-apache-example-deployment-a-kind-clusters-florin-end-->
\ No newline at end of file
+<!--quickstart-2-apache-example-deployment-a-kind-clusters-florin-end-->
```
MikeSpreitzer, line 17: What is the change here?

MikeSpreitzer:
Yeah, there is something strange happening to you.  I do not recall seeing someone need `sudo` for `kubectl`.  I can see how it might be an issue for `kind` (because of its use of a CRI socket), but it is normal to arrange things so that the user involved can use `kind` without `sudo`.

docs/content/Getting-Started/quickstart-subs/quickstart-2-apache-example-deployment-a-kind-clusters-florin.md
namasl: I only changed line number 5, I'm not sure why it's picking up line 17. A few but not all of the files I touched git says I changed the last line but shows no obvious difference. I made edits with vim, no clue...

namasl: It seems the original files for which I "edited" the last line did not terminate with LF, so Vim added those in.

MikeSpreitzer: In the case of adding a terminal linebreak, I know that I have seen GitHub call that out by showing a red circle-with-line to indicate lack of terminal linebreak.  Of course, there are two cases: adding and removing.  I wonder if that makes the difference.

namasl:
@MikeSpreitzer I redid the QuickStart on fresh installs of Ubuntu 22.04.3 LTS and Debian 12.1, and both require the addition of `sudo` at the same points.

namasl:
The requirement of `sudo` with kind/kubectl stems from the usage of Docker. In Debian 12, substituting Podman for Docker (just do a `sudo apt install podman`) allows the clusters to run rootless.

In principle this aught to work with Ubuntu, but there is a known bug and none of the fixes have worked properly for me (downgrading to podman=3.4.4+ds1-1ubuntu1, "upgrading" with the Kubic repo, or installing a different containernetworking-plugins).
https://bugs.launchpad.net/ubuntu/+source/libpod/+bug/2024394

MikeSpreitzer:
I am confused.  `kubectl` does not use the CRI (regardless of whether it is Docker or podman).

MikeSpreitzer:
Regarding `kind`, I see that indeed the merged quickstart does not say anything about enabling the ordinary user to use the CRI socket.  I think that is what should be fixed.

namasl:
> I am confused. `kubectl` does not use the container runtime (regardless of whether it is Docker or podman).

Sorry, I wasn't very precise in my last comment; `kind` requires the use of `sudo` with a vanilla install of Docker. This results in the clusters from step 3a of QuickStart being in root's space, so commands like `kubectl --context kind-florin apply -f florin-syncer.yaml` won't hit the right context unless executed with `sudo`.

> Regarding `kind`, I see that indeed the merged quickstart does not say anything about enabling the ordinary user to use the container runtime socket. I think that is what should be fixed.

I like this path.

namasl:
Latest push includes instructions for enabling rootless Docker. QuickStart has been tested to work with rootless Docker in Ubuntu 20.04 LTS, Debian 12, as well as Podman in Debian 12 (I can only assume Podman will work just as well in Ubuntu once their bug is resolved).

For RHEL/CentOS/Fedora I've not tested anything and only supplied the link to set up rootless Docker. Depending on the OS version different instructions need to be followed. Including pre-req install instructions for multiple versions of multiple OS's is cumbersome and can easily become stale if a pre-req update changes instructions, so I'm thinking links to the maintainer's install docs is a better approach anyway.

clubanderson (commented):
/lgtm
/approve

clubanderson:
@namasl looks like this needs a rebase...

clubanderson:
nevermind

PR #1008, opened 2023-09-19
✨ Draft API for requesting return of singleton reported state

MikeSpreitzer:
## Summary
This PR proposes an API by which (1) users explicitly request return of reported state when the number of executing copies is 1 (i.e., #772) and (2) the implementation reports the number of executing copies.

This PR is for review of the proposed API.  It would be confusing to merge the API extension without the accompanying implementation.

## Related issue(s)

Fixes #


MikeSpreitzer:
The force-push to 6ab57c2c8 is a rebase onto `main`.

PR #1010, opened 2023-09-19
🌱 Cleanup after EdgePlacement downsync unification

MikeSpreitzer:
## Summary
This PR removes a little junk and uses `ObjectName` in place of `string` more.

## Related issue(s)

Fixes #


MikeSpreitzer:
See #1011 instead.

PR #1028, opened 2023-09-21
✨ Draft alternate API for singleton reported state return

MikeSpreitzer:
## Summary
This PR proposes an alternate API (to #1008) for requesting the return of reported state from a single executing copy all the way back to the WDS.  This proposal has the advantage of not requiring modification to the workload objects, better supporting scenarios where those come from third parties.

## Related issue(s)

Fixes #


MikeSpreitzer:
Although GitHub has not detected it, this now conflicts with `main`.
See #1031 instead, which adds the implementation as well as the API.

PR #1034, opened 2023-09-25
🌱 Also tag center KS image with git commit

MikeSpreitzer:
## Summary
This PR updates the Makefile so that building the central container image now applies two tags, rather than just the previous one; the new tag identifies the git commit, while the old tag identifies the build time.  Adding this information helps with understandability of the produced artifacts, while keeping it out of the image contents avoids an offense against making the image depend only on its _functional_ input.

This PR also fixes a bug in the center.Dockerfile that was introduced in a recent space PR.

## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,8 +15,16 @@
 # We need bash for some conditional logic below.
 SHELL := /usr/bin/env bash -e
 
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
The fact that the Dockerfile bug slipped in here shows a gap in CI testing: the building of the central container image is not tested.

PR #1037, opened 2023-09-25
:bug: Remove patch used just for v0.4.0

francostellari:
## Summary

Remove the files that were used to patch the container for v0.4.0.
Updated the Docker file correspondingly.

## Related issue(s)

Fixes #


MikeSpreitzer:
The center image Docker file that we are now using is https://github.com/kubestellar/kubestellar/blob/main/center.Dockerfile and the make targets for building it are in https://github.com/kubestellar/kubestellar/blob/main/Makefile.  The instructions that we are directing users to and maintaining are in https://github.com/kubestellar/kubestellar/blob/main/docs/content/Coding%20Milestones/PoC2023q1/commands.md#deployment-into-a-kubernetes-cluster

francostellari:
@MikeSpreitzer this was just to fix what is currently in the repo. However, it sounds like you rather delete `user/container`. Is that what you want?

MikeSpreitzer:
The `entry.sh` needs to remain, somewhere.  The rest of `user/container` and `user/yaml` is not needed.  And I think that `user/helm-chart` should move out from under `user/`, we are maintaining it from an internal stance.

PR #1039, opened 2023-09-26
✨ Rename "center" image to "core", support extra tag

MikeSpreitzer:
## Summary
This PR does two things.

First, this PR makes a few improvements to the make targets for building the center image.  One is to introduce the option for a user-supplied tag. Another removes one source of difficulty when building on ARM for x86 (but other difficulties remain, https://github.com/docker/buildx/issues/2028). Another other fixes the `kubestellar-image-local` target to work for podman on ARM.

Second, this PR makes a terminology shift, from speaking of the "center image" to speaking of the "core image" (which is more consistent with our other terminology, such as KubeStellar Core Space).

## Related issue(s)

Fixes #


clubanderson (commented):
/approve

PR #1041, opened 2023-09-26
✨ Recognize that core helm chart is being maintained internally

MikeSpreitzer:
## Summary
This PR moves the central Helm chart out from under the directory designed for external-stance stuff, since it is being maintained from an internal stance.
This PR also removes the older stuff that was under `user/` that has already been moved and adopted to internal stance (the Dockerfile, CLI for contributors and for users, and documentation).

## Related issue(s)

Fixes #


francostellari:
I don't understand why we need to antepose `center-` to everything to distinguish it from what? There is no other Dockerfile, helm-chart

MikeSpreitzer:
There _is_ one other container image being made now, for the syncer.  We plan to make more images in the future.

clubanderson:
Why aren't we using "core" instead of "center"?  We use the term "core" to describe our components (sans the syncer) when we work with others. If we describe it as "center" it implies that there are parts above and below that we need to supply. While there is a "below", we don't have an "above" component. While I understand what you are trying to imply @MikeSpreitzer , I would prefer we use "core" if anything.

MikeSpreitzer:
Yes, I think "core" is an improvement on "center".

MikeSpreitzer:
The force-push above adds the terminology shift from "center" to "core".

clubanderson (commented):
/approve

francostellari:
@MikeSpreitzer when you have time, can you please rebase and merge.
I need this for my work on multi-containers.

PR #1044, opened 2023-09-28
✨ Change how kubectl-kubestellar-deploy determine the container tag

francostellari:
## Summary

I think that the plugin should select the tag image based on the plugin release version, or default to stable.

## Related issue(s)

Fixes #

### DIFF
--- a/scripts/kubectl-kubestellar-deploy
+++ b/scripts/kubectl-kubestellar-deploy
@@ -20,7 +20,10 @@
 
 KSDIR=$(cd $(dirname ${BASH_SOURCE[0]}); cd ..; pwd)
 
-image_tag=b23-09-15-04-52-13
+image_tag=$(kubestellar-release 2> /dev/null)
+if [ "$image_tag" == "" ]; then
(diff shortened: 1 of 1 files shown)


scripts/kubectl-kubestellar-deploy
```diff
@@ -20,7 +20,10 @@
 
 KSDIR=$(cd $(dirname ${BASH_SOURCE[0]}); cd ..; pwd)
 
-image_tag=b23-09-15-04-52-13
+image_tag=$(kubestellar-release 2> /dev/null)
```
MikeSpreitzer, line 23: This will not work because `kubestellar-release` is a special thing that only exists in release archives.  We need developers to be able to build and deploy whatever version they are working on.  I suspect the motivation here is the problem you reported with using recently built images. I think the right fix is the other one that you suggested, which I have put into #1047 .

MikeSpreitzer:
#1047 has merged now.
#1051 updates the default reference to match a build of what `main` points to right now.

PR #1042, opened 2023-09-27
📖 Start actually writing the user guide

MikeSpreitzer:
## Summary
This PR starts to fill in the user guide with consumer-oriented material.  This is still a work-in-progress.

## Related issue(s)

This is intended to eventually address #1023

PR #1052, opened 2023-09-29
✨ Add initial user spaces

### LINKED ISSUE #1048: feature: there should be at least 1 inventory and 1 workload management workspace created with KubeStellar by default
### Feature Description

KubeStellar is a bunch of parts that must be assembled before anything meaningful can be done with them.  

### Proposed Solution

In order to get closer to a useful solution it would be good to create 1 inventory and 1 workload management workspace created immediately following the containerized or non-containerized version of KubeStellar.

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR expands the functionality of `kubestellar init` to create an IMW and a WMW.

This is built on top of #1042, and will be rebased after that merges.

## Related issue(s)

Fixes #1048


MikeSpreitzer:
/cc @clubanderson 
/cc @dumb0002 
@cwiklik 
/cc @KPRoche 
/cc @francostellari

clubanderson:
@MikeSpreitzer please remove anything from this PR that is not directly related to creating an IMW and WMW out-of-the-box with KubeStellar.  KubeStellar is a simple tool that needs to take location/synctarget/edgeplacement out-of-the-box.  I have given guidance on a user guide and diagram in other communications.  We just need this PR to focus on creating and testing (docsecutable) the default creation of 1 imw and 1 wmw by default with initial installation of kubestellar.

scripts/kubestellar
```diff
@@ -256,31 +259,42 @@ function ensure_espw() {
...
+function ensure_imw1() {
+    kubectl ws root &> /dev/null
+    if kubectl get ws imw1 &> /dev/null
+    then kubectl ws imw1 &> /dev/null
+    else kubectl ws create imw1 --enter &> /dev/null
+    fi
+    
+    if ! kubectl get apibinding "edge.kubestellar.io" &> /dev/null; then
+	kubectl kcp bind apiexport root:espw:edge.kubestellar.io
+    fi
```
ezrasilvera, line 272: Do you want to go back to root after you finishe

scripts/kubestellar
```diff
@@ -256,31 +259,42 @@ function ensure_espw() {
...
+}
 
-# Check placement-translator is already running
-if [ $(process_running placement-translator) == "running" ]
-then
-    echo "An older deployment of placement-translator is already running - terminating it ...."
-    pkill -f placement-translator
+if [ "$subcommand" != init ]; then
+    
+    # Check mailbox-controller is already running
```
ezrasilvera, line 277: But this changes the functionality of "init" compared to what you had before. Before you just checked the existence of the spaces

scripts/kubestellar
```diff
@@ -256,31 +259,42 @@ function ensure_espw() {
...
+    # Check placement-translator is already running
+    if [ $(process_running placement-translator) == "running" ]
+    then
+	echo "An older deployment of placement-translator is already running - terminating it ...."
+	pkill -f placement-translator
+    fi
+    
 fi
 
 if [ $subcommand == stop ]; then
```
ezrasilvera, line 300: Not related to this PR. I find it confusing to have an explicit "stop" subcommand that is based on the fact that you kill the process by default ..

ezrasilvera:
Couldn't find a good place to put the comment: In the kubestellar script it seems you start the controllers also on `init`  and not only on `start`. 
BTW, why do we need two distinct command for init & start ?  What is the usage of "init" without start ?

MikeSpreitzer:
@ezrasilvera : in https://github.com/kubestellar/kubestellar/pull/1052#issuecomment-1746814832 you start with pointing out a bug that was fixed in #1057.

scripts/kubestellar
MikeSpreitzer: That was a design choice made long ago and not part of this PR. Feel free to open an issue or PR if you want to suggest a better implementation or a change to the design for the set of platform management commands.

scripts/kubestellar
MikeSpreitzer: You are debating the behavior of the internal script function `ensure_imw1`, right? #1057, which differs a little in this script, has been merged instead of this PR. If you have a better idea for the internal details of the merged script, please open an issue or PR.

PR #1057, opened 2023-09-29
✨ Make kubestellar init create first IMW and WMW

MikeSpreitzer:
## Summary
This PR expands the functionality of `kubestellar init` to create an IMW and a WMW.

This PR is an alternative to #1052; unlike that one, this one does NOT build on #1042.

## Related issue(s)

Fixes #1048

PR #1047, opened 2023-09-28
🐛 Adopt hack for OpenShift vs amd64/v2 problem

MikeSpreitzer:
## Summary
This PR stupidifies the built images to not mention that v2 of the x86-64 ISA is required, because OpenShift nodes refuse to pull images that _do_ say they are built for that (even when run on a machine that has those instructions). This PR also adds a note to the building instructions about this.

## Related issue(s)

Fixes #

PR #1077, opened 2023-10-03
✨Introduce Helm chart and entry.sh for implementing cluster deployment in separated containers

francostellari:
## Summary

- Updated `entry,sh` to support running separated processes
- Updated `core-helm-chart` to:
  - Removed the mandatory `kubestellar` namespace. Use `--namespace <name>` instead or install in a new-project
  - Persist the PKI and kcp kubeconfig
  - Run kubestellar as 5 separated containers in a pod: kcp core mailbox-controller where-resolver placement-translator
  - Enable setting the ESPW name via value `ESPW_NAME`
  - Enable ensuring 0, 1, or more inventories using value `ENSURE_INV`
  - Enable ensuring 0, 1, or more WMW using value `ENSURE_WMW`
  - Access individual containers logs:

```shell
kubectl logs $(kubectl get pod --selector=app=kubestellar -o jsonpath='{.items[0].metadata.name}') -c kcp
kubectl logs $(kubectl get pod --selector=app=kubestellar -o jsonpath='{.items[0].metadata.name}') -c init
kubectl logs $(kubectl get pod --selector=app=kubestellar -o jsonpath='{.items[0].metadata.name}') -c mailbox-controller
kubectl logs $(kubectl get pod --selector=app=kubestellar -o jsonpath='{.items[0].metadata.name}') -c where-resolver
kubectl logs $(kubectl get pod --selector=app=kubestellar -o jsonpath='{.items[0].metadata.name}') -c placement-translator
```

Retrieve the admin.kubeconfig:
```shell
kubectl get secrets kubestellar -o 'go-template={{index .data "external.kubeconfig"}}' | base64 --decode > admin.kubeconfig
````

## Related issue(s)
#995 not working for kcp v0.11.0

Fixes #
#992
#993
#1048
#1063

### CODE REVIEW

File: core-helm-chart/templates/deployment.yaml [hunk 1]
```diff
@@ -15,37 +15,106 @@
...
+          runAsNonRoot: true
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop: ["ALL"]
+          seccompProfile:
+            type: RuntimeDefault
         image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
         imagePullPolicy: {{ .Values.image.pullPolicy }}
+        # command: [ "sleep" , "infinity" ]
+        command: [ "/home/kubestellar/entry.sh" , "kcp" ]
```
MikeSpreitzer, line 43: Could we get verbosity control on this one too?

core-helm-chart/templates/deployment.yaml
```diff
@@ -15,37 +15,106 @@
...
         env:
```
MikeSpreitzer, line 44: Can we get verbosity control here too?

core-container/entry.sh
```diff
@@ -16,88 +16,402 @@
...
+    echo "Waiting for kcp to be ready... this may take a while."
+    (
+        KUBECONFIG=
+        until [ "$(kubectl logs $(kubectl get pod --selector=app=kubestellar -o jsonpath='{.items[0].metadata.name}') -c kcp | grep '***READY***')" != "" ]; do
+            sleep 5
+        done
+    )
+    echo "Succes!"
+    echo "Copying the admin.kubeconfig from kubestellar seret..."
+    mkdir -p /home/kubestellar/.kcp
```
MikeSpreitzer, line 38: This directory must already exist, the Deployment makes it so.

core-container/entry.sh
```diff
@@ -16,88 +16,402 @@
...
+    (
+        KUBECONFIG=
+        kubectl get secrets kubestellar -o 'go-template={{index .data "admin.kubeconfig"}}' | base64 --decode > /home/kubestellar/.kcp/admin.kubeconfig
+    )
```
MikeSpreitzer, line 42: This seems more complicated than necesssary. Why not just mount the persistent volume at `home/kubestellar/.kcp` for every container?

core-container/entry.sh
```diff
@@ -16,88 +16,402 @@
...
+    name=$(grep name: "$filename" | head -1 | awk '{ print $2 }')
+    if kubectl get "$kind" "$name" &> /dev/null ; then
+        kubectl replace -f "$filename"
+    else
+        kubectl create -f "$filename"
+    fi
+}
 
-mkdir -p ${PWD}/kubestellar-logs # required in Open Shift clusters
+function ensure_root_compute_configd() {
```
MikeSpreitzer, line 88: Why is there a copy of material from the `kubestellar` script here? Why not invoke the script?

core-helm-chart/templates/deployment.yaml
francostellari: @MikeSpreitzer can you please helm me by letting me know the verbosity argument for kcp start?
Also would the "value" be the same than the others.
For simplicity I'm using a common one at the moment.

core-container/entry.sh
francostellari: @MikeSpreitzer thank you for the suggestion.
I think this is "could" ideally be achieved by running `kubestellar init` but as the script stands right now, I think  there is a previous check that I cannot pass.
Perhaps in the future we can find a way to modify the script to work for both purposes.
My quick solution was to copy part of the code.
I'm not quite sure why some of the objects are in the script and some are in the cofig folder.
Perhaps a more clean solution would be to put everything in the config.
```
# Check if kcp is running
if [ $(process_running kcp) != "running" ]
then
    echo "kcp is not running - please start it ...."
    exit 1
fi
```

francostellari:
@MikeSpreitzer I think the most critical gap is to find a way to restart kcp properly with the pki CA being maintained and have a functional admin.kubeconfig. So far I have not been able to do it. Please help if you have time.

francostellari:
@clubanderson the latest commit should resolve https://github.com/kubestellar/kubestellar/issues/1048

core-helm-chart/templates/deployment.yaml
MikeSpreitzer: `kcp start` takes the usual verbosity argument, `-v`.
I think it makes sense to give individualized verbosity control, not force the same verbosity level to apply to all four processes.
It might be good to frame this such that if the user does not override any values then none of the processes is given a `-v` (each applies its own default), but there are four values that the user can optionally set to cause a specific `-v=N` to be given to a process.

None of this verbosity stuff is a blocker for this PR; we can do more in follow-ons.

core-container/entry.sh
MikeSpreitzer: I would prefer to not merge a duplication of scripting.  Please let me know what stops this from re-using `kubestellar init` and let's get that fixed first.

core-helm-chart/templates/deployment.yaml
francostellari: @MikeSpreitzer I will implement in a separated PR. I'm concerned to mess this one up.

MikeSpreitzer:
@francostellari : can you please bump the chart's major version number to 1?

MikeSpreitzer:
What's with the file named `0` introduced by this PR?

core-container/entry.sh
```diff
@@ -16,88 +16,487 @@
...
+        export EXTERNAL_HOSTNAME=$(guess_kcp_dns)
     fi
-else
-    if kubectl create secret generic kubestellar --from-file="${PWD}/.kcp/admin.kubeconfig" 2> /dev/null; then
-        echo " Created secret in the current namespace."
+    echo "EXTERNAL_HOSTNAME=${EXTERNAL_HOSTNAME}"
+
+    # Create the certificates
+    if [ -n "$EXTERNAL_HOSTNAME" ]; then
+        echo "Creating the TSL certificates"
```
MikeSpreitzer, line 352: Typo: "TSL" vs "TLS"

core-container/entry.sh
```diff
@@ -16,88 +16,487 @@
...
+        # mkdir -p .kcp
+        cd .kcp
+        eval pieces_external=($(kubestellar-ensure-kcp-server-creds ${EXTERNAL_HOSTNAME}))
+        eval pieces_cluster=($(kubestellar-ensure-kcp-server-creds ${KUBESTELLAR_SERVICE})) #############
+        cd ..
+        echo "TLS certificates for server ${EXTERNAL_HOSTNAME}:"
+        echo ${pieces_external[0]}
+        echo ${pieces_external[1]}
+        echo ${pieces_external[2]}
```
MikeSpreitzer, line 361: Let's not log the private key

core-container/entry.sh
```diff
@@ -16,88 +16,487 @@
...
+if [ "$VERBOSITY" == "" ]; then
+    VERBOSITY="2"
 fi
+if [ "$ESPW_NAME" == "" ]; then
+    ESPW_NAME="espw"
+fi
+echo "ESPW_NAME=${ESPW_NAME}"
+echo "VERBOSITY=${VERBOSITY}"
+if [ "$ENSURE_IMW" == "" ]; then
+    ENSURE_IMW=""
```
MikeSpreitzer, line 476: What's the point of changing the value from empty string to empty string?

core-container/entry.sh
```diff
@@ -16,88 +16,487 @@
...
+
+function run_init() {
+    echo "--< Starting init >--"
+    wait_kcp_ready
+    kubestellar init --local-kcp false
+    # ensure_root_compute_configd
+    # ensure_espw
+    kubectl ws root
+    ensure_invs $ENSURE_IMW
+    ensure_wmws $ENSURE_WMW
```
MikeSpreitzer, line 419: Please do not make the chart differ from the behavior of `kubestellar init`.  Either generalize `kubestellar init` to cover the expanded options or remove them from this edition of the chart and revisit in a later PR.

core-container/entry.sh
```diff
@@ -16,88 +16,487 @@
...
-    if kubectl get route kubestellar-route &> /dev/null; then
-        EXTERNAL_HOSTNAME=$(kubectl get route kubestellar-route -o yaml -o jsonpath={.spec.host} 2> /dev/null)
+
+function echoerr() {
+   echo "ERROR: $1" >&2
+}
+
+
+function ensure_inv() {
+    while IFS=':' read -ra items; do
```
MikeSpreitzer, line 28: Now that we avoid workspace hierarchy this can be much simpler.

core-container/entry.sh
```diff
@@ -16,88 +16,487 @@
...
+            ensure_inv "$i"
+            kubectl ws use $initial_ws &> /dev/null # go back to the initial state
+        done
+    done <<< "$1"
+
+    kubectl ws use $initial_ws &> /dev/null
+}
+
+
+function ensure_wmw() {
```
MikeSpreitzer, line 61: Now that we avoid workspace hierarchy this can be much simpler.

core-container/entry.sh
francostellari: Implemented

core-container/entry.sh
```diff
@@ -16,88 +16,487 @@
...
+# ---
+# apiVersion: rbac.authorization.k8s.io/v1
+# kind: ClusterRole
+# metadata:
+#   name: compute:apiexport:kubernetes:maximal-permission-policy-extended
+# rules:
+# - apiGroups:
+#   - ""
+#   - apps
+#   - networking.k8s.io
```
MikeSpreitzer, line 212: Not loving a PR that introduces a bunch of commented out code.

francostellari:
> @francostellari : can you please bump the chart's major version number to 1, so client code gets a strong signal that it creates different structure now? The chart's version number can describe the chart independent of (but I am not saying orthogonal to) the MCCM version that it deploys.

Done

core-container/entry.sh
francostellari: Removed, I commented it during test.
New code is working.

core-container/entry.sh
```diff
@@ -16,88 +16,487 @@
...
+# }
+
+
+function run_kcp() {
+    echo "--< Starting kcp >--"
+
+    echo Attempting to delete kubestellar secret...
+    (
+        KUBECONFIG=
+        ! kubectl delete secret kubestellar
```
MikeSpreitzer, line 339: Why the `!` here?
This is in the context of `set -e`, right? That means this command will fail the script if the Secret is not present at the start of the command, right? Wouldn't it be better to make the script not fail regardless of the presence or absence of the Secret before this command?

core-container/entry.sh
francostellari: I would rather keep this

core-container/entry.sh
francostellari: ops

core-container/entry.sh
francostellari: ok, I was using it for debug

core-container/entry.sh
francostellari: thanks, done

core-container/entry.sh
```diff
@@ -16,88 +16,487 @@
...
+        echo -n "Running kcp without TLS keys... "
+        kcp start &
+    fi
+    echo Started.
+
+    # Waiting to be ready
+    echo "Waiting for ${KUBECONFIG}..."
+    while [ ! -f "${KUBECONFIG}" ]; do sleep 5; done
+    echo 'Waiting for "root:compute" workspace...'
+    until [ "$(kubectl ws root:compute 2> /dev/null)" != "" ]; do sleep 5;done
```
MikeSpreitzer, line 384: I do not like the fact that we have multiple blocks of code to do this waiting, and they do it differently.
Not a blocker, but it should be cleaned up soon.

francostellari:
> What's with the file named `0` introduced by this PR?

sorry, removed

core-container/entry.sh
francostellari: my understanding and testing suggest that the ! avoids the script to fail on error with set -e

core-container/entry.sh
francostellari: I'm not sure I understand

core-container/entry.sh
francostellari: I'm not sure I understand

core-container/entry.sh
MikeSpreitzer: I would expect an invocation of `kubectl` before launching the server to fail!  Why is this command here at all?  What good do you think it does?

MikeSpreitzer:
Perhaps `core-helm-chart/README.md` should have been deleted earlier. I left it only for internal consumption. What users should be reading is the documentation in the https://docs.kubestellar.io/main/Coding%20Milestones/PoC2023q1/commands/#deployment-into-a-kubernetes-cluster section. And this PR needs to keep the commands in that section working.

MikeSpreitzer:
The title of this PR is currently wrong, right?  This PR does not _introduce_ the Helm chart.

francostellari:
back to /hold
config error

MikeSpreitzer:
I am running through example1 using the deploy-as-workload option. It gets stuck at the following point (I start this excerpt with a `get pods` to show state).

```
bash-3.2$ kubectl get pods -A
NAMESPACE            NAME                                            READY   STATUS      RESTARTS   AGE
default              kubestellar-84fdf66696-dl9p5                    5/5     Running     0          113s
ingress-nginx        ingress-nginx-admission-create-2k96v            0/1     Completed   0          5m46s
ingress-nginx        ingress-nginx-admission-patch-qglgh             0/1     Completed   0          5m46s
ingress-nginx        ingress-nginx-controller-7456fdc894-slr78       1/1     Running     0          5m46s
kube-system          coredns-5d78c9869d-248xx                        1/1     Running     0          5m57s
kube-system          coredns-5d78c9869d-jxrg8                        1/1     Running     0          5m57s
kube-system          etcd-ks-host-control-plane                      1/1     Running     0          6m11s
kube-system          kindnet-t2jpd                                   1/1     Running     0          5m58s
kube-system          kube-apiserver-ks-host-control-plane            1/1     Running     0          6m11s
kube-system          kube-controller-manager-ks-host-control-plane   1/1     Running     0          6m11s
kube-system          kube-proxy-ghtr7                                1/1     Running     0          5m58s
kube-system          kube-scheduler-ks-host-control-plane            1/1     Running     0          6m11s
local-path-storage   local-path-provisioner-6bc4bddd6b-5flmh         1/1     Running     0          5m57s

bash-3.2$ kubectl kubestellar get-external-kubeconfig -o kcs.kubeconfig -X
+ shift
+ ((  0 > 0  ))
+ '[' -z kcs.kubeconfig ']'
+ set -e
++ kubectl get pods -n kubestellar -l app=kubestellar-server -o 'jsonpath={.items[0].metadata.name}'
+ server_pod=
+ sleep 10
++ kubectl get pods -n kubestellar -l app=kubestellar-server -o 'jsonpath={.items[0].metadata.name}'
+ server_pod=
+ sleep 10
++ kubectl get pods -n kubestellar -l app=kubestellar-server -o 'jsonpath={.items[0].metadata.name}'
+ server_pod=
+ sleep 10
^C
```

MikeSpreitzer:
See git history problems here now.  Currently this PR diverges from `main` after commit 523ce181bd4b2b8268f71ac3b10a512fb03ce318 , which was the merge of #1047 last week. That won't pick up the recent improvements to `kubestellar init` (#1057, #1089), which merged later.

In this PR now, `scripts/kubectl-kubestellar-deploy` sets the default core image tag to a build of commit 8259fa7c7 , which is later than 523ce181bd4b2b8268f71ac3b10a512fb03ce318 but does not include the changes in this PR.

MikeSpreitzer:
Also, the tag `main-mc` currently refers to a multi-platform manifest that was built from commit c722cbdc , which also does not include all of the changes in this PR.

The good news is that this PR is still consistent with `main` regarding the default syncer image tag.

I am going to launch a build of the current version of this PR.

MikeSpreitzer:
BTW, when you rebase, it would probably be good to squash together all those README updates.

MikeSpreitzer:
Oh! "Merge branch 'kubestellar:main' into helm-chart-separated-containers"! Please do not do that. It violates the discipline we agreed to regarding keeping the git history simple.

MikeSpreitzer:
I just built a core image for commit cd4136ab4 , and while doing that noticed that the changes since c722cbdc are non-functional.

MikeSpreitzer:
So I updated `kubectl-kubestellar-deploy` to ask for core image tag "git-c722cbdc-clean" (same image as "main-mc" at the moment). Deleted the old chart instance and made a new one with the updated deploy plugin. But still the get-external-kubeconfig plugin is looking for the wrong pod.

```
bash-3.2$ kubectl kubestellar get-external-kubeconfig -o kcs.kubeconfig -X
+ shift
+ ((  0 > 0  ))
+ '[' -z kcs.kubeconfig ']'
+ set -e
++ kubectl get pods -n kubestellar -l app=kubestellar-server -o 'jsonpath={.items[0].metadata.name}'
+ server_pod=
+ sleep 10
++ kubectl get pods -n kubestellar -l app=kubestellar-server -o 'jsonpath={.items[0].metadata.name}'
+ server_pod=
+ sleep 10
...
```

MikeSpreitzer:
And this tells why: the namespace is not what's expected.

```
bash-3.2$ kubectl get pods -A -L app
NAMESPACE            NAME                                            READY   STATUS      RESTARTS       AGE    APP
default              kubestellar-cfb965b4d-gnvhm                     5/5     Running     1 (4m5s ago)   4m9s   kubestellar
...
```

francostellari:
> See git history problems here now. Currently this PR diverges from `main` after commit [523ce18](https://github.com/kubestellar/kubestellar/commit/523ce181bd4b2b8268f71ac3b10a512fb03ce318) , which was the merge of #1047 last week. That won't pick up the recent improvements to `kubestellar init` (#1057, #1089), which merged later.
> 
> In this PR now, `scripts/kubectl-kubestellar-deploy` sets the default core image tag to a build of commit [8259fa7](https://github.com/kubestellar/kubestellar/commit/8259fa7c7aab17533839ffbd975c91192e3f8836) , which is later than [523ce18](https://github.com/kubestellar/kubestellar/commit/523ce181bd4b2b8268f71ac3b10a512fb03ce318) but does not include the changes in this PR.

I rebased but did not squash the cmmits.
I know for sure that I would screw up something

francostellari:
> Oh! "Merge branch 'kubestellar:main' into helm-chart-separated-containers"! Please do not do that. It violates the discipline we agreed to regarding keeping the git history simple.

I'm sorry I don't understand and I don't know how to fix this.

francostellari:
> And this tells why: the namespace is not what's expected.
> 
> ```
> bash-3.2$ kubectl get pods -A -L app
> NAMESPACE            NAME                                            READY   STATUS      RESTARTS       AGE    APP
> default              kubestellar-cfb965b4d-gnvhm                     5/5     Running     1 (4m5s ago)   4m9s   kubestellar
> ...
> ```

The deployment ns is now a user choice, as requested.

MikeSpreitzer:
I am going to continue testing what I have, choosing to put the core into the `kubestellar` namespace.

MikeSpreitzer:
Ugh, this is why I opened #414.

```
bash-3.2$ kubectl kubestellar deploy --external-endpoint yep.yep:6443 -n kubestellar --create-namespace
/Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin/kubectl-kubestellar-deploy: missing value for long flag --create-namespace
```

MikeSpreitzer:
After manually creating the kubestellar namespace and doing the deploy, we are closer but still no cigar, different label.

```
bash-3.2$ kubectl get pods -A -L app
NAMESPACE            NAME                                            READY   STATUS      RESTARTS   AGE    APP
...
kubestellar          kubestellar-cfb965b4d-d5d6c                     5/5     Running     0          52s    kubestellar
...
```

MikeSpreitzer:
> I rebased

No, you did not.  You merged `main` into this PR's branch.  Please do not do that.  It violates the discipline that we agreed to for keeping simple git history.

MikeSpreitzer:
run_kcp insists on FAILURE of the command to delete the Secret.

the ready file is created in the init container, but currently other plugins seek it in the default container.

MikeSpreitzer:
out-of-the-box spaces a bit incoherent:

```
bash-3.2$ kubectl get Workspaces
NAME                                                       TYPE           REGION   PHASE   URL                                                  AGE
1lce1d0yfx2ti8lb-mb-53b3a371-5df9-4799-92f4-0462c3d24611   organization            Ready   https://10.244.0.29:6443/clusters/2a9byjmlatxgls74   23s
1lce1d0yfx2ti8lb-mb-7ccf8af2-deb3-4336-be7c-4a336f871464   organization            Ready   https://10.244.0.29:6443/clusters/27nbqr8af6tu7rpi   29s
compute                                                    universal               Ready   https://10.244.0.29:6443/clusters/3yd4rtw13irl9qkz   20m
espw                                                       organization            Ready   https://10.244.0.29:6443/clusters/2msbc26i71vb9b07   20m
imw1                                                       organization            Ready   https://10.244.0.29:6443/clusters/1lce1d0yfx2ti8lb   20m
inv1                                                       organization            Ready   https://10.244.0.29:6443/clusters/2wfiqqxbuhpm1rci   20m
wmw1                                                       organization            Ready   https://10.244.0.29:6443/clusters/ftdvvmtpu621oqtj   20m
```

MikeSpreitzer:
I successfully ran through example1 with deploying the core to a kind cluster.

/lgtm
/approve
Thanks!

PR #1089, opened 2023-10-05
✨ Improve `kubestellar init` wrt waiting for kcp

MikeSpreitzer:
## Summary
This PR does two things.
(1) Make waiting for a "kcp" process optional.
(2) Add non-optional waiting for the server to be fully ready.

## Related issue(s)

I think this needs to go in before #1077 so that can use this.

### DIFF
--- a/scripts/kubestellar
+++ b/scripts/kubestellar
@@ -24,6 +24,7 @@
 # Requirements:
 #    Download KubeStellar binaries
 #    KubeStellar controller binaries are on $PATH.
+#    kubectl is configured to talk to kcp server with high privilege
 
 bindir="$(dirname "$0")"
 
@@ -37,6 +38,7 @@ remove=0
 cleanup=0
 espw_name="espw"
 log_folder=$(pwd)/kubestellar-logs
+local_kcp=true
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
The force-push to 78b243b16: adds timing to the wait messages, and changes the checked workspace to `root:compute`.

MikeSpreitzer:
And the force-push to 810670655 is the doc update.

scripts/kubestellar
```diff
@@ -302,6 +305,21 @@ if [ $subcommand == stop ]; then
...
+# Check if kcp is running
+if [ "$local_kcp" == true ] && [ $(process_running kcp) != "running" ]; then
+    echo "kcp is not running - please start it ...."
+    exit 1
+fi
+
+let tries=1
+while ! ( kubectl get workspaces && kubectl ws root:compute ); do
+    if (( ( tries & (tries-1) ) == 0 ))
+    then echo "$0: At $(date), kcp server not running yet, waiting $(( (tries+2)/2 * 15 )) sec..." >& 2
```
ezrasilvera, line 317: let's also add "Or root:compute is not available"  to the message

MikeSpreitzer:
Recent force-pushes update the waiting message and wait time arithmetic.

scripts/kubestellar
```diff
@@ -302,6 +305,21 @@ if [ $subcommand == stop ]; then
...
 fi
 
+# Check if kcp is running
+if [ "$local_kcp" == true ] && [ $(process_running kcp) != "running" ]; then
+    echo "kcp is not running - please start it ...."
+    exit 1
+fi
+
+let tries=1
+while ! ( kubectl get workspaces && kubectl ws root:compute ); do
```
ezrasilvera, line 315: Do we have any test to verify that kubectl ws plugin is in the path ?

MikeSpreitzer: well, this is testing that.

MikeSpreitzer:
If we want the Helm chart to just launch pods without any synchronization (which I think we do), then each needs to wait for the preconditions that it needs.

I would be willing to entertain a time limit on the waiting for kcp and root:compute.  That would require deciding on the default time limit.

scripts/kubestellar
ezrasilvera: OK. I meant distinctly. Fine for now, for the future we may want separate tests for better root cause analysis

Issue #1058, opened 2023-09-29
bug: kubestellar kubectl plugins should detect Kubestellar core version and avoid throw error if incompatible


clubanderson:
### Describe the bug

I am able to run any version of the kubestellar kubetl plugins against a kubestellar endpoint.  The problem is that not all versions of the plugins are compatible with all versions of the Kubestellar Core version running.  we need to put a mechanism in place to stop any incompatibility and notify the user.

### Steps To Reproduce

1. install kubectl kubestellar plugin v 0.6
2. run plugin against v0.7
3. weird behavior occurs

### Expected Behavior

error should be thrown to avoid confusion from user or misconfiguration.

MikeSpreitzer:
Agreed. There is a non-trivial design issue here regarding what "compatible" means. We could strive for the utmost in accuracy, allowing any combination that will get what the user expects done. Thinking about all that entails scares me. I suspect the more practical approach is to define a more conservative check: success implies that what the user expects will get done, but failure does not guarantee that there would actually be a problem. But while I am willing to take some of that liberty in principle, the more we take in practice the less usable the resulting system will be. So there is some interesting design to do here.

To inform that design, I would like to understand what went wrong to prompt #1049 .

clubanderson:
I'm happy with exact match between version of plugin and version of core. That's good enough for now. And if there is a mismatch, offer the plugin user the brew command to get the version they need. Simple. 

The bug I closed turned out to be scripts that were mismatched. I deleted anything and everything related to kubestellar in my path, and then pull v0.7.0 bin and then it worked again. To before I did that, I asked braulio to access the core from his laptop and he didn't have my issue. So that confirmed it for me. No reason to follow a flake, let's just try to get to parity and a suggestion if not parity.

MikeSpreitzer:
It would be unsurprising, and thus probably not dissatisfying, to users for us to apply one [semantic version](https://semver.org/) to all of KS MCCM and do the checking based on that. To do that properly and achieve the behavior you desire, we would need to graduate from major version 0 to major version 2 (at first) and increment the major version again with any backward incompatible change. To avoid incrementing the major version unnecessarily, we would need to increment the minor version when needed and the patch version otherwise; this contrasts to current practice which makes only one release on each release branch.

[...]

PR #1092, opened 2023-10-06
⚠️ Increment major version in Kubernetes API objects

MikeSpreitzer:
## Summary
This PR increments the major version of the kube API for multi-cluster configuration management. We have made breaking changes since the last release, and it is normal to increment the major version in that case. Also, this is part of responding to #1058 .

I simply changed the version on the whole API. I did not worry about supporting a user who wants to do a rolling upgrade, because I think we have none of those.

## Related issue(s)

Fixes #


MikeSpreitzer:
CI successfully ran example1 with deployment as bare processes.
I just successfully ran example1 with deployment to a kind cluster on my laptop.

yana1205:
I also see no issue from the results of unit test/e2e test. I can approve this change but I'm not sure the current strategy to support backward compatibility in KubeStellar. If it's OK with supporting only v2alpha1, I'm fine with this PR. (I understand that supporting both versions at the development phase is too much effort to go forward.)

yana1205:
/lgtm

I also see no issue from the results of unit test/e2e test so the change looks good to me.

However, I'm not sure the current strategy to support backward compatibility in KubeStellar. I understand that supporting both versions at the development phase is too much effort to go forward.

MikeSpreitzer:
Currently we explicitly do NOT support any backward compatibility. So that is part of why this change is good to do now. Both because it is not harmful, and because NOT incrementing the major version would generate unnecessary confusion in users.

clubanderson (approved):
/approve

Issue #1102, opened 2023-10-09
Thanos expects rough play


MikeSpreitzer:
Thanos creates a ConfigMap named `user-workload-monitoring-config` in the namespace `openshift-user-workload-monitoring` and expects users to modify it. The syncer avoids modifying it because that object lacks the annotation that the syncer uses to mark objects managed by the syncer. The syncer is trying to play too nicely.

clubanderson:
@MikeSpreitzer is this issue closed as a matter of #1103 and #1105 merging?  Is there more to do?

MikeSpreitzer:
Implemented by #1103 and #1105

clubanderson:
Reopening to allow for a test first with upcoming release.

clubanderson:
we are now using customization to achieve the post-edit.

PR #1103, opened 2023-10-09
✨ Make the syncer more assertive

MikeSpreitzer:
## Summary
This makes the syncer modify existing objects in the WEC regardless of whether they have the annotation that the syncer applies. This is because Thanos expects such rough play. Hopefully this will not break anything!

## Related issue(s)

This is intended to address #1102 in the `release-0.7` branch.


clubanderson (commented):
/lgtm
/approve

PR #1105, opened 2023-10-09
✨ Make the syncer more assertive

MikeSpreitzer:
## Summary
This makes the syncer modify existing objects in the WEC regardless of whether they have the annotation that the syncer applies. This is because Thanos expects such rough play. Hopefully this will not break anything!

## Related issue(s)

This is intended to address https://github.com/kubestellar/kubestellar/issues/1102 in the `main` branch.


clubanderson (commented):
/lgtm
/approve

Issue #1098, opened 2023-10-07
🐛 kubestellar init needs a switch to suppress creation of default inv and wmw.


clubanderson:
### Describe the bug

Our helm chart creates a set of default inv and wmw based on values. The default values already have 1 of each. If we are to continue using KubeStellar init in the helm workflow, then we should not create 2 sets of workspaces. We should suppress the kibestellar init wmw and inv and use the helm wmw and inv set that is created. 

### Steps To Reproduce

1. Install kubestellar using helm
2. Observe inv and wmw from kubestellar init run
3. Observe wmw and inv set created by helm workflow. 

### Expected Behavior

We want the wmw and inv set that the helm chart creates. Not the ones that kubestellar init creates.

clubanderson:
@francostellari will this item be completely satisfied with your work on #1111?

francostellari:
Yes I believe so

clubanderson:
Closing. Thank you @francostellari

clubanderson:
Reopening to allow for a test first with upcoming release.

francostellari:
`kubestellar init --ensure-imw "" and --ensure-wmw ""` suppresses the creation of any IMW or WMW

PR #1111, opened 2023-10-10
✨ Ensure IMWs and WMWs across all scripts

francostellari:
## Summary

Create a consistent experience to ensure IMWs and WMWs across all scripts.
- update `kubestellar init` to accept `--ensure-imw "imw-list"`  and `--ensure-wnw "wmw-list"`, where the list are comma separated and accept fully qualified kcp paths (default is `imw1` and `wmw1` to respect current behavior, use "" to create nothing)
- update `bootstrap-kubestellar` to leverage the above `kubestellar init` code instead of its own copy of the code
- update container `entry.sh`  to leverage the above `kubestellar init` code instead of its own copy of the code

## Related issue(s)

Fixes #
#1098

### DIFF
--- a/docs/content/Coding Milestones/PoC2023q1/commands.md
+++ b/docs/content/Coding Milestones/PoC2023q1/commands.md
@@ -230,6 +230,8 @@ not meaningful for the `stop` subcommand.
   `${PWD}/kubestellar-logs`.
 - `--local-kcp $bool`: says whether to expect to find a local process
   named "kcp".  Defaults to "true".
+- `--ensure-imw`: provide a comma separated list of fully qualified inventory workspaces, _e.g._ "root:imw1,root:test:imw2". Defaults to "root:imw1". To prevet the creation of any inventory workspace, then pass "".
+- `--ensure-wmw`: provide a comma separated list of fully qualified workload management workspaces, _e.g._ "root:org:wmw1,root:org:imw2". Defaults to "root:wmw1". To prevet the creation of any workload management workspace, then pass "".
 - `-h` or `--help`: print a brief usage message and terminate.
 
 #### Kubestellar init
--- a/docs/content/Coding Milestones/PoC2023q1/commands.md
+++ b/docs/content/Coding Milestones/PoC2023q1/commands.md
@@ -220,7 +220,8 @@ kubestellar [flags] subcommand [flags]
 This command accepts the following command line flags, which can
 appear before and/or after the subcommand.  The `--log-folder` flag is
 only meaningful for the `start` subcommand. The `--local-kcp` flag is
-not meaningful for the `stop` subcommand.
+not meaningful for the `stop` subcommand. The `--ensure-imw` and 
+`--ensure-wmw` flags are only meaningful for the `start` or `init` subcommands.
 
 - `-V` or `--verbose`: calls for more verbose output.  This is a
   binary choice, not a matter of degree.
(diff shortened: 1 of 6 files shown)


clubanderson (commented):
/lgtm

scripts/kubestellar
```diff
@@ -39,9 +39,93 @@ cleanup=0
...
+}
+
+function ensure_imws() {
+    initial_ws=$(kubectl ws . --short)
+    while IFS=',' read -ra items; do
+        for i in "${items[@]}"; do
+            if [ "$i" == "" ] ; then
+                continue
+            fi
+            echo "Ensuring INV: $i"
```
MikeSpreitzer, line 79: The old name is "IMW" (Inventory Management Workspace), the new name is "IS" (Inventory Space).

scripts/kubestellar
```diff
@@ -39,9 +39,93 @@ cleanup=0
...
+    while IFS=':' read -ra items; do
+        len=${#items[@]}
+        if [[ ${len} -eq 0 ]]; then
+            break # nothing to do
+        fi
+        # separate the last item
+        last_item=${items[$len-1]}
+        unset items[$len-1]
+        for i in "${items[@]}"; do
+            if [ "$i" == "" ] ; then
```
MikeSpreitzer, line 98: What if `"$last_item" == ""` ?

MikeSpreitzer:
The doc page needs to be kept consistent.

scripts/kubestellar
francostellari: @MikeSpreitzer Thank you for the observation. I will change to IMW as it it used in a lot of spaces.
I think I need a bigger more coordinated and, breaking, intervention to change all of IMW into IS.

scripts/kubestellar
francostellari: @mike since we are working with a full path line "root:1:2:last"
Then all pieces except the last one can be `ws use` if the exist or `ws create --enter` if they don't
But the last piece must be either "ws use" if it exist or `ks ensure-wmw` if it does not exists

francostellari:
> The doc page needs to be kept consistent.

My kryptonite... I edited commands.mc.

scripts/kubestellar
MikeSpreitzer: Why the special case for empty string?

docs/content/Coding Milestones/PoC2023q1/commands.md
```diff
@@ -230,6 +230,8 @@ not meaningful for the `stop` subcommand.
   `${PWD}/kubestellar-logs`.
 - `--local-kcp $bool`: says whether to expect to find a local process
   named "kcp".  Defaults to "true".
+- `--ensure-imw`: provide a comma separated list of fully qualified inventory workspaces, _e.g._ "root:imw1,root:test:imw2". Defaults to "root:imw1". To prevet the creation of any inventory workspace, then pass "".
```
MikeSpreitzer, line 233: (A) Elsewhere the terminology of pathname is used, rather than fully qualified.

(B) Since we have removed the use of hierarchy, let's give only flat examples: "root:imw1,root:imw2".

francostellari: Ok, I think I made all the changes

scripts/kubestellar
```diff
@@ -50,63 +50,28 @@ function echoerr() {
...
-            fi
-            if ! kubectl ws "$i" &> /dev/null ; then #
-                if ! kubectl ws create "$i" --enter &> /dev/null ; then
-                    echoerr 'unable to create "$i"!'
+            imw=${i#"root:"}
+            echo "Ensuring IMW: $imw"
+            kubectl ws root &> /dev/null
+            if ! kubectl ws "$imw" &> /dev/null ; then #
+                if ! kubectl ws create "$imw" --enter &> /dev/null ; then
+                    echoerr 'unable to create "$imw"!'
```
MikeSpreitzer, line 65: It would be better to display the error message from `kubectl ws create` in this case.

scripts/kubestellar
```diff
@@ -117,12 +82,13 @@ function ensure_wmws() {
             if [ "$i" == "" ] ; then
                 continue
             fi
-            echo "Ensuring WMW: $i"
-            ensure_wmw "$i"
-            kubectl ws use $initial_ws &> /dev/null # go back to the initial state
+            wmw=${i#"root:"}
+            echo "Ensuring WMW: $wmw"
+            kubectl ws root &> /dev/null
+            ${bindir}/kubectl-kubestellar-ensure-wmw $wmw $xflag --with-kube true &> /dev/null
```
MikeSpreitzer, line 88: The `&> /dev/null` makes the `$xflag` moot, right?

scripts/kubestellar
```diff
@@ -117,12 +82,13 @@ function ensure_wmws() {
...
         done
     done <<< "$1"
-    kubectl ws use $initial_ws &> /dev/null
+    kubectl ws use $initial_ws &> /dev/null # go back to the initial state
```
MikeSpreitzer, line 91: What is this command for?  The next one is `kubectl ws root`.

docs/content/Coding Milestones/PoC2023q1/commands.md
MikeSpreitzer: I think so too.

MikeSpreitzer (commented):
I built the core image, tagged `v0.8.0-alpha.8.42d07da-clean`.  I tested it with example1 manually, deploying the core to a kind cluster. It worked.

/approve

Thanks!

PR #1114, opened 2023-10-11
✨ Draft API for identifying immobile parts of downsynced objects

### LINKED ISSUE #1112: feature: use trichotomy rather than dichotomy for downsynced object
### Feature Description

Currently a downsynced object is divided into two parts:
- desired state, which travels from core to WEC, and
- reported state, which travels from WEC to core.

Actually, there is a third part: system state, which is specific to the space and controlled by the apiserver and does not travel (e.g., UID, ResourceVersion).

The feature here is to introduce a third non-system part, state that does not travel. I have heard of two motivations so far.
1. Nick Mitchell wants to have some derived state that is large, derived by a controller in the core without a spurious copy to the WEC.
2. @clubanderson wants to have an object introduced in the core but a part of the object created in the WEC and not transported back to the core.

### Proposed Solution

Introduce a third non-system part of downsynced objects and an API by which users can identify it.

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR introduces the concept of "immobile" parts of downsynced objects and an API for users to identify immobile parts.

This is incomplete, so far there is only the API; the code generation and implementation remain to be done.

## Related issue(s)

Fixes #1112

### DIFF
--- a/pkg/apis/edge/v2alpha1/edge-placement.go
+++ b/pkg/apis/edge/v2alpha1/edge-placement.go
@@ -21,10 +21,18 @@ import (
 )
 
 // EdgePlacement exists in the center and binds (a) a collection of
-// Locations with (b) both (b1) objects in the center to downsync
-// (propagate desired state from center to edge and return reported
-// state from edge toward center), and (b2) a way of identifying objects
+// Locations with (b) both (b1) objects in the center to downsync,
+// and (b2) a way of identifying objects
 // (in edge clusters) to upsync (propagate from edge toward center).
+// For an object subject to downsync, its state is divided into
+// the following four categories.
+//  1. Desired state, which flows from center to edge;
+//  2. Reported state, which flows from edge to center;
+//  3. System-managed state, determined by Kubernetes and beyond the
+//     control of KubeStellar;
+//  4. Immobile state, which also does not propagate but is not determined
+//     by Kubernetes.
+//
 // Both downsyncing and upsyncing are with all of the Locations.  This
 // is not entirely unrelated to a TMC Placement, which directs the
 // selected Namespaces to propagate to _one_ of the selected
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
For example, the following EdgePlacement calls for downsync of a Secret but not propagation of any of its content (only the existence of the object flows from WDS to WEC).

```yaml
apiVersion: edge.kubestellar.io/v2alpha1
kind: EdgePlacement
spec:
  locationSelectors:
  - matchLabels: { "foo", "bar" }
  downsync:
  - apiGroup: ""
    resources: [ secrets ]
    namespaceSelectors:
    - matchLabels: { "foo": "baz" }
    objectNames: [ example-secret ]
    immobileParts:
    - [ data ]
    - [ stringData ]
```

MikeSpreitzer:
For another example, the following EdgePlacement calls for downsync of all Widgets but stops propagation of just one field in the Spec.

```yaml
apiVersion: edge.kubestellar.io/v2alpha1
kind: EdgePlacement
spec:
  locationSelectors:
  - matchLabels: { foo: bar }
  downsync:
  - apiGroup: "example.com"
    resources: [ widgets ]
    namespaceSelectors:
    - matchLabels: { foo: baz }
    objectNames: [ "*" ]
    immobileParts:
    - [ spec, oneWeirdField ]
```

MikeSpreitzer:
@yana1205

clubanderson:
This PR is interesting. It may come in handy at some point. I can't think of a use for it right now because anything that I don't want to sync downstream is also something I can store in Git. I can't store a secret's sensitive information in git, therefore I just store the non sensitive parts and sync the secret without sensitive information. Once synced, I would ask kubestellar not to overwrite it so that that an admin of the target can patch the secret with the information required. I don't think I have a use for this PR.

ezrasilvera:
@MikeSpreitzer (cc: @nirrozenbaum @pdettori ) Are we still going to use this with the pluggable transport ?  
I think we should re-visit the source requirement (in https://github.com/kubestellar/kubestellar/issues/1112) and make sure we still think it is a requirement we should support and what is the priority.

MikeSpreitzer:
The switch to pluggable transport does not remove the API requirement. The plugged in transports will need to be able to support this functionality.

nirrozenbaum:
@MikeSpreitzer I'd be happy to understand what is the use case of this requirement.
we're currently in a transition to ks-0.20 and we don't want to add "nice to have" features.
IMO if there is no requirement for this feature for any user, we should postpone this topic to next phases and discuss it later if needed or not.

MikeSpreitzer:
What I am trying to say is that the design here for representing the "what" decision is still relevant. So a _fragment_ of this PR is still relevant.

PR #1116, opened 2023-10-12
✨ a new user-based quickstart

clubanderson:
## Summary
This quickstart is targeted to the average kubernetes user/admin/operator that has an understanding of basic kubectl, helm, and brew commands to download, install cloud-native projects and knows how to interact with kubectl for the purpose of applying, editing/updating, deleting, and reading kube objects

## Related issue(s)

Fixes #


docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,80 @@
...
+brew
+kubectl
+3 kind clusters 
+   1 KubeStellar core cluster
+   2 KubeStellar edge clusters (edge-cluster1, edge-cluster2)
+   
+## 1. Install KubeStellar
+
+helm install kubestellar --set EXTERNAL_HOSTNAME="my-long-app-name.aregion.some.cloud.com" --set EXTERNAL_PORT=1024
+
```
pdettori, line 24: Is there any hint on what external hostname I could use if installing on kind (as the guide says). Can I use anything I like ?

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,80 @@
...
+-or-
+
+helm install kubestellar --set clusterType=OpenShift
+
+## 2. Install KubeStellar's Kubectl plugins
+
+brew install kubestellar-kubectl v0.X.0
+
```
pdettori, line 32: Shouldn't I have first something like `brew tap kubestellar/kubestellar https://github.com/kubestellar/kubestellar`

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,80 @@
...
+## 3. Access KubeStellar
+
+kubectl get secrets kubestellar -o 'go-template={{index .data "external.kubeconfig"}}' | base64 --decode > admin.kubeconfig
+
```
pdettori, line 36: Isn't the namespace missing ? e.g. `get secrets -n kubestellar ...` - also perhaps you may simplify into

```
kubectl get secrets kubestellar -n kubestellar -o jsonpath='{.data.external\.kubeconfig}' | base64 -d > > admin.kubeconfig
```

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,80 @@
...
+## 4. Install KubeStellar Syncers on your Edge Clusters
+
+KUBECONFIG=admin.config kubectl kubestellar prep-for-cluster --imw root:imw1 edge-cluster1 env=edge-cluster1
+KUBECONFIG=amin.config kubectl kubestellar prep-for-cluster --imw root:imw1 edge-cluster2 env=edge-cluster2
+
+change your kubeconfig to point at edge-cluster1 and edge-cluster2 and apply the files that prep-for-cluster prepared for you
+
+KUBECONFIG-edge-cluster1.config kubectl apply -f edge-cluster1
+KUBECONFIG-edge-cluster2.config kubectl apply -f edge-cluster1
+
```
pdettori, line 46: If all clusters are kind clusters most likely the user at this point does not have a different kubeconfig file for each kind cluster, unless they are told so in the guide. By default they will just have 3 contexts in their Kubeconfig, so they would use a command like 
```
KUBECONFIG=~/.kube/config kubectl --context kind-cluster1 apply -f edge-cluster1.yaml
```

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,80 @@
...
+{%
+   include-markdown "quickstart-subs/quickstart-2-apache-example-deployment-d-create-and-deploy-apache-into-clusters.md"
+   start="<!--quickstart-2-apache-example-deployment-d-create-and-deploy-apache-into-clusters-start-->"
+   end="<!--quickstart-2-apache-example-deployment-d-create-and-deploy-apache-into-clusters-end-->"
+%}
+
+## 6. Check the status of your Apache Server on edge-cluster1 and edge-cluster2
+
+KUBECONFIG=admin.config kubectl get ...
+
```
pdettori, line 58: Same comment as above, use context

pdettori (commented):
It looks like a good start, we need also to ensure that we have all in place matching this user guide. We should also have another guide for users installing on OpenShift clusters.

docs/content/Getting-Started/user-quickstart.md
effi-ofer: That's a good point and the answer is any valid dns accessible machine name.  In my case I use EXTERNAL_HOSTNAME=localhost.

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,80 @@
...
+## 0. Pre-reqs
+
+helm
+brew
+kubectl
+3 kind clusters 
+   1 KubeStellar core cluster
+   2 KubeStellar edge clusters (edge-cluster1, edge-cluster2)
+
```
effi-ofer, line 20: Maybe this is just a problem with the git preview, but the above does not format correctly for me.

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,80 @@
...
+[![QuickStart test]({{config.repo_url}}/actions/workflows/docs-ecutable-qs.yml/badge.svg?branch={{config.ks_branch}})]({{config.repo_url}}/actions/workflows/docs-ecutable-qs.yml)&nbsp;&nbsp;&nbsp;
+
+
+!!! tip "Estimated time to complete this example:" 
+    ~4 minutes (after installing prerequisites)
+
+### How to deploy and use KubeStellar
```
MikeSpreitzer, line 8: This heading should be less nested than the headings for the sections under it.

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,80 @@
...
+helm
+brew
+kubectl
+3 kind clusters 
+   1 KubeStellar core cluster
+   2 KubeStellar edge clusters (edge-cluster1, edge-cluster2)
+   
+## 1. Install KubeStellar
+
+helm install kubestellar --set EXTERNAL_HOSTNAME="my-long-app-name.aregion.some.cloud.com" --set EXTERNAL_PORT=1024
```
MikeSpreitzer, line 23: Our current release archive building script packs up `bin/` _and_ `config/` and `core-helm-chart`, and `bin/kubectl-kubestellar-deploy` uses that adjacent copy of the Helm chart.

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,80 @@
...
+
+helm install kubestellar --set clusterType=OpenShift
+
+## 2. Install KubeStellar's Kubectl plugins
+
+brew install kubestellar-kubectl v0.X.0
+
+## 3. Access KubeStellar
+
+kubectl get secrets kubestellar -o 'go-template={{index .data "external.kubeconfig"}}' | base64 --decode > admin.kubeconfig
```
MikeSpreitzer, line 35: This seems a rather fiddly thing to ask users to do. Given that the previous step delivered executables intended for users to use, why not use https://docs.kubestellar.io/main/Coding%20Milestones/PoC2023q1/commands/#fetch-kubeconfig-for-external-clients ?

docs/content/Getting-Started/user-quickstart.md
clubanderson: agreed

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,106 @@
...
+
+This guide is intended to show how to (1) quickly bring up a **KubeStellar** environment using helm, (2) install plugins with brew, (3) retrieve your kubeconfig, (4) install KubeStellar Syncers on 2 edge clusters, (5) and deploy an example workload to both edge clusters.
+
+## 0. Pre-reqs
+
+- helm
+- brew
+- kubectl
+- kind clusters (3) 
+   - 1 KubeStellar core cluster (kind-core)
```
MikeSpreitzer, line 18: Everywhere else we call this the "hosting" cluster. We should be consistent.

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,106 @@
...
+   - 2 KubeStellar edge clusters (kind-edge-cluster1, kind-edge-cluster2)
+   
+## 1. Install KubeStellar
+
+```
+kubectl cluster-info --context kind-core
```
MikeSpreitzer, line 24: This command does not _set_ the current context. It just uses the given one for the `cluste-info` action. It does not modify any state that affects later commands.

This comment applies to many exhibited command sequences.

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,106 @@
...
+helm install kubestellar --set EXTERNAL_HOSTNAME="localhost" --set EXTERNAL_PORT=1024
```
MikeSpreitzer, line 25: The choice of "localhost" leads to a world of hurt and confusion when someone tries to use the kubeconfig fetched in step 3 from inside a container or VM or from another machine.  The NOTE below could be clearer. The point is not whether the user is "using DNS" in their environment. Everybody is "using" DNS in their environment. What needs to be said here is words about _must_ use a FQDN if the kubeconfig will be used from within a container or VM or other machine, how the FQDN must resolve to an IP address that the client can use to open a TCP connection that will reach the host of the hosting cluster, and yes, how this requires control over DNS records or /etc/hosts as relevant.

clubanderson:
K=11
S=19

1119

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,253 @@
...
+         ```
+    === "when is KubeStellar ready?"
+         run the following to wait for KubeStellar to be ready to take requests:
+         ```
+         echo -n 'Waiting for KubeStellar to be ready'
+         while ! kubectl exec $(kubectl get pod --selector=app=kubestellar -o jsonpath='{.items[0].metadata.name}' -n kubestellar) -n kubestellar -c init -- ls /home/kubestellar/ready &> /dev/null; do
+            sleep 10
+            echo -n "."
+         done
+         echo "KubeStellar is now ready to take requests"
```
fab7, line 114: One could add 1 or 2 newline(s) before echoing "KubeStellar is now ready to take requests".
This would avoid to echo a cluttered output message when the user is too slow to cut and paste the wait command, such as:
  --> "Waiting for KubeStellar to be readyKubeStellar is now ready to take requests"

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,253 @@
...
+         KUBECONFIG=ks-core.kubeconfig kubectl ws --context root tree
+         ```
+    === "uh oh, error"
+         Did you received the following error:
+         ```Error: Get "https://some_hostname.some_domain_name:{{config.ks_port_num}}/clusters/root/apis/tenancy.kcp.io/v1alpha1/workspaces": dial tcp: lookup some_hostname.some_domain_name on x.x.x.x: no such host``
+
+         A common error occurs if you set your port number to a pre-occupied port number and/or you set your EXTERNAL_HOSTNAME to something other than "localhost" so that you can reach your KubeStellar Core from another host, check the following:
+         
+         Check if the port specified in the ks-core kind cluster configuration and the EXTERNAL_PORT helm value are unoccupied by another application:
+
```
fab7, line 157: "..are unoccupied by..." isn't it "Check if... are occupied by another application." ?

docs/content/Getting-Started/user-quickstart.md
```diff
@@ -0,0 +1,253 @@
...
+         &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;1. is the port specified in this example occupied by another proecess?  If so, delete the 'ks-core' kind cluster and create it again using an available port for your 'hostPort' value
+
```
fab7, line 159: [typo] proecess? --> process?

clubanderson:
to test this PR:

clubanderson:
to test this PR:
1) checkout kubestellar (not a fork)
2) checkout my pr for user-QuickStart locally
3) cd to the docs directory
4) mkdocs serve

how to install mkdocs:

clubanderson:
I am happy with the state of the user-quickstart for kind.  Please review

clubanderson:
added openshift quickstart.  openshift version syncs correctly but does pods do not start due to service account error.  Need some help on this one

docs/content/common-subs/create-ks-core-kind-cluster.md
```diff
@@ -0,0 +1,36 @@
...
+    hostPort: {{ config.ks_kind_port_num }}
+    protocol: TCP
+EOF
+```
+
+Be sure to apply an ingress control with SSL passthrough to **ks-core**. This is a special requirement for Kind that allows access to the <span class="Space-Bd-BT">KUBESTELLAR</span> core running on **ks-core**.
+```shell
+kubectl apply -f https://raw.githubusercontent.com/kubestellar/kubestellar/main/example/kind-nginx-ingress-with-SSL-passthrough.yaml
+```
+**Wait about 20 seconds** and then check if the ingress control is ready on **ks-core**:
```
MikeSpreitzer, line 27: The second command is also about waiting. It would be simpler to just have that one command. Is 90 seconds not long enough to wait? If we need another 20 seconds then just add 20s to the limit on the `kubectl wait`.

MikeSpreitzer:
The English text here does not even mention kcp, and installation of kcp is part of installation of KubeStellar. This is implicitly saying that kcp is part of KubeStellar. I thought we were agreed on a different position.

docs/content/Getting-Started/user-quickstart-kind.md
```diff
@@ -0,0 +1,276 @@
...
+           end="<!--kubestellar-show-available-spaces-end-->"
+         %}
+    
+    === "uh oh, error?"
+         Did you received the following error:
+         ```Error: Get "https://some_hostname.some_domain_name:{{config.ks_kind_port_num}}/clusters/root/apis/tenancy.kcp.io/v1alpha1/workspaces": dial tcp: lookup some_hostname.some_domain_name on x.x.x.x: no such host``
+
+         A common error occurs if you set your port number to a pre-occupied port number and/or you set your EXTERNAL_HOSTNAME to something other than "localhost" so that you can reach your <span class="Space-Bd-BT">KUBESTELLAR</span> Core from another host, check the following:
+         
+         Check if the port specified in the **ks-core** Kind cluster configuration and the EXTERNAL_PORT helm value are occupied by another application:
```
MikeSpreitzer, line 104: I would say specifically that it is the _host_ port of the Kind cluster to check, because what is configured is a port mapping --- _two_ ports are mentioned in the configuration.

docs/content/Getting-Started/user-quickstart-kind.md
```diff
@@ -0,0 +1,276 @@
...
+    === "show all available <span class="Space-Bd-BT">KUBESTELLAR</span> Core Spaces"
+         {%
+           include-markdown "../common-subs/kubestellar-show-available-spaces.md"
+           start="<!--kubestellar-show-available-spaces-start-->"
+           end="<!--kubestellar-show-available-spaces-end-->"
+         %}
+    
+    === "uh oh, error?"
+         Did you received the following error:
+         ```Error: Get "https://some_hostname.some_domain_name:{{config.ks_kind_port_num}}/clusters/root/apis/tenancy.kcp.io/v1alpha1/workspaces": dial tcp: lookup some_hostname.some_domain_name on x.x.x.x: no such host``
```
MikeSpreitzer, line 100: KS core port collision will not cause DNS lookups to fail.

docs/content/Getting-Started/user-quickstart-kind.md
```diff
@@ -0,0 +1,276 @@
...
+
+         A common error occurs if you set your port number to a pre-occupied port number and/or you set your EXTERNAL_HOSTNAME to something other than "localhost" so that you can reach your <span class="Space-Bd-BT">KUBESTELLAR</span> Core from another host, check the following:
```
MikeSpreitzer, line 102: The concern is not only "from another host" but also from inside a container on the same host.

docs/content/Getting-Started/user-quickstart-kind.md
```diff
@@ -0,0 +1,276 @@
...
+         
+         Check if the port specified in the **ks-core** Kind cluster configuration and the EXTERNAL_PORT helm value are occupied by another application:
+
+         &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;1. is the port specified in this example occupied by another process?  If so, delete the **ks-core** Kind cluster and create it again using an available port for your 'hostPort' value
+
+         &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2. if you change the port for your **ks-core** 'hostPort', remember to also use that port as the helm 'EXTERNAL_PORT' value
+
+         Check that your EXTERNAL_HOSTNAME helm value is reachable via DNS:
+
+         &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;1. use 'nslookup <value of EXTERNAL_HOSTNAME>' to make sure there is a valid IP address associated with the hostname you have specified
```
MikeSpreitzer, line 112: `nslookup` does not find stuff in `/etc/hosts`.
`ping` finds stuff in `/etc/hosts`. People often configure their machines to ignore ping requests, but that does not stop the address resolution step from getting the intended result.

```
bash-3.2$ nslookup yep.yep
Server:		192.168.58.47
Address:	192.168.58.47#53

** server can't find yep.yep: NXDOMAIN

bash-3.2$ ping yep.yep
PING yep.yep (192.168.58.125): 56 data bytes
Request timeout for icmp_seq 0
Request timeout for icmp_seq 1
^C
--- yep.yep ping statistics ---
3 packets transmitted, 0 packets received, 100.0% packet loss
```

docs/content/Getting-Started/user-quickstart-kind.md
```diff
@@ -0,0 +1,276 @@
...
+
+         &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2. make sure your EXTERNAL_HOSTNAME and associated ip address are listed in your /etc/hosts file.
+
+         &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;3. make sure the IP address is associated with the system where you have deployed the **ks-core** Kind cluster
```
MikeSpreitzer, line 116: We should also say that the IP address is one that can be used to open TCP connections from the syncers and the shell(s) where the user will be issuing commands. While it is true that 127.0.0.1 is an IP address that is "associated with the system where you have deployed the **ks-core** Kind cluster", it is not an IP address that EXTERNAL_HOSTNAME should resolve to.

docs/content/common-subs/create-ks-core-kind-cluster.md
```diff
@@ -0,0 +1,36 @@
+<!--create-ks-core-kind-cluster-start-->
+create the ks-core kind cluster
+```shell
+export KUBECONFIG=~/.kube/config
+kind create cluster --name ks-core --config - <<EOF
```
MikeSpreitzer, line 5: There is no need here to diverge from the pattern that is mainly used in this document, `KUBECONFIG=~/.kube/confg command`.

docs/content/Getting-Started/user-quickstart-kind.md
```diff
@@ -0,0 +1,276 @@
+---
+short_name: user-quickstart-kind
+manifest_name: 'docs/content/Getting-Started/user-quickstart-kind.md'
+---
+[![User QuickStart Kind test]({{config.repo_url}}/actions/workflows/docs-ecutable-user-quickstart-kind.yml/badge.svg?branch={{config.ks_branch}})]({{config.repo_url}}/actions/workflows/docs-ecutable-user-quickstart-kind.yml)&nbsp;&nbsp;&nbsp;
+
+<!-- 
+!!! tip "Estimated time to complete this example:" 
+    ~4 minutes (after installing prerequisites) -->
```
namasl, line 9: 4 minutes isn't realistic unless someone is only going copy/paste without errors, and not reading any of the text along the way. 20 - 30 minutes is closer to a good pace.

docs/content/common-subs/install-helm-kind.md
```diff
@@ -0,0 +1,14 @@
+<!--install-helm-kind-start-->
+```shell hl_lines="6"
+KUBECONFIG=~/.kube/config kubectl config use-context ks-core  
+kubectl create namespace kubestellar
```
MikeSpreitzer, line 4: I see that this document generally follows the pattern of `KUBECONFIG=~/.kube/config command`, and I think that is a good way to make it clear to the user that the choice of a kubeconfig file is a relevant thing to be paying attention to. It is a way of solving the problem of making context (broadly construed) visible without replacing text (very convenient and accessible) with pictures (horrible for copy-and-paste by will-sighted people and even worse for consumption by others) of sessions in shells graphically show context.

I assume that the lack of the KUBECONFIG prefix on the second command here is an oversight. As this currently stands, the second command only works if elsewhere the user has done `export KUBECONFIG=~/.kube/config` or equivalent. If that assumption were good enough for the second command here then it would also be good enough for the first.

docs/content/common-subs/install-helm-kind.md
```diff
@@ -0,0 +1,14 @@
+<!--install-helm-kind-start-->
+```shell hl_lines="6"
+KUBECONFIG=~/.kube/config kubectl config use-context ks-core  
+kubectl create namespace kubestellar  
+
+helm repo add kubestellar https://helm.kubestellar.io
+helm repo update
+helm install kubestellar/kubestellar-core \
```
MikeSpreitzer, line 8: The `helm` command also consumes the choice of kubeconfig file, in the regular ways: by a setting of the KUBECONFIG envar or with the `--kubeconfig` command line flag.

docs/content/common-subs/install-helm-kind.md
```diff
@@ -0,0 +1,14 @@
...
+  --set EXTERNAL_HOSTNAME="kubestellar.core" \
+  --set EXTERNAL_PORT={{ config.ks_kind_port_num }} \
+  --namespace kubestellar \
```
MikeSpreitzer, line 11: Don't you also need `--create-namespace` here?

docs/content/common-subs/check-kubestellar-helm-deployment-running.md
```diff
@@ -0,0 +1,11 @@
+<!--check-kubestellar-helm-deployment-running-start-->
+```shell
+echo -n 'Waiting for KubeStellar to be ready'
+while ! kubectl exec $(kubectl get pod --selector=app=kubestellar -o jsonpath='{.items[0].metadata.name}' -n kubestellar) -n kubestellar -c init -- ls /home/kubestellar/ready &> /dev/null; do
```
MikeSpreitzer, line 4: The two `kubectl` commands here are sensitive to the setting of `$KUBECONFIG`. Since this document generally uses an explicit setting of that envar on each command line, that could be done here. One way would be very direct:

```
while ! KUBECONFIG=~/.kube/config kubectl exec $(KUBECONFIG=~/.kube/config kubectl get pod --selector=app=kubestellar -o jsonpath='{.items[0].metadata.name}' -n kubestellar) -n kubestellar -c init -- ls /home/kubestellar/ready &> /dev/null; do
```

Another approach would be to break out the discovery of the pod name into a command that is done once at the start; it does not have to be rediscovered every time around the loop. But we _do_ have to wait for the pod to exist, which is covered by waiting for the pod to be ready.

```
KUBECONFIG=~/.kube/config kubectl wait -n kubestellar -for condition=ready pod --selector app=kubestellar
KS_CORE_POD=$(KUBECONFIG=~/.kube/config kubectl get pod --selector=app=kubestellar -o jsonpath='{.items[0].metadata.name}' -n kubestellar)
while ! KUBECONFIG=~/.kube/config kubectl exec $KS_CORE_POD -n kubestellar -c init -- ls /home/kubestellar/ready &> /dev/null; do
...
done
```

docs/content/Getting-Started/user-quickstart-openshift.md
```diff
@@ -0,0 +1,243 @@
...
+           end="<!--brew-install-end-->"
+         %}
+    === "remove"
+         {%
+           include-markdown "../common-subs/brew-remove.md"
+           start="<!--brew-remove-start-->"
+           end="<!--brew-remove-end-->"
+         %}
+
+#### 2. View your <span class="Space-Bd-BT">KUBESTELLAR</span> Core Space environment
```
MikeSpreitzer, line 178: This section also does something much more impactful that listing some workspaces: it fetches the external kubeconfig and puts in a particular file. This should be called out in the section title as well as getting explicit mention in the English text.

docs/content/common-subs/kubestellar-apply-apache.md
```diff
@@ -0,0 +1,111 @@
...
+metadata:
+  name: my-first-edge-placement
+spec:
+  locationSelectors:
+  - matchLabels: {"location-group":"edge"}
+  downsync:
+  - apiGroup: ""
+    resources: [ configmaps ]
+    namespaceSelectors:
+    - matchLabels: {"common":"sure-is"}
```
MikeSpreitzer, line 24: With release 0.8 and 0.9 you can replace this namespace label selection with simply `namespaces: [ my-namespace ]`.

docs/content/Getting-Started/user-quickstart-openshift.md
```diff
@@ -0,0 +1,243 @@
...
+    === "open a bug report"
+        Stuck? [open a bug report and we can help you out](https://github.com/kubestellar/kubestellar/issues/new?assignees=&labels=kind%2Fbug&projects=&template=bug_report.yaml&title=bug%3A+)
+
+
+#### 4. Install <span class="Space-Bd-BT">KUBESTELLAR</span> Syncers on your Edge Clusters
+!!! tip ""
+    === "Prep and apply"
+        prepare <span class="Space-Bd-BT">KUBESTELLAR</span> Syncers, with `kubestellar prep-for-cluster`, for **ks-edge-cluster1** and **ks-edge-cluster2** and then apply the files that `kubestellar prep-for-cluster`` prepared for you
+
+        **important:** make sure you created Kind clusters for **ks-edge-cluster1** and **ks-edge-cluster2** from the pre-req step above before proceeding [how-to-deploy-and-use-kubestellar](#how-to-deploy-and-use-kubestellar)
```
MikeSpreitzer, line 107: Overlooked edit here, the text refers to "Kind" clusters.

MikeSpreitzer:
Chris Milite showed us how to downsync the effect of an `oc adm policy add-scc-to-user`. The key object is a RoleBinding. You can use one whose subject is the `default` ServiceAccount in your workload namespace (`commonstuff`). Chris also created a Role object, but it would be simpler to bind to the pre-existing ClusterRole. The Role that Chris made is equivalent to the ClusterRole named `system:openshift:scc:anyuid`. I have not seen the error message from the OpenShift WECs, so I can not promise that this is the right scc to solve all the problems there.

docs/content/Getting-Started/user-quickstart-kind.md
```diff
@@ -0,0 +1,276 @@
...
+#### 2. View your <span class="Space-Bd-BT">KUBESTELLAR</span> Core Space environment
+
+!!! tip ""
+    === "show all available <span class="Space-Bd-BT">KUBESTELLAR</span> Core Spaces"
+         ```
+         KUBECONFIG=~/.kube/config kubectl --context ks-core get secrets kubestellar \
+           -o jsonpath='{.data.external\.kubeconfig}' \
+           -n kubestellar | base64 -d > ks-core.kubeconfig
+
+         KUBECONFIG=ks-core.kubeconfig kubectl ws --context root tree
```
MikeSpreitzer, line 215: Are you sure that `ws tree` is sensitive to the `--context` setting on the command line? It does not look that way to me:

```
mspreitz@mjs13 kubestellar % KUBECONFIG=ksc.kubeconfig kubectl ws .
Current workspace is "root:imw1".

mspreitz@mjs13 kubestellar % KUBECONFIG=ksc.kubeconfig kubectl ws --context root tree
.
└── imw1
```

docs/content/common-subs/kubestellar-apply-apache.md
```diff
@@ -0,0 +1,111 @@
+<!--kubestellar-apply-apache-start-->
+<span class="Space-Bd-BT">KUBESTELLAR</span>'s helm chart automatically creates a Workload Management
+Workspace (WMW) for you to store kubernetes workload descriptions and <span class="Space-Bd-BT">KUBESTELLAR</span> control objects in. The automatically created WMW is at `root:wmw1`.
+
+Create an EdgePlacement control object to direct where your workload runs using the 'location-group=edge' label selector. This label selector's value ensures your workload is directed to both clusters, as they were labeled with 'location-group=edge' when you issued the 'kubestellar prep-for-cluster' command above.
+
+In the `root:wmw1` workspace create the following `EdgePlacement` object: 
+```shell linenums="1" hl_lines="10 11 16 21"
```
MikeSpreitzer, line 8: The display of line numbers here is inconsistent with almost all of the rest of the document.

You do not need to _display_ line numbers in order to be able to _highlight_ lines; you can just remove the `linenums="1"` part.

Since the simpler way of selecting namespaces is available, I suppose you will want to use it, removing the need to highlight the (would not exist) Namespace label selectors.

docs/content/common-subs/kubestellar-test-apache.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+Finally, let's check that the workload is working in both clusters:
+For **ks-edge-cluster1**:
+```shell
+export KUBECONFIG=~/.kube/config
+while [[ $(kubectl --context ks-edge-cluster1 get pod \
+  -l "app=common" -n my-namespace \
+  -o jsonpath='{.items[0].status.phase}') != "Running" ]]; \
+  do sleep 5; done;curl http://localhost:8094
```
MikeSpreitzer, line 31: Editing oversight here, I assume you did not want to tack the `curl` onto the end of the `while`.

docs/content/Getting-Started/user-quickstart-kind.md
```diff
@@ -0,0 +1,276 @@
...
+          start="<!--create-ks-edge-cluster2-kind-cluster-start-->"
+          end="<!--create-ks-edge-cluster2-kind-cluster-end-->"
+        %}
+
+        {%
+          include-markdown "../common-subs/create-files-and-contexts-for-kind-clusters.md"
+          start="<!--create-files-and-contexts-for-kind-clusters-start-->"
+          end="<!--create-files-and-contexts-for-kind-clusters-end-->"
+        %}
+    === "uh oh, error?"
```
MikeSpreitzer, line 61: Very nice, I like adding a tab for troubleshooting.

clubanderson:
changes made to suit first round of reviews.  Please test review/test again.  I am running automation, in my branch, to make sure this passes docs-ecutable.

clubanderson:
@namasl saw your comment on timing and adjusted to 20 min for both docs.  I could not find your comment in this conversation, but it did show up in my email - don't know why it is not inline above.

clubanderson:
docs-ecutable passed for kind quickstart.  There is no docs-ecutable for openshift at this time - but there might be in the future.  Easy to test, not so easy to login to openshift - not to mention storing that info somewhere.  An issue for another time.

docs/content/common-subs/create-ks-core-kind-cluster.md
```diff
@@ -0,0 +1,36 @@
...
+- role: control-plane
+  kubeadmConfigPatches:
+  - |
+    kind: InitConfiguration
+    nodeRegistration:
+      kubeletExtraArgs:
+        node-labels: "ingress-ready=true"
+  extraPortMappings:
+  - containerPort: 443
+    hostPort: {{ config.ks_kind_port_num }}
```
namasl, line 17: Do you mean to specify a port number here? If I run this verbatim I get:
`ERROR: failed to create cluster: unable to decode config: yaml: unmarshal errors:
  line 13: cannot unmarshal !!map into int32`

clubanderson:
@namasl to test this PR you must:
1) checkout this PR locally on your system
2) install mkdocs
3) cd to /docs and run 'mkdocs serve'
4) open browser to http://127.0.0.1:8000/Getting-Started/user-quickstart-openshift/

you can't run any of the example term session content by just copying from source.

docs/content/common-subs/install-helm-kind.md
```diff
@@ -0,0 +1,14 @@
+<!--install-helm-kind-start-->
+```shell hl_lines="7"
+KUBECONFIG=~/.kube/config kubectl config use-context ks-core  
+KUBECONFIG=~/.kube/config kubectl create namespace kubestellar  
+
+helm repo add kubestellar https://helm.kubestellar.io
+helm repo update
+KUBECONFIG=~/.kube/config helm install kubestellar/kubestellar-core \
+  --set EXTERNAL_HOSTNAME="kubestellar.core" \
+  --set EXTERNAL_PORT={{ config.ks_kind_port_num }} \
```
namasl, line 10: Similar comment as for `docs/content/common-subs/create-ks-core-kind-cluster.md`, as the Quickstart stands the user needs to manually enter a port here (unless I'm overlooking something).

docs/content/common-subs/kubestellar-test-apache.md
```diff
@@ -0,0 +1,62 @@
...
+
+Finally, let's check that the workload is working in both clusters:
+For **ks-edge-cluster1**:
+```shell
+while [[ $(KUBECONFIG=~/.kube/config kubectl --context ks-edge-cluster1 get pod \
+  -l "app=common" -n my-namespace -o jsonpath='{.items[0].status.phase}') != "Running" ]]; do 
+    sleep 5; 
+  done;
+curl http://localhost:8094
+```
```
namasl, line 32: This is the point at which my attempt to execute the Quickstart fails. I've gone through twice to make sure I didn't skip any steps (still possible), but executing `KUBECONFIG=~/.kube/config kubectl --context ks-edge-cluster1 get pod \
  -l "app=common" -n my-namespace -o jsonpath='{.items[0].status.phase}'` gives the error

```
error: error executing jsonpath "{.items[0].status.phase}": Error executing template: array index out of bounds: index 0, length 0. Printing more information for debugging the template:
	template was:
		{.items[0].status.phase}
	object given to jsonpath engine was:
		map[string]interface {}{"apiVersion":"v1", "items":[]interface {}{}, "kind":"List", "metadata":map[string]interface {}{"resourceVersion":""}}
```

ks-edge-cluster1 log:
```
I1023 15:24:24.338879       1 reflector.go:307] Listing and watching *v2alpha1.EdgeSyncConfig from k8s.io/client-go@v0.0.0-20230210192259-aaa28aa88b2d/tools/cache/reflector.go:215
W1023 15:24:24.399611       1 reflector.go:376] k8s.io/client-go@v0.0.0-20230210192259-aaa28aa88b2d/tools/cache/reflector.go:215: failed to list *v2alpha1.EdgeSyncConfig: Get "https://kubestellar.core:4000/apis/edge.kubestellar.io/v2alpha1/edgesyncconfigs?limit=500&resourceVersion=0": dial tcp: lookup kubestellar.core on 10.96.0.10:53: no such host
E1023 15:24:24.399674       1 reflector.go:138] k8s.io/client-go@v0.0.0-20230210192259-aaa28aa88b2d/tools/cache/reflector.go:215: Failed to watch *v2alpha1.EdgeSyncConfig: failed to list *v2alpha1.EdgeSyncConfig: Get "https://kubestellar.core:4000/apis/edge.kubestellar.io/v2alpha1/edgesyncconfigs?limit=500&resourceVersion=0": dial tcp: lookup kubestellar.core on 10.96.0.10:53: no such host
```

ks-edge-cluster2 log:

```
I1023 15:25:53.899697       1 reflector.go:307] Listing and watching *v2alpha1.EdgeSyncConfig from k8s.io/client-go@v0.0.0-20230210192259-aaa28aa88b2d/tools/cache/reflector.go:215
W1023 15:25:53.956613       1 reflector.go:376] k8s.io/client-go@v0.0.0-20230210192259-aaa28aa88b2d/tools/cache/reflector.go:215: failed to list *v2alpha1.EdgeSyncConfig: Get "https://kubestellar.core:4000/apis/edge.kubestellar.io/v2alpha1/edgesyncconfigs?limit=500&resourceVersion=0": dial tcp: lookup kubestellar.core on 10.96.0.10:53: no such host
E1023 15:25:53.956738       1 reflector.go:138] k8s.io/client-go@v0.0.0-20230210192259-aaa28aa88b2d/tools/cache/reflector.go:215: Failed to watch *v2alpha1.EdgeSyncConfig: failed to list *v2alpha1.EdgeSyncConfig: Get "https://kubestellar.core:4000/apis/edge.kubestellar.io/v2alpha1/edgesyncconfigs?limit=500&resourceVersion=0": dial tcp: lookup kubestellar.core on 10.96.0.10:53: no such host
```

My `/etc/hosts`:

```
127.0.0.1	localhost
127.0.1.1	debian
192.168.122.144 kubestellar.core

# The following lines are desirable for IPv6 capable hosts
::1     localhost ip6-localhost ip6-loopback
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters

```

docs/content/common-subs/check-kubestellar-helm-deployment-running.md
```diff
@@ -0,0 +1,13 @@
...
+```shell
+echo -n 'Waiting for KubeStellar to be ready'
+while ! KUBECONFIG=~/.kube/config kubectl exec $(KUBECONFIG=~/.kube/config kubectl get pod \
+   --selector=app=kubestellar -o jsonpath='{.items[0].metadata.name}' -n kubestellar) \
+   -n kubestellar -c init -- ls /home/kubestellar/ready &> /dev/null; do
+   sleep 10
+   echo -n "."
+done
+
+echo "\n\nKubeStellar is now ready to take requests"
```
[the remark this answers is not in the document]
MikeSpreitzer: Further:

```
mspreitz@mjs13 ~ % echo -n "a\nb"
a
b%                                                                              mspreitz@mjs13 ~ % bash

The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.
bash-3.2$ echo -n "a\nb"
a\nbbash-3.2$ 
```

docs/content/common-subs/check-kubestellar-helm-deployment-running.md
MikeSpreitzer: Actually, I that`kubectl wait` for the pod to be ready might NOT do what we want if invoked before the pod exists. I had that fail immediately with a complaint about "no resource found" when that case happened to me recently.

MikeSpreitzer:
In the version I reviewed recently, I saw kubeconfig contexts being _renamed_ in the case of using `kind` and _copied_ in the case of using OpenShift. The renaming is much easier on the user. Why not do it in the OpenShift case too? Perhaps there was some thinking about what happens when the OpenShift credentials time out after a day? In either approach (copy or rename), what the user is doing stops working and they need to re-do the `oc login`.

clubanderson:
the rename is for kind because this is a new environment and is going to disappear at some point.  The thinking for openshift is that we do not want to override the naming in case the clusters are being used for other things.  I do not want to be the person they blame if they are using the cluster for other things and the context is hard coded somewhere.

KPRoche:
Hmmm... something odd. my clone didn't grab the right version somehow
Going to try again.

KPRoche:
OK, that looks like actual instructions (as opposed to 404). Can actually test it out now.

docs/content/Getting-Started/user-quickstart-kind.md
```diff
@@ -0,0 +1,250 @@
...
+
+        **important:** For this quickstart you will need to know how to use kubernetes' kubeconfig *context* to access multiple clusters.  You can learn more about kubeconfig context [here](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/)
+
+!!! tip ""
+    === "Pre-reqs"
+        + [__kubectl__](https://kubernetes.io/docs/tasks/tools/) (version range expected: 1.24-1.26)
+
+        + [__helm__](https://helm.sh/docs/intro/install/) - to deploy the <span class="Space-Bd-BT">KUBESTELLAR</span>-core helm chart
+        
+        + [__brew__](https://helm.sh/docs/intro/install/) - to install the <span class="Space-Bd-BT">KUBESTELLAR</span> user commands and kubectl plugins
```
KPRoche, line 32: This is pointing at the helm install instructions again.
Don't you want to point at https://docs.brew.sh/Installation ?

clubanderson:
> OK, that looks like actual instructions (as opposed to 404). Can actually test it out now. For the record, what this still-sometimes-git-newbie has trouble with is changing branches.
> 
> 1. checkout this PR locally on your system. Make sure to check out the right branch!
>    
>    * e.g. _git clone -b user-quickstart https://github.com/clubanderson/kubestellar.git_
> 2. install mkdocs
> 3. cd to /docs and run 'mkdocs serve'
> 4. open browser to http://127.0.0.1:8000/Getting-Started/user-quickstart-openshift/

@KPRoche just added a 'testing-doc-prs.md' to the operation guide based on your steps here.  Thank you

KPRoche:
Regarding installing brew -- it *really* does not want you to install it as root.

KPRoche:
ah, brew won't install on arm, either.
```[kproche@rpi4-centos-edge ~]$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
==> Checking for `sudo` access (which may request your password)...
[sudo] password for kproche:
Homebrew on Linux is not supported on ARM processors.
  https://docs.brew.sh/Homebrew-on-Linux#arm-unsupported
[kproche@rpi4-centos-edge ~]$```

KPRoche:
I *really* like the idea of this quickstart.
Having spent an hour or so in the "install anywhere from tarball" labyrinth for brew, without yet succeeding, if we really want a "user-quickstart" we need an alternative for systems where brew doesn't work.

Even a mention in the prereqs (maybe in the uh-oh, Errors? tab) and a link somewhere for a manual install of the kubestellar user commands and kubectl plugins (@clubanderson do you have a list of steps that were the basis of creating the brew formula?)  would be good.

KPRoche:
I just ran into an "interesting" glitch. I was viewing the document on a browser on a different machine than I wanted to launch my kind clusters on. The cut-and-paste code snippet reports the kind port for the machine the document is being *viewed* on.

KPRoche:
> I just ran into an "interesting" glitch. I was viewing the document on a browser on a different machine than I wanted to launch my kind clusters on. The cut-and-paste code snippet reports the kind port for the machine the document is being _viewed_ on.

OK, @MikeSpreitzer showed me where that is set in the mkdocs.yaml so it is NOT dynamic by machine, but is empty if you are looking at the source rather than viewing it fully rendered.

MikeSpreitzer:
In the website navigation panel on the left this PR currently includes the following three items in the "Getting Started" section:
- Quickstart
- Quickstart - OpenShift
- Detailed Quickstart

I think it would be better if the first one of those was also qualified, to make its relationship with the others clearer.  How about calling it "Quickstart - kind"?

MikeSpreitzer:
As I view the `kind` quickstart right  now (commit afab3875054b0e2bb60fe950d01b25602ecb2ceb), I see a "Special notes for Debian users" in the Pre-reqs section but the user who has a problem in this regard will not get an error message until step 3. The "uh oh, error?" tab for Step 3 does not reference the "Special notes for Debian users".

Also, the choices for Linux are not "Debian" vs. "not Debian". The [Linux family tree](https://www.google.com/search?q=linux+family+tree) is quite large.  And "Ubuntu", which we know does NOT need the "Special notes for Debian users", IS in the Debian branch of that tree. So I think we need to identify who needs that note not in terms of "Debian" but rather in terms of the critical functionality; perhaps something like "host OS does not run a DNS server" or "host OS does not run a DNS server that reads /etc/hosts".

KPRoche:
I found another "obvious in hindsight" gotcha. If you create the ks-edge clusters on a different system than ks-core, you have to rename the contexts *_there_* for the apply syncer manifest commands to work.

clubanderson:
> In the website navigation panel on the left this PR currently includes the following three items in the "Getting Started" section:
> 
> * Quickstart
> * Quickstart - OpenShift
> * Detailed Quickstart
> 
> I think it would be better if the first one of those was also qualified, to make its relationship with the others clearer. How about calling it "Quickstart - kind"?

I updated to the "Quickstart with Kind" and "Quickstart with OpenShift"

PR #1122, opened 2023-10-12
✨ Expand API to allow namespace selection by name

### LINKED ISSUE #1084: feature: ability to select namespaces by names rather than label selector
### Feature Description

In some use cases some workload objects --- and `Namespace` objects _are_ objects --- are authored by a third party (not the KS user). Requiring the KS user to add labels to these objects is problematic (not necessarily insoluble, but more trouble than we want to require for users). One example of this is in https://github.com/kubestellar/kubestellar/blob/a60700c53c83270394c804bc4e7ea6dd9d6be5e9/pkg/apis/edge/v1alpha1/edge-placement.go#L107-L111 , which says that the _only_ way that users can filter on namespace is by using a label selector on the namespace's labels. There should be a way to filter based on the namespace's name.

### Proposed Solution

Add a field that holds a list of strings, each of which is a dead-simple pattern (i.e., `*` can match any substring, all other characters are literals) for matching namespace names. A namespace matches the whole DownsyncObjectTest if it matches either the new field or the old list of label selectors.

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR expands the "what predicate" to allow testing a namespace's name rather than its labels.

## Related issue(s)

Fixes #1084

### DIFF
--- a/config/crds/edge.kubestellar.io_edgeplacements.yaml
+++ b/config/crds/edge.kubestellar.io_edgeplacements.yaml
@@ -63,9 +63,10 @@ spec:
                 items:
                   description: 'DownsyncObjectTest is a set of criteria that characterize
                     matching objects. An object matches if: - the `apiGroup` criterion
-                    is satisfied; - the `resources` criterion is satisfied; - the
-                    `namespaceSelectors` criterion is satisfied; and - EITHER the
-                    `objectNames` criterion or the `labelSelectors` criterion matches.'
+                    is satisfied; - the `resources` criterion is satisfied; - EITHER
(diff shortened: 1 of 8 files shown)


clubanderson (commented):
/lgtm

MikeSpreitzer:
My plan is to include the implementation in this PR.

MikeSpreitzer:
The force-push to df4839e8f adds some missing API annotations marking lists as optional.

MikeSpreitzer:
I have successfully run through example1, deploying the core to a kind cluster.

clubanderson (commented):
/lgtm

PR #1140, opened 2023-10-17
✨ Brew installer for kubestellar user commands and kubectl plugins

### LINKED ISSUE #908: feature: use Brew to install KubeStellar plugins

clubanderson:
## Related issue(s)

Fixes #908


MikeSpreitzer:
The brew install failed for me.  Here is the typescript.

[brew-install-ks-fail1.log](https://github.com/kubestellar/kubestellar/files/12924137/brew-install-ks-fail1.log)

MikeSpreitzer:
The command in the README is too specific about the file location. Not everybody has their local copy of this repo at `~/kubestellar`.

MikeSpreitzer:
After correcting for the location of the formula, I was able to successfully do the `brew install`.  The typescript is attached here.

[brew-install-kubestellar-231017a.log](https://github.com/kubestellar/kubestellar/files/12924197/brew-install-kubestellar-231017a.log)

I was surprised that it installed the brew version of kubectl.  I already had a different version of kubectl installed, deliberately chosen to not be very different from the version of Kubernetes that kcp v0.11 pins KubeStellar to. Is there a way to make this formula not install kubectl if it is already present in my `$PATH`?

MikeSpreitzer:
I see that this brew install currently does not install the kcp plugins. Remember that many of the kubectl-kubestellar-xxx scripts invoke `kubectl ws`.

I see that this brew install does not currently install helm.  `kubectl-kubestellar-deploy` uses helm. It also uses a local copy of the helm chart, found bin/../core-helm-chart/ ; the brew install could create /opt/homebrew/Cellar/kubestellar_cli/v0.8.0/core-helm-chart/ along with /opt/homebrew/Cellar/kubestellar_cli/v0.8.0/bin/ , right?

clubanderson:
@MikeSpreitzer I have updated the formula for KS to include check for non brew kubectl, I also added the other files from the tar to install alongside bin (includes core helm etc.), I added a quick formula for kcp, and I updated readme.

ready for another round of review

fab7:
Well done! The installer worked for me a Ubuntu 22.04.3 LTS (Jammy Jellyfish).
I was wondering if we could add 1-2 commands at the very end  to confirm the installation (e.g. `which kubectl` or `kubectl version`).
Here is my typescript: 
[pr1140-fab.log](https://github.com/kubestellar/kubestellar/files/12930183/pr1140-fab.log)

brew/README.md
```diff
@@ -0,0 +1,19 @@
...
+    HOMEBREW_NO_INSTALL_FROM_API=1 brew install -s --formula kubestellar_cli.rb
+
+to remove: 
+    brew remove kubestellar_cli
+    brew remove kcp_cli
+
+
+after PR is accepted:
+
+brew tap kubestellar/kubestellar https://github.com/kubestellar/kubestellar/brew
```
MikeSpreitzer, line 17: What is going to make this happen?  When I put `https://github.com/kubestellar/kubestellar/bootstrap` in the address bar of my browser, I get a 404.

MikeSpreitzer:
The technique that `kubectl-kubestellar-deploy` uses to locate the local copy of the helm chart does not work when that executable is invoked from a symlinked location. This is not only a problem for the brew install, it is really a general issue; the brew install forces the issue.

```
mspreitz@mjs13 kubestellar % which kubectl-kubestellar-deploy
/opt/homebrew/bin/kubectl-kubestellar-deploy

mspreitz@mjs13 kubestellar % ls -l /opt/homebrew/bin/kubectl-kubestellar-deploy
lrwxr-xr-x  1 mspreitz  admin  63 Oct 17 17:22 /opt/homebrew/bin/kubectl-kubestellar-deploy -> ../Cellar/kubestellar_cli/v0.8.0/bin/kubectl-kubestellar-deploy

mspreitz@mjs13 kubestellar % ls /opt/homebrew/Cellar/kubestellar_cli/v0.8.0
INSTALL_RECEIPT.json	README.md		config
LICENSE			bin			core-helm-chart

mspreitz@mjs13 kubestellar % kubectl kubestellar deploy -X -n kubestellar --external-endpoint yep.yep:6443 --create-namespace
+ shift
+ ((  5 > 0  ))
+ case "$1" in
+ helm_flags[${#helm_flags[*]}]=-n
+ ((  5 > 1  ))
+ helm_flags[${#helm_flags[*]}]=kubestellar
+ shift
+ shift
+ ((  3 > 0  ))
+ case "$1" in
+ ((  3 >1  ))
+ external_endpoint=yep.yep:6443
+ shift
+ shift
+ ((  1 > 0  ))
+ case "$1" in
+ helm_flags[${#helm_flags[*]}]=--create-namespace
+ ((  1 > 1  ))
+ shift
+ ((  0 > 0  ))
+ case "$openshift" in
+ cluster_type=Kubernetes
+ case "$external_endpoint" in
++ cut -d: -f1
+ external_hostname=yep.yep
++ cut -d: -f2
+ external_port=6443
+ set -e
+ helm upgrade -i -n kubestellar --create-namespace kubestellar /opt/homebrew/core-helm-chart --set clusterType=Kubernetes --set-string EXTERNAL_HOSTNAME=yep.yep --set EXTERNAL_PORT=6443 --set image.tag=v0.8.0-alpha.13.git180d07e-clean
Error: path "/opt/homebrew/core-helm-chart" not found
```

MikeSpreitzer:
There is this great utility named `realpath` in Linux. I did not have it on my previous MacBook. And now here it is!

```
mspreitz@mjs13 kubestellar % which realpath
/bin/realpath
```

clubanderson:
@MikeSpreitzer interesting dilemma.  What do you suggest we can do to make the command less sensitive?

MikeSpreitzer:
I documented the current problem in #1152 .
Three approaches occur to me. One is to insert a use of `realpath` into the scripts that navigate to peer directories, to remove confusion from symlinks; I think this would be the simplest and I have seen it used elsewhere. Another approach is to have `make build` put something in `bin/` that writes the needed content to stdout. A third approach is to have `make build` compile the needed content into the executable in `bin/` that uses it.

clubanderson:
My vote is for realpath. If it is commonly used by others in the opensource and cloud native landscape, then it should be good enough for us. I'm also wondering how much this problem affects our kubectl plugins vs our scripts. To me, this problem appears to be script specific.

clubanderson:
@MikeSpreitzer is approval pending on #1152 or can we go ahead here?

clubanderson:
@MikeSpreitzer  I think the 2 scripts effected by #1152 are:

kubectl-kubestellar-deploy
kubestellar init

These aren't scripts our users need. They are for collaborators/contributors (maybe). If we are focused on users of kubestellar, then I think there is no harm to get this PR merged and then work on fixing the scripts for the next release.

brew/README.md
MikeSpreitzer: This looks like #1153

clubanderson: going to update this - I do not think we need the url at the end.

clubanderson:
@francostellari @MikeSpreitzer ready for final review.  I won't know if the tap works until this gets merged

PR #1146, opened 2023-10-17
🌱 Align default core image ref in chart with coming release

MikeSpreitzer:
## Summary
This PR updates the default core image tag in the Helm chart's `values.yaml` to be named after the coming release. The theory is that a tag naming a release branch can be advanced along backward-compatible changes to pick up improvements and bug fixes while being specific enough to insulate a user of one branch from work on the next branch.

## Related issue(s)

Fixes #

### DIFF
--- a/core-helm-chart/values.yaml
+++ b/core-helm-chart/values.yaml
@@ -16,7 +16,7 @@
 image:
   repository: quay.io/kubestellar/kubestellar
   pullPolicy: IfNotPresent
-  tag: main-mc
+  tag: release-0.9
 
 # Storage size
 storageSize: 8Gi


clubanderson:
is this still needed @MikeSpreitzer ?

MikeSpreitzer:
We have not discussed and agreed on a plan for image tagging, so I have to temper my answer with a qualification that I am only speaking for myself at this point.

Because the default tag in the `v0.8.0` chart is "main-mc", we have to take the approach that "main-mc" is synonymous with "release-0.8". This means that as soon as we make any change that is not intended to be part of the 0.8 release stream, the default tag in the chart has to be something other than "main-mc". We are not there today, but I expect we will get there soon.

clubanderson:
I understand your reasoning. Since this is in main branch only, let's give it a whirl.

PR #1155, opened 2023-10-18
:sparkles:  Increase kind version to v0.20 for ubuntu

fab7:
## Related issue(s)

Fixes #


MikeSpreitzer:
Why? Why only the one OS? Is this tested?

fab7:
@MikeSpreitzer - Correct, only Ubuntu OS was updated because I did not test the other ones. However one could update the version of all our links because the KinD Web site refers to v0.20 for all the  systems now (Linux, macOS and Windows). Also, this was simply an exercise to try out the procedure for creating my first PR.

PR #1168, opened 2023-10-19
✨ Update values.yaml image tag to `latest`

francostellari:
## Summary

Update values.yaml image tag to latest

## Related issue(s)

Fixes #

### DIFF
--- a/core-helm-chart/values.yaml
+++ b/core-helm-chart/values.yaml
@@ -16,7 +16,7 @@
 image:
   repository: quay.io/kubestellar/kubestellar
   pullPolicy: IfNotPresent
-  tag: release-0.9
+  tag: latest
 
 # Storage size
 storageSize: 8Gi


clubanderson (commented):
/lgtm
/approve

MikeSpreitzer:
I think that this is a bad idea. It invites users to make Deployment objects using the image tag "latest". Doing that means that any work we do concurrently can sabotage that user, assuming we update what "latest" means.

The idea behind "release-0.9" is that it identifies a _release branch_ . That is not one point release, it is a branch of development that only includes backward-compatible changes.

MikeSpreitzer:
For a point of comparison, see https://docs.bitnami.com/tutorials/understand-rolling-tags-containers/

@francostellari : are you recommending that users always override the `image.tag` setting?

clubanderson:
/lgtm
/approve

we can always change this to something else in the future if it does not serve us well.

clubanderson:
@francostellari please rebase so this can merge

MikeSpreitzer:
This is moot now.

PR #1185, opened 2023-10-23
✨ Introduce kubestellar-list-syncing-objects

MikeSpreitzer:
## Summary
This PR adds a demo/utility command that lists or watches the objects of a given kind in the mailbox workspaces.

Example usage:
```
kubestellar-list-mid-dynamic --api-group apps --api-kind ReplicaSet
```

## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,7 +114,7 @@ ldflags:
 require-%:
 	@if ! command -v $* 1> /dev/null 2>&1; then echo "$* not found in \$$PATH"; exit 1; fi
 
-build: WHAT ?= ./cmd/kubectl-kubestellar-syncer_gen ./cmd/kubestellar-version ./cmd/kubestellar-where-resolver ./cmd/mailbox-controller ./cmd/placement-translator
+build: WHAT ?= ./cmd/kubectl-kubestellar-syncer_gen ./cmd/kubestellar-version ./cmd/kubestellar-where-resolver ./cmd/mailbox-controller ./cmd/placement-translator ./cmd/kubestellar-list-syncing-objects
 build: require-jq require-go require-git verify-go-versions ## Build all executables
 	GOOS=$(OS) GOARCH=$(ARCH) CGO_ENABLED=0 go build $(BUILDFLAGS) -ldflags="$(LDFLAGS)" -o bin $(WHAT)
 	cp outer-scripts/* bin/
@@ -122,7 +122,7 @@ build: require-jq require-go require-git verify-go-versions ## Build all executa
 	cp inner-scripts/* bin/
 .PHONY: build
 
-userbuild: WHAT ?= ./cmd/kubectl-kubestellar-syncer_gen ./cmd/kubestellar-version
+userbuild: WHAT ?= ./cmd/kubectl-kubestellar-syncer_gen ./cmd/kubestellar-version ./cmd/kubestellar-list-syncing-objects
 userbuild: require-jq require-go require-git verify-go-versions ## Build executables needed by users outside the core image
 	GOOS=$(OS) GOARCH=$(ARCH) CGO_ENABLED=0 go build $(BUILDFLAGS) -ldflags="$(LDFLAGS)" -o bin $(WHAT)
 	cp outer-scripts/* bin/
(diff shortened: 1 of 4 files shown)


MikeSpreitzer:
/cc @waltforme 
/cc @clubanderson 
@yana1205

clubanderson:
I like the idea, absolutely hate the name.  Can we make something that is super simple here?  simply "kubestellar-monitor-resources" would be great.

MikeSpreitzer:
I do not love the name either.
I put "mid" in the name suggesting "middle" because it is specifically looking in a middle type of place, _between_ the WECs and the WDSes. The objects of interest also appear in the WECs, and in the WDSes in the 1:1 case. So I thought it important that the name be specific to the fact that it is looking in the middle.

This command is ambidextrous: it can produce a one-shot listing, and it can do on-going monitoring. I expect that for your new quickstarts the one-shot listing will be the more interesting case. There is some precedent for taking the "on-going" part as a modulation of "list": `kubectl get` can get a list of objects, and takes an optional command line flag that turns it into an on-going watch.

cmd/kubestellar-list-mid-dynamic/main.go
```diff
@@ -0,0 +1,180 @@
...
+	if len(kind) == 0 {
+		logger.Error(nil, "The --kind must not be the empty string")
+	}
+	if len(gvr.Resource) == 0 {
+		gvr.Resource = strings.ToLower(kind) + "s"
+	}
+
+	parentClientConfig, err := parentClientOpts.ToRESTConfig()
+	if err != nil {
+		logger.Error(err, "failed to make ESPW client config")
```
waltforme, line 94: Maybe change from 'ESPW' to 'parent' here as well?

cmd/kubestellar-list-mid-dynamic/main.go
```diff
@@ -0,0 +1,180 @@
...
+	parentInformerFactory := kcpinformers.NewSharedScopedInformerFactory(parentClient, 0, "")
+	mbPreInformer := parentInformerFactory.Tenancy().V1alpha1().Workspaces()
+
+	informer := mailboxwatch.NewSharedInformer[k8sdynamic.NamespaceableResourceInterface, *unstructured.UnstructuredList](ctx, listGVK, mbPreInformer, dynamicClusterResource, exampleObj, 0, upstreamcache.Indexers{})
+
+	informer.AddEventHandler(upstreamcache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj any) { log(logger, "add", obj) },
+		UpdateFunc: func(oldObj, newObj any) { log(logger, "update", newObj) },
+		DeleteFunc: func(obj any) { log(logger, "delete", obj) },
+	})
```
waltforme, line 132: It is fun and creative to me that, logic other than enqueuing are written here in the ResourceEventHandlerFuncs. I like it 👍

MikeSpreitzer:
The force-push to 1004e8dab fixes import ordering and the editing oversight called out in review above.

clubanderson:
Let's talk this through, while this command works, it will be lost in obscurity if it is not renamed.

MikeSpreitzer:
I am open to renaming. I mentioned some of my concerns above. The "dynamic" part is descriptive for Kubernetes experts but is perhaps not something that needs to be in the name. Following are some suggestions.

- `kubestellar-list-middle-objects`
- `kubestellar-list-objects-in-middle`
- `kubestellar-list-full-objects`

clubanderson:
how about 'kubestellar-list-objects-pending-sync'

cmd/kubestellar-list-mid-dynamic/main.go
MikeSpreitzer: Thanks, fixed.

MikeSpreitzer:
Remember that an object subject to downsync has state flowing in two directions: the desired state flows from core to WEC, and the reported state flows from WEC to core.

A mailbox space holds both objects subject to downsync and objects subject to upsync.

These objects appear there regardless of whether any propagation remains to be done, so "pending" is not quite on-point.

MikeSpreitzer:
An object in a mailbox space may already have been synced.  Yes, qualifying with "pending sync" and "to sync" have the same issue: they imply that the list will only have objects that have not yet been synced.

MikeSpreitzer:
How about even simpler, `kubestellar-list-syncing-object` ?

clubanderson:
sold

MikeSpreitzer:
The force-push to 25d8d279c is a rebase plus the agreed renaming plus adding documentation.

PR #1191, opened 2023-10-25
✨ Allow IP address instead of domain name as external hostname

MikeSpreitzer:
## Summary
This PR generalizes the generation of external kubeconfigs to allow for the possibility of using an IP address rather than a domain name to identify where the hosting cluster's Ingress controller is listening.

This PR also does a few updates that should have been done a long time ago.

## Related issue(s)

Fixes #

### DIFF
--- a/core-container/entry.sh
+++ b/core-container/entry.sh
@@ -112,11 +112,15 @@ function run_kcp() {
 
     # Create the certificates
     if [ -n "$EXTERNAL_HOSTNAME" ]; then
-        echo "Creating the TLS certificates"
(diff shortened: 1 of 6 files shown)


MikeSpreitzer:
Oh dang, this looks like a showstopper.

```
bash-3.2$ kubectl kubestellar deploy --external-endpoint 192.168.1.42:1119 -n kubestellar --create-namespace
Error: 1 error occurred:
	* Ingress.extensions "kubestellar-ingress" is invalid: spec.rules[0].host: Invalid value: "192.168.1.42": must be a DNS name, not an IP address
```

MikeSpreitzer:
Closing until such time, if any, as there is an idea for how to make the ingress work.

PR #1201, opened 2023-10-26
✨ Start drafting v3 API

MikeSpreitzer:
## Summary
This PR proposes a v3 of the `edge.kubestellar.io` Kubernetes API group.

This is just a statement of what a nicer API might look like. I do not yet have a position on coexistence with v2 vs. jump from v2 to v3.

## Related issue(s)

Fixes #


MikeSpreitzer:
We went with a smaller change, in #1202

PR #1202, opened 2023-10-26
✨ Propose dirty change to v2beta1 kube API

MikeSpreitzer:
## Summary
This PR proposes a somewhat backward-compatible change to v2beta1 of KubeStellar's Kubernetes API regarding how objects are identified for downsync.

This draws on ideas also expressed in #1201 but, as a change to the existing version, does not raise issues of coexist-vs-jump. The changes are _somewhat_ backward compatible: for the EdgePlacement YAMLs that I have seen, the same objects would be selected for downsync. It is not fully backward compatible: I can imagine other YAMLs whose meaning is changed by this PR (and a unit test had an EdgePlacement whose meaning changed). This is by no means an ideal situation, but may be a pragmatic move for a project at this stage.

The main idea is to make `DownsyncObjectTest` a simple conjunction of tests, with each field defaulting to matching everything. That way the user can write what they want to filter on and omit the rest. For example, to select all the namespaced objects in namespace `fooey` one could put the following simple member in the `downsync` list of an EdgePlacementSpec.

```
- namespaces: [fooey]
```

BTW, we always needed to answer the question of what happens if the same object is selected by both the downsync and upsync criteria --- but we did not. Now that this encourages broader downsync selection, it is no longer tolerable to dodge that question. This PR proposes an answer.

## Related issue(s)

Fixes #


clubanderson:
If I am understanding this correctly, this proposes to default open for edgeplacement. Meaning that if I omit a field then it is naturally "*". This is great. When we have more examples of workloads and 1:m or m:1 then we can re-examine how this logic works. This gets us closer to making the sync work without guessing what a helm chart might include. Does this change also include the ability to sync just a namespace without needing to specify sync of the objects in a namespace?

MikeSpreitzer:
@clubanderson : yes, the opening comment here shows an example of that.

MikeSpreitzer:
/cc @waltforme 
/cc @yana1205 
/cc @clubanderson 
/cc @dumb0002 
@cwiklik 
@francostellari 
@KPRoche

clubanderson:
/lgtm

@waltforme can you give a review?

waltforme:
> /lgtm
> 
> @waltforme can you give a review?

Sure, will do right after I submit my work-in-progress PRs in hand.

pkg/apis/edge/v2alpha1/edge-placement.go
```diff
@@ -74,6 +74,7 @@ type EdgePlacement struct {
 
 // EdgePlacementSpec holds a desired binding between (a) a set of Locations and
 // (b) a way of identifying objects to downsync and a way of identifying objects to upsync.
+// An object that matches both the downsync and the upsync is upsynced (i.e., upsync has precedence).
```
waltforme, line 77: So this object is upsynced **only**, not downsynced? May I know why upsync takes precedence? (not against the idea at all, just want to understand the thoughts/reasoning behind it)

MikeSpreitzer: Remember that we define the jargon very carefully. "downsynced" and "upsynced" are adjectives that apply to an object, _not_ the past tense of a verb. A downsynced object has state flowing in both directions: desired state from core to WEC, reported state from WEC to core.

In the quest to make it easy for a user to identify all the objects that she wants downsynced, we inevitably make the user cast a wide net. If there **are** any objects to be upsynced, it is not unlikely that they will be a subset of those that match the downsync criteria.

waltforme: I see. Thanks for your help to understand this.

PR #1211, opened 2023-10-27
🐛 Update the ready check for core helm chart

waltforme:
## Summary
The command loops forever for me, without specifying the `kubestellar` namespace.


MikeSpreitzer (commented):
/lgtm
Thanks!

PR #1214, opened 2023-10-27
🐛Change the way entry.sh determine if the kcp container has completed …

### LINKED ISSUE #1213: bug: KubeStellar controller container fails to determine that kcp container is ready if the log overflows
### Describe the bug

If one of KubeStellar container restart after the kcp container has been running has been running for a long time, it may fail to find the `***READY***` flag because in the kc container, due to log overflow.

### Steps To Reproduce

1. run KubeStellar relwase v0.9.0 using the helm chart
2. wait a sufficient time until the `***READY***` flag is not visible in the `kcp` container log
3. restart any of the controller containers
4. observe that the container get stuck at the step of waiting for kcp to be ready

### Expected Behavior

The container should restart and find kcp ready, then proceed to the next step

### Additional Context

This happens because the kcp log overflows

francostellari:
## Summary

To fix #1213 we revert back to check on the existence of file `ready` in `kcp` container instead of flag `***READY***` in `kcp` container log, which may overflow

## Related issue(s)

Fixes #
#1213

### DIFF
--- a/core-container/entry.sh
+++ b/core-container/entry.sh
@@ -28,12 +28,12 @@ function wait_kcp_ready() {
     echo "Waiting for kcp to be ready... this may take a while."
     (
         KUBECONFIG=
-        # while ! kubectl exec $(kubectl get pod --selector=app=kubestellar -o jsonpath='{.items[0].metadata.name}') -c kcp -- ls /home/kubestellar/ready &> /dev/null; do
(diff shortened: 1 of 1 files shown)


francostellari:
\cc @dumb0002

clubanderson (commented):
/lgtm

Issue #1080, opened 2023-10-04
feature: automatic configuration of nonnamespaced apigroup, resources, and names based on the resources in a wmw


clubanderson:
### Feature Description

Why are we making our customers indicate the nonNameSpacedObjects they need to support their workloads, when the information is in the resources that are in the wmw?

### Proposed Solution

remove the need for users to handcraft the nonNameSpacedObjects in the edgeplacement.  Either remove this section in the spec or create it automatically for them.

### Want to contribute?

- [ ] I would like to work on this issue.

pdettori:
Are there use cases where there are resources that may exist in the WMW that should not be delivered to the WECs ?

ezrasilvera:
@clubanderson How do you suggest we decide which resources are downsynced? How do we decide today which cluster-wide-resources we downsync ?

MikeSpreitzer:
@clubanderson, just to make sure I understand: the suggestion here is that users should be unable to configure which cluster-scoped objects are downsynced, instead every EdgePlacement object implicitly selects all cluster-scoped objects? If so, that seems a bit blunt to me. Convenient in some cases, but blunt. For one example, note that the KubeStellar control objects are cluster-scoped, but they do not make sense to downsync. There are also the APIBinding objects that are needed to even make the KubeStellar API available. Perhaps the suggestion is that every EdgePlacement object implicitly selects all cluster-scoped objects except the KubeStellar control objects and the APIBinding objects? I suspect that the list of desired exceptions will not end there. Switching to kube-bind will add other control objects, for one example. Remember that we also have an open issue for how to deal with RedHat-style operators; that may add additional control or other sorts of infrastructure cluster-scoped objects that should not be downsynced. Would you settle instead for an explicit list of APIGroups with no more details per APIGroup?

clubanderson:
I am not implying that all clusterscoped resources in a cluster should be downsynced.  What I am saying is that it is easy to identify which clusterscoped resources are required by a workload.  Once identified, those resources, resourcenames, apigroups should be automatically added to the spec for edgeplacement.  We should not require an operator to iteratively guess at what clusterscoped resources they need in their MB.  And, to add to this, the clusterscoped resources likely already exist at the synctarget, so the only reason why they are needed in the MB is so that the workload can be applied there.  So the cluster-scoped objects are just transitory and are (likely) not needed at the target.  So that is even more the reason for not asking the operator to do the work manually.

clubanderson:
[...]

Issue #1064, opened 2023-10-02
feature: add 'customresourcedefinitions' as a default resourceName in our examples for edgeplacements


clubanderson:
### Feature Description

customeresourcedefinitions are very common and it would be a good idea to include this in our examples on kubestellar.io

### Proposed Solution

replace:

    resourceNames: [ "bind-kubernetes", "bind-apps" ]

with:

    resourceNames: [ "bind-kubernetes", "bind-apps", "customresourcedefinitions" ]

### Want to contribute?

- [ ] I would like to work on this issue.

MikeSpreitzer:
Here is my take on this. First, for downsynced objects: just like a namespace is automatically created when/where needed to transport a namespaced object that the explicit configuration says is downsynced, I think that it would be reasonable to automatically carry the needed CustomResourceDefinition (if any) along too. There is a little more to say, regarding the ObjectMeta of the Namespace object, and thus corresponding things to say about the ObjectMeta of the CustomResourceDefinition object.

Multiple WDSes can direct workload objects to the same WEC. You might expect this to be uncommon. For upsynced objects, you might expect it to be common for mulitple WECs to direct the same kind/namespace/name to the same WDS. This makes the question of consistency of colliding CRDs more prominent.

clubanderson:
thank you for finding the backlog item here.  I think this will really help our customers.

MikeSpreitzer:
FYI, here is what an APIBinding has to say for itself.

```
bash-3.2$ kubectl get apibindings tenancy.kcp.io-7axfd -o yaml
apiVersion: apis.kcp.io/v1alpha1
kind: APIBinding
metadata:
  annotations:
    kcp.io/cluster: 24ea110rxpjxryp1
  creationTimestamp: "2023-10-13T01:16:18Z"
  finalizers:
  - apis.kcp.io/apibinding-finalizer
  generation: 1
  labels:
    claimed.internal.apis.kcp.io/bmCdly9xXiUpEHe3ypvDwvXMTfoVZUE92mqAQf: LstvmbbzVDDOn90ZbhzSQO5U3DMCf88h1pZ
    internal.apis.kcp.io/export: bmCdly9xXiUpEHe3ypvDwvXMTfoVZUE92mqAQf
  name: tenancy.kcp.io-7axfd
  resourceVersion: "1663"
  uid: a0bb1ef6-97b6-4c5c-81d7-46dd6327b889
spec:
  reference:
    export:
[...]

PR #1221, opened 2023-10-30
✨ Add implicit selection of CRDs, APIBindings

MikeSpreitzer:
## Summary
This PR modifies the meaning of the downsync part of the "what predicate" in an EdgePlacement to implicitly include `CustomResourceDefinitions` and `APIBinding` objects that define the kinds of other selected objects.

This PR also makes the meaning downsync part of the "what predicate" exclude the things that match upsync part.

## Related issue(s)

I think this addresses the concerns behind #1064 and #1080.


MikeSpreitzer:
This PR makes it so that you do not have to explicitly talk about the CRD objects in the "what predicate" _regardless_ of whether they define a namespaced resource or a cluster-scoped resource. Implicit matching of CRD objects is demonstrated in the update to example1.

MikeSpreitzer:
This is a backward-compatible change.

Issue #909, opened 2023-08-16
feature: UX - help users debug their "what predicates"


MikeSpreitzer:
### Feature Description

In #903 we see a case where an error in a "what predicate" was mistaken for a malfunction of the KubeStellar platform.

In #545 we see a request to allow sloppier "what predicates".

Both could be addressed by giving users an easy way to debug their "what predicates".

### Proposed Solution

One way to address this would be to add a command that takes an EdgePlacement and a kubeconfig pointing to a workload description space and evaluates the EdgePlacement's "what predicate" against the current contents of the workload description space.

Another way would be to make the placement translator maintain, analogously to the SinglePlacementSlice objects that show the resolution of the "where predicate", API objects that show the resolution of the "what predicate".

Another would be through logging at a numerically-low V level.

### Want to contribute?

- [ ] I would like to work on this issue.

MikeSpreitzer:
Another approach would be extending the status section of an EdgePlacement object to reflect what is currently matching the "what predicate".  Currently, for namespaced resources, this is the outer product of two things: the namespaces selected and all the namespaced resources in the same space.  That listing of namespaced resources could get large.  If we make a change along the lines of #884 then that could greatly mitigate that scaling problem.

clubanderson:
adding to backlog, no commitment for sprint 3

MikeSpreitzer:
A common error that may deserve particular attention is the user writing a resource or object name in the "what predicate" that matches nothing. Perhaps if there were some way of letting the user know that a name they wrote is matching nothing, along with one or a few examples of things that it was compared with, this might be helpful.

MikeSpreitzer:
I think that @clubanderson's recent experience as a user makes it clear that this is urgently needed, so I am bringing it out of the backlog.

clubanderson:
this should be a kubectl command that shows what is going to be synced (matches the 'what' predicate, is in the WDS and is in the MB).  If its not matching any of those criteria (wds, mb, label) then alert and give some info on which one it did not match.

clubanderson:
this could also be an enumerated list in the SPS

clubanderson:
SPS is an internal interface.  Not something we would like to expose necessarily.  We externalize the 'where' already, now we need to find a place to deposit/externalize the 'what'

PR #1222, opened 2023-10-30
✨ UX - Report what resolution

MikeSpreitzer:
## Summary
This PR introduces a kind of API object for reporting the objects that match an EdgePlacement's "what predicate" and extends the placement controller to maintain these objects.

This PR is currently built on top of #1221. Once that merges, I will rebase this one.

## Related issue(s)

This is intended to address #909.


MikeSpreitzer:
This is a backward-compatible change.

MikeSpreitzer:
The force-push to a973815d9 is a rebase onto `main`.

MikeSpreitzer:
The force-push to d0c3213b0 is a rebase and extends the reported info about each workload part to include the mode bits handled by the what-resolver.

ezrasilvera:
@MikeSpreitzer is this still relevant ?  ks-0.20 is using completely different code, right ?

MikeSpreitzer:
@ezrasilvera : the API part is still relevant, even though the implementation part needs major adaptation.

PR #1224, opened 2023-10-30
✨ Reproduce kube-bind example for wmw1 and imw1 in the init container

waltforme:
This PR reproduces the kube-bind README.md scenario, but inside our kubestellar deployment's init container, via some initial scripting.

After `helm install`, I checked the syncing of a mangodb object between `wmw1` and `espw` to make sure it behaves as expected, including all CRUD operations.

/cc @ezrasilvera

### DIFF
--- a/inner-scripts/kubestellar
+++ b/inner-scripts/kubestellar
@@ -323,8 +323,31 @@ function ensure_espw() {
     else
         kubectl ws create "$espw_name" --enter
     fi
+    run_dex
+    run_kb_backend
     eval kubectl apply -f "$bindir/../config/exports" $verbdir
     echo "Finished populating the espw with kubestellar apiexports"
+}
+
+function run_dex() {
+    echo "--< starting dex in background >--"
+    dex serve /home/kubestellar/dex/dex-config-dev.yaml &
+    if ! pgrep dex &> /dev/null; then
+        echo "dex not started"
+        exit 1
+    fi
+}
+
+function run_kb_backend() {
+    echo "--< starting kube-bind example-backend in background >--"
+    kubectl ws root:espw
+    kubectl apply -f /home/kubestellar/kube-bind/deploy/crd
+    kubectl apply -f /home/kubestellar/kube-bind/deploy/examples/crd-mangodb.yaml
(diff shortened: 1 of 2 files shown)


core-container/entry.sh
```diff
@@ -178,6 +178,10 @@ function run_init() {
     echo "--< Starting init >--"
     wait_kcp_ready
     kubestellar init --local-kcp false --ensure-imw $ENSURE_IMW --ensure-wmw $ENSURE_WMW
+    run_dex
```
ezrasilvera, line 181: Don't you need to make sure it's running as you do it in the background ?

core-container/entry.sh
```diff
@@ -218,6 +222,41 @@ function run_placement_translator() {
...
+function run_kb_backend() {
+    echo "--< Starting kube-bind example-backend in background >--"
+    kubectl ws root:espw
+    kubectl apply -f /home/kubestellar/kube-bind/deploy/crd
+    kubectl apply -f /home/kubestellar/kube-bind/deploy/examples/crd-mangodb.yaml
+    example-backend --oidc-issuer-client-id=kube-bind --oidc-issuer-client-secret=ZXhhbXBsZS1hcHAtc2VjcmV0 --oidc-issuer-url=http://127.0.0.1:5556/dex --oidc-callback-url=http://127.0.0.1:8080/callback --pretty-name="BigCorp.com" --namespace-prefix="kube-bind-" --cookie-signing-key=bGMHz7SR9XcI9JdDB68VmjQErrjbrAR9JdVqjAOKHzE= --cookie-encryption-key=wadqi4u+w0bqnSrVFtM38Pz2ykYVIeeadhzT34XlC1Y= --consumer-scope="Cluster" &
+}
+
+
+function run_kb_konnector_for_wmw() {
```
ezrasilvera, line 240: Don't we want to do this in the ensure-wmw ?

core-container/entry.sh
```diff
@@ -218,6 +222,41 @@ function run_placement_translator() {
...
 
 
+function run_dex() {
+    echo "--< Starting dex in background>--"
+    dex serve /home/kubestellar/dex/dex-config-dev.yaml &
+}
+
+
+function run_kb_backend() {
+    echo "--< Starting kube-bind example-backend in background >--"
```
ezrasilvera, line 232: don't we want to do this when we create espw ?

core-container/entry.sh
waltforme: Yes, moved to ensure-wmw.

core-container/entry.sh
waltforme: Yes, added check for it.

core-container/entry.sh
waltforme: Yes, moved into `function ensure_espw()`

waltforme:
Also added binding for `imw1`.
```
bash-5.1$ kc ws root:imw1
Current workspace is "root:imw1".
bash-5.1$ kc get crd | grep mango
mangodbs.mangodb.com              2023-10-31T20:01:45Z
bash-5.1$ kc ws root:wmw1
Current workspace is "root:wmw1".
bash-5.1$ kc get crd | grep mango
mangodbs.mangodb.com              2023-10-31T20:01:49Z
```

PR #1228, opened 2023-10-31
:bug: support status update for both cases of subresource and non subresource

yana1205:
## Summary

MCAD has fixed to move `status` field of [appWrapper](https://github.com/cwiklik/multi-cluster-app-dispatcher/blob/mcad133/config/crd/bases/mcad.ibm.com_appwrappers.yaml) in subresource since v1.33. In the previous versions (at least before v1.31), the `status` hadn't been subresource. Such `status` can not be updated by [UpdateStatus method.](https://github.com/kubernetes/client-go/blob/v0.28.3/dynamic/interface.go#L36C4-L36C4) In order to address this issue, we made a tentative fix so that we use UpdateStatus method for built-in k8s resources while just [Update method](https://github.com/kubernetes/client-go/blob/v0.28.3/dynamic/interface.go#L35). (The corresponding git issue is   https://github.com/kubestellar/kubestellar/pull/947/files#diff-8f3c8f861fc2c427af64605adf617851b657a18cb8518cc4b46267b11014fa48R336)

To support both cases, I add a check logic if the status field of the resource to be status-upsynced is defined in subresource or not in this PR.
 
FYI: @dumb0002 

## Related issue(s)

Fixes #947, #945

### DIFF
--- a/pkg/syncer/clientfactory/client.go
+++ b/pkg/syncer/clientfactory/client.go
@@ -28,12 +28,17 @@ import (
 )
 
 type Client struct {
-	ResourceClient dynamic.NamespaceableResourceInterface
-	scope          meta.RESTScope
+	ResourceClient          dynamic.NamespaceableResourceInterface
+	scope                   meta.RESTScope
+	hasStatusInSubresources bool
 }
 
 func (c *Client) IsNamespaced() bool {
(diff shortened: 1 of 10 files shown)


MikeSpreitzer (commented):
/lgtm
Thanks!

Issue #1026, opened 2023-09-21
bug: Documentation no longer specifies go (and version) in prerequisites to install


KPRoche:
### Describe the bug

The "required packages" include files no longer mention go, and any of the examples which include a "make build" command may fail as a result.

### Steps To Reproduce

1. go to https://docs.kubestellar.io/main/Coding%20Milestones/PoC2023q1/example1/
2. Read and follow the instructions to install prerequisites
3. Continue down to https://docs.kubestellar.io/main/Coding%20Milestones/PoC2023q1/example1/#get-kubestellar
4. reach the make build command and run it
5. if go is not installed, or too old a version is installed, this will fail

### Expected Behavior

Add mention that _if you will be building kubestellar from the source, you need to have go>= 1.19 installed_. This used to be in the prerequisites include file.
If it was removed from there as excess because its not needed for the containerized quickstart using bootstrap, which also pulls in those required software include files, call out the requirement for go in the main file after the prereqs section.

clubanderson:
@KPRoche I think that is in the top of the page under required packages, no?

clubanderson:
hmm, well, that was there a while back - guess it needs to be re-added

KPRoche:
nope 
![image](https://github.com/kubestellar/kubestellar/assets/25445603/0522597b-22a1-4523-b593-1da50fb17f26)
![image](https://github.com/kubestellar/kubestellar/assets/25445603/f2b13166-1223-4985-be1e-8cffc13f72ff)

I think it was pulled out in the name of simplifying the quickstart

KPRoche:
> hmm, well, that was there a while back - guess it needs to be re-added
It is a conditional requirement IFF you are doing a build. Might be worth standing it out from the main prereqs as an optional requirement

KPRoche:
QUESTION: **when** is go actually required? only for doing a make build or code development, or does it need to be installed for the using one of the release archives?
In other words, if I'm not building my own binaries, but just want to run from one of the releases, do I need to install go?

MikeSpreitzer:
Right, it depends on the role of the person involved.  For a user of a release, there is no go compilation involved.  For a contributor who is building, go compilation is involved.

clubanderson:
@KPRoche @fab7 can you work together to knock this out?

KPRoche:
> @KPRoche @fab7 can you work together to knock this out?
@clubanderson 
[...]

PR #1233, opened 2023-10-31
:book: Restore and Update installation info for go in required packages documentation

KPRoche:
## Summary
Restore notes about installing go as a prerequisite if the user intends to build kubestellar locally

Instructions for all OS tabs were added to the file

##  in response to issue #1026 

Fixes #

### DIFF
--- a/docs/content/common-subs/required-packages.md
+++ b/docs/content/common-subs/required-packages.md
@@ -14,6 +14,12 @@
 
         + [__helm__](https://helm.sh/docs/intro/install/) (required when deploying as workload)
 
+        __If you intend to build kubestellar from source__  you will also need 
+
+        + [__go__](https://go.dev/doc/install) (Go version >=1.19 required; 1.19 recommended) [__go releases__] (https://go.dev/dl)
+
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
The new instructions are only for human reading, not docs-ecutable running in bash, right? That would be different from the other instructions; is that deliberate?

MikeSpreitzer:
Also, as a general rule: if I already have something installed by a different package manager, please do not do an additional install.

KPRoche:
> The new instructions are only for human reading, not docs-ecutable running in bash, right? That would be different from the other instructions; is that deliberate?

Yes. As I was reading the assorted instructions on the go website, it was abundantly clear that go is SUPER fussy about how it is installed (or re-installed). So I copied the instructions for different OS configurations from there as well as referring back to that page.

clubanderson (commented):
/lgtm

MikeSpreitzer:
And the GitHub workflows cover for go by having a step that installs it.
I wonder if there is a GitHub action that installs ko?

/approve

PR #1245, opened 2023-11-02
:bug: Fix some SM bugs

ezrasilvera:
## Related issue(s)

Fixes #

### DIFF
--- a/space-framework/pkg/space-manager/reconcile_spaceprovider.go
+++ b/space-framework/pkg/space-manager/reconcile_spaceprovider.go
@@ -51,9 +51,8 @@ func (c *controller) reconcileSpaceProviderDesc(key string) error {
 }
 
 func (c *controller) handleAdd(providerDesc *spacev1alpha1.SpaceProviderDesc) error {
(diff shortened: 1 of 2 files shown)


space-framework/pkg/space-manager/reconcile_spaceprovider.go
```diff
@@ -51,9 +51,8 @@ func (c *controller) reconcileSpaceProviderDesc(key string) error {
 }
 
 func (c *controller) handleAdd(providerDesc *spacev1alpha1.SpaceProviderDesc) error {
-	if providerDesc.Status.Phase == spacev1alpha1.SpaceProviderDescPhaseInitializing {
```
andreyod, line 54: So what will happen with the event when provider status phase is Initializing? We will try to set it again to Initializing and create a provider?

ezrasilvera: yes. I went through the code and it seems our Create flow is idempotent and doesn't duplicate things

PR #1247, opened 2023-11-03
:bug: fix syncer-gen to generate SA token for later K8Ss that no longer generate SA token automatically

yana1205:
## Summary

- Fix syncer-gen (`kubectl kubestellar syncer-gen`) to generate the service account token for the service account for Syncer to connect to MBS for K8Ss environments that no longer generate SA token automatically.

(Thanks Paolo for giving the manual sequences.)

## Verification
I verified in two kcp environments (regular k8s and k8s compounded in kubestellar v0.12.0 where workspace is configured so that SA token is not generated automatically) that the generated kubeconfig can be accessible to MBS.

## Related issue(s)

Fixes #


clubanderson:
/lgtm
/approve

Thank you @yana1205

PR #1250, opened 2023-11-05
:sparkles: Get space config by reference in annotation

andreyod:
## Summary
Today the space aware client can return space config by space name. This PR will add an API to return space config by reference that is set in space object annotation. This will allow us to use kube-bind special namespace which is set for service provider resource after binding, to get config for the consumer space that owns the resource.
We assume that there is one to one mapping between service provider and consumer space.

## Related issue(s)

Fixes #

### DIFF
--- a/space-framework/pkg/msclientlib/controller.go
+++ b/space-framework/pkg/msclientlib/controller.go
@@ -150,6 +150,8 @@ func (c *controller) handleAdd(space interface{}, spaceKey string) {
 		c.handleDelete(spaceKey)
 		return
 	}
+	// add space reference to cache
+	c.cacheSpaceReference(nSpace, spaceKey)
 
 	// TODO We assume the space is in a pod in the same cluster where the controllers are deployed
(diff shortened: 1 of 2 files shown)


space-framework/pkg/msclientlib/controller.go
```diff
@@ -168,8 +170,20 @@ func (c *controller) handleAdd(space interface{}, spaceKey string) {
...
+		}
 	}
 	delete(c.msClient.configs, spaceKey)
 }
+
+func (c *controller) cacheSpaceReference(space *spacev1alpha1.Space, spaceKey string) {
+	annotations := space.GetAnnotations()
+	c.msClient.lock.Lock()
+	defer c.msClient.lock.Unlock()
+	if ref, ok := annotations[refAnnotationKey]; ok {
```
ezrasilvera, line 186: Need to check for annotations != nil

andreyod: Done

PR #1251, opened 2023-11-05
:sparkles:  SF docs

ezrasilvera:
Space Framework overall doc

## Related issue(s)

Fixes #


MikeSpreitzer:
English has inherited this nasty "silent e" thing. There is one at the end of "control plane".

MikeSpreitzer:
Is this intended to be user-facing documentation showing the simplest possible usage scenario, or architecture documentation showing all the modularity?

I thought the cluster that hosts the abstraction layer does not have to be the same as the cluster that does any particular other thing (e.g., host the KubeStellar controllers).

MikeSpreitzer:
The API server shown in space-framework/docs/SAC.drawio.svg is the same as the API server shown in space-framework/docs/SF-All.drawio.svg , right?  If so then it would be good to label them the same.

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+**Space Object**, **Space** - A kind of Kubernetes API object that _represents_ a pSpace.  Its "kind" is `Space`, we sometimes include the "object" part when writing specifically about the API objects to clearly distinguish them from what they represent.
+
+**Space Provider** - A Thing that manages the life cycle of multiple pSpaces. This includes at minimal create/delete/List operations on pSpaces. KIND is an example for Space Provider (that manages KIND clusters), KCP is another example - KCP manages KCP's workspaces. 
+
+## High level architecture  & Main componenets
+![SF overall view](SF-All.drawio.svg "SF overall viewgit s")
+
+The Space Framework (SF) is a generic management framework for space providers and pSpaces. The framework defines an abstraction layer for space providers and pSpace management that allows clients (both CLI/script based and client-go based clients) to use spaces while mantaining the clients decoupled from the specific pSpace and space provider that is being used.
```
MikeSpreitzer, line 13: Is there a CLI to the Space Framework?  I thought that there is only a kube-style API.  Of course, `kubectl` can be used to manipulate any kube API objects, that goes without saying.

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+The The space framework uses two CRDs:
+- **SpaceProviderDesc:** This CRD represnts a space provider it holds the information needed in order to interact with a specific Space  provider. 
+- **Space:** This CRD represnts a space as defined above. It includes reference to secrets that allow clients to connect to the pSpace that it represents and also several attributes used by the SF that will be described later on. 
+
+The space fromework supports several ways to add new spaces to the frameowrk, this is controlled by definiing differet space types: 
+
+1. Managed Space: The pSpace is created through an explicit request to the management api server to create a new Space. For Managed spaces the space framework initiates a request to the Space provider to create the pSpace, the space frame work will then continously reconsile between the Space and the pSpace states. For managed spaces the desired state is controled through the Space object. For example, in order to delete the pSpace the client need to delete the corresponding Space object and the SF will interact with the space provider in order to delete the pSpace.
```
MikeSpreitzer, line 21: "reconcile"

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+2. Imported spaces: A Space can be created by importing an existing pSpace. In this case the Space is not linked to any space provider and the client who imports that Space need to supply the access information that allows other clients to connect to the pSpace. For imported spaces the desired state is derived by the pSpace. This means that the SF is not responsible for the life cycle management of the pSpace but only update the corresponding Space object. For example, when deleteing the space object of an existing imported pSpace the space framework will recreate the space object.
```
MikeSpreitzer, line 22: > when deleteing the space object of an existing imported pSpace the space framework will recreate the space object

Where is the information about this space stored besides the Space object?

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+3. Unmanaged Space: The SF supports discovery of existing pSpaces on a space provider and adding them as "Unmanaged Spaces" into the framework. This mode is similar to the import mode in terms of the "source of truth" - the desired state is defined by the pSpace. However, as oppose to import, here the pSpaces are automatically discovered through the space provider regular interface and the pSpace access information is retrieved from teh space provider similary to how it is done in the managed space case. The SF supports pattern based filtering rules to limit the discovery of pSpaces. 
+
+### Space Manager 
+The Space Manager (SM) is a Kuberentes controller that is responsible for maintaing the state of the Space objects. The SM reconcile the Space and SpaceProviderDesc objects, and monitor the state of the pSpaces through the space provider.
```
MikeSpreitzer, line 26: "reconciles"

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Space Manager 
+The Space Manager (SM) is a Kuberentes controller that is responsible for maintaing the state of the Space objects. The SM reconcile the Space and SpaceProviderDesc objects, and monitor the state of the pSpaces through the space provider.   
+The SM uses a library of space provider adaptors to communicate with the space providers. Currently the SF includes 3 space provider adaptors KCP, KIND, and KubeFlex.
+
+### Space provider adaptors
+The space manager uses a set of space provider adaptors that interacts with the space provider. All space provider adaptors implement a simple "space client interface" that includes basic pSpace life cycle operatios. In addition the interface also defines the events that sent from the adaptors to the space manager.
```
MikeSpreitzer, line 30: "operations"

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Space aware client (SAC)
+The Space aware client (msclient) allows clients/controllers to easily get access to the underlying space by simply using the space name. There is no direct interaction between the msclient and the Space  provider, and therefore the msclient is transparent to the specific provider of the spaces. 
+![Space Aware Client](SAC.drawio.svg "Space Aware Client")
+The only requirement is that the space can be accessed through regular Kube APIs when using the appropriate kubeconfig information. 
+
+**Main features**  
+- Constantly watch for changes in the available spaces. The SAC is using an informer on Space objecs using the SF Space client-go APIs.
+    - Holds a cache of the pSpace access info for each Space 
+- Exposes utility functions to get RestConfig by simpy specifying the space name
```
MikeSpreitzer, line 40: .. and Space object's namespace?

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+The Space aware client (msclient) allows clients/controllers to easily get access to the underlying space by simply using the space name. There is no direct interaction between the msclient and the Space  provider, and therefore the msclient is transparent to the specific provider of the spaces. 
+![Space Aware Client](SAC.drawio.svg "Space Aware Client")
+The only requirement is that the space can be accessed through regular Kube APIs when using the appropriate kubeconfig information. 
+
+**Main features**  
+- Constantly watch for changes in the available spaces. The SAC is using an informer on Space objecs using the SF Space client-go APIs.
+    - Holds a cache of the pSpace access info for each Space 
+- Exposes utility functions to get RestConfig by simpy specifying the space name  
+
+Currently the SAC exposes a simple function that supply the config access info for a space. In the future we will add a SAC space aware ClientSets
```
MikeSpreitzer, line 42: "the config access" is not quite English. Perhaps something like the following?

> ... that supplies the [`rest.Config`](https://github.com/kubernetes/client-go/blob/v0.24.3/rest/config.go#L53) data that gives complete access to a physical space.

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Architecture details 
+When a new SpaceProviderDesc is created the SM creates a new namespace that will be used to host all the Space objects representing pSpaces on that space provider.  
+The name of the generated namespace will be `paceprovider-<provider name>`
+
+![Space, SpaceProviderDesc and the Space provider namespace](SF-NS.drawio.svg "SF Namespaces")
+
+Note: In the future the spaces will be created in their own namespace and not in namespace per provider. 
+
+The Space also holds a reference to the SpaceProviderDesc associated with this Space and some capabilities (e.g., the space type). The status of the space holds a secret reference to the secret holding the access information for that pSpace. As described before for managed spaces the SF retrive that information from teh space provider, create the needed secrets and update the Space object with the secret reference
```
MikeSpreitzer, line 52: "retrieves"
"the"
"creates ... and updates ..."

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+**Discovery**  
+As mentioned before, the Space manager also supports discovery of spaces created out of band (i.e., not through creating Space objects).
+
+Each provider adaptor constantly watch for pSpace creation/deletion events and feeds the SM with those events. In case a pSpace that matches the discovery filtering is detected and there is no corrresponding Space object the SM creates a Space object representing this new detected pSpace and sets its mode to “unmanaged” .
```
MikeSpreitzer, line 57: "constantly watches"

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+**Cration & Deletion flow of managed Spaces**
```
MikeSpreitzer, line 59: "Creation"

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+## Architecture details 
+When a new SpaceProviderDesc is created the SM creates a new namespace that will be used to host all the Space objects representing pSpaces on that space provider.  
+The name of the generated namespace will be `paceprovider-<provider name>`
+
+![Space, SpaceProviderDesc and the Space provider namespace](SF-NS.drawio.svg "SF Namespaces")
+
+Note: In the future the spaces will be created in their own namespace and not in namespace per provider. 
+
+The Space also holds a reference to the SpaceProviderDesc associated with this Space and some capabilities (e.g., the space type). The status of the space holds a secret reference to the secret holding the access information for that pSpace. As described before for managed spaces the SF retrive that information from teh space provider, create the needed secrets and update the Space object with the secret reference
+
```
MikeSpreitzer, line 53: Somewhere around here there needs to be a discussion of the mapping between the name of the Space object and the relevant provider's identifier(s) for the corresponding physical space. For example, in the kcp provider there are two identifiers: the `logicalcluster.Name` that lies in a flat namespace and there is the pathname of the corresponding workspace (when there is one). It seems to me that this mapping (really, these mappings --- one for each sort of identifier the provider has) is a degree of freedom that the provider adapter has, and thus needs to be documented for each provider adapter.

space-framework/docs/space-framework.md
MikeSpreitzer, line 57: Presumably the SM _deletes_ Space objects that it discovered too. The omission of this from the text leaves the reader wondering. Also, are there any _update_ cases?

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+### Space provider adaptors
+The space manager uses a set of space provider adaptors that interacts with the space provider. All space provider adaptors implement a simple "space client interface" that includes basic pSpace life cycle operatios. In addition the interface also defines the events that sent from the adaptors to the space manager.
+
+### Space aware client (SAC)
+The Space aware client (msclient) allows clients/controllers to easily get access to the underlying space by simply using the space name. There is no direct interaction between the msclient and the Space  provider, and therefore the msclient is transparent to the specific provider of the spaces. 
+![Space Aware Client](SAC.drawio.svg "Space Aware Client")
+The only requirement is that the space can be accessed through regular Kube APIs when using the appropriate kubeconfig information. 
+
+**Main features**
```
MikeSpreitzer, line 37: Why not use markdown markup on this subsection header?

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+Each provider adaptor constantly watch for pSpace creation/deletion events and feeds the SM with those events. In case a pSpace that matches the discovery filtering is detected and there is no corrresponding Space object the SM creates a Space object representing this new detected pSpace and sets its mode to “unmanaged” .
+
+**Cration & Deletion flow of managed Spaces**  
+The SM sets the status of the Space according to the actual staus of the pSpace. When a space is created a creation command is sent to the space provider and the state of the the Sapce is set to `Initializing`. Only when the pSpace is becoming avaliable (each space provider validate this in its own way) the SM sets the Space state to `Ready`.   
+The SM uses finalizers for the space deletion flow - the Space object is not removed until the corresponding pSpace is deleted by the space provider.
+
+## Examples
+
+**SpaceProviderDesc**
+The folowing is SpaceProviderDesc for for a KubeFlex provider type. The secret refernced by `secretRef` should be created before the SpaceProviderDesc object is created.
```
MikeSpreitzer, line 66: (1) ".. is a SpaceProviderDesc .."

(2) What goes wrong if the Secret is not created first? FYI, Kubernetes style generally frowns on such ordering requirements.

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+spec:
+  ProviderType: "kubeflex"
+  SpacePrefixForDiscovery: "ks-"
+  secretRef:
+    namespace: default
+    name: pkflex
+```
+
+**Space**
+The folowing is a Space yaml is used to create a managed space on the previously defined KubeFlex provider.
```
MikeSpreitzer, line 81: Uppercase "YAML".

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+**Main features**  
+- Constantly watch for changes in the available spaces. The SAC is using an informer on Space objecs using the SF Space client-go APIs.
+    - Holds a cache of the pSpace access info for each Space 
+- Exposes utility functions to get RestConfig by simpy specifying the space name  
+
+Currently the SAC exposes a simple function that supply the config access info for a space. In the future we will add a SAC space aware ClientSets
+
+## Architecture details 
+When a new SpaceProviderDesc is created the SM creates a new namespace that will be used to host all the Space objects representing pSpaces on that space provider.
```
MikeSpreitzer, line 45: But there can be more than one SpaceProviderDesc for a given provider (ProviderType), right?

space-framework/docs/space-framework.md
ezrasilvera: CLI here means kubectl   will explicitly say

space-framework/docs/space-framework.md
ezrasilvera: Secrets that the Space object reference

space-framework/docs/space-framework.md
ezrasilvera: Yes. But the above say it's a NS per provider (instance) not for provider type.

space-framework/docs/space-framework.md
ezrasilvera: The SpaceProviderDesc will stay at Initializing and not move to Ready. Also an error message will be send.

space-framework/docs/space-framework.md
ezrasilvera: Added

space-framework/docs/space-framework.md
ezrasilvera: add explanation on namespace

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+## High level architecture  & Main componenets
+![SF overall view](SF-All.drawio.svg "SF overall viewgit s")
+
+The Space Framework (SF) is a generic management framework for space providers and pSpaces. The framework defines an abstraction layer for space providers and pSpace management that allows clients (both CLI/script based and client-go based clients) to use spaces while mantaining the clients decoupled from the specific pSpace and space provider that is being used. 
+
+The The space framework uses two CRDs:
+- **SpaceProviderDesc:** This CRD represnts a space provider it holds the information needed in order to interact with a specific Space  provider. 
+- **Space:** This CRD represnts a space as defined above. It includes reference to secrets that allow clients to connect to the pSpace that it represents and also several attributes used by the SF that will be described later on. 
+
+The space fromework supports several ways to add new spaces to the frameowrk, this is controlled by definiing differet space types:
```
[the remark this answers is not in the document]
ezrasilvera: Changed accordingly

space-framework/docs/space-framework.md
ezrasilvera: Correct. We still not have that mapping well defined in the SM and we use Space name == pSpace name . We definetly need to maintain the mapping between the Space name and the  provider's pSpace ID

MikeSpreitzer: I find saying "Space name == pSpace name" to be confusing.  I think of the physical space as the actual thing that the provider provides. So its name or names or IDs are the names/IDs that the provider uses. Which I think you agreed is not necessarily the same as the name of the Space object.

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
+# Space Framework
+
+## terminoloogy
```
michalmalka, line 3: Typo... Terminology

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
+# Space Framework
+
+## terminoloogy 
+**Physical Space**, **pSpace** A thing that behaves like a Kubernetes kube-apiserver (including the persistent storage behind it) and the subset of controllers in the kube-controller-manager that are concerned with API machinery generalities (not management of containerized workloads). A kcp logical cluster is an example. A regular Kubernetes cluster is another example.
+
```
michalmalka, line 5: Need to rephrase the above... hard to understand currently and better not to use thing... 
Maybe something like: 
Physical Space, pSpace - refers to an entity similar to a Kubernetes kube-apiserver, encompassing its persistent storage, along with the subset of controllers from the kube-controller-manager that focus on API machinery principles rather than the management of containerized workloads. Examples are: a kcp logical cluster and a standard Kubernetes cluster.

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+**Space Object**, **Space** - A kind of Kubernetes API object that _represents_ a pSpace.  Its "kind" is `Space`, we sometimes include the "object" part when writing specifically about the API objects to clearly distinguish them from what they represent.
+
+**Space Provider** - A Thing that manages the life cycle of multiple pSpaces. This includes at minimal create/delete/List operations on pSpaces. KIND is an example for Space Provider (that manages KIND clusters), KCP is another example - KCP manages KCP's workspaces. 
+
```
michalmalka, line 9: need to rephrase again and not use "Thing" ... 
maybe: An entity that manages the life cycle of multiple pSpaces. This entails, at the very least, performing create/delete/list operations on pSpaces. examples: KIND is a space provider that manages KIND cluster, and KCP manages KCP's workspaces.

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+## High level architecture  & Main componenets
+![SF overall view](SF-All.drawio.svg "SF overall viewgit s")
+
+The Space Framework (SF) is a generic management framework for space providers and pSpaces. The framework defines an abstraction layer for space providers and pSpace management that allows clients (both CLI/script based and client-go based clients) to use spaces while mantaining the clients decoupled from the specific pSpace and space provider that is being used. 
+
+The The space framework uses two CRDs:
```
michalmalka, line 15: remove one "The"

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+- **SpaceProviderDesc:** This CRD represnts a space provider it holds the information needed in order to interact with a specific Space  provider. 
+- **Space:** This CRD represnts a space as defined above. It includes reference to secrets that allow clients to connect to the pSpace that it represents and also several attributes used by the SF that will be described later on. 
+
+Spaces can be devided into two categories:
+1. Provider based spaces: The SM uses a space provider to get information (or perform operations) on the spaces. We can think of such pSpaces as "belong" to the space provider. There are two types of spaces in this category which control the way theses spaces are added to the frameowrk:
```
michalmalka, line 20: what is SM?

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+- **Space:** This CRD represnts a space as defined above. It includes reference to secrets that allow clients to connect to the pSpace that it represents and also several attributes used by the SF that will be described later on. 
+
+Spaces can be devided into two categories:
+1. Provider based spaces: The SM uses a space provider to get information (or perform operations) on the spaces. We can think of such pSpaces as "belong" to the space provider. There are two types of spaces in this category which control the way theses spaces are added to the frameowrk: 
+   * **Managed Space**: The pSpace is created through an explicit request to the management api server to create a new Space. For Managed spaces the space framework initiates a request to the Space provider to create the pSpace, the space frame work will then continously reconcile between the Space and the pSpace states. For managed spaces the desired state is controled through the Space object. For example, in order to delete the pSpace the client need to delete the corresponding Space object and the SF will interact with the space provider in order to delete the pSpace.
```
michalmalka, line 21: space frame work - should be framework

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+  2. Providerless spaces: This spaces don't belong to a space provider and are called "Imported spaces". Such Space is created by importing an existing pSpace. In this case the Space is not linked to any space provider and the client which imports that Space need to supply the access information that allows other clients to connect to the pSpace. For imported spaces the desired state is derived by the pSpace. This means that the SF is not responsible for the life cycle management of the pSpace but only update the corresponding Space object. For example, when deleteing the space object of an existing imported pSpace the space framework will recreate the space object. 
+### Space Manager 
+The Space Manager (SM) is a Kuberentes controller that is responsible for maintaing the state of the Space objects. The SM reconciles the Space and SpaceProviderDesc objects, and monitors the state of the pSpaces through the space provider.   
+The SM uses a library of space provider adaptors to communicate with the space providers. Currently the SF includes 3 space provider adaptors KCP, KIND, and KubeFlex.
```
michalmalka, line 26: I think you meant that SM includes 3 space provider... and not SF

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Space provider adaptors
+The space manager uses a set of space provider adaptors that interacts with the space provider. All space provider adaptors implement a simple "space client interface" that includes basic pSpace life cycle operatios. In addition the interface also defines the events that are sent from the adaptors to the space manager.
+
+### Space aware client (SAC)
+The Space aware client (msclient) allows clients/controllers to easily get access to the underlying space by simply using the space name. There is no direct interaction between the msclient and the Space  provider, and therefore the msclient is transparent to the specific provider of the spaces.
```
michalmalka, line 32: SAC and msclient are the same? need to be consistent and use only one...

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Space aware client (SAC)
+The Space aware client (msclient) allows clients/controllers to easily get access to the underlying space by simply using the space name. There is no direct interaction between the msclient and the Space  provider, and therefore the msclient is transparent to the specific provider of the spaces. 
+![Space Aware Client](SAC.drawio.svg "Space Aware Client")
+The only requirement is that the space can be accessed through regular Kube APIs when using the appropriate kubeconfig information. 
+
+**Main features**  
+- Constantly watch for changes in the available spaces. The SAC is using an informer on Space objecs using the SF Space client-go APIs.
+    - Holds a cache of the pSpace access info for each Space 
+- Exposes utility functions to get RestConfig by simpy specifying the space name, and optionally the provider name (if the provider is not specified the default provider is being used). The namespace where the Space resides can be deived from the name of the provider.
```
michalmalka, line 39: deived == derived

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+**Discovery**  
+As mentioned before, the Space manager also supports discovery of spaces created out of band (i.e., not through creating Space objects).
+
+Each provider adaptor constantly watches for pSpace creation/deletion events and feeds the SM with those events. In case a pSpace that matches the discovery filtering is detected and there is no corrresponding Space object the SM creates a Space object representing this new detected pSpace and sets its mode to “unmanaged” .
+When the pSpace is deleted the SM deletes the corresponding Space object.
+
+**Creation & Deletion flow of managed Spaces**  
+The SM sets the status of the Space according to the actual staus of the pSpace. When a space is created a creation command is sent to the space provider and the state of the the Sapce is set to `Initializing`. Only when the pSpace is becoming avaliable (each space provider validate this in its own way) the SM sets the Space state to `Ready`.
```
michalmalka, line 60: Typo - Sapce == Space

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+When the pSpace is deleted the SM deletes the corresponding Space object.
+
+**Creation & Deletion flow of managed Spaces**  
+The SM sets the status of the Space according to the actual staus of the pSpace. When a space is created a creation command is sent to the space provider and the state of the the Sapce is set to `Initializing`. Only when the pSpace is becoming avaliable (each space provider validate this in its own way) the SM sets the Space state to `Ready`.   
+The SM uses finalizers for the space deletion flow - the Space object is not removed until the corresponding pSpace is deleted by the space provider.
+
+## Examples
+
+**SpaceProviderDesc**
+The folowing is a SpaceProviderDesc for for a KubeFlex provider type. The secret refernced by `secretRef` should be created before the SpaceProviderDesc object is created.
```
michalmalka, line 66: Typo - refernced == referenced

space-framework/docs/space-framework.md
MikeSpreitzer: The Space Manager is part of the Space Framework. The Space Manager process includes three "provider adapters", in addition to the generic space management code. The _providers_ themselves are external things, like kcp, kind, and KubeFlex.

space-framework/docs/space-framework.md
ezrasilvera: Currently the "Space Name" == "the pSpace identifier used by the space provider"

space-framework/docs/space-framework.md
ezrasilvera: yes. will fix

ezrasilvera:
@MikeSpreitzer @michalmalka Apparently I pushed many spelling mistakes. Fixing now.

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+**Space Object**, **Space** - A kind of Kubernetes API object that _represents_ a pSpace.  Its "kind" is `Space`, we sometimes include the "object" part when writing specifically about the API objects to clearly distinguish them from what they represent.
+
+**Space Provider** - A Thing that manages the life cycle of multiple pSpaces. This includes at minimal create/delete/List operations on pSpaces. KIND is an example for Space Provider (that manages KIND clusters), KCP is another example - KCP manages KCP's workspaces. 
+
+## High level architecture  & Main components
+![SF overall view](SF-All.drawio.svg "SF overall view")
+
+The Space Framework (SF) is a generic management framework for space providers and pSpaces. The framework defines an abstraction layer for space providers and pSpace management that allows clients (both CLI/script based and client-go based clients) to use spaces while maintaining the clients decoupled from the specific pSpace and space provider that is being used.
```
MikeSpreitzer, line 13: There is an inconsistency here. In the Terminology section it is said that "KIND is an example for Space Provider". The sentence on which I am commenting here does not speak specifically of SpaceProviderDesc and SpaceProviderType, it only speaks of "space providers". If such a "space provider" is a SpaceProviderDesc then KIND is not an example of that (KIND is an example of SpaceProviderType). If "space provider" here means SpaceProviderType then omission of SpaceProviderDesc is leaving out a critical part.

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+The Space Framework (SF) is a generic management framework for space providers and pSpaces. The framework defines an abstraction layer for space providers and pSpace management that allows clients (both CLI/script based and client-go based clients) to use spaces while maintaining the clients decoupled from the specific pSpace and space provider that is being used. 
+
+The space framework uses two CRDs:
+- **SpaceProviderDesc:** This CRD represents a space provider it holds the information needed in order to interact with a specific Space  provider. 
+- **Space:** This CRD represents a space as defined above. It includes reference to secrets that allow clients to connect to the pSpace that it represents and also several attributes used by the SF that will be described later on. 
+
+Spaces can be divided into two categories:
+1. Provider based spaces: The Space Manager (SM) uses a space provider to get information (or perform operations) on the spaces. We can think of such pSpaces as "belong" to the space provider. There are two types of spaces in this category which control the way these spaces are added to the space framework:
```
MikeSpreitzer, line 20: s/belong/belonging/

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+- **Space:** This CRD represents a space as defined above. It includes reference to secrets that allow clients to connect to the pSpace that it represents and also several attributes used by the SF that will be described later on. 
+
+Spaces can be divided into two categories:
+1. Provider based spaces: The Space Manager (SM) uses a space provider to get information (or perform operations) on the spaces. We can think of such pSpaces as "belong" to the space provider. There are two types of spaces in this category which control the way these spaces are added to the space framework: 
+   * **Managed Space**: The pSpace is created through an explicit request to the management api server to create a new Space. For Managed spaces the space framework initiates a request to the Space provider to create the pSpace, the space frame work will then continuously reconcile between the Space and the pSpace states. For managed spaces the desired state is controlled through the Space object. For example, in order to delete the pSpace the client need to delete the corresponding Space object and the SF will interact with the space provider in order to delete the pSpace.
```
MikeSpreitzer, line 21: s/frame work/framework/

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+  2. Providerless spaces: This spaces don't belong to a space provider and are called "Imported spaces". Such Space is created by importing an existing pSpace. In this case the Space is not linked to any space provider and the client which imports that Space need to supply the access information that allows other clients to connect to the pSpace. For imported spaces the desired state is derived by the pSpace. This means that the SF is not responsible for the life cycle management of the pSpace but only update the corresponding Space object. For example, when deleting the space object of an existing imported pSpace the space framework will recreate the space object.
```
MikeSpreitzer, line 23: > This means that the SF is not responsible for the life cycle management of the pSpace but only update the corresponding Space object.

Update based on what? What updates are we talking about here?

> For example, when deleting the space object of an existing imported pSpace the space framework will recreate the space object.

When _who_ is deleting the Space object? I think you are trying to say that if anyone deletes the Space object then it will be re-created with the same content as it originally had. Is that right? I still find this surprising. Why is all that information being stored redundantly? Is there a way for the user (or an authorized colleague) who originally imported a space to undo that?

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+### Space Manager 
+The Space Manager (SM) is a Kubernetes controller that is responsible for maintaining the state of the Space objects. The SM reconciles the Space and SpaceProviderDesc objects, and monitors the state of the pSpaces through the space provider.   
+The SM uses a library of space provider adaptors to communicate with the space providers. Currently the SF supports 3 space providers KCP, KIND, and KubeFlex, and includes 3 space provider adaptors that the SM uses.
+
+### Space provider adaptors
+The space manager uses a set of space provider adaptors that interacts with the space provider. All space provider adaptors implement a simple "space client interface" that includes basic pSpace life cycle operations. In addition the interface also defines the events that are sent from the adaptors to the space manager.
```
MikeSpreitzer, line 29: s/interacts/interact/

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Space aware client (SAC)
+The Space aware client (SAC) allows clients/controllers to easily get access to the underlying space by simply using the space name. There is no direct interaction between the SAC and the Space  provider, and therefore the SAC is transparent to the specific provider of the spaces. 
+![Space Aware Client](SAC.drawio.svg "Space Aware Client")
+The only requirement is that the space can be accessed through regular Kube APIs when using the appropriate kubeconfig information. 
+
+**Main features**  
+- Constantly watch for changes in the available spaces. The SAC is using an informer on Space objects using the SF Space client-go APIs.
+    - Holds a cache of the pSpace access info for each Space 
+- Exposes utility functions to get RestConfig by simply specifying the space name, and optionally the provider name (if the provider is not specified the default provider is being used). The namespace where the Space resides can be derived from the name of the provider.
```
MikeSpreitzer, line 39: "RestConfig" is not the most common way of referring to this data structure.  It more commonly appears as `rest.Config`. Better still, tell the reader what the `rest` part refers to. A hyperlink would be a great way to do that.

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+![Space Aware Client](SAC.drawio.svg "Space Aware Client")
+The only requirement is that the space can be accessed through regular Kube APIs when using the appropriate kubeconfig information. 
+
+**Main features**  
+- Constantly watch for changes in the available spaces. The SAC is using an informer on Space objects using the SF Space client-go APIs.
+    - Holds a cache of the pSpace access info for each Space 
+- Exposes utility functions to get RestConfig by simply specifying the space name, and optionally the provider name (if the provider is not specified the default provider is being used). The namespace where the Space resides can be derived from the name of the provider.  
+
+Currently the SAC exposes a simple function that that supplies a rest.Config data structure that gives access to a physical space with the identity of a user that is authorized to do everything. In the future we will add a SAC space aware ClientSets
```
MikeSpreitzer, line 41: The first sentence here is largely redundant with the previous paragraph. Do we really need so much duplication?

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+### Space Manager 
+The Space Manager (SM) is a Kubernetes controller that is responsible for maintaining the state of the Space objects. The SM reconciles the Space and SpaceProviderDesc objects, and monitors the state of the pSpaces through the space provider.   
+The SM uses a library of space provider adaptors to communicate with the space providers. Currently the SF supports 3 space providers KCP, KIND, and KubeFlex, and includes 3 space provider adaptors that the SM uses.
+
+### Space provider adaptors
+The space manager uses a set of space provider adaptors that interacts with the space provider. All space provider adaptors implement a simple "space client interface" that includes basic pSpace life cycle operations. In addition the interface also defines the events that are sent from the adaptors to the space manager.
+
+### Space aware client (SAC)
+The Space aware client (SAC) allows clients/controllers to easily get access to the underlying space by simply using the space name. There is no direct interaction between the SAC and the Space  provider, and therefore the SAC is transparent to the specific provider of the spaces.
```
MikeSpreitzer, line 32: "simply using the space name" is not accurate, the `providerNS string` (which is actually specific to not just the provider but, more specifically, the SpaceProviderDesc --- right?

space-framework/docs/space-framework.md
MikeSpreitzer: If "SpaceProviderDesc" is synonymous with "space provider" then you have answered the question but invalidated the earlier statement that KIND is an example of a space provider.

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+**Main features**  
+- Constantly watch for changes in the available spaces. The SAC is using an informer on Space objects using the SF Space client-go APIs.
+    - Holds a cache of the pSpace access info for each Space 
+- Exposes utility functions to get RestConfig by simply specifying the space name, and optionally the provider name (if the provider is not specified the default provider is being used). The namespace where the Space resides can be derived from the name of the provider.  
+
+Currently the SAC exposes a simple function that that supplies a rest.Config data structure that gives access to a physical space with the identity of a user that is authorized to do everything. In the future we will add a SAC space aware ClientSets
+
+## Architecture details 
+When a new SpaceProviderDesc is created the SM creates a new namespace that will be used to host all the Space objects representing pSpaces on that space provider.  
+The name of the generated namespace will be `paceprovider-<provider name>`
```
MikeSpreitzer, line 45: Missing "s" at the start of that generated name?

space-framework/docs/space-framework.md
MikeSpreitzer: Ah, perhaps you are using "the pSpace identifier used by the space provider" to mean some particular technical thing that I am not aware of.  It is an identifier in the interface between Space Manager and provider adapter?

space-framework/docs/space-framework.md
MikeSpreitzer: Perhaps replace that second sentence with something like the following.

> The framework defines an abstraction layer that is a common interface to space functionality. Clients of the abstraction layer interact with it rather than directly with the space providers. The abstraction layer's interface is a state-based interface expressed in Kubernetes API objects.

space-framework/docs/space-framework.md
MikeSpreitzer: I still see two copies of "The" here.

space-framework/docs/space-framework.md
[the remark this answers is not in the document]
MikeSpreitzer: Better now, thanks.

space-framework/docs/space-framework.md
MikeSpreitzer: continued in a new comment

space-framework/docs/space-framework.md
MikeSpreitzer: continued in a new comment.

space-framework/docs/space-framework.md
MikeSpreitzer: If it is allowed for the Secret to be created later and after that the relevant controller will react and proceed, that is the normal pattern and hardly needs comment in a high-level doc; certainly this is not a _requirement_ that the Secret be created first. If the only deviation from this pattern is that the controller logs an error message rather than a "waiting..." message while waiting, then I suggest fixing the logging to make a totally normal controller.

space-framework/docs/space-framework.md
ezrasilvera: SpaceProviderDesc is the object/CRD that holds information about the space provider itself.

space-framework/docs/space-framework.md
ezrasilvera: No, it is identifyer between the adaptor and the space provider itself (or the iterface it exposes) For example, for KIND provider the KIND cluster name is the id of the pSpace. and as stated above currently the space name will be used as the KIND cluster name

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+  2. Providerless spaces: This spaces don't belong to a space provider and are called "Imported spaces". Such Space is created by importing an existing pSpace. In this case the Space is not linked to any space provider and the client which imports that Space need to supply the access information that allows other clients to connect to the pSpace. For imported spaces the desired state is derived by the pSpace. This means that the SF is not responsible for the life cycle management of the pSpace but only update the corresponding Space object. For example, when deleting the space object of an existing imported pSpace the space framework will recreate the space object. 
+### Space Manager 
+The Space Manager (SM) is a Kubernetes controller that is responsible for maintaining the state of the Space objects. The SM reconciles the Space and SpaceProviderDesc objects, and monitors the state of the pSpaces through the space provider.
```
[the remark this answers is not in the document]
ezrasilvera: I'm not sure what are the other "componenets" in the diagram that you think shoud be under tge umbrella of "Space Manager"

space-framework/docs/space-framework.md
MikeSpreitzer: I am not sure I understand that answer. If a user deletes a Space object, does that undo the import? Your earlier text seems to say "no, the SM will re-create the Space object". Which is it? Are there other cases where the SM updates the Space object? If the SM will re-create an imported Space object after it is deleted, why was the SM keeping all the needed information somewhere else?

space-framework/docs/space-framework.md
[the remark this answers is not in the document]
MikeSpreitzer: There is a process that has the generic (in the plain English sense of the word, not as in golang "generics") Space Manager code PLUS the provider adapters, right? Do you consider the provider adapters to be part of the Space Manger? The picture does not show it that way.

space-framework/docs/space-framework.md
MikeSpreitzer: In that case some other text is wrong, the text that says KIND is an example of a space provider. In fact KIND is an example of a SpaceProviderType, _not_ a provider (SpaceProviderDesc).

space-framework/docs/space-framework.md
MikeSpreitzer: In the kcp case, what is "the pSpace identifier used by the space provider"? kcp identifies its spaces in two ways: `logicalcluster.Name` and workspace pathname. So I find an assertion that implies there is only to be confounding.

space-framework/docs/space-framework.md
MikeSpreitzer: There can be two SpaceProviderDesc with the same SpaceProviderType, right?

space-framework/docs/space-framework.md
ezrasilvera: OK, let me clarify. There is a difference between Import (providerless) and unmanaged spaces (using discovery over a provider):
For Imported spaces - deletion of the space object is basically an "undo import" the Space object will not be re-created
For unmanaged spaces - if the Space is deleted, the SM will recreate it if the pSpace exists. 
I think I indeed made a mistake in the description of Imported spaces. Will fix

ezrasilvera:
> Is this intended to be user-facing documentation showing the simplest possible usage scenario, or architecture documentation showing all the modularity?

It's HL architecture.  A simple "user guide" will be published as separate document.. 

> I thought the cluster that hosts the abstraction layer does not have to be the same as the cluster that does any particular other thing (e.g., host the KubeStellar controllers).

It is. There shouldn't a reference to KubeStellar in this document, or to any specific usage of the SF.

space-framework/docs/space-framework.md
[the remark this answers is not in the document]
ezrasilvera: It is part of the process, and I thought the picture shows that (I'll add text to emphasis this fact)

space-framework/docs/space-framework.md
ezrasilvera: Yes. However, The assumption is that we will have one SpaceProviderDesc per space provider instance. So you can have to KCP servers running and create 2 SpaceProviderDesc objects to represent them . Both will be with the same "provider type" of KCP. 
In theory , nothing prohibit you to have 2 SpaceProviderDesc  pointing to the SAME space provider instance, there is a potential of management collision  when a single space is being managed through 2 different SpaceProviderDesc  and basically represented as 2 different Spaces

space-framework/docs/space-framework.md
ezrasilvera: I'm not sure I understand the issue here. I don't see any reference to KIND here. Maybe it's outdated text ?

space-framework/docs/space-framework.md
ezrasilvera: for KCP we use th eWorkspace path name

space-framework/docs/space-framework.md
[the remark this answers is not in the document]
MikeSpreitzer: I figured out that the provider adapters are in the same process as the generic Space Manager code. My question is whether (a) the term "Space Manger" refers ONLY to the generic Space Manger code or (b) the term "Space Manger" refers to the process that contains the generic code PLUS the provider adapters.

space-framework/docs/space-framework.md
MikeSpreitzer: Please do not mark a conversation as "resolved" without a clue about _why_ it is being marked as resolved.

space-framework/docs/space-framework.md
MikeSpreitzer: I was not asking about the case of two SpaceProviderDesc referring to the same provider, I was asking about the case of two SpaceProviderDesc referring to two providers of the same _type_. Since you seem to be confirming that this is allowed, that means you have a conceptual mismatch in the text that says that "kcp" --- which the reader will interpret as referring to the product from the kcp repo --- is an example of a provider (i.e., SpaceProviderDesc); instead "kcp" (the product) is an example of a SpaceProviderType, and multiple SpaceProviderDesc may refer to the same product.

space-framework/docs/space-framework.md
MikeSpreitzer: A kcp workspace pathname is a structured thing in the abstract, with a string representation using colon (`:`) to separate parts of the path. Are you saying that when using a kcp provider, the Space object name will be a kcp workspace pathname, such as "root:espw"?

MikeSpreitzer:
There needs to be a CLI for dealing with spaces, analogous to the kcp CLI for dealing with workspaces. There should be only one copy/version of this CLI, suitable for use in the core container and by contributors deploying the central components as bare processes. This CLI should be documented, at least for contributors.

The work I see so far follows a direction I like: rather than modifying a kubeconfig file like `kcp ws` does, instead make new kubeconfig files. This avoids the problems with modifying files, and fits nicely together with the documentation style recently adopted (@clubanderson , @eaepstein) of explicitly writing the kubeconfig file being used on each command line.

We should make each command consume only one kubeconfig file, for simplicity and familiarity. For commands that inherently manipulate just one pre-existing space, this would be the kubeconfig file for that space. For commands that create or delete a space (conditionally or otherwise), this would be the kubeconfig for the space that holds the Space objects --- and such commands that may create a space should have an option for writing a kubeconfig file for that space.

For example, what about the following CLI?

`kubectl sm ensure space $space_name [--provicer $desc_name] [-o $space_kubeconfig]`

`kubectl sm remove space $space_name [--provider $desc_name]`

MikeSpreitzer:
@ezrasilvera:

> There shouldn't a reference to KubeStellar in this document, or to any specific usage of the SF

So it seems a bit odd to talk about the "core" cluster or space. I think "hosting" is the more apt term.

space-framework/docs/space-framework.md
MikeSpreitzer: Why was this marked as resolved?

space-framework/docs/space-framework.md
MikeSpreitzer: I still see "KIND is an example for Space Provider" in the Terminology section entry for "Space Provider".

space-framework/docs/space-framework.md
MikeSpreitzer: Why was this marked as resolved?

space-framework/docs/space-framework.md
MikeSpreitzer: Why was this marked as resolved?

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,92 @@
...
+   * **Managed Space**: The pSpace is created through an explicit request to the management api server to create a new Space. For Managed spaces the space framework initiates a request to the Space provider to create the pSpace, the space frame work will then continuously reconcile between the Space and the pSpace states. For managed spaces the desired state is controlled through the Space object. For example, in order to delete the pSpace the client need to delete the corresponding Space object and the SF will interact with the space provider in order to delete the pSpace.
+   * **Unmanaged Space**: The SF supports discovery of existing pSpaces on a space provider and adding them as "Unmanaged Spaces" into the framework. This mode is similar to the import mode in terms of the "source of truth" - the desired state is defined by the pSpace. However, as oppose to import, here the pSpaces are automatically discovered through the space provider regular interface and the pSpace access information is retrieved from the  space provider similarly to how it is done in the managed space case. The SF supports pattern based filtering rules to limit the discovery of pSpaces.
```
MikeSpreitzer, line 22: s/oppose/opposed/

space-framework/docs/space-framework.md
MikeSpreitzer: Yes, please do.

MikeSpreitzer, line 23: s/that Space need to/that Space needs to/

space-framework/docs/space-framework.md
MikeSpreitzer: Why was this marked as resolved?

space-framework/docs/space-framework.md
MikeSpreitzer: Why was this marked resolved?

MikeSpreitzer: https://github.com/kubestellar/kubestellar/blob/fecfae8b77db8aaad965ea7032fc2295abaedaa0/space-framework/pkg/msclientlib/client.go#L52 does _not_ show an optional provider name, it shows an unconditional provider namespace name.

space-framework/docs/space-framework.md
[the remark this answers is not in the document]
ezrasilvera: I agree there is no good separation throughout the doc between the provider adaptor and the SM. I’ll review it.

space-framework/docs/space-framework.md
ezrasilvera: I'm not sure I understand the issue. For the space provider we named the "providerType" in the same name as the name of the product which is also the actual provider backend. Maybe I should have mentioned that there can be multiple provider backend instances represented by multiple providerDesc  and they all of type KCP

space-framework/docs/space-framework.md
ezrasilvera: I agree. I tend to resolve only when I changed/fix the exact proposal in the comments (mainly spelling mistakes). If that was not the case this is a mistake

ezrasilvera: Regarding the markdown comment - I wasnt sure what you meant here

space-framework/docs/space-framework.md
ezrasilvera: I wasn't accurate. The space name isused as the workspace object's name  (in meta) .

space-framework/docs/space-framework.md
ezrasilvera: Because it was fixed. Sometimes git still shows old code in the browser. not sure why.

ezrasilvera: last week commit 19ec68eff83b5d829c7a0d2158f3efeb43a98970

space-framework/docs/space-framework.md
ezrasilvera: fixed last week commit 19ec68eff83b5d829c7a0d2158f3efeb43a98970

space-framework/docs/space-framework.md
ezrasilvera: Already did, last week. Hopefully it's OK

space-framework/docs/space-framework.md
ezrasilvera: Because I changed to "by simply using the space name and optionally the SpaceProviderDesc name"

space-framework/docs/space-framework.md
MikeSpreitzer: I mean that what I see here currently is `**Main features**` and I think this intended to be the heading of a subsection nested under `### Space aware client (SAC)`. So my question is why not continue with regular markdown for section headings and use `#### Main features`?

space-framework/docs/space-framework.md
MikeSpreitzer: My point is that this design has one extra level of multiplicity that is not properly introduced and consistently accounted for, it only appears in passing in some places. The English text often talks about the concept of a "space provider", and the Terminology section defines that term and says that "KIND" is an example of a "space provider". But in the places where the precision of code is used, there are *two* concepts --- SpaceProviderDesc and SpaceProviderType --- and they are different. "KIND" is an example of a SpaceProviderType, and yet you also want to treat SpaceProviderDesc as equivalent to "space provider".

space-framework/docs/space-framework.md
MikeSpreitzer: You finally stated the missing information for the kcp provider adapter. It is a choice made by that provider adapter, about the relation it maintains between Space object name and the ways that kcp identifies its logicalclusters. My point is that in general this is a question that has to be answered (i.e., documented) for each provider adapter: the relation it maintains between Space object name and the ways that the provider identifies its native cluster-ish thingies.

space-framework/docs/space-framework.md
MikeSpreitzer: But the second parameter is actually (a) the provider _namespace_ name and (b) not optional (as far as the reader of the interface sees).

space-framework/docs/space-framework.md
MikeSpreitzer: I see that the spelling has been fixed.

space-framework/docs/space-framework.md
MikeSpreitzer: In https://github.com/ezrasilvera/edge-mc/blob/bccd082b5f4ad29740a668e8bb987bc0d18bde4e/space-framework/docs/space-framework.md?plain=1 I still see "For example, when deleting the space object of an existing imported pSpace the space framework will recreate the space object".

space-framework/docs/space-framework.md
MikeSpreitzer: thanks

space-framework/docs/space-framework.md
MikeSpreitzer: thanks

MikeSpreitzer:
Also, `KubestellarSpaceInterface` has a second method now, named `ConfigForSpaceByExportRef`.

Also, that is not quite explained well enough. What exactly is the `ref string` that it is given?

space-framework/docs/space-framework.md
ezrasilvera: Correct. I'll add a section on "Provider Adaptors"  And will put there the details for a specific provider adaptor

space-framework/docs/space-framework.md
ezrasilvera: Will review the doc again. I agree that for space we have distinct pSpace Vs Space, where on the backend we have space provider, space provider object (SpaceProviderDesc) and space provider adaptor (and space provider type)

space-framework/docs/space-framework.md
ezrasilvera: Will update the interface in the doc. To you request we added a another interface that uses the default provider so you don't need to specify it. We pass the provider name, there is a fixed rule that relates the providerNS and the provider name

MikeSpreitzer:
In reviewing the code, I discovered that each Space has _two_ kubeconfigs for clients, one for clients in the space hosting cluster and one for clients outside that cluster. This document should make that factor clear. And since the msclientlib does NOT have that factor, it is only usable within the space hosting cluster; this is an architectural restriction that should be made clear.

ezrasilvera:
> There needs to be a CLI for dealing with spaces, analogous to the kcp CLI for dealing with workspaces. There should be only one copy/version of this CLI, suitable for use in the core container and by contributors deploying the central components as bare processes. This CLI should be documented, at least for contributors.
> 
> The work I see so far follows a direction I like: rather than modifying a kubeconfig file like `kcp ws` does, instead make new kubeconfig files. This avoids the problems with modifying files, and fits nicely together with the documentation style recently adopted (@clubanderson , @eaepstein) of explicitly writing the kubeconfig file being used on each command line.
> 
> We should make each command consume only one kubeconfig file, for simplicity and familiarity. For commands that inherently manipulate just one pre-existing space, this would be the kubeconfig file for that space. For commands that create or delete a space (conditionally or otherwise), this would be the kubeconfig for the space that holds the Space objects --- and commands that may create a space should have an option for writing a kubeconfig file for that space.
> 
> For example, what about the following CLI? The following commands would all consume a kube client config for the space holding the Space objects, accepting all the usual `kubectl` ways of telling it what kube client config to use.
> 
> `kubectl sm space ensure $space_name [--provider $desc_name] [-o $space_kubeconfig]`
> 
> `kubectl sm space remove $space_name [--provider $desc_name]`
> 
> `kubectl sm space list [$space_name] [--provider $desc_name]`
> 
> `kubectl sm provider list`

@MikeSpreitzer Seems reasonable  my only concern is with the . `ensure `     , if indeed we want a single function that will also act as the "get config if exists"  , we would need to add a Boolean  of `--create-if-missing`

ezrasilvera:
> In reviewing the code, I discovered that each Space has _two_ kubeconfigs for clients, one for clients in the space hosting cluster and one for clients outside that cluster. This document should make that factor clear. And since the msclientlib does NOT have that factor, it is only usable within the space hosting cluster; this is an architectural restriction that should be made clear.

The SAC (mscientlib) does support external or internal. This code is still not documented here . Will update

MikeSpreitzer:
> my only concern is with the ensure , if indeed we want a single function that will also act as the "get config if exists" , we would need to add a Boolean of --create-if-missing

Well "ensure" is all about "create if missing", so it would be odd to juxtapose those names. I think it would be better to have a distinct command, something like the following.

```
kubectl sm space get-kubeconfig $space_name [--provider $desc_name] [--internal $bool] -o $space_kubeconfig
```

space-framework/docs/space-framework.md
ezrasilvera: Oh I see. Changed to markdown

ezrasilvera:
> Is this intended to be user-facing documentation showing the simplest possible usage scenario, or architecture documentation showing all the modularity?
> 
> I thought the cluster that hosts the abstraction layer does not have to be the same as the cluster that does any particular other thing (e.g., host the KubeStellar controllers).

It suppose to be a general Doc. The SM hosting cluster doesn't have to be any specific cluster. Where did you see a reference to KS ?

ezrasilvera:
> The SAC (mscientlib) does support external or internal. This code is still not documented here as it's still not exposed through the interface API (it is supported already by the internal implementation
> `func (mcc *multiSpaceClient) getRestConfigFromSecret(internalAccess bool, space *spacev1alpha1.Space) (*rest.Config, error)`

Added information about the 2 secrets

space-framework/docs/space-framework.md
```diff
@@ -0,0 +1,146 @@
...
+The KCP space provider exposes KCP's workspaces as the pSpaces. The KCP SPA interacts with the KCP server to manage these workspaces. 
+We don't support workspace hierarchy and all workspaces are created under the ```root``` workspace and therefor the KCP SPA uses the Space name as the unique identifier for the workspace.  
+Example: When the user creates a Space with name ```mySpace``` the SPA will create a workspace named ```mySpace``` with path of ```root:mySpace```
+
+### KIND SPA
+
+#### Example: Create a Space on the KCP space provider
+Suppose we created a SpaceProviderDesc of type KCP named sp-kcp1, and we now want to create a Space. As mentioned before when the SpaceProviderDesc was created the SM also created a corresponding SPA instance. 
+1. Create a Space object with a reference to sp-kcp1
+2. The SM finds the SPA the corresponding SPA
```
MikeSpreitzer, line 141: I am having trouble parsing this sentence.

MikeSpreitzer (commented):
This needs more work; it can be done in follow-on PRs.

/lgtm
/approve

Issue #1208, opened 2023-10-26
Replace KCP-bind with kube-bind


ezrasilvera:
The main goal of this item is to replace KCP bind with kube-bind, transparently so KS is still fully functional (using KCP workspaces)
We can separate into two different sections/PRs

This work will be done through the branch:  https://github.com/kubestellar/kubestellar/tree/kube-bind

1. Use kube-bind, continue using the logical-cluster ID when processing the List/Watch objects comings from the Export
2. Remove the logical-cluster id with space-name coming from an internal mapping of the kube-bind annotation & space name

###Phase1:  Use kube-bind  PR:  https://github.com/kubestellar/kubestellar/pull/1265
1. the KS container image (used by all KS containers) should include the KB needed executables 
2. Setup scripts that build the ESPW, WDS, and IS should be changed to:
    * Run dex, 
    * ESPW: Run backend engine + import all KS CRDs + add KB export label + import the KB CRDs
    * IS, WDS: Run konnector, bind all KS resources
3. Mailbox controller - create bind for new MBs (can call script) - Note the above is done in the mb-controller container
4. Need to replace the check on all controllers that verify the export exists
Note: some of the KB components may reside on different containers (e.g., dex and konnectors for MBs) - need to verify access

### Phase2: Stop using the cluster-id
Background:
KCP objects have special annotation("kcp.io/cluster") that holds a reference to the cluster. From that reference is is also possible to obtain the workspace. This annotation is set by KCP and will not be available in other providers.
For example, when we use a watch or List over an APIExport using kube-bind, we get a list of objects and such annotation is the only way to know which cluster the object belongs to.

1. During bind generate a map of [provider NS for the consumer : space name] 
2. Change the relevant controllers code

Notes regarding SM integration:
1. When spaces are created through the SM, the SM will start the konnector process & (optionally) the backend engine process
2. When the space-manager will be used, the mapping of the provider NS of the consumer will be added to the space annotation

MikeSpreitzer:
The two steps enumerated in the opening comment definitely belong in distinct PRs. The first step has no dependency on the space framework and can be taken before any introduction of the space framework. The second step pre-reqs use of the space framework, and can be delayed until some provider other than kcp is used.

clubanderson:
@ezrasilvera can this be closed?

PR #1254, opened 2023-11-06
✨ Invocation of kube-bind script from the mailbox-controller

waltforme:
## Summary
This PR finishes item 3 in Section "Use kube-bind" of #1208.

/cc @ezrasilvera
/cc @andreyod

### DIFF
--- a/cmd/mailbox-controller/controller.go
+++ b/cmd/mailbox-controller/controller.go
@@ -19,6 +19,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"os/exec"
 	"strings"
 
 	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


cmd/mailbox-controller/controller.go
```diff
@@ -252,10 +253,18 @@ func (ctl *mbCtl) sync(ctx context.Context, refany any) bool {
...
-	// ISSUE kube-bind objects(apiservicebindings) does not have clusterclientset like in KCP. Maybe dynamicClusterClient can help.
+	// The script ensures that a mailbox workspace has the needed KubeStellar CRDs.
+	shellScriptName := "kubectl-kubestellar-ensure-mbw"
 
-	// TODO bind all resources. Note: running in pod, need to access the dex
+	// The script is idempotent.
+	logger.V(2).Info("Executing shell script", "script", shellScriptName)
+	cmdLine := shellScriptName + " " + workspace.Name
+	cmd := exec.Command("/bin/sh", "-c", cmdLine)
+	_, err := cmd.Output()
```
andreyod, line 263: The script should provide simple, easy to parse output, in order to be able to map provider namespace for this space and the space name(see TODO in line 269 below).
Something like: "Completed. Service provider namespace is:<ns for this MBW>"

waltforme: Thank you. Exactly, because of this, I didn't touch line 269 in this PR, so it can continue to remind me there is parsing to do here.

I am thinking finishing item 3 in Section "Use kube-bind" of https://github.com/kubestellar/kubestellar/issues/1208 in this PR, and then support "Stop using the cluster-id" Section in the next PR. @andreyod please let me know if we want to also do the parsing here.

andreyod: I think the parsing should be in another PR, together with other changes for "Stop using the cluster-id" that we will need to do after we start using space framework(we are not there yet). But  `kubectl-kubestellar-ensure-mbw` can provide the needed output in this PR.

waltforme: I got the point. But if we go with that way, we will have to do the parsing twice. First we'll parse the stdout/stderr of kubectl-bind in the script, second we'll parse the output of the script in mailbox-controller go code.

I was planning to do the parsing only once and for all, in mailbox-controller go code. What do you think?

ezrasilvera: I think we should postpone that to another PR AFTER we have merged the space framework. The integration would be
1. Create the space through SM
2. Get the space config info from the SM
3. Execute bind using the above config
4. Parse the output to get the NS on the export provider 
5. Set the Space annotation with the info in (4) 
In all the scripts the above will be done by the script (including bring up konnector process) 

In the MB case 1,2 and 3 are done from code. for 4 & 5 we have 2 options:
1. Do (4) & (5)  in code  in order to do (4) the exec need to get all output of the bind, and (5) will be done directly through client-go
2. Have the script do 3, 4 & 5

waltforme: Option 1 makes more sense to me.

PR #1257, opened 2023-11-07
✨ Make test user quickstart exercise ServiceAccount

### LINKED ISSUE #1060: feature: CI needs to test deployment of KubeStellar core in a kube cluster
### Feature Description

We need the regular testing for every PR to include checking that the result of the merging the PR is a KubeStellar version that can be deployed as workload in a Kubernetes cluster and pass an end-to-end test.

This particular test will necessarily include building the container image(s) involved, arranging that those will be the ones used in the remainder of the test, creating a Kubernetes cluster to deploy the core into, and creating a WEC to exercise.

### Proposed Solution

Write a test that is like example1 but follows the deploy-as-workload path and first builds the core image(s) that will be used.

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR extends the test user quickstart to exercise a downsynced ServiceAccount object.

This PR also replaces the use of brew to get user executables with building the local sources and fetching a patched kcp release archive. This is the last step in the campaign to address #1060 .

This PR also reacts to a change snuck in through an earlier update to `go.mod`, advancing the required version of `klog/v2` (from v2.70.1 to v2.100.1) through a big change (https://github.com/kubernetes/klog/pull/375) in the way it logs `any` values. Now they are logged by marshaling as JSON! Previously those values were rendered with `fmt.Sprintf("%v", .)`. This klog change means that every private field of a struct being logged is now invisible in the log. This PR changes many structs to make their fields exported (i.e., the field name is capitalized). This PR also adds the method needed to make `WorkloadParts` a `json.Marshaler`.

## Related issue(s)

Fixes #1060

### DIFF
--- a/docs/content/Getting-Started/user-quickstart-test.md
+++ b/docs/content/Getting-Started/user-quickstart-test.md
@@ -110,25 +110,15 @@ qs_sort: test
 %}
 
 #### 2. Install <span class="Space-Bd-BT">KUBESTELLAR</span>'s user commands and kubectl plugins
-!!! tip ""
-    === "install"
-         {%
-           include-markdown "../common-subs/brew-install.md"
-           start="<!--brew-install-start-->"
-           end="<!--brew-install-end-->"
-         %}
-    === "remove"
-         {%
-           include-markdown "../common-subs/brew-remove.md"
-           start="<!--brew-remove-start-->"
-           end="<!--brew-remove-end-->"
-         %}
-    === "uh oh, no brew?"
-         {%
-           include-markdown "../common-subs/brew-no.md"
-           start="<!--brew-no-start-->"
-           end="<!--brew-no-end-->"
-         %}
+
+```shell
+pwd
+rm -f bin/*
+make userbuild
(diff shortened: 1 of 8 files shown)


MikeSpreitzer:
I canceled the "docs-ecutable - User Quickstart Test" because it looked like it was waiting forever for something that was not going to appear. This needs to be debugged.

MikeSpreitzer:
The force-push to 5bcbda5b3 is a rebase plus debugging work.

MikeSpreitzer:
/cc @clubanderson 
/cc @dumb0002 
@yana1205

MikeSpreitzer:
I just realized that the user-quickstart-test continues with docs-ecutable content in its second part, using an existing core. This second part uses brew. That is acceptable right now, because the second part does not exercise features that are absent from the user executables in the brew repository.  Also, the discussion of using `kubestellar-list-syncing-objects` is currently only hypothetical and only in the second part of this doc. It should be promoted to something tested in the first part. But let's leave that for a follow-on PR.

PR #1258, opened 2023-11-07
🐛 Make core.Dockerfile accept advance in golang version

MikeSpreitzer:
## Summary
This PR changes `core.Dockerfile` to suppress the golang version check in the `make innerbuild`. This is needed because RedHat recently updated what `dnf` in the `redhat/ubi9` image finds when asked to install `golang`, from release 1.19.something to 1.20.10, causing the aforementioned version check to fail.

This PR also makes a few improvements in building and testing: add the temp script generated by docs-ecutable to .gitignore; adjust the way the syncer image is recursively built to work more broadly.

## Related issue(s)

Fixes #

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ coverage.*
 tools
 build
 .vscode
+docs/scripts/generated_script.sh
 
 # JetBrains IDE files
 .idea/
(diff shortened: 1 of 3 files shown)


MikeSpreitzer:
/cc @waltforme 
/cc @francostellari 
@clubanderson

.gitignore
```diff
@@ -4,6 +4,7 @@ coverage.*
 tools
 build
 .vscode
+docs/scripts/generated_script.sh
```
ezrasilvera, line 7: Who is generating this?  What's the relation to golang version?

MikeSpreitzer: the docs-ecutable metascript

PR #1169, opened 2023-10-19
:bug: Physical Cluster -> Workload Execution Cluster swap

namasl:
## Summary
Updates old terminology `pcluster` with `wec`, and "physical cluster" with "workload execution cluster" or "WEC".

## Related issue(s)

Fixes #1071

PR #1259, opened 2023-11-07
✨ updated verify go version script

nirrozenbaum:
this change is checking go version is at least 1.19 instead of strictly eqaulity.

## Related issue(s)

Fixes #

### DIFF
--- a/hack/verify-go-versions.sh
+++ b/hack/verify-go-versions.sh
@@ -17,14 +17,19 @@
 set -e
 set -o pipefail
 
-VERSION=$(grep "go 1." go.mod | sed 's/go //')
+MINIMAL_VERSION=$(grep "go 1." go.mod | sed 's/go //')
 
-# grep "FROM golang:" Dockerfile | { ! grep -v "${VERSION}"; } || { echo "Wrong go version in Dockerfile, expected ${VERSION}"; exit 1; }
-# grep -w "go-version:" .github/workflows/*.yaml | { ! grep -v "go-version: v${VERSION}"; } || { echo "Wrong go version in .github/workflows/*.yaml, expected ${VERSION}"; exit 1; }
+# grep "FROM golang:" Dockerfile | { ! grep -v "${MINIMAL_VERSION}"; } || { echo "Wrong go version in Dockerfile, expected ${MINIMAL_VERSION}"; exit 1; }
+# grep -w "go-version:" .github/workflows/*.yaml | { ! grep -v "go-version: v${MINIMAL_VERSION}"; } || { echo "Wrong go version in .github/workflows/*.yaml, expected ${MINIMAL_VERSION}"; exit 1; }
 # Note CONTRIBUTING.md isn't copied in the Dockerfile
 # if [ -e CONTRIBUTING.md ]; then
-#   grep "golang.org/doc/install" CONTRIBUTING.md | { ! grep -v "${VERSION}"; } || { echo "Wrong go version in CONTRIBUTING.md expected ${VERSION}"; exit 1; }
+#   grep "golang.org/doc/install" CONTRIBUTING.md | { ! grep -v "${MINIMAL_VERSION}"; } || { echo "Wrong go version in CONTRIBUTING.md expected ${MINIMAL_VERSION}"; exit 1; }
 # fi
+
 if [ -z "${IGNORE_GO_VERSION}" ]; then
-  go version | { ! grep -v go${VERSION}; } || { echo "Unexpected go version installed, expected ${VERSION}. Use IGNORE_GO_VERSION=1 to skip this check."; exit 1; }
+  ENV_VERSION=$(go version | awk '{print $3}' | sed 's/go//')
+  if [[ "$ENV_VERSION" < "$MINIMAL_VERSION" ]]; then
+    echo "Unexpected go version installed. expected minimal go version ${MINIMAL_VERSION} while your environment has version ${ENV_VERSION}. Use IGNORE_GO_VERSION=1 to skip this check."
+    exit 1
+  fi
 fi
--- a/hack/verify-go-versions.sh
+++ b/hack/verify-go-versions.sh
@@ -32,13 +32,17 @@ if ! [ -x "$(command -v go)" ]; then # validate go is installed
 fi
 
 if [ -z "${IGNORE_GO_VERSION}" ]; then # validate go version is sufficient
-  MINIMAL_VERSION_MAJOR=$(echo ${MINIMAL_VERSION} | awk -F '.' '{printf $1}')
-  MINIMAL_VERSION_MINOR=$(echo ${MINIMAL_VERSION} | awk -F '.' '{printf $2}')
+  REQUIRED_VERSION_MAJOR=$(echo ${MINIMAL_VERSION} | awk -F . '{printf $1}')
+  MINIMAL_VERSION_MINOR=$(echo ${MINIMAL_VERSION} | awk -F . '{printf $2}')
   ENV_VERSION=$(go version | awk '{print $3}' | sed 's/go//')
-  ENV_VERSION_MAJOR=$(echo ${ENV_VERSION} | awk -F '.' '{printf $1}')
-  ENV_VERSION_MINOR=$(echo ${ENV_VERSION} | awk -F '.' '{printf $2}')
+  ENV_VERSION_MAJOR=$(echo ${ENV_VERSION} | awk -F . '{printf $1}')
+  ENV_VERSION_MINOR=$(echo ${ENV_VERSION} | awk -F . '{printf $2}')
 
-  if [ ${ENV_VERSION_MAJOR} -lt ${MINIMAL_VERSION_MAJOR} ] || [ ${ENV_VERSION_MINOR} -lt ${MINIMAL_VERSION_MINOR} ]; then
+  if [ "${ENV_VERSION_MAJOR}" != "${REQUIRED_VERSION_MAJOR}" ]; then
+    echo "Unexpected go version installed. expected go version major ${REQUIRED_VERSION_MAJOR}, while your environment has version ${ENV_VERSION}. Use IGNORE_GO_VERSION=1 to skip this check."
+	  exit 1
+  fi
+  if ! [ "${ENV_VERSION_MINOR}" -ge "${MINIMAL_VERSION_MINOR}" ]; then
 	  echo "Unexpected go version installed. expected minimal version ${MINIMAL_VERSION}, while your environment has version ${ENV_VERSION}. Use IGNORE_GO_VERSION=1 to skip this check."
 	  exit 1
   fi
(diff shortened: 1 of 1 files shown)


hack/verify-go-versions.sh
```diff
@@ -17,14 +17,19 @@
...
 # Note CONTRIBUTING.md isn't copied in the Dockerfile
 # if [ -e CONTRIBUTING.md ]; then
-#   grep "golang.org/doc/install" CONTRIBUTING.md | { ! grep -v "${VERSION}"; } || { echo "Wrong go version in CONTRIBUTING.md expected ${VERSION}"; exit 1; }
+#   grep "golang.org/doc/install" CONTRIBUTING.md | { ! grep -v "${MINIMAL_VERSION}"; } || { echo "Wrong go version in CONTRIBUTING.md expected ${MINIMAL_VERSION}"; exit 1; }
 # fi
+
 if [ -z "${IGNORE_GO_VERSION}" ]; then
-  go version | { ! grep -v go${VERSION}; } || { echo "Unexpected go version installed, expected ${VERSION}. Use IGNORE_GO_VERSION=1 to skip this check."; exit 1; }
+  ENV_VERSION=$(go version | awk '{print $3}' | sed 's/go//')
+  if [[ "$ENV_VERSION" < "$MINIMAL_VERSION" ]]; then
```
MikeSpreitzer, line 31: I think it is safer to insist on a positive outcome from test for an acceptable relationship than accept a negative outcome from a test for an unacceptable relationship --- the latter can accept unexpected failures.

MikeSpreitzer:
The nit-pickers are complaining because the "title" of this PR does not follow the rules stated in https://github.com/kubestellar/kubestellar/blob/main/.github/pull_request_template.md?plain=1 .

MikeSpreitzer:
If the change in this PR is made, and someone using go 1.20 or 1.21 does a `make codegen`, will that produce code that the repo's CI checking will reject? You could test that.

nirrozenbaum:
> If the change in this PR is made, and someone using go 1.20 or 1.21 does a `make codegen`, will that produce code that the repo's CI checking will reject? You could test that.

@MikeSpreitzer I'm sorry if the question is basic. my team is new in the project and I'm still not very familiar with the CI checking used in KS. 
how can I verify that CI checking will not reject the produced code? is there a way to manually test it?

hack/verify-go-versions.sh
nirrozenbaum: I tend to disagree with this comment.

if you had to write a function with your custom logic and had to verify a condition before starting, how would you write your function?

image

Or:
image

of course it is a matter of code styling and both options will work, but best practices definitely mention the first option as the preferred approach. it reduces code indentation which results in a more readable and maintainable code and it is easier for understanding.

another point - the original code of this script was checking negative test (pay attention for the NOT) so I'm not sure what exactly has changed. 
also adding here another example from a well known CNCF incubation project, KubeEdge, which also checks go version and does it using negative test. see example here:
https://github.com/kubeedge/kubeedge/blob/master/hack/lib/golang.sh#L35-#L43.

I can easily add a check that verifies go is installed before the versions comparison if you think it's safer.
other than that, I don't really see which unexpected failures may happen and I'm happy to hear if you have any example.

MikeSpreitzer: My point is not about indentation (you can get the indentation you like with either sort of test by appropriate engineering on the script) but about the ways that a command in `bash` can have a non-0 exit code. Bash has ways to fail that you typically do not see in programming languages like Go. In the case at hand, what if `$MINIMAL_VERSION` did not get set to the expected string but rather the empty string?

In the older script, the test `grep -v go${VERSION}` is testing for the acceptable condition; any failure to demonstrate the acceptable condition is treated as a problem.

MikeSpreitzer:
@nirrozenbaum : here is a way that you could check the scenario I outlined, using go 1.20.

```
git checkout patch-1
git checkout -b patch-1-test
make codegen
git commit -a
```

The push your new `patch-1-test` branch to your fork and open a PR based on it; the regular CI testing will get applied to that new `patch-1-test` branch.

nirrozenbaum:
> @nirrozenbaum : here is a way that you could check the scenario I outlined, using go 1.20.
> 
> ```
> git checkout patch-1
> git checkout -b patch-1-test
> make codegen
> git commit -a
> ```
> 
> Then push your new `patch-1-test` branch to your fork and open a PR based on it; the regular CI testing will get applied to that new `patch-1-test` branch.

as I mentioned in previous comment, I tested using go 1.20 by running `make codegen`. 
no file has changed (also tried to delete whatever is under `pkg/client/*` before running make). 
since nothing has changed I have nothing to commit. 

I guess that means that the answer to your original question is - CI checking should work exactly as today, cause the make command produces the exact same files.

hack/verify-go-versions.sh
nirrozenbaum: That is incorrect.
In older script, if `{VERSION}` is empty, the script will return successfully which is a problem (is it?).
I think the main question here is - **what is this script trying to verify**??

IMO we should make a clear distinction between whatever is part of KS code, and the environment of the machine which tries to run KS locally. 
**This script is NOT checking KS code, it's checking the machine environment.** 
this was the situation in older script and in this new script.

go.mod and this verify script are both part of KS code. 
if a user clones the repo and starts playing with `go.mod` file, he shouldn't be surprised if things breaks.
the assumption should be that in the older script the var ${VERSION} contains the correct version as written in `go.mod` (in newer script nothing has changed, var just renamed). 

if one of KS team members will change `go.mod` in a way that break things, the PR will fail and we will catch it.

this script only checks the machine env and reports if the env doesn't match the minimal requirements of KS.

I've added another check to verify that go is installed on the machine to verify we don't get garbage.

MikeSpreitzer:
Oh, sorry, I did not notice your mention that `make codegen` with go 1.20 causes no changes. That settles the question for go 1.20. Can you also test for go 1.21?

hack/verify-go-versions.sh
MikeSpreitzer: > if {VERSION} is empty, the script will return successfully

Oh, good point. That is indeed a bug, unless something else is failing when go is not installed (which I think was already true, but no harm in adding an explicit check).

namasl:
I ran `make codegen` with go 1.21.1 where I had originally seen an issue, but strangely I'm now seeing no difference between this and go 1.19. I also checked 1.21.4 (latest), likewise no difference.

namasl:
To add more context, the original issue I saw with `make codegen` happened on PR #1169, where I did a blind find/replace of text across files, including those that are generated. A subsequent `make codegen` after those changes resulted in differences between go 1.19 and 1.21.1. I'm not sure why that happened, but modifying the generated files outside of `make codegen` isn't a normal operation. I see no reason not to accept this PR.

MikeSpreitzer:
The version sensitivity question has been settled, in a positive way.

I still would like to see the test for a bad version replaced with a test for a good version. That can be done in follow-on, or in this PR (I will re-review quickly).

/lgtm

Thanks!

nirrozenbaum:
changed the test. now versions major/minor are separated and compared as integers, which is safer.

hack/verify-go-versions.sh
```diff
@@ -32,9 +32,14 @@ if ! [ -x "$(command -v go)" ]; then # validate go is installed
 fi
 
 if [ -z "${IGNORE_GO_VERSION}" ]; then # validate go version is sufficient
+  MINIMAL_VERSION_MAJOR=$(echo ${MINIMAL_VERSION} | awk -F '.' '{printf $1}')
+  MINIMAL_VERSION_MINOR=$(echo ${MINIMAL_VERSION} | awk -F '.' '{printf $2}')
```
MikeSpreitzer, line 36: You do not need to quote a dot.

hack/verify-go-versions.sh
```diff
@@ -32,9 +32,14 @@ if ! [ -x "$(command -v go)" ]; then # validate go is installed
...
   ENV_VERSION=$(go version | awk '{print $3}' | sed 's/go//')
-  if [[ "$ENV_VERSION" < "$MINIMAL_VERSION" ]]; then
-    echo "Unexpected go version installed. expected minimal go version ${MINIMAL_VERSION} while your environment has version ${ENV_VERSION}. Use IGNORE_GO_VERSION=1 to skip this check."
-    exit 1
+  ENV_VERSION_MAJOR=$(echo ${ENV_VERSION} | awk -F '.' '{printf $1}')
+  ENV_VERSION_MINOR=$(echo ${ENV_VERSION} | awk -F '.' '{printf $2}')
+
+  if [ ${ENV_VERSION_MAJOR} -lt ${MINIMAL_VERSION_MAJOR} ] || [ ${ENV_VERSION_MINOR} -lt ${MINIMAL_VERSION_MINOR} ]; then
```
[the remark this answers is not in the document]
MikeSpreitzer: .. or, exactly equivalently but simpler,

```
if [ "$ENV_VERSION_MAJOR" != "$REQUIRED_VERSION_MAJOR" ] || ! [ "$ENV_VERSION_MINOR" -ge "$MINIMAL_VERSION_MINOR" ]; then
```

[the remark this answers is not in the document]
nirrozenbaum: I'm totally fine with your suggested change. I'll push this change soon.
just one question for my understanding, I'm probably missing something.
you claim that :
```
if ! [ ${ENV_VERSION_MINOR} -ge ${MINIMAL_VERSION_MINOR} ]; then
```
is safer than:
```
if [ ${ENV_VERSION_MINOR} -lt ${MINIMAL_VERSION_MINOR} ]; then
```

I don't understand this claim. 
I'd be happy if you can give any example (doesn't have to be related to this code change), where one of the above returns `true` and the other returns `false`. or any other explanation that shows any difference between these tests.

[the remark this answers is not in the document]
MikeSpreitzer: The numeric comparisons fail when either of the strings is not actually a number. Thus:

```
bash-3.2$ ENV_VERSION_MINOR=""
bash-3.2$ MINIMAL_VERSION_MINOR=19
bash-3.2$ if [ ${ENV_VERSION_MINOR} -lt ${MINIMAL_VERSION_MINOR} ]; then echo unhappy; else echo happy; fi
bash: [: -lt: unary operator expected
happy
bash-3.2$ if [ "${ENV_VERSION_MINOR}" -lt "${MINIMAL_VERSION_MINOR}" ]; then echo unhappy; else echo happy; fi
bash: [: : integer expression expected
happy
bash-3.2$ if ! [ "${ENV_VERSION_MINOR}" -ge "${MINIMAL_VERSION_MINOR}" ]; then echo unhappy; else echo happy; fi
bash: [: : integer expression expected
unhappy
```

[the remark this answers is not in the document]
nirrozenbaum: Thanks for the example Mike!

MikeSpreitzer (commented):
/lgtm
/approve
Thanks!

PR #1262, opened 2023-11-07
🐛 Rename APIExports to avoid collisions

### LINKED ISSUE #1261: bug: two different APIExport objects named "core.k8s.io"
### Describe the bug

```shell
mspreitz@mjs13 kubestellar % grep "name: core.k8s.io" config/kube/exports/*/*
config/kube/exports/cluster-scoped/apiexport-core.k8s.io.yaml:  name: core.k8s.io
config/kube/exports/namespaced/apiexport-core.k8s.io.yaml:  name: core.k8s.io
```

This is bad because it means that only one of them actually takes effect. A user who tries to take advantage of the other gets error messages about the relevant resources not being defined.

### Steps To Reproduce

See above.

### Expected Behavior

No name collisions among objects in tree.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR renames some APIExport objects, to avoid name collisions.

## Related issue(s)

Fixes #1261

### DIFF
--- a/config/kube/exports/cluster-scoped/apiexport-core.k8s.io.yaml
+++ b/config/kube/exports/cluster-scoped/apiexport-core.k8s.io.yaml
@@ -2,7 +2,7 @@ apiVersion: apis.kcp.io/v1alpha1
 kind: APIExport
 metadata:
   creationTimestamp: null
-  name: core.k8s.io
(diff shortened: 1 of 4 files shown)


clubanderson (commented):
/lgtm

PR #1256, opened 2023-11-07
🐛 Rebased update bootstrap and container image to use the patched kcp 0.11

MikeSpreitzer:
## Summary
This PR is #1249 but without the problematic merge of `main` into the PR's branch. I also added a commit that updates example1 to use the patched kcp when deploying bare processes.

## Related issue(s)

Fixes #1248

PR #1267, opened 2023-11-08
:bug: fix syncer integration test to create SA token in downstream workspace

yana1205:
## Summary

This is a followup fix of https://github.com/kubestellar/kubestellar/pull/1256#issuecomment-1797812534.
Now kcp used in KS does not create SA token automatically. Syncer integration test code relies on this feature. The fix is to create SA token in tests. 

## Related issue(s)

Fixes #1256#issuecomment-1797812534

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "go.sum|docs/package-lock.json|cmd/kube-watchall/go.sum|docs/config.toml|docs/resources/_gen|docs/static|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2023-11-03T15:34:50Z",
+  "generated_at": "2023-11-08T22:15:00Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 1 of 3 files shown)


MikeSpreitzer:
Thank you!
Don't you also need to update the ServiceAccount so that it refers to the Secret?
Can you remove https://github.com/kubestellar/kubestellar/blob/8a6873525c1505ffbd8abbce092277f1b4ba90d8/docs/content/Coding%20Milestones/PoC2023q1/kubestellar-syncer-subs/kubestellar-syncer-1-syncer-gen-plugin.md?plain=1#L2-L4 so that the fix gets tested?

yana1205:
>  Don't you also need to update the ServiceAccount so that it refers to the Secret?

No, it's OK with just creating a token and using the token to access the downstream.

> Can you please remove so that the fix gets tested?

Done. `docs-ecutable for syncer e2e` now passes. https://github.com/yana1205/kubestellar/actions/runs/6818011887/job/18542732770


FYI: I rebased the branch from the latest main and force-pushed it.

MikeSpreitzer:
We are talking here about a document that is written for both human readers and to be an executable test. That document, as merged and as in this PR currently, is not accurate in what it tells the reader is done by syncer-gen. The text for human readers should be accurate, as well as the executable part succeeding. @yana1205 can you please update this PR to make the text accurate as well as the execution succeed? We have discussed simplifying syncer-gen in the future, and I have noted that plan in #1274. It is OK with me if you update this PR to describe syncer-gen as it will be after that future simplification.

PR #1272, opened 2023-11-09
✨ First draft of image repo listing utility

MikeSpreitzer:
## Summary
This PR adds a utility in hack/ for fetching the manifests for all the tags of a given image (repository). This is better than `curl` or `wget` of just the first page of tags, and takes the extra step of fetching the associated manifests.

## Related issue(s)

Fixes #

### DIFF
--- a/hack/inspect-image.py
+++ b/hack/inspect-image.py
@@ -1,0 +1,127 @@
+#!/usr/bin/env python3
+
+# Copyright 2023 The KubeStellar Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


clubanderson (commented):
Is this for internal use only?  Why would we need this?

MikeSpreitzer:
This is for use by contributors, which is why I put it in `hack/`.  It is useful to understand the provenance of images. You may have noticed that the quay.io web UI often ignores requests to go beyond the first page of tags. This utility gathers a bunch of useful information in one place.

clubanderson (commented):
/lgtm
/approve

PR #1099, opened 2023-10-08
:sparkles: add control annotation for making auto-downsync optional

yana1205:
## Summary

This PR is for a feature request (https://github.com/kubestellar/kubestellar/issues/1094).

### What's changes
- Introduce an annotation `edge.kubestellar.io/downsync-overwrite: "true" or "false"`. If an object in Workload Description Space (WDS) is set `edge.kubestellar.io/downsync-overwrite: "false"`, Syncer does not overwrite the object in Workload Execution Cluster (WEC).
- Added the test case

### Behavior
When an object is updated,
- if `downsync-overwrite` is true, the downstream object is overwritten by the upstream one (This is nothing difference from the current behavior)
- if `downsync-overwrite` is false, the downstream object is not overwritten

When an object is deleted,
- if `downsync-overwrite` is true, the downstream object is deleted (This is nothing difference from the current behavior)
- if `downsync-overwrite` is false, the downstream object is not deleted

When an object is created,
- if `downsync-overwrite` is true, the downstream object is created (This is nothing difference from the current behavior)
- if `downsync-overwrite` is false, the downstream object is also created

The E2E test case covers above.

### Verification

@clubanderson I pushed the image to my repo:
- `ghcr.io/yana1205/kubestellar/syncer:dev-2023-1008` for main branch
- `ghcr.io/yana1205/kubestellar/syncer:dev-2023-1008-v0.7.0` for release-0.7 branch

Could you verify this change works for the requirement? 

From my end, I verified the following scenario on example1 scenario with `ghcr.io/yana1205/kubestellar/syncer:dev-2023-1008` (not attempt for v0.7).

1. I finished setting up exampl1 scenario. Now I am in WDS for specialstuff.
    ```
    $ kubectl ws .
    Current workspace is "root:my-org:wmw-s".
    ```
1. Make sure that configmap `httpd-htdocs` does not have annotation `edge.kubestellar.io/downsync-overwrite`
    ```
    $ kubectl -n specialstuff get cm httpd-htdocs -o yaml | grep 'edge.kubestellar.io/downsync-overwrite'
    ```
1. `edge.kubestellar.io/downsync-overwrite` is not defined so Syncer works as same as today
    1. Update configmap in WEC (add `aaa: value1`)
        ```
        $ KUBECONFIG=~/.kube/config kubectl --context kind-guilder -n specialstuff edit cm httpd-htdocs     
        apiVersion: v1
        data:
          aaa: value1
          index.html: |
        ...
        ```
[...]

PR #1273, opened 2023-11-09
🐛 Make placement translator respect create-only setting on object

MikeSpreitzer:
## Summary
This PR makes the placement translator respect the create-only annotation introduced in #1099 .  This will allow us to go back to using kcp v0.11.0, which is good because kube-bind depends on some of the functionality that is disabled in our patched version of kcp.

This PR also moves the declaration of the annotation key into the relevant API file, because it is part of the interface.

This PR also makes the code consistently decode the annotation by comparing its value to "false".

## Related issue(s)

### DIFF
--- a/pkg/placement/what-resolver.go
+++ b/pkg/placement/what-resolver.go
@@ -108,8 +108,8 @@ type NamespacedName = Pair[NamespaceName, ObjectName]
 // holds the data for a given object (necessarily in a particular WDS)
 type objectDetails struct {
 	// PlacementBits holds an entry for each EdgePlacement whose what predicate
-	// matches the object, and the bool value is WantSingletonReportedState.
-	PlacementBits MutableMap[ObjectName, bool]
+	// matches the object.
+	PlacementBits MutableMap[ObjectName, DistributionBits]
 }
 
 // NewWhatResolver returns a WhatResolver;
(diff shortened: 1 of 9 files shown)


MikeSpreitzer:
The force-push to 31bf804fa expands the unit test of the what-resolver.

MikeSpreitzer:
The force-push to 7ae992d66 expands the set-binder unit test.

pkg/placement/what-resolver.go
```diff
@@ -639,22 +639,32 @@ func whatMatchingPlacements(logger klog.Logger, wsd *workspaceDetails, candidate
 
 // returns `(changed bool, success bool)`
 func (od *objectDetails) setByMatch(logger klog.Logger, wsd *workspaceDetails, spec *edgeapi.EdgePlacementSpec, epName ObjectName, whatResource string, whatObj mrObject) (bool, bool) {
-	wantSingletonReturn, found := od.PlacementBits.Get(epName)
+	oldDistrBits, found := od.PlacementBits.Get(epName)
+	newDistrBits := DistributionBits{ReturnSingletonState: spec.WantSingletonReportedState,
+		CreateOnly: whatObj != nil && isCreateOnly(whatObj)}
```
ezrasilvera, line 644: Not crucial - if the next condition doesn't depend on the above oldDistrBits and newDistrBits, why not moving this to after the if ! success ?

ezrasilvera:
Just to make sure I got it - As far as I saw, you do allow a createOnly resource to still change the createOnly  annotation back to false , right ?

MikeSpreitzer:
The placement translator is a state-based controller, continually reacting to the latest desired state. The kube API machinery has no way to prevent changes to annotations.

pkg/placement/what-resolver.go
MikeSpreitzer: Yes, there is some freedom in the ordering of these statements.

Issue #1274, opened 2023-11-10
Need to not disable the ServiceAccount Token controller


MikeSpreitzer:
kube-bind relies on some of the behavior of the ServiceAccount Token controller.  Specifically, its creating an actual authorized token and putting it in the ServiceAccount's Secret (which kube-bind's bind operation creates). We considered trying to hack kube-bind's bind operation to add this behavior, but decided it would be better to stop disabling that controller. This will entail a few things.

- [x] Make the Placement Translator respect the `edge.kubestellar.io/downsync-overwrite` annotation, so that it does not get into a fight with that controller over the content of a ServiceAccount so annotated. This is #1273 
- [x] Currently `kubectl-kubestellar-syncer_gen` gives that controller 3 seconds to create the Secret and, if time runs out, requests a token and creates the Secret; it should not be so impatient. This is #1276 
- [ ] Update the syncer doc to track the change in how syncer-gen works and re-enable testing of that part.
- [x] Go back to using plain kcp v0.11.0. This is #1277 
- [ ] Extend EdgePlacement so that it can specify that "create-only" mode externally to a workload object, to support third-party workload. This is #1279 
- [x] Make KubeFlex not disable that controller.
- [ ] Simplify syncer-gen to insist that something else inject the actual token into the Secret (delay this until we are sure the syncer-gen will never be used with the patched kcp).

MikeSpreitzer:
@waltforme 
@pdettori 
@ezrasilvera 
@yana1205

clubanderson:
@yana1205 can you look at the remaining work items in this issue above?

PR #1275, opened 2023-11-10
🌱 Syncer gen patience

MikeSpreitzer:
## Summary
This PR makes the patience of syncer-gen --- the amount of time it waits for the ServiceAccount Token controller to do its job before assuming that will not happen --- configurable, and makes the default longer than the current fixed value (20 sec instead of 3).

This is part of the campaign of #1274 .

This is built on top of #1273 and I will rebase this if that merges first.

## Related issue(s)

Fixes #


MikeSpreitzer:
See #1276 instead.

PR #1276, opened 2023-11-10
🌱 Syncer gen patience2

MikeSpreitzer:
## Summary
This PR is an alternative to #1275 and makes syncer-gen follow the pattern shown in https://eng.d2iq.com/blog/service-account-tokens-in-kubernetes-v1.24/#how-does-this-affect-you if a token Secret for the syncer's ServiceAccount does not show up within a limited amount of time. This also adds a second wait for a token to appear in the Secret. This behavior should work for kcp v0.11.0, Kubernetes 1.23 and later, and OpenShift (which is unlikely to be relevant as a space provider).

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cliplugins/kubestellar/syncer-gen/edgesync.go
+++ b/pkg/cliplugins/kubestellar/syncer-gen/edgesync.go
@@ -91,6 +91,12 @@ type EdgeSyncOptions struct {
 	SyncTargetLabels []string
 	// Lifetime is the requested token lifetime. Optional. (default: 87600h (10 years))
 	Lifetime time.Duration
+	// ServiceAccountTokenWait is how long syncer-gen will wait for the ServiceAccount Token controller
+	// to supply a token Secret for the syncer's ServiceAccount, before syncer-gen does the deed itself.
+	// Default is 20 seconds.
+	// Set it short if you know the ServiceAccount Token controller will never do the job.
+	// Set it long if you know the ServiceAccount controller _is_ eventually going to do the job.
+	ServiceAccountTokenWait time.Duration
 }
 
 // NewSyncOptions returns a new EdgeSyncOptions.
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
The force-push to cc7340c99 fixes a bug (`tokenSecretName` was left empty if the Secret had to be created).

pkg/cliplugins/kubestellar/syncer-gen/edgesync.go
```diff
@@ -500,24 +499,62 @@ func (o *EdgeSyncOptions) enableSyncerForWorkspace(ctx context.Context, config *
...
 		}
+		logger := klog.FromContext(ctx)
+		logger.V(4).Info("Failed to fetch ServiceAccount Secret", "secretName", tokenSecretName, "err", err)
+		fetchError = err
+		return false, nil
+	})
+	if err != nil && err != wait.ErrWaitTimeout {
+		return "", "", nil, fmt.Errorf("failed to get ServiceAccount token, namespace=%v, name=%v, err=%v, fetchError=%v", namespace, tokenSecretName, err, fetchError)
 	}
+	if err == wait.ErrWaitTimeout {
```
yana1205, line 525: Is this block required?
I tried syncer-gen on k8s v1.27 (where SA token is not automatically generated) and k8s v1.23 (where SA token is automatically generated). For both cases, the token in placed in `.data.token` of the SA token secret. 
For v1.27, when the SA token secret is manually created at L503 (`kubeClient.CoreV1().Secrets(namespace).Create(ctx, tokenSecret, metav1.CreateOptions{})`), the SA token is also created and populated in `.data.token`.
For v1.23, when SA is created, the SA token secret is automatically created in which the token is populated in `.data.token`.

yana1205: Thank for updating syncer-gen for the campaign in https://github.com/kubestellar/kubestellar/issues/1274.

yana1205: It is safe to introduce this block for just in case that the SA token is successfully populated in the SA token secret.

MikeSpreitzer: Right. This block covers the context where it currently occurs, using the patched version of kcp v0.11.0 that _does not_ inject an actual token into a service-account-token Secret. Once that is well and truly behind us, I think this block could be removed.

PR #1277, opened 2023-11-10
🌱 Revert to plain kcp

MikeSpreitzer:
## Summary
This PR undoes most of #1256, to revert to using plain kcp v0.11.0 instead of the patched variant that disables the Token controller for ServiceAccounts.

## Related issue(s)

This is part of the campaign #1274

PR #1279, opened 2023-11-10
✨ Enable EdgePlacement to say "create-only"

MikeSpreitzer:
## Summary
This PR expands the EdgePlacement API so that an EdgePlacement can apply the "create-only" setting to selected objects. This is good because it enables the "create-only" setting to be applied to third-party workload objects.

## Related issue(s)

This is built on #1277 and will be rebased if that merges first.


clubanderson:
If our syncer reacts to "overwrite downsync" true or false, then our edgeplacement should match that terminology. We should not have two different terms to mean the same thing.

MikeSpreitzer:
Yes, consistency is good.  Consistency with ArgoCD, FluxCD, and OCM in the naming of this concept would also be good, if they happen to have the same concept and call it by the same name.

MikeSpreitzer:
In the "application" side of OCM (not the "policy" side), they use the term "create only" for this concept, albeit motivated by a different use case.  See https://open-cluster-management.io/concepts/manifestwork/#resource-race-and-adoption

clubanderson:
Right. We are not OCM. We use overwrite downsync.

Issue #1266, opened 2023-11-08
Tracking of decoupling integration to KS


ezrasilvera:
|ID| Item | Status| Issue | PR | comments|
| --------| -------- | ------- | -------- | ------- | ------- |
|1 |Configure and install the SM and space providers | PR waiting for approval |#1215 |#1218|
|2| Change KS scripts (install, tests, etc..) to use the space manager instead of `kcp ws` commands | Being implemented |#1215 |TBD| Dependent on **1**
|3| Replace kcp-bind with kube-bind | <ul><li> PR waiting for review</li><li>Fixing fails in tests</li></ul> |#1208 |#1265| This should work only on KCP until item (4) is completed and merged
|3| <ul><li> Stop using KCP's cluster-id annotation in the KS controllers code</li><li>Change the KS controllers to use the Space Aware Client when interacting with spaces</li></ul>   | Implemented |#1208 |TBD| Dependent on items 1, 2, and 3 to be able to test

MikeSpreitzer:
" KS quick start" is too vague. There are four quickstarts. I suppose you mean the one whose filename is `user-quickstart-test.md`.

MikeSpreitzer:
@1218 has been approved and merged.

MikeSpreitzer:
Regarding the "issue" column: 1208 has two parts (the first itemized enumeration in there), and they are very different. I think it would help to update this column to refer to "1208 part 1" and "1208 part 2".

ezrasilvera:
> Is item 4 really in issue 1208? I am having trouble seeing it there.
It's not part of the issue. It's just that the code we implemented already contain both the changes for the cluster-id on top of the changes t support spaces . I'll update the issue

ezrasilvera:
> What is the difference between the two documentation items? They read the same now.

updated

MikeSpreitzer:
Since PR #1265 only removes the use of kcp APIExport+APIBinding for the `edge.kubestellar.io` API group, and APIExport+APIBinding are also used for the kube API groups, we need another row in the table to cover removing that second use of APIExport+APIBinding.

clubanderson:
@ezrasilvera please keep the description of this issue updated regularly so that it is easy for anyone to see it's status

MikeSpreitzer:
A. Need a row for switching tests to use KubeFlex, or duplicate tests to test with kcp and with KubeFlex. And debugging whatever turns up.

[...]

PR #1265, opened 2023-11-08
:sparkles: kube bind  integration

ezrasilvera:
This PR replaces the usage of kcp-bind with kube-bind.  

## Related issue(s)

Fixes #

PR #1286, opened 2023-11-13
:sparkles: Use space framework in KubeStellar controllers

andreyod:
## Summary
This PR should be deployed after https://github.com/kubestellar/kubestellar/pull/1265 . It doesn't depend on the kube-bind functionality but the two PRs touches the same files and functions, and it will be easier to apply this PR on top of #1265 rather than the opposite 

This PR is also depends on https://github.com/kubestellar/kubestellar/pull/1268 and CICD tests expected to pass after 1268 is merged.
This PR introduce initial use of space framework and space aware client in Kubestellar controllers.
Note that this PR does not complete the KCP runtime decouple and should be followed by another PR to remove remaining dependencies on KCP due to the cluster-id VS space-name issues. 

## Related issue(s)
https://github.com/kubestellar/kubestellar/issues/1266 , Item 4

Fixes #


MikeSpreitzer:
Presumably fixing the merge conflict will include updating the opening comment (which currently was written before #1265 and #1268 merged).

PR #1291, opened 2023-11-14
✨ Add kube-bind ID mapping support

MikeSpreitzer:
## Summary
This PR adds the implementation of the ID mapping support for users of kube-bind documented in #1289 .

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/kbuser/object_ids.go
+++ b/pkg/kbuser/object_ids.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2023 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kbuser
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// AnalyzeObjectID examines an object in a kube-bind service provider
+// cluster and returns the object's ID as known in the service consumer
+// cluster and the kube-bind ID for the consumer, or an error if
+// the object does not appear to be a provider's copy of a consumer's object.
+func AnalyzeObjectID(obj metav1.Object) (namespace, name, kbSpaceID string, err error) {
+	namespace = obj.GetNamespace()
+	name = obj.GetName()
+	annotations := obj.GetAnnotations()
+	if annotations != nil {
+		kbSpaceID = annotations["kube-bind.io/cluster-namespace"]
+	}
+	if kbSpaceID == "" {
+		return "", "", "", errors.New("no 'kube-bind.io/cluster-namespace' annotation found")
+	}
+	var ok bool
+	if namespace == "" {
+		name, ok = cutPrefix(name, kbSpaceID+"-")
+		if !ok {
+			err = fmt.Errorf("name %q does not have prefix for comsumer %q", name, kbSpaceID)
+		}
+		return
+	}
+	namespace, ok = cutPrefix(namespace, kbSpaceID+"-")
+	if !ok {
+		err = fmt.Errorf("namespace %q does not have prefix for comsumer %q", namespace, kbSpaceID)
+	}
+	return
+}
+
+func cutPrefix(s, sep string) (string, bool) {
+	if strings.HasPrefix(s, sep) {
+		return s[len(sep):], true
+	}
+	return s, false
+}
+
+// ComposeClusterScopedName translates the name of a cluster-scoped object
+// from what appears in the service consumer cluster to what appears in the
+// service provider cluster. A Namespace is an example of a cluster-scoped
+// object.
+// For namespaced objects, kube-bind translates the name of their namespace.
+func ComposeClusterScopedName(kbSpaceID string, name string) string {
+	return kbSpaceID + "-" + name
+}
--- a/pkg/kbuser/space_ids.go
+++ b/pkg/kbuser/space_ids.go
@@ -1,0 +1,151 @@
+/*
+Copyright 2023 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kbuser
+
+import (
+	"context"
+	"strings"
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	informerscorev1 "k8s.io/client-go/informers/core/v1"
+	kubernetes "k8s.io/client-go/kubernetes"
+	cache "k8s.io/client-go/tools/cache"
+)
+
+const KubeBindLabelName string = "kubestellar.io/kube-bind-id"
+
+const CMNamePrefix string = "kbmap-"
+
+// KubeBindSpaceRelation is specific to one kube-bind service provider
+// and reveals the 1:1 relation between the kube-bind identifier for
+// a consumer (i.e., the name of the "cluster namespace" for that consumer
+// in the provider) and the underlying space identifier (i.e., the
+// `logicalcluster.Name`).
+type KubeBindSpaceRelation interface {
+	// SpaceIDToKubeBind maps an underlying space ID to a kube-bind space ID.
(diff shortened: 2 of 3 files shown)


MikeSpreitzer:
My initial draft of `parseMappedName` is just a guess. Please let me know whether and, if so, how it needs to be adjusted.

pkg/kbuser/space_ids.go
```diff
@@ -0,0 +1,151 @@
...
+// in the provider) and the underlying space identifier (i.e., the
+// `logicalcluster.Name`).
+type KubeBindSpaceRelation interface {
+	// SpaceIDToKubeBind maps an underlying space ID to a kube-bind space ID.
+	// Returns empty string if there is no relationship.
+	SpaceIDToKubeBind(spaceID string) string
+
+	// SpaceIDFromKubeBind maps a kube-bind space ID to an underlying space ID.
+	// Returns empty string if there is no relationship.
+	SpaceIDFromKubeBind(kubeBindID string) string
```
ezrasilvera, line 47: why not symmetrical naming e.g., KubeBindToSpaceID   ?

pkg/kbuser/space_ids.go
```diff
@@ -0,0 +1,151 @@
...
+	toKubeBind   map[string]string
+	fromKubeBind map[string]string
+}
+
+func (reln *kubeBindSpaceRelation) OnAdd(obj any) {
+	cm := obj.(*corev1.ConfigMap)
+	if !strings.HasPrefix(cm.Name, CMNamePrefix) || len(cm.Name) <= len(CMNamePrefix)+1 {
+		return
+	}
+	underID := cm.Name[len(CMNamePrefix):]
```
ezrasilvera, line 87: As above - I find the "under" part confusing

pkg/kbuser/object_ids.go
```diff
@@ -0,0 +1,56 @@
...
+// cluster and the kube-bind ID for the consumer, or an error if
+// the object does not appear to be a provider's copy of a consumer's object.
+func AnalyzeObjectID(obj metav1.Object) (namespace, name, kbSpaceID string, err error) {
+	namespace = obj.GetNamespace()
+	name = obj.GetName()
+	if namespace == "" {
+		name, kbSpaceID, err = parseMappedName(name)
+		return
+	}
+	namespace, kbSpaceID, err = parseMappedName(name)
```
ezrasilvera, line 37: Is there a case for name spaced object ?

waltforme:
> My initial draft of `parseMappedName` is just a guess. Please let me know whether and, if so, how it needs to be adjusted.

We don't know the number of `-`s in a 'mapped name', because the cluster namespace's prefix is configurable through a flag of the example-backend binary. I saw the prefix been set to `kube-bind` in kube-bind's documents, so a cluster namespace reads like `kube-bind-jlchp`, I also saw the prefix been set to `cluster` in the tests, so a cluster namespace reads like `cluster-khk4n`.

What we do know is that a 'mapped name' is a cluster namespace name plus exactly one '-' plus the original name on consumer side. Documented [here](
https://github.com/kube-bind/kube-bind/blob/428391bf62fa7b6be01d470e017610af72761096/pkg/konnector/controllers/cluster/serviceexport/cluster-scoped/utils.go#L32).

I think one possible way to parse is to use the annotation keyed `kube-bind.io/cluster-namespace` on an object, but up to you to decide. Let me paste the metadata of a synced object here for example:
```
bash-5.1$ kc get synctargets.edge kube-bind-jgpwb-dev -oyaml | yq .metadata
annotations:
  kcp.io/cluster: 1zblnr99qzqhto3b
  kube-bind.io/cluster-namespace: kube-bind-jgpwb
creationTimestamp: "2023-11-09T20:04:11Z"
generation: 1
labels:
  env: dev
name: kube-bind-jgpwb-dev
ownerReferences:
  - apiVersion: v1
    kind: Namespace
    name: kube-bind-jgpwb
    uid: ac37228f-2d29-4abd-ad01-2ef69b65b16c
resourceVersion: "2164"
uid: 14c9a3e5-40a3-420b-8c12-ce26cd4759b3
```

waltforme:
Just a general question. I imaged that there will be one configmap whose `data` persists the all the mappings for all the spaces. But in this PR I see that there are multiple configmaps, where each configmap persists one mapping, for one space. May I know the thoughts to use multiple configmaps instead of one?

pkg/kbuser/space_ids.go
```diff
@@ -0,0 +1,151 @@
...
+	// SpaceIDFromKubeBind maps a kube-bind space ID to an underlying space ID.
+	// Returns empty string if there is no relationship.
+	SpaceIDFromKubeBind(kubeBindID string) string
+}
+
+// NewKubeBindSpaceRelation creates a KubeBindSpaceRelation given a
+// Kubernetes client for the provider's space and a context bounding
+// the relation's lifetime.
+func NewKubeBindSpaceRelation(ctx context.Context, client kubernetes.Interface) *kubeBindSpaceRelation {
+	informer := informerscorev1.NewFilteredConfigMapInformer(client, "kubestellar-system", 0, cache.Indexers{}, tweakListOptions)
```
waltforme, line 54: Maybe we can continue to use `kubestellar` namespace?

ezrasilvera:
> Just a general question. I imaged that there will be one configmap whose `data` persists the all the mappings for all the spaces. But in this PR I see that there are multiple configmaps, where each configmap persists one mapping, for one space. May I know the thoughts to use multiple configmaps instead of one?

In general you are correct, but we thought that for the general space provider  case, in the future, we will save that info in the space object so it seems similar to what you have here ..
In any case we don't envision many consumers :-)

pkg/kbuser/space_ids.go
MikeSpreitzer: yes

MikeSpreitzer:
Before we go any further, I realized that I should double-check something.
Consider the case of a service provider cluster that exports multiple CRDs.
Consider a single service consumer cluster that imports more than one of those CRDs. Does kube-bind assign the same "cluster namespace" for that (provider, consumer) pair, across all CRDs?

pkg/kbuser/object_ids.go
MikeSpreitzer: I am not sure what we have in the future. I thought it best to not bet against caring about namespaced objects.

MikeSpreitzer:
@ezrasilvera , you made several comments about one issue: terminology. I struggled with that, and am not sure what to do. Here are some of my thoughts.

- I think that it is fair to say that kube-bind is dealing with spaces. Sure, the authors were thinking and writing "cluster", but the focus of what kube-bind deals with is only what we call a "space" (yeah, the konnector has to run somewhere, but we have demonstrated that this does not need to be "in a space").
- We could try to use "cluster" when we are talking kube-bind and "space" otherwise, but I would not find this clearest because a "space" can be a "cluster".
- Thus, I view kube-bind as having its own space identifiers (i.e., the "cluster namespace" name), and the issue is relating those space identifiers with the other space identifiers in play.
- What to call the other space identifiers in our view? At the moment they are `logicalcluster.Name`, but we plan to change that soon. Is there a term that makes sense now and in the near future? I tried "underlying space identifier" on the grounds that these "underly" what kube-bind is doing in the sense of being prior or depended on: what kube-bind does is a layer of functionality added on top of the more fundamental space functionality. I am open to other suggestions.

MikeSpreitzer:
The force-push to 57b125ff0 corrects the namespace of the ConfigMap objects and changes `AnalyzeObjectID` to get kube-bind's ID for the consumer from the annotation.

MikeSpreitzer:
And the force-push to 313410b47 removes an unintended blank line.

MikeSpreitzer:
> Just a general question. I imaged that there will be one configmap whose data persists the all the mappings for all the spaces. But in this PR I see that there are multiple configmaps, where each configmap persists one mapping, for one space. May I know the thoughts to use multiple configmaps instead of one?

Putting a database in one object invites several problems.
- not botching concurrent updates.
- It makes access control harder, requiring row-level access control rather than object-level access control to do a great job.
- frequent updates to a large object is a scaling botch. I know we do not expect a large data volume or update rate, but still I saw no reason to invite this problem.

ezrasilvera:
> Before we go any further, I realized that I should double-check something. Consider the case of a service provider cluster that exports multiple CRDs. Consider a single service consumer cluster that imports more than one of those CRDs. Does kube-bind assign the same "cluster namespace" for that (provider, consumer) pair, across all CRDs?

See slack as well. As far as I remember the "cluster namespace" in the provider is per consumer cluster. So yes, all resources/CRDs  coming from that consumer will get the same "cluster namespace" and the same annotation.

ezrasilvera:
@MikeSpreitzer regarding terminology - my concern was not related to "cluster" Vs "space" but rather to "underlying" Vs consumer. It felt to me that the variables should clearly indicates that we are talking about the "system wide ID" of the consumer VS the private ID of that consumer in the specific provider.  kubeBindID, kbSpaceID, etc.. all seem reasonable for the second. I'm having difficulties with the "underline" for the first. 
This is just terminology - so I maybe a clear definition at the beginning will be enough.  
The code is LGTM to me - if you are 100% fine with underlining , let's merge it with that.

waltforme:
> Putting a database in one object invites several problems.
> 
>     * not botching concurrent updates.
> 
>     * It makes access control harder, requiring row-level access control rather than object-level access control to do a great job.
> 
>     * frequent updates to a large object is a scaling botch. I know we do not expect a large data volume or update rate, but still I saw no reason to invite this problem.

I see. Thanks for sharing the thoughts!

MikeSpreitzer:
@ezrasilvera : yes, "system-wide" vs. "provider's" is another way of looking at the distinction.

Let's merge this as it is, the code is good enough, and work on the variable names and comments in follow-on.

/approve

PR #1292, opened 2023-11-14
:sparkles: Created EdgePlacementDecision API

vMaroon:
## Summary
Created initial API of the EdgePlacementDecision resource that will serve as a part of the new architecture, that serves pluggable transports.

EdgePlacementDecision is bound to a single EdgePlacement resource, and reflects the resolution of the latter's placement selectors. The decision resource explicitly states **what** workload resources are to be propagated to **where**.


pkg/apis/edge/v2alpha1/edge-placementdecision.go
```diff
@@ -0,0 +1,115 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SourceEdgePlacementLabelKey is the key of the label used in a EdgePlacementDecision.
+// to reference the EdgePlacement that it is part of the response to.
+// We use a label rather than a field because field selectors do not work
+// on a resource defined by a CRD.
+const SourceEdgePlacementLabelKey string = "edge.kubestellar.io/source-edgeplacement"
```
nirrozenbaum, line 27: remove this label.
if we use single decision object per edge placement, no need for this label since we would use the same object name as the placement.

pkg/apis/edge/v2alpha1/edge-placementdecision.go
```diff
@@ -0,0 +1,115 @@
...
+// All resources present in this spec are propagated to all destinations present.
+type EdgePlacementDecisionSpec struct {
+	// `Workload` is a collection of namespaced and cluster-scoped resources to be propagated to destination clusters.
+	Workload EdgePlacementDecisionSpecResources `json:"workload,omitempty"`
+
+	// `destinations` is a list of cluster-identifiers that the resources should be propagated to.
+	Destinations []Destination `json:"destinations,omitempty"`
+}
+
+// EdgePlacementDecisionSpecResources explicitly defines the resources to be synced.
```
nirrozenbaum, line 66: fix comment - "to be downsynced". people who are familiar with current upsync/downsync capabilites might get confused.

pkg/apis/edge/v2alpha1/edge-placementdecision.go
```diff
@@ -0,0 +1,115 @@
...
+// Currently, the fields inside are taken from syncer-config.go.
```
nirrozenbaum, line 67: remove this line. if this is merged there will be no syncer config.

pkg/apis/edge/v2alpha1/edge-placementdecision.go
```diff
@@ -0,0 +1,115 @@
...
+	// +optional
+	NamespacedObjects []NamespaceScopeDownsyncObjects `json:"namespacedObjects,omitempty"`
+
+	// `clusterScope` holds a list of individual cluster-scoped objects
+	// to downsync, organized by resource.
+	// Remember that a "resource" is a kind/type/sort of objects,
+	// not an individual object.
+	// +optional
+	ClusterScope []ClusterScopeDownsyncResource `json:"clusterScope,omitempty"`
+}
```
nirrozenbaum, line 82: where is the `NamespaceScopeDownsyncs`, `NamespaceScopeDownsyncObjects` and `ClusterScopeDownsyncResource` structs?
if this is referencing the structs from SyncerConfig this is not good.
while we keep syncer config for now, we should keep in mind that this will not be there.
we need to create the structs here and use them. you can use different names to avoid conflicts.

vMaroon: I think using the exact fields from the SyncerConfig as we claimed makes sense. It is not a bad practice as these structs are exported. Once the SyncerConfig is gone, we can move them here and also remove the comment on top of the struct. 

What do you think?

nirrozenbaum: while from one hand it makes sense, I thought we might use this opportunity to revisit the structure fields.
for example - what is the difference between `NamespaceScopeDownsyncs` and `NamespaceScopeDownsyncObjects` and why do we need both? I don't know the answer myself yet.
if both represent namespace scoped objects, it would be good to consolidate them into a single list.

since we don't want to touch SyncerConfig at this point not to break existing code functionality, if we want to do any change to the structures we cannot change in syncer config file.

vMaroon: I assumed that no field is redundant, and that we'll want to keep the exact same fields until further design development.

If you insist on duplicating the fields I can do it, I personally prefer using SyncerConfig's exported structs, which clearly highlights the information duplication while not duplicating code.

nirrozenbaum: do you see a case where the transport controller needs 3 lists and not 2?
this structure affects the controller implementation (whether the controller should handle 2 lists of 3).
if the answer is yes, then we can keep existing structs.
if on the other hand we agree that there is no need in more than two lists, than we know for sure that we will change this struct soon or later. 
and if we plan on changing the structs, we need our own structs since we cannot modify syncer config.

vMaroon: @nirrozenbaum after further investigation, the use-case seems to be that NamespaceScope is for specifying a broader, more general selection of resources. It allows the user to define entire namespaces and types of resources within those namespaces that should be synced. This is useful when you want to sync a whole set of resources within specified namespaces. For example, syncing all deployments and services from certain namespaces.

It makes sense to use at scale, assuming that the calculation of the field is efficient, **BUT**, it is not actually used anywhere outside tests, therefore removing.

On the matter of duplicating structs to gain control over insides, @shmuelk and I agree that it makes sense to do that lazily, as in create new structs when we want to change the content. For now it makes sense to use the structs defined in SyncerConfig, within the same package. What do you think?

nirrozenbaum: ok, acceptable. let's do it lazily and create the new structs when needed.

PR #1294, opened 2023-11-15
:sparkles: Partial conversion of kubectl-kubestellar* Bash commands to Go

namasl:
## Summary

This is a partial conversion of the kubectl-kubestellar* Bash scripts to Go, covering the commands:
- kubectl kubestellar ensure location
- kubectl kubestellar ensure wds
- kubectl kubestellar remove location
- kubectl kubestellar remove wds
- kubectl kubestellar get-external-kubeconfig
- kubectl kubestellar get-internal-kubeconfig

The following commands are not covered
- kubectl kubestellar prep-for-syncer
- kubectl kubestellar prep-for-cluster
- kubectl kubestellar deploy

This PR is for general awareness as well as a request for review, rather than an expectation of merging given the KCP decoupling work. If a merge were desired, the Makefile and some docfiles would need minor updates, and the Bash implementations should be deleted. A drawback to such a merge is that the help printed by `kubectl kubestellar` would neglect to inform the user of the unimplemented commands that still exist as Bash scripts.

The commands in this PR may be compiled by going into the `cmd/kubectl-kubestellar/` directory and executing `go build -o kubectl-kubestellar main.go`, and the resultant `kubectl-kubestellar` executable moved to a directory in the search path. Alternatively, in lieu of compilation these commands may be run by substituting `kubectl kubestellar` with `go run main.go` while in the `cmd/kubectl-kubestellar/` directory.

Here are some example commands that can be executed upon completion of the Quickstart using Kind. These can be executed in any order, and repeatedly, as all commands are idempotent.

- `KUBECONFIG=~/ks-core.kubeconfig kubectl kubestellar ensure location --imw imw1 ks-edge-cluster1 "env=ks-edge-cluster1" "location-group=edge"`
- `KUBECONFIG=~/ks-core.kubeconfig kubectl kubestellar remove location --imw imw1 ks-edge-cluster1`
- `KUBECONFIG=~/ks-core.kubeconfig kubectl kubestellar ensure wds wmw1 --with-kube`
- `KUBECONFIG=~/ks-core.kubeconfig kubectl kubestellar remove wds wmw1`
- `KUBECONFIG=~/.kube/config kubectl kubestellar get-external-kubeconfig -o ks-config-ext.yaml`
- `KUBECONFIG=~/.kube/config kubectl kubestellar get-internal-kubeconfig -o ks-config-int.yaml`

## Related issue(s)
#414 

Fixes #


MikeSpreitzer:
What about `kubectl kubestellar ensure wmw` ?

namasl:
> What about `kubectl kubestellar ensure wmw` ?

The main `wmw` commands are now named `wds` to adjust to the newer lingo, but `wmw` survives an alias to `wds`.

cmd/kubectl-kubestellar/cmd/ensure/location.go
```diff
@@ -0,0 +1,167 @@
...
+	plugin "github.com/kubestellar/kubestellar/pkg/cliplugins/kubestellar/ensure"
+)
+
+var imw string // IMW workspace path
+
+// Create the Cobra sub-command for 'kubectl kubestellar ensure location'
+func newCmdEnsureLocation() *cobra.Command {
+	// Make location command
+	cmdLocation := &cobra.Command{
+		Use:     "location --imw <IMW_NAME> <LOCATION_NAME> <\"KEY=VALUE\" ...>",
```
MikeSpreitzer, line 47: I am not sure what the angle brackets are telling the reader here. If they are intended to surround a variable then `KEY` and `VALUE` should get their own individual pair of angle brackets. If they are intended to highlight what the shell parses as a "word" then I think that is overkill, readers can pretty easily see that without them.

cmd/kubectl-kubestellar/cmd/ensure/location.go
```diff
@@ -0,0 +1,167 @@
...
+			// want the help to be displayed when the error is due to an
+			// invalid command.
+			cmd.SilenceUsage = true
+			err := ensureLocation(cmd, args)
+			return err
+		},
+	}
+
+	// Add flag for IMW workspace
+	cmdLocation.Flags().StringVar(&imw, "imw", "", "IMW workspace")
```
MikeSpreitzer, line 66: "IMW workspace" is like a "PIN number".

cmd/kubectl-kubestellar/cmd/ensure/location.go
```diff
@@ -0,0 +1,167 @@
...
+	err = plugin.CheckLabelArgs(labels)
+	if err != nil {
+		logger.Error(err, fmt.Sprintf("Problem with label arguments %s", labels))
+		return err
+	}
+
+	// Options for IMW workspace
+	imwClientOpts := clientopts.NewClientOpts("imw", "Access to the IMW workspace")
+	// Set default context to "root"; we will need to append the IMW name to the root server
+	imwClientOpts.SetDefaultCurrentContext("root")
```
MikeSpreitzer, line 111: No, you want to use the `genericclioptions.NewConfigFlags(true)` from `cmd/ensure/ensure.go`.

cmd/kubectl-kubestellar/cmd/ensure/location.go
```diff
@@ -0,0 +1,166 @@
...
+// - Check for SyncTarget of provided name in IMW, create if not
+// - Check that SyncTarget has an "id" label matching the Location name
+// - Ensure that SyncTarget has the labels provided by the user
+// - Check for Location of provided name in IMW, create if not
+// - Ensure that Location has the labels provided by the user
+// - If Location "default" exists, delete it
+func ensureLocation(cmdLocation *cobra.Command, args []string, cliOpts *genericclioptions.ConfigFlags) error {
+	locationName := args[0]
+	labels := args[1:]
+	ctx := context.Background()
```
MikeSpreitzer, line 86: Wouldn't it be better to use `cmdLocation.Context()`?

cmd/kubectl-kubestellar/cmd/ensure/location.go
```diff
@@ -0,0 +1,166 @@
...
+	// Make sure user provided labels are valid
+	err = plugin.CheckLabelArgs(labels)
+	if err != nil {
+		logger.Error(err, fmt.Sprintf("Problem with label arguments %s", labels))
+		return err
+	}
+
+	// Set context to root, later on we will append the IMW name to the root server
+	configContext := "root"
+	cliOpts.Context = &configContext
```
MikeSpreitzer, line 110: This unconditionally sets the context. The normal behavior is to let the user control the context through the command line flag. To do that, set `cliOpts.Context` _before_ calling `cliOpts.AddFlags(fs)`.

cmd/kubectl-kubestellar/cmd/ensure/location.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// Get client config from flags
+	config, err := cliOpts.ToRESTConfig()
+	if err != nil {
+		logger.Error(err, "Failed to get config from flags")
+		return err
+	}
+
+	// Update host to work on objects within IMW
+	config.Host += ":" + imw
```
MikeSpreitzer, line 120: This only works if `config.Host` was referring to the `root` workspace. I see how hijacking the `context` setting above ensures that, but _that_ assumes that there _is_ a context named `root` and it refers to the workspace named `root`. I have not looked at how `kubectl ws root:imw1` is implemented; is it more robust than this?

cmd/kubectl-kubestellar/cmd/ensure/location.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// Check that APIBinding exists, create if not
+	// This function prints its own log messages, so no need to add any here.
+	err = plugin.VerifyOrCreateAPIBinding(kcpClient, ctx, "edge.kubestellar.io", "edge.kubestellar.io", "root:espw")
+	if err != nil {
+		return err
+	}
+
+	// Create client-go instance from config
+	client, err := clientset.NewForConfig(config)
```
MikeSpreitzer, line 138: I would qualify `client` and `clientset` here, to make the complementary to the `kcp` ones above; perhaps with `k8s`?

cmd/kubectl-kubestellar/cmd/ensure/wds.go
```diff
@@ -0,0 +1,144 @@
...
+			// want the help to be displayed when the error is due to an
+			// invalid command.
+			cmd.SilenceUsage = true
+			err := ensureWds(cmd, args, cliOpts)
+			return err
+		},
+	}
+
+	// Add flag for
+	cmdWds.Flags().BoolVar(&withKube, "with-kube", true, "Include root:compute API bindings")
```
MikeSpreitzer, line 61: I think it would be clearer if the help string spoke of "Kubernetes API bindings" rather than "root:compute API bindings".

cmd/kubectl-kubestellar/cmd/ensure/wds.go
```diff
@@ -0,0 +1,144 @@
...
+	cmdWds.MarkFlagRequired("with-kube")
```
MikeSpreitzer, line 62: Why make this required?  It is not required in the current bash script.

cmd/kubectl-kubestellar/cmd/ensure/wds.go
```diff
@@ -0,0 +1,144 @@
...
+//   - At the root KCP level, check for a WDS workspace having the user provided
+//     name, create if needed
+//   - At the root KCP level, check for an APIBinding "bind-espw" with export path
+//     "root:espw" and export name "edge.kubestellar.io"
+//   - If --with-kube is true, ensure a list of APIBindings exist with export path
+//     "root:compute" (create any that are missing). If --with-kube is false, make
+//     sure none of these exist (delete as needed).
+func ensureWds(cmdWds *cobra.Command, args []string, cliOpts *genericclioptions.ConfigFlags) error {
+	wdsName := args[0] // name of WDS
+	ctx := context.Background()
```
MikeSpreitzer, line 78: Why not get the context from the Command?

cmd/kubectl-kubestellar/cmd/ensure/wds.go
```diff
@@ -0,0 +1,144 @@
...
+
+	// Check for WDS workspace, create if it does not exist
+	// This function prints its own log messages, so no need to add any here.
+	err = plugin.VerifyOrCreateWDS(client, ctx, wdsName)
+	if err != nil {
+		return err
+	}
+
+	// Update host to work on objects within WDS workspace
+	config.Host += ":" + wdsName
```
MikeSpreitzer, line 119: `config` here is a _pointer_ to a data structure that is already in use from the earlier `NewForConfig()`, right? I think it would be safer to copy the Config struct and modify the copy.

cmd/kubectl-kubestellar/cmd/root.go
```diff
@@ -0,0 +1,117 @@
...
+	// Add cliOpts flags to fs (flow from syntax is confusing, goes -->)
+	cliOpts.AddFlags(fs)
+
+	// Allow logging to pick up flags
+	klog.InitFlags(flag.CommandLine)
+	// Add logging flags to fs
+	fs.AddGoFlagSet(flag.CommandLine)
+	// Add flags to our command; make these persistent (available to this
+	// command and all sub-commands)
+	rootCmd.PersistentFlags().AddFlagSet(fs)
```
MikeSpreitzer, line 103: You don't want two different `cliOpts` adding the same flags to the persistent flag set. If you add the root command's `cliOpts`'s flags then do not also add the flags of different `cliOpts` created for subsidiary commands.

Remember that `kubectl kubestellar deploy` is different from all the others, it does NOT want the flags of a `cliOpts`, it wants the flags for `helm create` / `helm apply`.

cmd/kubectl-kubestellar/cmd/ensure/location.go
namasl: OK, I can make this a little more robust. At the moment I make the following assumptions:

1) We only ever have a flat KCP workspace structure (should be safe at this point).

2) The user does not provide a full workspace path, just the workspace name (i.e. `imw1` not `root:imw1`). I don't explicitly check this, so `root:imw1` will give an "access denied" error rather than a more helpful message. I can add a check to make this more user friendly, or even accept and remove `root:`.

3) There is a "root" workspace. Instead I could take the `config.Host` from whatever context we are in, chop off any subdirectories to get something like "https://hostname:1119", and append to it "/clusters/root:imwName". This is more or less what ends up happening in https://github.com/kcp-dev/kcp/blob/v0.11.0/pkg/cliplugins/workspace/plugin/kubeconfig.go#L277-L296

clubanderson:
per @namasl - closing in favor of working with the new space framework architecture.

PR #1301, opened 2023-11-16
✨ Let the mailbox-controller use the new `kube-bind` script

waltforme:
## Summary
As shown in the title.

Tested by running example 1 locally that we have the planned configmaps showing up:
```
ubuntu@ip-172-31-67-225:~$ kc ws root
Current workspace is "root".
ubuntu@ip-172-31-67-225:~$ kc get ws | grep -v ^compute | grep -v ^espw
NAME                                                       TYPE           REGION   PHASE   URL                                                    AGE
2t187v44xbbdc4kr-mb-d9fcabfa-605c-4b74-9901-c2e232e96734   organization            Ready   https://172.31.67.225:6443/clusters/2hyw8w88zrmbqhhs   27m
2t187v44xbbdc4kr-mb-f92233a7-e137-437f-8908-76428bfa3f7a   organization            Ready   https://172.31.67.225:6443/clusters/vzixinlk05nz6y6s   27m
imw1                                                       organization            Ready   https://172.31.67.225:6443/clusters/33vmovfryc2wtjxr   27m
wmw-c                                                      organization            Ready   https://172.31.67.225:6443/clusters/1g3ae893ute3zz9a   26m
wmw-s                                                      organization            Ready   https://172.31.67.225:6443/clusters/22k9lrouumzk6tml   25m
wmw1                                                       organization            Ready   https://172.31.67.225:6443/clusters/34655jlznbibixys   27m
ubuntu@ip-172-31-67-225:~$ kc ws root:espw
Current workspace is "root:espw".
ubuntu@ip-172-31-67-225:~$ kc -n kubestellar get cm -o=custom-columns=Name:.metadata.name,Labels:.metadata.labels | grep kbmap
kbmap-1g3ae893ute3zz9a   map[kubestellar.io/kube-bind-id:kube-bind-cqwxz]
kbmap-22k9lrouumzk6tml   map[kubestellar.io/kube-bind-id:kube-bind-4gq8p]
kbmap-2hyw8w88zrmbqhhs   map[kubestellar.io/kube-bind-id:kube-bind-v4krf]
kbmap-33vmovfryc2wtjxr   map[kubestellar.io/kube-bind-id:kube-bind-lrgkj]
kbmap-34655jlznbibixys   map[kubestellar.io/kube-bind-id:kube-bind-xg66t]
kbmap-vzixinlk05nz6y6s   map[kubestellar.io/kube-bind-id:kube-bind-npr5f]
ubuntu@ip-172-31-67-225:~$ kc get ns | egrep 'kube-bind-[a-z0-9]+\s'
kube-bind-4gq8p                Active   25m
kube-bind-cqwxz                Active   26m
kube-bind-lrgkj                Active   27m
kube-bind-npr5f                Active   27m
kube-bind-v4krf                Active   27m
kube-bind-xg66t                Active   27m
ubuntu@ip-172-31-67-225:~$
```


cmd/mailbox-controller/controller.go
```diff
@@ -255,76 +258,76 @@ func (ctl *mbCtl) sync(ctx context.Context, refany any) bool {
...
+	for idx, resource := range resourcesToBind {
+		logger.V(2).Info("Ensuring binding", "script", shellScriptName, "resource", resource)
 
-	r, _ := regexp.Compile("🔒 (Created|Updated) secret.*namespace (.*)")
-	found, namespaceName := false, ""
-	for {
-		line, _, err := errBuf.ReadLine()
+		// Make sandboxes for the concurrent access of the kubeconfig from the mailbox-controller's multiple workers
+		// This is a compromise due to the situation that 1) we are using kcp and 2) we are running tests on bare process of this controller
+		workerIdx, _ := ctx.Value(ctxKey("worker")).(int)
```
MikeSpreitzer, line 270: Discarding the error is dangerous. `workerIdx := ctx.Value(ctxKey("worker")).(int)` is valid and panics if the type assertion fails --- which is safer. Safer still would be to generate a fresh number every time (e.g., with `atomic.AddUint64(.., 1)`) and include deletion of the temp file in the command line (e.g., with a command like `( cp $src $temp; trap "rm -f $temp" EXIT; use $temp )`).

cmd/mailbox-controller/controller.go
```diff
@@ -255,76 +258,76 @@ func (ctl *mbCtl) sync(ctx context.Context, refany any) bool {
...
+		if idx == 0 {
+			cmdLine = cmdLine + " --start-konnector true"
+		}
+		logger.V(2).Info(cmdLine)
+		cmd := exec.Command("/bin/sh", "-c", cmdLine)
+		stdout, err := cmd.StdoutPipe()
 		if err != nil {
-			logger.V(4).Info("End of stderr")
-			break
+			logger.Error(err, "Unable to return a pipe for stdout")
```
MikeSpreitzer, line 286: "return" is not what is being attempted here.

cmd/mailbox-controller/controller.go
waltforme: Cool, good to know something new today and happy to see that works as added [here](https://github.com/kubestellar/kubestellar/pull/1301/commits/2d4bfd6c1d4d069ff4c8eec1a5eaed39fbe68af0).

cmd/mailbox-controller/controller.go
waltforme: I copied this from the go doc of the `StderrPipe()` function which says 'returns a pipe'. May I know what should be the better message here?

MikeSpreitzer: The general idea here is that the error message should be written from the perspective of the code using `StderrPipe` rather than the definition of `StderrPipe`. So while the latter talks about what `StderrPipe` "returns", the code using that is not in the business of "returning" the same thing. The error message should be about what problem the _using_ code is having.

cmd/mailbox-controller/controller.go
```diff
@@ -255,76 +258,82 @@ func (ctl *mbCtl) sync(ctx context.Context, refany any) bool {
...
+	resourcesToBind := []string{"syncerconfigs", "edgesyncconfigs"}
+	for idx, resource := range resourcesToBind {
+		logger.V(2).Info("Ensuring binding", "script", shellScriptName, "resource", resource)
 
-	r, _ := regexp.Compile("🔒 (Created|Updated) secret.*namespace (.*)")
-	found, namespaceName := false, ""
-	for {
-		line, _, err := errBuf.ReadLine()
+		// suffix helps isolate this controller's multiple workers to prevent concurrent access of a single kubeconfig file.
+		// This is a compromise due to the situation that 1) we are using kcp and 2) we are running tests on bare process of this controller.
```
MikeSpreitzer, line 269: Part 2 is not important here. Even when run in a container, the mailbox controller may be doing concurrent script invocations using the same filesystem.

cmd/mailbox-controller/controller.go
```diff
@@ -62,6 +63,8 @@ type mbCtl struct {
 	queue                   workqueue.RateLimitingInterface // of mailbox workspace Name
 }
 
+var initialSuffix uint64 = 142857
```
MikeSpreitzer, line 66: But this _variable_ does not hold the _initial_ suffix in general, its job is to hold the suffix.

cmd/mailbox-controller/controller.go
```diff
@@ -255,76 +258,82 @@ func (ctl *mbCtl) sync(ctx context.Context, refany any) bool {
...
-	// There is a contact between the script and the code that this message is the signal for successful bindings
-	r, _ = regexp.Compile("CRDs already in place, returning")
-	established := false
-	for {
-		line, _, err := outBuf.ReadLine()
+		err = cmd.Start()
 		if err != nil {
-			logger.V(4).Info("End of stdout")
-			break
+			logger.Error(err, "Unable to start executing the script")
```
MikeSpreitzer, line 302: Tell the reader _which_ script or what this exec is trying to accomplish.

cmd/mailbox-controller/controller.go
```diff
@@ -255,76 +258,82 @@ func (ctl *mbCtl) sync(ctx context.Context, refany any) bool {
...
+			if err != nil {
+				if err == io.EOF {
+					logger.V(4).Info("End of stdout")
+					break
+				} else {
+					logger.Error(err, "Unable to read stdout")
+					return true
+				}
+			}
+			logger.V(2).Info(string(line))
```
MikeSpreitzer, line 318: I would add an explanation of what this is. For example:
```
logger.V(2).InfoS("Stdout from kube-bind", "resource", resource, "line", line)
```

inner-scripts/kube-bind
```diff
@@ -78,15 +79,28 @@ else
...
 fi
 
+if [ "$start_konnector" == true ]
+then
+    if ! kubectl get namespace kube-system &> /dev/null
+    then kubectl create namespace kube-system &> /dev/null
+    fi
+    echo "starting kube-bind konnector for space $space_name in background"
+    konnector &> /dev/null &
+    sleep 2
```
MikeSpreitzer, line 96: As a general rule, a fixed delay is bad because you never know how overloaded the node may be and thus how long you have to wait. A better approach is a loop that waits until the needed condition is met, possibly giving up after a long time.

MikeSpreitzer (commented):
This is good enough to merge with more work done in follow-on PR(s).
/lgtm
Thanks

cmd/mailbox-controller/controller.go
waltforme: That's crystal clear. I see the point now. Changed in latest push.

cmd/mailbox-controller/controller.go
waltforme: Changed in latest push.

cmd/mailbox-controller/controller.go
waltforme: Changed in latest push.

cmd/mailbox-controller/controller.go
waltforme: Changed in latest push.

cmd/mailbox-controller/controller.go
waltforme: Changed in latest push.

inner-scripts/kube-bind
waltforme: Yes. This issue is also related to another issue, which is there might be multiple konnectors, each for one space. I need to figure out which one is the one just started. Will address this in another PR.

cmd/mailbox-controller/controller.go
```diff
@@ -255,76 +258,82 @@ func (ctl *mbCtl) sync(ctx context.Context, refany any) bool {
...
+			removeCopyWhenExits,
+			invokeScript,
+		}, "; ")
+		logger.V(2).Info("About to exec", "cmdLine", cmdLine)
+		cmd := exec.Command("/bin/sh", "-c", cmdLine)
+		stdout, err := cmd.StdoutPipe()
 		if err != nil {
-			logger.V(4).Info("End of stderr")
-			break
+			logger.Error(err, "Failed to extract stdout pipe for exec of script", "script", shellScriptName)
```
MikeSpreitzer, line 292: It would be better if this message also identified the resource involved.

MikeSpreitzer (commented):
This is good enough to merge.
/lgtm
/approve
Thanks!

PR #1306, opened 2023-11-20
:sparkles: Kube-bind provider-consumer object relationship

andreyod:
## Summary
1. In mailbox-controller construct a mailbox space name from the original SyncTarget object
2. In where-resolver controller create and update SPS in the WMW space and not in the ESPW.
3. In placement-translator TBD

## Related issue(s)

Fixes #

### DIFF
--- a/cmd/kubestellar-where-resolver/cmd/where-resolver-cmd.go
+++ b/cmd/kubestellar-where-resolver/cmd/where-resolver-cmd.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/component-base/version"
 	"k8s.io/klog/v2"
 
@@ -31,6 +32,7 @@ import (
 	edgeclientset "github.com/kubestellar/kubestellar/pkg/client/clientset/versioned"
 	edgeclusterclientset "github.com/kubestellar/kubestellar/pkg/client/clientset/versioned/cluster"
 	edgeinformers "github.com/kubestellar/kubestellar/pkg/client/informers/externalversions"
+	"github.com/kubestellar/kubestellar/pkg/kbuser"
 	wheresolver "github.com/kubestellar/kubestellar/pkg/where-resolver"
 )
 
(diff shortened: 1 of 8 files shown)


cmd/kubestellar-where-resolver/cmd/where-resolver-cmd.go
```diff
@@ -104,13 +106,20 @@ func Run(ctx context.Context, options *resolveroptions.Options) error {
 		logger.Error(err, "failed to create edge clientset for controller")
 		return err
 	}
+	serviceProviderClient, err := kubernetes.NewForConfig(espwRestConfig)
```
MikeSpreitzer, line 109: I would not use the words "service provider" in this variable name. That only identifies its significance in the theory of kube-bind and `kbuser.NewKubeBindSpaceRelation`. It would be better if the variable name more specifically identified _which_ service provider it is talking about. That would be the KCS (née ESPW). Even better, since this is not the only clientset for that space, would be for the variable to identify which API group(s) _this_ clientset is for. Perhaps something like `kubeClient` (and perhaps change `espwClientset` above to `edgeClient`).

MikeSpreitzer:
Has the where-resolver been updated to produce SinglePlacement structs according to https://github.com/kubestellar/kubestellar/blob/5fe587ecf309234a4f23cb8c97b5171d5efac364/docs/content/Coding%20Milestones/PoC2023q1/outline.md?plain=1#L778-L781 ?

MikeSpreitzer:
Please see #1308

cmd/kubestellar-where-resolver/cmd/where-resolver-cmd.go
andreyod: Changed

andreyod:
> Has the where-resolver been updated to produce SinglePlacement structs according to
> 
> https://github.com/kubestellar/kubestellar/blob/5fe587ecf309234a4f23cb8c97b5171d5efac364/docs/content/Coding%20Milestones/PoC2023q1/outline.md?plain=1#L778-L781
> 
> ?

Yes

PR #1307, opened 2023-11-21
🌱 Clarify single placement

MikeSpreitzer:
## Summary
The PR updates the comments on the fields of SinglePlacement to specify which of the two copies of the objects involved are identified. This is documenting an agreement already reached and appearing in https://github.com/kubestellar/kubestellar/blob/5fe587ecf309234a4f23cb8c97b5171d5efac364/docs/content/Coding%20Milestones/PoC2023q1/outline.md?plain=1#L778-L781 but somehow still not widely known.

## Related issue(s)

Fixes #


MikeSpreitzer:
wrong base.

PR #1330, opened 2023-11-28
🐛 fixed make codegen to work correctly in case of deleteion from pkg/apis

nirrozenbaum:
## Summary

if one of the apis defined under `pkg/apis/` is deleted, current command `make codegen` leaves garbage generated code. 
This can be easily solved by removing generated code under `pkg/client` before the code generation.

in order to test this fix I tried the following:
1. deleted syncer config api file.
2. ran `make codegen`.

BEFORE this fix, the following screenshot is the result:
Screenshot 2023-11-28 at 15 24 54

AFTER this fix, I ran the same and the following screenshot is the result:
Screenshot 2023-11-28 at 15 27 26

as it can be seen in the screenshots, if this fix is not introduced, a lot of unnecessary generated code is still there even if we delete a file from the api.

## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -248,6 +248,7 @@ crds: $(CONTROLLER_GEN) $(API_GEN) $(YAML_PATCH)
 .PHONY: crds
 
 codegen: crds $(CODE_GENERATOR)
+	rm -rf pkg/client/*
 	go mod download
 	./hack/update-codegen-clients.sh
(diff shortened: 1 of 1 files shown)


MikeSpreitzer (commented):
/lgtm
/approve
Thanks!

PR #1333, opened 2023-11-28
✨ remove some instances of kcp from non-active doc

clubanderson:
## Summary
first step of many

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/Contribution guidelines/operations/packaging-and-versioning.md
+++ b/docs/content/Contribution guidelines/operations/packaging-and-versioning.md
@@ -56,8 +56,8 @@ Following are some recent tags and what they correspond to.
 
 ## Core container image
 
-The core container runs the kcp server and the KubeStellar core
-controllers and initializes the kcp workspaces. It is built from some
+The core container runs a space provider server and the KubeStellar core
+controllers and initializes the space provider's spaces. It is built from some
 contents of [the KubeStellar GitHub
 repo](https://github.com/kubestellar/kubestellar), specifically
 excluding some that are only needed by users outside this
@@ -79,7 +79,7 @@ Following are some recent tags and what they correspond to.
 ## KubeStellar core Helm chart
 
 This Helm chart describes a deployment of the KubeStellar core
-components (kcp server, KubeStellar controllers, initialization) as
+components (a space provider server, KubeStellar controllers, initialization) as
 workload for a Kubernetes cluster.
 
 The source of this Helm chart is in the `core-helm-chart` subdirectory
@@ -137,9 +137,8 @@ KubeStellar. According to homebrew convention, users refer to it in
 homebrew as `kubestellar/kubestellar` (e.g., `brew tap
 kubestellar/kubestellar` connects directly to this GitHub repository).
 
-This HomeBrew repository has two formulae.  One called `kcp_cli` has
-the kubectl plugins of kcp. Release v0.11.0 of kcp is always used. A
-formula called `kubestellar_cli` has contents that are needed by a
+This HomeBrew repository has two formulae.  One to control the space
+provider and another formula called `kubestellar_cli` containing tools needed by a
 user who deploys the KubeStellar core as workload in a Kubernetes
 cluster or another user of that deployment. This formula fetches and
 installs a `kubestellaruser` release archive from GitHub. Users invoke
@@ -148,7 +147,6 @@ formulae.
 
 Following are the versions available.
 
-- `kcp_cli` `v0.11.0` for that release of kcp
 - `kubestellar_cli` `v0.9.0`, for that release of KubeStellar.
 
 ### kubestellar/kubeflex
--- a/docs/content/Contribution guidelines/CONTRIBUTING.md
+++ b/docs/content/Contribution guidelines/CONTRIBUTING.md
@@ -69,7 +69,7 @@ If you have any questions about contributing, don't hesitate to reach out to us 
 
 Our [QuickStart](../../Getting-Started/quickstart/)
  guide shows a user how to install a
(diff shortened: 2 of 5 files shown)


clubanderson:
@KPRoche can you review?

docs/content/Contribution guidelines/operations/packaging-and-versioning.md
```diff
@@ -56,8 +56,8 @@ Following are some recent tags and what they correspond to.
 
 ## Core container image
 
-The core container runs the kcp server and the KubeStellar core
-controllers and initializes the kcp workspaces. It is built from some
+The core container runs a space provider server and the KubeStellar core
```
MikeSpreitzer, line 59: The proposed new language here is not accurate for providers other than kcp. For kind there isn't even a "server", and KubeFlex has its own container image.

docs/content/Contribution guidelines/operations/packaging-and-versioning.md
```diff
@@ -56,8 +56,8 @@ Following are some recent tags and what they correspond to.
...
+controllers and initializes the space provider's spaces. It is built from some
```
MikeSpreitzer, line 60: I think it would be better to say more specifically that it is the spaces used by KubeStellar that are being initialized, or perhaps say that spaces are created and initialized for use by KubeStellar.

docs/content/Contribution guidelines/operations/packaging-and-versioning.md
```diff
@@ -79,7 +79,7 @@ Following are some recent tags and what they correspond to.
 ## KubeStellar core Helm chart
 
 This Helm chart describes a deployment of the KubeStellar core
-components (kcp server, KubeStellar controllers, initialization) as
+components (a space provider server, KubeStellar controllers, initialization) as
```
MikeSpreitzer, line 82: I would remove the word "server" here: "space provider" logically covers all of its machinery, including a server if there is one.

docs/content/Contribution guidelines/operations/packaging-and-versioning.md
```diff
@@ -137,9 +137,8 @@ KubeStellar. According to homebrew convention, users refer to it in
 homebrew as `kubestellar/kubestellar` (e.g., `brew tap
 kubestellar/kubestellar` connects directly to this GitHub repository).
 
-This HomeBrew repository has two formulae.  One called `kcp_cli` has
-the kubectl plugins of kcp. Release v0.11.0 of kcp is always used. A
-formula called `kubestellar_cli` has contents that are needed by a
+This HomeBrew repository has two formulae.  One to control the space
```
MikeSpreitzer, line 140: I think it is better to leave the specificity here, since the brew formula in question really is specific and the point of this document is to talk about dependencies so specificity is important here.

docs/content/Contribution guidelines/operations/packaging-and-versioning.md
```diff
@@ -148,7 +147,6 @@ formulae.
 
 Following are the versions available.
 
-- `kcp_cli` `v0.11.0` for that release of kcp
```
MikeSpreitzer, line 151: Similar comment as above.

clubanderson: same as comment above

docs/content/Contribution guidelines/operations/packaging-and-versioning.md
clubanderson: removed 'server' - my mistake

docs/content/Contribution guidelines/operations/packaging-and-versioning.md
clubanderson: I think this is a bit too specific, but added language to make it so

docs/content/Contribution guidelines/operations/packaging-and-versioning.md
MikeSpreitzer: The problem here is not just the word "server". KubeFlex has its own machinery, including container image and Pod that runs it.

BTW, this is also entangled with a problem that existed before this PR. The text "The core container runs the kcp server and the KubeStellar core containers" is a bit mixed up.  For example, it is saying "The core container runs ... containers". That is not right.  That text confuses container _image_ with _container_ and suggests (with the singular "the") that there is only one core container. We currently run five core containers, all using one container image: kcp server, init, and three controllers.

Here is a suggested revision that is accurate:

> The core container image is used in containers that run the KubeStellar core containers, initialize the KubeStellar core, and run one of the supported space providers.

docs/content/Contribution guidelines/operations/packaging-and-versioning.md
MikeSpreitzer: The brew formula for kcp is not obviated by the change in Helm charting; the brew formula is supplying executables that clients need.

clubanderson: I am not adding this reference into our doc.  it is unnecessary

docs/content/Contribution guidelines/operations/packaging-and-versioning.md
clubanderson: right - changed

docs/content/Contribution guidelines/operations/packaging-and-versioning.md
MikeSpreitzer: We are talking about a reference that is already in the doc. You are insisting on removing it. I disgree with removing it, but think I have more productive things to spend my time on regarding the word count. So, other than this,
/lgtm

PR #1338, opened 2023-11-29
🌱 Minor fixes in response to review comments to 1265

waltforme:
## Summary
Same as the title.

### DIFF
--- a/scripts/outer/kubectl-kubestellar-prep_for_syncer
+++ b/scripts/outer/kubectl-kubestellar-prep_for_syncer
@@ -134,7 +134,7 @@ spaceID=$(kubectl "${kubectl_flags[@]}" get synctargets.edge.kubestellar.io "$st
 
 kubectl ws "${kubectl_flags[@]}" "$espw"
 if ! cluster_ns_name="$(kubectl "${kubectl_flags[@]}" -n kubestellar get configmap kbmap-"$spaceID" -ojsonpath="{.metadata.labels['kubestellar\.io\/kube-bind-id']}")"; then
-	echo "error reading kube-bind cluster namespace from the configmap"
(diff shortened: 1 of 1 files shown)


scripts/outer/kubectl-kubestellar-prep_for_syncer
```diff
@@ -177,15 +177,11 @@ echo "now that workspace _exists_, but is it _ready_?"
...
-	kubectl ws .
-	echo "tried once to switch to $mbwsname"
+    echo "tried once to switch to $mbwsname"
     sleep 15
     if ! kubectl ws "${kubectl_flags[@]}" "$mbwsname" &> /dev/null; then
-	kubectl ws .
-	echo "tried twice to switch to $mbwsname"
-	sleep 15
-	kubectl ws .
-	kubectl ws "${kubectl_flags[@]}" "$mbwsname"
```
MikeSpreitzer, line 188: Did you really intend to delete this line? It was deliberately put there to try one last time and err out if unsuccessful.

waltforme: If the last (could be either the 3rd, like in the current revision; or the 2nd, in the changed revision) try fails, we don't want to proceed, instead we want to be aware of the failure and explicitly exit, right?

MikeSpreitzer: Yes, we want the script to fail so that the caller is aware that there is a problem. It is not necessary for the script to explicitly say `exit`; any command that fails will cause an exit because of the earlier `set -e`.
The original version of this script deliberately tries three times.

MikeSpreitzer (commented):
/lgtm
/approve
Thanks!

PR #1340, opened 2023-11-30
✨ More changes in response to review comments to 1265

waltforme:
## Summary
As show in the title.

/cc @MikeSpreitzer

### DIFF
--- a/scripts/inner/kubestellar-kube-bind
+++ b/scripts/inner/kubestellar-kube-bind
@@ -18,6 +18,7 @@ space_name=""
 resource_to_bind=""
 start_konnector=false
 cm_namespace=kubestellar # the namespace where the configmaps live in on the provider side
+log_dir=$(pwd)/kubestellar-kube-bind-logs
 
 while (( $# > 0 )); do
     case "$1" in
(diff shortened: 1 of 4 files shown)


scripts/inner/kubestellar-kube-bind
```diff
@@ -85,17 +86,22 @@ if kubectl get crd $resource_to_bind.edge.kubestellar.io &> /dev/null; then
     exit 0
 fi
 
+if [[ ! -d $log_dir ]]; then
```
MikeSpreitzer, line 89: Isn't this what the `-p` flag on `mkdir` means?

scripts/inner/kubestellar-kube-bind
```diff
@@ -18,6 +18,7 @@ space_name=""
 resource_to_bind=""
 start_konnector=false
 cm_namespace=kubestellar # the namespace where the configmaps live in on the provider side
+log_dir=$(pwd)/kubestellar-kube-bind-logs
```
MikeSpreitzer, line 21: Is there any promise what is the current working directory at this point? Any promise that it is appropriate for putting logs into?

scripts/inner/kubestellar-kube-bind
waltforme: Removed this in #1351.

PR #1351, opened 2023-12-01
 ✨ Make kubestellar-kube-bind idempotent

waltforme:
## Summary
As shown in the title.

### DIFF
--- a/scripts/inner/kubestellar-kube-bind
+++ b/scripts/inner/kubestellar-kube-bind
@@ -16,24 +16,18 @@
 
 space_name=""
 resource_to_bind=""
-start_konnector=false
 cm_namespace=kubestellar # the namespace where the configmaps live in on the provider side
 log_dir=$(pwd)/kubestellar-kube-bind-logs
 
 while (( $# > 0 )); do
     case "$1" in
     (-h|--help)
-        echo "Usage: $0 (--start-konnector boolean | -X)* space_name resource_to_bind"
+        echo "Usage: $0 [-X] space_name resource_to_bind"
         echo "space_name is the name of a kube-bind consumer space, the space currently must be a direct child of the kcp 'root' workspace, note that space_name doesn't contain the 'root:' prefix"
         echo "resource_to_bind is a resource offered by the kube-bind provider, in lower case and plural form"
         exit 0;;
     (-X)
         set -o xtrace;;
-    (--start-konnector)
-        if (( $# >1 ))
-        then start_konnector="$2"; shift
-        else echo "$0: missing start-konnector value" >&2; exit 1
-        fi;;
     (-*)
         echo "$0: flag syntax error" >&2
         exit 1;;
@@ -60,12 +54,6 @@ then
     exit 1
 fi
 
-case "$start_konnector" in
-    (true|false) ;;
-    (*) echo "$0: start-konnector should be true or false" >&2
-	  exit 1;;
-esac
-
 initial_ws=$(kubectl ws . --short)
 trap "kubectl ws use $initial_ws &> /dev/null # go back to the initial state" EXIT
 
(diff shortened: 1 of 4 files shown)


scripts/inner/kubestellar-kube-bind
```diff
@@ -86,21 +74,18 @@ if kubectl get crd $resource_to_bind.edge.kubestellar.io &> /dev/null; then
     exit 0
 fi
 
-if [[ ! -d $log_dir ]]; then
-    mkdir -p $log_dir
-fi
-
-if [ "$start_konnector" == true ]
-then
+if pgrep konnector | xargs ps e -p 2>/dev/null | grep "KB_CONSUMER=$space_name" &> /dev/null; then
```
MikeSpreitzer, line 77: (1) What if one space name is a prefix of another? You could add a terminal character that is valid in an envar value but not in a Space name.

(2) I think you need some `w`s on that `ps` command:
```
mspreitz@mjs-dev7a:~$ ps e -p 1146462
    PID TTY      STAT   TIME COMMAND
1146462 pts/0    Ss     0:00 -bash LANG=C.UTF-8 USER=mspreitz LOGNAME=mspreitz H

mspreitz@mjs-dev7a:~$ ps ew -p 1146462
    PID TTY      STAT   TIME COMMAND
1146462 pts/0    Ss     0:00 -bash LANG=C.UTF-8 USER=mspreitz LOGNAME=mspreitz HOME=/home/mspreitz PATH=/usr/local/sbin:/usr/local/b

mspreitz@mjs-dev7a:~$ ps eww -p 1146462
    PID TTY      STAT   TIME COMMAND
1146462 pts/0    Ss     0:00 -bash LANG=C.UTF-8 USER=mspreitz LOGNAME=mspreitz HOME=/home/mspreitz PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin SHELL=/bin/bash TERM=xterm-256color XDG_SESSION_ID=531 XDG_RUNTIME_DIR=/run/user/1002 DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1002/bus XDG_SESSION_TYPE=tty XDG_SESSION_CLASS=user MOTD_SHOWN=pam SSH_CLIENT=9.67.15.143 58072 22 SSH_CONNECTION=9.67.15.143 58072 9.59.199.171 22 SSH_TTY=/dev/pts/0
```

On Ubuntu and Darwin, two `w` means unlimited width.

(3) On Darwin, you need `E` rather than `e` to get the environment displayed.

scripts/inner/kubestellar-kube-bind
```diff
@@ -86,21 +74,18 @@ if kubectl get crd $resource_to_bind.edge.kubestellar.io &> /dev/null; then
     exit 0
 fi
 
-if [[ ! -d $log_dir ]]; then
-    mkdir -p $log_dir
-fi
-
-if [ "$start_konnector" == true ]
-then
+if pgrep konnector | xargs ps e -p 2>/dev/null | grep "KB_CONSUMER=$space_name\s" &> /dev/null; then
```
MikeSpreitzer, line 77: What if KB_CONSUMER is the last variable listed on the line?

```
mspreitz@mjs13 kubestellar % grep 'foo\s'
bar
foo
doo
```

waltforme: I saw that this `KB_CONSUMER` always be the 1st in my tests, and it can't be the last because there are `PATH`, `PWD` etc.

MikeSpreitzer: On my Mac the environment variables are not listed in alphabetical order.
I think it would be best to not rely on assumptions about how `ps` orders that listing.

waltforme: Addressed in latest push.

PR #1353, opened 2023-12-01
✨ Introduce word counting utility

MikeSpreitzer:
## Summary
This PR introduces a utility that counts uses of "kcp", broken down by directory and file type. It is intended to compile a record over time that can be examine for evolution.

## Related issue(s)

Fixes #


MikeSpreitzer:
Now it can count a given commit, and it does so without doing a `git clone`.

PR #1354, opened 2023-12-01
✨ Shift left the labeling of edge CRDs for kube-bind to export

waltforme:
## Summary
This PR uses some more kubebuilder markers to label the edge CRDs earlier for them to be later exported via kube-bind.

### DIFF
--- a/config/crds/edge.kubestellar.io_customizers.yaml
+++ b/config/crds/edge.kubestellar.io_customizers.yaml
@@ -3,8 +3,9 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.10.0
-  creationTimestamp: null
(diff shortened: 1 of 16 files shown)


config/crds/edge.kubestellar.io_customizers.yaml
```diff
@@ -3,8 +3,9 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.10.0
-  creationTimestamp: null
```
MikeSpreitzer, line 7: Why was this line here (and in the other YAMLs)? Does removing it break something?

waltforme: As far as what I saw in local test and the CI, I didn't see it break anything. I think it is introduced by the more recent version of kubebuilder.

waltforme: Yes

PR #1392, opened 2023-12-06
✨ Use exact match when looking for dex process

waltforme:
## Summary
As shown in the title.

### DIFF
--- a/scripts/inner/kubestellar
+++ b/scripts/inner/kubestellar
@@ -48,7 +48,13 @@ function echoerr() {
 # Check if a given process name is running
 process_running() {
   SERVICE="$1"
-  if pgrep -f "$SERVICE" >/dev/null
+  NAME_LEN=${#SERVICE}
+  if [ $NAME_LEN -gt 15 ]; then
+    PGREP_FLAG="-f"
+  else
+    PGREP_FLAG="-x"
+  fi
+  if pgrep $PGREP_FLAG "$SERVICE" >/dev/null
   then
       echo "running"
   else
@@ -381,7 +387,6 @@ function run_dex() {
         echo "dex started"
     else
         echo "dex not started"
-        sleep 2
         exit 1
     fi
 }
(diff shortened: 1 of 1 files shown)


scripts/inner/kubestellar
```diff
@@ -375,13 +375,12 @@ function run_dex() {
         echo "Error looking for dex config file at $dex_config_file" >&2
     fi
     dex serve $dex_config_file &
+    sleep 1
 
-    run_status=$(wait_for_process dex)
-    if [ $run_status -eq 0 ]; then
+    if pgrep -x dex &> /dev/null; then
```
MikeSpreitzer, line 380: Wouldn't it be better to fix and call wait_for_process ?

waltforme: Tried before opening the PR. But we'd better not to use `-x` there because wait_for_process is used by mailbox-controller etc which have longer names (>15 chars) and pgrep's man pages says

```
       The  process  name  used  for  matching is limited to the 15 characters present in the output of
       /proc/pid/stat.  Use the -f option to match against the complete  command  line,  /proc/pid/cmd‐
       line.
```
Maybe the `-f` flag is there for this reason.

waltforme: `grep -x` won't work for KS controllers unless we truncate like this
```
ubuntu@ip-172-31-67-225:~/kubestellar$ pgrep -x mailbox-controller
ubuntu@ip-172-31-67-225:~/kubestellar$ pgrep -x mailbox-control
1853417
ubuntu@ip-172-31-67-225:~/kubestellar$ 
```

MikeSpreitzer: What I am suggesting is to use a wait function that does exact match for dex, continue using the old wait function for controllers with distinctive names.

waltforme: I see. Tried in latest push.

MikeSpreitzer: done

PR #1412, opened 2023-12-07
🐛 Fix `kubectl-kubestellar-ensure-location`

waltforme:
## Summary
With an idempotent script for kube-bind, we can now addresses https://github.com/kubestellar/kubestellar/issues/1337#issuecomment-1832146503.

## Related issue(s)

Closes #1337

PR #1415, opened 2023-12-08
🌱 Redo1412

MikeSpreitzer:
## Summary
This PR repeats #1412, undoing the undo in #1414.

## Related issue(s)

Fixes #


MikeSpreitzer:
This PR does introduce a problem now. When a user, outside the hosting cluster, issues a `kubectl kubestellar prep-for-cluster` command, that invokes `kubectl kubestellar ensure location` outside the hosting cluster, which in turn invokes `kubestellar-kube-bind` (e.g., for imw1) outside the hosting cluster --- which does not notice that there is already a konnector running inside the hosting cluster and thus tries to run another konnector. That fails because user-quickstart-test has not put `konnector` on the user's `$PATH`.

MikeSpreitzer:
Moot now.

PR #1437, opened 2023-12-08
🐛 Replace secret delete;create with apply

### LINKED ISSUE #1429: bug: Concurrency bugs wrt Secret objects
### Describe the bug

In `core-container/entry.sh` there are two places where the script is trying to make sure a Secret object exists with certain contents, and starts by deleting that Secret object if it exists. The first is in create_kcp_provider_object, and it surely is a concurrency bug. Considering various restart scenarios, it is possible for a concurrent process to be trying to read that Secret while it is being deleted. The other is in run_kcp, and that is likely a concurrency bug too. This pattern is also used in `space-framework/entry.sh`.

### Steps To Reproduce

Look at the cited scripts.

### Expected Behavior

Tolerating various crash and restart scenarios is expected of cloud-native code.

### Additional Context

Originally mentioned in https://github.com/kubestellar/kubestellar/pull/1363#pullrequestreview-1766204621

MikeSpreitzer:
## Summary
This PR replaces scripting that ensures the existence and content of a Secret object by deleting it and then creating it with using `kubectl apply`. This is good because concurrent processes can exist and get trouble from the deletion.

## Related issue(s)

Fixes #1429

### DIFF
--- a/core-container/entry.sh
+++ b/core-container/entry.sh
@@ -18,6 +18,10 @@ set -e
 
 KUBESTELLAR_SERVICE="kubestellar"
 
+if base64 -w 0 <<<"" &>/dev/null
+then nolbs='-w 0'
+else nolbs=''
+fi
 
 function echoerr() {
    echo "ERROR: $1" >&2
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
In testing on my laptop:

```bash
...
.+++ KUBECONFIG=/Users/mspreitz/.kube/config
+++ kubectl get pod --selector=app=kubestellar -o 'jsonpath={.items[0].metadata.name}' -n kubestellar
++ KUBECONFIG=/Users/mspreitz/.kube/config
++ kubectl exec kubestellar-69d6d997b8-7m8p8 -n kubestellar -c init -- ls /home/kubestellar/ready
++ sleep 10
^Cmake: *** [docs-ecutable] Interrupt: 2

mspreitz@mjs13 kubestellar % kubectl get pod --selector=app=kubestellar
No resources found in default namespace.

mspreitz@mjs13 kubestellar % kubectl get pod --selector=app=kubestellar -A
NAMESPACE     NAME                           READY   STATUS             RESTARTS      AGE
kubestellar   kubestellar-69d6d997b8-7m8p8   5/6     CrashLoopBackOff   4 (37s ago)   3m15s

mspreitz@mjs13 kubestellar % 
```

MikeSpreitzer:
The force-push to f1242a992 fixes stupid typos.

MikeSpreitzer:
I canceled the user-quickstart-test because it was taking forever, still not finding kubestellar ready.

MikeSpreitzer:
The force-push to 1665a1685 fixes another stupid typo.

waltforme:
The changes made here all looks good to me.

I just want to confirm that I understand the bug #1429 itself:
If container A restarts after a crash, it may `run_init` or `run_kcp`. During the runs, there is a duration after a secret is deleted and before the secrete is recreated. If a read from container B happens within this duration then that read is doomed to fail.
Is my understanding correct?

Issue #1438, opened 2023-12-08
bug: captuing pod's bearer token is a time bomb


MikeSpreitzer:
### Describe the bug

https://github.com/kubestellar/kubestellar/blob/d22d4272d1571ccc9f6ad836bb5a5e0b02605c22/core-container/entry.sh#L264 captures for use forevermore a bearer token that will expire within an hour

### Steps To Reproduce

read the code

### Expected Behavior

The system stays up for months

ezrasilvera:
@MikeSpreitzer what is your recommendation instead?

MikeSpreitzer:
I recommend that for scripting that runs in a pod in the host cluster, kube client programs (`kubectl`, controllers) wanting to access the host cluster/ SMAS be given no `--kubeconfig` and `$KUBECONFIG` be empty, so that the regular behavior of client-go in that case happens --- which uses the _current_ bearer token mounted and rotated by the kubelet.

MikeSpreitzer:
The problem in the KubeStellar `entry.sh` has been fixed in #1444 .
The problem also appears in the Space Framework's `entry.sh`.

MikeSpreitzer:
The space-manager process should take kubectl arguments in the regular way, using https://github.com/kubernetes/cli-runtime/blob/v0.24.3/pkg/genericclioptions/config_flags.go#L414 , like https://github.com/kubestellar/kubestellar/blob/main/cmd/kube-watchall/main.go does.

MikeSpreitzer:
Moot

PR #1363, opened 2023-12-04
:sparkles: Switch to use the Space Framework

ezrasilvera:
This PR replaces #1286.   It is still WIP until all changes are pushed into the space-mgt branch

It includes changing all the scripts, tests, and KS code to use the SF instead of KCP workspaces.
It also switch from using a "cluster ID" to using "space name"

## Related issue(s)

Fixes #

PR #1444, opened 2023-12-09
🌱 Revamp commands

MikeSpreitzer:
## Summary
This PR revises the design of the scripts of KubeStellar for simplicity, regularity, and avoiding #1438 and #1447 .

This work is incomplete. The following changes have been done.

- Introduce subcommand`kubectl kubestellar space` into the tree, to head a branch of commands for manipulating spaces.
- Rename `kubectl kubestellar get config for space` to `kubectl kubestellar space get-kubeconfig`, to keep all the space subcommands in a branch of the subcommand tree. Make it take kube client config for SMAS access like a regular kubectl plugin. Change required parameters from flag to positional.
- Rename `kubectl kubestellar create space` to `kubectl kubestellar space ensure`, for tree discipline and because this is "ensure exists" rather than "create". Make it take kube client config for SMAS access like a regular kubectl plugin. Change required parameter from flag to positional. Introduce an option to write a space kubeconfig file. Document this command.
- Make `kubestellar-kube-bind` take a kube client config for SMAS access like kubectl. Changed its convention for where konnector log and kubeconfig are stored.
- Make `kubectl kubestellar ensure wmw` take a kube client config for SMAS access like a regular kubectl plugin. Move this script from `scripts/overlap/` to `scripts/inner/` because it only works "in cluster".
- Introduce `kubectl kubestellar ensure inventory-space`.
- Update `kubectl kubestellar ensure location` to take the inventory space kubeconfig in the regular way for kubectl.
- Update `kubectl kubestellar prep-for-syncer` and `kubectl kubestellar prep-for-cluster` to take SMAS client config in the regular way for kubectl.
- For commands that have an ephemeral need for a space kubeconfig file: use `mktemp` to make a filename and `rm` it at the end of the script (thus script crashes leave this file around for inspection).
- Update the `kubestellar` script and its usage in example1.

Planned additional changes include the following.

- Update the core Helm chart's entry.sh
- Update user-quickstart-test
- Fix commands that #1363 left broken.

## Related issue(s)

Fixes #


scripts/inner/kubestellar-kube-bind
```diff
@@ -14,22 +14,37 @@
...
+
 cm_namespace=kubestellar # the namespace where the configmaps live in on the provider side
 log_dir=$(pwd)/kubestellar-kube-bind-logs
 
 while (( $# > 0 )); do
     case "$1" in
     (-h|--help)
-        echo "Usage: $0 [-X] space_name resource_to_bind"
+        echo "Usage: $0 (-h | --help | -X | \$kubectl_flag)* space_name resource_to_bind"
         echo "space_name is the name of a kube-bind consumer space, the space currently must be a direct child of the kcp 'root' workspace, note that space_name doesn't contain the 'root:' prefix"
```
ezrasilvera, line 32: This comment is correct only for KCP, I suggest to remove it completely. We explain in the create-space that for KCP , space name is represents a workspace under root, with path "root:spacename"

scripts/inner/kubestellar-kube-bind
```diff
@@ -56,15 +71,28 @@ then
...
 fi
 
-conf_file="/tmp/$space_name.conf"
-kubectl-kubestellar-get-config-for-space --space-name $space_name --sm-core-config $SM_CONFIG --sm-context $SM_CONTEXT --in-cluster --output $conf_file
-KUBECONFIG=$conf_file
+# Some place protectable to store kubeconfig files.
+# We do not have promises about what the $PWD is.
+# Some users administer multiple KubeStellar systems,
+# so the name must not be a pure function of the space name.
+konnect_dir="${HOME}/kubestellar-konnectors/$(date +%y%m%d-%H%M%S)-${space_name}-${resource_to_bind}-$$"
```
ezrasilvera, line 78: Adding resource_to_bind is not reasonable. The konnector is using a config matching a specific space, and is used for all resources bounded into that space.  
Also, the konnector process is tagged and validated using env variable  that uses only the space name so a new KS instance using the same space name will mistakenly  avoid creating a konnector .

scripts/inner/kubestellar-kube-bind
```diff
@@ -56,15 +71,28 @@ then
     exit 1
 fi
 
-conf_file="/tmp/$space_name.conf"
-kubectl-kubestellar-get-config-for-space --space-name $space_name --sm-core-config $SM_CONFIG --sm-context $SM_CONTEXT --in-cluster --output $conf_file
-KUBECONFIG=$conf_file
+# Some place protectable to store kubeconfig files.
+# We do not have promises about what the $PWD is.
+# Some users administer multiple KubeStellar systems,
+# so the name must not be a pure function of the space name.
```
ezrasilvera, line 77: This is the first time I see requirement of running multiple KS instances in parallel, I believe this is not supported by bare-processes, and I'm not sure it was ever tested for QS tests (that use  some local files as well ...)

scripts/inner/kubestellar-kube-bind
MikeSpreitzer: In the case of the Helm chart there is a fixed usage pattern, this is easier.

I was thinking also of the bare processes case. But I forgot that bare process deployment is only for contributors doing development and testing --- not for users administering possibly multiple KubeStellar deployments. So it _is_ easier than I was thinking.

I am going to revise this further to push responsibility up to the caller (kubectl kubestellar ensure wmw and is).

scripts/inner/kubestellar-kube-bind
MikeSpreitzer: Yes, this is outdated and needs to be fixed.

scripts/inner/kubestellar-kube-bind
MikeSpreitzer: I _know_ that @clubanderson is managing multiple MCAD/kubeflow stacks.

scripts/inner/kubestellar-kube-bind
MikeSpreitzer: Fixed in the force-push to ebfb6293b

scripts/inner/kubestellar-kube-bind
MikeSpreitzer: On second thought, I am not going to move that responsibility up the stack, because the mailbox-controller is not currently dealing in kubeconfig files for the MBSes.

MikeSpreitzer:
There is also a lot of bug fixing in here for old bugs, I do not understand how they did not cause CI failure.

MikeSpreitzer:
Current problem in example1 is that kubestellar-list-syncing-objects has not been updated.  Following is the end of a try.

```
...
++ echo '* CronTab my-new-cron-object is in its mailbox'
* CronTab my-new-cron-object is in its mailbox
++ kubestellar-list-syncing-objects --api-group apps --api-kind ReplicaSet
E1212 03:39:58.887378   35818 main.go:94] "failed to make parent client config" err="context \"root\" does not exist"
make: *** [docs-ecutable] Error 2
```

MikeSpreitzer:
Oh, right. kubestellar-list-syncing-objects needs to be given the KCP_KUBECONFIG.

docs/content/Getting-Started/user-quickstart-test.md
```diff
@@ -173,12 +173,12 @@ echo The mailbox for ks-edge-cluster1 is $MB1
 MB2=$(KUBECONFIG=$SM_CONFIG kubectl get spaces -n spaceprovider-default -o json | jq -r '.items | .[] | .metadata | select(.annotations ["edge.kubestellar.io/sync-target-name"] == "ks-edge-cluster2") | .name')
 echo The mailbox for ks-edge-cluster2 is $MB2
 
-MB1_SPACE="${PWD}/temp-space-config/${MB1}"
-kubectl-kubestellar-get-config-for-space --space-name $MB1 --sm-core-config $SM_CONFIG --sm-context $SM_CONTEXT --output $MB1_SPACE
-MB2_SPACE="${PWD}/temp-space-config/${MB2}"
-kubectl-kubestellar-get-config-for-space --space-name $MB2 --sm-core-config $SM_CONFIG --sm-context $SM_CONTEXT --output $MB2_SPACE
-WMW1_SPACE_CONFIG="${PWD}/temp-space-config/spaceprovider-default-wmw1"
-kubectl-kubestellar-get-config-for-space --space-name wmw1 --sm-core-config $SM_CONFIG --sm-context $SM_CONTEXT --output $WMW1_SPACE_CONFIG
+MB1_KUBECONFIG="${PWD}/${MB1}.kubeconfig"
```
effi-ofer, line 176: This is a general comment that I'll put here.  These generated files are not always cleaned up.  Sometimes we run these tests on our own machine and then have to clean things up manually.  If the generated files go into a subdirectory the cleanup is easier.  Can we decide on a convention where the stored kubeconfig files are in a sub dir?

effi-ofer: Or in tmp.

scripts/inner/kubectl-kubestellar-ensure-wmw
```diff
@@ -98,25 +79,27 @@ esac
...
+		kubectl --kubeconfig "$wds_kubeconfig" apply --server-side=true -R -f "$bindir/../config/kube/crds/"
 	else
-		kubectl --kubeconfig $space_config "${kubectl_flags[@]}" delete -R -f "$bindir/../config/kube/crds/"
+		kubectl --kubeconfig "$wds_kubeconfig" delete -R -f "$bindir/../config/kube/crds/"
 	fi
 }
 
 reconcile_kube_CRDs
+
+if [ "$kc_is_temp" == true ]
```
effi-ofer, line 103: Perhaps switch to using trap at the point where you define wds_kubeconfig?  I learned this from Jun:
```
wds_kubeconfig="$(mktemp "${wmw_name}.kubeconfig#XXXX")"
trap "rm -f $wds_kubeconfig" EXIT
```

scripts/inner/kubestellar
```diff
@@ -19,12 +19,20 @@
 #           (a) 1 kcp workspace: edge service provider workspace (espw)
 #           (b) 3 kubestellar controllers: kubestellar-where-resolver, mailbox-controller and placement-translator
 
-# Assumption: kcp server is running.
+# Pre-req: the space manager is running and there is a space provider
+# named "default" and that is the one to use.
+
+# Pre-req: `kubectl` with no flags will access the Space Management
+# API Space (SMAS) with high privilege.
```
effi-ofer, line 26: You're going to add the Space Management API Space to the terminology page?

scripts/inner/kubestellar
```diff
@@ -63,10 +73,10 @@ process_running() {
 }
 
 wait_for_process(){
-  status=$(process_running $1)
-  MAX_RETRIES=10
-  retries=0
-  status_code=0
+  local status=$(process_running $1)
+  local MAX_RETRIES=10
```
effi-ofer, line 77: General comment: we should standardize on variable names - upper vs lower case.

scripts/inner/kubestellar
```diff
@@ -154,58 +160,24 @@ while (( $# > 0 )); do
...
 function ensure_espw() {
-    in_cluster=
-    if $IN_CLUSTER; then
-        in_cluster="--in-cluster"
-    fi 
-    space_config="${KUBECONFIG_DIR}/${PROVIDER_NAMESPACE}-${espw_name}"
-    kubectl-kubestellar-create-space --space-name $espw_name --provider-name $PROVIDER_NAME --sm-core-config $SM_CONFIG 
-    kubectl-kubestellar-get-config-for-space --space-name $espw_name --sm-core-config $SM_CONFIG ${in_cluster} --sm-context $SM_CONTEXT --output $space_config
-    run_dex
+    local kcs_kubeconfig="$(mktemp kcs.kubeconfig#XXXX)"
```
effi-ofer, line 170: Is this kcs_kubeonfig deleted anywhere?  Can it be done with bash trap?

scripts/inner/kubestellar
```diff
@@ -154,58 +160,24 @@ while (( $# > 0 )); do
...
+    kubectl-kubestellar-space-ensure $espw_name $in_cluster --output-kubeconfig $kcs_kubeconfig
+    ensure_dex
     sleep 2
-    run_kb_backend
-    KUBECONFIG=$space_config kubectl apply -f "$bindir/../config/crds/"
+    ensure_kb_backend "$kcs_kubeconfig"
+    KUBECONFIG=$kcs_kubeconfig kubectl apply -f "$bindir/../config/crds/"
+    rm "$kcs_kubeconfig"
```
effi-ofer, line 176: Perhaps use a bash trap?

scripts/inner/kubestellar
```diff
@@ -227,23 +199,27 @@ function run_dex() {
...
     if [ $(process_running example-backend) == "running" ]; then
         echo "example-backend is already running"
         return
     fi
 
     echo "starting kube-bind example-backend in background"
 
-    kubectl-kubestellar-get-config-for-space --space-name "espw" --sm-core-config $SM_CONFIG ${in_cluster} --sm-context $SM_CONTEXT --output "/tmp/espw-conf"
+    local kcs_kubeconfig="/tmp/kcs.kubeconfig"
+    cp "$1" $kcs_kubeconfig
```
effi-ofer, line 214: Does kcs_kubeconfig ever get deleted?

scripts/inner/kubestellar
```diff
@@ -227,23 +199,27 @@ function run_dex() {
...
     
     ( 
-        KUBECONFIG="/tmp/espw-conf"
+        export KUBECONFIG="$kcs_kubeconfig"
```
effi-ofer, line 217: Is the scope of this KUBECONFIG only within the round brackets?  Or does this change the value of KUBECONFIG outside?

scripts/inner/kubestellar-kube-bind
```diff
@@ -129,6 +170,8 @@ metadata:
 data: {}
 EOF
 
+rm $KUBECONFIG
```
effi-ofer, line 173: use trap?

scripts/outer/kubectl-kubestellar-ensure-location
```diff
@@ -114,10 +101,7 @@ set -e
 
 echo "--- current directory is $PWD"
 
-imw_space_config="${PWD}/temp-space-config/spaceprovider-default-${imw}"
-kubectl-kubestellar-get-config-for-space --space-name ${imw} --sm-core-config $SM_CONFIG --sm-context $SM_CONTEXT --output $imw_space_config
-
-if ! kubectl "${kubectl_flags[@]}" --kubeconfig $imw_space_config get synctargets.edge.kubestellar.io "$objname" &> /dev/null; then
+if ! kubectl "${kubectl_flags[@]}" get synctargets.edge.kubestellar.io "$objname" &> /dev/null; then
```
effi-ofer, line 104: Do we need to run these kubectl commands with a particular kubeconfig?  If so how do you make sure it happens?  By assuming the --kubeconfig <config> is passed on the kubectl_flags?

scripts/outer/kubectl-kubestellar-prep_for_syncer
```diff
@@ -110,68 +113,86 @@ fi
...
 
-if ! KUBECONFIG=$SM_CONFIG kubectl get space -n spaceprovider-default "$mbwsname" &> /dev/null; then
-    echo "$0: did not find the mailbox workspace $mbwsname; is the mailbox controller running?" >&2
+if ! kubectl get space -n spaceprovider-default "$mbsname" &> /dev/null; then
+    echo "$0: did not find the mailbox workspace $mbsname; is the mailbox controller running?" >&2
     exit 5
 fi
 
 echo "now that workspace _exists_, but is it _ready_?"
+set -x
```
effi-ofer, line 164: Do we need to set -x here?

docs/content/Getting-Started/user-quickstart-test.md
MikeSpreitzer: Good point.

scripts/inner/kubectl-kubestellar-ensure-wmw
MikeSpreitzer: I deliberately did _not_ use `trap` in this case, in order to leave some evidence behind to aid in investigation in the failure case.

scripts/inner/kubestellar
MikeSpreitzer: We discussed this in Slack a while ago and agreed on this term.
Yes, it should go in the glossary.
I put an introduction to this term in the introductory material in commands.md.

scripts/inner/kubestellar
MikeSpreitzer: There is already a lot of mixed usage, I am not sure it is worth the trouble. Does the fact that the usage is mixed cause a problem?

scripts/inner/kubestellar
MikeSpreitzer: This, too, was a deliberate choice for the sake of debuggability.

scripts/inner/kubestellar
MikeSpreitzer: No, that is why it is a distinct file. Remember that the example-backend continues running after this script exits.

scripts/inner/kubestellar
MikeSpreitzer: Round brackets make a subshell. A subprocess does not affect the environment of its ancestors or siblings.

scripts/inner/kubestellar-kube-bind
MikeSpreitzer: Another deliberate choice for debuggability.

scripts/outer/kubectl-kubestellar-ensure-location
MikeSpreitzer: The caller gets to decide how to pass in the desired kubeconfig; as a kubectl plugin, all the usual ways are available (and as a kubectl plugin all the usual ways SHOULD be supported):
- via `--kubeconfig` on the command line;
- referenced in `$KUBECONFIG`;
- in `~/.kube/config`.

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-space-manager.md
```diff
@@ -5,10 +5,12 @@
 ```shell
 kind create cluster --name sm-mgt
 KUBECONFIG=~/.kube/config kubectl config rename-context kind-sm-mgt sm-mgt
-export SM_CONFIG=~/.kube/config
-export SM_CONTEXT=sm-mgt
+SM_CONFIG=~/.kube/config
```
effi-ofer, line 8: I am no bash expert so I am just asking - we don't need to export the SM_CONFIG in case the generated script calls another script?  Or are all the scripts being converted to accept SM_CONFIG as parameter?

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-space-manager.md
```diff
@@ -30,13 +32,13 @@ export PATH=$(pwd)/bin:$PATH
...
 ```shell
-KUBECONFIG=$SM_CONFIG kubectl --context sm-mgt apply -f config/crds/
+KUBECONFIG=$SM_CONFIG kubectl apply -f config/crds/
 cd ..
 ```
 Finally, start the space-manager controller.
 
 ```shell
-space-manager --kubeconfig $SM_CONFIG --context sm-mgt -v 2 &
+space-manager --kubeconfig $SM_CONFIG --context sm-mgt -v 4 &> /tmp/space-manager.log &
```
effi-ofer, line 41: With the assumption in line 11 about the current context, do we still need to pass --context to space-manager?  I think the space-manager has a default of context of sm-mgt.

docs/content/Getting-Started/user-quickstart-test.md
MikeSpreitzer: We may categorize them as follows:
- user (e.g., example1 or a quickstart script) chose where to put the file;
- script chose where to put the file.

There is the potential for conflicts if both categories go to the same directory, because each author does not know what the other is doing.

In the latter case my intent has been that the script deletes the file when done, except for the cases where there is an on-going need for the file (konnector, example-backend).

MikeSpreitzer: I think I will change the example1 and quickstarts to locate those kubeconfigs in a directory that is cleared at the _start_.

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-stage-3.md
```diff
@@ -15,26 +15,24 @@ use the following commands to launch the placement translator; it
...
+GUILDER_MB_KUBECONFIG="${PWD}/${GUILDER_SPACE}.kubeconfig"
+kubectl-kubestellar-space-get_kubeconfig $FLORIN_SPACE --kubeconfig $SM_CONFIG $in_cluster $FLORIN_MB_KUBECONFIG
+kubectl-kubestellar-space-get_kubeconfig $GUILDER_SPACE --kubeconfig $SM_CONFIG $in_cluster $GUILDER_MB_KUBECONFIG
 
 # wait until SyncerConfig, ReplicaSets and Deployments are ready
 mbxws=($FLORIN_SPACE $GUILDER_SPACE)
 for ii in "${mbxws[@]}"; do
   (
-    export KUBECONFIG="${PWD}/temp-space-config/$ii"
+    export KUBECONFIG="${PWD}/${ii}.kubeconfig"
```
effi-ofer, line 35: Do we need to export in this case?  Each kubectl statement in the body of the for loop would still get the correct KUBECONFIG even if it was local and there is no call to a sub script.

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-stage-3.md
```diff
@@ -53,7 +51,7 @@ for ii in "${mbxws[@]}"; do
   )
 done
 (
-  export KUBECONFIG=$GUILDER_MB_CONFIG
+  export KUBECONFIG=$GUILDER_MB_KUBECONFIG
```
effi-ofer, line 54: Same as line 35.  Do we need to export?

docs/content/Getting-Started/user-quickstart-test.md
effi-ofer: Excellent decision.  Thank you.

scripts/inner/kubestellar
effi-ofer: Just looks unprofessional.  But you are right and we have bigger fish to fry.

docs/content/Getting-Started/user-quickstart-test.md
MikeSpreitzer: Done in the push to 2781ec836

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-space-manager.md
MikeSpreitzer: This PR is changing the KubeStellar scripts to not read the `SM_CONFIG` or `SM_CONTEXT` environment variables. This is part of fixing #1438 : getting the kube go client library to read, when running in a pod, the pod's credentials.

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-space-manager.md
MikeSpreitzer: I first tried omitting the `--context` and the space manager crashed with a complaint about not being able to use the context `kind-mgt`, if I recall correctly. Please look into the default behavior of the space manager.

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-stage-3.md
MikeSpreitzer: You seem to be assuming that `KUBECONFIG` is automatically exported even if export is not explicitly requested here. You may have forgotten that prior to this PR there was another command that marked `KUBECONFIG` for export. It was very early in example1 and easy to forget about later. In this PR I am removing that, as it is a confusing long-range dependency. So in this subshell (the round brackets you see here) it is necessary to either `export KUBECONFIG=...` as a standalone command or specify `KUBECONFIG=...` on each `kubectl ...` command.

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-stage-3.md
MikeSpreitzer: Same answer.

scripts/outer/kubectl-kubestellar-prep_for_syncer
```diff
@@ -110,68 +113,85 @@ fi
...
     exit 1
 fi
 
-espw_space_config="${PWD}/temp-space-config/spaceprovider-default-${espw}"
-kubectl-kubestellar-get-config-for-space --space-name ${espw} --sm-core-config $SM_CONFIG --sm-context $SM_CONTEXT --output $espw_space_config
+espw_kubeconfig=$(mktemp espw.kubeconfig#XXXX)
+kubectl-kubestellar-space-get_kubeconfig "${sub_flags[@]}" "${kubectl_flags[@]}" ${espw} "$espw_kubeconfig"
 
-if ! cluster_ns_name="$(KUBECONFIG=$espw_space_config kubectl "${kubectl_flags[@]}" -n kubestellar get configmap kbmap-"$imw" -ojsonpath="{.metadata.labels['kubestellar\.io\/kube-bind-id']}")"; then
+if ! cluster_ns_name="$(KUBECONFIG=$espw_kubeconfig kubectl -n kubestellar get configmap kbmap-"$imw" -ojsonpath="{.metadata.labels['kubestellar\.io\/kube-bind-id']}")"; then
```
ezrasilvera, line 134: I see we are mixing KUBECONFIG=  and --kubeconfig.  I suggest using the --kubeconfig for direct kubectl commands

scripts/inner/kubestellar-kube-bind
```diff
@@ -61,36 +75,55 @@ then
...
-konnector_log_file=$kb_dir/kube-bind-konnector-$space_name.log
+konnector_dir="konnector-${space_name}"
+konnector_log_file="${konnector_dir}/log"
+finally=""
 
 if pgrep konnector | xargs ps e -p 2>/dev/null | egrep "KB_CONSUMER=$space_name(\s|$)" &> /dev/null; then
     echo "kube-bind konnector for space $space_name already running"
+    consumer_kubeconfig=$(mktemp "$space_name".kubeconfig#XXXX)
+    finally="rm -f '$consumer_kubeconfig'; "
+    trap "$finally" EXIT
```
ezrasilvera, line 88: Why two traps in a single script ? Can we do conditional cleanup and define a single trap at the beginning ?

scripts/inner/kubestellar-kube-bind
```diff
@@ -61,36 +75,55 @@ then
...
+    launch_konnector=false
 else
-    if ! kubectl get namespace kube-system &> /dev/null
-    then kubectl create namespace kube-system &> /dev/null
+    mkdir -p "$konnector_dir"
+    consumer_kubeconfig="${konnector_dir}/kubeconfig"
+    launch_konnector=true
+fi
+
+kubectl kubestellar space get-kubeconfig "${sub_flags[@]}" "${kubectl_flags[@]}" $space_name "$consumer_kubeconfig"
```
ezrasilvera, line 96: I see you replaced explicit flag based arguments with positional arguments.  I find positional arguments very confusing and hard to maintain

scripts/overlap/kubectl-kubestellar-space-get_kubeconfig
```diff
@@ -0,0 +1,83 @@
...
+	    if (( $# > 1 )); then 
+		 kubectl_flags[${#kubectl_flags[*]}]="$2"
+		 shift
+	    fi;;
+	(-*)
+	    echo "$0: flag syntax error" >&2
+	    exit 1;;
+	(*)
+	    if [ -z "$space_name" ]; then space_name="$1"
+	    elif [ -z "$output" ]; then output="$1"
```
ezrasilvera, line 45: wrote it also on one of the usages:
I see you replaced explicit flag based arguments with positional arguments. I find positional arguments very confusing and hard to maintain

scripts/inner/kubestellar-kube-bind
```diff
@@ -61,36 +75,55 @@ then
...
+ORIGINAL_KUBECONFIG="$KUBECONFIG"
+export KUBECONFIG="$consumer_kubeconfig"
```
ezrasilvera, line 98: I specifically changed this. I think such manipulation with a global KUBECONFIG are trouble.  (also some signals are not detected by trap like SIGKILL   etc ..)

scripts/inner/kubestellar-kube-bind
MikeSpreitzer: When there are a lot of them, yes. When there are only two, I think it is quite manageable.

scripts/outer/kubectl-kubestellar-prep_for_syncer
MikeSpreitzer: OTOH, consistently starting lines with `KUBECONFIG=<something> ` has a certain appeal.

What do you mean by _direct_ kubectl commands?  Do you mean ones that do not involve a kubectl plugin? If so, why should that matter to this question?

scripts/inner/kubestellar-kube-bind
MikeSpreitzer: There is already one conditional (the `if pgre konnector ...` just above), I thought it simpler to not have a more complicated command in the trap.

scripts/inner/kubestellar-kube-bind
MikeSpreitzer: I was following the lead of what was already there. The scope of this assignment is only about 50 lines I think that is OK.

docs/content/Getting-Started/user-quickstart-test.md
```diff
@@ -343,7 +351,7 @@ how to create, but not overwrite/update a synchronized resource
            -o jsonpath='{.data.external\.kubeconfig}' \
            -n kubestellar | base64 -d > ks-core.kubeconfig
```
MikeSpreitzer, line 352: Oops, forgot to put this in `MY_KUBECONFIGS`. I think that can be fixed in follow-on.

scripts/inner/kubestellar-kube-bind
```diff
@@ -138,10 +181,22 @@ else
...
 fi
 
-KUBECONFIG=$space_conf_file
+if fgrep -x "${resource_to_bind}" "$request_file"; then
+    echo "The request file has junk in it; see?" >&2
+    cat "$request_file" >&2
+    mv "$request_file" "$request_file".junk
+    fgrep -x -v "${resource_to_bind}" "$request_file".junk > "$request_file"
+    echo Removed >&2
+fi
```
MikeSpreitzer, line 190: I added this because I got the following failure from `kubestellar-kube-bind` when testing on my laptop.

```
...
++ kubectl bind apiservice --remote-kubeconfig-namespace kube-bind --remote-kubeconfig-name kubeconfig-ms6gs --skip-konnector -f konnector-imw1/locations-request.yaml
+ sub_result='DISCLAIMER: This is a prototype. It will change in incompatible ways at any time.

Error: failed to unmarshal manifest: error converting YAML to JSON: yaml: line 2: mapping values are not allowed in this context'
+ echo 'Error occurred during binding locations for imw1'
Error occurred during binding locations for imw1
+ echo 'Result of kubectl bind apiservice: DISCLAIMER: This is a prototype. It will change in incompatible ways at any time.

Error: failed to unmarshal manifest: error converting YAML to JSON: yaml: line 2: mapping values are not allowed in this context'
Result of kubectl bind apiservice: DISCLAIMER: This is a prototype. It will change in incompatible ways at any time.

Error: failed to unmarshal manifest: error converting YAML to JSON: yaml: line 2: mapping values are not allowed in this context
+ echo 'Request file:'
Request file:
+ cat konnector-imw1/locations-request.yaml
locations
apiVersion: kube-bind.io/v1alpha1
kind: APIServiceExportRequest
metadata:
  name: locations.edge.kubestellar.io
spec:
  resources:
  - group: edge.kubestellar.io
    resource: locations
status: {}

+ echo
...
```

docs/content/Getting-Started/user-quickstart-test.md
ezrasilvera: Sounds good

scripts/inner/kubestellar
ezrasilvera: Let's push it for a (much) later PR :-)

docs/content/Coding Milestones/PoC2023q1/example1-subs/example1-space-manager.md
ezrasilvera: yes the SM doesn't use the "current" we need to change the code

ezrasilvera:
Mike - thanks for the hard work :-)  Let's merge it
/lgtm

docs/content/Getting-Started/user-quickstart-test.md
MikeSpreitzer: Addressed in #1463

Issue #1452, opened 2023-12-11
bug: concurrency bug wrt kube-bind - space name relationship


MikeSpreitzer:
### Describe the bug

The `kubestellar-kube-bind` script writes the ConfigMap that holds the relationship _after_ doing the `kubectl bind` operation, so projected objects can arrive in the KCS before the ConfigMap.  The mailbox controller has an `IndexFunc` that returns an error if it can not follow the relationship. Did you know that returning an error from an `IndexFunc` causes a panic? Following is an excerpt from a mailbox controller log.

```
...
I1211 12:33:27.365162   93943 reflector.go:271] Starting reflector *v2alpha1.SyncTarget (0s) from k8s.io/client-go@v0.0.0-20230210192259-aaa28aa88b2d/tools/cache/reflector.go:215
I1211 12:33:27.365176   93943 reflector.go:307] Listing and watching *v2alpha1.SyncTarget from k8s.io/client-go@v0.0.0-20230210192259-aaa28aa88b2d/tools/cache/reflector.go:215
I1211 12:33:27.365182   93943 reflector.go:271] Starting reflector *v1alpha1.Space (0s) from k8s.io/client-go@v0.0.0-20230210192259-aaa28aa88b2d/tools/cache/reflector.go:215
I1211 12:33:27.365190   93943 reflector.go:307] Listing and watching *v1alpha1.Space from k8s.io/client-go@v0.0.0-20230210192259-aaa28aa88b2d/tools/cache/reflector.go:215
E1211 12:33:27.366187   93943 controller.go:382] "failed to get consumer space ID from a provider's copy" err="failed to retrive spaceID" syncTarget="kube-bind-4hzzv-florin"
E1211 12:33:27.366324   93943 runtime.go:79] Observed a panic: &errors.errorString{s:"unable to calculate an index entry for key \"kube-bind-4hzzv-florin\" on index \"mbsName\": failed to retrive spaceID"} (unable to calculate an index entry for key "kube-bind-4hzzv-florin" on index "mbsName": failed to retrive spaceID)
goroutine 104 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x106414ce0?, 0x140003e1c00})
	/Users/mspreitz/go/pkg/mod/github.com/kcp-dev/kubernetes/staging/src/k8s.io/apimachinery@v0.0.0-20230210192259-aaa28aa88b2d/pkg/util/runtime/runtime.go:75 +0x7c
[...]

PR #1453, opened 2023-12-11
✨ Create configmap before binding

waltforme:
## Summary
In this PR, the configmaps are created before doing `kubectl bind apiservice`, so that the problem documented in #1452 can be avoided.

### DIFF
--- a/scripts/inner/kubestellar-kube-bind
+++ b/scripts/inner/kubestellar-kube-bind
@@ -19,7 +19,7 @@
 space_name=""
 resource_to_bind=""
 cm_namespace=kubestellar # the namespace where the configmaps live in on the provider side
-log_dir=$(pwd)/kubestellar-kube-bind-logs
+kb_dir=$(pwd)/kubestellar-kube-bind-files
 
 in_cluster=
 if $IN_CLUSTER; then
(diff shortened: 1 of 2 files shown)


waltforme:
Force-push to https://github.com/kubestellar/kubestellar/commit/19829d0639b38aea5c9a92e25a4a1f4dc9a6b42e was a rebase onto main.

scripts/inner/kubestellar-kube-bind
```diff
@@ -89,8 +89,11 @@ else
     fi
 fi
 
+request_file=$kb_dir/$resource_to_bind-request.yaml
+trap 'rm $request_file' EXIT
+
 echo "binding $resource_to_bind for $space_name"
-if ! result="$(kubectl bind http://127.0.0.1:8080/export --skip-konnector --unattended --resource $resource_to_bind 2>&1)"; then
+if ! result="$(kubectl bind http://127.0.0.1:8080/export --dry-run --skip-konnector --unattended --resource $resource_to_bind 2>&1 1>$request_file)"; then
```
MikeSpreitzer, line 96:
```
bash-3.2$ result="$(date 2>&1 1>datefile)"
bash-3.2$ echo "result=<$result>"
result=<>
```

MikeSpreitzer: So this works because `kubectl bind` sends the sought output to stderr?

waltforme: Yes!

PR #1457, opened 2023-12-12
:sparkles: Add cli flag for external access to space provider

andreyod:
## Summary
This PR introduces new boolean flag for KubeStellar controllers. This flag indicates whether a controller needs external or in-cluster access to the spaces. The flag value is `true` for external, or default `false` for in cluster access.

## Related issue(s)

Fixes #


pkg/placement/what-resolver_test.go
```diff
@@ -152,7 +152,7 @@ func TestWhatResolver(t *testing.T) {
 	kbSpaceRelation := kbuser.NewKubeBindSpaceRelation(ctx, fakeUpKubeClient)
 	spaceProviderNs := "spaceprovider-fake"
 	// TODO fake
-	spaceclient, _ := msclient.NewMultiSpace(ctx, nil)
+	spaceclient, _ := msclient.NewMultiSpace(ctx, nil, true)
```
ezrasilvera, line 155: Why true?

PR #1465, opened 2023-12-13
🌱 updated kube bind `AnalyzeObject` func not to return the namespace of the original object.

nirrozenbaum:
this return value wasn't used anywhere in the code. in case the need arises in the future, the original function was commented out.

## Summary

the namespace of the original object wasn't used anywhere in the code. 
in case the need arises in the future, the original function was commented out.

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/Coding Milestones/PoC2023q1/outline.md
+++ b/docs/content/Coding Milestones/PoC2023q1/outline.md
@@ -176,8 +176,8 @@ mapping can be inverted with the following interface.
 ```go
 
 // AnalyzeClusterScopedObject examines a cluster-scoped object in a kube-bind service provider cluster
-// and returns the object's ID as known in the service consumer cluster and the kube-bind ID for the
-// consumer, or an error if the object does not appear to be a provider's copy of a consumer's object.
+// and returns the object's name as known in the service consumer cluster and the kube-bind space id for
+// the consumer, or an error if the object does not appear to be a provider's copy of a consumer's object.
 func AnalyzeClusterScopedObject(obj metav1.Object) (name, kbSpaceID string, err error) {..}
 
 // ComposeClusterScopedName translates the name of a cluster-scoped object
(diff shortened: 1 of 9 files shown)


pkg/kbuser/object_ids.go
```diff
@@ -24,32 +24,55 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// If in the future a need arises to analyze both cluster-scoped and namespace-scoped objects, switch to using
```
MikeSpreitzer, line 27: I would put this note _after_ the live code rather than _before_, expecting readers to be more interested in the live code.

nirrozenbaum: done

docs/content/Coding Milestones/PoC2023q1/outline.md
```diff
@@ -175,11 +175,10 @@ mapping can be inverted with the following interface.
 
 ```go
 
-// AnalyzeObjectID examines an object in a kube-bind service provider
-// cluster and returns the object's ID as known in the service consumer
-// cluster and the kube-bind ID for the consumer, or an error if
-// the object does not appear to be a provider's copy of a consumer's object.
-func AnalyzeObjectID(obj metav1.Object) (namespace, name, kbSpaceID string, err error) {..}
+// AnalyzeClusterScopedObject examines a cluster-scoped object in a kube-bind service provider cluster
+// and returns the object's ID as known in the service consumer cluster and the kube-bind ID for the
```
MikeSpreitzer, line 179: Now that we do not need a term for Namespace+Name, we can simply say "object name" (which is a well-defined technical term) instead of "object ID" (which is less so).

pkg/kbuser/object_ids.go
MikeSpreitzer: Good.

docs/content/Coding Milestones/PoC2023q1/outline.md
nirrozenbaum: done

MikeSpreitzer:
The changes here look good to me. I do not know why this PR is failing CI.

PR #1466, opened 2023-12-14
🐛 Don't ignore available discovery results in err case

MikeSpreitzer:
Also simplify a for loop.

## Summary
This PR fixes a bug in the apiwatch package, where it was incorrectly assuming that when a discovery call returns an error the associated data structure(s) are empty. That meant that one problematic API group would make them all disappear.

This PR also simplifies a loop statement to stop the nagging from linters.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/apiwatch/resources.go
+++ b/pkg/apiwatch/resources.go
@@ -341,7 +341,7 @@ func (am arMap) toList(logger klog.Logger, prefix []string, consume func(ksmetav
 func (rlw *resourcesListWatcher) listWithSubresources(logger klog.Logger, resourceVersionS string) ([]ksmetav1a1.APIResource, error) {
 	groupList, resourceList, err := rlw.cache.ServerGroupsAndResources()
 	if err != nil {
-		return nil, err
+		rlw.logger.V(3).Info("Did not get all api groups and resources", "err", err.Error())
 	}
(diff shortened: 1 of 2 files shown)


pkg/apiwatch/resources.go
```diff
@@ -341,7 +341,7 @@ func (am arMap) toList(logger klog.Logger, prefix []string, consume func(ksmetav
 func (rlw *resourcesListWatcher) listWithSubresources(logger klog.Logger, resourceVersionS string) ([]ksmetav1a1.APIResource, error) {
 	groupList, resourceList, err := rlw.cache.ServerGroupsAndResources()
 	if err != nil {
-		return nil, err
+		rlw.logger.V(3).Info("Did not get all api groups and resources", "err", err.Error())
```
ezrasilvera, line 344: Is that enough ?  Are you sure there are no (other) errors that may result in a nil groupList or resourceList that currently are not checked afterwards as you originally returned here ?

MikeSpreitzer: I do not understand the question. This change is (1) removing the incorrect assumption that non-nil error implies empty list and (2) moving the error handling from the caller into here. (1) is surely harmless. (2) is harmless as far as I know about callers.

PR #1473, opened 2023-12-17
Update example.md

msteinder:
Script below includes ../... I think it should be consistent

## Related issue(s)

Fixes #


msteinder:
Nope, the script has a mistake. As the following commands require you to be in kubestellar folder

PR #1482, opened 2023-12-20
:bug: Update remove-all.md

andreyod:
## Summary
1. Kill the space-manager and example-backend processes.
2. Delete the sm-mgt Kind cluster.
3. Remove files and directories created by docs-executable in Kubestellar root directory.

## Related issue(s)

Fixes #


ezrasilvera:
@waltforme Can you please try to add a safe way to kill dex and konnector ?  Maybe we should add env variable and kill by matching to that as well ?
/lgtm
/approve

waltforme:
Sure, will track this once back to office.

PR #1488, opened 2023-12-20
🌱 switched from controller runtime client to dynamic client

nirrozenbaum:
below are the list of changes in this PR:
- replaced controller runtime client with dynamic client
- renamed receiver name in all funcs to c instead of ctrl
- added GVR of wrapped object to the transport interface.  

the last point is a temp code that will be changed soon such that the controller will calculate GVR of the wrapped object internally. when adding this piece of code, we will remove the GVR requirement from transport interface.

## Related issue(s)

Fixes #


pkg/transport/transport.go
```diff
@@ -18,11 +18,13 @@ package transport
 
 import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 type Transport interface {
 	InformerSynced() bool
 	WrapObjects([]*unstructured.Unstructured) *unstructured.Unstructured // wrap multiple objects into a single wrapped object.
+	GetWrappedObjectGVR() schema.GroupVersionResource
```
vMaroon, line 27: Perhaps add a TODO on what we discussed (the work that results in this interface eventually containing only wrapObjects)

vMaroon (approved):

pkg/transport/transport.go
nirrozenbaum: I hope to complete this part today or the day after. otherwise I'll add a TODO

PR #1489, opened 2023-12-20
🌱 removed the use of ObjectMeta directly in space-framework (part2)

nirrozenbaum:
best practices is to use the Get/Set functions instead of ObjectMeta directly.
in addition, in order to check whether an object has marked for deletion, one should use the built-in IsZero() function

## Related issue(s)

Fixes #


clubanderson:
@ezrasilvera to review

ezrasilvera (approved):

PR #1490, opened 2023-12-21
🌱 calculate wrapped object GVR from the wrapped object.

nirrozenbaum:
removed the requirement to get wrapped object GVR from transport interface

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/transport/generic_transport_controller.go
+++ b/pkg/transport/generic_transport_controller.go
@@ -30,8 +30,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
+	cacheddiscovery "k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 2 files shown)


pkg/transport/generic_transport_controller.go
```diff
@@ -54,7 +57,16 @@ const (
 // NewTransportController returns a new transport controller
 func NewTransportController(ctx context.Context, edgePlacementDecisionInformer edgev2alpha1informers.EdgePlacementDecisionInformer, transport Transport,
 	kubeBindSpaceRelation kbuser.KubeBindSpaceRelation, spaceManagementClient msclient.KubestellarSpaceInterface, spaceProviderNs string,
-	transportSpaceClient dynamic.Interface) *genericTransportController {
+	transportSpaceConfig *rest.Config) (*genericTransportController, error) {
+	transportSpaceDynamicClient, err := dynamic.NewForConfig(transportSpaceConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create dynamic k8s clientset for transport space - %w", err)
+	}
+	wrappedObjectGVK := transport.WrapObjects(make([]*unstructured.Unstructured, 0)).GroupVersionKind() // empty wrapped object to get GVK from it.
```
shmuelk, line 65: This depends on the fact that transport.WrapObjects returns a wrapped object with an empty list of objects inside it. You need to at least document this.

vMaroon (approved):

PR #1492, opened 2023-12-22
🌱 Transport wrappedobj informer

nirrozenbaum:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/transport/generic_transport_controller.go
+++ b/pkg/transport/generic_transport_controller.go
@@ -154,7 +154,7 @@ func (c *genericTransportController) enqueueEdgePlacementDecision(obj interface{
 	var key string
 	var err error
 	if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
-		c.logger.Error(err, "failed to extract key from EdgePlacementDecision object")
+		c.logger.Error(err, "failed to enqueue EdgePlacementDecision object")
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/transport/generic_transport_controller.go
```diff
@@ -140,18 +147,44 @@ type genericTransportController struct {
...
-// string which is then put onto the work queue. This method should *not* be passed resources of any type other than EdgePlacementDecision.
+// enqueueEdgePlacementDecision takes an EdgePlacementDecision resource and
+// converts it into a namespace/name string which is put onto the workqueue.
+// This func *shouldn't* handle any resource other than EdgePlacementDecision.
 func (c *genericTransportController) enqueueEdgePlacementDecision(obj interface{}) {
 	var key string
 	var err error
 	if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
-		c.logger.Error(err, "failed to enqueue EdgePlacementDecision object")
+		c.logger.Error(err, "failed to extract key from EdgePlacementDecision object")
```
vMaroon, line 157: I think this change does make the error more accurate, but for debugging clarity later (or log clarity overall) - it should probably also still include that the enqueuing has failed, since no other error level around calling this function reports it.

vMaroon (approved):
looks good

pkg/transport/generic_transport_controller.go
nirrozenbaum: reverted to previous log message.
merging the PR

Issue #1499, opened 2024-01-02
The build of the core container image can be optimized


waltforme:
Right now it takes very long to finish `make kubestellar-image`. The time varies from machine to machine. My experiences was roughly 50 minutes, while colleagues report 'over one hour'.

There are potential optimizations to streamline the build process. We can use prebuilt images for infrequently changed binaries and/or cross-compilation to speed it up.

clubanderson:
#1487 is related

Issue #1487, opened 2023-12-20
Recover the supports to `linux/ppc64le` after including the automated kubectl-bind into core image


waltforme:
Our recently automated `kubectl-bind` uses `surf` to simulate the user's browser. However, `surf` lacks supports to ppc64le at build time.
```
#0 1029.4 /root/go/pkg/mod/github.com/headzoo/surf@v1.0.1/agent/agent_linux.go:19:23: cannot use buf.Sysname (variable of type [65]uint8) as [65]int8 value in argument to charsToString
#0 1029.4 /root/go/pkg/mod/github.com/headzoo/surf@v1.0.1/agent/agent_linux.go:29:23: cannot use buf.Release (variable of type [65]uint8) as [65]int8 value in argument to charsToString
```
Temporarily we can remove `linux/ppc64le` from Makefile's `CORE_PlATFORMS`, then build a multi-arch image which supports `linux/amd64` and `linux/arm64`.

In the long run we should resume the supports to `linux/ppc64le`.

clubanderson:
work in progress.  @waltforme has opened a feature request in the upstream web sim lib to support ppc.  going to update the ks build to re-use existing lib instead of building binaries to support this feature.

waltforme:
Done by #1505. Closing.

PR #1505, opened 2024-01-05
✨ Use prebuilt images for kube-bind related binaries

waltforme:
## Summary
Previously, kube-bind binaries are built from three different branches from waltforme/kube-bind, namely `autobind`, `syncmore` and `main`. Now the three branches are consolidated into one branch `kubestellar`.

With that, we first simplify the builds in example1.

Next, on the `kubestellar` branch, images were built for the kube-bind binaries. In `core.Dockerfile`, we then use the prebuilt kube-bind images to accelerate the build of kubestellar core image. Also, we use a prebuilt dex image for the same reason.

Note that the prebuilt kube-bind images are multi-platform images, including ppc64le, so #1487 is resolved.

A side effect of this PR is the reduced build time for `make kubestellar-image`, which is now roughly half an hour on my setup.
```
[+] Building 2060.3s (139/139) FINISHED                                      docker-container:kubestellar
```

## Related issue(s)
#1499

### DIFF
--- a/core.Dockerfile
+++ b/core.Dockerfile
@@ -40,12 +40,7 @@ RUN mkdir -p .kcp && \
 
 RUN git clone https://github.com/waltforme/kube-bind.git && \
     pushd kube-bind && \
-    mkdir bin && \
-    IGNORE_GO_VERSION=1 go build -o ./bin/example-backend ./cmd/example-backend/main.go && \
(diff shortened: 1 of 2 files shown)


clubanderson (commented):
/lgtm
/approve

PR #1509, opened 2024-01-09
✨ Use cross-compilation to speed up core image build

waltforme:
## Summary
This PR uses cross-compilation instead of emulation to speed up `make kubestellar-image`.

On my MacBook with an M1 Max chip it now takes 200+ seconds to build the image:
```
[+] Building 214.7s (145/145) FINISHED                                       docker-container:kubestellar
```

The time above is now mostly consumed by `make innerbuild` --- as it should be.

The time above doesn't include the extra time to push to quay.io. Pushing time varies from machine to machine. On my MacBook at home it takes roughly 200 secnods; on my dev Ubuntu on AWS it takes roughly 100 seconds.

## Related issue(s)
#1499

### DIFF
--- a/core.Dockerfile
+++ b/core.Dockerfile
@@ -1,7 +1,7 @@
 ###############################################################################
 # Builder image
 ###############################################################################
-FROM redhat/ubi9 AS builder
+FROM --platform=$BUILDPLATFORM redhat/ubi9 AS builder
 
 ARG TARGETOS
 ARG TARGETARCH
@@ -57,7 +57,7 @@ ADD test/            test/
 ADD .git/            .git/
 ADD .gitattributes Makefile Makefile.venv go.mod go.sum .
 
-RUN make innerbuild GIT_DIRTY=$GIT_DIRTY IGNORE_GO_VERSION=yesplease
+RUN make innerbuild GOOS=$TARGETOS GOARCH=$TARGETARCH GIT_DIRTY=$GIT_DIRTY IGNORE_GO_VERSION=yesplease
 
 FROM ghcr.io/waltforme/kube-bind/example-backend:latest AS example-backend-binary
 


core.Dockerfile
```diff
@@ -1,7 +1,7 @@
 ###############################################################################
 # Builder image
 ###############################################################################
-FROM redhat/ubi9 AS builder
+FROM --platform=$BUILDPLATFORM redhat/ubi9 AS builder
```
ezrasilvera, line 4: Who should set those variables?  Explicitly in our Makefile ?

waltforme: The docker doc says they are set when we use BuildKit:
https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

So when we use `buildx`, values about the build platform are automatically available, values about the target platform (Yes, you are right) are passed in by the `--platform` flag of `buildx build` which is in our Makefile.

MikeSpreitzer:
Has anybody tested whether the cross-compilations actually produce images that work on a different platform than the build platform?  I find a suspicious line in core.Dockerfile:

```
COPY --from=builder /root/go/bin                                         /usr/local/bin/
```

MikeSpreitzer:
I tested by doing a `make kubestellar-image` in a Linux VM on my Apple Silicon M2 Mac, then using the image from a Linux VM in the IBM Cloud.  Got a failure on yq.

```
mspreitz@mjs-dev7a:~$ docker create --name foo quay.io/kubestellar/kubestellar:git-bb1c09328-clean
Unable to find image 'quay.io/kubestellar/kubestellar:git-bb1c09328-clean' locally
git-bb1c09328-clean: Pulling from kubestellar/kubestellar
...
Digest: sha256:9fd4c9a740a0cecda6fef3083ebc00b2e5655de2534b9b399980887eb3a7b3cf
Status: Downloaded newer image for quay.io/kubestellar/kubestellar:git-bb1c09328-clean
1d958405dd008ad9bccdf9b8e8845e51b04cb9acccd5b50025dfeab35c80d058
mspreitz@mjs-dev7a:~$ 
mspreitz@mjs-dev7a:~$ docker cp foo:/usr/local/bin ulb
                               Successfully copied 57.3MB to /home/mspreitz/ulb
mspreitz@mjs-dev7a:~$ 
mspreitz@mjs-dev7a:~$ find ulb
ulb
ulb/kubectl
ulb/yq
mspreitz@mjs-dev7a:~$ 
mspreitz@mjs-dev7a:~$ file ulb/yq
ulb/yq: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, BuildID[sha1]=18f653ba6b10f7ea0a6cc9fc5816475ee52df46f, for GNU/Linux 3.7.0, with debug_info, not stripped
mspreitz@mjs-dev7a:~$ 
mspreitz@mjs-dev7a:~$ ulb/yq -h
-bash: ulb/yq: cannot execute binary file: Exec format error
```

PR #1511, opened 2024-01-10
✨  add contributor to README

grahamwhiteuk:
## Summary
Add [grahamwhiteuk](https://github.com/grahamwhiteuk) to contributors in README as suggested by @clubanderson (note: I've not opened a related issue in the tracker for this PR)

It also looks like my vscode settings of removing whitespace at the end of lines has adjusted a few additional things here but I think that's OK.

Note also: I've linked to my list of PRs in the README which is hinted at by the `title` element within the anchor reference that says "Contributed PRs".  This is different to all the other contributors in the README that have their assigned issues linked.  I have no assigned issues since I'm not currently a member of the project.  I wonder if the README would be better updated such that everyone else is shown with their contributions rather than their assigned issues???


clubanderson (commented):
/lgtm
/approve

PR #1525, opened 2024-01-11
⚠️ Partially converge old and new placement Kinds

MikeSpreitzer:
## Summary
This PR does some convergence between the old and new placement Kinds.

This PR _only_ includes the IDL change. How do I update the generated files?

## Related issue(s)

Fixes #

PR #1528, opened 2024-01-12
🌱 Remove some string encoding and parsing

MikeSpreitzer:
## Summary
This PR simplifies the controllers by removing some string encoding and decoding.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/placement/controller.go
+++ b/pkg/placement/controller.go
@@ -391,7 +391,7 @@ func (c *Controller) reconcile(ctx context.Context, key util.Key) error {
 		if err != nil {
 			// The resource no longer exist, which means it has been deleted.
 			if errors.IsNotFound(err) {
-				utilruntime.HandleError(fmt.Errorf("resource '%s' for lister '%s' in work queue no longer exists", key.NamespaceNameKey, key.GvkKey))
+				utilruntime.HandleError(fmt.Errorf("object %#v for lister '%s' in work queue no longer exists", key.NamespacedName, key.GvkKey))
 				return nil
 			}
 			return err
@@ -454,13 +454,7 @@ func (c *Controller) getObjectFromKey(key util.Key) (runtime.Object, error) {
 	}
 	lister := *pLister
 
-	namespace, name, err := cache.SplitMetaNamespaceKey(key.NamespaceNameKey)
-	if err != nil {
-		utilruntime.HandleError(fmt.Errorf("invalid resource key: %s %s", key.NamespaceNameKey, err))
-		return nil, nil
-	}
-
-	return getObject(lister, namespace, name)
+	return getObject(lister, key.NamespacedName.Namespace, key.NamespacedName.Name)
 }
 
 func getObject(lister cache.GenericLister, namespace, name string) (runtime.Object, error) {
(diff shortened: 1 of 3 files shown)


MikeSpreitzer:
This PR modifies both controller processes, so both will need to be tested.

pdettori:
> This PR modifies both controller processes. @pdettori how do I test the status controller?

@MikeSpreitzer - please follow step 3.a to install status add-on in https://github.com/kubestellar/kubestellar/tree/ks-0.20/docs/content/v0.20#setup, then use this scenario to test: https://github.com/kubestellar/kubestellar/tree/ks-0.20/docs/content/v0.20#scenario-4---singleton-status

MikeSpreitzer:
@pdettori : my question is how to test _this PR's_ version of the status controller.

pdettori:
> @pdettori : my question is how to test _this PR's_ version of the status controller.

@MikeSpreitzer you can follow the same steps to start the controller manager with `make run` you already followed for https://github.com/kubestellar/kubestellar/pull/1525 . The status controller is started by the same controller manager, and only starts if it detects the presence of the WorkStatus resource in the `imbs` space. Installing the status add-on on the imbs1 will install the CRD for WorkStatus and will create WorkStatus instances in mailbox namespaces for objects deployed by the KS controller, which would allow you test the status controller.

pdettori:
@MikeSpreitzer as you pointed out, we need to document the steps to test in general. Based on the issues you encountered (conflicts etc) the steps I provided in the other PR could be also be improved, for example, instead of creating a wds with the option  `create wds1 -p kslight` you may create the wds1 without starting the controller with -p kslight and then label the wds with `kflex.kubestellar.io/cptype: wds` to avoid the conflicts at startup. (this is likely a temporary issue until we start building the new images for the refactored code)

pkg/placement/controller.go
```diff
@@ -391,7 +391,7 @@ func (c *Controller) reconcile(ctx context.Context, key util.Key) error {
 		if err != nil {
 			// The resource no longer exist, which means it has been deleted.
 			if errors.IsNotFound(err) {
-				utilruntime.HandleError(fmt.Errorf("resource '%s' for lister '%s' in work queue no longer exists", key.NamespaceNameKey, key.GvkKey))
+				utilruntime.HandleError(fmt.Errorf("object %#v for lister '%s' in work queue no longer exists", key.NamespacedName, key.GvkKey))
```
pdettori, line 394: Good point, this was meant to be an object indeed.

pkg/util/keys.go
```diff
@@ -31,9 +31,9 @@ import (
 // since they are no longer in the cache, the key stores a shallow copy of the
 // deleted object.
 type Key struct {
-	GvkKey           string
-	NamespaceNameKey string
-	DeletedObject    *runtime.Object
+	GvkKey         string
+	NamespacedName cache.ObjectName
```
pdettori, line 35: Good point, this removes the need for parsing strings as was doing before, which was based on https://github.com/kubernetes/sample-controller/blob/master/controller.go

pdettori (approved):
/lgtm

pkg/util/keys.go
MikeSpreitzer: Thanks for the explanation. I hope to open a PR on that sample.

MikeSpreitzer:
The force-push to 2c3d7e67d is a rebase onto the latest edition of #1573

MikeSpreitzer:
The force-push to bd6d379c5 is the promised rebase after merge of #1573

pdettori (approved):

PR #1552, opened 2024-01-14
🌱 Clarify helm tweak

MikeSpreitzer:
## Summary
This PR updates the commentary at the start of scripts/add-helm-code.sh to clarify what it does.

## Related issue(s)

Fixes #


MikeSpreitzer:
Also, a few things that struck me as odd: why does this script take an optional subcommand when there is only one possibility? And why use the word "add" in the script name if it also appears as a subcommand? And what this does is not, to my mind, an "add" --- it is more of a "specialize".

PR #1572, opened 2024-01-17
🌱 Brushup e2e 1

MikeSpreitzer:
## Summary
This PR makes a few minor improvements to the one e2e test.

- Move "cleanup" from the end to the start, so that users can examine and build on the state established by the test. Modify the cleanup commands so that they nominally succeed regardless of the starting state.
- Make the scripts echo their commands, so that users can understand what is going on.
- Remove the self-referential and unnecessary instructions for how to use `git`.
- Rename `DEFAULT_KIND_CLUSTER` to `KIND_HOSTING_CLUSTER` to better describe its meaning/use.
- Rename and simplify the make target `ko-build` to `ko-build-local`, because it is only used to build an image used locally in testing and ko with podman does not really succeed to build a multi-platform manifest.
- Update the make target `ko-build-local` to support users who have rootless podman pretending to be docker.
- Remove the hack of the Helm chart for installing the WDS operator. The hack was setting the ImagePullPolicy to `Never` but the default `IfNotFound` works just fine, the image gets loaded into the kind cluster before that Helm chart is instantiated there.
- Remove the use of `kubectl config use-context`, making the choice of context explicit in every kubectl and helm command.
- Slow down the wait loops. Checking every second was annoyingly chatty.

## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ DEFAULT_NAMESPACE=default
 # Default WDS name to use for make deploy (mainly for local testing) 
 DEFAULT_WDS_NAME=wds1
 # default kind hosting cluster name
-DEFAULT_KIND_CLUSTER ?= kubeflex
+KIND_HOSTING_CLUSTER ?= kubeflex
 
 # We need bash for some conditional logic below.
 SHELL := /usr/bin/env bash -e
@@ -151,9 +151,9 @@ test: manifests generate fmt vet envtest ## Run tests.
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./cmd/kubestellar-operator/main.go $(ARGS)
 
-.PHONY: ko-build
-ko-build: test ## Build docker image with ko
-	KO_DOCKER_REPO=ko.local ko build --push=false -B ./cmd/${CMD_NAME} -t ${IMAGE_TAG} --platform linux/amd64,linux/arm64
+.PHONY: ko-build-local
+ko-build-local: test ## Build local container image with ko
+	KO_DOCKER_REPO=ko.local ko build -B ./cmd/${CMD_NAME} -t ${IMAGE_TAG}
 	docker tag ko.local/${CMD_NAME}:${IMAGE_TAG} ${IMG}
 
 .PHONY: docker-push
@@ -161,13 +161,13 @@ docker-push: ## Push docker image
 	docker push ${IMG}
 
 .PHONY: ko-build-push
-ko-build-push: test ## Build and push docker image with ko
+ko-build-push: test ## Build and push container image with ko
 	KO_DOCKER_REPO=${DOCKER_REGISTRY} ko build -B ./cmd/${CMD_NAME} -t ${IMAGE_TAG} --platform linux/amd64,linux/arm64
 
 # this is used for local testing 
 .PHONY: kind-load-image
 kind-load-image: 
-	kind load --name ${DEFAULT_KIND_CLUSTER} docker-image ${IMG}
+	kind load --name ${KIND_HOSTING_CLUSTER} docker-image ${IMG}
 
 .PHONY: chart
 chart: manifests kustomize
@@ -209,7 +209,6 @@ undeploy: ## Undeploy manager from the K8s cluster specified in ~/.kube/config. 
 .PHONY: install-local-chart
 install-local-chart: kind-load-image
 	helm upgrade --install kubestellar -n ${DEFAULT_WDS_NAME}-system ./core-helm-chart  --set ControlPlaneName=${DEFAULT_WDS_NAME}
-	kubectl -n ${DEFAULT_WDS_NAME}-system patch deployment kubestellar-controller-manager --type='json' -p='[{"op": "replace", "path": "/spec/template/spec/containers/1/imagePullPolicy", "value":"Never"}]'
 
 ##@ Build Dependencies
 
--- a/Makefile
+++ b/Makefile
@@ -153,7 +153,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 
 .PHONY: ko-build-local
 ko-build-local: test ## Build local container image with ko
-	KO_DOCKER_REPO=ko.local ko build -B ./cmd/${CMD_NAME} -t ${IMAGE_TAG}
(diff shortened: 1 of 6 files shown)


ezrasilvera:
@MikeSpreitzer The problem is that kflex is using blocking commands and the github workflow sample that and bombard the output with a lot of messages which make the output unusable. There is no kflex output on that command.
We need to differentiate between local usage (where we want to see the output) and cicd test where we want pus to /dev/null.
Maybe add a redirection flag that can be set by the run-test which is not used in cicd

Makefile
```diff
@@ -204,12 +204,12 @@ deploy: manifests kustomize ## Deploy manager to the K8s cluster specified in ~/
...
-# installs the chart from ./core-helm-chart for local dev/testing on default WDS using image loaded in kind
-# the deployment is patched to load the image pre-loaded in kind
+# installs the chart from ./core-helm-chart for local dev/testing a WDS using image loaded in kind.
+# The Helm chart should be instantiated into the KubeFlex hosting cluster.
+# If $(KUBE_CONTEXT) is set then that indicates where to install the chart; otherwise it goes to the current kubeconfig context.
 .PHONY: install-local-chart
 install-local-chart: kind-load-image
-	helm upgrade --install kubestellar -n ${DEFAULT_WDS_NAME}-system ./core-helm-chart  --set ControlPlaneName=${DEFAULT_WDS_NAME}
-	kubectl -n ${DEFAULT_WDS_NAME}-system patch deployment kubestellar-controller-manager --type='json' -p='[{"op": "replace", "path": "/spec/template/spec/containers/1/imagePullPolicy", "value":"Never"}]'
+	helm upgrade $(if $(KUBE_CONTEXT),--kube-context $(KUBE_CONTEXT),) --install kubestellar -n ${DEFAULT_WDS_NAME}-system ./core-helm-chart  --set ControlPlaneName=${DEFAULT_WDS_NAME}
```
ezrasilvera, line 212: I know it's not part of theis PR - Can we change the DEFAULT_WDS_NAME  to WDS_NAME ?  We basically set a default value in the makefile. User can override it with external setting when running the Makefile

test/e2e/multi-cluster-deployment/run-test.sh
```diff
@@ -16,8 +16,9 @@
 # This is an end to end test of multi cluster deployement.  
 # For readable instructions, please visit https://github.com/kubestellar/kubestellar/tree/ks-0.20/docs/content/v0.20
 
+set -x # so users can see what is going on
 set -e # exit on error
 
+./cleanup.sh
```
effi-ofer, line 22: I am a bit conflicted here.  On the one hand, I personally love having a cleanup phase before the test.  On the other hand, this test case might be running against an existing kind cluster that have just been used for some other testing and it won't be right to remove them.  I won't object if you feel strongly about this change.  But let me list yet a third option - check for existence of needed clusters and fail the test here if pre-existing with a message recommending the user runs cleanup.sh.  I leave this in your very capable hands.

test/e2e/multi-cluster-deployment/setup-kubestellar.sh
```diff
@@ -13,27 +13,27 @@
...
 echo "Create a Kind hosting cluster with nginx ingress controller and KubeFlex operator"
 echo "-------------------------------------------------------------------------"
-kflex init --create-kind &> /dev/null
+kflex init --create-kind
 echo "Kubeflex kind cluster created."
 
 echo "Create an inventory & mailbox space of type vcluster running OCM (Open Cluster Management) directly in KubeFlex. Note that -p ocm runs a post-create hook on the vcluster control plane which installs OCM on it."
 echo "-------------------------------------------------------------------------"
-kflex create imbs1 --type vcluster -p ocm &> /dev/null
+kflex create imbs1 --type vcluster -p ocm
```
effi-ofer, line 26: Because of the way this test runs in CICD and the output is captured, this kflex create and other like it generate plenty of useless messages that overwhelm the test output.  Perhaps we should add a parameter to the test case that by default will display the output and in CICD will pipe to /dev/null or to a file?

README.md
```diff
@@ -55,7 +55,10 @@ TODO: Need update!
 
 ## QuickStart
 TBD
```
effi-ofer, line 57: Perhaps remove the TBD?

effi-ofer (approved):

test/e2e/multi-cluster-deployment/run-test.sh
MikeSpreitzer: This test _is_ destructive of the clusters involved. I have never heard anybody complain about a test destroying clusters that they wanted to preserve. OTOH, I have heard many complaints about tests failing due to conflict with pre-existing state. I suggest we start with this forceful initial cleanup, and if we get complaints then we can later introduce a choice between (a) forceful cleanup or (b) pre-flight checks with good error messages.

README.md
MikeSpreitzer: I think we still have significant work to do here, but do not really care about this line. Does not have to be changed in this PR, IMHO.

effi-ofer (approved):

Makefile
```diff
@@ -151,23 +151,23 @@ test: manifests generate fmt vet envtest ## Run tests.
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./cmd/kubestellar-operator/main.go $(ARGS)
 
-.PHONY: ko-build
-ko-build: test ## Build docker image with ko
-	KO_DOCKER_REPO=ko.local ko build --push=false -B ./cmd/${CMD_NAME} -t ${IMAGE_TAG} --platform linux/amd64,linux/arm64
+.PHONY: ko-build-local
+ko-build-local: test ## Build local container image with ko
+	$(shell (docker version | { ! grep -qi podman; } ) || echo "DOCKER_HOST=unix://$$HOME/.local/share/containers/podman/machine/qemu/podman.sock ") KO_DOCKER_REPO=ko.local ko build -B ./cmd/${CMD_NAME} -t ${IMAGE_TAG}
```
pdettori, line 156: @MikeSpreitzer this only works for podman ... what about users using docker engine on linux or rancher desktop on Mac ?

MikeSpreitzer: I will try removing the DOCKER_HOST setting from here and making it in my shell.

PR #1573, opened 2024-01-17
🌱 Add singleton test

MikeSpreitzer:
## Summary
This PR adds a test of the return of status in the 1:1 distribution case.

## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/pr-test-singleton-status.yml
+++ b/.github/workflows/pr-test-singleton-status.yml
@@ -1,0 +1,45 @@
+name: test singleton status
+
+on:
+  workflow_call:
+  # So we can trigger manually if needed
+  workflow_dispatch:
+  # To confirm any changes to docs build successfully, without deploying them
+  pull_request:
+    branches:
+      - ks-0.20
+  push:
+    branches:
+      - ks-0.20
+    tags:
+      - 'v*'
+
+jobs:
+  test-pr:
+    name: test-pr
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4.1.1
+      
+      - uses: actions/setup-go@v5
+        with:
+          go-version: v1.19
+          cache: true
+
+      - name: Install kubectl
+        uses: azure/setup-kubectl@v3
+        id: install
+
+      - uses: ko-build/setup-ko@v0.6
+      
+      - name: Install dependencies
+        run: |
+          curl -L https://raw.githubusercontent.com/open-cluster-management-io/clusteradm/main/install.sh | bash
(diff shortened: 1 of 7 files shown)


ezrasilvera:
@MikeSpreitzer If we say the status is a mandatory feature, why should we have 2 tests ?  It seems that the singleton test includes the pr-test, so let's just use the singleton test.

test/e2e/multi-cluster-deployment/setup-kubestellar.sh
```diff
@@ -13,27 +13,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -x # echo so users can understand what is happening
 set -e # exit on error
 
 echo "Create a Kind hosting cluster with nginx ingress controller and KubeFlex operator"
 echo "-------------------------------------------------------------------------"
-kflex init --create-kind &> /dev/null
+kflex init --create-kind
```
ezrasilvera, line 21: The problem is that kflex is using blocking commands and the github workflow sample that and bombard the output with a lot of messages which make the output unusable. 
We need to differentiate between local usage (where we want to see the output) and cicd test where we want pus to /dev/null.
Maybe add a redirection flag that can be set by the run-test which is not used in cicd

MikeSpreitzer: This change is from #1572, see the discussion there.

MikeSpreitzer:
The force-push to af40e2a67 is a rebase onto the latest edition of #1572

MikeSpreitzer:
@ezrasilvera : the reason for two tests is that they have different coverage. The multi-cluster test exercises 1:N distribution and no status return. The new test exercises 1:1 distribution AND status return.

MikeSpreitzer:
The force-push to af15c88af is a rebase, now that #1572 has merged.

ezrasilvera:
> The force-push to [af15c88](https://github.com/kubestellar/kubestellar/commit/af15c88afdec2116523e1eacc9ab089fe4ae37fb) is a rebase, now that #1572 has merged.

I think we can have the status be done on one of the two nodes of the other test

test/e2e/multi-cluster-deployment/setup-kubestellar.sh
pdettori: @ezrasilvera  we can open an issue in kubeflex to add a flag to the cli so that for ci/cd tests we do not do the noisy updates

MikeSpreitzer: https://github.com/kubestellar/kubeflex/issues/173

PR #1578, opened 2024-01-18
:bug: Fix syncer for some bugs reported during MCAD integration 

yana1205:
## Summary

This PR contains two fixes.
- Fix Syncer to avoid adding an entry to nil map
  - There is a case that an object is annotated as `downsync-overwrite` in MBS but the object on WEC has empty annotation. In this case, Syncer crashed due to adding an entry to nil map at [downsyncer.go#L303](https://github.com/kubestellar/kubestellar/blob/main/pkg/syncer/syncers/downsyncer.go#L303.)
- Fix Syncer not to overwrite managedFields

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/syncer/syncers/common.go
+++ b/pkg/syncer/syncers/common.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"strings"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


yana1205:
This PR is for former ks. This is no longer valid since ks-0.20 was merged into main branch.

MikeSpreitzer:
The `release-0.14` branch still exists.

PR #1582, opened 2024-01-18
✨ Accurate labeling on ManifestWorks

### LINKED ISSUE #1541: bug: assign correctly placement labels in transport manifest when different placements select overlapping groups of clusters
### Describe the bug

 The current logic for assigning placement labels  for an object that has overlapping placement with overlapping matching clusters is not correct. This is because of a shortcut that was taken in the placement logic implementation.
 
For example, object 1 has matching placement 1, that selects clusters 1 and 2, and placement 2, that selects clusters 2 and 3. The list of clusters for the object is cluster 1, cluster 2, cluster 3. The shortcut implementation adds the label for placement 1 and 2 to all copies of the manifest in cluster 1, 2 and 3.

### Steps To Reproduce

Test by creating overlapping placements as described in the bug, and examine the labels in the ManifestWork.

### Expected Behavior

the correct implementation should add the label for placement 1 on cluster 1, for placement 1 and 2 on cluster 2, and for placement 2 on cluster 3.

### Additional Context

_No response_

waltforme:
## Summary
This PR applies accurate 'managed-by' labels on Manifestworks which are delivered by the placement controller, thus handles #1541.

#1541 already clearly described the issue. Below is a visualization/example of that situation.
image

## Related issue(s)
Closes #1541

### DIFF
--- a/pkg/placement/controller.go
+++ b/pkg/placement/controller.go
@@ -30,6 +30,7 @@ import (
 	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -441,7 +442,46 @@ func (c *Controller) reconcile(ctx context.Context, key util.Key) error {
 	}
 
 	c.logger.Info("Delivering", "object", util.GenerateObjectInfoString(obj), "to clusters", clusters)
-	deliverObjectToManagedClusters(c.logger, c.ocmClient, obj, clusters, managedByPlacements, c.wdsName, singletonStatus)
+	for _, clName := range clusters {
+		// find which placement(s) select this managedCluster
+		placementNames := []string{}
+		for _, plName := range managedByPlacements {
+			plObj, err := c.getPlacementByName(plName)
+			if err != nil {
+				return err
+			}
+			pl, err := runtimeObjectToPlacement(plObj)
+			if err != nil {
+				return err
+			}
+			cl, err := ocm.GetClusterByName(c.ocmClient, clName)
+			if err != nil {
+				return err
+			}
+			// a placement selects a managedCluster iff the managedCluster is selected by every single selector
+			// i.e. selectors are ANDed
+			matches := true
+			for _, s := range pl.Spec.ClusterSelectors {
+				selector, err := metav1.LabelSelectorAsSelector(&s)
+				if err != nil {
+					return err
+				}
+				if !selector.Matches(labels.Set(cl.Labels)) {
+					matches = false
+					break
+				}
+			}
+			if matches {
+				placementNames = append(placementNames, plName)
+			}
+		}
+		manifest := ocm.WrapObject(obj)
+		util.SetManagedByPlacementLabels(manifest, c.wdsName, placementNames, singletonStatus)
+		err := reconcileManifest(c.ocmClient, manifest, clName)
+		if err != nil {
+			c.logger.Error(err, "Error delivering object to mailbox")
+		}
+	}
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


waltforme:
Here are the tests I did in my setup with 3 managed clusters. The workload object I'm checking is a namespace `nginx`.
```
ubuntu@ip-172-31-38-158:~/kubestellar$ kc --context imbs1 -n cluster1 get manifestwork v1-namespace--nginx -oyaml | yq .metadata.labels
managed-by.kubestellar.io/wds1.nginx-edge: "true"
ubuntu@ip-172-31-38-158:~/kubestellar$ kc --context imbs1 -n cluster2 get manifestwork v1-namespace--nginx -oyaml | yq .metadata.labels
managed-by.kubestellar.io/wds1.nginx-dev: "true"
managed-by.kubestellar.io/wds1.nginx-edge: "true"
ubuntu@ip-172-31-38-158:~/kubestellar$ kc --context imbs1 -n cluster3 get manifestwork v1-namespace--nginx -oyaml | yq .metadata.labels
managed-by.kubestellar.io/wds1.nginx-dev: "true"
```

nirrozenbaum:
we’re in the process of reviewing and merging of pluggable transport and all of this code will not exist. in addition, the way that transport works is different than what was done here (we wrap multiple objects not one).
I suggest holding with this PR until pluggable transport is merged and then to solve it with the new design and code.

pkg/placement/controller.go
```diff
@@ -441,7 +442,46 @@ func (c *Controller) reconcile(ctx context.Context, key util.Key) error {
...
+					break
+				}
+			}
+			if matches {
+				placementNames = append(placementNames, plName)
+			}
+		}
+		manifest := ocm.WrapObject(obj)
+		util.SetManagedByPlacementLabels(manifest, c.wdsName, placementNames, singletonStatus)
+		err := reconcileManifest(c.ocmClient, manifest, clName)
```
ezrasilvera, line 480: Do we need to do the above operations in case there were no matches at all and placementNames is empty ?

pkg/placement/controller.go
```diff
@@ -441,7 +442,46 @@ func (c *Controller) reconcile(ctx context.Context, key util.Key) error {
 	}
 
 	c.logger.Info("Delivering", "object", util.GenerateObjectInfoString(obj), "to clusters", clusters)
-	deliverObjectToManagedClusters(c.logger, c.ocmClient, obj, clusters, managedByPlacements, c.wdsName, singletonStatus)
+	for _, clName := range clusters {
+		// find which placement(s) select this managedCluster
+		placementNames := []string{}
+		for _, plName := range managedByPlacements {
+			plObj, err := c.getPlacementByName(plName)
+			if err != nil {
```
ezrasilvera, line 450: should we return an error if the error was for specific name? Maybe continue ?

pkg/placement/controller.go
waltforme: Good question. My understanding is that we should always have a non-empty list because in order for each cluster to be included into the `clusters` list, there must be at least one placement(s) that select it.

pkg/placement/controller.go
waltforme: I think we should return. Thoughts are: If we continue, we may end up with an incomplete list of `placementNames` for current cluster, thus an inaccurate set of labels on the ManifestWork. But I'm not 100% sure. Maybe @pdettori has a better view on this?

pdettori: @waltforme overall looks fine, but could all the added logic be implemented in `deliverObjectToManagedClusters` (changing / rewriting that function and its signature as you see fit) ?

pkg/placement/controller.go
pdettori: At this point there should be a non-empty list of clusters (that was checked before) and  a non empty list of managedByPlacements - since the latter is not checked it seems like a good idea to check

pdettori (changes requested):

pkg/placement/controller.go
waltforme: Thanks for the suggestion. I packed the logic into a method (not a standalone function) on `c *Controller` because I see the logic relies on `c` in multiple places.

pkg/placement/controller.go
waltforme: Done in the latest push.

pdettori (approved):
/lgtm

Issue #1537, opened 2024-01-12
feature: have client-go code generation


nirrozenbaum:
### Feature Description

in ks-0.20 branch, makefile doesn't include code generation for client-go informer/lister/etc.
all code is using dynamic client with unstructured objects instead.
it is preferred to have typed clients like we had in ks0.15 and use a dynamic client with unstructured only when there is no other option.

### Proposed Solution

implementation + tests

### Want to contribute?

- [ ] I would like to work on this issue.

nirrozenbaum:
@MikeSpreitzer can you help with this?
I'm assuming you know this part best and could do it quickly.

MikeSpreitzer:
@nirrozenbaum : yes, I can start on it Tuesday. I expect I will start from what is in KubeStellar release 0.15 and then switch from kcp-special to regular tools, and probably remove the patching (which I suspect was there only for kcp-specific reasons).

@pdettori : where is WorkStatus defined?

nirrozenbaum:
@MikeSpreitzer sounds good. thank you.

pdettori:
> @nirrozenbaum : yes, I can start on it Tuesday. I expect I will start from what is in KubeStellar release 0.15 and then switch from kcp-special to regular tools, and probably remove the patching (which I suspect was there only for kcp-specific reasons).
> 
> @pdettori : where is WorkStatus defined?

@MikeSpreitzer it is defined in the status-addon project in our internal GitHub, which we need to bring under https://github.com/kubestellar . We have also another issue here, there are incompatible dependencies between the libraries imported by the status-addon (from https://github.com/open-cluster-management-io/addon-framework ) and libraries we use in ks-0.20 (deep down the issue is about different and incompatible versions of the open telemetry agent). I could not find a way to resolve those conflicts with replace directives in the go.mod file, so for the time being  I have been using the dynamic client packages to avoid importing the status-add on project. We may need to think a better way for this, perhaps defining a separate project for all APIs we define that all projects can import without dragging along other dependencies that can be problematic, or have circular deps issues.

ezrasilvera:
@MikeSpreitzer The space-framework is already doing vanilla k8s code generation without any dependency on KCP. You can simply copy the generating script under hack and it should work as is (once you move the resource into its place)

ezrasilvera:
Also note that the dependency on KCP is  coming implicitly through the inclusion of tools.go which pulls in go.mod with its replacements of k8s --> kcp version. 
[...]

Issue #1533, opened 2024-01-12
feature: strong modularity boundaries between KubeFlex and ks-0.20


MikeSpreitzer:
### Feature Description

Almost all of ks-0.20 does not depend on KubeFlex. Almost all of KubeFlex makes no mention of ks-0.20. ks-0.20 should be usable in a simple deployment scenario that does not involve KubeFlex, using one cluster as WDS and IS and mailbox namespace holder. The only connections between KubeFlex and ks-0.20 are conveniences that could be moved out to a third place, leaving KubeFlex and ks-0.20 orthogonal to each other.

### Proposed Solution

details TBD

### Want to contribute?

- [ ] I would like to work on this issue.

ezrasilvera:
Agree. The plan is to support a flow in which we separate the creation of spaces and the deployment into them (e.g., of the transport and the 0.20 controller(s) ) .

pdettori:
Agreed on the orthogonality. I think though we may still allow to option to use the KF's post-create-hook to install the ks 0.20 controller. User may use that when creating the wds or may not use the -p option and then run helm to install the controller.

MikeSpreitzer:
@pdettori : yes, and the PostCreateHook that installs the ks-0.20 controller does not need to be in the kubeflex repo, right? This is part of what I said about the connections between the two being something that can be a third module that depends on KubeFlex and ks-0.20.

pdettori:
@MikeSpreitzer 
> the PostCreateHook that installs the ks-0.20 controller does not need to be in the kubeflex repo, right?

It does not, but we may need to find a 3rd party repo where to host the yaml for the hook, currently is in https://github.com/kubestellar/kubeflex/blob/main/config/samples/postcreate-hooks/kslight.yaml

nirrozenbaum:
totally agree on the orthogonality and also with Ezra that we should separate the creation of spaces and the deployment into them (as it was in SF).

ezrasilvera:
@MikeSpreitzer This is phase1 in my opinion. You can say high priority phase1.  People can start working on phase1 items while this is being implemented.

andreyod:
To make the Kubestellar operator independent from Kubeflex we need a way to provide the access(kubeconfig) to WDS and IMBS. The initial proposal is to use secrets. Kubeflex is already creates those secrets, and for other scenarios, the user(script) will be responsible to create them in the same way(name/namespace).

pdettori:
[...]

Issue #1517, opened 2024-01-11
feature: modify pluggable transport code to work with ks-0.20 design


nirrozenbaum:
### Feature Description

Current transport implementation used Space-Manager + Kube-Bind. 
the code has to be modified to match ks0.2 design and architecture, controller per WDS. 

### Proposed Solution

change implementation + tests.

### Want to contribute?

- [ ] I would like to work on this issue.

nirrozenbaum:
closed as completed

PR #1588, opened 2024-01-18
✨ pluggable transport initial work

nirrozenbaum:
## Summary

This PR is the initial work done of pluggable transport with the new ks-0.20 design as stated in #1517.
pay attention - currently there is no client-go code generation in this branch. this would be done in issue #1537 by @MikeSpreitzer in the next few days.
until that is done, for the completion of the initial skeleton I used generated code that I generated on a side project.
Once #1537 is completed I will make sure to validate the generated code usage in transport code.
issue #1533 was also taken into account, transport cmd receives kubeconfigs, so that transport controller is totally unaware of kubeflex or spaces.

## Related issue(s)

#1517
#1533
#1537

Fixes #


nirrozenbaum:
pay attention that placement decision current structure is not final and may change.
this is just an initial skeleton to initiate the review process.
I prefer doing the structure changes (if there are any) after we have code generation integrated.

another comment - most of the files are auto generated (go-client generation).
the only relevant files to review are:

whatever is under cmd/transport.
2 files under pkg/transport
pkg/apis/edge/v2alpha1/placementdecision.go - is the CRD definition.
all the rest is code generation and no need to review.

would appreciate your review and comments.
/cc @ezrasilvera
/cc @pdettori
/cc @vMaroon
/cc @shmuelk

ezrasilvera:
I see you use temporary codegen so I think It's a good practice to put don't merge label so it won't be merged by mistake with the wrong api generation

nirrozenbaum:
this PR has dependency on client-go generation. therefore we will merge it only after PR #1608 is merged into main.

nirrozenbaum:
rebased the PR on top of main and used codegen from main.

/unhold

api/edge/v1alpha1/types.go
```diff
@@ -171,6 +173,133 @@ type PlacementList struct {
...
+	Namespace string `json:"namespace,omitempty"`
+}
+
+type PlacementDecisionStatus struct {
+	// `specGeneration` identifies the generation of the spec that this is the status for.
+	// Zero means that no status has yet been written here.
+	// +optional
+	SpecGeneration int32 `json:"specGeneration,omitempty"`
+
+	// `propagatedWorkloadsCount` is the number of destinations that received all the resources in the `spec.workload`.
```
MikeSpreitzer, line 284: Any generation of that spec.workload?
What exactly does "received" mean? The agent got it? The agent successfully applied all of it?

api/edge/v1alpha1/types.go
```diff
@@ -171,6 +173,133 @@ type PlacementList struct {
...
+	// Standard object metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// `spec` explicitly describes a desired binding between workloads and Locations.
+	// It reflects the resolution of a Placement's selectors.
+	Spec PlacementDecisionSpec `json:"spec,omitempty"`
+
+	// `status` describes the status of the process of binding workloads to Locations.
```
MikeSpreitzer, line 199: The described status is not the status of a "decision", it is the status of the _distribution_ . Something needs to be renamed or refactored here.

api/edge/v1alpha1/types.go
```diff
@@ -171,6 +173,133 @@ type PlacementList struct {
...
+}
+
+// NamespaceScopeDownsyncObjects matches some objects of one particular namespaced resource.
+type NamespaceScopeDownsyncObjects struct {
+	// GroupResource holds the API group and resource name.
+	metav1.GroupResource `json:",inline"`
+
+	// `apiVeresion` holds just the version, not the group too.
+	// This is the version to use both upstream and downstream.
+	APIVersion string `json:"apiVersion"`
```
MikeSpreitzer, line 238: Why not just use `metav1.GroupVersionResource` ?

api/edge/v1alpha1/types.go
```diff
@@ -171,6 +173,133 @@ type PlacementList struct {
...
+
+	// `objects` holds the names of the objects of this kind to downsync.
+	// Empty list means none of them.
+	// +optional
+	Objects []string `json:"objects,omitempty"`
```
MikeSpreitzer, line 270: I think "ObjectNames" would be clearer here.

api/edge/v1alpha1/types.go
```diff
@@ -171,6 +173,133 @@ type PlacementList struct {
...
+}
+
+// PlacementDecisionSpecResources explicitly defines the resources to be down-synced.
+// The ClusterScope list defines the cluster-scope resources, while NamespacedObjects packs individual resources
+// identifiable by namespace & name.
+type PlacementDecisionSpecResources struct {
+	// `clusterScope` holds a list of individual cluster-scoped objects
+	// to downsync, organized by resource.
+	// Remember that a "resource" is a kind/type/sort of objects,
+	// not an individual object.
```
MikeSpreitzer, line 222: Yes. So it should be `PlacementDecisionSpecObjects` rather than `PlacementDecisionSpecResources`. And this field's type should be a slice of `ClusterScopeDownsyncObject` rather than `ClusterScopeDownsyncResource`.

api/edge/v1alpha1/types.go
```diff
@@ -171,6 +173,133 @@ type PlacementList struct {
...
+	Workload PlacementDecisionSpecResources `json:"workload,omitempty"`
+
+	// `destinations` is a list of cluster-identifiers that the resources should be propagated to.
+	Destinations []Destination `json:"destinations,omitempty"`
+}
+
+// PlacementDecisionSpecResources explicitly defines the resources to be down-synced.
+// The ClusterScope list defines the cluster-scope resources, while NamespacedObjects packs individual resources
+// identifiable by namespace & name.
+type PlacementDecisionSpecResources struct {
```
MikeSpreitzer, line 218: I think we could use a shorter name, describing what one of these is rather than where it gets used. And that is a clearer way to name something anyway, right? How about `DownsyncObjectReferences` ? Or `DownsyncWorkloadParts` ?

api/edge/v1alpha1/types.go
```diff
@@ -171,6 +173,133 @@ type PlacementList struct {
...
+
+	// `objects` holds the names of the objects of this kind to downsync.
+	// Empty list means none of them.
+	// +optional
+	Objects []string `json:"objects,omitempty"`
+}
+
+// Destination wraps the identifiers required to uniquely identify a destination cluster.
+type Destination struct {
+	Namespace string `json:"namespace,omitempty"`
```
MikeSpreitzer, line 275: So I was expecting something like a cluster name?

MikeSpreitzer:
Did you compare #1222 ?

api/edge/v1alpha1/types.go
nirrozenbaum: changed it to clusterId

api/edge/v1alpha1/types.go
nirrozenbaum: right. this was inherited from syncer config. with this structure.
fixed.

nirrozenbaum:
> Did you compare #1222 ?

@MikeSpreitzer I looked in #1222 at the beginning of the work on the initial design.
the design we agreed on is different in many aspects from what it was in this PR and therefore I do see the point in consulting with this PR. we also don't want to be biased by design that was agreed by the team that should be abandoned.

I saw many of the comments are about names/description. We took the initial structures from syncer config to align with the existing names/descriptions (who knew that we will redesign even before merging transport). the comments are all legitimate and I'll update the names/descriptions.

api/edge/v1alpha1/types.go
nirrozenbaum: done. changed to `DownsyncObjectsReferences`

api/edge/v1alpha1/types.go
nirrozenbaum: done

api/edge/v1alpha1/types.go
nirrozenbaum: done

api/edge/v1alpha1/types.go
MikeSpreitzer: yes

api/edge/v1alpha1/types.go
```diff
@@ -171,6 +173,125 @@ type PlacementList struct {
...
+	Workload DownsyncObjectsReferences `json:"workload,omitempty"`
+
+	// `destinations` is a list of cluster-identifiers that the objects should be propagated to.
+	Destinations []Destination `json:"destinations,omitempty"`
+}
+
+// DownsyncObjectsReferences explicitly defines the objects to be down-synced.
+// The ClusterScope list defines the cluster-scope objects, while NamespacedObjects packs individual objects
+// identifiable by namespace & name.
+type DownsyncObjectsReferences struct {
```
MikeSpreitzer, line 218: We do not need two plurals. I think we want "DownsyncObjectReferences".

api/edge/v1alpha1/types.go
```diff
@@ -171,6 +173,125 @@ type PlacementList struct {
...
+	// `namespace` identifies the namespace
+	Namespace string `json:"namespace"`
+
+	// `names` holds the names of the objects that match.
+	// Empty list means none of them.
+	// +optional
+	Names []string `json:"names,omitempty"`
+}
+
+type ClusterScopeDownsyncObject struct {
```
MikeSpreitzer, line 255: Since this identifies generally multiple objects, I would put a plural in the type name. "ClusterScopeDownsyncObjects".

api/edge/v1alpha1/types.go
MikeSpreitzer: yes

api/edge/v1alpha1/types.go
MikeSpreitzer: yes

api/edge/v1alpha1/types.go
nirrozenbaum: done

api/edge/v1alpha1/types.go
nirrozenbaum: done

api/edge/v1alpha1/types.go
nirrozenbaum: I updated the descriptions in status fields. let me know if it's clearer now.

api/edge/v1alpha1/types.go
nirrozenbaum: how would you call it?

api/edge/v1alpha1/types.go
MikeSpreitzer: will open a new thread on the current edition.

api/edge/v1alpha1/types.go
```diff
@@ -171,6 +173,129 @@ type PlacementList struct {
...
+type PlacementDecisionStatus struct {
+	// observedGeneration represents the .metadata.generation that the status is based upon.
+	// For instance, if .metadata.generation is currently 12, but the .status.observedGeneration is 9, the status is out of date
+	// with respect to the current state of the PlacementDecision instance.
+	// Zero means that no status has yet been written here.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// `propagatedWorkloadsCount` represents the number of different destinations that got a copy of the wrapped object
```
MikeSpreitzer, line 279: _the_ wrapped object? I thought the current code (pre-pluggable-transport) is wrapping each workload object individually and that the switch to pluggable transport would preserve that behavior. Certainly the API should not preclude that behavior.

api/edge/v1alpha1/types.go
MikeSpreitzer: Note also that propagation status has several facets. Propagation from WDS to mailbox. Propagation from mailbox to agent in WEC. Success or failure of agent to create/update object in WEC.

My thinking has been to (A) define something that exposes all these details in the mailbox and (B) let the user define summarization(s) of the aspects in the way that works best for them.

api/edge/v1alpha1/types.go
nirrozenbaum: the pluggable-transport do not wrap each workload object individually.
I think you were out of office when we discussed transport with Gosia and the others.
we put all objects that are selected by a placement into a single wrapped object.
we also discussed that with Paolo.

api/edge/v1alpha1/types.go
nirrozenbaum: this is the common k8s usage. status part reports the current state while spec part is the desired state.

nirrozenbaum: you're right about the several facets.
in this PR the facet is propagation from WDS to mailbox. 
this is what the transport controller is responsible for.
from the mailbox the objects are taken care by the agent.
we do plan on adding the other status of propagation from mailbox to WEC, but that is not part of this PR.
 
 please note that this PR do NOT cover the whole functionality of pluggable-transport. it just gives the starting point for phase1 as it was discussed with the broader team.
 
 I agree with you on both points (A) and (B), but both of them are out of scope of this PR. adding the status of mb to WEC propagation is more straight forward and can be added easily in an additional PR after this one is merged.
 the idea was to split PRs into smaller and easier to review PRs rather than one huge PR.
  
 regarding status and/or summarization, this should be designed more carefully before adding anything. 
 this topic was clearly agreed by the broader team and executives as out of scope for current release.

nirrozenbaum: btw, I documented that the status is for propagation from wds to mailbox in comments:
https://github.com/kubestellar/kubestellar/blob/transport-ks0.20/api/edge/v1alpha1/types.go#L279-#L280

api/edge/v1alpha1/types.go
```diff
@@ -20,6 +20,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+/******************** Start Placement Type *********************/
```
ezrasilvera, line 23: Why do we need these start/end comments ?

nirrozenbaum: we do not. removed.

api/edge/v1alpha1/types.go
ezrasilvera: @nirrozenbaum - @MikeSpreitzer was talking about the pre-pluggable-transport. 
So the answer is - yes the current  pre-pluggable-transport code wrap each object. As part of the pluggable transport mode the team suggested to move to a different approach of wrapping all object in the placement into single wrapping object. There are pros/cons for each of the approaches,  It was agreed to continue with this new approach the team started:
1. The specific approach doesn't change the user facing API
2. When going with the new approach all current functionality need to be supported (The team need to validate)
3. @pdettori had some concerns - the team addressed the issues he raised and he approved
4. changing back to the original approach is estimated as small effort (up to a week) - so we can always decide to switch back if we encounter major reason in the future.

api/edge/v1alpha1/types.go
MikeSpreitzer: I also think that we should replace the one monolithic `Spec PlacementDecisionSpec` with two collections of slices (where decision slices and what decision slices) API objects. If we take the Spec out of the PlacementDecision object and leave WDS-to-mailbox propagation status then it clearly is no longer a placement _decision_ object.

For the WDS-to-mailbox propagation status, think also about overlap between Placement objects.

MikeSpreitzer: There is no WDS-to-mailbox propagation status returned in release-0.14, and not currently in `main` either (if I understand correctly). Maybe we can separate the work on that status from the initial pluggable transport work.

api/edge/v1alpha1/types.go
MikeSpreitzer: Did that discussion include consideration of overlapping Placement objects and the consequent duplication in wrapped bundles?

MikeSpreitzer: Defining a status field that only makes sense in one approach precludes the other.

api/edge/v1alpha1/types.go
nirrozenbaum: I'm not sure I understand the point of the last comment.
The design is completely different than what it was in release 0-14, obviously there are many different things + things that were not part of previous releases. same goes for current main.

nirrozenbaum: regarding the point about replacing `Spec` - I disagree with the comment.
every CR has spec and status fields, this is the k8s way of defining CRs. we don't want to re-invent the wheel.
we use the k8s common patterns by defining spec and status fields and under the spec you can find the two collections you specified. these two collections represent the desired state of this CR (desired state in k8s CRs is what's under spec, not an arbitrary definition of two collections).

api/edge/v1alpha1/types.go
nirrozenbaum: yes. the discussion included overlapping placement objects and duplication of wrapped bundles.
there is no issue with that.
I didn't understand why you think the status field doesn't make sense.
transport controller watches CR with generation 'x' and creates a manifest work in destination mailbox ns.
we need to be sure that the manifest work content matches whatever there is in our CR with generation 'x'.

api/edge/v1alpha1/types.go
MikeSpreitzer: An object having spec and status is the common pattern but not the only pattern. For examples in Kubernetes: ConfigMap, Secret, Endpoints, EndpointsSlice. The last one is an inspiration for what we did about the "where" resolution in the release 0.15 and prior, and the extension I proposed in #1222 . In the world of "compliance" and "policy" it is not uncommon to have reports in different objects than policies. See the work of the Compliance-to-Policy project for an example.

I think there is design work still to do on status reporting. I suggested leaving the status reported out of the initial pluggable transport work so that the status reporting design work not hold up the other initial pluggable transport work.

nirrozenbaum: I'm OK leaving the status fields out of this PR.
let me update the api definition quickly.

api/edge/v1alpha1/types.go
MikeSpreitzer: The go comment here refers to "the [thus, one] wrapped object that contains all the objects in the" workload. That asserts one approach and precludes the other.

api/edge/v1alpha1/types.go
nirrozenbaum: done

api/edge/v1alpha1/types.go
nirrozenbaum: remove status from this PR.
we'll leave it for next PRs

MikeSpreitzer (commented):
/LGTM
Thanks!

MikeSpreitzer:
Oh, spoke too soon, also need to review the other stuff here.
/remove-lgtm

cmd/transport/base-controller.go
```diff
@@ -0,0 +1,99 @@
...
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+
+	transportoptions "github.com/kubestellar/kubestellar/cmd/transport/options"
+	edgeclientset "github.com/kubestellar/kubestellar/pkg/generated/clientset/versioned"
+	edgeinformers "github.com/kubestellar/kubestellar/pkg/generated/informers/externalversions"
+	"github.com/kubestellar/kubestellar/pkg/transport"
+)
+
```
MikeSpreitzer, line 37: A comment here outlining what this thing does would be very helpful, particularly since it is not described anywhere else.

cmd/transport/base-controller.go
```diff
@@ -0,0 +1,99 @@
...
+	options := transportoptions.NewTransportOptions()
+	fs := pflag.NewFlagSet(transport.ControllerName, pflag.ExitOnError)
+	klog.InitFlags(flag.CommandLine)
+	fs.AddGoFlagSet(flag.CommandLine)
+	options.AddFlags(fs)
+	fs.Parse(os.Args[1:])
+
+	fs.VisitAll(func(flg *pflag.Flag) {
+		logger.Info(fmt.Sprintf("Command line flag '%s' - value '%s'", flg.Name, flg.Value)) // log all arguments
```
MikeSpreitzer, line 54: Why not use the defined structure of Logger::Info --- the keysAndValues? Something like

```
logger.Info("Command line flag", flg.Name, flg.Value)
```

or

```
logger.Info("Command line flag", "name", flg.Name, "value", flg.Value)
```

cmd/transport/base-controller.go
```diff
@@ -0,0 +1,99 @@
...
+		klog.FlushAndExit(klog.ExitFlushTimeout, 1)
+	}
+
+	edgeSharedInformerFactory := edgeinformers.NewSharedInformerFactoryWithOptions(edgeClientset, defaultResyncPeriod)
+	edgePlacementDecisionInformer := edgeSharedInformerFactory.Edge().V1alpha1().PlacementDecisions()
+
+	transportController, err := transport.NewTransportController(ctx, edgePlacementDecisionInformer, transportImplementation, edgeClientset, wdsDynamicClient, transportRestConfig, options.WdsName)
+
+	// notice that there is no need to run Start method in a separate goroutine.
+	// Start method is non-blocking and runs all registered informers in a dedicated goroutine.
```
MikeSpreitzer, line 90: s/all registered informers in a/each of the factory's informers in its own/

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+const (
+	ControllerName                 = "transport-controller"
+	transportFinalizer             = "transport.kubestellar.io/object-cleanup"
+	notFoundErrorSuffix            = "not found"
+	originOwnerReferenceAnnotation = "transport.kubestellar.io/originOwnerReferencePlacementDecisionKey"
+	originWdsAnnotation            = "transport.kubestellar.io/originWdsName"
+)
+
+// NewTransportController returns a new transport controller
+func NewTransportController(ctx context.Context, placementDecisionInformer edgev1alpha1informers.PlacementDecisionInformer, transport Transport,
```
MikeSpreitzer, line 59: FYI, Kubernetes has established confused terminology:
```
type PlacementDecisionInformer interface {
	Informer() cache.SharedIndexInformer
	Lister() v1alpha1.PlacementDecisionLister
}
```

Thus, one "informer" of PlacementDecisions has a method `Informer` that produces another "informer" of PlacementDecisions (alongside a method that produces a lister).

FYI, I tend to name variables of the former type "...PreInformer" rather than "Informer", to make the distinction clearer.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+	wdsClientset *edgeclientset.Clientset, wdsDynamicClient *dynamic.DynamicClient, transportRestConfig *rest.Config, wdsName string) (*genericTransportController, error) {
```
MikeSpreitzer, line 60: We will want to have unit tests. They will be easier if this func takes the relevant clients instead of `transportRestConfig *rest.Config`, because there are standard fake clients for testing but not standard fake `*rest.Config`s.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+
+func convertObjectToUnstructured(object runtime.Object) (*unstructured.Unstructured, error) {
+	unstructuredObject, err := runtime.DefaultUnstructuredConverter.ToUnstructured(object)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert given object to unstructured - %w", err)
+	}
+	return &unstructured.Unstructured{Object: unstructuredObject}, nil
+}
+
+func getGvrFromWrappedObject(restConfig *rest.Config, wrappedObject runtime.Object) (*schema.GroupVersionResource, error) {
```
MikeSpreitzer, line 118: Returning a pointer here requires maintainers to think harder than return just the `schema.GroupVersionResource`. With return of a pointer _into_ a data structure returned from an upstream library, maintainers have to be sure that nothing assigns into that pointer or that such an assignment will not cause grief to the upstream library. In the error cases, this func could return a `schema.GroupVersionResource{}`.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+		return nil, fmt.Errorf("failed to create dynamic k8s clientset for transport space - %w", err)
+	}
+
+	emptyWrappedObject := transport.WrapObjects(make([]*unstructured.Unstructured, 0)) // empty wrapped object to get GVR from it.
+	wrappedObjectGVR, err := getGvrFromWrappedObject(transportRestConfig, emptyWrappedObject)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get transport wrapped object GVR - %w", err)
+	}
+
+	dynamicInformer := dynamicinformer.NewDynamicSharedInformerFactory(transportDynamicClient, 0)
```
MikeSpreitzer, line 72: I would put "Factory" in this variable's name.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+		wdsDynamicClient:                wdsDynamicClient,
+		wdsName:                         wdsName,
+	}
+
+	transportController.logger.Info("Setting up event handlers")
+	// Set up an event handler for when PlacementDecision resources change
+	placementDecisionInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    transportController.enqueuePlacementDecision,
+		UpdateFunc: func(_, new interface{}) { transportController.enqueuePlacementDecision(new) },
+		DeleteFunc: transportController.enqueuePlacementDecision,
```
MikeSpreitzer, line 94: The DeleteFunc needs to be able to handle a `cache.DeletedFinalStateUnknown` (see https://github.com/kubernetes/client-go/blob/v0.28.2/tools/cache/controller.go#L213-L216).

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+// extracts the origin PlacementDecision of the given wrapped object and
+// enqueue that PlacementDecision object for processing. This way, we
+// don't need to implement custom logic for handling WrappedObject resources.
+// More info on this pattern here:
+// https://github.com/kubernetes/community/blob/8cafef897a22026d42f5e5bb3f104febe7e29830/contributors/devel/controllers.md
+func (c *genericTransportController) handleWrappedObject(obj interface{}) {
+	wrappedObject := obj.(metav1.Object)
+	ownerPlacementDecisionKey, found := wrappedObject.GetAnnotations()[originOwnerReferenceAnnotation] // safe if GetAnnotations() returns nil
+	if !found {
+		c.logger.Info("failed to extract placementdecision key from transport wrapped object", "WrappedObjectName", wrappedObject.GetName())
```
MikeSpreitzer, line 186: The namespace should also be logged, right?

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+	wdsName          string
+}
+
+// enqueuePlacementDecision takes an PlacementDecision resource and
+// converts it into a namespace/name string which is put onto the workqueue.
+// This func *shouldn't* handle any resource other than PlacementDecision.
+func (c *genericTransportController) enqueuePlacementDecision(obj interface{}) {
+	var key string
+	var err error
+	if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
```
MikeSpreitzer, line 168: If this line were to call https://github.com/kubernetes/client-go/blob/v0.28.2/tools/cache/store.go#L121 instead and the queue held `ObjectName` values then ` func (c *genericTransportController) process` could be simpler --- no parsing needed there.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+// isObjectBeingDeleted is a helper function to check if object is being deleted.
+func isObjectBeingDeleted(object metav1.Object) bool {
+	return !object.GetDeletionTimestamp().IsZero()
+}
+
+func (c *genericTransportController) deleteWrappedObjectsAndFinalizer(ctx context.Context, placementDecision *v1alpha1.PlacementDecision) error {
+	for _, destination := range placementDecision.Spec.Destinations { // TODO need to revisit the destination struct and see how to use it properly
+		if err := c.transportClient.Resource(c.wrappedObjectGVR).Namespace(destination.ClusterId).Delete(ctx, fmt.Sprintf("%s-%s", placementDecision.GetName(), c.wdsName),
+			metav1.DeleteOptions{}); err != nil { // wrapped object name is in the format (PlacementDecision.GetName()-WdsName). see updateWrappedObject func for explanation.
+			if !strings.HasSuffix(err.Error(), notFoundErrorSuffix) { // if object is already not there, we do not report an error cause desired state was achieved.
```
MikeSpreitzer, line 321: Wouldn't it be better to do something with data structure rather than string parsing?

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+
+	if err := c.removeFinalizerFromPlacementDecision(ctx, placementDecision); err != nil {
+		return fmt.Errorf("failed to remove finalizer from PlacementDecision object '%s' - %w", placementDecision.GetName(), err)
+	}
+
+	return nil
+}
+
+func (c *genericTransportController) removeFinalizerFromPlacementDecision(ctx context.Context, placementDecision *v1alpha1.PlacementDecision) error {
+	return c.updatePlacementDecision(ctx, placementDecision, func(placementDecision *v1alpha1.PlacementDecision) bool {
```
MikeSpreitzer, line 336: The objects returned from a PlacementDecisionLister must be treated as read-only.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+	}
+
+	objectsToPropagate, err := c.getObjectsFromWDS(ctx, placementDecision)
+	if err != nil {
+		return fmt.Errorf("failed to get objects to propagate to WECs from PlacementDecision object '%s' - %w", placementDecision.GetName(), err)
+	}
+
+	wrappedObject, err := convertObjectToUnstructured(c.transport.WrapObjects(objectsToPropagate))
+	if err != nil {
+		return fmt.Errorf("failed to wrap objects to a single wrapped object - %w", err)
```
MikeSpreitzer, line 359: The failure here is not in the wrapping but just in the conversion to unstructured.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+	}
+	// wrapped object name is (PlacementDecision.GetName()-WdsName).
```
MikeSpreitzer, line 361: This name composition scheme has two problems.
1. Length. What guarantees that the composite name is not too long? I do not see anything. One solution would be to use object UIDs instead of names.
2. Uniqueness. Both of the input names are under use control and can contain a dash ("-"), if I understand correctly. That means that concatenating them with a dash in between is not guaranteed to create unique names. I think that you could avoid this problem by choosing a separator that is not a valid charater to appear in `c.wdsName` (these are fairly restricted, right? Where is it documented?).

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+}
+
+// isObjectBeingDeleted is a helper function to check if object is being deleted.
+func isObjectBeingDeleted(object metav1.Object) bool {
+	return !object.GetDeletionTimestamp().IsZero()
+}
+
+func (c *genericTransportController) deleteWrappedObjectsAndFinalizer(ctx context.Context, placementDecision *v1alpha1.PlacementDecision) error {
+	for _, destination := range placementDecision.Spec.Destinations { // TODO need to revisit the destination struct and see how to use it properly
+		if err := c.transportClient.Resource(c.wrappedObjectGVR).Namespace(destination.ClusterId).Delete(ctx, fmt.Sprintf("%s-%s", placementDecision.GetName(), c.wdsName),
```
MikeSpreitzer, line 319: You might not want to give users access to all namespaces in the transport space.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+	// add cluster-scoped objects to the 'objectsToPropagate' slice
+	for _, clusterScopedObject := range placementDecision.Spec.Workload.ClusterScope {
+		if clusterScopedObject.ObjectNames == nil {
+			continue // no objects from this gvr, skip
+		}
+		gvr := schema.GroupVersionResource{Group: clusterScopedObject.Group, Version: clusterScopedObject.Version, Resource: clusterScopedObject.Resource}
+		for _, objectName := range clusterScopedObject.ObjectNames {
+			object, err := c.wdsDynamicClient.Resource(gvr).Get(ctx, objectName, metav1.GetOptions{})
+			if err != nil {
+				return nil, fmt.Errorf("failed to get required cluster-scoped object '%s' with gvr %s from WDS - %w", objectName, gvr, err)
```
MikeSpreitzer, line 389: This reaction is a little harsh. This `Get` is happening some time after something else determined that the object should be propagated, and the user might have deleted the object in the meantime.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, placementDecision *v1alpha1.PlacementDecision) ([]*unstructured.Unstructured, error) {
+	objectsToPropagate := make([]*unstructured.Unstructured, 0)
+	// add cluster-scoped objects to the 'objectsToPropagate' slice
+	for _, clusterScopedObject := range placementDecision.Spec.Workload.ClusterScope {
+		if clusterScopedObject.ObjectNames == nil {
+			continue // no objects from this gvr, skip
+		}
+		gvr := schema.GroupVersionResource{Group: clusterScopedObject.Group, Version: clusterScopedObject.Version, Resource: clusterScopedObject.Resource}
+		for _, objectName := range clusterScopedObject.ObjectNames {
+			object, err := c.wdsDynamicClient.Resource(gvr).Get(ctx, objectName, metav1.GetOptions{})
```
MikeSpreitzer, line 387: `c.wdsDynamicClient.Resource(gvr)` could be computed once, before the loop

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+	}
+	// add namespace-scoped objects to the 'objectsToPropagate' slice
+	for _, namespaceScopedObject := range placementDecision.Spec.Workload.NamespaceScope {
+		gvr := schema.GroupVersionResource{Group: namespaceScopedObject.Group, Version: namespaceScopedObject.Version, Resource: namespaceScopedObject.Resource}
+		for _, objectsByNamespace := range namespaceScopedObject.ObjectsByNamespace {
+			if objectsByNamespace.Names == nil {
+				continue // no objects from this namespace, skip
+			}
+			for _, objectName := range objectsByNamespace.Names {
+				object, err := c.wdsDynamicClient.Resource(gvr).Namespace(objectsByNamespace.Namespace).Get(ctx, objectName, metav1.GetOptions{})
```
MikeSpreitzer, line 402: `c.wdsDynamicClient.Resource(gvr)` could be computed once, before the outer loop; `.Namespace(objectsByNamespace.Namespace)` could be done once, before the inner loop.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+				if err != nil {
+					return nil, fmt.Errorf("failed to get required namespace-scoped object '%s' with gvr '%s' from WDS - %w", objectName, gvr, err)
```
MikeSpreitzer, line 404: The namespace should be reported too.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+		}
+	}
+
+	return objectsToPropagate, nil
+}
+
+func (c *genericTransportController) createOrUpdateWrappedObject(ctx context.Context, namespace string, wrappedObject *unstructured.Unstructured) error {
+	existingWrappedObject, err := c.transportClient.Resource(c.wrappedObjectGVR).Namespace(namespace).Get(ctx, wrappedObject.GetName(), metav1.GetOptions{})
+	if err != nil {
+		if !strings.HasSuffix(err.Error(), notFoundErrorSuffix) { // if object is not there, we need to create it. otherwise report an error.
```
MikeSpreitzer, line 417: Why not use `errors.IsNotFound(err)` here?

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+		}
+		// object not found when using get, create it
+		_, err = c.transportClient.Resource(c.wrappedObjectGVR).Namespace(namespace).Create(ctx, wrappedObject, metav1.CreateOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to create wrapped object '%s' in destination WEC mailbox namespace '%s' - %w", wrappedObject.GetName(), namespace, err)
+		}
+		return nil
+	}
+	// // if we reached here object already exists, try update object
+	wrappedObject.SetResourceVersion(existingWrappedObject.GetResourceVersion())
```
MikeSpreitzer, line 428: What goes wrong if this line is omitted?

cmd/transport/options/options.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/spf13/pflag"
+)
+
+const (
+	defaultConcurrency = 4
+)
+
+type TransportOptions struct {
+	Concurrency         int
+	WdsKubeConfig       string
```
MikeSpreitzer, line 29: The kubeconfig context seems to be an important variable too. There are other common `kubectl` flags that could be exposed. See https://github.com/kubestellar/kubestellar/blob/v0.14.0/cmd/placement-translator/main.go#L84-L88 for example.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,512 @@
...
+
+	// workqueue is a rate limited work queue.
+	// This is used to queue work to be processed instead of performing it as soon as a change happens.
+	// This means we can ensure we only process a fixed amount of resources at a time, and makes it
+	// easy to ensure we are never processing the same item simultaneously in two different workers.
+	workqueue workqueue.RateLimitingInterface
+
+	//transport is a specific implementation for the transport interface.
+	transport        Transport
+	transportClient  dynamic.Interface // dynamic client to transport wrapped object. since object kind is unknown during complilation, we use dynamic
```
MikeSpreitzer, line 154: Have you considered using golang generics?

pkg/transport/generic_transport_controller.go
nirrozenbaum: if we don't set the resource version update fails with the following error:
```
metadata.resourceVersion: Invalid value: "": must be specified for an update
```

cmd/transport/base-controller.go
nirrozenbaum: this is how I wrote it initially but it had some linter warnings.
(at least it used to have linter warnings when I worked on transport against release 0.15).
if linter configuration has changed and this doesn't create warnings I'm happy to do this change.

pkg/transport/generic_transport_controller.go
nirrozenbaum: yes. I'm totally aware of this.
the names I used are following the patterns from the community sample controller:
https://github.com/kubernetes/sample-controller/blob/master/controller.go#L94

whoever is reading this go-client code should be able to understand it.

pkg/transport/generic_transport_controller.go
nirrozenbaum: no. using dynamic client here for a given GVR is very straight forward and simple.
don't think we need generics here.

cmd/transport/base-controller.go
nirrozenbaum: done

pkg/transport/generic_transport_controller.go
nirrozenbaum: done

pkg/transport/generic_transport_controller.go
nirrozenbaum: yes, missed it. 
done.

pkg/transport/generic_transport_controller.go
nirrozenbaum: done.

pkg/transport/generic_transport_controller.go
nirrozenbaum: right. done.

pkg/transport/generic_transport_controller.go
nirrozenbaum: good point. done.

pkg/transport/generic_transport_controller.go
nirrozenbaum: the mapping of clusterId into namespace and transport space shard will be done as part of the iteration on cluster inventory. for now keeping it minimal to get things working.

pkg/transport/generic_transport_controller.go
ezrasilvera: Indeed it's confusing. I'm not sure PreXX is helping here (Personally I was always confused by Pre - I thought it performs operations prior to those of the informer method )
It's more like InformerFactory  but probably not that as well.   
I'm fine leaving it like this unless we comup with a good naming convention . @pdettori ?

nirrozenbaum: I'm not sure I understand why do we need to come up with new naming convention while using `informer` IS THE naming convention in the community.
look for your favorite controller in k8s org, this is the pattern:
https://github.com/kubernetes/kubernetes/blob/81986587efb506c54632babf0ec79592da79597a/pkg/controller/volume/persistentvolume/pv_controller_base.go#L70-#L74

pkg/transport/generic_transport_controller.go
nirrozenbaum: done.

cmd/transport/base-controller.go
nirrozenbaum: done

pkg/transport/generic_transport_controller.go
nirrozenbaum: fixed

pkg/transport/generic_transport_controller.go
nirrozenbaum: done.
handling function is now using `cache.DeletionHandlingMetaNamespaceKeyFunc`

pkg/transport/generic_transport_controller.go
nirrozenbaum: I prefer keeping it simple. 
I don't feel like adding data structure would be better in any way.

pkg/transport/generic_transport_controller.go
nirrozenbaum: I prefer using the `cache.DeletionHandlingMetaNamespaceKeyFunc` function which covers also the case of `cache.DeletedFinalStateUnknown` and have only this single handling function for all event types.

pkg/transport/generic_transport_controller.go
nirrozenbaum: done

cmd/transport/options/options.go
nirrozenbaum: done.
I've added under `cmd` a new dir `client-options` which can be reused by the different controllers (e.g., placement controller needs it as well).
I then used it in the transport controller options+cmd.

pkg/transport/generic_transport_controller.go
nirrozenbaum: under the basic rule that wdsName is unique, the concatenated name suffix is `-unique_suffix`.
unless really unreasonable names are chosen, we get uniqueness.
of course someone can choose unreasonable names but I think that doesn't justify blocking the PR.
we can take care of this in a later PR.

cmd/transport/base-controller.go
MikeSpreitzer: The first of the two I suggested causes linter warnings; the second does not (in my experience in this project).

nirrozenbaum: done

pkg/transport/generic_transport_controller.go
MikeSpreitzer: The value is "pre" in the sense that you get this value _before_ you get the actual informer.

It is OK with me to leave this code as it is. I was just suggesting something that I find helpful. I do not think we have to follow examples in all details. Remember that the code upstream has evolved over time, and has a lot of warts due to its history and the fact that it evolved by changing only what _had_ to be changed and got limited review.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,507 @@
...
+	if err != nil {
+		return schema.GroupVersionResource{}, fmt.Errorf("failed to convert wrapped object to unstructured - %w", err)
+	}
+
+	gvk := unstructuredWrappedObject.GroupVersionKind()
+	mapper := restmapper.NewDeferredDiscoveryRESTMapper(cacheddiscovery.NewMemCacheClient(clientset.Discovery()))
+
+	restMapping, err := mapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+	if err != nil {
+		return schema.GroupVersionResource{}, fmt.Errorf("failed to discover GroupVersionResource from given GroupVersionKind - %w", err)
```
MikeSpreitzer, line 122: I would show the GVK in the error.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,507 @@
...
+const (
+	ControllerName                 = "transport-controller"
+	transportFinalizer             = "transport.kubestellar.io/object-cleanup"
+	originOwnerReferenceAnnotation = "transport.kubestellar.io/originOwnerReferencePlacementDecisionKey"
+	originWdsAnnotation            = "transport.kubestellar.io/originWdsName"
+)
+
+// NewTransportController returns a new transport controller
+func NewTransportController(ctx context.Context, placementDecisionInformer edgev1alpha1informers.PlacementDecisionInformer, transport Transport,
+	wdsClientset *edgeclientset.Clientset, wdsDynamicClient *dynamic.DynamicClient, transportClientset *kubernetes.Clientset,
```
MikeSpreitzer, line 57: This func could take an `edgeclientset.Interface` instead of a `*edgeclientset.Clientset`, right?
Same for transport and dynamic, right?
That would make it easy to write unit tests that pass the standard fakes.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: Even with unique wdsName values, the concatenations are not guaranteed to be unique. For example: "momma-mia" + "-" + "farrow" == "momma" + "-" + "mia-farrow". I think we should not rely on users to just know that they have to avoid doing this. There is also the length issue.

OK with me to address later, if you have an Issue to do so.

pkg/transport/generic_transport_controller.go
nirrozenbaum: yes. from some reason I used in one place in interface and in others I didn't.
(in the struct transport dynamic client was interface).
that was a glitch.
fixed.

pkg/transport/generic_transport_controller.go
nirrozenbaum: I prefer getting this PR merged asap cause it's blocking Maroon in the continuation PR.
minor items like this can be addressed later in very small PR.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: Thanks.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,507 @@
...
+	_, err = c.transportClient.Resource(c.wrappedObjectGVR).Namespace(namespace).Update(ctx, wrappedObject, metav1.UpdateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to update wrapped object '%s' in destination WEC mailbox namespace '%s' - %w", wrappedObject.GetName(), namespace, err)
+	}
+
+	return nil
+}
+
+// updateObjectFunc is a function that updates the given object. returns true if object was updated, otherwise false.
+type updateObjectFunc func(*v1alpha1.PlacementDecision) bool
```
MikeSpreitzer, line 430: This type is not for _any_ object, only a specific kind of objects; I would name it something more specific like `updateDecisionFunc` to `updatePlacementDecisionFunc`. This is not a blocker.

pkg/transport/generic_transport_controller.go
```diff
@@ -0,0 +1,507 @@
...
+
+func (c *genericTransportController) updatePlacementDecision(ctx context.Context, placementDecision *v1alpha1.PlacementDecision, updateObjectFunc updateObjectFunc) error {
+	// objects returned from a PlacementDecisionLister must be treated as read-only.
+	// Therefore, create a deep copy before updating the object.
+	newPlacementDecision := placementDecision.DeepCopy()
```
MikeSpreitzer, line 435: If you wanted to save some memory in some cases, you could move this responsibility into the updateFunc.
Not a blocker.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: I guess my meaning was not clear. You made the change I was thinking of, using `errors.IsNotFound`.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: It had better be allocating some new memory, otherwise two calls with different namespaces (which could happen concurrently) will interfere with each other!
Not a blocker.

cmd/transport/options/options.go
MikeSpreitzer: It's a little odd to put something under `cmd/` that is not a command. I myself would have put it in `pkg/`.
Not a blocker.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: I have opened Issue #1651 to represent the need to fix this later.

MikeSpreitzer (commented):
I think this could be merged as is, or a little more work done.
/lgtm
Thanks!

nirrozenbaum:
@MikeSpreitzer
in order to unblock Maroon, let's get it merged and I'll address the last comments in a new PR.
thanks!

pkg/transport/generic_transport_controller.go
nirrozenbaum: you're right. it does allocate some new memory.
my mistake.
will take care in later PR

pkg/transport/generic_transport_controller.go
nirrozenbaum: handled in PR #1675

PR #1593, opened 2024-01-18
:bug: Fix Syncer to avoid adding entry to nil map

yana1205:
## Summary

Syncer to avoid adding an entry to nil map
  - There is a case that an object is annotated as `downsync-overwrite` in MBS but the object on WEC has empty annotation. In this case, Syncer crashed due to adding an entry to nil map at [downsyncer.go#L303](https://github.com/kubestellar/kubestellar/blob/release-0.14/pkg/syncer/syncers/downsyncer.go#L303)

We were reported that Syncer was crashed by MCAD integration teams and found that the above is the root cause.  

## Related issue(s)

Fixes #


MikeSpreitzer (commented):
/lgtm
/approve
Thanks!

PR #1597, opened 2024-01-19
✨ Start v0.20 doc re-org

MikeSpreitzer:
Create top-level structure, add architecture and packaging stubs.

Start re-org of examples.

## Summary
This PR is a start on some doc work. Adding some basic docs about architecture and packaging+delivery. Some discussed re-org of the examples. Some removal of duplicate material.

## Related issue(s)

Fixes #


docs/content/v0.20/architecture.md
```diff
@@ -0,0 +1,36 @@
+# KubeStellar Architecture
+
+**NOTE**: This is only a very rudimentary start.
+
+## Higher Level
+
+At the higher level there are three modules.
+
+1. KubeFlex
```
ezrasilvera, line 9: In a KubeStellar project I would assume KubeStellar should be the main and first module and the others should be some components it is using.

docs/content/v0.20/architecture.md
```diff
@@ -0,0 +1,36 @@
...
+2. [the status addon](https://github.ibm.com/dettori/status-addon)
```
ezrasilvera, line 10: the is implementation details, unless you are going to list here all the controllers, I would say the status addon is part of KubeStellar

docs/content/v0.20/examples.md
```diff
@@ -0,0 +1,588 @@
+# KubeStellar Usage Examples
+
```
ezrasilvera, line 2: Maybe put X specific files one for each scenario and put here a one liner title + a link ?

ezrasilvera:
I think we need a more detailed pre-req file (if it will come as next PR it's fine ..) , it should at least include expected versions (KIND, ko, etc..) and a pointer to the site (if we don't supply instructions)

docs/content/v0.20/examples.md
```diff
@@ -0,0 +1,588 @@
...
+```
+
+Verify that statefulset has been created in both clusters
+
+```shell
+kubectl --context cluster1 get statefulsets -n postgres-system
+kubectl --context cluster2 get statefulsets -n postgres-system
+```
+
+### Propagate helm metadata Secret to managed clusters
```
MikeSpreitzer, line 339: I misunderstood earlier. It seems to me now that this subsection _is_ optional. We are talking about only something that makes `helm list` able to work on the WEC(s), _not_ about something that makes the workload functional.

docs/content/v0.20/architecture.md
MikeSpreitzer: The point here is that the status addon is NOT in the KubeStellar repo, so at a high level --- where we look at repos, go modules, and dependencies --- we have to say that the status addon is currently separate. I think that it _should_ move into the KubeStellar repo, but right now we have to document the current state of affairs.

MikeSpreitzer:
Yes, more work on pre-reqs can be done, I suggest in follow-on PRs.

docs/content/v0.20/examples.md
```diff
@@ -0,0 +1,588 @@
...
+Verify that deployment has been created in both clusters
+
+```shell
+kubectl --context cluster1 get deployments -n nginx-res
+kubectl --context cluster2 get deployments -n nginx-res
+```
+
+## Scenario 6 - multi-cluster workload deployment of app with ServiceAccount with ArgoCD 
+
+This scenario is something you can do after the [common setup](#common-setup).
```
waltforme, line 532: Looks like when Scenario 3 is deleted, the instructions for Argo CD installation, i.e. the reference to thirdparties.md, which was at the beginning of the deleleted Scenario 3, also gets deleted. We need to get that reference back, right?

MikeSpreitzer: Good catch. Restored.

docs/content/v0.20/architecture.md
MikeSpreitzer: Re-ordered in latest edition.

docs/content/v0.20/examples.md
MikeSpreitzer: Restored to language about "optional" in latest edition.

MikeSpreitzer:
One of the changes I have made here is to the singleton status return scenario. The problem with that was that it was exercising semantics that we have not agreed on, namely having our controller making a _choice_ of which WEC to distribute to when more than one matches. That is not the semantics of the legacy code. Also, the scenario was making an _assumption_ of which WEC the controller would chose. I have changed the scenario so that there is only one matching WEC, to avoid these problems.

docs/content/v0.20/packaging.md
```diff
@@ -0,0 +1,28 @@
+# Packaging and Delivery
+
+## status-addon
+
+The [status-addon](https://github.ibm.com/dettori/status-addon) repo is the source of a RedHat-style operator. 
+
+The operator is delivered by a Helm chart at `quay.io/pdettori/status-addon-chart`. Chart versioning? How is the chart built, pushed?
+
+There are probably one or more container images involved.
+
```
pdettori, line 10: There is one. Plan is to open source the status add on as well and make the image part of our official set of images.

docs/content/v0.20/packaging.md
```diff
@@ -0,0 +1,28 @@
...
+## KubeStellar
+
+### Central container image
+
+Appears to be at `ghcr.io/kubestellar/kubestellar/kubestellar-operator:0.20.0-alpha.1`. `make ko-build-push` will create that image. `make ko-build-local` will make a local image for just the local platform.
```
pdettori, line 15: Maybe we can say "current snapshot available at... " - "appears to be" seems to indicate something we are guessing about, but we do know

docs/content/v0.20/packaging.md
```diff
@@ -0,0 +1,28 @@
...
+
+### KubeStellar Operator Helm Chart
+
+The source for a Helm chart is in [core-helm-chart](../../../core-helm-chart). `make chart` (re)derives it from loal sources.
```
pdettori, line 19: loal->local

docs/content/v0.20/packaging.md
```diff
@@ -0,0 +1,28 @@
...
+
+This chart creates (among other things) a `Deployment` object that runs a container from the image `ghcr.io/kubestellar/kubestellar/kubestellar-operator:0.20.0-alpha.1`.
+
+### KubeFlex PostCreateHooks
+
+There are two `PostCreateHook` objects defined in [config/postcreate-hooks](../../../config/postcreate-hooks).
+
+- `ocm.yaml` adds `clusteradm` by running a container using the image `quay.io/kubestellar/clusteradm:0.7.1` which is built from ... what?
```
pdettori, line 27: It's built from https://github.com/open-cluster-management-io/clusteradm - since it's a CLI there are no community images available for it so we are building our own. @francostellari submitted https://github.com/kubestellar/kubestellar/pull/1602 to add the build script for this.

docs/content/v0.20/packaging.md
```diff
@@ -0,0 +1,28 @@
...
+- `kubestellar.yaml` runs container image `quay.io/kubestellar/helm:v3.13.2` (which is built by what? From what?) to instantiate the chart from `oci://ghcr.io/kubestellar/kubestellar/kubestellar-operator-chart` (tag?), which is built by (what?) from (what? probably answered above).
```
pdettori, line 28: Similar answer to above. No community image is available for helm so we build our own from https://github.com/helm/helm using ko. We should also ad a script for that - perhaps @francostellari can create one based on the one he ddid for clusteradm.

docs/content/v0.20/architecture.md
pdettori: We should definitely move it under the org https://github.com/kubestellar . As of moving into https://github.com/kubestellar/kubestellar - tried previously and found go deps issues could not resolve. But the main reason to keep it separate is modularity - we wanted to separate OCM - specific code in the transport-related repo(s).

pdettori (changes requested):
Overall I am fine with the changes, I just pointed out couple of updates.

docs/content/v0.20/architecture.md
MikeSpreitzer: OK, so the future move is not entirely clear yet.

docs/content/v0.20/packaging.md
MikeSpreitzer: thanks, noted.

docs/content/v0.20/packaging.md
MikeSpreitzer: thanks, noted.

docs/content/v0.20/packaging.md
MikeSpreitzer: Revised.

docs/content/v0.20/packaging.md
MikeSpreitzer: Updated, thanks.

docs/content/v0.20/architecture.md
pdettori: Well, I propose to move the private repo to a new repo  https://github.com/kubestellar/status-addon - we just have to agree on that.

pdettori (approved):
/lgtm

docs/content/v0.20/packaging.md
francostellari: @pdettori the current version that you have is also single arch. I can create script and build the multi arch. Do you specifically want v3.13.2 or latest is good?

ezrasilvera:
Can we please make sure that issues are opened for  all the "will be done in the future" comments? Thanks

docs/content/v0.20/packaging.md
francostellari: Please see https://github.com/kubestellar/kubestellar/pull/1604/files

MikeSpreitzer:
The force-push to 8f3476370 is a rebase, done to trigger re-test --- which passed.

MikeSpreitzer:
Created #1605 for a "will be done".

Issue #1584, opened 2024-01-18
bug:  `kflex create imbs1 --type vcluster -p ocm` not working


francostellari:
### Describe the bug

Referring to step 3 of https://github.com/kubestellar/kubestellar/tree/main/docs/content/v0.20:

After the command: 

```shell
$ kflex create imbs1 --type vcluster -p ocm
✔ Checking for saved initial context...
✔ Switching to initial context...
✔ Creating new control plane imbs1 of type vcluster ...
✔ Waiting for API server to become ready...
```

The OCM pods do not seem to be working:

```shell
$ k --context kind-kubeflex get pods -A
NAMESPACE            NAME                                                READY   STATUS      RESTARTS   AGE
imbs1-system         coredns-68559449b6-swvrd-x-kube-system-x-vcluster   1/1     Running     0          46s
imbs1-system         ocm-cq5xm                                           0/1     Error       0          45s
imbs1-system         ocm-fbt5q                                           0/1     Error       0          27s
imbs1-system         update-cluster-info-5kqtf                           0/1     Completed   0          74s
imbs1-system         vcluster-0                                          2/2     Running     0          74s
```

The reported error is:

```shell
$ k --context kind-kubeflex logs ocm-cq5xm -n imbs1-system
exec /ko-app/clusteradm: exec format error
```

### Steps To Reproduce

See above

### Expected Behavior

Pods run

pdettori:
@francostellari what is your arch/os ? perhaps we do not have the build for that ?

francostellari:
Linux amd64

francostellari:
I could not figure out how to see the tags and arch of the ghrc.io registry. Pasting portions of the url does not work

pdettori:
The image is creating the issue is the job defined in https://github.com/kubestellar/kubestellar/blob/main/config/postcreate-hooks/ocm.yaml - the image is quay.io/kubestellar/clusteradm:0.7.1

pdettori:
Not sure I did the build for linux/amd64 - can you tell by looking at https://quay.io/repository/kubestellar/clusteradm?tab=tags ?

francostellari:
> Not sure I did the build for linux/amd64 - can you tell by looking at https://quay.io/repository/kubestellar/clusteradm?tab=tags ?

@pdettori It looks like that only linux/arm64 is available.
We need to build for other arches, at least linux/amd64

pdettori:
ok, will look into that

[...]

PR #1602, opened 2024-01-19
🐛 Update clusteradm image 0.7.1-->0.7.2 and provide multi-arch

francostellari:
## Summary

- Fix issue #1584
- Update clusteradm version 0.7.1-->0.7.2
- Provide convenience script to build future images of clusteradm
  Example of usage: `build-clusteradm-image.sh --version 0.7.2 --registry quay.io/kubestellar`

## Related issue(s)

Fixes #
#1584


MikeSpreitzer:
Thanks.

Development-time scripts go in `hack/`.

We should have a doc about the commands that contributors use for the various development-time activities that derive some things from other things. We do not have such a doc yet, so I can not fault this PR for updating that doc to mention this derivation step.

francostellari:
> Development-time scripts go in `hack/`.

I moved as you requested

hack/build-clusteradm-image.sh
```diff
@@ -0,0 +1,99 @@
...
+
+# Usage: $0 [--version|-v version] [--repo registry] [--platform platforms] [--verbose|-V] [-X]
+# Working directory does not matter.
+
+set -e
+
+clusteradm_version="" # ==> latest
+verbose="false"
+clusteradm_folder=clusteradm
+repo=quay.io/kubestellar
```
MikeSpreitzer, line 27: In container image jargon, this is a "registry" rather than a "repo" (as correctly stated in the command line flags).

hack/build-clusteradm-image.sh
```diff
@@ -0,0 +1,99 @@
...
+platform=linux/amd64,linux/arm64,linux/ppc64le
```
MikeSpreitzer, line 28: This is a plural rather than a singular.

francostellari:
Yes I was undecided, I ended up picking KO nomenclature that calls it a
REPO. I can change it if it is important. Being a build tool probably
rarely used... Probably nobody is going to care

hack/build-clusteradm-image.sh
```diff
@@ -0,0 +1,99 @@
...
+    shift
+done
+
+if [ "$clusteradm_version" == "" ]; then
+    clusteradm_version=$(get_latest_version)
+fi
+
+# Remove the initial "v" if present
+if [[ "$clusteradm_version" == v* ]]; then
+    clusteradm_version=${clusteradm_version//v}
```
MikeSpreitzer, line 75: This removes all appearances of "v". For example, it converts `v0.20.0-preview` to `0.20.0-preiew`. I think that instead you want to remove just a prefix, as follows.
```
clusteradm_version=${clusteradm_version#v}
```

francostellari:
Again both docker and ko use the singular for this flag

francostellari:
Ok

hack/build-clusteradm-image.sh
```diff
@@ -0,0 +1,99 @@
...
+        then { repo="$2"; shift; }
+        else { echo "$0: missing registry url" >&2; exit 1; }
+        fi;;
+    (--platform)
+        if (( $# > 1 ));
+        then { platform="$2"; shift; }
+        else { echo "$0: missing comma separated list of platforms" >&2; exit 1; }
+        fi;;
+    (--verbose|-V)
+        verbose="true";;
```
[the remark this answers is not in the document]
francostellari: ok

MikeSpreitzer:
On "repo" vs. "registry": I think the definitions in OCI and Docker are more dominant than what `ko` does. As long as the command lines that contributors issue use "registry", that is the main thing; the internal variable is less important.

hack/build-clusteradm-image.sh
```diff
@@ -0,0 +1,99 @@
...
+}
+
+while (( $# > 0 )); do
+    case "$1" in
+    (--version|-v)
+        if (( $# > 1 ));
+        then { clusteradm_version="$2"; shift; }
+        else { echo "$0: missing release version" >&2; exit 1; }
+        fi;;
+    (--repo)
```
MikeSpreitzer, line 42: Should be "--registry".

francostellari: ok

hack/build-clusteradm-image.sh
```diff
@@ -0,0 +1,99 @@
...
+else
+    git clone -b "v$clusteradm_version" --depth 1 https://github.com/open-cluster-management-io/clusteradm.git "$clusteradm_folder" > /dev/null
+fi
+
+cd "$clusteradm_folder"
+
+export KO_DOCKER_REPO=$repo
+
+if [ $verbose == "true" ]; then
+    ko build -B ./cmd/clusteradm -t $clusteradm_version --sbom=none --platform $platform
```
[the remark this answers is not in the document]
francostellari: because it create garbage in quay.io

hack/build-clusteradm-image.sh
```diff
@@ -0,0 +1,85 @@
...
+        fi;;
+    (--platform)
+        if (( $# > 1 ));
+        then { platform="$2"; shift; }
+        else { echo "$0: missing comma separated list of platforms" >&2; exit 1; }
+        fi;;
+    (-X)
+    	set -x;;
+    (-h|--help)
+        echo "Usage: $0 [--version|-v release] [--registry registry] [-V|--verbose] [-X]"
```
MikeSpreitzer, line 53: remove `--verbose` here too.

francostellari: ok

hack/build-clusteradm-image.sh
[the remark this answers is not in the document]
MikeSpreitzer: "garbage"? Are you referring to the SBOMs?

MikeSpreitzer:
This is good enough to merge, although I would prefer to reach an agreement on SBOMs. It is OK if we pursue that in follow-on work.
/lgtm

hack/build-clusteradm-image.sh
```diff
@@ -0,0 +1,85 @@
...
+    esac
+    shift
+done
+
+if [ "$clusteradm_version" == "" ]; then
+    clusteradm_version=$(get_latest_version)
+fi
+
+# Remove the initial "v", if present
+if [[ "$clusteradm_version" == v* ]]; then
```
MikeSpreitzer, line 70: BTW, the parameter expansion below makes this test unnecessary --- `#` strips a prefix IFF it is there.

MikeSpreitzer (commented):
/lgtm
/approve
Thanks!

PR #1603, opened 2024-01-19
🌱 Re-organize the e2e tests for de-duplication, doc consistency

MikeSpreitzer:
## Summary
This PR re-organizes the e2e test sources, to reduce duplication and track the scenario change in #1597 . This PR improves alignment between the user-facing docs in docs/content/v0.20 and the end-to-end tests. This also fixes a source of test flakiness.

## Related issue(s)

Fixes #1603


test/e2e/common/setup-kubestellar.sh
```diff
@@ -0,0 +1,95 @@
...
+:
+function create_cluster() {
+  cluster=$1
+  kind create cluster --name $cluster
+  kubectl config rename-context kind-${cluster} $cluster
+  clusteradm --context imbs1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk '{print $0 " --context '${cluster}' --force-internal-endpoint-lookup"}' | sh
+}
+
+create_cluster cluster1
+create_cluster cluster2
```
ezrasilvera, line 68: This is limiting to 2 clusters even when we need 1 or more to 2, can we add a parameter so that setup-kubestellar create N clusters?  (named cluster{n})

ezrasilvera:
I would rename the "workload-deployment" to "execute-test" /"execute" / "run-test"  or something (as it's not just deploying workloads), and change the `run-test` to something like "run-e2e-test" or something

test/e2e/common/setup-kubestellar.sh
MikeSpreitzer: How many would the cleanup delete? Right now it is very simple: we have two tests that both create and initially delete two WECs.

Can we address generalization in follow-on PRs?

MikeSpreitzer:
@ezrasilvera : I agree, "workload-deployment" was poorly named.  I did not want to rename it to something that sounds like it is the whole test. Instead I renamed it to "use-kubestellar", hopefully making clear the fact that it is complementary to "setup-kubestellar" (and thus the other component files in a test).

test/e2e/common/setup-kubestellar.sh
ezrasilvera: Yes we can. I thought the singleton uses only a single cluster.

PR #1604, opened 2024-01-19
✨ Provide build script for Helm image

francostellari:
## Summary

- provide script for building the Helm container image
- created multi arch image `3.14.0`
- updated the reference in the yaml

## Related issue(s)

Fixes #

### DIFF
--- a/hack/build-helm-image.sh
+++ b/hack/build-helm-image.sh
@@ -1,0 +1,83 @@
+#!/bin/env bash
+
+# Copyright 2023 The KubeStellar Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Purpose: build a helm image and push it to the registry.
+
+# Usage: $0 [--version|-v release] [--registry registry] [--platform platforms] [-X]
+# Working directory does not matter.
+
+set -e
+
+helm_version="" # ==> latest
+helm_folder=helm
+registry=quay.io/kubestellar
+platform=linux/amd64,linux/arm64,linux/ppc64le
+
+get_latest_version() {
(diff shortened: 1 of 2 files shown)


hack/build-helm-image.sh
```diff
@@ -0,0 +1,83 @@
...
+
+# Usage: $0 [--version|-v release] [--registry registry] [--platform platforms] [-X]
+# Working directory does not matter.
+
+set -e
+
+helm_version="" # ==> latest
+helm_folder=helm
+registry=quay.io/kubestellar
+platform=linux/amd64,linux/arm64,linux/ppc64le
```
ezrasilvera, line 27: Do we have OCM for linux/ppc64le  ?

francostellari: This is not for OCM but yes, I have build for linux/amd64, linux/arm64, and linux/ppc64le as you can see in quay.io.

hack/build-helm-image.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/env bash
+
+# Copyright 2023 The KubeStellar Authors.
+#
```
pdettori, line 4: Since these are files authored this year, let's use 2024 instead of 2023

hack/build-helm-image.sh
pdettori: FYI - open cluster management only has  images for linux/amd64 and linux/arm64 - see for example https://quay.io/repository/open-cluster-management/work?tab=tags (on the agent side) and https://quay.io/repository/open-cluster-management/placement?tab=tags (on the hub side). So even if we build for other archs and we use OCM that will not help.

pdettori: @francostellari just to confirm,  If I don't pass any flags it will just build with the defaults in lines 24-27, right ?

pdettori (changes requested):
Overall looks good. I suggest that until we sort out out to support other archs we should just build for the two supported by OCM, and so update the platform defaults, and the header as suggested.

francostellari:
https://github.com/kubestellar/kubestellar/pull/1604#discussion_r1461230681

Yes

hack/build-helm-image.sh
francostellari: ok, done

francostellari:
I think that the underlying issue OCM image availability is not quite related to the helm image. In my opinion, it is better to have the image we build with multi arches in case we find a solution in the future, rather than coming back and adding more arches.
What is preventing OCM from build the ppc64le image? Can we aske them to add it to future releases?

pdettori (approved):
/lgtm

PR #1606, opened 2024-01-20
🌱 Improved ordering of go imports

MikeSpreitzer:
## Summary
This PR updates the rule for ordering go-language import, putting kubeflex in place of kcp.

This PR is built on #1603 (because I have been plagued by the test flakiness) and will be rebased after that merges.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/crd/crds.go
+++ b/pkg/crd/crds.go
@@ -25,7 +25,6 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	kfutil "github.com/kubestellar/kubeflex/pkg/util"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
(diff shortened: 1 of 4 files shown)


pkg/util/kubeconfig.go
```diff
@@ -33,6 +32,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	kfv1aplha1 "github.com/kubestellar/kubeflex/api/v1alpha1"
```
ezrasilvera, line 35: Why do we need KF related code in KS?  I don't even thing we need GetWDSKubeconfig and GetIMBSKubeconfig at all. controllers that need to access those cp/spaces can get the actual kubeconfig directly instead of getting a name and then retrieve a config. 
The KS code can and should be agnostic to who supply that config

pkg/crd/crds.go
```diff
@@ -36,6 +35,8 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 
+	kfutil "github.com/kubestellar/kubeflex/pkg/util"
```
ezrasilvera, line 38: Also here - the same comment, as the kubeconfig.go Why do I need KF util functions in order to apply CRDs on a CP ?

MikeSpreitzer:
@ezrasilvera : the point you raised is part of the thinking in #1633  and why I hope it can be accomplished sooner rather than later. I hope you agree that it should not be part of this PR.

MikeSpreitzer:
The force-push to 06e558f8e is a rebase onto the latest edition of #1603

MikeSpreitzer:
The force-push to 9970c0683 is the promised rebase onto `main`.

PR #1607, opened 2024-01-20
✨ Cleanup gomod

MikeSpreitzer:
## Summary
This PR does some cleanup of go.mod: removing unneeded replacements, making the required versions of k8s stuff consistent.

## Related issue(s)

Fixes #

PR #1608, opened 2024-01-20
✨ Add basic client generation

### LINKED ISSUE #1537: feature: have client-go code generation

MikeSpreitzer:
## Summary
This PR adds generation of client code, including informers and listers, using the plain Kubernetes code-generator module (k8s.io/code-generator). The generated code appears in `pkg/generated`.

In order to get this working, I had to insert a directory in the api branch. The v1alpha1 API is now at `api/edge/v1alpha1` (it used to be at `api/v1alpha1`).

I also had to rename the file `placement_types.go` to `types.go` (https://github.com/kubernetes/kubernetes/issues/119930).

Bug #1611 predates this PR and this PR does not fix that bug. I do not know whether there is any (relevant) incompatibility between code-generator and controller-gen.

## Related issue(s)

Fixes #1537


hack/update-codegen-clients.sh
```diff
@@ -19,29 +19,20 @@ set -o nounset
...
-# "core:v2alpha1 workload:v2alpha1 apiresource:v2alpha1 tenancy:v2alpha1 tenancy:v1beta1 apis:v2alpha1 scheduling:v2alpha1 topology:v2alpha1" \
-bash "${CODEGEN_PKG}"/generate-groups.sh "deepcopy,client" \
-  github.com/kubestellar/kubestellar/pkg/client github.com/kubestellar/kubestellar/pkg/apis \
-  "edge:v2alpha1" \
-  --go-header-file "${SCRIPT_ROOT}"/hack/boilerplate/boilerplate.generatego.txt \
-  --output-base "${SCRIPT_ROOT}" \
-  --trim-path-prefix github.com/kubestellar/kubestellar
+source "${CODE_GEN_DIR}/kube_codegen.sh"
+
+rm -rf "${SCRIPT_ROOT}/pkg/generated"
```
ezrasilvera, line 26: Why do we need to delete the whole directory every time ?  BTW, If so this is a "create" not "update" function.

ezrasilvera:
@MikeSpreitzer So changing the package from Edge will be in a different PR ?

hack/update-codegen-clients.sh
nirrozenbaum: @ezrasilvera this is needed in case you remove an existing API. 
in such a case you want to also remove the generated code for that API and not to leave garbage.

MikeSpreitzer:
The force-push to dc2947c04 is a rebase onto the current edition of #1607 .

@ezrasilvera : yes, the API group rename will be a separate PR.

MikeSpreitzer:
waiting for #1607 to merge, so
/hold

MikeSpreitzer:
The force-push to 0ff436967 is a rebase onto the current edition of #1607 .

MikeSpreitzer:
The force-push to 3e640fd1e is the rebase onto `main`.
/remove-hold

api/edge/register.go
```diff
@@ -0,0 +1,22 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package edge
+
```
ezrasilvera, line 18: Why Edge ?

ezrasilvera: Sorry - just saw the reply

Issue #1611, opened 2024-01-21
bug: `make manifests` and `make generate` fail


MikeSpreitzer:
### Describe the bug

```
mspreitz@mjs13 kubestellar % git status
On branch main
Your branch is up to date with 'upstream/main'.

nothing to commit, working tree clean

mspreitz@mjs13 kubestellar % git rev-parse HEAD
12b4d02d6844ee55dc339b78f9a8d89147ddddd4

mspreitz@mjs13 kubestellar % make manifests
test -s /Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin/controller-gen && /Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin/controller-gen --version | grep -q v0.11.3 || \
	GOBIN=/Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.11.3
/Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
HyphenGroup/apis/example/v1/register.go:20:2: k8s.io/apimachinery@v0.0.0: replacement directory ../../apimachinery does not exist
HyphenGroup/apis/example/v1/register.go:21:2: k8s.io/apimachinery@v0.0.0: replacement directory ../../apimachinery does not exist
HyphenGroup/apis/example/v1/register.go:22:2: k8s.io/apimachinery@v0.0.0: replacement directory ../../apimachinery does not exist
HyphenGroup/applyconfiguration/example/v1/clustertesttype.go:23:2: k8s.io/apimachinery@v0.0.0: replacement directory ../../apimachinery does not exist
HyphenGroup/applyconfiguration/example/v1/clustertesttype.go:24:2: k8s.io/client-go@v0.0.0: replacement directory ../../client-go does not exist
HyphenGroup/clientset/versioned/scheme/register.go:25:2: k8s.io/apimachinery@v0.0.0: replacement directory ../../apimachinery does not exist
HyphenGroup/clientset/versioned/scheme/register.go:26:2: k8s.io/apimachinery@v0.0.0: replacement directory ../../apimachinery does not exist
HyphenGroup/clientset/versioned/typed/example/v1/clustertesttype.go:27:2: k8s.io/api@v0.0.0: replacement directory ../../api does not exist
[...]

PR #1623, opened 2024-01-22
✨ Deep copy by controller-gen

MikeSpreitzer:
## Summary
This PR changes the responsibility for generating the deep copy code back to controller-gen, so that we do not have two generators generating the same subject matter (note: they do it slightly differently). This undoes most (but not all) of #1615, now that #1611 has been resolved.

## Related issue(s)

Fixes #


MikeSpreitzer:
@nirrozenbaum

PR #1627, opened 2024-01-22
🌱 Update go version to track Kubernetes release 1.28

MikeSpreitzer:
## Summary
This PR updates the `go.mod` to require at least go 1.20, to be consistent with Kuberentes release 1.28 (which is what is imported).

## Related issue(s)

Fixes #

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -5,7 +5,7 @@ presubmits:
     clone_uri: "https://github.com/kubestellar/kubestellar"
     spec:
       containers:
-        - image: ghcr.io/kcp-dev/infra/build:1.19.9-2
+        - image: ghcr.io/kcp-dev/infra/build:1.20.9-3
           command:
             - make
             - verify-boilerplate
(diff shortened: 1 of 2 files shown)


waltforme:
I saw the pull-kubestellar-verify check is complaining. So do we want to update the `image: ghcr.io/kcp-dev/infra/build:1.19.9-2` in `.prow.yaml`?

MikeSpreitzer:
@waltforme : thanks, yes. Wondering where to find the list of what is available... besides using the utility I wrote to get a full listing of the container repo metadata.

MikeSpreitzer:
Oh and the funky login is required there.
Oh, and maybe we should use an upstream edition.

MikeSpreitzer:
https://github.com/kubestellar/kubestellar/blob/main-legacy/hack/inspect-image.py

fortunately, the list is short:
```
{"name":"kcp-dev/infra/build","tags":["1.19.9-1","1.19.9-2","1.19.9-3","1.20.7-1","1.20.8-1","1.20.9-1","1.20.9-2","1.20.9-3"]}
```

MikeSpreitzer:
The force-push to c30e09d5a modifies the pull-kubestellar-verify prow job to use a go-1.20-based image.

waltforme:
> BTW, we are going to have trouble when we want go 1.21 (which we will when we go to Kubernetes release 1.29), because there is no corresponding ghcr.io/kcp-dev/infra/build image.

Maybe we can switch to some other image at that time, seems the container is used to do basic/general stuff only.

PR #1630, opened 2024-01-22
🌱 Tidy up code generation

MikeSpreitzer:
## Summary
This PR rubs off a few warts in the code generation.
1. Removes the unneeded and ineffectual invocation of code-generator to generate deep copy code.
2. Adds a `make` target to do all the code generation.
3. Updates documentation.

## Related issue(s)

Fixes #


MikeSpreitzer:
The force-push to 678793405 is a rebase onto `main`.

MikeSpreitzer:
For some reason unclear to me, `make generate` now outputs very slightly different --- but entirely equivalent --- code. The second commit here memorializes that change.

docs/content/v0.20/packaging.md
```diff
@@ -12,19 +20,24 @@ There is also a container image involved.
...
+## Files generated by controller-gen
 
-- `make generated` generates the deep copy code.
+- `make manifests` generates the CustomeResourceDefinition files,
+  which exist in two places: [config/crd/bases](config/crd/bases) and
+  [pkg/crd/files](pkg/crd/files).
 
-Both of the above are currently broken, see Issue 1611.
+- `make generated` generates the deep copy code, which exists in
+  `zz_generated.deepcopy.go` next to the API source.
```
nirrozenbaum, line 36: what about `make codegenclients` and `make all-generated`?

MikeSpreitzer: I am not sure I understand the question. The text already says that `make all-generated` is equivalent to the combination of the three other make targets for generating files. The text already says that `make codegenclients` makes the files in `pkg/generated/`, not files in `api/`.

nirrozenbaum: oops. I missed it in the doc. saw the make all-generated only in makefile.
then all good.

PR #1636, opened 2024-01-23
✨ Restore verification of derived files

MikeSpreitzer:
## Summary
This PR restores the `make` target `verify-codegen`, which was doing nothing. Now it checks that all the derived files have been correctly derived.

## Related issue(s)

Fixes #


hack/verify-codegen.sh
```diff
@@ -21,15 +21,24 @@ set -o errexit
...
-if ! git diff --quiet --exit-code -- pkg/client; then
+if [[ "$(git status --porcelain=1 -- api config/crd/bases pkg/crd/files pkg/generated)" != "" ]]; then
+    echo "Sorry, some relevant files are not in the current git commit."
+    echo "Correct checking can not be done."
+    git status -- api config/crd/bases pkg/crd/files pkg/generated
+    exit 1
+fi
+
+pwd
+make all-generated
```
ezrasilvera, line 32: This seems a little circular. This is supposed to be just verify not verify-and-build-if-missing

MikeSpreitzer: This follows the established pattern from upstream.
I added the preflight check you see earlier in this script.
Having `make` do the derivations would make it impossible for this script to do the preflight check.

Issue #1585, opened 2024-01-18
Enable the legacy workflows for release  0.14  branch


ezrasilvera:
Need to re-enable the workflows so people can validate fixes for release 0.14

MikeSpreitzer:
I just did a survey of which workflows are **disabled** in the GitHub web UI. I found the following disabled.

- create KubeStellar-Syncer image
- docs-ecutable - ALL
- docs-ecutable - example1
- docs-ecutable - Mailbox
- docs-ecutable - Placement
- docs-ecutable - QS
- docs-ecutable - Scheduler
- docs-ecutable - Syncer
- docs-ecutable - User Quickstart Kind
-  docs-ecutable - User Quickstart Test
- docs-ecutable - Where Resolver
- Greetings

MikeSpreitzer:
Is there any record of who disabled a given workflow, when, why?

MikeSpreitzer:
The "create KubeStellar-Syncer image" workflow only reacts to workflow_dispatch events, so I am re-enabling it right now. Not that it matters, GitHub says it has never been run.

MikeSpreitzer:
I have re-enabled the example1 and user-quickstart-test docs-ecutables in the GitHub web UI, now that their definitions in `main` exclude PR activity on `main`.

MikeSpreitzer:
Not sure what to do about the Greetings workflow.

MikeSpreitzer:
That missing example file was restored in #1661

MikeSpreitzer:
I have re-enabled the remaining docs-ecutable workflows in the GitHub web UI, now that they only trigger for legacy branches and tags.

The only remaining question is what to do about the "Greetings" workflow.

MikeSpreitzer:
I have re-enabled the Greetings workflow in the GitHub web UI.

PR #1652, opened 2024-01-25
🌱 Restrict CI workflows to legacy KubeStellar, for main

MikeSpreitzer:
## Summary
This PR adjusts when the example1 and user-quickstart-test workflows trigger, to restrict that to legacy KubeStellar. These workflows are currently disabled in the GitHub web UI and we would like to re-enable them for maintenance of legacy KubeStellar.

I do not understand GitHub workflows well enough to know whether I need to make this change to `main` or `release-0.14` or both. Unless and until I learn otherwise, I will attempt both.

This is part of issue #1585 

## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/docs-ecutable-example1.yml
+++ b/.github/workflows/docs-ecutable-example1.yml
@@ -8,16 +8,14 @@ on:
   # To confirm any changes to docs build successfully, without deploying them
   pull_request:
     branches:
-      - main
-      - "release-*"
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
/cc @clubanderson 
/cc @ezrasilvera 
/cc @yana1205 
@dumb0002

yana1205 (approved):
/lgtm

PR #1653, opened 2024-01-25
🌱 Restrict CI workflows to legacy KubeStellar, for release-0.14

MikeSpreitzer:
## Summary
This PR adjusts when the example1 and user-quickstart-test workflows trigger, to restrict that to legacy KubeStellar. These workflows are currently disabled in the GitHub web UI and we would like to re-enable them for maintenance of legacy KubeStellar.

I do not understand GitHub workflows well enough to know whether I need to make this change to `main` or `release-0.14` or both. Unless and until I learn otherwise, I will attempt both. The complement to this PR is #1652 .

This is part of issue #1585 

## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/docs-ecutable-example1.yml
+++ b/.github/workflows/docs-ecutable-example1.yml
@@ -8,14 +8,12 @@ on:
   # To confirm any changes to docs build successfully, without deploying them
   pull_request:
     branches:
-      - main
-      - "release-*"
+      - "release-0.1*"
   push:
     branches:
-      - main
-      - "release-*"
+      - "release-0.1*"
     tags:
-      - 'v*'
+      - 'v0.1*'
 
 env:
   docs-ecutable-filename: example1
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
/cc @clubanderson 
/cc @ezrasilvera 
/cc @yana1205 
@dumb0002

clubanderson:
I would only change main for this

yana1205:
Probably, the change is ok with only main.

> An event occurs on your repository. The event has an associated commit SHA and Git ref.

> GitHub searches the .github/workflows directory in your repository for workflow files that are present in the associated commit SHA or Git ref of the event.

ezrasilvera:
@MikeSpreitzer Is it indeed necessary to enable all workflows for the legacy release ? I imagine that for validation tests 1 or 2 will be enough. It's adding a support and maintenance complexity to have so many workflows that are not part of our main anymore.
Can we move all to legacy sub directory ? Or prefix all with legacy- ?

ezrasilvera:
> @MikeSpreitzer Is it indeed necessary to enable all workflows for the legacy release ? I imagine that for validation tests 1 or 2 will be enough. It's adding a support and maintenance complexity to have so many workflows that are not part of our main anymore. Can we move all to legacy sub directory ? Or prefix all with legacy- ?

PR #1654, opened 2024-01-26
✨ Add script to check pre-requisites

francostellari:
## Summary

Provide a utility script to check several pre-requisites. 
This may be useful in support of other scripts and for user debugging.

Command line:

```shell
$ ./check_pre_req.sh --help
Usage: ./check_pre_req.sh [-A|--assert] [-L|--list] [-V|--verbose] [-X] [program1] [program2] ...
```

Listing all programs that can be checked:

```shell
$ ./check_pre_req.sh --list
argo brew docker go helm jq kind ko make ocm yq
```

Check several Go related dependencies:

```shell
$ ./check_pre_req.sh go ko
Checking KubeStellar pre-requisites:
✅ Go
✅ KO
```

Check verbosely, showing resolution suggestions:

```shell
$ ./check_pre_req.sh -V argo
Checking KubeStellar pre-requisites:
❌ ArgoCD CLI
   install: https://argo-cd.readthedocs.io/en/stable/cli_installation/
```

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/v0.20/contributor.md
+++ b/docs/content/v0.20/contributor.md
@@ -4,7 +4,7 @@
 
 ## Check pre-requisites for KubeStellar
 
-The [check_pre_req](../../../hack/check_pre_req.sh) script offers a convenient way to check for pre-requisites, including the ones needed for [KubeStellar](./pre-reqs.md) deployment and [use case scenarios](./examples.md).
+The [check_pre_req](../../../hack/check_pre_req.sh) script offers a convenient way to check for pre-requisites, needed for [KubeStellar](./pre-reqs.md) deployment and [use case scenarios](./examples.md).
 
 The [check_pre_req](../../../hack/check_pre_req.sh) script check for a pre-requisite presence in the path, by using the `which` command, and it can optionally provide version and path information for pre-requisites that are present, or installation information for missing pre-requisites.
 
@@ -25,27 +25,26 @@ The [check_pre_req](../../../hack/check_pre_req.sh) script accepts a list of spe
 
 ```shell
 $ check_pre_req.sh --list
-argo brew docker go helm jq kcp kflex kind ko kubectl make ocm yq
+argo brew docker go helm jq kflex kind ko kubectl make ocm yq
 ```
 
-For example, the complete list of pre-requisites could be checked with the command below:
+For example, list of pre-requisites required by Kubestellar can be checked with the command below (add the `-V` flag to get the version of each program na dusggestions on how to install missing pre-requisites):
 
 ```shell
-$ check_pre_req.sh
-X ArgoCD CLI
-✔ Home Brew
+$ hack/check_pre_req.sh
+Checking pre-requisites for using KubeStellar:
 ✔ Docker
-✔ Go
-✔ Helm
-✔ jq
-X kcp
-✔ KubeFlex
-✔ Kind
-✔ KO
 ✔ kubectl
-✔ GNU Make
+✔ KubeFlex
 ✔ OCM CLI
-✔ yq
+✔ Helm
+Checking additional pre-requisites for running the examples:
+✔ Kind
+X ArgoCD CLI
+Checking pre-requisites for building KubeStellar:
+✔ GNU Make
+✔ Go
+✔ KO
 ```
 
 In another example, a specific list of pre-requisites could be asserted by an higher-level script, while providing some installation information, with the command below (note that the script will terminate upon finding a missing pre-requisite):
--- a/docs/content/v0.20/contributor.md
+++ b/docs/content/v0.20/contributor.md
@@ -1,0 +1,61 @@
+# KubeStellar Contributors
+
+**NOTE**: work in progress... write more
+
+## Check pre-requisites for KubeStellar
+
+The [check_pre_req](../../../hack/check_pre_req.sh) script offers a convenient way to check for pre-requisites, including the ones needed for [KubeStellar](./pre-reqs.md) deployment and [use case scenarios](./examples.md).
+
+The [check_pre_req](../../../hack/check_pre_req.sh) script check for a pre-requisite presence in the path, by using the `which` command, and it can optionally provide version and path information for pre-requisites that are present, or installation information for missing pre-requisites.
+
+We envision that the [check_pre_req](../../../hack/check_pre_req.sh) script could be useful for user-side debugging as well as for asserting the presence of pre-requisites in higher-level automation scripts.
+
+The [check_pre_req](../../../hack/check_pre_req.sh) script accepts a list of optional flags and arguments.
+
+**Supported flags:**
+
+- `-A|--assert`: exits with error code 2 upon finding the fist missing pre-requisite
+- `-L|--list`: prints a list of supported pre-requisites
+- `-V|--verbose`: displays version and path information for installed pre-requisites or installation information for missing pre-requisites
+- `-X`: enable `set -x` for debugging the script
+
+**Supported arguments:**
+
+The [check_pre_req](../../../hack/check_pre_req.sh) script accepts a list of specific pre-requisites to check, among the list of available ones:
+
+```shell
+$ check_pre_req.sh --list
+argo brew docker go helm jq kcp kflex kind ko kubectl make ocm yq
+```
+
+For example, the complete list of pre-requisites could be checked with the command below:
+
+```shell
+$ check_pre_req.sh
+X ArgoCD CLI
+✔ Home Brew
+✔ Docker
+✔ Go
+✔ Helm
+✔ jq
+X kcp
+✔ KubeFlex
+✔ Kind
+✔ KO
+✔ kubectl
+✔ GNU Make
+✔ OCM CLI
+✔ yq
+```
+
+In another example, a specific list of pre-requisites could be asserted by an higher-level script, while providing some installation information, with the command below (note that the script will terminate upon finding a missing pre-requisite):
+
+```shell
+$ check_pre_req.sh --assert --verbose helm argo docker kind
+Checking KubeStellar pre-requisites:
+✔ Helm
+  version: version.BuildInfo{Version:"v3.14.0", GitCommit:"3fc9f4b2638e76f26739cd77c7017139be81d0ea", GitTreeState:"clean", GoVersion:"go1.21.5"}
+     path: /usr/sbin/helm
+X ArgoCD CLI
+  how to install: https://argo-cd.readthedocs.io/en/stable/cli_installation/
+```
--- a/docs/content/v0.20/examples.md
+++ b/docs/content/v0.20/examples.md
@@ -7,8 +7,6 @@
 
 See [pre-reqs](pre-reqs.md).
 
-See also [check_pre_req](contributor.md#check-key-pre-requisites-for-kubestellar) script.
-
 ## Common Setup
 
 The following steps establish an initial state used in the examples below.
--- a/docs/content/v0.20/examples.md
+++ b/docs/content/v0.20/examples.md
@@ -57,7 +57,6 @@ manager which connects to the `wds1` front-end and the `imbs1` OCM control plane
 
 6. Follow the steps to [create and register two clusters with OCM](example-wecs.md).
 
-
 7. (optional) Check all deployments and statefulsets running in the hosting cluster. Expect to
 see the wds1 kubestellar-controller-manager created in the wds1-system namespace and the imbs1
 statefulset created in the imbs1-system namespace.
@@ -554,6 +553,7 @@ spec:
     - matchLabels: {"argocd.argoproj.io/instance":"nginx-sa"}
 EOF
 ```
+
 Switch context to hosting cluster and argocd namespace (this is required by argo to
 create an app with the CLI)
 
--- a/docs/content/v0.20/examples.md
+++ b/docs/content/v0.20/examples.md
@@ -7,6 +7,8 @@
 
 See [pre-reqs](pre-reqs.md).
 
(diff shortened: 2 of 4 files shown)


hack/check_pre_req.sh
```diff
@@ -0,0 +1,180 @@
...
+set -e # exit on error
+
+echoerr() {
+    echo "$@" 1>&2;
+}
+
+is_installed() {
+    # $1 == name
+    # $2 == command name to search
+    # $3 == command to get the evrsion
```
MikeSpreitzer, line 26: s/evrsion/version/

hack/check_pre_req.sh
```diff
@@ -0,0 +1,180 @@
...
+    # $4 == help
+    if which $2 > /dev/null ; then
+        echo -e "\u2705 $1"
+        if [ $# -gt 2 ]; then
+            echov "   version: $(eval "$3" 2> /dev/null)"
```
MikeSpreitzer, line 31: On my mac:

```
bash-3.2$ echov bac
bash: echov: command not found
```

MikeSpreitzer: Wouldn't plain `echo` suffice here?

hack/check_pre_req.sh
```diff
@@ -0,0 +1,180 @@
...
+        fi
+        echov "      path: $(which $2)"
+    else
+        echo -e "\u274c $1"
+        if [ $# -gt 3 ]; then
+            echov "   install: $4"
```
MikeSpreitzer, line 37: Maybe make it a little more explicit: say "how to install" rather than just "install"?

hack/check_pre_req.sh
```diff
@@ -0,0 +1,180 @@
...
+        'brew' \
+        'brew --version' \
+        '/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"'
+}
+
+is_installed_docker() {
+    is_installed 'Docker' \
+        'docker' \
+        'docker --version' \
+        'curl -L https://raw.githubusercontent.com/open-cluster-management-io/clusteradm/main/install.sh | bash'
```
MikeSpreitzer, line 63: This is not the install for _docker_

MikeSpreitzer:
```
mspreitz@mjs13 kubestellar % hack/check_pre_req.sh jq
zsh: permission denied: hack/check_pre_req.sh
```

I think GitHub will remember it if you `chmod +x hack/check_pre_req.sh`

hack/check_pre_req.sh
```diff
@@ -0,0 +1,180 @@
+#!/bin/env bash
```
MikeSpreitzer, line 1:
```
mspreitz@mjs13 kubestellar % chmod +x hack/check_pre_req.sh 
mspreitz@mjs13 kubestellar % 
mspreitz@mjs13 kubestellar % hack/check_pre_req.sh jq      
zsh: hack/check_pre_req.sh: bad interpreter: /bin/env: no such file or directory
```

OTOH, `/usr/bin/env` seems to be working in other scripts.

MikeSpreitzer:
After fixing the problems I found, it ran like this on my Mac:

```
mspreitz@mjs13 kubestellar % hack/check_pre_req.sh         
\u274c ArgoCD CLI
\u2705 Home Brew
\u2705 Docker
\u2705 Go
\u2705 Helm
\u2705 jq
\u2705 Kind
\u2705 KO
\u2705 GNU Make
\u2705 OCM CLI
\u2705 yq
```

MikeSpreitzer:
The following is what `man bash` has to say about the builtin command "echo" on my Mac.

```
       echo [-neE] [arg ...]
              Output the args, separated by spaces, followed by a newline.  The return status is always 0.  If -n is specified, the trailing newline is suppressed.  If
              the -e option is given, interpretation of the following backslash-escaped characters is enabled.  The -E option disables the interpretation of these escape
              characters, even on systems where they are interpreted by default.  The xpg_echo shell option may be used to dynamically determine whether or not echo
              expands these escape characters by default.  echo does not interpret -- to mean the end of options.  echo interprets the following escape sequences:
              \a     alert (bell)
              \b     backspace
              \c     suppress trailing newline
              \e     an escape character
              \f     form feed
              \n     new line
              \r     carriage return
              \t     horizontal tab
              \v     vertical tab
              \\     backslash
              \0nnn  the eight-bit character whose value is the octal value nnn (zero to three octal digits)
              \xHH   the eight-bit character whose value is the hexadecimal value HH (one or two hex digits)
```

OTOH, there is this:

```
mspreitz@mjs13 kubestellar % echo $'\u2705 jq'
✅ jq
```

hack/check_pre_req.sh
```diff
@@ -0,0 +1,180 @@
...
+        verbose="true";;
+    (-X)
+    	set -x;;
+    (--help|-help|-h)
+        echo "Usage: $0 [-A|--assert] [-L|--list] [-V|--verbose] [-X] [program1] [program2] ..."
+        exit 0;;
+    ($PROGRAMS)
+        programs+=("$1");;
+    (-*)
+        echoerr "$0: unknown flag"
```
MikeSpreitzer, line 149: Tell the reader what that flag is.

hack/check_pre_req.sh
```diff
@@ -0,0 +1,180 @@
...
+        exit 1;;
+    (*)
+        echoerr "$0: unknown positional argument"
```
MikeSpreitzer, line 152: Tell the reader what that argument is.

hack/check_pre_req.sh
francostellari: done, thank you

hack/check_pre_req.sh
francostellari: ok, thank you

hack/check_pre_req.sh
francostellari: done, thank you

hack/check_pre_req.sh
francostellari: done, thank you

francostellari:
Added kcp and kflex

MikeSpreitzer:
```
mspreitz@mjs13 kubestellar % hack/check_pre_req.sh 
zsh: permission denied: hack/check_pre_req.sh
```

MikeSpreitzer:
Oh, ugh, the `$'` that I showed earlier was from zsh not bash. On the Mac, `$'` is ALSO limited to one-byte code points.

MikeSpreitzer:
So, after `chmod +x`, I get this:
```
mspreitz@mjs13 kubestellar % hack/check_pre_req.sh argo jq
\u274c ArgoCD CLI
\u2705 jq
```

MikeSpreitzer:
So, sadly, I suggest replacing the checkmark and X with "installed:" and "missing: ".

MikeSpreitzer:
Success on my mac!

Screenshot 2024-01-26 at 1 43 15 PM

MikeSpreitzer (commented):
It would be good to have some user-facing doc telling users and contributors about this new script.

/lgtm
Thanks!

docs/content/v0.20/contributor.md
```diff
@@ -0,0 +1,60 @@
...
+```shell
+$ check_pre_req.sh --list
+argo brew docker go helm jq kcp kflex kind ko make ocm yq
+```
+
+For example, the complete list of pre-requisites could be checked with the command below:
+
+```shell
+$ check_pre_req.sh
+X ArgoCD CLI
```
ezrasilvera, line 35: I wouldn't put ArgoCD as pre-req

docs/content/v0.20/contributor.md
```diff
@@ -0,0 +1,60 @@
...
+✔ Home Brew
+✔ Docker
+✔ Go
+✔ Helm
+✔ jq
+X kcp
```
ezrasilvera, line 41: I wouldn't put KCP as pre-req for the default (i.e. no parameters) of this script. Obviously one can still explicitly check for it using this scrip

docs/content/v0.20/contributor.md
MikeSpreitzer: Our user-facing pre-req.md distinguishes multiple categories of prereqs. I see that this script makes no distinctions. I assumed that was deliberate. We need a doc that points users at this script and explains how to use it; presumably that doc will make it clear that this script checks all categories of prereqs.

docs/content/v0.20/contributor.md
```diff
@@ -0,0 +1,60 @@
+# KubeStellar Contributors
+
+**NOTE**: work in progress... write more
+
+## Check key pre-requisites for KubeStellar
+
+The [check_pre_req](../../../hack/check_pre_req.sh) script offers a convenient way to check key pre-requisites needed for KubeStellar deployment and use case scenarios.
```
MikeSpreitzer, line 7: "key" is too vague. Point to the existing pre-reqs.md and explain that the check_pre_req.sh script checks all categories in that doc and more.

docs/content/v0.20/contributor.md
ezrasilvera: My point is that the **default** behavior of this script (i.e., when no parameters are given) should be the default pre-req for our base user scenario, which doesn't include argo and KCP. 
Argo and KCP should/can be checked when specifying special category.

francostellari:
@ezrasilvera @MikeSpreitzer I reworked the docs a bit more.

francostellari:
@ezrasilvera the script is designed to check a bunch of pre-req that may be useful now or potentially in the future. As difference scenarios have different requirements, the default behavior is to check them all. The default behavior is non blocking.
In the future, when the scenarios are more settled, it would be a good idea to have those as parameters: e.g. check_pre_req sceario1|example2|etc

ezrasilvera:
> @ezrasilvera the script is designed to check a bunch of pre-req that may be useful now or potentially in the future. As difference scenarios have different requirements, the default behavior is to check them all. The default behavior is non blocking. In the future, when the scenarios are more settled, it would be a good idea to have those as parameters: e.g. check_pre_req sceario1|example2|etc

I think this may confuse contributors. This script is pointed through the contributor MD file where we expect 99.999% of the users to not use either Argo and surely not KCP (which may not be even supported in 0.20 ) . If a user tries KS for the first time, and see a red X near argo he will conclude he needs Argo and follow your link to install it. In the output you show there isn't any message that indicates this is an optional dependency ..

docs/content/v0.20/contributor.md
```diff
@@ -0,0 +1,61 @@
+# KubeStellar Contributors
+
+**NOTE**: work in progress... write more
+
+## Check pre-requisites for KubeStellar
+
+The [check_pre_req](../../../hack/check_pre_req.sh) script offers a convenient way to check for pre-requisites, including the ones needed for [KubeStellar](./pre-reqs.md) deployment and [use case scenarios](./examples.md).
```
MikeSpreitzer, line 7: The existing pre-reqs.md file covers prerequisites for more than "deployment". Have a look at the categories there, and the way the reference to ArgoCD is qualified.

docs/content/v0.20/examples.md
```diff
@@ -7,6 +7,8 @@
 
 See [pre-reqs](pre-reqs.md).
 
+See also [check_pre_req](contributor.md#check-key-pre-requisites-for-kubestellar) script.
+
```
MikeSpreitzer, line 11: Probably this reference should be moved into pre-reqs.md.

docs/content/v0.20/examples.md
```diff
@@ -25,7 +27,7 @@ The following steps establish an initial state used in the examples below.
    ```
 
 3. Create an inventory & mailbox space of type `vcluster` running *OCM* (Open Cluster Management)
-in KubeFlex. Note that `-p ocm` runs a post-create hook on the *vcluster* control plane 
+in KubeFlex. Note that `-p ocm` runs a post-create hook on the *vcluster* control plane
```
MikeSpreitzer, line 30: Thanks for finding and fixing all these extra spaces at end of line.
If these fixes were in a separate PR then it could be merge without any debate.

MikeSpreitzer:
I am sympathetic to the argument that the checking script should not be checking for kcp now because it is not needed at all. We can add checking for kcp in the future when it becomes relevant.

MikeSpreitzer:
I am sympathetic to the idea of making some distinctions among categories. I am sympathetic to wanting a simple script. Rather than introducing the complexity of a way on the command line to limit the categories checked, a simpler approach would be to simple organize the output by category, with an explicit indication of category (e.g., group the output by category, with an introductory line per category).

francostellari:
I have pushed a new commit to try addressing several comments:
1. I removed kcp the list of checked programs
2. I have reworked the default output as:
```shell
$ hack/check_pre_req.sh
Checking pre-requisites for using KubeStellar:
✔ Docker
✔ kubectl
✔ KubeFlex
✔ OCM CLI
✔ Helm
Checking additional pre-requisites for running the examples:
✔ Kind
X ArgoCD CLI
Checking pre-requisites for building KubeStellar:
✔ GNU Make
✔ Go
✔ KO
```

docs/content/v0.20/examples.md
```diff
@@ -25,7 +25,7 @@ The following steps establish an initial state used in the examples below.
    ```
 
 3. Create an inventory & mailbox space of type `vcluster` running *OCM* (Open Cluster Management)
-in KubeFlex. Note that `-p ocm` runs a post-create hook on the *vcluster* control plane
```
MikeSpreitzer, line 28: I do not understand why GitHub is saying that there is a trailing space to be deleted here. This PR was rebased onto `main` after that space was deleted, right?

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #1658, opened 2024-01-26
🐛 minor fix in helper functions

nirrozenbaum:
## Summary

1. isCRD function is using always version "v1". this is incorrect (a bug) since CRDs might have different versions. 
for example - KS CRDs are using v1alpha1 version.

2. making the getObjectGVK function simpler

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/util/unstructured.go
+++ b/pkg/util/unstructured.go
@@ -88,8 +88,6 @@ func getObjectGVK(o interface{}) (schema.GroupVersionKind, error) {
 	switch obj := o.(type) {
 	case runtime.Object:
 		return obj.GetObjectKind().GroupVersionKind(), nil
-	case unstructured.Unstructured:
-		return obj.GetObjectKind().GroupVersionKind(), nil
 	}
 
 	return schema.GroupVersionKind{}, fmt.Errorf("object is of wrong type: %#v", o)
--- a/pkg/util/unstructured.go
+++ b/pkg/util/unstructured.go
@@ -32,7 +32,7 @@ import (
 const (
 	CRDKind                              = "CustomResourceDefinition"
 	CRDGroup                             = "apiextensions.k8s.io"
-	CRDVersion                           = "v1"
+	AnyVersion                           = "*"
 	ServiceVersion                       = "v1"
 	ServiceKind                          = "Service"
 	PlacementKind                        = "Placement"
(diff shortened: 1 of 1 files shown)


pkg/util/unstructured.go
```diff
@@ -77,28 +77,20 @@ func matchesGVK(o interface{}, group, version, kind string) bool {
...
-		gvk.Kind = obj.GetObjectKind().GroupVersionKind().Kind
-	case unstructured.Unstructured:
-		gvk.Group = obj.GetObjectKind().GroupVersionKind().Group
-		gvk.Version = obj.GetObjectKind().GroupVersionKind().Version
-		gvk.Kind = obj.GetObjectKind().GroupVersionKind().Kind
-	default:
-		return gvk, fmt.Errorf("object is of wrong type: %#v", obj)
-	}
-	return gvk, nil
+		return obj.GetObjectKind().GroupVersionKind(), nil
```
pdettori, line 90: Removing the case for `unstructured.Unstructured` may have unintended consequences on other parts of the code using this function. I had scenarios where the object could only be cast to `unstructured.Unstructured` but not to `runtime.Object`.

nirrozenbaum: how is that possible?
`Unstructured` implements `runtime.Object` interface.
so if it can be cast to `unstructured` is should be possible to cast also to `runtime.Object` interface.

pdettori: You are correct. I remember having that issue once but now I cannot reproduce. Maybe it was about creating directly an object as `unstructured.Unstructured` which did not have the deep copy functions and therefore the cast threw an error saying it was missing those. But I did a quick test with this change and I don't see that happing, so I will go ahed and approve.

pdettori (approved):
/lgtm

pkg/util/unstructured.go
MikeSpreitzer: `func (in *Unstructured) DeepCopy() *Unstructured` exists.

MikeSpreitzer: So does `func (obj *Unstructured) GetObjectKind() schema.ObjectKind`.  Those are the two methods that make up `runtime.Object`.

MikeSpreitzer: The following code compiles.

```
	test := &unstructured.Unstructured{}
	var testR runtime.Object = test
```

MikeSpreitzer (commented):
/lgtm

PR #1659, opened 2024-01-26
🌱 Don't expose spec.numberOfClusters of placements

### LINKED ISSUE #1641: bug: spec.numberOfClusters is not supported and should be removed
### Describe the bug

See issue #1594 - since there is no agreement yet on supporting it we decided to remove it before the 0.20 release.

### Steps To Reproduce

Run the scenario in #1594

### Expected Behavior

We should not have a parameter that is not supported at this time

### Additional Context

_No response_

waltforme:
## Summary
Resolves #1641.


pdettori (approved):
/lgtm

pdettori:
Thanks @waltforme !

waltforme:
YW @pdettori !
/approve

PR #1660, opened 2024-01-26
Fix Syncer not to overwrite managedFields

yana1205:
## Summary

This is the fix for Syncer not to overwrite managedFields. This bug was discovered while performance test. Thanks @dumb0002.

(This is same change as [the first commit](https://github.com/kubestellar/kubestellar/pull/1578/commits/e48f8ea765bd796958f91cc7127a6c9e1bb95fa6) in this PR (https://github.com/kubestellar/kubestellar/pull/1578))
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/syncer/syncers/common.go
+++ b/pkg/syncer/syncers/common.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"strings"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/klog/v2"
@@ -182,7 +183,7 @@ func diff(logger klog.Logger, srcResourceList *unstructured.UnstructuredList, de
 		if ok {
 			srcResource.SetResourceVersion(destResource.GetResourceVersion())
 			srcResource.SetUID(destResource.GetUID())
-			srcResource.SetManagedFields(destResource.GetManagedFields())
+			srcResource.SetManagedFields([]metav1.ManagedFieldsEntry{})
 
 			// Avoid to overwrite status field. Though, not sure this workaround is required.
 			// Actually, when Syncer donwsynces, Syncer doesn't call UpdateStatus() method. Status fields at downstream side aren't updated by downsyncing.
@@ -197,6 +198,7 @@ func diff(logger klog.Logger, srcResourceList *unstructured.UnstructuredList, de
 		} else {
 			srcResource.SetResourceVersion("")
 			srcResource.SetUID("")
+			srcResource.SetManagedFields([]metav1.ManagedFieldsEntry{})
 			setAnnotation(&srcResource)
 			newResources = append(newResources, srcResource)
 		}
--- a/pkg/syncer/syncers/downsyncer.go
+++ b/pkg/syncer/syncers/downsyncer.go
@@ -22,6 +22,7 @@ import (
 	"sync"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/klog/v2"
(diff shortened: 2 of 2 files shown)


MikeSpreitzer:
Has this been tested?

pkg/syncer/syncers/common.go
```diff
@@ -182,7 +183,7 @@ func diff(logger klog.Logger, srcResourceList *unstructured.UnstructuredList, de
 		if ok {
 			srcResource.SetResourceVersion(destResource.GetResourceVersion())
 			srcResource.SetUID(destResource.GetUID())
-			srcResource.SetManagedFields(destResource.GetManagedFields())
+			srcResource.SetManagedFields([]metav1.ManagedFieldsEntry{})
```
MikeSpreitzer, line 186: If the given value were `nil` instead of an empty slice then this would hit the even more satisfying special case at the start of SetManagedFields:

```
func (u *Unstructured) SetManagedFields(managedFields []metav1.ManagedFieldsEntry) {
	if managedFields == nil {
		RemoveNestedField(u.Object, "metadata", "managedFields")
		return
	}
...
```

MikeSpreitzer: But an empty slice is equivalent on the wire.  See https://github.com/kubernetes/apimachinery/blob/v0.24.3/pkg/apis/meta/v1/types.go#L275

MikeSpreitzer:
The CI test "docs-ecutable - User Quickstart Test" failed because of https://github.com/kubestellar/kubestellar/issues/1585#issuecomment-1912458152 , which I hope to fix with #1661

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

pkg/syncer/syncers/common.go
yana1205: Thanks. Using `nil` might be better. I tried it at local and confirmed that managedFields are also not overwritten by Syncer.

yana1205:
@MikeSpreitzer Thank you for your review and approval!

> Has this been tested?

I forgot to add it in PR message. Yes, I tested it when debugging it in the environment of the performance test. I confirmed that the managedFields in downstream objects are not updated with this fix (setting empty ManagedFields) at that time.

pkg/syncer/syncers/common.go
MikeSpreitzer: I think it would only be better in the sense of making its intended effect more obvious. But, as I noted above, because of the way that objects are serialized and deserialized --- both to/from JSON and protobuf --- `nil` is equivalent to empty slice.

MikeSpreitzer:
@dumb0002 also reports successful testing. He used, I believe, the multi-platform manifest with digest  b309c83c634f6743af618860b17b0c8fcd61fcd3ef5f0563155057d46cdeb7cf (referencing image digests 7659eea87de0e2607ec22b7957486ae6369053800bd9bc7a480a3f83f02ebea0 for arm64, d36933bb0105c982ec77f7e433b28327ab9c27b69b6806eee2b658c7b003c256 for amd64), which I built from commit fa4c1b4ee and tagged `git-fa4c1b4ee-clean` and `v0.14.2`. Because of the successful tests, I am now moving the `release-0.14` image tag to that same multi-platform manifest.

MikeSpreitzer:
Using `go version -m $binary_file`, I was able to make a little more forensic progress. In the `/ko-app/syncer` in 93f1c416a250bc60d5ccc20540d4011ab03304c262c5f6d58389d56c3ef107d8, I found only these `build` entries:

```
	build	-compiler=gc
	build	-trimpath=true
	build	CGO_ENABLED=0
	build	GOARCH=arm64
	build	GOOS=linux
```

Looking in 7659eea87de0e2607ec22b7957486ae6369053800bd9bc7a480a3f83f02ebea0 I found the following.

```
	build	-compiler=gc
	build	-ldflags="-s -w -X main.version=0.14.2 -X main.commit=fa4c1b4eec34eb1c0f154e38148d2c0da4cc4903 -X main.date=2024-01-26T19:42:28Z -X main.builtBy=goreleaser"
	build	CGO_ENABLED=0
	build	GOARCH=arm64
	build	GOOS=linux
	build	vcs=git
	build	vcs.revision=fa4c1b4eec34eb1c0f154e38148d2c0da4cc4903
	build	vcs.time=2024-01-26T19:06:21Z
	build	vcs.modified=true
```

So whoever built b309c83c634f6743af618860b17b0c8fcd61fcd3ef5f0563155057d46cdeb7cf did _not_ use `make build-kubestellar-syncer-image` (which says `GOFLAGS=-buildvcs=false`) but instead used goreleaser, and they did it at 2:42 PM EST on Friday.

clubanderson:
there was that GH action for goreleaser that someone was playing with last week. Perhaps that did this?

PR #1662, opened 2024-01-26
🐛 Fix Syncer not to overwrite managedFields

MikeSpreitzer:
## Summary
This is the fix for Syncer not to overwrite managedFields. This bug was discovered while performance test. Thanks @dumb0002.

(This is same change as [the first commit](https://github.com/kubestellar/kubestellar/pull/1578/commits/e48f8ea765bd796958f91cc7127a6c9e1bb95fa6) in this PR (https://github.com/kubestellar/kubestellar/pull/1578))

This is @yana1205's work and is exactly the same as #1660 but that one is blocked by a bogus CI test failure and there is no way for me to re-run that git workflow.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/syncer/syncers/common.go
+++ b/pkg/syncer/syncers/common.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"strings"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
Now I can proceed with #1660.

PR #1664, opened 2024-01-26
🌱 Restore other docs-ecutable workflows for legacy releases, for release-0.14

MikeSpreitzer:
## Summary
This PR narrows what the remaining docs-ecutable workflows say they apply to, restricting them to legacy releases.

This addresses part of #1585 

## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/docs-ecutable-mailbox.yml
+++ b/.github/workflows/docs-ecutable-mailbox.yml
@@ -9,8 +9,7 @@ on:
   pull_request:
     types: [labeled]
     branches:
-      - main
-      - "release-*"
+      - "release-0.1*"
     paths:
       - "docs/content/Coding Milestones/PoC2023q1/mailbox-controller.md"
       - "docs/content/Coding Milestones/PoC2023q1/mailbox-controller-subs/**"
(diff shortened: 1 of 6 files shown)


ezrasilvera:
@MikeSpreitzer Is it indeed necessary to enable all workflows for the legacy release ? I imagine that for validation tests 1 or 2 will be enough. It's adding a support and maintenance complexity to have so many workflows that are not part of our main anymore.
Can we move all to legacy sub directory ? Or prefix all with legacy- ?

ezrasilvera:
> @MikeSpreitzer Is it indeed necessary to enable all workflows for the legacy release ? I imagine that for validation tests 1 or 2 will be enough. It's adding a support and maintenance complexity to have so many workflows that are not part of our main anymore. Can we move all to legacy sub directory ? Or prefix all with legacy- ?

ezrasilvera:
@MikeSpreitzer do we still need this ?   If not please close.

MikeSpreitzer:
I am not sure whether we will need this. It may depend on related things we do or do not do for non-legacy KubeStellar.

PR #1665, opened 2024-01-26
🌱 Restore other docs-ecutable workflows for legacy releases, for main

MikeSpreitzer:
## Summary
This PR narrows the applicability of the remaining docs-ecutable workflows to legacy releases, in the workflow definitions in the `main` branch.

This is part of addressing #1585 

## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/docs-ecutable-mailbox.yml
+++ b/.github/workflows/docs-ecutable-mailbox.yml
@@ -9,8 +9,7 @@ on:
   pull_request:
     types: [labeled]
     branches:
-      - main
-      - "release-*"
+      - "release-0.1*"
     paths:
       - "docs/content/Coding Milestones/PoC2023q1/mailbox-controller.md"
       - "docs/content/Coding Milestones/PoC2023q1/mailbox-controller-subs/**"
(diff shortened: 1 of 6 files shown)


ezrasilvera:
@MikeSpreitzer Is it indeed necessary to enable all workflows for the legacy release ?  I imagine that for validation tests 1 or 2 will be enough. It's adding a support and maintenance complexity to have so many workflows that are not part of our main anymore. 
Can we move all to `legacy` sub directory ?  Or prefix all with `legacy-`  ?

ezrasilvera:
> @MikeSpreitzer Is it indeed necessary to enable all workflows for the legacy release ? I imagine that for validation tests 1 or 2 will be enough. It's adding a support and maintenance complexity to have so many workflows that are not part of our main anymore. Can we move all to `legacy` sub directory ? Or prefix all with `legacy-` ?

MikeSpreitzer:
@ezrasilvera : we seem to have some interest in maintaining the release-0.14 branch, so that is why I am making the workflows that test a release usable.

MikeSpreitzer:
I suspect that renaming them in `main` will also require renaming them in `release-0.14` because for _some_ --- and not _all_ --- triggering events, what matters is what is in the relevant PR or branch rather than `main`. @clubanderson can you clarify?

PR #1667, opened 2024-01-27
🐛 added `WorkloadGeneration` field to the downsync objects section of PlacementDecision

### LINKED ISSUE #1671: feature: handle update to an object that was selected by a placement
### Feature Description

Suppose you create a deployment in the WDS. The placement controller writes the placement decision, the transport controller sees there is a new (or updated) placement decision, it wraps it up and deliver it to the mailbox namespace. Perhaps there are also other objects in that placement decision that are targeted by the placement. 
At this point, assume that the deployment is updated - e.g., changing the number of replicas from 1 to 3, or added an annotation/label. How is the transport informed of the change in the deployment in spec/annotations/labels? 

### Proposed Solution

There are several options to tackle this issue:
Option (1) - the placementdecision might have a `revision` field for every object that exists in the `ClusterScope` and `NamespaceScope` lists. this way, when placement controller gets an event, if something has changed in the specific object the placement controller increments the revision field of the changed object. This triggers an event in transport controller that will wrap up the objects and take if from there to the WECs.
(suggested by Nir & Maroon)

Option (2) - same as option (1), but instead of having a `revision` field per object, we have a single `revision` field per the two lists (ClusterScope + NamespaceScope). if any object in those lists has changed, placement controller increments the revision field which in turn triggers an event in transport controller, continues the same as in option (1).
(suggested by Nir&Maroon)

Option (3) - the placement controller is only responsible for maintaining correct lists in the PlacementDecision, and the transport controller is responsible for maintaining the correct relationships between workload objects and wrapped-up contents in the mailboxes.
(suggested by Mike)

Option (4) - implement the go interface defined for transport based on the work API and for every pluggable transport implement the code in a separate go project / package and import that code/package in KubeStellar controller manager. Assuming we have other transport plugins, the selection of one or the other would be based on some flag on the KS controller manager. We may think this approach may make more difficult to add other plugins, but after all, how many do we expect. Most likely we will have the OCM-based one and the KS syncer-based one.
(suggested by Paolo)

Option (5) - Expose the go transport interface as a gRPC interface. The transport would run in a separate container in the same pod as the placement controller or in a separate pod and exposed as a service. Whenever the placement controller sees an update on the tracked objects, it runs all the placement matching logic for objects and clusters and invokes the gRPC transport service which takes care of the transport part.
(suggested by Paolo)

 

### Want to contribute?

### Additional Context

_No response_

nirrozenbaum:
This field is required in order to trigger reconciliation in cases where the list of objects to downsync hasn't changed (their identifiers hasn't changed), but spec of at least one of the objects has changed. in such cases we need to trigger distribution of the updates changes.

## Related issue(s)

Fixes #1671

### DIFF
--- a/api/edge/v1alpha1/types.go
+++ b/api/edge/v1alpha1/types.go
@@ -205,9 +205,9 @@ type PlacementDecisionSpec struct {
 }
 
 // DownsyncObjectReferences explicitly defines the objects to be down-synced.
-// The ClusterScope list defines the cluster-scope objects, NamespacedObjects packs individual objects
-// identifiable by namespace & name, and generation represents the internal spec generation of the listed
-// objects in the ClusterScope and NamespaceScope lists.
+// The ClusterScope list defines the cluster-scope objects, NamespaceScope list
+// defines the namespace-scope objects and generation represents the internal spec
+// generation of the listed objects in the ClusterScope and NamespaceScope lists.
 type DownsyncObjectReferences struct {
 	// `clusterScope` holds a list of individual cluster-scoped objects
 	// to downsync, organized by resource.
--- a/api/edge/v1alpha1/types.go
+++ b/api/edge/v1alpha1/types.go
@@ -205,8 +205,9 @@ type PlacementDecisionSpec struct {
 }
 
 // DownsyncObjectReferences explicitly defines the objects to be down-synced.
-// The ClusterScope list defines the cluster-scope objects, while NamespacedObjects packs individual objects
-// identifiable by namespace & name.
+// The ClusterScope list defines the cluster-scope objects, NamespacedObjects packs individual objects
+// identifiable by namespace & name, and generation represents the internal spec generation of the listed
+// objects in the ClusterScope and NamespaceScope lists.
 type DownsyncObjectReferences struct {
 	// `clusterScope` holds a list of individual cluster-scoped objects
 	// to downsync, organized by resource.
@@ -218,6 +219,15 @@ type DownsyncObjectReferences struct {
 	// `NamespaceScope` matches if and only if at least one member matches.
 	// +optional
 	NamespaceScope []NamespaceScopeDownsyncObjects `json:"namespaceScope,omitempty"`
+
+	// `ObjectsInternalSpecGeneration` is a sequence number representing a specific generation of the downsync objects internal spec.
+	// For example, if ClusterScope and NamespaceScope lists haven't changed but the
+	// internal spec of at least one object has changed, this field should be incremented.
+	// ObjectsInternalSpecGeneration should be updated upon a change in the spec of one
+	// of the objects from ClusterScope and NamespaceScope in order to trigger the
+	// distribution of the updated object.
+	// +optional
+	ObjectsInternalSpecGeneration int64 `json:"generation,omitempty"`
 }
 
 // NamespaceScopeDownsyncObjects matches some objects of one particular namespaced object.
--- a/api/edge/v1alpha1/types.go
+++ b/api/edge/v1alpha1/types.go
@@ -206,8 +206,10 @@ type PlacementDecisionSpec struct {
 
 // DownsyncObjectReferences explicitly defines the objects to be down-synced.
 // The ClusterScope list defines the cluster-scope objects, NamespaceScope list
-// defines the namespace-scope objects and generation represents the internal spec
-// generation of the listed objects in the ClusterScope and NamespaceScope lists.
+// defines the namespace-scope objects and WorkloadGeneration represents the
+// generation of the objects in the ClusterScope and NamespaceScope lists.
+// Upon a change in any of workload objects that should be distributed
+// (e.g., spec, annotations or labels) the workload generation field should be incremented.
 type DownsyncObjectReferences struct {
 	// `clusterScope` holds a list of individual cluster-scoped objects
 	// to downsync, organized by resource.
@@ -220,14 +222,15 @@ type DownsyncObjectReferences struct {
 	// +optional
 	NamespaceScope []NamespaceScopeDownsyncObjects `json:"namespaceScope,omitempty"`
 
-	// `ObjectsInternalSpecGeneration` is a sequence number representing a specific generation of the downsync objects internal spec.
-	// For example, if ClusterScope and NamespaceScope lists haven't changed but the
-	// internal spec of at least one object has changed, this field should be incremented.
-	// ObjectsInternalSpecGeneration should be updated upon a change in the spec of one
-	// of the objects from ClusterScope and NamespaceScope in order to trigger the
-	// distribution of the updated object.
+	// `WorkloadGeneration` is a sequence number representing a specific generation of
+	// the workload objects to be downsynced.
+	// For example, if ClusterScope and NamespaceScope lists haven't changed but at least
+	// one object has changed, this field should be incremented.
+	// Upon a change in any of workload objects that should be distributed
+	// (e.g., spec, annotations or labels) the workload generation field should be incremented.
+	// `WorkloadGeneration` field is monotonically increasing.
 	// +optional
-	ObjectsInternalSpecGeneration int64 `json:"generation,omitempty"`
+	WorkloadGeneration int64 `json:"workloadGeneration,omitempty"`
 }
 
 // NamespaceScopeDownsyncObjects matches some objects of one particular namespaced object.
--- a/config/crd/bases/edge.kubestellar.io_placementdecisions.yaml
+++ b/config/crd/bases/edge.kubestellar.io_placementdecisions.yaml
@@ -81,6 +81,17 @@ spec:
                       - version
                       type: object
                     type: array
+                  generation:
+                    description: '`ObjectsInternalSpecGeneration` is a sequence number
+                      representing a specific generation of the downsync objects internal
+                      spec. For example, if ClusterScope and NamespaceScope lists
+                      haven''t changed but the internal spec of at least one object
+                      has changed, this field should be incremented. ObjectsInternalSpecGeneration
+                      should be updated upon a change in the spec of one of the objects
+                      from ClusterScope and NamespaceScope in order to trigger the
+                      distribution of the updated object.'
+                    format: int64
+                    type: integer
                   namespaceScope:
                     description: '`NamespaceScope` matches if and only if at least
                       one member matches.'
--- a/config/crd/bases/edge.kubestellar.io_placementdecisions.yaml
+++ b/config/crd/bases/edge.kubestellar.io_placementdecisions.yaml
@@ -81,17 +81,6 @@ spec:
                       - version
                       type: object
                     type: array
-                  generation:
-                    description: '`ObjectsInternalSpecGeneration` is a sequence number
-                      representing a specific generation of the downsync objects internal
-                      spec. For example, if ClusterScope and NamespaceScope lists
-                      haven''t changed but the internal spec of at least one object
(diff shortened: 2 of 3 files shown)


api/edge/v1alpha1/types.go
```diff
@@ -205,8 +205,9 @@ type PlacementDecisionSpec struct {
 }
 
 // DownsyncObjectReferences explicitly defines the objects to be down-synced.
-// The ClusterScope list defines the cluster-scope objects, while NamespacedObjects packs individual objects
-// identifiable by namespace & name.
+// The ClusterScope list defines the cluster-scope objects, NamespacedObjects packs individual objects
+// identifiable by namespace & name, and generation represents the internal spec generation of the listed
```
MikeSpreitzer, line 209: Instead of "NamespacedObjects packs individual objects identifiable by namespace & name" why not just do the complement to the cluster-scoped clause --- "NamespacedObjects list defines the namespaced objects"?

MikeSpreitzer, line 209: (1) How does "internal space" differ from "spec"? The standard terminology is just "spec" --- or "desired state" if you mean something more than the spec section.

(2) Do you really mean that no changes to anything besides the spec section matter here? No changes to labels or annotations? No changes to ConfigMap or Secret objects?

MikeSpreitzer:
This has got a bad smell for me. It seems to be based on a paradigm different from managing based on comparing desired state and reported state. The field added here is not a description of desired state but rather a signalling mechanism between the implementation of the controller that produces PlacementDecision objects and the implementation of the controller that consumes PlacementDecision objects. In a desired-vs-reported state approach, the producer would list the objects and the consumer would maintain the relationships between the listed objects and their wrapped appearances in the mailboxes.

api/edge/v1alpha1/types.go
nirrozenbaum: right. this comment was left from legacy code (since we took structure from syncerconfig).
will change this.

nirrozenbaum: @MikeSpreitzer thanks for the comments. good questions. 
I will try to better explain. you're right that this comment doesn't explain the issue very good.
I need to update the comment (we do care about labels/annotations, not only spec).

the point we're tackling is how the update of an object is handled. 
Suppose you create a deployment in the WDS and a placement that selects this deployment. 
The placement controller writes the placement decision, the transport controller sees there is a new (or updated) placement decision, it wraps it up and deliver it to the mailbox namespace. 
At this point, the deployment may be updated - e.g., changing the number of replicas from 1 to 3. 
How the transport can informed of the change in spec of the deployment (or annotations or labels)? 

One valid option is to add the above field. 
I should point out that without this field, the PlacementDecision spec is not changed (cause list of objects and list of destinations remain the same), therefore placementdecision generation is not increased. 
with this field it triggers the increment of PlacementDecision.generation.
then we work with the desired state vs current state approach - 
PlacementDecision generation was increased (for example) to 4, while the placementdecision generation that was used to create the wrapped object is 3, so we need to update.
if both equal we don't need to update.
Think about it the following way - if in every object we would add the generation of it, would it be valid?
cause then we specify the concrete generation of the object to distribute which is the desired-state.

There is another valid approach which is to force the placement controller to update the placementdecision object even if nothing in the lists has changed, which will trigger the incrementation of the placementdecision.generation and then we do the same in the part of desired-state vs current-state.
wdyt?

MikeSpreitzer: Does a no-change update increment `ObjectMeta.Generation`?

There is another valid approach, which I tried to outline above. In this approach the placement controller is only responsible for maintaining correct lists in the PlacementDecision, and the transport controller is responsible for maintaining the correct relationships between workload objects and wrapped-up contents in the mailboxes.

MikeSpreitzer: If I understand correctly, in the approach suggested in this PR: (a) the placement controller has to update the ObjectsInternalSpecGeneration field of every PlacementDecision every time the placement controller starts up, and (b) the transport controller has to construct a fresh wrapped-up contents for every PlacementDecision object every time the transport controller starts up.

nirrozenbaum: ```does a no-change update increment ObjectMeta.Generation? I would expect not.```
I'm pretty sure I tested it and there is a field to force the update, including the increment of generation.

option (3) is indeed another valid option, although it sounds to me like a more complex one.
I think options (1) or (2) are very simple and easy for understanding and maintaining, while option (3) might be trickier. 

regarding the startup - yes, we should trigger the update sequence. it might happen if any of the objects has changed or using this field.

nirrozenbaum: another comment on option (3) - 
in our design there is a clear separation of responsibilities between the placement controller and the transport controller.
placement controller is responsible for watching placement and all objects that are included in any of the placements. the placement controller should maintain the list of objects to distribute and list of destinations. upon any change in the objects or destinations, this controller should handle the change.

the transport controller is responsible for watching placement decision, fetch the objects as listed in the decision and distribute it to the destinations as listed in the decision object.

option (3) you suggested violates this separation of concerns. it's the placement controller responsibility to handle any change in placement or any of the objects the are selected by a placement.

nirrozenbaum: @MikeSpreitzer I want to correct myself regarding option (2). 
Maroon and me tested it today and seems like generation is not increasing. I'm pretty sure I was testing something with force and that increased generation, but maybe I had a bug or something that modified the spec and that was the reason for generation change.
so - forget about option (2), sounds like this is not an option :)
we're left with the other options.

pdettori: Since we are talking about options there are yet other two (let's call them option 4 and 5) which might be potentially more efficient. In the current implementation (kslight) as you know there is no creation or update of a PlacementDecision object (which 1) creates the issues described above to manage objects updates and 2) requires writes to the WDS API server for each object update). 

Option 4 - just implement the go interface defined for transport based on the work API and for every pluggable transport implement the code in a separate go project / package (as I think you are doing now) and import that code/package in the KubeStellar controller manager. Assuming we have other transport plugins, the selection of one or the other would be based on some flag on the KS controller manager. We may think this approach may make more difficult to add other plugins, but after all, how many do we expect. Most likely we will have the OCM-based one and the KS syncer-based one.

Option 5 - Expose the go transport interface as a gRPC interface. The transport would run in a separate container in the same pod as the placement controller or in a separate pod and exposed as a service. Whenever the placement controller sees an update on the tracked objects, it runs all the placement matching logic for objects and clusters and invokes the gRPC transport service which takes care of the transport part.

nirrozenbaum: @pdettori correct me if I'm wrong, but option (5) is kind of equal to option (1) in terms of triggering transport reconciliation from the placement controller. upon an internal change in one of the objects that are selected by a placement, option (5) uses gRPC to trigger reconciliation while option (1) increments a generation field in PlacementDecision object to trigger the reconciliation.

At the end, one of the main aspects I'm looking at is simplicity, and triggering reconciliation using a field in a CR sounds like to me like the most straight forward and simple approach.

nirrozenbaum: fixed

pdettori: @nirrozenbaum: in terms of simplicity option 4 is the simplest of all (and probably most performant). Option 5 was adding the gPRC interface on the go interface you already defined for "work"  to avoid the tight coupling of the code. The gPRC interface is not about "triggering" the reconciliation but to do a gRPC call to the transport service that has all the objects reference and list of clusters as payload (basically the content of what you have in the PlacementDecision). The reconciliation happens on the client side (the placement controller) - if the gRPC call does not succeed it will re-enque and retry until succeeds.

api/edge/v1alpha1/types.go
```diff
@@ -205,8 +205,9 @@ type PlacementDecisionSpec struct {
 }
 
 // DownsyncObjectReferences explicitly defines the objects to be down-synced.
-// The ClusterScope list defines the cluster-scope objects, while NamespacedObjects packs individual objects
-// identifiable by namespace & name.
+// The ClusterScope list defines the cluster-scope objects, NamespaceScope list
+// defines the namespace-scope objects and generation represents the internal spec
```
ezrasilvera, line 209: Maybe revision number?  version ? sequence?  personally I find generation confusing

ezrasilvera:
@nirrozenbaum I have a problem with a PR without any issue that describe what is the actual problem we are trying to solve. I tried to read between the lines in this PR but I think this should be clearly stated in some issue. 
I see we now have 5 options that are scattered around all the comments. Can we please open an issue, list all these options (pros, cons), and then we can have a focused discussion? 
Intuitively, I don't feel such a field is the best solution. First it just say that "something" was changed, don't we want to also indicate which of the objects changed (so maybe we don't need to go over all of them again) ?  Do we have a race condition here (+ optimization) ? If there are 2 changes just one after the other - would we go twice over everything ? There are probably some other aspects but I'm not sure I got the intent in all the options

api/edge/v1alpha1/types.go
nirrozenbaum: I agree that generation might get confusing. same probably goes for version which might be confusing with ResourceVersion.
revision sounds good to me.
if we can get an agreement on the general direction with @pdettori and @MikeSpreitzer I'm happy to rename the field to `revision`.

nirrozenbaum:
> @nirrozenbaum I have a problem with a PR without any issue that describe what is the actual problem we are trying to solve. I tried to read between the lines in this PR but I think this should be clearly stated in some issue. I see we now have 5 options that are scattered around all the comments. Can we please open an issue, list all these options (pros, cons), and then we can have a focused discussion? Intuitively, I don't feel such a field is the best solution. First it just say that "something" was changed, don't we want to also indicate which of the objects changed (so maybe we don't need to go over all of them again) ? Do we have a race condition here (+ optimization) ? If there are 2 changes just one after the other - would we go twice over everything ? There are probably some other aspects but I'm not sure I got the intent in all the options

@ezrasilvera @MikeSpreitzer @pdettori @vMaroon 
I created an issue to list the problem and the options discussed here:
https://github.com/kubestellar/kubestellar/issues/1671

ezrasilvera:
> 
> @ezrasilvera @MikeSpreitzer @pdettori @vMaroon I created an issue to list the problem and the options discussed here: #1671

Excellent. Thanks @nirrozenbaum

MikeSpreitzer:
Regarding the name of the new field:
- I dislike "ObjectsInternalSpecGeneration" because (a) there are "objects" everywhere and (b) "spec" is too narrow and (c) I do not see how "internal" is functional at all.
- I dislike "revision" because it is a well-known and very generic term; it begs the question of "revision of _what_ exactly?"
- How about "WorkloadGeneration"?

nirrozenbaum:
> Regarding the name of the new field:
> 
> * I dislike "ObjectsInternalSpecGeneration" because (a) there are many different sorts of "objects" in this system and most of the interfaces are in terms of "objects", so the word "objects" tells the reader approximately nothing, (b) "spec" is too narrow, and (c) I do not see how "internal" is functional at all.
> * I dislike "revision" because it is a well-known and very generic term; it begs the question of "revision of _what_ exactly?"
> * How about "WorkloadGeneration"?

I agree with your points. we were thinking a lot on a name that will be good and not confusing and had no good ideas.
`WorkloadGeneration` sounds excellent to me.

nirrozenbaum:
pushed a commit to rename the field to `WorkloadGeneration` as suggested by @MikeSpreitzer

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #1666, opened 2024-01-26
📖 Update packaging.md

MikeSpreitzer:
## Summary
This PR does various corrections, extensions, and updates to packaging.md.

## Related issue(s)

Fixes #

Issue #1538, opened 2024-01-12
feature:  Handling of workload conflicts when using multiple WDSs


pdettori:
### Feature Description

In kcp TMC's, there was namespaces names mangling and objects were pretty much all namespaced, so there wee no possible conflicts of different workloads names (say deployment with name "deployment1" in namespace "namespace1" ). KubeStellar does not mangle namespaces names (and not even cluster scoped names AFAIK), so we could have scenarios where different workloads from different WDS want to end up in the same WEC. E.g. deployment1/namespace1 from WDS1 in WEC1 and deployment1/namespace1 from WEC2 - note that even if names are the same workloads could be different - in WDS1 could be a deployment for nginx, in WDS2 for something else.

### Proposed Solution

Manifests work have a label to mark owning placement and WDS
When the placement controller update a manifest work, should check if there is a conflict (different WDS ) for a workload (manifest) with the same name (manifest name is derived from apigroup, version, resource name, object name and namespace (the latter if namespaced))
if there is a conflict, it should stop the copy to the conflicting manifest and throw an error so user can correct the placement
It is assumed that end user may not have access to KSLight logs, so in general (not just for this scenario) the error should be shown in the placement conditions - user should always be able to look at placement object status and see what went wrong.

### Want to contribute?

- [ ] I would like to work on this issue.

pdettori:
Reporting a previous discussion that was not tracked in this repo:

[mspreitz](https://github.ibm.com/mspreitz) commented [9 days ago](https://github.ibm.com/dettori/kslight/blob/main/README.md#issuecomment-69844810)
Shouldn't this be a question for individual objects rather than bundles?

The error message in question would be one amalgamated error message per placement object, listing all such errors involving that placement, right? I wonder if something more structured would be better. A structured error section. Or using Kubernetes Event objects.

This design posts errors whenever there is any overlap. That forces users to write non-overlapping placement objects in order to avoid the errors. It might be more convenient to allow overlapping placements and post errors only if there is an actual conflict.

dettori commented 8 days ago • 
@mspreitz thank you for you comments, some more comments/answers:

> Shouldn't this be a question for individual objects rather than bundles?
Yes. In the current KSLight design there is only one object per wrapper. When a conflict is detected, is at the level of the wrapper bundling the object. As mentioned above, manifest name is derived from apigroup, version, resource name, object name and namespace.

>The error message in question would be one amalgamated error message per placement object, listing all such errors involving that placement, right?
Correct

[...]

PR #1669, opened 2024-01-29
📖 add architecture and user interaction docs

### LINKED ISSUE #1575: feature: document the new architecture
### Feature Description

We need a current architecture document.

### Proposed Solution

Create one.

### Want to contribute?

### Additional Context

_No response_

### LINKED ISSUE #1670: Architecture document for the status adds on
We need to have  an architecture document for the status adds on that will include:
1. High level diagram of the participating controllers and objects
   * Description of the policy (what is updated, on which condition)
   * Dependency on external packages/solutions (e.g., is it OCM only ? ) 
3. Relation to the clusters (hubcluster, wec cluster, etc..) 
4.

pdettori:
## Summary

Add architecture and user interaction docs

## Related issue(s)

Fixes #1575
Fixes #1670


.gitignore
```diff
@@ -6,6 +6,7 @@ tools
 build
 .vscode
 docs/scripts/generated_script.sh
+.*.png.bkp
```
MikeSpreitzer, line 9: Is this really something that should go in GitHub? What about backups of other kinds of files?

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+4. Define Placement Policies to specify *what" workloads and *where* should be 
+deployed on the WECs.
+5. Submit Workload Definitions in the native Kubernetes format to the WDSs, 
+and let the Placement Policies govern which WECs should receive them.
+6. Check the status of submitted workloads from the WDS.
 
-The last two are independent and KubeStellar depends on them.
+KubeStellar users can have different roles and responsibilities. 
+For example, a System Admin can set up a KubeStellar instance, 
+create WDSs and ITSs, and register WECs. An Application Owner can
```
MikeSpreitzer, line 29: I think that creating a KubeStellar instance is a much heaver-weight thing that creating a WDS; I would not put them together in the same role.

Since we are not trying to agree here on roles, let's not list hypothetical roles that will just cause arguments. I suggest instead just listing the things that a "user" does.

docs/content/v0.20/user-interaction.md
```diff
@@ -0,0 +1,193 @@
+# User Interaction
```
MikeSpreitzer, line 1: This looks very redundant with the material in the examples. Maybe just point to them, maybe with a very brief outline of what the reader should look for there?

docs/content/v0.20/assets/README.md
```diff
@@ -0,0 +1,5 @@
+# How to edit these pictures
```
MikeSpreitzer, line 1: Why is there an `assets` subdirectory? If the intent is to group related things together, the name `assets` does not really identify their subject matter.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+The user interaction with KubeStellar is described in the 
+[user interaction](./user-interaction.md) section.
 
-The status addon defines the `WorkStatus` Kind of objects and some (what?) associated stuff.
+The KubeStellar architecture has these main modules:
 
-## KubeStellar
+- *KubeStellar Controller Manager*: this module is responsible of delivering workload
+objects from the WDS to the ITS according to placement policies, and 
+updating the status of objects in the WDS.
```
MikeSpreitzer, line 41: You have also been calling this thing the "kubestellar operator", right? That is the way I went in #1666 . This thing not only runs controllers but it also delivers CRD(s) into the WDS, right?

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
-At the higher level there are three modules.
+KubeStellar relies on the concept of *spaces*, which are abstractions 
+that represent a Kubernetes API Server and its controllers. Users can use 
+spaces to perform these tasks:
 
-1. KubeStellar
-2. [the status addon](https://github.ibm.com/dettori/status-addon)
-3. KubeFlex
+1. Create *Workload Definition Spaces* (WDSs) to store the definitions of their workloads.
+2. Create *Inventory and Transport Spaces* (ITSs) to manage the inventory of clusters and
```
MikeSpreitzer, line 17: These are not just _spaces_, they have to be full Kubernetes _clusters_, right? If so then it would be less confusing to call them clusters.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+events, work queues for parallel processing of tasks, and a reconciler
+to ensure the actual state matches the desired state. However, that
+pattern has been extended to provide the following features:
+
+- Using dynamic informers
+- Starting informers on all API Resources (except some that do not need
+  watching)
+- Informers and Listers references are maintained in a hash map and
+  indexed by GVK (Group, Version, Kind) of the watched objects.
+- Using a common work queue and set of workers, where the key is defined as follows:
```
MikeSpreitzer, line 254: FYI: There are lots of kinds of keys floating around. I generally use the term "object reference" here, to make it clearer what these things are.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+  - Key is a struct instead than a string, and contains the following:
+    - GVK of the informer and lister for the object that generated the
+      event
+    - Structured Name of the object 
+    - For delete event: Shallow copy of the object being deleted. This
```
MikeSpreitzer, line 259: A given finalizer is typically managed by one controller, for the benefit of that controller. Why would it matter here whether an object has any finalizers at all?

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+based on placement policies.
+
+The architecture and the event flow for create/update object events is
+illustrated in Figure 3 (some details might be omitted to make the flow easier
+to understand).
+
+<figure>
+  <img src="assets/placement-controller.png"  alt="Placement Controller">
+  <figcaption align="center">Figure 3 - Placement Controller</figcaption>
+</figure>
```
MikeSpreitzer, line 288: That picture says the informers have no handlers for deletion events. That is untrue, if I recall correctly.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+    gets a copy from the work queue. Workers should always receive the
+    key and use it to retrieve the object from the cache.
+
+3.  A worker pulls a key from the work queue, and then does the
+    following processing:
+
+    -  Uses the GVK key to get the reference to the lister for the
+        object
+    -  Gets the object from the lister cache using the key that was
+        generated with the MetaNamespaceKeyFunc.
```
MikeSpreitzer, line 331: Not any more, right?

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+          find the matching clusters using the label selector expression
+          for clusters.
+        - If there are matching clusters, add the names of the cluster
+          to a hashmap setting the name of the cluster as a key. Cluster
+          groups from different placements are merged together.
+    -  If there are no matching clusters, the worker returns without
+        actions and is ready to process other events from the queue.
+    -  If there are matching clusters:
+        - Wraps the object into a ManifestWork
+        - Adds a label to the ManifestWork that is used to track the
```
MikeSpreitzer, line 346: I find it confusing to start by talking about _the_ label and later recognize that there is a label per (Placement, cluster). I think it would be better to start off by explaining that there is a label per (Placement, cluster).

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+re-evaluate all objects vs. the changed placement only, but it saves
+some additional complexity in the code.
+
+#### Placement Deleted
+
+When a new placement is created, the placement controller sets a finalizer
+on it. The Worker pulls a key from queue; if it is a placement and it has been
+deleted (deletion timestamp is set) it follows the flow below:
+
+- Uses lister to list all ManifestWorks on all namespaces in OCM by
```
MikeSpreitzer, line 416: s/OCM/ITS/ or /ITC/

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+    key to store a reference to the object in the work queue. The key
+    contains the GVK key used to retrieve a reference to the informer
+    and lister for that kind of object, and a namespace/name key to
+    retrieve the actual object. Storing the key in the work queue is a
+    common pattern in client-go as the object may have changed in the
+    cache (which is kept updated by the reflector) by the time a worker
+    gets a copy from the work queue. Workers should always receive the
+    key and use it to retrieve the object from the cache.
+
+3.  A worker pulls a key from the work queue, and then does the
```
MikeSpreitzer, line 325: What about a workload object update that shrinks the set of Placements that match?

There is special handling for `Namespace` objects, right?

What about cluster add/change/delete?

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+#### Placement Created or Updated 
+
+Worker pulls key from queue; if it is a placement and it has not been
+deleted (deletion timestamp not set) it follows the following flow:
+
+- Iterates all GVK-indexed listers
+- For each lister
+  - List objects
+    - For each object
+      - Checks object placement label
```
MikeSpreitzer, line 396: You mean the generally plural set of labels on the corresponding wrapped object in the ITS(ITC)? If so then (a) being explicit about that would be helpful, and (b) explicitly mentioning the read (GET or cache fetch) that deliver access to that label set would be helpful.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+invoked. Since most objects do not have a finalizer (which prevents
+deletion from the API server until the controller that sets the
+finalizer does some cleanup and then removes the object), a shallow copy
+of the object is added to a field in the key before pushing it to the
+work queue. Then:
+
+- Worker pulls the key from the work queue
+
+- Flow continues the same way as in the create/update scenario, however
+  the deletedObject field in the key indicates that the object has been
```
MikeSpreitzer, line 376: What if, at this time, there is now an object matching the key pulled from the work queue?

.gitignore
pdettori: Using Draw.io to save the editable png files creates automatically backup files with that extension. Should we rely on contributors to care for not adding those to the git repo or just have an entry in .gitignore for that? If the concern is for the `.*.png.bkp` extension, should we use something like `.*.bkp` instead?

docs/content/v0.20/assets/README.md
pdettori: I was undecided if using the name `images` for that folder, but then I thought that could be confusing as well as observers may think of OCI images or something like that. I found other CNCF projects such as Argo CD use the name `assets` (https://github.com/argoproj/argo-cd/tree/master/docs ) - but if that creates confusion we could use the name `diagrams`. Wdyt ?

docs/content/v0.20/user-interaction.md
pdettori: The main reason I created this file is there is because we are still debating if document by examples or just provide a simple quick start and then document by features. I think we may go with the latter approach in the end, but then we need still to document the user interaction. If this helps, I can point to the example 1 and remove some of the content that overlaps with it  in the user interaction section, as it is likely that we will use example 1 as the "Quick start example"

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
 # KubeStellar Architecture
 
-**NOTE**: This is only a very rudimentary start.
+KubeStellar provides multi-cluster deployment of Kubernetes objects, controlled by a simple
```
ezrasilvera, line 3: deployment and configuration ?  Or is that implied by deployment ?

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+or bundling. The high-level architecture for KubeStellar is illustrated in Figure 1.
 
-## Higher Level
+<figure>
+  <img src="assets/high-level-architecture.png"  alt="High Level Architecture">
+  <figcaption align="center">Figure 1 - High Level Architecture </figcaption>
+</figure>
 
-At the higher level there are three modules.
+KubeStellar relies on the concept of *spaces*, which are abstractions
```
ezrasilvera, line 12: You may take the definition as we had it in our glossary in KS

docs/content/v0.20/architecture.md
ezrasilvera: @MikeSpreitzer Why "full k8s clusters" ?  they need to be running pods, not sure about anything else. Also, it's only for this specific OCM case, not sure what other transports would need.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+3. Register and label Workload Execution Clusters (WECs) with the Inventory and 
+Transport Space, to keep track of the available clusters and their characteristics.
+4. Define Placement Policies to specify *what" workloads and *where* should be 
+deployed on the WECs.
+5. Submit Workload Definitions in the native Kubernetes format to the WDSs, 
+and let the Placement Policies govern which WECs should receive them.
+6. Check the status of submitted workloads from the WDS.
 
-The last two are independent and KubeStellar depends on them.
+KubeStellar users can have different roles and responsibilities.
```
ezrasilvera, line 27: If we want to define roles, we should have a special subsection for roles with description of the different operations each role can perform and how the roles can be assigned

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+provide one or more spaces. We plan to make this optional in the near future.
+
+KubeFlex is a flexible framework that supports various kinds of control planes, such as:
+
+- *k8s*: a basic Kubernetes API Server with a subset of kube controllers. 
+This control plane does not run workloads (e.g., pods), and 
+it uses about 350 MB of memory per instance with a shared Postgres Database Backend.
+- *vcluster*: a virtual cluster that runs on the hosting cluster, 
+based on the  [vCluster Project](https://www.vcluster.com). This type control plane can run pods using worker nodes of the hosting cluster.
+- *host*: the KubeFlex hosting cluster, which is exposed as a control plane.
```
ezrasilvera, line 90: Do we need to say something about the denaturing capabilities?   We knows that k8s API server is a space , however a space is not necessarily a cluster .. So there is a difference in what you can do with each such "space"

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+placement and workload objects on the Workload Definition Space 
+(WDS) and wraps workload objects into a manifest for delivery through the 
+Inventory and Transport Space (ITS). The status controller watches for `workstatuses` objects on the 
+ITS and updates the status of objects in the WDS when singleton status is requested 
+in the placement for those objects. There is one instance of a KubeStellar Controller 
+Manager for each WDS. Currently this controller-manager runs in the KubeFlex hosting cluster.
+More details on the internals of this module are provided in [KubeStellar Controllers Architecture](#kubestellar-controllers-architecture)
+
+## Space Manager
+
```
ezrasilvera, line 76: In general there are sections here that go into explaining what Kflex is , we need to think what should be included here and what should be simply pointed to the KF documentation

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+When using KubeStellar with KubeFlex, users interact with the API server 
+of the hosting cluster to create or delete control planes, and with the 
+hosted API servers to manage *Workloads* and *Placements* or *ManagedClusters*.
+KubeFlex defines a ControlPlane CRD that represents a Control Plane. When a user
+initiates the creation of a Control Plane Custom Resource (CR) by
+executing the `kflex create <cp>` command or the `kubectl apply -f mycontrolplane.yaml` 
+command, the KubeFlex controller creates a new namespace within the hosting cluster, 
+and then deploys the following
+artifacts in that namespace:
+
```
ezrasilvera, line 118: This section is an example to something that maybe should be explained in KF docs and just pointed from here

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+              apply of the manifest into the namespace.
+        - Worker returns and is ready to pick other keys from the queue.
+
+There are other event flows, based on the object GVK and type of event.
+The following sections broadly describe these flows:
+
+#### Object Deleted 
+
+When an object is deleted from the WDS, the handler’s *DeleteFunc* is
+invoked. Since most objects do not have a finalizer (which prevents
```
MikeSpreitzer, line 367: What about a workload object that is deleted after one run on this controller and before the next?

docs/content/v0.20/assets/README.md
MikeSpreitzer: Do we need a subdirectory at all?

.gitignore
MikeSpreitzer: I think it is OK to have tool specific exclusions, if they are commented with an explanation.

docs/content/v0.20/user-interaction.md
MikeSpreitzer: I do not think "document by examples" and "a simple quickstart [presumably plus non-full-example-based documentation]" are the only choices. I think a few examples plus non-full-example-based documentation is also a possibility, and is the path that we are currently on.

Since the material proposed here for this file overlaps so strongly with the existing example 1, I would indeed replace the material here with a pointer to example 1 and, perhaps, a list of points for the reader to look out for in example 1.

docs/content/v0.20/architecture.md
MikeSpreitzer: I think these clusters need support for more than Pod objects. Sure, the current usage probably does not exploit the full Kubernetes API, but unless we identify a subset that we can confidently restrict transports to use, I think we should be saying something like "full kube cluster" or "regular kube cluster".

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+          -  Uses the client connected to the ITS to do a server-side
+              apply of the manifest into the namespace.
+        - Worker returns and is ready to pick other keys from the queue.
+
+There are other event flows, based on the object GVK and type of event.
+The following sections broadly describe these flows:
+
+#### Object Deleted 
+
+When an object is deleted from the WDS, the handler’s *DeleteFunc* is
```
MikeSpreitzer, line 366: Did I overlook the discussion of what happens when the handler's add or update handler is invoked and, after fetching the object, it is found to have a non-zero DeletionTimestamp?

.gitignore
pdettori: Ok, I will add a comment in .gitignore for that line

docs/content/v0.20/architecture.md
pdettori: Sounds good, I will change to be less specific

docs/content/v0.20/user-interaction.md
pdettori: Sounds good.

docs/content/v0.20/assets/README.md
pdettori: We can do without it. We'll just have several .md files and .png files under the same directory. I think it's fine.

docs/content/v0.20/architecture.md
pdettori: The claim for vcluster is that they are a certified distro that "passes 100% of Kubernetes conformance tests" - https://www.businesswire.com/news/home/20210622005444/en/vcluster-Becomes-First-Certified-Kubernetes-Distribution-for-Virtual-Kubernetes-Clusters. Do we need to be so specific when talk about ITS - there are potentially other options for ITS, based on just API server + some OCM controllers, such as https://github.com/open-cluster-management-io/multicluster-controlplane (also available in KubeFlex) - it's just that is not well maintained at this time.

docs/content/v0.20/architecture.md
pdettori: The current name for the deployment is "kubestellar-controller-manager". The idea was to align the name of the module to the name of the deployment.

docs/content/v0.20/architecture.md
pdettori: You are correct it should not matter, as even if there is a finalizer there is no guarantee that that finalizer won't be removed by the owning controller and object will be deleted by the time the worker picks the object ref from the queue.

docs/content/v0.20/architecture.md
pdettori: That is correct, thanks, I fixed the picture to show the deletion event as well for the handler.

docs/content/v0.20/architecture.md
pdettori: Correct. Fixed.

docs/content/v0.20/architecture.md
pdettori: fixed

docs/content/v0.20/architecture.md
pdettori: > What about a workload object update that shrinks the set of Placements that match?
This is not currently not covered. Should be listed in "not currently addressed". I suggest we open a new issue to discuss approaches for this.

> There is special handling for Namespace objects, right?
Can you be more specific? What is the expectation for a namespace object ? At this time it is treated like any other object.

> What about cluster add/change/delete?
This is not currently not covered. Should be listed in "not currently addressed". I suggest we open a new issue to discuss approaches for this.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+  work for the object from each namespace associated with the cluster.
+
+- Worker returns and is ready to pick other keys from the queue.
+
+#### Placement Created or Updated 
+
+Worker pulls key from queue; if it is a placement and it has not been
+deleted (deletion timestamp not set) it follows the following flow:
+
+- Iterates all GVK-indexed listers
```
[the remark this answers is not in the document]
pdettori: Correct

docs/content/v0.20/architecture.md
pdettori: I realized my description here was not accurate and updated to be more clear and consistent with the code. Pls. check after I push the update to the PR.

docs/content/v0.20/architecture.md
pdettori: If the key has a not null `DeletedObject` it will not get the object from the cache (thus it will not get the new object at this time), therefore it will go to the delete flow and that object gets deleted. The creation of a new object with same (GVK,namespace,name) right after the deletion of an object with the same (GVK,namespace,name) will trigger a create event and event handler will create a new key with a null `DeletedObject` which would then be handled like a new object to evaluate vs. placements.

docs/content/v0.20/architecture.md
pdettori: I think it is implied by deployment, since we are talking about Kubernetes objects that can be anything (workloads, operators, policies etc.)

docs/content/v0.20/architecture.md
pdettori: Do you have a pointer to the glossary ? I could not locate it.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
 
-1. KubeStellar
-2. [the status addon](https://github.ibm.com/dettori/status-addon)
-3. KubeFlex
+1. Create *Workload Definition Spaces* (WDSs) to store the definitions of their workloads.
+2. Create *Inventory and Transport Spaces* (ITSs) to manage the inventory of clusters and 
+the transport of workloads.
+3. Register and label Workload Execution Clusters (WECs) with the Inventory and 
+Transport Space, to keep track of the available clusters and their characteristics.
+4. Define Placement Policies to specify *what" workloads and *where* should be
```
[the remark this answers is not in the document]
pdettori: Updated to add the definition of workload (to connect it with objects) - since we use the term WDS etc.it is still appropriate to mention workloads.

docs/content/v0.20/architecture.md
pdettori: @MikeSpreitzer suggested not to list all the ops for roles for now, as we may get stuck into that discussion for a while, so I simplified that section to just mention that there are different roles and when we talk about a "user" he may have all of them.

docs/content/v0.20/architecture.md
pdettori: added some clarification here.

docs/content/v0.20/architecture.md
MikeSpreitzer: I am OK with converging on that name. But the description here should make it clear that this thing also installs CRD(s).

docs/content/v0.20/architecture.md
MikeSpreitzer: Right, I do not want to be too specific. Maybe just say "cluster" instead of "space", to give the reader a heads-up that we do need more than "space" promises.

docs/content/v0.20/architecture.md
MikeSpreitzer: https://docs.kubestellar.io/release-0.15/Getting-Started/user-guide/#glossary

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
 
+- *Status Add-On Controller*: This module installs the OCM status add-on agent 
+on all WECs and sets the RBAC permissions for it using the OCM add-on framework.
+
+- *OCM Agent*: This module registers the WEC to the OCM Hub, watches for 
+ManifestWorks and unwraps and syncs the objects into the WEC.
+
+- *OCM Status Add-On Agent*: This module watches *AppliedManifestWorks* 
+to find objects that are synced by the OCM agent, gets their status 
+and updates *workstatuses* objects in the ITS namespace associated with the WEC.
```
MikeSpreitzer, line 58: In English text I would use the Kind rather than the resource name (as, for example, is done for ManifestWork).

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+A typical event flow for a create/update object event will run as
+follows:
+
+1.  Informer invokes the event handler AddFunc or UpdateFunc
+
+2.  The event handler does some filtering (for example, to ignore update
+    events where the object content is not modified) and then creates a
+    key to store a reference to the object in the work queue. The key
+    contains the GVK key used to retrieve a reference to the informer
+    and lister for that kind of object, and a namespace/name key to
```
MikeSpreitzer, line 322: I would use a plus sign here rather than a slash, between "namespace" and "name".

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+3.  A worker pulls a key from the work queue, and then does the
+    following processing:
+
+    -  Uses the GVK key to get the reference to the lister for the
+        object
+    -  Gets the object from the lister cache using the NamespacedName of 
+       the object.
+    -  Gets the lister for Placement objects and list all placements
+    -  Iterates on all placements, and for each placement:
+        - evaluates the placement label selector vs. object labels to
```
MikeSpreitzer, line 338: There is more to the test now. I do not think it is necessary to go into the details, just say something like "evaluates whether the object matches the downsync selection criteria in the Placement".

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+              apply of the manifest into the namespace.
+        - Worker returns and is ready to pick other keys from the queue.
+
+There are other event flows, based on the object GVK and type of event.
+The following sections broadly describe these flows:
+
+#### Object Deleted 
+
+When an object is deleted from the WDS, the handler’s *DeleteFunc* is
+invoked. Since most objects do not have a finalizer (which prevents
```
MikeSpreitzer, line 369: I do not understand why this text is talking about finalizers. Even if a workload object _did_ have a finalizer, it would be managed by some workload controller and not give any useful guarantee to KubeStellar's placement controller.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+  labels. If yes, remove the label, if not, delete the manifestwork.
+
+Re-enqueuing all object keys forces the re-evaluation of all objects vs.
+all placements. This is a shortcut as it would be more efficient to
+re-evaluate all objects vs. the changed placement only, but it saves
+some additional complexity in the code.
+
+#### Placement Deleted
+
+When a new placement is created, the placement controller sets a finalizer
```
MikeSpreitzer, line 412: s/When a new placement is created/When the placement controller first processes a new Placement/

docs/content/v0.20/architecture.md
MikeSpreitzer: I think GitHub has lost track of where this thread belongs. The same topic is addressed in another thread.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+on it. The Worker pulls a key from queue; if it is a placement and it has been
+deleted (deletion timestamp is set) it follows the flow below:
+
+- Uses lister to list all ManifestWorks on all namespaces in ITS by
```
MikeSpreitzer, line 416: What "lister"? There are no informers and listers on ManifestWork objects, right? At any rate, they are not mentioned in the picture or the outline above.

docs/content/v0.20/architecture.md
MikeSpreitzer: Yes, this is functionally correct but has a performance down-side of doing unnecessary work.

docs/content/v0.20/architecture.md
MikeSpreitzer: I created #1674 for this.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+    contains the GVK key used to retrieve a reference to the informer
+    and lister for that kind of object, and a namespace/name key to
+    retrieve the actual object. Storing the key in the work queue is a
+    common pattern in client-go as the object may have changed in the
+    cache (which is kept updated by the reflector) by the time a worker
+    gets a copy from the work queue. Workers should always receive the
+    key and use it to retrieve the object from the cache.
+
+3.  A worker pulls a key from the work queue, and then does the
+    following processing:
```
MikeSpreitzer, line 330: This does not explain how https://github.com/kubestellar/kubestellar/blob/efc8b6ce0cf65b795a7aeda6e7087d6da840265f/api/edge/v1alpha1/types.go#L55 is handled. As it is currently defined, that option affects the forward propagation, right?

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+          -  Uses the client connected to the ITS to do a server-side
+              apply of the manifest into the namespace.
+        - Worker returns and is ready to pick other keys from the queue.
+
+There are other event flows, based on the object GVK and type of event.
+The following sections broadly describe these flows:
+
+#### Object Deleted 
+
+When an object is deleted from the WDS, the handler’s *DeleteFunc* is
```
MikeSpreitzer, line 368: What if a worker dequeues a key that refers to an object that is _now_ deleted (but wasn't when the event handler ran)?

You may say "no problem, the fact that the object is _now_ deleted implies that there was a following event handler invocation that inserted a key with a copy of the object".  There are two failures of the sequencing assumptions there. One is that the two keys are not equal and thus not combined in the workqueue and thus may be processed concurrently in different workers, thus not serialized. And even if all the work wrt a given object _were_ serialized, the first attempt to handle the non-deletion key might suffer a transient failure causing the key to be requeued for later retry, possibly after the work on the deletion key.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+        - Adds a label for each matched (Placement, WDS) to the ManifestWork that 
+          is used to track the placement that caused the object to be 
+          delivered to one or more clusters. The label contains both the 
+          placement name (note that placement is cluster-scoped) and the WDS name. 
+          This way, when a placement is deleted or updated it is possible to 
+          locate the associated ManifestWorks for deletion or label removal 
+          (if more than one label is present, as other placements may “own” the object).
+        - For each cluster:
+          -  Sets the manifest namespace == name of the cluster
+          -  Uses the client connected to the ITS to do a server-side
```
MikeSpreitzer, line 359: Just saying "a server-side apply" begs some questions. What is the change being submitted? What is the FieldManager set to? In terms of behavioral results, what happens if there is a controller fight between two different WDS supplying different content for the same object?

docs/content/v0.20/architecture.md
pdettori: Ok, added that to the description of the kubestellar-controller-manager.

docs/content/v0.20/architecture.md
pdettori: We are already saying when we talk about the Space Manager/ KubeFlex that there are different type of spaces - it is quite clear from there that they have different capabilities - some are conformant Kube clusters, some are "denatured" to some extent. So far when we talk about clusters we have been mainly talking about the WECs. Perhaps it may confuse the reader now to talk about clusters when we talk about the KS control plane in the center?

docs/content/v0.20/architecture.md
pdettori: Thanks, I have updated with that def. Going over comments before updating the PR.

docs/content/v0.20/architecture.md
pdettori: I will leave the minimum that is needed for readers concerned with understanding KubeStellar overall arch and modules and move/point to KS docs for more details.

docs/content/v0.20/architecture.md
pdettori: Moved out to https://github.com/kubestellar/kubeflex/blob/main/docs/architecture.md and pointed to that section for more info.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
 
-A WDS holds user workload objects and the user's objects that form the interface to KubeStellar control. Currently the only control objects are `Placement` objects. We plan to add `Customization` objects soon, and later objects to specify summarization.
+The status add-on tracks objects applied by the work agent by watching 
+*appliedmanifestworks* objects. These objects list the GVR, name
+and namespace (the latter for namespaced objects) of each object applied
+by the related *ManifestWork*. The status add-on then uses this information 
+to ensure that a singleton informer is started for each GVR, 
+and to track status updates of each tracked object. The status add-on
+then creates/updates *workstatuses* objects in the ITS with the status
+of tracked objects. A `WorkStatus` object contains status for exactly one
```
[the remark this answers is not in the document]
pdettori: In the MB NS. I have updated the doc to say that.

docs/content/v0.20/architecture.md
pdettori: Is this question still related to the discussion on deleted and re-created object with same identifiers (lines 373-376) ? Can you elaborate more on the scenario?

docs/content/v0.20/architecture.md
pdettori: ok, done for all the occurrences in the text.

docs/content/v0.20/architecture.md
pdettori: done

docs/content/v0.20/architecture.md
pdettori: ok, done

docs/content/v0.20/architecture.md
pdettori: Ok, I removed the sentence about finalizers, and as for the logic, we can track this in https://github.com/kubestellar/kubestellar/issues/1674>

docs/content/v0.20/architecture.md
pdettori: Done

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+- Re-enqueues all objects to force re-evaluation: this is done by 
+  iterating all GVK-indexed listers, listing objects for each lister
+  and re-enqueuing the key for each object.
+- Remove ManifestWorks from ITS for objects no longer matching: generate
+  the ManagedByPlacementLabelKey for the current (Placement, WDS) and use
+  that to retrieve all the manifestworks associated with the (Placement, WDS).
+  Then, for each manifestwork, extract the wrapped object, and re-evaluate the
+  object vs. the current placement. If no longer a match (either for the
+  "what" or the "where" part) check if each manifestwork has other (Placement, WDS)
+  labels. If yes, remove the label, if not, delete the manifestwork.
```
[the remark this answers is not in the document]
pdettori: Let's discuss this in https://github.com/kubestellar/kubestellar/issues/1674 - the arch doc at this time describes what is implemented.

docs/content/v0.20/architecture.md
pdettori: That is correct. Setting up informers and listers for `ManifestWork` and `ManagedClusters` was a planned task I could not complete but we should definitely have. But in spirit of describing the current implementation, I updated the text to just say `Lists`.

docs/content/v0.20/architecture.md
pdettori: Correct. I updated the text to describe what happens when that option is used.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+objects, caches to store the objects, event handlers to react to
+events, work queues for parallel processing of tasks, and a reconciler
+to ensure the actual state matches the desired state. However, that
+pattern has been extended to provide the following features:
+
+- Using dynamic informers
+- Starting informers on all API Resources (except some that do not need
+  watching)
+- Informers and Listers references are maintained in a hash map and
+  indexed by GVK (Group, Version, Kind) of the watched objects.
```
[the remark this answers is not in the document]
pdettori: Since `DynamicSharedInformerFactory.ForResource(gvr schema.GroupVersionResource)` require a GVR to create the informer, and when we get an object we have the GVK, the idea was to index by GVK to avoid looking up the GVR every time. Re. your question - are you implying that there is a way to create an informer just with Group and Resource (without version) to avoid creating one for GVR when there is more than one version ? Also, in which scenario two versions of the same CRD are present at the same time?

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,479 @@
...
+
+Currently there are two controllers in the KubeStellar controller
+manager: the placement controller and the status controllor.
+
+### Placement Controller
+
+The Placement controller is responsible for watching workload and 
+placement objects, and wrapping and delivering objects to the ITS
+based on placement policies.
+
```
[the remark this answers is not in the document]
pdettori: Yes, that is correct. When there is an error there is re-enqueing that will result in retries. I have added a sentence to explain that.

docs/content/v0.20/architecture.md
pdettori: The worker tries to retrieve the object if it was not deleted when the event handler ran:

https://github.com/kubestellar/kubestellar/blob/e516d75cc05dcfe16b041d8616381ceaaaea4f1a/pkg/placement/controller.go#L390-L402

As for the concurrency, we can discuss in https://github.com/kubestellar/kubestellar/issues/1674

docs/content/v0.20/architecture.md
pdettori: I think we had the discussion about conflicts for objects with same identifiers from WDSs in https://github.com/kubestellar/kubestellar/issues/1538 and we concluded we want to detect the conflict and report it to the user. Currently the field manager name is the same for all WDSs, but a first step to detect conflict is to use name that includes the WDS name as well.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,433 @@
...
-2. [the status addon](https://github.ibm.com/dettori/status-addon)
-3. KubeFlex
+1. Create *Workload Definition Spaces* (WDSs) to store the definitions of their workloads.
+A Kubernetes workload is an application that runs on Kubernetes. A workload can be made by a 
+single Kubernetes object or several objects that work together.
+2. Create *Inventory and Transport Spaces* (ITSs) to manage the inventory of clusters and 
+the transport of workloads.
+3. Register and label Workload Execution Clusters (WECs) with the Inventory and 
+Transport Space, to keep track of the available clusters and their characteristics.
+4. Define Placement Policies to specify *what" objects and *where* should be
```
waltforme, line 27: A typo in Markdown for 'what'.

docs/content/v0.20/architecture.md
MikeSpreitzer: Thanks. There is still one on line 74.

docs/content/v0.20/architecture.md
MikeSpreitzer: No, this question is about handling changes that result in no event notifications from the informer, because they happen during an interval of time when the controller is crashed.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,433 @@
...
+5. Submit objects in the native Kubernetes format to the WDSs, 
+and let the Placement Policies govern which WECs should receive them.
+6. Check the status of submitted objects from the WDS.
 
-The last two are independent and KubeStellar depends on them.
+In KubeStellar, users can assume a variety of roles and responsibilities. 
+These roles could range from system administrators and application owners 
+to CISOs and DevOps Engineers. However, for the purpose of this document, 
+we will not differentiate between these roles. Instead, we will use the 
+term ‘user’ to collectively refer to any individual who encompasses all these roles.
```
MikeSpreitzer, line 37: "encompasses all these roles" is restrictive --- it excludes individuals who play a strict subset of those roles. I think you want to say something like "any individual in any of those roles". But that would still imply that we think that we have enumerated a definitive set of roles. I think the best approach is something like the following.

> Instead we will use the term 'user' broadly, without attempting to make distinctions among roles.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,433 @@
...
+    -  Iterates on all placements, and for each placement:
+        - evaluates whether the object matches the downsync selection 
+          criteria in the Placement.
+        - If there is a match, list ManagedClusters and
+          find the matching clusters using the label selector expression
+          for clusters.
+        - If there are matching clusters, add the names of the cluster
+          to a hashmap setting the name of the cluster as a key. Cluster
+          groups from different placements are merged together.
+        - If the one of the matched Placements has ` WantSingletonReportedState`
```
MikeSpreitzer, line 299: Did you want to s/^If the one/^If any/ ?

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,433 @@
...
+          set to true, clusters are sorted in alphanumerical order and
+          only the first cluster is selected for delivery.  
+    -  If there are no matching clusters, the worker returns without
+        actions and is ready to process other events from the queue.
+    -  If there are matching clusters:
+        - Wraps the object into a ManifestWork
+        - Adds a label for each matched (Placement, WDS) to the ManifestWork that
```
MikeSpreitzer, line 306: This is not in the scope of an iteration over WDSes, so speaking of "each matched ... WDS" looks weird.

MikeSpreitzer, line 306: Speaking of "each matched Placement" grates against the consideration for WantSingletonReportedState. Does having that bit set anywhere change the "each matched Placement" here to "the Placement selected for delivery"? And what about the Placements that matched but were not selected for delivery?

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,433 @@
...
+          (if more than one label is present, as other placements may “own” the object).
+        - For each cluster:
+          -  Sets the manifest namespace == name of the cluster
+          -  Uses the client connected to the ITS to do a server-side
+              apply of the manifest into the namespace.
+        - Worker returns and is ready to pick other keys from the queue.
+
+There are other event flows, based on the object GVK and type of event. 
+Error conditions cause the re-enqueing of keys, resulting in retries.
+The following sections broadly describe these flows:
```
MikeSpreitzer, line 321: s/:/./

docs/content/v0.20/architecture.md
pdettori: Thanks @waltforme - fixed

docs/content/v0.20/architecture.md
pdettori: done

docs/content/v0.20/architecture.md
pdettori: thanks, my bad

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,433 @@
...
+          criteria in the Placement.
+        - If there is a match, list ManagedClusters and
+          find the matching clusters using the label selector expression
+          for clusters.
+        - If there are matching clusters, add the names of the cluster
+          to a hashmap setting the name of the cluster as a key. Cluster
+          groups from different placements are merged together.
+        - If the one of the matched Placements has ` WantSingletonReportedState`
+          set to true, clusters are sorted in alphanumerical order and
+          only the first cluster is selected for delivery.
```
MikeSpreitzer, line 301: Is anything else done to get the status to come back to WorkStatus objects, or does merely installing the status addon cause status to be returned to WorkStatus objects for all downsynced objects? The reader should be told.

docs/content/v0.20/architecture.md
pdettori: good point, fixed

docs/content/v0.20/architecture.md
pdettori: fixed

docs/content/v0.20/architecture.md
pdettori: Ok, this is listed in https://github.com/kubestellar/kubestellar/issues/1674 item 2, so we can keep track there.

docs/content/v0.20/architecture.md
MikeSpreitzer: There is a description right here, which also should mention defining the API.

docs/content/v0.20/architecture.md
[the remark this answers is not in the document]
MikeSpreitzer: My comment is not about Kind vs resource. It is about the fact that in Kubernetes API machinery a given resource (kind of object) can be available in multiple versions, equivalently, a given object can be available in multiple versions. If the hash map here has an entry for `v1` of a given (group, Kind) and _also_ an entry for `v1beta3` of the same (group, Kind) then this controller will be working with two representations of the same object (for every object of that Kind).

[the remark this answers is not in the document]
MikeSpreitzer: CustomResourceDefinitions support the multiplicity of versions.  See https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/types.go#L60 . As for when that would be exploited, it is similar to the situation for built-in types. It would be done when the owning team needs to do some evolution of the definition.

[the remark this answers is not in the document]
MikeSpreitzer: Remember also that there are also some resources that are NOT custom.

docs/content/v0.20/architecture.md
```diff
@@ -1,36 +1,433 @@
...
+          locate the associated ManifestWorks for deletion or label removal 
+          (if more than one label is present, as other placements may “own” the object).
+        - For each cluster:
+          -  Sets the manifest namespace == name of the cluster
+          -  Uses the client connected to the ITS to do a server-side
+              apply of the manifest into the namespace.
+        - Worker returns and is ready to pick other keys from the queue.
+
+There are other event flows, based on the object GVK and type of event. 
+Error conditions cause the re-enqueing of keys, resulting in retries.
```
MikeSpreitzer, line 320: Presumably not _all_ error conditions: the ones that can not be cured by retry should not be retried.

docs/content/v0.20/architecture.md
MikeSpreitzer: I think this doc should say something about the change submitted and the FieldManager.

docs/content/v0.20/architecture.md
MikeSpreitzer: This doc should say somewhere what happens when the worker is processing a `DeletedObject`-free key and gets `errors.IsNotFound`.

docs/content/v0.20/architecture.md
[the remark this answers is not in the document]
pdettori: But at least for CRDs only one version can be marked as the storage version, and that is the one stored in etcd, right ?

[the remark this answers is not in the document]
pdettori: >  a given object can be available in multiple versions

What is an example of that ? Are you saying that an object may exist in multiple versions simultaneously?

docs/content/v0.20/architecture.md
pdettori: Agree, not an error, normal behavior in Kube.

As for what happens, I am updating the section about create/update (step 3) with this text:

```

File: docs/content/v0.20/architecture.md (hunk 34 above)
Comment (pdettori) [reply docs/content/v0.20/architecture.md:315]: Added some description. Will wait for more comments before updating the PR.

File: docs/content/v0.20/architecture.md (hunk 43 above)
Comment (pdettori) [reply docs/content/v0.20/architecture.md:321]: done

File: docs/content/v0.20/architecture.md (hunk 44 above)
Comment (pdettori) [reply docs/content/v0.20/architecture.md:301]: At this time, the latter. I have added that in the text.

File: docs/content/v0.20/architecture.md (hunk 45 above)
Comment (pdettori) [reply docs/content/v0.20/architecture.md:320]: Changed into "Error conditions may cause..." to account for that.

File: docs/content/v0.20/architecture.md (hunk 42 above)
Comment (pdettori) [reply docs/content/v0.20/architecture.md]: Currently, setting `WantSingletonReportedState` in one of the placements that matches the object affects the behavior for all matching placements,  those selected for delivery (with matching clusters) and those not selected for delivery.

File: docs/content/v0.20/architecture.md (hunk 45 above)
Comment (MikeSpreitzer) [reply docs/content/v0.20/architecture.md]: My point is that not all error conditions do cause retries (if I understand correctly), only a subset. The error conditions that clearly will persist are not retried, right?

File: docs/content/v0.20/architecture.md [hunk 46]
```diff
@@ -1,36 +1,446 @@
...
+
+3.  A worker pulls a key from the work queue, and then does the
+    following processing:
+
+    -  Uses the GVK key to get the reference to the lister for the
+        object
+    -  Gets the object from the lister cache using the NamespacedName of 
+       the object.
+    -  If the object was not found (because it was deleted) after the
+       key was enqueued by the event handler, worker returns. But after that,
```
MikeSpreitzer, line 292: "after" is too strong here. The `DeletedObject`-free key can be pulled from the queue and worked on either/both before and after the `DeletedObject`-bearing key is enqueued. This is for two reasons. One is that unequal keys can be processed concurrently. The other is retries.

docs/content/v0.20/architecture.md
MikeSpreitzer: The text that I am commenting on here says unequivocally that there is a label added for every matching Placement. Is this really true even if some of those matching Placements have `WantSingletonReportedState == true` ?  Also, the following bullet speaks simply of "each cluster". Is that really each cluster matching any Placement that matches the object, regardless of `WantSingletonReportedState` ? I doubt it. I think these pieces of text need updating. Is anything undone in the `WantSingletonReportedState == true` case for clusters that match a Placement that matches the object but are _not_ selected for delivery? The text should say.

docs/content/v0.20/architecture.md
MikeSpreitzer: So the doc issue has been resolved. There remains the code issue of reporting a non-error as an error.

docs/content/v0.20/architecture.md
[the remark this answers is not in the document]
MikeSpreitzer: The current doc text does not speak of storage versions.  In the current revision lines 209--212 are not explicit about which versions are involved. Here is the text I currently see:

> Starting informers on all API Resources (except some that do not need watching)
> Informers and Listers references are maintained in a hash map and indexed by GVK (Group, Version, Kind) of the watched objects.

If you do `kubectl api-versions` I think you will see that some API groups are available in multiple versions (which those are will vary depending on the Kube release you query). While an object is stored in only one version, a client can issue get, list, put, etc requests against any of the available versions. If the controller's hash map contains an entry for each version of the API group that is available, then the controller will be accessing all the objects of that API group via requests for those available versions.

docs/content/v0.20/architecture.md
MikeSpreitzer: That's not unreasonable. In any case, this naming discussion is not limited to this doc.

docs/content/v0.20/architecture.md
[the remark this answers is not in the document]
pdettori: I see, thank you for the explanation. But since the discovery client is calling ServerPreferredResources(), then we should have only one preferred version for each api group, right ? 

https://github.com/kubestellar/kubestellar/blob/c861dc3932d9499b6da565a0f2ec96f958cfc262/pkg/placement/controller.go#L163C22-L163C79

Is then your point here about explaining this in the doc ?

docs/content/v0.20/architecture.md
pdettori: Correct.

docs/content/v0.20/architecture.md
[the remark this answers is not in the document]
MikeSpreitzer: Yes.

docs/content/v0.20/architecture.md
pdettori: I see. I reworded the text to avoid assumptions on sequencing of events.

docs/content/v0.20/architecture.md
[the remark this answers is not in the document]
pdettori: fixed

docs/content/v0.20/architecture.md
pdettori: Since we agreed with @ezrasilvera and @MikeSpreitzer to move ahead and merge, I moved this discussion in this issue https://github.com/kubestellar/kubestellar/issues/1688

MikeSpreitzer (approved):
We agreed to merge this now and continue working on it in follow-on Issues and PRs.
/lgtm
/approve
Thanks!

PR #1675, opened 2024-01-30
✨ transport handling of converging actual state to the desired state

nirrozenbaum:
## Summary

In #1588, the initial code that was merged into main did not handle the current state, but only the desired state.
That was done as part of splitting PRs into smaller pieces.
This PR adds the functionality of getting current state, calculating the desired state as specific in the PlacementDecision CR and converging current state to the desired state (if needed).

This PR also handled some minor points that were raised by @MikeSpreitzer as part of the code review of #1588, for example: 
- in add/remove finalizer functions the deep copy is done only if there is a need to update the object
- clean object uses deep copy in order not to update an object that was fetched using the client 

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/transport/generic_transport_controller.go
+++ b/pkg/transport/generic_transport_controller.go
@@ -308,8 +308,7 @@ func (c *genericTransportController) deleteWrappedObjectsAndFinalizer(ctx contex
 	for _, destination := range placementDecision.Spec.Destinations {
 		if err := c.deleteWrappedObject(ctx, destination.ClusterId, fmt.Sprintf("%s-%s", placementDecision.GetName(), c.wdsName)); err != nil {
 			// wrapped object name is in the format (PlacementDecision.GetName()-WdsName). see updateWrappedObject func for explanation.
-			return fmt.Errorf("failed to delete wrapped object '%s' in destination WEC mailbox namespace '%s' - %w", fmt.Sprintf("%s-%s", placementDecision.GetName(),
-				c.wdsName), destination.ClusterId, err)
+			return fmt.Errorf("failed to delete wrapped object from all destinations' - %w", err)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/transport/generic_transport_controller.go
```diff
@@ -304,13 +305,11 @@ func isObjectBeingDeleted(object metav1.Object) bool {
...
-		if err := c.transportClient.Resource(c.wrappedObjectGVR).Namespace(destination.ClusterId).Delete(ctx, fmt.Sprintf("%s-%s", placementDecision.GetName(), c.wdsName),
-			metav1.DeleteOptions{}); err != nil { // wrapped object name is in the format (PlacementDecision.GetName()-WdsName). see updateWrappedObject func for explanation.
-			if !errors.IsNotFound(err) { // if object is already not there, we do not report an error cause desired state was achieved.
-				return fmt.Errorf("failed to delete wrapped object '%s' in destination WEC with namespace '%s' - %w", fmt.Sprintf("%s-%s", placementDecision.GetName(),
-					c.wdsName), destination.ClusterId, err)
-			}
+	for _, destination := range placementDecision.Spec.Destinations {
+		if err := c.deleteWrappedObject(ctx, destination.ClusterId, fmt.Sprintf("%s-%s", placementDecision.GetName(), c.wdsName)); err != nil {
+			// wrapped object name is in the format (PlacementDecision.GetName()-WdsName). see updateWrappedObject func for explanation.
+			return fmt.Errorf("failed to delete wrapped object '%s' in destination WEC mailbox namespace '%s' - %w", fmt.Sprintf("%s-%s", placementDecision.GetName(),
```
MikeSpreitzer, line 311: FYI, this is repeating information in the wrapped error.

pkg/transport/generic_transport_controller.go
```diff
@@ -403,65 +417,123 @@ func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, plac
...
+			break
+		}
+	}
+	if !found {
+		return nil
+	}
+	// otherwise, object was found and was replaced with last object in the list
+	poppedObject := &list.Items[length-1]
+	list.Items = list.Items[:length-1]
+	return poppedObject
```
MikeSpreitzer, line 466: The control flow would be simpler if these statements were moved to replace the `break` above. And that would make some more simplification obvious (no need to put the returned object at the end of the list).

MikeSpreitzer (commented):
I left a few individual comments.

This is good enough to merge, or you could address those comments in this PR.
/lgtm
Thanks!

nirrozenbaum:
@MikeSpreitzer Thanks for the review, fixed all comments and pushed.

pkg/transport/generic_transport_controller.go
nirrozenbaum: fixed

pkg/transport/generic_transport_controller.go
nirrozenbaum: fixed

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #1676, opened 2024-01-30
🐛 updated placement code to work with interfaces to allow testing

nirrozenbaum:
## Summary

This PR fixes a few issues in the current placement controller:
- current  controller is using concrete clients instead of using an interface for the clients. this usage doesn't allow unit tests cause it's not possible to replace the clients. by replacing all clients to interface, it should be possible to replace it with a fake implementation.
- both placement + status controllers have `map[string]*cache.SharedIndexInforer` and `map[string]*cache.GenericLister`. having a pointer to an interface (which is already a pointer) looks strange and is not needed. changed the type to `map[string]INTERFACE` without the pointer.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/crd/crds.go
+++ b/pkg/crd/crds.go
@@ -48,7 +48,7 @@ var embeddedFiles embed.FS
 
 const FieldManager = "kubestellar"
 
-func ApplyCRDs(dynamicClient *dynamic.DynamicClient, clientset *kubernetes.Clientset, clientsetExt *apiextensionsclientset.Clientset, logger logr.Logger) error {
+func ApplyCRDs(dynamicClient dynamic.Interface, clientset kubernetes.Interface, clientsetExt apiextensionsclientset.Interface, logger logr.Logger) error {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
@@ -128,7 +128,7 @@ func DecodeYAML(yamlBytes []byte) (*unstructured.Unstructured, error) {
 	return obj, nil
 }
 
-func waitForCRDAccepted(ctx context.Context, clientset *apiextensionsclientset.Clientset, crdName string) error {
+func waitForCRDAccepted(ctx context.Context, clientset apiextensionsclientset.Interface, crdName string) error {
 	return wait.PollUntilContextCancel(ctx, 1*time.Second, true, func(ctx context.Context) (bool, error) {
 		crd, err := clientset.ApiextensionsV1().CustomResourceDefinitions().Get(ctx, crdName, metav1.GetOptions{})
 		if err != nil {
--- a/pkg/placement/controller.go
+++ b/pkg/placement/controller.go
@@ -76,12 +76,12 @@ type Controller struct {
 	ctx              context.Context
 	logger           logr.Logger
 	ocmClient        client.Client
-	dynamicClient    *dynamic.DynamicClient
-	kubernetesClient *kubernetes.Clientset
-	extClient        *apiextensionsclientset.Clientset
-	listers          map[string]*cache.GenericLister
+	dynamicClient    dynamic.Interface
+	kubernetesClient kubernetes.Interface
+	extClient        apiextensionsclientset.Interface
+	listers          map[string]cache.GenericLister
 	gvksMap          map[string]*schema.GroupVersionResource
-	informers        map[string]*cache.SharedIndexInformer
+	informers        map[string]cache.SharedIndexInformer
 	stoppers         map[string]chan struct{}
 	workqueue        workqueue.RateLimitingInterface
 	initializedTs    time.Time
@@ -119,8 +119,8 @@ func NewController(mgr ctrlm.Manager, wdsRestConfig *rest.Config, imbsRestConfig
 		dynamicClient:    dynamicClient,
 		kubernetesClient: kubernetesClient,
 		extClient:        extClient,
-		listers:          make(map[string]*cache.GenericLister),
-		informers:        make(map[string]*cache.SharedIndexInformer),
+		listers:          make(map[string]cache.GenericLister),
+		informers:        make(map[string]cache.SharedIndexInformer),
 		stoppers:         make(map[string]chan struct{}),
 		gvksMap:          make(map[string]*schema.GroupVersionResource),
 		workqueue:        workqueue.NewRateLimitingQueue(ratelimiter),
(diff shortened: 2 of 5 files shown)


pdettori:
@nirrozenbaum I am fine in general with the changes, but it looks like the tests are not passing, have you tried the tests locally?

nirrozenbaum:
@pdettori yes I know why it fails. 
there is a function that has to be changed also in kubeflex..
```func GroupVersionKindToResource(clientset kubernetes.Interface, gvk schema.GroupVersionKind) (*schema.GroupVersionResource, error) {
under the file `pkg/util/unstructured.go`. 
I can open also a PR to KF but how can I sync it to this PR?

**Comment by pdettori** [comment]
@nirrozenbaum I think there was some general discussion about reducing dependencies from KubeFlex where possible. We could do the change with a PR in KF and then update the go mod import, or we could just copy that function in the KS and then you do not have that dep. That may make overall updates easier.

**Comment by ezrasilvera** [comment]
@nirrozenbaum Why do we need to pass Interface instead of the actual type for unittest ?  In the unit test if we create a fake NewSimpleDynamicClient   won't we be able to pass it as a dynamic ClientSet ?

**Comment by nirrozenbaum** [comment]
> @nirrozenbaum Why do we need to pass Interface instead of the actual type for unittest ? In the unit test if we create a fake NewSimpleDynamicClient won't we be able to pass it as a dynamic ClientSet ?

@ezrasilvera 

below is the declaration of NewSimpleDynamicClient:
```
func NewSimpleDynamicClient(scheme *runtime.Scheme, objects ...runtime.Object) *FakeDynamicClient
```
both `*FakeDynamicClient` and `*dynamic.DynamicClient` implement the interface `dynamic.Interface`.
if you pass to the controller arg from type `*dynamic.DynamicClient`, you cannot replace it in unit-tests with fake.
if on the other hand you pass arg from type `dynamic.Interface` - you can run with real dynamic client on runtime and with fake on unit-tests.

**Comment by ezrasilvera** [comment]
> > @nirrozenbaum Why do we need to pass Interface instead of the actual type for unittest ? In the unit test if we create a fake NewSimpleDynamicClient won't we be able to pass it as a dynamic ClientSet ?
> 
> @ezrasilvera
> 
> below is the declaration of NewSimpleDynamicClient:
> 
> ```
> func NewSimpleDynamicClient(scheme *runtime.Scheme, objects ...runtime.Object) *FakeDynamicClient
> ```
> 
> both `*FakeDynamicClient` and `*dynamic.DynamicClient` implement the interface `dynamic.Interface`. if you pass to the controller arg from type `*dynamic.DynamicClient`, you cannot replace it in unit-tests with fake. if on the other hand you pass arg from type `dynamic.Interface` - you can run with real dynamic client on runtime and with fake on unit-tests.

Thanks for the clarification. I somehow missed the dynamic.DynamicClient in the function parameters ...

File: pkg/crd/crds.go [hunk 1]
```diff
@@ -81,6 +87,27 @@ func ApplyCRDs(dynamicClient *dynamic.DynamicClient, clientset *kubernetes.Clien
 	return nil
 }
 
+// Convert GroupVersionKind to GroupVersionResource
+func groupVersionKindToResource(clientset kubernetes.Interface, gvk schema.GroupVersionKind) (*schema.GroupVersionResource, error) {
+	resourceList, err := clientset.Discovery().ServerPreferredResources()
+	if err != nil {
+		// ignore the error caused by a stale API service
+		if !strings.Contains(err.Error(), UnableToRetrieveCompleteAPIListError) {
+			return nil, err
```
MikeSpreitzer, line 96: This is erring too aggressively. The problem in `err` might not be relevant to the particular `gvk` of interest here. See #1466

MikeSpreitzer (commented):
I left an individual comment.

This is good enough to merge, or you could address that comment in this PR.
/lgtm
Thanks!

pkg/crd/crds.go
nirrozenbaum: @MikeSpreitzer completely agree with you.
the code that was pulled from KS Light is returning errors as is without wrapping them. 
this should be addressed all over the code (except for new transport code which already wraps errors).
In this PR I wasn't addressing this issue, but only replaced the concrete clients with interfaces to allow testing.

ezrasilvera: @nirrozenbaum please open an issue on that

ezrasilvera: It's a little strange to introduce a new function groupVersionKindToResource  that includes a bad behavior on th error handling.   It's not old code you ignore in this PR , it's new code you introduce .

nirrozenbaum: created an issue for this problem:
https://github.com/kubestellar/kubestellar/issues/1677

@ezrasilvera please pay attention, this is not new code, this function was pulled from KubeFlex code. 
In the existing code before this PR, KS code is calling functions from KF and this is one of the functions.
I pulled the function here since KF function still uses `clientset *kubernetes.Clientset` and without pulling the function here this change is not mergable. 
as I wrote to Mike, I completely agree this is an issue that should be addressed all over the code that was pulled from KSLight. let's try to handle this in #1677

MikeSpreitzer: My OP here is not about wrapping errors. It is about the code failing when it could succeed.

FYI: the code in KubeFlex was cloned from code in KubeStellar _before_ #1466 merged.

ezrasilvera: @nirrozenbaum please add an issue to fix this specific error , to make sure only the relevant errors fail the function (others might be OK in term of the logic here).

ezrasilvera: Let's merge it and push a different PR to fix this bug

nirrozenbaum: @MikeSpreitzer I misunderstood. initially I thought you mean errors are not wrapped and also not explain the problem correctly in some of the case.
getting back to your point - you are correct.

my intention in this PR was not to fix this error but only to replace the specific clients with interface to allow testability.
copying this function from KF here is a side effect since the specific client appeared there as well.
I agree with the general approach to merge this PR to allow unit-test and will create an issue for the specific error.

MikeSpreitzer: Yes, you did.

@pdettori : is there still unnecessary failure in KubeFlex?

pdettori: The imported function was from an older version of KubeFlex (which was used in KSLight). I changed since then that function, so no issues there.

https://github.com/kubestellar/kubeflex/blob/2287f51e6346f6f2b05b00ee52eb415cd3784eed/pkg/util/unstructured.go#L47-L62

nirrozenbaum: @MikeSpreitzer @pdettori please let me know if you want me to change the function to the newer version Paolo pointed to.

Issue #1705, opened 2024-02-05
bug:  client-side throttling when starting dynamic informers for (many) new CRDs


pdettori:
### Describe the bug

When starting of dynamic informers for new CRDs is enabled, in an OpenShift cluster which has ~ 280 CRDs installed, the kubestellar controller manager shows many `client-side throttling` messages in the log.

```
I0205 02:01:39.499569       1 request.go:697] Waited for 17.144362886s due to client-side throttling, not priority and fairness, request: GET:https://kubernetes.default.svc/apis/kubeflow.org/v1alpha1
I0205 02:01:49.499937       1 request.go:697] Waited for 27.144577753s due to client-side throttling, not priority and fairness, request: GET:https://kubernetes.default.svc/apis/operator.tigera.io/v1
2024-02-05T02:01:53Z    INFO    Placement       Got object event        {"apiextensions.k8s.io/v1": "CustomResourceDefinition", "": "trustyaiservices.trustyai.opendatahub.io.trustyai.opendatahub.io"}
I0205 02:01:59.700193       1 request.go:697] Waited for 37.343447865s due to client-side throttling, not priority and fairness, request: GET:https://kubernetes.default.svc/apis/authorization.openshift.io/v1
I0205 02:02:09.899934       1 request.go:697] Waited for 47.543101288s due to client-side throttling, not priority and fairness, request: GET:https://kubernetes.default.svc/apis/observability.open-cluster-management.io/v1beta2
I0205 02:02:20.099804       1 request.go:697] Waited for 57.742738621s due to client-side throttling, not priority and fairness, request: GET:https://kubernetes.default.svc/apis/workload.codeflare.dev/v1beta1
I0205 02:02:30.100048       1 request.go:697] Waited for 8.794602429s due to client-side throttling, not priority and fairness, request: GET:https://kubernetes.default.svc/apis/packages.operators.coreos.com/v1
I0205 02:02:40.100055       1 request.go:697] Waited for 18.794475039s due to client-side throttling, not priority and fairness, request: GET:https://kubernetes.default.svc/apis/apiserver.openshift.io/v1
I0205 02:02:50.300102       1 request.go:697] Waited for 28.994482276s due to client-side throttling, not priority and fairness, request: GET:https://kubernetes.default.svc/apis/tekton.dev/v1
[...]

PR #1707, opened 2024-02-05
✨ add startup option to select allowed resources

### LINKED ISSUE #1697: bug: The Kubestellar Controller Manager is crashing on OCP for CP of type "host"
### Describe the bug

While trying to test Scenario#2 on OCP, I'd noticed that neither the AppWrapper CRD nor the CR were being downsynced to the WECs. This scenario involves a control plane of type host. On closer inspection the controller manager is crashing:

k get pods -n wds2-system
NAME                                              READY   STATUS             RESTARTS       AGE
kubestellar-controller-manager-675ffffd74-9kvch   1/2     CrashLoopBackOff   40 (36s ago)   3h24m

In the log, there are a bunch these:

I0201 20:15:55.356108       1 trace.go:236] Trace[1503902376]: "Reflector ListAndWatch" name:k8s.io/client-go@v0.28.2/tools/cache/reflector.go:229 (01-Feb-2024 20:15:38.046) (total time: 17309ms):
Trace[1503902376]: ---"Objects listed" error:<nil> 17309ms (20:15:55.356)

### Steps To Reproduce

Install kubeflex and kubestellar on OCP
Onboard WEC(s)
Follow setup for Scenario#2 described here: https://github.com/kubestellar/kubestellar/blob/main/docs/content/v0.20/examples.md

### Expected Behavior

Both the AppWrapper CRD and the CR should be down synced to the WEC(s). 

### Additional Context

Paolo's explanation:
"
in a CP of type k8s there are about 40 resource types. In this OCP I found about 395. Since the controller starts an informer for each,the api server does throttling and the informers never complete loading the caches. I need to give an option for running in OCP host to specify the resource to track rather than all
"

pdettori:
## Summary

Add startup option with flag `--resources` to start informers only for selected resources. This feature is mostly useful when the WDS has a large number of API resources and users want to track only a subset of resources. For example, in an OCP cluster used as WDS with ~ 380 API resources, as we recently found in issue #1697.

Also, improve logging with contextual logging for Placement and Status controllers, and fix a race condition found on Status controller that showed up in testing after enabling the `--resource` flag.

## Related issue(s)

Fixes #1697

### DIFF
--- a/cmd/kubestellar-operator/main.go
+++ b/cmd/kubestellar-operator/main.go
@@ -65,7 +65,7 @@ func main() {
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&wdsName, "wds-name", "", "name of the workload description space to connect to")
 	flag.StringVar(&wdsLabel, "wds-label", "", "label of the workload description space to connect to")
-	flag.StringVar(&allowedGroupsString, "api-groups", "", "list of allowed api groups, comma separated. If not specified, all api groups are allowed.")
+	flag.StringVar(&allowedGroupsString, "api-groups", "", "list of allowed api groups, comma separated. Empty string means all API groups are allowed")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
--- a/cmd/kubestellar-operator/main.go
+++ b/cmd/kubestellar-operator/main.go
@@ -60,12 +60,12 @@ func main() {
 	var probeAddr string
 	var wdsName string
 	var wdsLabel string
-	var resourceGroups string
+	var allowedGroupsString string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&wdsName, "wds-name", "", "name of the workload description space to connect to")
 	flag.StringVar(&wdsLabel, "wds-label", "", "label of the workload description space to connect to")
-	flag.StringVar(&resourceGroups, "resource-groups", "", "list of allowed resource groups, comma separated. If not specified, all resources are allowed.")
+	flag.StringVar(&allowedGroupsString, "api-groups", "", "list of allowed api groups, comma separated. If not specified, all api groups are allowed.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
@@ -79,7 +79,7 @@ func main() {
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
 	// parse allowed resources string
-	allowedResources := util.ParseResourceGroupsString(resourceGroups)
+	allowedGroupsSet := util.ParseAPIGroupsString(allowedGroupsString)
 
 	// setup manager
 	// manager here is mainly used for leader election and health checks
@@ -134,7 +134,7 @@ func main() {
 	setupLog.Info("Got config for IMBS", "name", imbsName)
 
 	// start the binding controller
-	bindingController, err := binding.NewController(mgr, wdsRestConfig, imbsRestConfig, wdsName, allowedResources)
+	bindingController, err := binding.NewController(mgr, wdsRestConfig, imbsRestConfig, wdsName, allowedGroupsSet)
 	if err != nil {
 		setupLog.Error(err, "unable to create binding controller")
 		os.Exit(1)
--- a/cmd/kubestellar-operator/main.go
+++ b/cmd/kubestellar-operator/main.go
@@ -60,12 +60,12 @@ func main() {
 	var probeAddr string
 	var wdsName string
 	var wdsLabel string
-	var resources string
+	var resourceGroups string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&wdsName, "wds-name", "", "name of the workload description space to connect to")
 	flag.StringVar(&wdsLabel, "wds-label", "", "label of the workload description space to connect to")
-	flag.StringVar(&resources, "resources", "", "list of allowed resources, in the form <resource>.<api group>/version, comma separated. If not specified, all resources are allowed.")
+	flag.StringVar(&resourceGroups, "resource-groups", "", "list of allowed resource groups, comma separated. If not specified, all resources are allowed.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
@@ -79,11 +79,7 @@ func main() {
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
 	// parse allowed resources string
-	allowedResources, err := util.ParseResourcesString(resources)
-	if err != nil {
-		setupLog.Error(err, "error parsing allowed resources")
-		os.Exit(1)
-	}
+	allowedResources := util.ParseResourceGroupsString(resourceGroups)
 
 	// setup manager
 	// manager here is mainly used for leader election and health checks
--- a/cmd/kubestellar-operator/main.go
+++ b/cmd/kubestellar-operator/main.go
@@ -60,10 +60,12 @@ func main() {
 	var probeAddr string
 	var wdsName string
 	var wdsLabel string
+	var resources string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&wdsName, "wds-name", "", "name of the workload description space to connect to")
 	flag.StringVar(&wdsLabel, "wds-label", "", "label of the workload description space to connect to")
+	flag.StringVar(&resources, "resources", "", "list of allowed resources, in the form <resource>.<api group>/version, comma separated. If not specified, all resources are allowed.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
@@ -75,6 +77,13 @@ func main() {
 	flag.Parse()
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
+	// parse allowed resources string
+	allowedResources, err := util.ParseResourcesString(resources)
+	if err != nil {
+		setupLog.Error(err, "error parsing allowed resources")
+		os.Exit(1)
+	}
 
 	// setup manager
 	// manager here is mainly used for leader election and health checks
@@ -129,7 +138,7 @@ func main() {
 	setupLog.Info("Got config for IMBS", "name", imbsName)
 
 	// start the binding controller
-	bindingController, err := binding.NewController(mgr, wdsRestConfig, imbsRestConfig, wdsName)
+	bindingController, err := binding.NewController(mgr, wdsRestConfig, imbsRestConfig, wdsName, allowedResources)
 	if err != nil {
 		setupLog.Error(err, "unable to create binding controller")
 		os.Exit(1)
@@ -161,4 +170,5 @@ func main() {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
+	select {}
 }
--- a/core-helm-chart/values.yaml
+++ b/core-helm-chart/values.yaml
@@ -5,9 +5,9 @@
 # name of control plane to set as WDS
 ControlPlaneName: wds1
 
-# list of allowed resource groups, comma separated.
-# If not specified, all resources are allowed. 
+# list of allowed Kubernetes api groups, comma separated.
+# If not specified, all api groups are allowed. 
 # When set by --set with helm CLI, use the format
 # "<group1>\, <group2>\, <groupN>" with "" and escaped \,
-ResourceGroups:
+APIGroups:
 
--- a/core-helm-chart/values.yaml
+++ b/core-helm-chart/values.yaml
@@ -2,5 +2,12 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+# name of control plane to set as WDS
 ControlPlaneName: wds1
+
+# list of allowed resource groups, comma separated.
+# If not specified, all resources are allowed. 
+# When set by --set with helm CLI, use the format
+# "<group1>\, <group2>\, <groupN>" with "" and escaped \,
+ResourceGroups:
 
--- a/core-helm-chart/values.yaml
+++ b/core-helm-chart/values.yaml
@@ -8,6 +8,6 @@ ControlPlaneName: wds1
 # list of allowed Kubernetes api groups, comma separated.
 # If not specified, all api groups are allowed. 
 # When set by --set with helm CLI, use the format
-# "<group1>\, <group2>\, <groupN>" with "" and escaped \,
+# "<group1>\,<group2>\,<groupN>" with "" and escaped \,
 APIGroups:
 
--- a/pkg/binding/controller.go
+++ b/pkg/binding/controller.go
@@ -87,26 +87,15 @@ type Controller struct {
 	informers         map[string]cache.SharedIndexInformer
 	stoppers          map[string]chan struct{}
 	placementResolver PlacementResolver
-<<<<<<< HEAD:pkg/binding/controller.go
-
-	workqueue     workqueue.RateLimitingInterface
-	initializedTs time.Time
-	wdsName       string
-}
-
-// Create a new binding controller
-func NewController(mgr ctrlm.Manager, wdsRestConfig *rest.Config, imbsRestConfig *rest.Config, wdsName string) (*Controller, error) {
-=======
 	workqueue         workqueue.RateLimitingInterface
 	initializedTs     time.Time
 	wdsName           string
-	allowedResources  map[string]bool
+	allowedGroupsSet  sets.Set[string]
 }
 
 // Create a new placement controller
 func NewController(mgr ctrlm.Manager, wdsRestConfig *rest.Config, imbsRestConfig *rest.Config,
-	wdsName string, allowedResources map[string]bool) (*Controller, error) {
->>>>>>> 8a41ee7a4 (use api group only to filter resources):pkg/placement/controller.go
+	wdsName string, allowedGroupsSet sets.Set[string]) (*Controller, error) {
 	ratelimiter := workqueue.NewMaxOfRateLimiter(
 		workqueue.NewItemExponentialFailureRateLimiter(5*time.Millisecond, 1000*time.Second),
 		&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(50), 300)},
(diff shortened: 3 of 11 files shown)


MikeSpreitzer:
I really do not think the user should have to supply the API version.

I also think that asking the user to specify the 'resource' part is asking for pain. The goal here is to greatly cut down on the number of informers, and constraining just by API group will be enough for that. I would not get in the business of asking users for more than that.

`schema.GroupResource` has an established string format.

ezrasilvera:
I agree this is not the preferred solution (see my comments and discussion in the issue #1697 ). However this is a mandatory fix even for the initial integration tests with MCAD. 
@nirrozenbaum what is the time estimation to implement your proposal? 
Can we go with this workaround and remove it later on when the correct fix is created ?

nirrozenbaum:
yes, we can. 
if we all agree the correct fix is what was stated above I can open an issue for that and we can get back to it later.
@vMaroon what is your time estimation for the above proposal?
I would estimate it with couple of days at most, but Maroon knows placement code small details, so I'll let him answer the estimation question.

ezrasilvera:
@MikeSpreitzer @nirrozenbaum @pdettori Let me try to summarize some conclusions and proposed plan  as  I think (from the comments) that we are on the same page:

1. Mandating a user to specify the objects up front is probably not a good solution. Adding this as an option for users who prefer to do it (or use cases where it is reasonable) seems to be OK
2. We need  a working solution to fix the current crash issue. Starting with the above PR and adding, at following PRs, the option that doesn't require explicit specification of API objects at setup, seems like a reasnoble approach

pdettori:
> I agree with @MikeSpreitzer that we do not want to mandate that users have to specify api resources in every placement. That was the initial UX and it was a big pain for end users (e.g. for helm charts deployment through KS). The two-pronged approach @MikeSpreitzer proposes may work, but I think there might be some potential ambiguities there too. How would we enforce that if one ObjectTest in a BindingPolicy specify an API Group all the other ObjectTests in the same BindingPoliciy or other BPs have to do that? If we want to do dynamic configuration of the restrictions, there are also other approaches we could discuss, for example we could also create a new CRD for resource configuration (plus other KS configs) that a user may populate to put restrictions if required. 

For the time being, we have an immediate need for the MCAD use case, where we know in advance which resource we care about for the "host" WDS (the app wrapper). There is no need for this use case to change anything dynamically. Even if there is, it can be still done by editing the kubestellar CM deployment (not great, but it does work). 

Here is my proposal: I will update the PR based on @MikeSpreitzer's comments to just require the API Group (and not the version and name of the resource) which should be enough for filtering out most of the unrequired informers, and the we document this feature for 0.20. For longer term, let's design an approach that can maintain the current UX simplicity but allow users to add dynamically constraints on watched resources.

MikeSpreitzer:
@pdettori asked:

> How would we enforce that if one ObjectTest in a BindingPolicy specify an API Group all the other ObjectTests in the same BindingPoliciy or other BPs have to do that?

That is not the constraint we would have to enforce. The constraint would be that one or more of the following things is true.

- The number of resources reported by discovery is not "too large".  We might simply build in a constant number for this. Or we might make it a command line flag with a reasonable default.
- The command line flags include a list of acceptable API groups.
- Each ObjectTest specifies an API group.

Perhaps your question was how to report it when none of those conditions is true? A variety of techniques comes to mind. Certainly there would be controller log messages. There could be Event objects, particularly connected to the offending Placement(s) --- the one(s) with an ObjectTest that does not specify an API group. We could define a status field in Placement (BindingPolicy) specifically to report offending ObjectTests.

Perhaps the question is what to do when none of those conditions is true? In that case I suggest that the controller, in addition to reporting the problem, simply decline to create informers for API groups not listed in any ObjectTest.

MikeSpreitzer:
@pdettori also asked:

> If we want to do dynamic configuration of the restrictions, there are also other approaches we could discuss, for example we could also create a new CRD for resource configuration (plus other KS configs) that a user may populate to put restrictions if required.

I agree, we could invent more ways to post the constraints. I was trying for a kind of simplicity, not inventing new things, just leveraging the existing ones. Admittedly that (number of concepts) is just one measure of simplicity. Another measure is the total size of the input. Putting constraints in one or a few API objects of a new Kind may lead to a smaller total size than specifying an API group in every ObjectTest.

pdettori:
> Perhaps your question was how to report it when none of those conditions is true? A variety of techniques comes to mind. Certainly there would be controller log messages. There could be Event objects, particularly connected to the offending Placement(s) --- the one(s) with an ObjectTest that does not specify an API group. We could define a status field in Placement (BindingPolicy) specifically to report offending ObjectTests.

@MikeSpreitzer - I think we should assume that an end user accessing the wds may not have access to KubeStellar controller logs. It maybe simpler just look at conditions / structured errors in the placement object(s), that would always be available to the user accessing the WDS.

MikeSpreitzer:
@pdettori : we have been talking about design ideas for work to do later, right? We are agreed that this PR should just be simplified to deal with API groups instead of GroupVersionResource, right?

pdettori:
@MikeSpreitzer : correct

ezrasilvera:
@pdettori once you update this PR to just use the API group, let's merge it and move the discussion to #1725

cmd/kubestellar-operator/main.go
```diff
@@ -60,10 +60,12 @@ func main() {
 	var probeAddr string
 	var wdsName string
 	var wdsLabel string
+	var resourceGroups string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&wdsName, "wds-name", "", "name of the workload description space to connect to")
 	flag.StringVar(&wdsLabel, "wds-label", "", "label of the workload description space to connect to")
+	flag.StringVar(&resourceGroups, "resource-groups", "", "list of allowed resource groups, comma separated. If not specified, all resources are allowed.")
```
MikeSpreitzer, line 68: The official term is "API group".

core-helm-chart/values.yaml
```diff
@@ -2,5 +2,12 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+# name of control plane to set as WDS
 ControlPlaneName: wds1
 
+# list of allowed resource groups, comma separated.
```
MikeSpreitzer, line 8: I would say "Kubernetes API groups" instead of "resource groups". Both because it uses standard Kubernetes terminology and because it does not use a term that means something else in everybody's cloud.

pkg/placement/controller.go
```diff
@@ -76,26 +78,26 @@ const placementDecisionQueueingDelay = 2 * time.Second
...
+	stoppers          map[string]chan struct{}
 	placementResolver PlacementResolver
-
-	workqueue     workqueue.RateLimitingInterface
-	initializedTs time.Time
-	wdsName       string
+	workqueue         workqueue.RateLimitingInterface
+	initializedTs     time.Time
+	wdsName           string
+	allowedResources  map[string]bool
```
MikeSpreitzer, line 95: FYI: now that we are no longer stuck on Kubernetes release 1.24 we can use https://github.com/kubernetes/apimachinery/blob/v0.29.1/pkg/util/sets/set.go#L24

pkg/placement/controller.go
```diff
@@ -42,6 +42,8 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
```
MikeSpreitzer, line 46: FYI: Since controller-runtime and zap were developed, Kubernetes got klog and logr-style logging (i.e., both leveled and structured). I think modern Kubernetes code uses the structured logging in klog.

pkg/util/resources.go
```diff
@@ -0,0 +1,65 @@
...
+
+// ParseResourcesString takes a comma separated string list of resources in the form of
+// <api-group1>, <api-group2> .. and returns a map[string]bool
+func ParseResourceGroupsString(resourceGroups string) map[string]bool {
+	if resourceGroups == "" {
+		return nil
+	}
+
+	// trim single and double quotes to make this safe from
+	// user passing the resource option quoted in the container args
```
MikeSpreitzer, line 33: FYI, I do not understand the usage scenario that requires trimming quotes here. It seems odd to me.

pkg/util/resources.go
```diff
@@ -0,0 +1,65 @@
...
+	resourceGroups = strings.Trim(resourceGroups, `'"`)
+
+	groupsMap := map[string]bool{}
+	for _, g := range strings.Split(resourceGroups, ",") {
+		groupsMap[strings.Trim(g, " ")] = true
```
MikeSpreitzer, line 38: I am also not clear on why spaces would need to be trimmed here.

pkg/util/resources.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	return groupsMap
+}
+
+// IsResourceGroupAllowed checks if an infomer for the resource is allowed to start.
+// an empty or nil allowedResources slice is equivalent to allow all,
+func IsResourceGroupAllowed(resourceGroup string, allowedResourceGroups map[string]bool) bool {
+	if allowedResourceGroups == nil || allowedResourceGroups != nil && len(allowedResourceGroups) == 0 {
```
MikeSpreitzer, line 47: (A) FYI: X || ((!X) && Y) == X || Y, for all X and Y (go `||` is lazy).

(B) FYI: "The length of a nil slice, map or channel is 0." (https://go.dev/ref/spec#Length_and_capacity)

pkg/util/resources.go
```diff
@@ -0,0 +1,65 @@
...
+	for _, g := range strings.Split(resourceGroups, ",") {
+		groupsMap[strings.Trim(g, " ")] = true
+	}
+
+	return groupsMap
+}
+
+// IsResourceGroupAllowed checks if an infomer for the resource is allowed to start.
+// an empty or nil allowedResources slice is equivalent to allow all,
+func IsResourceGroupAllowed(resourceGroup string, allowedResourceGroups map[string]bool) bool {
```
MikeSpreitzer, line 46: FYI: https://github.com/kubernetes/apimachinery/blob/v0.29.1/pkg/util/sets/set.go#L82 is a thing

pkg/util/resources.go
```diff
@@ -0,0 +1,65 @@
...
+
+// append the minimal set of resources that are required to operate
+func addRequiredResourceGroups(allowedResourceGroups map[string]bool) {
+	// if resources are provided, we need to ensure that at least CRD and KS API
+	// resources are watched
+
+	allowedResourceGroups[v1alpha1.GroupVersion.Group] = true
+
+	// disabled until https://github.com/kubestellar/kubestellar/issues/1705 is resolved
+	// to avoid client-side throttling
```
MikeSpreitzer, line 63: I do not understand this. #1705 is about a high request rate from this controller for discovery. I do not see how disallowing that one API group makes a difference --- the full discovery is going to be done anyway, right? And even if only allowed groups were queried in discovery, adding this one API group would not greatly increase the requests for discovery.

pkg/util/unstructured.go
```diff
@@ -31,6 +31,7 @@ import (
 
 const (
 	CRDKind                              = "CustomResourceDefinition"
+	CRDResource                          = "customresourcedefinitions"
 	CRDGroup                             = "apiextensions.k8s.io"
```
MikeSpreitzer, line 35: FYI: we do not need our own copy of this string, it is available in https://github.com/kubernetes/apiextensions-apiserver/blob/v0.28.6/pkg/apis/apiextensions/register.go#L24

cmd/kubestellar-operator/main.go
pdettori: fixed in all occurrences

cmd/kubestellar-operator/main.go
```diff
@@ -76,6 +78,9 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	// parse allowed resources string
+	allowedResources := util.ParseResourceGroupsString(resourceGroups)
```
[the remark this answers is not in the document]
pdettori: changed in all occurrences

core-helm-chart/values.yaml
pdettori: done

pkg/placement/controller.go
pdettori: replaced use of map with sets.Set[string]

pkg/placement/controller.go
pdettori: Are you ok in leaving switching to klog in a further PR just focused on logging infrastructure. ?

pkg/util/resources.go
pdettori: Seems like this is no longer an issue when using helm to deploy. Removed the trimming.

pkg/placement/controller.go
MikeSpreitzer: yes

pkg/util/resources.go
pdettori: removed the trimming

pkg/util/resources.go
pdettori: Simplified to only check on len

pkg/util/resources.go
pdettori: changed to use sets.Set

pkg/util/resources.go
pdettori: It's not just that one extra informer for CRD is enabled. When that informer is enabled, https://github.com/kubestellar/kubestellar/blob/a357f6b0783627fd90ff883ad4716d386e8ee4dc/pkg/placement/crd.go#L71 is invoked at startup for each CRD on the system (and there are ~ 300). The call to https://github.com/kubestellar/kubestellar/blob/a357f6b0783627fd90ff883ad4716d386e8ee4dc/pkg/placement/crd.go#L82 was generating a lot of client-side throttling. This issue only shows up when you have a large number of CRDs. I have opened an issue to address that in a different PR.

MikeSpreitzer: So the proposal here is to break everything involving CRDs and fix it in follow-on?

cmd/kubestellar-operator/main.go
```diff
@@ -60,10 +60,12 @@ func main() {
 	var probeAddr string
 	var wdsName string
 	var wdsLabel string
+	var allowedGroupsString string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&wdsName, "wds-name", "", "name of the workload description space to connect to")
 	flag.StringVar(&wdsLabel, "wds-label", "", "label of the workload description space to connect to")
+	flag.StringVar(&allowedGroupsString, "api-groups", "", "list of allowed api groups, comma separated. If not specified, all api groups are allowed.")
```
MikeSpreitzer, line 68: FYI:
`-h` prints the default value, right? I would make the description here say "Empty string means all API groups are allowed".

core-helm-chart/values.yaml
```diff
@@ -2,5 +2,12 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+# name of control plane to set as WDS
 ControlPlaneName: wds1
 
+# list of allowed Kubernetes api groups, comma separated.
+# If not specified, all api groups are allowed. 
+# When set by --set with helm CLI, use the format
+# "<group1>\, <group2>\, <groupN>" with "" and escaped \,
```
MikeSpreitzer, line 11: But no spaces, right?

pkg/util/resources.go
```diff
@@ -0,0 +1,65 @@
...
+
+// ParseAPIGroupsString takes a comma separated string list of api groups in the form of
+// <api-group1>, <api-group2> .. and returns a sets.Set[string]
+func ParseAPIGroupsString(apiGroups string) sets.Set[string] {
+	if apiGroups == "" {
+		return nil
+	}
+
+	groupsSet := sets.Set[string]{}
+	for _, g := range strings.Split(apiGroups, ",") {
```
MikeSpreitzer, line 37: FYI: https://github.com/kubernetes/apimachinery/blob/v0.28.6/pkg/util/sets/set.go#L31 makes the loop here unnecessary.

pkg/util/resources.go
```diff
@@ -0,0 +1,65 @@
...
+	return groupsSet
+}
+
+// IsResourceGroupAllowed checks if a API group is allowed
+// an empty or nil allowedResources slice is equivalent to allow all,
+func IsAPIGroupAllowed(apiGroup string, allowedAPIGroups sets.Set[string]) bool {
+	if len(allowedAPIGroups) == 0 {
+		return true
+	}
+	addRequiredResourceGroups(allowedAPIGroups)
```
MikeSpreitzer, line 50: FYI: I would do this once, after parsing the comma-delimited string

pkg/util/resources_test.go
```diff
@@ -0,0 +1,104 @@
...
+	// Define some test cases with inputs and expected outputs
+	testCases := []struct {
+		name     string
+		input    string
+		expected sets.Set[string]
+	}{
+		{
+			name:     "valid input with api group",
+			input:    "apps, networking.k8s.io, policy",
+			expected: sets.New("apps", "networking.k8s.io", "policy"),
```
MikeSpreitzer, line 37: How does this work, with spaces no longer trimmed?

pkg/util/resources_test.go
```diff
@@ -0,0 +1,104 @@
...
+		},
+	}
+
+	// Iterate over the test cases, run the function with the input and compare expected vs. returned
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			actual := ParseAPIGroupsString(tc.input)
+
+			// Check if the output matches the expected output
+			if !reflect.DeepEqual(actual, tc.expected) {
```
MikeSpreitzer, line 57: FYI, https://github.com/kubernetes/apimachinery/blob/v0.28.6/pkg/util/sets/set.go#L192

pkg/util/resources_test.go
pdettori: you are right, fixed

pkg/util/resources.go
pdettori: yes, but only for this specific case of restricting api groups, because this is what was needed for the current test with MCAD and OCP as hosting cluster.

cmd/kubestellar-operator/main.go
pdettori: done

core-helm-chart/values.yaml
pdettori: correct, fixed

pkg/util/resources.go
MikeSpreitzer: If I understand correctly, this PR stops the controller from properly reacting to CRD add/change/delete but does not stop the controller from handling resources defined by CRDs that were fully in effect before the controller started. Is that right? And you are saying that the plan for the MCAD stack is to get the MCAD CRDs fully in effect before launching this controller. Is that right?

pkg/util/resources.go
pdettori: You are saying to do something like `sets.New(strings.Split(apiGroups, ","))` ? But that returns sets.Set[[]string], which I am not sure how to use

pkg/util/resources.go
pdettori: done

pkg/util/resources_test.go
pdettori: done

pkg/util/resources.go
MikeSpreitzer: `sets.New(strings.Split(apiGroups, ",")...)`

pkg/util/resources.go
pdettori: that's right, on both counts. MCAD CRD is installed on host cluster (and wecs) before starting the controller.

pkg/util/resources.go
pdettori: Ok, thanks, updated

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

pkg/binding/controller.go
```diff
@@ -194,6 +194,9 @@ func (c *Controller) run(workers int) error {
 			if _, excluded := excludedResourceNames[resource.Name]; excluded {
 				continue
 			}
+			if !util.IsAPIGroupAllowed(gv.Group, c.allowedGroupsSet) {
```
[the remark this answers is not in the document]
pdettori: @vMaroon you are right, it looks like it got overwritten in multiple rebases I had to do. Will open another PR to fix.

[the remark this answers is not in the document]
pdettori: adressed in https://github.com/kubestellar/kubestellar/pull/1743

Issue #1534, opened 2024-01-12
Cleanup the hack directory


ezrasilvera:
Remove all unused scripts from hack

clubanderson:
closing - outdated

PR #1709, opened 2024-02-05
🐛 remove "edge" leftovers

nirrozenbaum:
## Related issue(s)

Fixes #


waltforme:
I actually noticed the hack/update-codegen-crds.sh but realized it is part of the legacy things and is covered by #1534 so didn't touch it.

/lgtm

Issue #1557, opened 2024-01-15
feature: rename IMBS to transport space


nirrozenbaum:
### Feature Description

I suggest rename imbs to transport space.
imbs is the acronym for inventory + mailbox space.
in our terms, mailbox+inventory is exactly the transport abstraction. so we could just call it transport space, which makes it clearer.

### Proposed Solution

rename it

### Want to contribute?

- [ ] I would like to work on this issue.

nirrozenbaum:
> @nirrozenbaum just to clarify, you are suggesting that `imbs1` becomes `transport1` in all docs/examples/ arch description ?

yes

nirrozenbaum:
another point to mention - once we add the cluster inventory abstraction, we might want to host the abstract inventory CRs in a space. this would probably be called inventory space (the inventory of KS, not of the transport specific).
therefore, IMBS doesn't make much sense at this point.
so potentially we can have - 
WDS, Transport Space, Inventory Space.
and inventory space can run on the same space as WDS or transport to minimize number of spaces.

ezrasilvera:
I have no strong opinion on that.  Transport seems like not accurate to me, but I don't have a good suggestion :-)

ezrasilvera:
@nirrozenbaum is it urgent ?   Can we do a renaming PR later on ?  (i.e., during phase1 but after we stabilize our docs a little ) ?

PR #1712, opened 2024-02-05
🌱 renamed imbs to transport in all places

nirrozenbaum:
## Related issue(s)

Fixes #

#1557

### DIFF
--- a/cmd/kubestellar-operator/main.go
+++ b/cmd/kubestellar-operator/main.go
@@ -119,17 +119,17 @@ func main() {
 	}
 	setupLog.Info("Got config for WDS", "name", wdsName)
 
-	// get the config for IMBS
-	setupLog.Info("Getting config for IMBS")
-	imbsRestConfig, imbsName, err := util.GetIMBSKubeconfig(setupLog)
+	// get the config for transport space
+	setupLog.Info("Getting config for transport space")
+	transportRestConfig, transportName, err := util.GetTransportKubeconfig(setupLog)
 	if err != nil {
-		setupLog.Error(err, "unable to get IMBS kubeconfig")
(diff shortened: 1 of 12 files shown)


MikeSpreitzer:
I thought the new name is "Inventory and Transport Space" (ITS).

pdettori:
> I thought the new name is "Inventory and Transport Space" (ITS).

I thought the same. Until we come up with a solution for decoupling the inventory from the transport (which may happen with adoption of SIG MC inventory) the inventory and transport are still tightly coupled together

nirrozenbaum:
we have transport space that includes the transport specifics and should have inventory space that includes the KS unified representation of the clusters (with no relation of which transport was selected).
transport space includes also the cluster representation that is specific to the transport (in our case, ManagedCluster CR which is specific to OCM).
inventory space should include the KS representation of the clusters which is calculated from the OCM MC.
Inventory space is out of version 0.20, but we need to leave room for adding inventory space once inventory is added.

nirrozenbaum:
btw, the reason for this PR is because we talked about this in our of the calls and the team decided to do #1557 on phase1.
(prepare for introducing inventory after phase1)

MikeSpreitzer:
I do not recall discussing #1557 

In release 0.20 we _do_ have inventory. It is in the same space as the transport. That is why it makes sense to call it the "Inventory and Transport Space".

nirrozenbaum:
I assume one of us missed something, cause issue #1557 was labeled with phase1 (we went over this item as part of going over all items).

I see that we don't have consensus for this atm, so I can close PR for now and we can get back to this when we do the inventory work.

nirrozenbaum:
closing this PR for now.
will be handled as part of the inventory work (when we split transport and inventory).

PR #1714, opened 2024-02-05
✨ Let `make ko-build-local` produce platform-specific container image

### LINKED ISSUE #1706: bug: `make ko-build-local` is not completely 'local'
### Describe the bug

In the sense that the built image may not fit the build platform. There is an example in the 'Steps To Reproduce' section.

### Steps To Reproduce

1. On a linux/arm64 platform, in the root dir of the KubeStellar repository, do the following.
```
. test/e2e/common/setup-shell.sh
test/e2e/common/setup-kubestellar.sh
```
2. Check the log of the KS controller manager.
```
kubectl --context kind-kubeflex -n wds1-system logs deploy/kubestellar-controller-manager
```
The output should be similar to
```
exec /ko-app/kubestellar-operator: exec format error
```

### Expected Behavior

`make ko-build-local` produces image for the local platform.

### Additional Context

Looks like 'ko build' defaults to amd64. We may specify `--platform` to override the default.

waltforme:
## Related issue(s)

Fixes #1706

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 
 .PHONY: ko-build-local
 ko-build-local: test ## Build local container image with ko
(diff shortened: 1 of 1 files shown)


pdettori:
There is also #1715 - need to decide which one to merge

waltforme:
Latest force-push is a correction on the OS of the image, with the help from @MikeSpreitzer . Thanks!

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

Issue #1544, opened 2024-01-12
feature: more work needed on the Placement API


MikeSpreitzer:
### Feature Description

- We might not want to call the Kind "Placement", since OCM uses exactly the same name for something conceptually different.
- The API Group should not have "edge" in it.
- Should the upsync criteria examine only immutable object attributes?

### Proposed Solution

We have discussed a few possibilities regarding Placement:
- A different name. `Binding` (if the API group provides enough context)? `WhatWhereBinding`? `WorkloadPlacement`? `WorkloadBinding`?
- Factor as in OCM, leaving `Placement` to talk only about "where" and have a distinct binding object. We would need to decide on names for both. This assumes that we continue with our own "where" API.
- Adopt the OCM placement API. It has the advantage of having an implementation that we could just use. It has the disadvantage of directly exposing OCM, hinding transport pluggability. It has the disadvantage of being more complex than we want for our use cases.

For the API group, it is convenient for that to be of the form `<something>.kubestellar.io` --- for a few reasons. For one, the scripting for code generation is oriented around that form.  For another, it leaves space for branching out when we decide we want to make a meaningful decision. Note that in release 0.15 we have two `<something>`s.

### Want to contribute?

- [ ] I would like to work on this issue.

MikeSpreitzer:
@pdettori

MikeSpreitzer:
On the API group, we will take a vote between these two choices:
- `control.kubestellar.io`
- `mc.kubestellar.io` (mc for Multi-Cluster)

MikeSpreitzer:
On the subject of downsync vs upsync criteria: we decided that the upsync criteria would be modified to exclude the tests of object and namespace labels, and to refactor so that we have BasicObjectTest and that is inlined into ExtendedObjectTest.

MikeSpreitzer:
During the architecture meeting we decided to delete upsync from the API for now, and re-opened the question of what to rename/refactor `ObjectTest` to.

pdettori:
@MikeSpreitzer We are still going to have a DownSync field in the types, no? Then we could rename ObjectTest to DownSyncObjectTest, so when decide to add back UpSync we can just add UpSyncObjectTest and the UpSync field without having to change again ObjectTest.

MikeSpreitzer:
@pdettori : yes, think that is probably the most practical approach.

nirrozenbaum:
[...]

PR #1717, opened 2024-02-05
🌱 Changes to the Placement API

waltforme:
## Summary
This PR removes upsync from current Placement API, and renames ObjectTest since upsync becomes absent.

## Related issue(s)
This PR continues to address #1544.


pdettori (approved):
/lgtm

PR #1720, opened 2024-02-06
📖 updated architecture doc with pluggable-transport

nirrozenbaum:
## Related issue(s)

Fixes #


MikeSpreitzer:
This is in the midst of being edited, right? The description of the Placement Controller, AKA Binding Controller, is mostly not updated to account for the way PlacementDecisions, AKA Bindings, are produced.

nirrozenbaum:
> This is in the midst of being edited, right? The description of the Placement Controller, AKA Binding Controller, is mostly not updated to account for the way PlacementDecisions, AKA Bindings, are produced.

yes. we are in the middle of the update for placement controller (Maroon will complete this part).
the PR is currently on hold. we will try to get it done today and unhold.

docs/content/v0.20/architecture.md
```diff
@@ -1,11 +1,12 @@
+**NOTE**: It was decided by the community to rename `Placement` -> `BindingPolicy`.  
+This document uses the new terminology `BindingPolicy` for the resource that was called previously `Placement`, so on every place `BindingPolicy` is mentioned in this doc, we mean `Placement` as it was used up until today.
```
MikeSpreitzer, line 2: This doc also talks about the `Binding` Kind, so its relation to the current truth (`PlacementDecision`) should be noted.

docs/content/v0.20/architecture.md
```diff
@@ -62,21 +63,21 @@ to find objects that are synced by the OCM agent, gets their status
...
-placement and workload objects on the Workload Definition Space 
-(WDS) and wraps workload objects into a manifest for delivery through the 
-Inventory and Transport Space (ITS). The status controller watches for *WorkStatus* objects on the 
-ITS and updates the status of objects in the WDS when singleton status is requested 
-in the placement for those objects. There is one instance of a KubeStellar Controller 
-Manager for each WDS. Currently this controller-manager runs in the KubeFlex hosting cluster and 
-is responsible of installing the required CRDs in the associated WDS.
-More details on the internals of this module are provided in [KubeStellar Controllers Architecture](#kubestellar-controllers-architecture)
+This module manages binding controller (a.k.a placement controller before the rename mentioned at the very beginning of this doc) and status controller. The binding controller watches `BindingPolicy` and workload objects on the Workload Definition Space (WDS) and wraps workload objects into a manifest for delivery through the Inventory and Transport Space (ITS). The status controller watches for *WorkStatus* objects on the ITS and updates the
```
MikeSpreitzer, line 72: The statement that this controller "wraps workload objects into **a** manifest" mixes English plural and singular in a way that is unclear in the presence of a general number of BindingPolicy objects. More explicit quantifiers/relations are needed. I suspect the intent is to describe the future in which only pluggable transport is operating. This doc also needs to be clear about the present, in which only the kslight POC behavior is operating.

MikeSpreitzer: Also, the language "for delivery through the ... ITS" is not entirely clear about whether the wrapped manifests are already in the ITS or will go into them.  So I would reword further, perhaps something like the following.

> ... and maintains in the ITS a wrapped object per workload object to be delivered.

docs/content/v0.20/architecture.md
```diff
@@ -62,21 +63,21 @@ to find objects that are synced by the OCM agent, gets their status
...
+There is one instance of a KubeStellar Controller Manager for each WDS. Currently this controller-manager runs in the KubeFlex hosting cluster and is responsible of installing the required CRDs in the associated WDS.
+More details on the internals of this module are provided in [KubeStellar Controllers Architecture](#kubestellar-controllers-architecture).
+
+### Planned changes
+
+As a step towards a broader change we're currently working on, we plan to decouple the transport handling from the binding controller and to separate them into two different controllers (binding controller & transport controller).
+As part of this change, current binding controller also creates from the `BindingPolicy` a matching `Binding` object in the WDS which contains the concrete workload objects and concrete list of clusters that were selected by the `BindingPolicy`. The future transport controller should watch `Binding` and get from the WDS api server the objects that are specified in the workload section, wrap them into a wrapped object and deliver the wrapped object through the ITS to the WECs specifid in the `Binding` destinations.
```
MikeSpreitzer, line 80: The `Binding` does not "contain" the workload objects, it contains _references to_ the workload objects.

MikeSpreitzer: "WDS api server(s)"

MikeSpreitzer, line 80: "WDS apiserver(s)"

MikeSpreitzer, line 80: s/through the ITS/from the ITS/

docs/content/v0.20/architecture.md
```diff
@@ -229,20 +225,39 @@ pattern has been extended to provide the following features:
...
 
 Currently there are two controllers in the KubeStellar controller
-manager: the placement controller and the status controllor.
+manager: the binding controller and the status controllor.
+
+### Planned changes to the above:
+
+as part of the decoupling of transport from binding as described above, the following changes are included:
+- Binding Controller - one client connected to the WDS space and 
+  (or more in the future) to connect to one or more ITS shards.
```
MikeSpreitzer, line 234: The "(or more in the future)" is not properly attached to anything, I do not know what it is modifying.

docs/content/v0.20/architecture.md
```diff
@@ -229,20 +225,39 @@ pattern has been extended to provide the following features:
...
 
-### Placement Controller
+### Binding Controller (a.k.a. Placement Controller)
 
-The Placement controller is responsible for watching workload and 
-placement objects, and wrapping and delivering objects to the ITS
-based on placement policies.
+The Binding controller is responsible for watching workload objects and 
+`BindingPolicy` objects, and wrapping and delivering objects to the ITS
+based on binding policies. as a step towards the planned change that is describe in this document, Binding Controller also creates for each `BindingPolicy` a matching `Binding` object in WDS. a `Binding` object is mapped 1:1 to a `BindingPolicy` object and contains the concrete list of workload objects and concrete list of destinations that were selected by the policy and
```
MikeSpreitzer, line 251: Need initial cap on new sentence here.

MikeSpreitzer, line 251: s/in WDS/in the WDS/

MikeSpreitzer, line 251: I would make it clear here that the Binding contains _references_ to, not _copies_ of, workload objects.

docs/content/v0.20/architecture.md
```diff
@@ -229,20 +225,39 @@ pattern has been extended to provide the following features:
...
+should get all the workload objects.
+Until the planned change is completed, the `Binding` objects are not used.
```
MikeSpreitzer, line 253: 'are not used' is very vague. It could even mean that the concept of binding object is not used in the code. This is not what we mean here. I would say it a little clearer, perhaps something like "nothing besides this controller touches the `Binding` objects".

docs/content/v0.20/architecture.md
```diff
@@ -229,20 +225,39 @@ pattern has been extended to provide the following features:
...
+`BindingPolicy` objects, and wrapping and delivering objects to the ITS
+based on binding policies. as a step towards the planned change that is describe in this document, Binding Controller also creates for each `BindingPolicy` a matching `Binding` object in WDS. a `Binding` object is mapped 1:1 to a `BindingPolicy` object and contains the concrete list of workload objects and concrete list of destinations that were selected by the policy and
+should get all the workload objects.
+Until the planned change is completed, the `Binding` objects are not used. 
 
-The architecture and the event flow for create/update object events is
+The architecture and the event flow for **CURRENT** code for create/update object events is
 illustrated in Figure 3 (some details might be omitted to make the flow easier
-to understand).
+to understand). once the planned change that is describe in this document is integrated we will update this figure accordingly.
```
MikeSpreitzer, line 257: Initial cap on the first word of a sentence.

docs/content/v0.20/architecture.md
```diff
@@ -292,42 +307,43 @@ follows:
...
+    -  Iterates on all binding-policies, and for each of them:
+        - Evaluates whether the object matches the downsync selection 
+          criteria in the `BindingPolicy`.
+        - If there is a match:
+          - Notes the object as a match for the in-memory representation of the relevant `Binding`.
+            - Enqueues the representation of the relevant `Binding` for syncing if changed. 
+          - Lists ManagedClusters and finds the matching clusters using the label selector expression for clusters.
+            - If there are matching clusters, adds the names of the cluster
+            to a hashmap setting the name of the cluster as a key. Cluster
+            groups from different bindind policies are merged together.
```
MikeSpreitzer, line 320: FYI: Should be indented a little more. Not a functional problem, it seems to work anyway, but would be clearer to read.

docs/content/v0.20/architecture.md
```diff
@@ -353,43 +372,64 @@ before pushing it to the work queue. Then:
...
-  Then, for each manifestwork, extract the wrapped object, and re-evaluate the
-  object vs. the current placement. If no longer a match (either for the
-  "what" or the "where" part) check if each manifestwork has other (Placement, WDS)
-  labels. If yes, remove the label, if not, delete the manifestwork.
+ the ManagedByPlacementLabelKey for the current (`BindingPolicy`, WDS) and use
+ that to retrieve all the manifestworks associated with the (`BindingPolicy`, WDS).
+ Then, for each manifestwork, extract the wrapped object, and re-evaluate the
+ object vs. the current placement. If no longer a match (either for the
+ "what" or the "where" part) check if each manifestwork has other (`BindingPolicy`, WDS)
+ labels. If yes, remove the label, if not, delete the manifestwork.
```
MikeSpreitzer, line 393: The indenting here (2 spaces) was good.

docs/content/v0.20/architecture.md
```diff
@@ -353,43 +372,64 @@ before pushing it to the work queue. Then:
...
-When the placement controller first processes a new Placement, the placement 
+When the binding controller first processes a new `BindingPolicy`, the binding 
 controller sets a finalizer on it. The Worker pulls a key from queue; if it is 
-a placement and it has been deleted (deletion timestamp is set) it follows the flow below:
+a `BindingPolicy` and it has been deleted (deletion timestamp is set) it follows the flow below:
 
-- Lists all ManifestWorks on all namespaces in ITS by
-  placement label.
+- Deletes the in-memory representation of the `Binding`. Note that the actual `Binding` object
+is garbage collected due to the deletion of the `BindingPolicy` and the latter being an owner of the former (using `OwnerReference`).
```
MikeSpreitzer, line 407: FYI: indenting

docs/content/v0.20/architecture.md
```diff
@@ -444,3 +484,25 @@ of tracked objects in the namespace associated with the WEC cluster.
...
 status updates for one object do not require updates of a whole bundle. 
 
+
+### Transport Controller - PLANNED TO BE ADDED
+
+The transport controller is pluggable and allows the option to plug different
+implementations of the transport interface. In order to implement transport plugin, the plugin should supply the following:
+- upon registration of a new WEC, plugin should create a namespace per WEC in ITS (mailbox namespace per WEC);
+- plugin must be able to wrap multiple objects into a single wrapped object;
+- have an agent that runs on the WEC and watches the wrapped object in the corresponding namespace in the central hub and is able to unwrap it and apply the objects to the WEC;
```
MikeSpreitzer, line 494: (A) any number of wrapped objects, not necessarily one.

(B) "an agent that runs on  the WEC and watches the wrapped objects in the corresponding namespace" is overly specific. Much more indirect or complicated techniques should be allowed.

docs/content/v0.20/architecture.md
```diff
@@ -444,3 +484,25 @@ of tracked objects in the namespace associated with the WEC cluster.
...
+- have inventory representation for the clusters.
+
+The above list is required in order to comply with [<u>SIG Multi-Cluster Work API</u>](https://multicluster.sigs.k8s.io/concepts/work-api/).
+
+KubeStellar currently have one transport plugin implementation that is based on [Open Cluster Management](https://open-cluster-management.io) which is also a CNCF Sandbox project. we expect to have more transport plugin options in the future.
```
MikeSpreitzer, line 499: KubeStellar "has", not "have".

docs/content/v0.20/architecture.md
```diff
@@ -444,3 +484,25 @@ of tracked objects in the namespace associated with the WEC cluster.
...
+
+The following section describes how transport controller works, while the described behavior remains the same no matter which transport plugin is selected. The high level flow for the transport controller is describted in Figure 5.
+
+<figure>
+  <img src="./images/transport-controller.png"  alt="Transport Controller">
+  <figcaption align="center">Figure 5 - Transport Controller</figcaption>
+</figure>
+
+The transport controller watches for `Binding` objects on the WDS. `Binding` objects are mapped 1:1 to `BindingPolicy` and contain the list of concrete objects that were selected for distribution by the policy, and the concrete list of destination that were selected by the policy. upon a new `Binding` event (add/update), the transport controller gets from the WDS api-server the objects listed in the `Binding` workload section. it then wraps all objects into a single wrapped object and puts a copy of the wrapped object in every matching cluster mailbox namespace in the ITS. once the wrapped object is in the mailbox namespace of a cluster on the ITS, it's the agent responsibility to pull the wrapped object from there and apply/update/remove the workload objects.
```
MikeSpreitzer, line 508: "into a single wrapped object" is too specific, right? I thought we agreed that wrapping them individually would be an option.

MikeSpreitzer, line 508: (A) What about a Binding update that removes some WECs from the list of destinations?
(B) What about delete of a Binding?

docs/content/v0.20/architecture.md
nirrozenbaum: fixed

docs/content/v0.20/architecture.md
nirrozenbaum: fixed

nirrozenbaum: fixed

docs/content/v0.20/architecture.md
nirrozenbaum: fixed

docs/content/v0.20/architecture.md
nirrozenbaum: fixed

nirrozenbaum: fixed

nirrozenbaum: fixed

docs/content/v0.20/architecture.md
nirrozenbaum: fixed

docs/content/v0.20/architecture.md
nirrozenbaum: fixed

docs/content/v0.20/architecture.md
nirrozenbaum: fixed

docs/content/v0.20/architecture.md
```diff
@@ -444,3 +484,25 @@ of tracked objects in the namespace associated with the WEC cluster.
 A `WorkStatus` object contains status for exactly one object, so that 
 status updates for one object do not require updates of a whole bundle. 
 
+
+### Transport Controller - PLANNED TO BE ADDED
+
+The transport controller is pluggable and allows the option to plug different
+implementations of the transport interface. In order to implement transport plugin, the plugin should supply the following:
+- upon registration of a new WEC, plugin should create a namespace per WEC in ITS (mailbox namespace per WEC);
```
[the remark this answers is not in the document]
nirrozenbaum: fixed

docs/content/v0.20/architecture.md
nirrozenbaum: (A) I changed but not exactly as you described it. I wrote:
plugin must be able to wrap any number of objects into a single wrapped object. once the plugin can wrap any number of objects, it can wrap individual object to a wrapped object multiple times (e.g., each individual object gets its own wrapped object), or wrap multiple objects to a single wrapped object. 

(B) fixed

docs/content/v0.20/architecture.md
nirrozenbaum: fixed

nirrozenbaum: fixed

nirrozenbaum:
@MikeSpreitzer fixed all comments.

docs/content/v0.20/architecture.md
MikeSpreitzer: Since writing my original comment here I have realized that you are documenting a different design than I thought. Let's pursue the debate about what the design should be separately.

If I understand correctly, the design you are trying to document has an additional interface: the interface between the generic transport controller code and the plugin code. The interface to the plugin is NOT the kube API objects, it is a golang interface that the generic transport controller calls. Have I got this right? If so then it should be explicitly discussed in this document. If not then I am still puzzled by the language that I see.

docs/content/v0.20/architecture.md
MikeSpreitzer: My original comment here was based on a misunderstanding of what you were trying to document. If I understand your intent correctly now then no revision was needed here, and the revision that was done made the text more ambiguous. This is not technically incorrect.

MikeSpreitzer: The revised text still only speaks of adding wrapped objects into the mailbox namespaces, not removing.

nirrozenbaum: maybe latest changes are not visible?
I've added at the end the following:
```
Transport controller is based on the controller design patten and aims to bring the current state to the desired state. That is, if a WEC was removed from the Binding, the transport controller will also make sure to remove the matching wrapped object(s) from the WEC's mailbox namespace. Additionally, if a Binding is removed, transport controller will remove the matching wrapped object(s) from all mailbox namespaces.
```

nirrozenbaum: I reverted back to previous version. I also thought about it later and I agree we don't need this change.
this is in the section `Transport Controller - Planned to be added`. This does **NOT** represent current main. it's the planned changes (as you requested, we added the aspirational part).

docs/content/v0.20/architecture.md
nirrozenbaum: I've added 2 more sentences describing the go interface. I don't want to go too much into the details, I think it should be described at the high level. I've added a link to `transport.go` interface and to `ocm-transport-plugin` repo, so if someone wants more details he can go to the links.

docs/content/v0.20/architecture.md
MikeSpreitzer: I see that now.

MikeSpreitzer (approved):
This is good enough to merge.
/lgtm
/approve
Thanks!

PR #1728, opened 2024-02-07
✨ Improve post-test state dump

MikeSpreitzer:
## Summary
This PR revises the state dump at the end of the test workflows, to more clearly separate the different parts.

This is in place of https://github.com/vMaroon/kubestellar/pull/9

## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/pr-test-multicluster.yml
+++ b/.github/workflows/pr-test-multicluster.yml
@@ -45,23 +45,34 @@ jobs:
           cd test/e2e/multi-cluster-deployment 
           KFLEX_DISABLE_CHATTY=true ./run-test.sh 
 
-      - name: Investigate the kubestellar container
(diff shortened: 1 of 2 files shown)


ezrasilvera:
1. I'm not sure that separate the outputs into different sections is better for debugging usage, but will leave that for now.
2. Can we add some prefix to all this sections , anything reasonable is fine state-   debug- info-  summary-

MikeSpreitzer:
The force-push to 18432f4d0 changes the job step names based on the review above.

effi-ofer:
I am late to the party here seeing how this just got merged.  In the future it would be nice to also dump the list of deployments on the WEC clusters.

MikeSpreitzer:
@effi-ofer : I agree. You could offer that in another PR.

Issue #1729, opened 2024-02-07
feature: Rename two Kinds in our API


waltforme:
### Feature Description

We decided by a [vote](https://kubernetes.slack.com/archives/C058SUSL5AA/p1707155424985459?thread_ts=1706903581.862249&cid=C058SUSL5AA) to rename two Kinds in our API so that
- `Placement` becomes `BindingPolicy`;
- `PlacementDecision` becomes `Binding`.

Right now (immediately after merging of #1681), there are 1728 occurrences of 'placement', including 492 occurrences of 'placementdecisions', case insensitive.

I was suggested to create an issue to specifically track the renaming so here it is.

### Proposed Solution

We have a two-step plan for the renaming.

1. (Small and hard part) Update everything in api/ and test/, plus whatever necessary code changes, to get the new API code available and controllers passing the e2e tests. So that others can begin to contribute their new code growing on the fresh APIs.
2. (Big and easy part) Update everything left: variable names, file/directory names, (go) docs, etc. Also make sure we don't touch things that should not be renamed.

I'll create two PRs, one for each of the two steps, to complete the plan. If there are any (transient) inconsistencies between code and doc in the middle, we should make explicit statements about those inconsistencies.

### Want to contribute?

- [ ] I would like to work on this issue.

waltforme:
We have a two-step plan for the renaming.
1. (Small and hard part) Update everything in api/ and test/, plus whatever necessary code changes, to get the new API code available and controllers passing the e2e tests. So that others can begin to contribute their new code growing on the fresh APIs.
2. (Big and easy part) Update everything left: variable names, file/directory names, (go) docs, etc. Also make sure we don't touch things that should not be renamed.

I'll create two PRs, one for each of the two steps, to complete the plan. If there are any (transient) inconsistencies between code and doc in the middle, we should make explicit statements about those inconsistencies.

vMaroon:
@waltforme please note that you must also lookup "decision" on its own.

Issue #1701, opened 2024-02-03
Remember to update the API group (in KS/KS) for WorkStatus


waltforme:
The WorkStatus API is defined in the status addon repo. Then in the KS/KS repo, the KubeStellar controller manager uses its GVR to conditionally start the status controller.

However, because of a [documented difficulty](https://github.com/kubestellar/kubestellar/blob/71e77d281d86741b05a46f3e5ae5c0a08940a398/pkg/util/unstructured.go#L48-L51), the GVR of WorkStatus is explicitly coded today in KS/KS, not imported.

So, if we change the API group of WorkStatus from `edge.kubestellar.io` to something else in the status addon repo, we should remember to make changes in the KS/KS repo accordingly.

pdettori:
we have completed the open sourcing of ocm-status-addon, so we should be able now to work on that part of refactoring

pdettori:
https://github.com/kubestellar/ocm-status-addon/pull/24 is ready for review. As soon as it is merged and a new release for ocm-status-addon is published this task can be completed.

PR #1735, opened 2024-02-08
✨ First step to rename two Kinds in our API

waltforme:
## Summary
This PR takes step 1/2 of https://github.com/kubestellar/kubestellar/issues/1729.


.github/workflows/pr-test-multicluster.yml
```diff
@@ -57,17 +57,17 @@ jobs:
         if: always()
         run: kubectl --context kind-kubeflex logs -n wds1-system $(kubectl --context kind-kubeflex get pod -n wds1-system --selector=control-plane=controller-manager -o jsonpath='{.items[0].metadata.name}')
 
-      - name: show placements
+      - name: show bindingpolicies
         if: always()
-        run: kubectl --context wds1 get placements -o yaml
+        run: kubectl --context wds1 get bindingpolicies.control.kubestellar.io -o yaml
```
MikeSpreitzer, line 62: +1 on adding the API group, BTW.

api/control/v1alpha1/types.go
```diff
@@ -49,19 +49,19 @@ type PlacementSpec struct {
...
-	Conditions         []PlacementCondition `json:"conditions"`
-	ObservedGeneration int64                `json:"observedGeneration"`
+// BindingPolicyStatus defines the observed state of BindingPolicy
+type BindingPolicyStatus struct {
+	Conditions         []BindingPolicyCondition `json:"conditions"`
+	ObservedGeneration int64                    `json:"observedGeneration"`
 }
 
-// Placement is the Schema for the placementpolicies API
+// BindingPolicy defines in which ways the workloads ('what') and the destinations ('where') are bound together.
```
MikeSpreitzer, line 64: +1 on writing a meaningful description here.
I think we have been saying "workload objects" rather than "workloads".

api/control/v1alpha1/types.go
```diff
@@ -161,37 +161,37 @@ type DownsyncObjectTest struct {
...
+// Binding is mapped 1:1 to a single BindingPolicy object.
+// The decision resource reflects the resolution of the BindingPolicy's selectors,
 // and explicitly reflects which objects should go to what destinations.
 //
 // +genclient
 // +genclient:nonNamespaced
 // +kubebuilder:object:root=true
-// +kubebuilder:resource:scope=Cluster,shortName={pd}
-type PlacementDecision struct {
+// +kubebuilder:resource:scope=Cluster,shortName={bd}
```
MikeSpreitzer, line 178: FYI: "bd" is not bad. I could read it as "Binding Decision", and that would warm my heart, until I remember that is not what it means. If the full name is "Binding" then I think "bdg" would be a better shortname.

config/samples/control_v1alpha1_placement.yaml
```diff
@@ -1,11 +1,11 @@
...
+kind: BindingPolicy
 metadata:
-  name: placement1
+  name: bindingpolicy1
 spec:
   clusterSelectors:
   - matchLabels: {"location-group":"edge"}
   downsync:
-  - labelSelectors:
+  - objectSelectors:
```
MikeSpreitzer, line 9: I see that this change is not due to the renaming in this PR but an API change made earlier. Since this inconsistency went undetected since then I wonder: is this file even being used? Should we delete it? If not then there is a test or procedure that needs to be enhanced so that this bug would have been detected.

pkg/util/unstructured.go
```diff
@@ -35,10 +35,10 @@ const (
...
 	ServiceKind                          = "Service"
-	PlacementKind                        = "Placement"
-	PlacementResource                    = "placements"
-	PlacementDecisionKind                = "PlacementDecision"
-	PlacementDecisionResource            = "placementdecisions"
+	BindingPolicyKind                    = "BindingPolicy"
+	BindingPolicyResource                = "bindingpolicies"
+	BindingKind                          = "Binding"
+	BindingResource                      = "bindings"
 	WorkStatusGroup                      = "edge.kubestellar.io" // TODO: update/import after status-addon is rehomed and its api group for WorkStatus is updated
```
MikeSpreitzer, line 42: Red flag here. Need some separate follow-up.

api/control/v1alpha1/conditions.go
```diff
@@ -43,8 +43,8 @@ const (
 	ReasonReconcilePaused  ConditionReason = "ReconcilePaused"
 )
 
-// PlacementCondition describes the state of a control plane at a certain point.
-type PlacementCondition struct {
+// BindingPolicyCondition describes the state of a control plane at a certain point.
```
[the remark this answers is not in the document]
waltforme: Fixed.

api/control/v1alpha1/types.go
waltforme: Changed to "workload objects".

api/control/v1alpha1/types.go
waltforme: Agreed and changed.

config/samples/control_v1alpha1_placement.yaml
waltforme: I asked myself the same question and recorded the question in the history in this [commit](https://github.com/kubestellar/kubestellar/commit/0510ca35d808cebee3d5bfed7fe1c3f47d559abc). I thought someone created these files for their continence of quick tests, didn't want to create any surprise to them so I kept them and updated them.

Today I traced back to the very beginning of the history and realized that 'someone' is actually myself. With that I can safely delete them now.

pkg/util/unstructured.go
waltforme: Yes, so I created #1701 as a reminder.

MikeSpreitzer (commented):
/lgtm
Not sure how we want to interleave this with the other PRs in flight.

waltforme:
Yeah, hate the feeling to make others rebase but I think it is inevitable and better be done early so
/approve

PR #1736, opened 2024-02-09
✨ Add final state dump of Deployment objs in WEC(s)

MikeSpreitzer:
## Summary
This PR extends the state dumping at the end of the e2e tests to include the Deployment objects on the WEC(s).

## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/pr-test-multicluster.yml
+++ b/.github/workflows/pr-test-multicluster.yml
@@ -49,7 +49,7 @@ jobs:
         if: always()
         run: kubectl config get-contexts
 
-      - name: show pods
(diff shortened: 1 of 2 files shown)


effi-ofer:
Again I am late to the party. You guys never sleep :P.  I think the multi cluster test should dump both cluster1 and cluster2 deployment.  I'll add it via a separate PR when I have a few spare moments.

MikeSpreitzer:
@effi-ofer : I think this PR added that? Am I misunderstanding something?

effi-ofer:
You're right.  I must have been temporarily blind.

Issue #1590, opened 2024-01-18
Add CICD tests for user examples running from images


ezrasilvera:
Currently we have 2 CICD tests that build the KS code, need to have the same tests running on the "latest official stable" images

effi-ofer:
@MikeSpreitzer you just pushed #1795 which adds a --released flag to the test and enable switching from using a local image and chart to using the kflex -p option to load the global image.  Does -p installs the image via helm?  If so we can close this issue.  Otherwise, I have a branch that changes the test case to use the global image and charts but was waiting for #1779 to be merged first.

effi-ofer:
@MikeSpreitzer do we still need this issue?  How are you testing newly created release images ?

PR #1738, opened 2024-02-09
✨ add step to install post-install hooks in e2e tests and upgrade kflex to v0.4.2

### LINKED ISSUE #1737: bug: post-create-hooks are missing in e2e after update to kubeflex v0.4.2
### Describe the bug

As agreed previously, kubeflex removed the default pch for ocm and kubestellar. The e2e script now fails with kubeflex v0.4.2 because of missing hooks.

```
$ k --context kind-kubeflex get cps imbs1 -o yaml
apiVersion: tenancy.kflex.kubestellar.org/v1alpha1
kind: ControlPlane
metadata:
  creationTimestamp: "2024-02-09T20:59:32Z"
  finalizers:
  - kflex.kubestellar.org/finalizer
  generation: 1
  name: imbs1
  resourceVersion: "1135"
  uid: dab6f932-60c7-4f65-a158-8f967b756d15
spec:
  backend: shared
  postCreateHook: ocm
  type: vcluster
status:
  conditions:
  - lastTransitionTime: "2024-02-09T21:00:46Z"
    lastUpdateTime: "2024-02-09T21:00:46Z"
    message: ""
    reason: Available
    status: "True"
    type: Ready
  - lastTransitionTime: "2024-02-09T21:00:46Z"
    lastUpdateTime: "2024-02-09T21:00:46Z"
    message: error retrieving post create hook ocm PostCreateHook.tenancy.kflex.kubestellar.org
      "ocm" not found
    reason: ReconcileError
    status: "False"
    type: Synced
  observedGeneration: 0
  secretRef:
    inClusterKey: config-incluster
    key: config
    name: vc-vcluster
    namespace: imbs1-system
```    

### Steps To Reproduce

upgrade KubeFlex CLI to v0.4.2 and run the e2e tests.

### Expected Behavior

Should not fail.

### Additional Context

_No response_

pdettori:
## Summary

add step to install post-install hooks in e2e tests and upgrade kflex to v0.4.2

## Related issue(s)

Fixes #1737

### DIFF
--- a/test/e2e/common/setup-kubestellar.sh
+++ b/test/e2e/common/setup-kubestellar.sh
@@ -31,6 +31,14 @@ kflex init --create-kind $disable_chatty_status
 
 :
 : -------------------------------------------------------------------------
+: Install the post-create-hooks for ocm and kubstellar controller manager
+:
+kubectl apply -f ${SRC_DIR}/../../../config/postcreate-hooks/ocm.yaml
+kubectl apply -f ${SRC_DIR}/../../../config/postcreate-hooks/kubestellar.yaml
+: Kubestellar post-create-hooks applied.
+
+:
+: -------------------------------------------------------------------------
 : 'Create an inventory & mailbox space of type vcluster running OCM (Open Cluster Management) directly in KubeFlex. Note that -p ocm runs a post-create hook on the vcluster control plane which installs OCM on it.'
 :
 kflex create imbs1 --type vcluster -p ocm $disable_chatty_status
--- a/.github/workflows/pr-test-multicluster.yml
+++ b/.github/workflows/pr-test-multicluster.yml
@@ -35,10 +35,10 @@ jobs:
       - name: Install dependencies
         run: |
           curl -L https://raw.githubusercontent.com/open-cluster-management-io/clusteradm/main/install.sh | bash
-          wget https://github.com/kubestellar/kubeflex/releases/download/v0.4.1/kubeflex_0.4.1_linux_amd64.tar.gz
-          tar -xvf kubeflex_0.4.1_linux_amd64.tar.gz bin/kflex
+          wget https://github.com/kubestellar/kubeflex/releases/download/v0.4.2/kubeflex_0.4.2_linux_amd64.tar.gz
+          tar -xvf kubeflex_0.4.2_linux_amd64.tar.gz bin/kflex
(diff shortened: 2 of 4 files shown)


test/e2e/common/setup-kubestellar.sh
```diff
@@ -29,6 +29,14 @@ fi
 kflex init --create-kind $disable_chatty_status
 : Kubeflex kind cluster created.
 
+:
+: -------------------------------------------------------------------------
+: Install the post-create-hooks for ocm and kubstellar controller manager
+:
+kubectl apply -f ${SRC_DIR}/../../../config/postcreate-hooks/ocm.yaml
+kubectl apply -f ${SRC_DIR}/../../../config/postcreate-hooks/kubestellar.yaml
```
MikeSpreitzer, line 37: I think this one is irrelevant, because the commands below for creating the WDS do not use a PostCreateHook

waltforme: I agree that the `kubestellar` hook is not used in today's tests (which builds the KS controller manager from source). The reason that I think this line should be in this script is that it can be useful in the future, for https://github.com/kubestellar/kubestellar/issues/1590, as a 'common setup' step.

effi-ofer: I am not commenting on whether this is right PR for these lines, but I wanted to point out that we will need them for tests of release images.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -29,6 +29,14 @@ fi
 kflex init --create-kind $disable_chatty_status
 : Kubeflex kind cluster created.
 
+:
+: -------------------------------------------------------------------------
+: Install the post-create-hooks for ocm and kubstellar controller manager
+:
+kubectl apply -f ${SRC_DIR}/../../../config/postcreate-hooks/ocm.yaml
```
MikeSpreitzer, line 36: BTW, this needs to become irrelevant too. Currently this line is using something built from an older version of this same git repo. Instead it should use the local copy of the repo, as is done in place of the other PostCreateHook.

pdettori: @MikeSpreitzer can you explain more precisely what do you mean by "local copy of the repo" for the ocm post-create-hook ?

PR #1741, opened 2024-02-10
✨ Context and Logging improvements

MikeSpreitzer:
## Summary
This PR makes a few improvements in Context handling and logging.

It uses one `context.Context` created in `func main` and passed down in the usual way, instead of making many new `context.Background()` in several places.

FYI: https://go.dev/ref/spec#Receive_operator says that it is valid to apply the receive operator to a `nil` channel, and it will (as you would expect) never actually receive anything. This PR simplifies more complex code that used `Context.WithCancel` unnecessarily.

This PR switches from zap to using normal klog leveled and structured logging, getting better log messages: the timestamps have microsencond precision rather than whole second, and the source filename is mentioned.

Log statements that refer to an API object have been changed to not stringify the reference unless and until it is decided that this is actually needed.

This PR adds logging of the command line flag values at the start of the controller-manager.

This PR also turns up the logging level on the controller-manager.

This PR also fixes a couple of bugs where the key/value alternation of the `Info` method was not being followed (half the values appeared where a key belonged).

## Related issue(s)

Fixes #

### DIFF
--- a/core-helm-chart/templates/operator.yaml
+++ b/core-helm-chart/templates/operator.yaml
@@ -202,6 +202,7 @@ spec:
         - --leader-elect
         - --wds-name={{.Values.ControlPlaneName}}
         - --api-groups={{.Values.APIGroups}}
+        - -v=2
         image: ghcr.io/kubestellar/kubestellar/kubestellar-operator:0.20.0-alpha.1
         imagePullPolicy: IfNotPresent
         livenessProbe:
--- a/pkg/status/controller.go
+++ b/pkg/status/controller.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	ctrlm "sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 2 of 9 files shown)


pkg/status/controller.go
```diff
@@ -49,7 +49,6 @@ import (
 // a placement that cuase an object to be delivered to a cluster has singleton statsus specified
 // the full status will be copied to the object.
 type Controller struct {
-	ctx                context.Context
```
MikeSpreitzer, line 52: BTW, nothing was using this field.

MikeSpreitzer:
@waltforme

pkg/util/keys.go
```diff
@@ -106,7 +107,9 @@ func KeyFromGVKandNamespacedName(gvk schema.GroupVersionKind, name types.Namespa
 	return fmt.Sprintf("%s/%s/%s/%s", gvk.Group, gvk.Version, gvk.Kind, name.String())
 }
 
-// Used for generating a single string unique representation of the object for logging info
+// Used for generating a single string unique representation of the object for logging info.
+//
+// Deprecated: use RefToRuntimeObj instead.
 func GenerateObjectInfoString(obj runtime.Object) string {
```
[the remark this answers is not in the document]
MikeSpreitzer: I was thinking of not breaking concurrent work that adds code.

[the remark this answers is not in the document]
MikeSpreitzer: Removed now.

vMaroon:
BTW unrelated to this PR, but check this out: https://github.com/kubestellar/kubestellar/pull/1707#discussion_r1485589271

If it does not violate any flow, do you think you can slip fixing this issue into this PR?

core-helm-chart/templates/operator.yaml
```diff
@@ -202,6 +202,7 @@ spec:
         - --leader-elect
         - --wds-name={{.Values.ControlPlaneName}}
         - --api-groups={{.Values.APIGroups}}
+        - -v=2
```
pdettori, line 205: Is the intent to allow changing verbosity only editing the deployment at runtime ? Or should we expose this at the helm chart values level ?

pdettori (approved):
/lgtm

core-helm-chart/templates/operator.yaml
MikeSpreitzer: This is a matter for future PRs. I definitely want to give deployers more control over verbosity. I also suspect we will want to be a little more organized and principled about how we use various verbosity levels.

Issue #1745, opened 2024-02-12
bug: pluggable transport attempts to wrap empty workloads


vMaroon:
### Describe the bug

currently the transport tries to produce empty wrapped objects when reconciling a binding with empty workload. this leads to errors in OCM's case, since a manifestwork cannot be created while wrapping no resources.

### Steps To Reproduce

read above

### Expected Behavior

syncing bindings with empty workloads should be checking and enforcing that no relevant wrapped object exists.

nirrozenbaum:
this bug may happen in the following scenario (putting here only one example):
a user creates object with label X (e.g., a deployment) and a BindingPolicy that selects objects with label X.
at some point the the user may delete the deployment but BindingPolicy is still there, which results in re-evaluation of the BindingPolicy and the creation of Binding with empty workload section.
This is a valid scenario.

The solution:
transport controller should check if workload section is empty and if yes it should not distribute anything to any cluster. it should remove all existing manifest work from all destinations (empty binding should remove the manifest work).

nirrozenbaum:
fixed in PR #1747

PR #1747, opened 2024-02-12
🐛 fixed empty workload transport distribution

nirrozenbaum:
## Summary

As described in bug #1745 there are scenarios where a binding object with empty workload section is created. 
This might happen when objects selectors selects objects with label 'X' when no such object exists in multiple variations.
This is a valid scenario and transport controller should handle it properly.

## Related issue(s)

Fixes #
#1745

### DIFF
--- a/pkg/transport/generic_transport_controller.go
+++ b/pkg/transport/generic_transport_controller.go
@@ -356,17 +356,10 @@ func (c *genericTransportController) updateWrappedObjectsAndFinalizer(ctx contex
 		return fmt.Errorf("failed to build wrapped object from PlacementDecision '%s' - %w", placementDecision.GetName(), err)
 	}
 	// converge actual state to the desired state
-	for _, destination := range placementDecision.Spec.Destinations {
-		currentWrappedObject := c.popWrappedObjectByNamespace(currentWrappedObjectList, destination.ClusterId)
-		if currentWrappedObject != nil && currentWrappedObject.GetAnnotations() != nil &&
-			currentWrappedObject.GetAnnotations()[originOwnerGenerationAnnotation] == desiredWrappedObject.GetAnnotations()[originOwnerGenerationAnnotation] {
-			continue // current wrapped object is already in the desired state
-		}
-		// othereise, need to create or update the wrapped object
-		if err := c.createOrUpdateWrappedObject(ctx, destination.ClusterId, desiredWrappedObject); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/transport/generic_transport_controller.go
```diff
@@ -438,6 +435,31 @@ func (c *genericTransportController) initializeWrappedObject(ctx context.Context
 	return wrappedObject, nil
 }
 
+func (c *genericTransportController) propagateWrappedObjectToClusters(ctx context.Context, desiredWrappedObject *unstructured.Unstructured,
+	currentWrappedObjectList *unstructured.UnstructuredList, destinations []v1alpha1.Destination) error {
+	// if the desired wrapped object is nil, that means we should not propagate this object.
+	// this may happen when the workload section is empty.
+	// this is not an error state but a valid scenario.
+	// return without propagating, the delete section will remove existing instances of the wrapped object from all current destinations.
+	if desiredWrappedObject == nil {
```
vMaroon, line 444: Wouldn't ::initializeWrappedObject return an empty wrapped object, and not a nil pointer? That is according to transport's WrapObjects documentation:
> WrapObjects gets slice of objects and wraps them into a single wrapped object. In case slice is empty, the function should return an empty wrapped object.

pkg/transport/generic_transport_controller.go
```diff
@@ -438,6 +435,31 @@ func (c *genericTransportController) initializeWrappedObject(ctx context.Context
...
+		return nil // this is not considered an error.
```
MikeSpreitzer, line 445: What about deleting the existing wrapped objects, if any, for this BindingPolicy?

MikeSpreitzer: Oh, I see, the trimming of `currentWrappedObjectList` is suppressed in this case.

MikeSpreitzer (commented):
I have not tested, only inspected the code change. It
/lgtm

ezrasilvera:
@nirrozenbaum is this code being used in the current cicd tests ?  Or you tested it manually on the side ?
/hold

nirrozenbaum:
> @nirrozenbaum is this code being used in the current cicd tests ? Or you tested it manually on the side ? /hold

@ezrasilvera transport controller still not used in cicd (will be soon hopefully). 
this bug was found in manual e2e tests and the fix was tested in a manual test.

Issue #1750, opened 2024-02-12
bug: confusing content in cmd/


MikeSpreitzer:
### Describe the bug

Normal usage is that each subdirectory of `cmd/` corresponds to an executable to build. Today we have two exceptions.
1. `cmd/client-options` has an expected-to-be-shared package. That would normally go in `pkg/`.  **Fixed**
2. `cmd/transport` is just a fragment of a transport controller. If I understand correctly, there could be multiple transport controllers, and each would be a distinct executable and each could use the fragment currently in `cmd/transport`. If I have that right then this is also something that would normally go in `pkg/`.### 

### Steps To Reproduce

See above

### Expected Behavior

Each subdirectory of `cmd/` corresponds to an executable to build.

nirrozenbaum:
@MikeSpreitzer thanks for opening this issue.
1. you are correct. client options doesn't belong to `cmd`, although I also feel it doesn't belong to `pkg`.
I've looked for something similar in k8s github and the closest example I could find is the following:
https://github.com/kubernetes/kubernetes/blob/5909bd2ddf9b4a4ee2c6507be7dc50d1920e5035/staging/src/k8s.io/controller-manager/options/generic.go
I think maybe we can put it under options dir. (so we have cmd, options, pkg). wdyt? 

2. In a running KS system, there are no multiple transport controllers, there is just one. The single transport controller is pluggable and in order to plug the specific implementation, I've put a function that gets the impl as arg and then runs the single controller.
This is not a fragment of the transport controller, **this is the only transport controller**. Whoever initializes this controller just need to supply the specific impl of the interface. 
there could be different plugins that initialize this controller with different implementations of the interface.
so about (2) I still believe this should reside in cmd.
I could find an example for such a controller in k8s github. maybe we can look for examples in other CNCF repos (like prometheus, etc..)

nirrozenbaum:
btw, about point (2) - if you look in the base-controller.go file contents:
if we would rename Run function to main and remove the plugin injection, the code is a classic cmd file.

MikeSpreitzer:
[...]

PR #1751, opened 2024-02-13
🌱 move client options from cmd to options dir

nirrozenbaum:
## Summary

fix point (1) in issue #1750 

## Related issue(s)

Fixes #


MikeSpreitzer (approved):
I think `github.com/kubestellar/kubestellar/pkg/cli-options` would be a better choice than `github.com/kubestellar/kubestellar/options`, but this PR is an improvement.

/lgtm
/approve
Thanks!

Issue #1694, opened 2024-02-01
bug: incorrect handling of Placement and workload objects in placement controller


MikeSpreitzer:
### Describe the bug

From reviewing the architecture document, and first noted in #1674, problems in that controller include the following.

1. An update to a workload object that removes some Placements from the matching set is not handled correctly.
2. If a workload object is deleted, or changed to remove some Placements from the matching set, while the controller is down this will not be handled correctly.
3. A Placement update that (a) removes workload objects or clusters from their respective matching sets and (b) happens while the controller is not running, is not handled correctly.

### Steps To Reproduce

This was from based on reviewing the architecture document.

### Expected Behavior

Controller maintains intended eventually consistent relations between input and output objects.

waltforme:
I'm reading https://github.com/kubestellar/kubestellar/pull/1669#discussion_r1472208727 and tracking to this issue. I'm so curious about the "change/deletion happens while the controller is down" case. How are we going to solve this?

nirrozenbaum:
@MikeSpreitzer now that pluggable transport was merged, can we check which of the points still happen and which is fixed?
/cc @vMaroon 
/cc @ezrasilvera

vMaroon:
@MikeSpreitzer I think this is fixed, right?

PR #1753, opened 2024-02-13
:sparkles:  Initial Release limitations doc

ezrasilvera:
## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/v0.20/release-notes.md
+++ b/docs/content/v0.20/release-notes.md
@@ -1,0 +1,11 @@
+# Release notes
+
+The following sections list the known issues for each release. The issue list is not differential (i.e., compared to previous releases) but a full list representing the overall state of the specific release. 
+
+## Known constraints and limitations release: v0.20-alpha.2
+* Dynamic changes to WECs are not supported. Existing placements will not be updated when new WECs are added or when labels are added/deleted on existing WECs
+* Removing of WorkStatus objects (on the transport namespace) is not supported and may not result in recreation of that object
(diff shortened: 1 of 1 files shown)


docs/content/v0.20/release-notes.md
```diff
@@ -0,0 +1,11 @@
+# Release notes
+
+The following sections list the known issues for each release. The issue list is not differential (i.e., compared to previous releases) but a full list representing the overall state of the specific release. 
+
+## Known constraints and limitations release: v0.20-alpha.2
+* Dynamic changes to WECs are not supported. Existing placements will not be updated when new WECs are added or when labels are added/deleted on existing WECs
+* Removing of WorkStatus objects (on the transport namespace) is not supported and may not result in recreation of that object
+* Singleton: It's the user responsibility to make sure there are no shared objects in two different (singleton) placements that target two different WECs. Currently there is no enforcement on on that. 
+* Objects on two different WDSs shouldn't have the exact same identifier (same group, version, kind, name and namespace). Such a conflict is currently not identified.
+
```
pdettori, line 10: There are also some limitations related to removing labels from a workload object and changes when controller is down that are listed in https://github.com/kubestellar/kubestellar/issues/1694

effi-ofer:
Might it be worth while to add an indication on when a particular limitation would be lifted? Some of these limitations (dynamic WEC comes to mind) are going to be very short lived.

docs/content/v0.20/release-notes.md
ezrasilvera: Added. Thanks

Issue #1535, opened 2024-01-12
Update kubestellar.io content (remove all non relevant documents from the docs folder)


ezrasilvera:
1. Remove all documents related to our legacy code (example1, milestones, getting started, etc..)
2. Add a link to our updated 20.X.X content

Issue #1612, opened 2024-01-21
feature: choose a testing framework for the e2e tests


effi-ofer:
### Feature Description

We are currently using shell scripts for end to end testing.  Is that the best approach?  Should we switch to something else?

### Proposed Solution

Choose a testing framework.

### Want to contribute?

- [X] I would like to work on this issue.

effi-ofer:
Ginkgo is a widely used BDD framework.  It is used by kubevirt and OCM for e2e testing.  I am looking at it as a contender to replace the shell scripts.

pdettori:
@effi-ofer do you have a pointer to a e2e test written with Ginkgo for one of the two projects mentioned, just to get a feel of how it would look like?

MikeSpreitzer:
We need documentation for contributors showing a few scenarios of exercising the version of the repo that they are working on (not a previous release). The current bash-based E2E tests do that (contributors can be presumed to be able to read and understand bash). I have no objection to adding tests written in other ways, but we should retain instructions for contributors and these instructions must be tested in CI. Having E2E tests written in bash accomplishes that.

effi-ofer:
> @effi-ofer do you have a pointer to a e2e test written with Ginkgo for one of the two projects mentioned, just to get a feel of how it would look like?

Sorry @pdettori for replying only now.  Check out the PR associated with this issue.

effi-ofer:
> We need documentation for contributors showing a few scenarios of exercising the version of the repo that they are working on (not a previous release). The current bash-based E2E tests do that (contributors can be presumed to be able to read and understand bash). I have no objection to adding tests written in other ways, but we should retain instructions for contributors and these instructions must be tested in CI. Having E2E tests written in bash accomplishes that.

Once the pluggable transporter is fully online, we expect the examples in the associated PR will pass and then we can revisit this.

effi-ofer:
We choose ginkgo for e2e and the tests will be delivered in PR #1713.  Another issue (#1865) has been opened to add bash based examples.  I am therefore closing this issue.

MikeSpreitzer:
I will also comment in #1865. I think that it is important to retain at least one example of how a contributor would issue shell commands to exercise their own local copy/version of the system.

PR #1758, opened 2024-02-13
📖 Start drafting release process

### LINKED ISSUE #1732: feature: design and document procedure for making KS releases and snapshots
### Feature Description

We need to be able to make KubeStellar snapshots and releases, and do so without breaking users of previous ones. Other repositories in the kubestellar org should also have their own release processes.

What is a snapshot? Maybe it is a release that we do not advertise widely because it does not have all the user-visible features that we want in the next release. Maybe it is a git tag and equal "image" (published container and published Helm chart(s)) tag with instructions.

One part of the problem is just defining the criteria that a release process should meet.

### Proposed Solution

I do not have a proposal yet.

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR is intended to design and document the release process for this repo. This PR is currently incomplete.

## Related issue(s)

Fixes #1732

This interacts with the website design, whose updating is (sort of?) in #1535 .

### DIFF
--- a/docs/content/v0.20/release.md
+++ b/docs/content/v0.20/release.md
@@ -1,0 +1,84 @@
+# Making KubeStellar Releases
+
+This document defines how releases of the KubeStellar repository are made.
+
+See the associated [packaging and delivery doc](packaging.md) for some
+clues about the problem.
+
+## Goals and limitations
+
+The release process has the following goals.
+
+- A release is identified using [semantic versioning](semver.org). This means that the associated semantics are followed, in terms of what sort of changes to the repo require what sort of changes to the release identifier.
+- A user can pick up and use a given existing release without being perturbed by on-going contributor work. A release is an immutable thing.
+- A release with a given semver identifier is built from a commit of this Git repository tagged with a tag whose name is "v" followed by the release identifier.
+- There is a concept of a release branch, with a name following the pattern `release-$major.$minor`. This allows contributors to do maintenance work on older release branches. This also allows a user to track work along such a line.
+- The contents of `main` always work. This includes passing CI tests. This includes documentation being accurate. We allow point-in-time specific documentation, such as a document that says "Here is how to use release 1.2.3" --- which would refer to a release made in the past. We do not require the documentation in `main` to document all releases.
+- A git tag is immutable. Once associated with a given Git commit, that association is not changed later.
+- We do not put self-references into Git. For example, making release `1.2.3` doe not require changing any file in Git to have the string `1.2.3` in it.
+
+We have the following limitations.
+
+- The only way to publish artifacts (broadly construed, not (necessarily) GitHub "release artifacts") is to make a release.
+- The only way to test published artifacts is to make a release and test it.
+- Thus, it is necessary to keep users clearly appraised of the quality (or status of evaluating the quality) of each release.
+- Because of the lack of self references, most user instructions (e.g., examples) and tests do not have concrete release identifiers in them; instead, the user has to chose and supply the release identifier. There can also be documentation of a specific past release (e.g., the latest stable release) that uses the literal identifier for that past release.
+- **PAY ATTENTION TO THIS ONE**: Because of the prohibition of self references, **Git will not contain the exact bytes of our Helm chart definitions**. Where a Helm chart states its own version or has a container image reference to an image built from the same release, the bytes in Git have a placeholder for that image's tag and the process of creating the published release artifacts fills in that placeholder. Think of this as being analogous to the linking done when building a binary executable file.
+- The design below fails to completely meet the goal of not putting self-references in files under Git control. The failure is in the KubeFlex PostCreateHook that installs the kubestellar-controller-manager (KCM), where the version of the container image for the KCM appears.
+
+## Technology
+
+There will be a GitHub workflow that creates the published artifacts for each Git tag whose name starts with "v". The rest of the tag name is required to be a semver release identifier. Note that this document does not (yet, anyway) specify **how** that GitHub workflow gets its job done.
+
+For each tag `v$version` the following published artifacts will be created.
+
+- The container image for the kubestellar-controller-manager (KCM), at `ghcr.io/kubestellar/kubestellar/controller-manager`. Image tag will be `$version`. This GitHub "package" will be connected to the ks/ks repo (this connection is something that an admin will do once, it will stick for all versions).
+- The Helm chart (for installing the KCM for a WDS), at `ghcr.io/kubestellar/kubestellar/controller-manager-chart` with version `$version` and Helm "appVersion" `$version`. This GitHub "package" will also be connected to the ks/ks repo. The chart has a reference to container image for the KCM and that reference is `ghcr.io/kubestellar/kubestellar/controller-manager:$version`. **In Git the chart has only placeholders in these places, _not_ `$version`; the `$version` is inserted into a distinct copy by the GitHub workflow, which then publishes this specialized copy.**
+- Note that there is no automation (yet) concerning the KubeFlex PostCreateHook that installs the KCM.
+
+## Testing and Examples
+
+The unit tests (of which we have almost none right now), integration tests (of which we have none yet), and end-to-end (E2E) tests in this repository are run in the context of a local copy of this repository and test that version of this repository --- not using any published release artifacts.
+
+The end-to-end tests include ones written in `bash`, and these are the only documentation telling a user how to use the present version of this repository. Again, these tests do not use any published artifacts from a release of this repo.
+
+We will introduce another category of tests, _release tests_. These test a given release, using the published artifacts of that release. These tests will be executables that a contributor invokes, passing the identifier of the desired release on the command line. We will create a GitHub workflow that can be explicitly triggered and will apply the release tests (in the branch that the workflow is directed at) to the latest release in that branch. We will also create a GitHub workflow that is simply told which release to test.
+
+We will maintain a document that lists releases that pass our quality bar. The latest of those is thus the latest stable release. This document is updated in `main` as quality evaluations come in.
+
+We will maintain an examples document that shows users how to exercise a few scenarios using the latest stable release.
+
+## Policy
+
+We aim for all regular releases to be working. In order to do that, we have to make test releases and test them. The widely recognized pattern for doing that is to make "release candidates" (i.e., releases for testing purposes) `1.2.3-rc0`, `1.2.3-rc1`, `1.2.3-rc2`, and so on, while trying to get to a quality release `1.2.3`. Once one of them is judged to be of passing quality, we make a release without the `-rc<N>` suffix. Due to the self-reference in the KCM PostCreateHook, this will involve making a new commit.
+
+When reaching a new major or minor version, we make a release branch for it. That is, a branch named `release-$major.$minor`. At first this branch differs from `main` in one commit, the one that sets the self-reference in the KCM PostCreateHook. After that, the branches continue to evolve in the usual ways.
+
+### Deliberately feature-incomplete releases
+
+We plan a few deliberately feature-incomplete releases. They will be regular releases as far as the technology here is concerned. They will be announced only to selected users who acknowledge that they are getting something that is incomplete. In GitHub, these will be marked as "pre-releases". The status of these releases will be made clear in their documentation (which currently appears in [the release notes](release-notes.md).
+
+## Step-by-Step
+
+Making a new release requires a contributor to do the following things. Here `$version` is the semver identifier for the release (e.g., `1.2.3-rc2`).
+
+- Edit [the source for the KCM PCH](../../../config/postcreate-hooks/kubestellar.yaml) and update the tag in the reference to the KCM container image (it appears in the last object, a `Job`).
+
+- Make a new Git commit with that change.
+
+- Apply the Git tag `v$versin` to that new commit.
+
+- Push the commit and its tag to GitHub.
+
+After that, the GitHub workflow then creates and publishes the
+artifacts for that release, as discussed [above](#technology).
+
+## Future Process Development
+
+We intend to get rid of the self-reference in the KCM PCH, as follows. Define a Helm chart for installing the PCH. Update the release workflow to specialize that Helm chart, similarly to the specialization done for the KCM Helm chart.
+
+## Open questions
+
+What about the clusteradm container image?
+
+Playing nice with the website technology.
--- a/docs/content/v0.20/release.md
+++ b/docs/content/v0.20/release.md
@@ -24,7 +24,7 @@ We have the following limitations.
 - Thus, it is necessary to keep users clearly appraised of the quality (or status of evaluating the quality) of each release.
 - Because of the lack of self references, most user instructions (e.g., examples) and tests do not have concrete release identifiers in them; instead, the user has to chose and supply the release identifier. There can also be documentation of a specific past release (e.g., the latest stable release) that uses the literal identifier for that past release.
 - **PAY ATTENTION TO THIS ONE**: Because of the prohibition of self references, **Git will not contain the exact bytes of our Helm chart definitions**. Where a Helm chart states its own version or has a container image reference to an image built from the same release, the bytes in Git have a placeholder for that image's tag and the process of creating the published release artifacts fills in that placeholder. Think of this as being analogous to the linking done when building a binary executable file.
-- The design below fails to completely meet the goal of not putting self-references in files under Git control. The failure is in the KubeFlex PostCreateHook that installs the kubestellar-controller-manager (KCM), where the version of the container image for the KCM appears.
+- The design below **fails** to completely meet the goal of not putting self-references in files under Git control. The failure is in the KubeFlex PostCreateHook that installs the kubestellar-controller-manager (KCM), where the version of the container image for the KCM appears.
 
 ## Technology
 
@@ -78,6 +78,10 @@ artifacts for that release, as discussed [above](#technology).
 We intend to get rid of the self-reference in the KCM PCH, as follows. Define a Helm chart for installing the PCH. Update the release workflow to specialize that Helm chart, similarly to the specialization done for the KCM Helm chart.
 
 ## Open questions
+
+What to do about the dependency cycle between ks/ks and ks/ocm-transport-plugin?
+
+Exactly when does a new release branch diverge from `main`? What about cherry-picking between `main` and the latest (or also earlier?) release branch?
 
 What about the clusteradm container image?
 
--- a/docs/content/v0.20/release.md
+++ b/docs/content/v0.20/release.md
@@ -24,7 +24,7 @@ We have the following limitations.
 - Thus, it is necessary to keep users clearly appraised of the quality (or status of evaluating the quality) of each release.
 - Because of the lack of self references, most user instructions (e.g., examples) and tests do not have concrete release identifiers in them; instead, the user has to chose and supply the release identifier. There can also be documentation of a specific past release (e.g., the latest stable release) that uses the literal identifier for that past release.
 - **PAY ATTENTION TO THIS ONE**: Because of the prohibition of self references, **Git will not contain the exact bytes of our Helm chart definitions**. Where a Helm chart states its own version or has a container image reference to an image built from the same release, the bytes in Git have a placeholder for that image's tag and the process of creating the published release artifacts fills in that placeholder. Think of this as being analogous to the linking done when building a binary executable file.
-- The design below **fails** to completely meet the goal of not putting self-references in files under Git control. The failure is in the KubeFlex PostCreateHook that installs the kubestellar-controller-manager (KCM), where the version of the container image for the KCM appears.
+- The design below **fails** to completely meet the goal of not putting self-references in files under Git control. One failure is in the KubeFlex PostCreateHook that installs the kubestellar-controller-manager (KCM), where the version of the container image for the KCM appears. Another failure is in the examples document, which also holds references to its own release.
 
 ## Technology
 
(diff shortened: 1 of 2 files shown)


docs/content/v0.20/release.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Goals and limitations
+
+The release process has the following goals.
+
+- A release is identified using [semantic versioning](semver.org). This means that the associated semantics are followed, in terms of what sort of changes to the repo require what sort of changes to the release identifier.
+- A user can pick up and use a given existing release without being perturbed by on-going contributor work. A release is an immutable thing.
+- A release with a given semver identifier is built from a commit of this Git repository tagged with a tag whose name is "v" followed by the release identifier.
+- There is a concept of a release branch, with a name following the pattern `release-$major.$minor`. This allows contributors to do maintenance work on older release branches. This also allows a user to track work along such a line.
+- The contents of `main` always work. This includes passing CI tests. This includes documentation being accurate. We allow point-in-time specific documentation, such as a document that says "Here is how to use release 1.2.3" --- which would refer to a release made in the past. We do not require the documentation in `main` to document all releases.
```
pdettori, line 16: Would "point-in-time specific documentation" be created and tagged together with the release ?

docs/content/v0.20/release.md
```diff
@@ -0,0 +1,60 @@
...
+- A git tag is immutable. Once associated with a given Git commit, that association is not changed later.
+- We do not put self-references into Git. For example, making release `1.2.3` doe not require changing any file in Git to have the string `1.2.3` in it.
```
pdettori, line 18: What about docs that reference that release (e.g., "Here is how to use release 1.2.3" above)

docs/content/v0.20/release.md
```diff
@@ -0,0 +1,60 @@
...
+- Because of the prohibition of self references, Git will not contain the exact bytes of our Helm chart definitions. Where a Helm chart states its own version or supplies a default value for a container image reference to an image built from the same release, the bytes in Git have a placeholder for that image's tag and the process of creating the published release artifacts fills in that placeholder. Think of this as being analogous to the linking done when building a binary executable file.
+
+## Technology
+
+There will be a GitHub workflow that creates the published artifacts for each Git tag whose name starts with "v". The rest of the tag name is required to be a semver release identifier. For each tag `v$version` the following published artifacts will be created.
+
+- The container image for the kubestellar-controller-manager (KCM), at `ghcr.io/kubestellar/kubestellar-controller-manager` (or should it be `ghcr.io/kubestellar/kubestellar/controller-manager` from the ks/ks repo? or just `ghcr.io/kubestellar/controller-manager`?). Image tag will be `$version`.
```
pdettori, line 32: I realize now that our naming scheme for packages published to ghcr.io is quite inconsistent. Some package has the scheme `ghcr.io/<org>/<repo>/<name>` and some  `ghcr.io/<org>/<name>`. It looks `ghcr.io/<org>/<repo>/<name>` may be better as we could have for example `ghcr.io/kubestellar/kubestellar/controller-manager` and `ghcr.io/kubestellar/ocm-status-addon/controller-manager`.

docs/content/v0.20/release.md
```diff
@@ -0,0 +1,60 @@
...
+- The Helm chart (for installing the KCM for a WDS), at `ghcr.io/kubestellar/kubestellar-controller-manager-chart` with version `$version` and Helm "appVersion" `$version`. The `values.yaml` of this chart version has a default value for the image reference to the KCM and that value is `ghcr.io/kubestellar/kubestellar-controller-manager:$version`.
+
```
pdettori, line 34: Not so. The value of the image reference is not exposed in values.yaml. That was a deliberate choice (which we can discuss) as the yaml templates for the chart has been tested only with a specific image reference, and in general we would not want users to try to provide a different image version to a chart that was not tested with that version.

docs/content/v0.20/release.md
```diff
@@ -0,0 +1,60 @@
...
+## Testing and Instructions
+
+The unit tests (of which we have almost none right now), integration tests (of which we have none yet), and end-to-end (E2E) tests in this repository are run in the context of a local copy of this repository and test that version of this repository --- not using any published release artifacts.
+
+The end-to-end tests include ones written in `bash`, and these are the only documentation telling a user how to use the present version of this repository. Again, these tests do not use any published artifacts from a release of this repo.
```
pdettori, line 39: This is a point in time, right ? And we have also several examples, although we make the disclaimer that they are specific to a particular version and not maintained beyond that.

docs/content/v0.20/release.md
```diff
@@ -0,0 +1,60 @@
...
+
+We will introduce another category of tests, _release tests_. These test a given release, using the published artifacts of that release. These tests will be executables that a contributor invokes, passing the identifier of the desired release on the command line. We will create a GitHub workflow that can be explicitly triggered and will apply the release tests (in the branch that the workflow is directed at) to the latest release in that branch. We will also create a GitHub workflow that is simply told which release to test.
+
+We will maintain a document that lists releases that pass our quality bar. The latest of those is thus the latest stable release. This document is updated in `main` as quality evaluations come in.
+
+We will maintain an examples document that shows users how to exercise a few scenarios using the latest stable release.
```
pdettori, line 45: Is the idea to update the current examples doc for that ?

docs/content/v0.20/release.md
```diff
@@ -0,0 +1,60 @@
...
+
+When reaching a new major or minor version, we make a release branch for it. That is, a branch named `release-$major.$minor`. At first this branch points to the same commit as `main`, and we make release `$major.$minor.0-rc0` from it. After that work can continue, if necessary, on this release branch while work on `main` proceeds to develop the next minor (or major) version.
+
+## Open questions
+
+What to do about a deliberately feature-incomplete release (development snapshot)? Would we be making releases identified `0.20.0-milestone.1` (with `0.20.0-milestone.1.rc.0`, `0.20.0-milestone.1.rc.1` and so on for testing) and `0.20.0-milestone.2` (also with its own test releases) before getting to full-featured rlease `0.20.0`? Or would we simply decleare releases `0.20.0` and `0.21.0` to be deliberately feature-incomplete and release `0.22.0` would be the full-featured release?
+
```
pdettori, line 56: I would prefer the latter.

docs/content/v0.20/release.md
```diff
@@ -0,0 +1,60 @@
...
+What about the clusteradm container image?
+
+Do we publish any PostCreateHook(s)? How does a user use it(them)? I can think of two options for the KCM-for-WDS PCH, neither of which is great. One is simpler but a little more tedious for users: delete the PostCreateHook that installs the KCM for a WDS, instead ask the user to install a Helm chart per WDS. The other option piles on more technology: define a Helm chart that the deployer installs once in the host cluster to define the KCM-for-WDS PCH (and remember: what this thing does is install the KCM-for-WDS Helm chart).
+
```
pdettori, line 60: PostCreateHooks for kubestellar use are now part of the KubeStellar git repo, so they would be tagged with the release. Why can't we just require to do the step

```
kubectl --context kind-kubeflex  apply -f https://raw.githubusercontent.com/kubestellar/kubestellar/<release-tag>/config/postcreate-hooks/kubestellar.yaml
kubectl --context kind-kubeflex  apply -f https://raw.githubusercontent.com/kubestellar/kubestellar/<release-tag>/config/postcreate-hooks/ocm.yaml
```

and then the user can just use `kflex create wds<n> -p kubeflex` how many times he/she wants to create the WDS and deploy the controller ?

docs/content/v0.20/release.md
nirrozenbaum: IMO the names should be following the pattern `ghcr.io/kubestellar/{package_name}`. 
I do not see the point in duplicating kubestellar identifier.

I don't think we should re-invent the wheel. 
controller manager is an abstract term. from a quick search of CNCF projects images I haven't found any project that uses such an abstract name for image (I've checked 5-6 projects, not the whole list). 
We should use specific package names instead, e.g., `binding-controller`, `ocm-transport-plugin-controller`, `ocm-status-addon-controller`, etc. 
there are many examples for how CNCF projects are naming their images, most of them are using the pattern of `{Registry}/{Org}/{Package}`. 
 
some example for such are:
`quay.io/coreos/etcd-operator`
`quay.io/open-cluster-management/registration-operator`.
`openpolicyagent/opa-kube-scheduler` (in docker hub)
and more..

docs/content/v0.20/release.md
effi-ofer: I think this section will need to be updated in the next few weeks.  We have an outstanding e2e tests PR that is not based on bash.

docs/content/v0.20/release.md
effi-ofer, line 34: Should this section include a list of all the images/artifacts produced by the kubestellar organization, including kubeflex, status-addon, etc, and how they should be synchronized?

docs/content/v0.20/release.md
```diff
@@ -0,0 +1,60 @@
+# Making KubeStellar Releases
+
+This document defines how releases of the KubeStellar repository are made.
```
nirrozenbaum, line 3: do we want to create a release of KS repos at the same time and with the same identifier? 
like it is done in kubernetes?

I think that's a little confusing when user needs to use different versions.
for example: 
KS controllers version `v0.2.1`, kubeflex version `v0.5.0`, ocm-status-addon `0.4.0` and ocm-transport-plugin version `v0.1.0`.

instead I would expect that all versions should be aligned. than we create release notes for a specific version - 
for example - 
when releasing v0.6.0 let's assume we put in the release notes that status conflict is solved. 
if users want to use KS version where status conflict is solved, one should use v0.6.0 for KS controllers and for status-addon (kubeflex is a different story cause it's a space management framework, not really related to KS).

docs/content/v0.20/release.md
pdettori: @nirrozenbaum there is more than one controller within the kubestellar controller manager, that is why is a controller manager. An example of that is the kube-controller-manager in Kubernetes. Just calling it binding controller may be inaccurate (it includes also status controller). Using `ghcr.io/kubestellar/{package-name}` is fine with me as long as we have distinct and meaningful names. Possible choices are `ghcr.io/kubestellar/ks-controller-manager`, or `ghcr.io/kubestellar/kubestellar-controller-manager`. If we don't include the name of the repo in the path (`ghcr.io/kubestellar/<repo-name>/`) I think we should at least have the repo name or some abbreviated form of the repo name in the package name.

docs/content/v0.20/release.md
ezrasilvera: I'm not sure this is needed.  The fact that it's in the same ORG is not critical. I see the other repos as external dependencies similar to KIND, OCM CLI, etc..  
The main thing is to avoid circular dependency.  Other than that it's a regular A depends on B, and it's a common procedure how we update A to use a new version of B.

pdettori: I would definitely leave KubeFlex to have its independent release cycle as it is going to be decoupled anyway. As for the other repos, I thought that once you define the interface for a "pluggable transport" that transport adapter (which I assume will include status in the future) can evolve independently as long as the interface stays the same. What about other possible transports in the future? Do they all have to release with the same release cycle? It seems to defeat the purpose of having created a pluggable transport in the first place.

docs/content/v0.20/release.md
ezrasilvera: @nirrozenbaum why not including the repo ?  I think we should treat our repos as independent, and therefore the repo "owner" can use any package name. In that case we must have repo in order to avoid collisions

docs/content/v0.20/release.md
ezrasilvera: just as FYI We also plan to have tests that will use artifacts instead of building the code on the fly , but that's indeed a future statement.

docs/content/v0.20/release.md
ezrasilvera: I think the examples already point to a "release" (or artifact) , it's only the e2e tests that use the current code

docs/content/v0.20/release.md
```diff
@@ -0,0 +1,60 @@
...
+clues about the problem.
+
+## Goals and limitations
+
+The release process has the following goals.
+
+- A release is identified using [semantic versioning](semver.org). This means that the associated semantics are followed, in terms of what sort of changes to the repo require what sort of changes to the release identifier.
+- A user can pick up and use a given existing release without being perturbed by on-going contributor work. A release is an immutable thing.
+- A release with a given semver identifier is built from a commit of this Git repository tagged with a tag whose name is "v" followed by the release identifier.
+- There is a concept of a release branch, with a name following the pattern `release-$major.$minor`. This allows contributors to do maintenance work on older release branches. This also allows a user to track work along such a line.
```
ezrasilvera, line 15: is the branch name different than the tag ?

docs/content/v0.20/release.md
```diff
@@ -0,0 +1,60 @@
...
+- The contents of `main` always work. This includes passing CI tests. This includes documentation being accurate. We allow point-in-time specific documentation, such as a document that says "Here is how to use release 1.2.3" --- which would refer to a release made in the past. We do not require the documentation in `main` to document all releases.
+- A git tag is immutable. Once associated with a given Git commit, that association is not changed later.
+- We do not put self-references into Git. For example, making release `1.2.3` doe not require changing any file in Git to have the string `1.2.3` in it.
+
+We have the following limitations.
+
+- The only way to publish artifacts (broadly construed, not (necessarily) GitHub "release artifacts") is to make a release.
```
ezrasilvera, line 22: This seems a technical limitation  and it also add a lot of complexity to testing and naming.

pdettori: I assume you the technical limitation is around GoReleaser, right ? As we have seen, we should avoid manual publishing to ghcr.io because of the ownership/permissions issues. If we do not publish artifacts for testing via GoReleaser, we can still create GitHub workflows just for that purpose. We can still trigger the workflows based on a tag, but perhaps use a different pattern from that used for a release (we would need also to change the pattern that trigger the workflow invoking goreleaser)

docs/content/v0.20/release.md
MikeSpreitzer: Think of the existing `examples.md`. It reports on usage of a specific "snapshot". Being in the git repo, it is part of a commit, we have no choice about that. It is a commit that gets tagged in git, nothing less.

docs/content/v0.20/release.md
MikeSpreitzer: This is on the same point as your other comment.

docs/content/v0.20/release.md
MikeSpreitzer: @pdettori : sorry I was not explicit about this. Yes, I realize that this is a change from the current YAML. Now that you mention it, we can keep it a constant. Still, a constant that has to be injected in the process of releasing the chart.

pdettori: No. Not manually. See for example https://github.com/kubestellar/ocm-status-addon/blob/d66eeb5fc9624011cf86268a1ac475600979ca8a/.github/workflows/goreleaser.yml#L60-L64
`make chart` takes the image reference as a input parameter and sets it in the chart before publishing.

docs/content/v0.20/release.md
MikeSpreitzer: @ezrasilvera : I addressed that in what I called "release tests".

@pdettori : you effectively asked about `examples.md` above in two other comments. My point in this part of the draft text is that we need a way to tell contributors how to exercise the contents of `main`. I specifically want `bash` scripts because contributors can read them and they are very close to what a contributor would do. If these tests were written in `go` it would be much harder for a contributor to translate that to the shell commands to execute. I am trying to not duplicate material here; duplication would introduce a requirement for checking consistency.

docs/content/v0.20/release.md
MikeSpreitzer: Because a specific version is needed in such commands. But we want no self-references.

docs/content/v0.20/release.md
MikeSpreitzer: Normally separate repos are versioned separately.

docs/content/v0.20/release.md
MikeSpreitzer: @ezrasilvera : a release branch is specific only to major.minor; an individual release is one of many in a release branch.

docs/content/v0.20/release.md
MikeSpreitzer: It is not just because of goreleaser. You would do better to ask if it is because of the authorization problem at ghcr.io. But the answer is the "release tests". How else would we test a release without making a release?

docs/content/v0.20/release.md
nirrozenbaum: you are all correct. I haven’t thought about it to the end.
no need to align versions.

docs/content/v0.20/release.md
nirrozenbaum: @pdettori yes, using meaningful names was my intention. 

there is no problem with any of the approaches, I just stated that:
(a) no need for duplicate identifier
(b) the CNCF projects I checked are using this pattern. 

@ezrasilvera I actually think package names in this org should be validated, and not “repo owner can choose any name”. this is important for consistency.

eventually the difference will be:
ghcr.io/kubestellar/kubestellar-controller-manager Vs ghcr.io/kubestellar/kubestellar/controller-manager. 
which is better? 
there is no one correct answer but the other CNCF projects are mostly using the former.

docs/content/v0.20/release.md
MikeSpreitzer: Right. This design talks about the way the chart definition is modified in the process of creating the published artifact, and the exact published bytes are never in git.

docs/content/v0.20/release.md
MikeSpreitzer: I think the latter would work better with the website automation that we have. It codifies a concept of release branch, at least as far as website matters are concerned.

docs/content/v0.20/release.md
MikeSpreitzer: For repos ks/ks depends on, "no". I think we can just be ordinary here, it repo takes care of itself.

Referring to https://github.com/kubestellar/kubestellar/blob/main/docs/content/v0.20/packaging.md#outline-of-publishing, and adding in the dashed where ks/ocm-status-addon depends on ks/ks (look in the instructions in ks/ocm-status-addon), that leaves two repos of concern: ks/ocm-transport-plugin and ks/ocm-status-addon. In some sense there is actually a cycle of dependency here, because you can not use ks/ks without also using those repos.

I think it would be cleanest to split ks/ks into two parts. Think of it as two layers. The lower layer holds the bulk of the functionality and does not depend on ks/ocm-anything. The upper layer brings it all together, depending on ks/ks, ks/ocm-status-addon, and ks/ocm-transport-plugin.

docs/content/v0.20/release.md
MikeSpreitzer: I also commented in https://github.com/kubestellar/kubestellar/issues/1612#issuecomment-1944302541

docs/content/v0.20/release.md
MikeSpreitzer: Note that the question of how a use refers to an OCI image is connected to a bit of GitHub structure, whether a package is "from" a repo or "from" the org.

It makes more sense to me for each repo to "own" its GitHub packages. I could live with packages owned by the org (necessarily with coordinated names).

nirrozenbaum: not sure I got the point about the listing in the browser UI. I used the pattern I suggested and I see in the packages that my package is from my repo. 
I've attached screenshot of packages list and if I go into that package I also see it's linked to my repo.

can you please explain the point you were mentioning? 

image

MikeSpreitzer: @nirrozenbaum : my point is about the difference between the pattern you followed and the pattern followed by, e.g., the kubestellar-operator-chart. See how they are described differently?  And on the browser page for https://github.com/kubestellar/kubestellar/, you see that it says that repo has no packages.

MikeSpreitzer:
The force-push to 300dd23e6 attempts to more clearly answer @eaepstein's question about the Helm chart source.

docs/content/v0.20/release.md
nirrozenbaum: is this a limitation? 
I'm almost sure if you go into the package settings there is an option to link the package to the repo (should be done only once and will be kept for all releases)

nirrozenbaum: https://docs.github.com/en/packages/learn-github-packages/connecting-a-repository-to-a-package#connecting-a-repository-to-an-organization-scoped-package-on-github

MikeSpreitzer: Thanks for the pointer. BTW, I did not know any way to tell GitHub that a particular package is "from" or "in" a particular repo. Are there any other ways? Maybe part of the process of publishing the package? @nirrozenbaum , @pdettori : how did you do it, in the cases where you did?

docs/content/v0.20/release.md
MikeSpreitzer: Yes, I think that would work.

docs/content/v0.20/release.md
nirrozenbaum: the initial image I created was done manually and wasn’t linked. I linked it to the repo as described in the pointer doc.
later I removed the manually created images and created through go releaser and link was already there. 

I’m sure there is way to have as part of the automated process. maybe it is already there, I’m not sure cause in my case I created the link manually before.

docs/content/v0.20/release.md
MikeSpreitzer: @pdettori : So the PostCreateHook source in GitHub in release `1.2.3` has the string "1.2.3" in it? That is a self-reference. It means that each commit has to refer to a release that does not yet exist when the commit is made, and each commit can be in only one release.

pdettori: Understood. One solution is what you outlined with publishing a helm chart to install the two PCHs where we use the same approach to tagging used with the other helm charts to avoid the self referencing.

MikeSpreitzer: Yes, that is what I was trying to say in the second approach listed.

docs/content/v0.20/release.md
MikeSpreitzer: @pdettori: put another way: we need a process to publish artifacts for a release. I am trying to not multiply entities needlessly.

MikeSpreitzer:
The recent force-pushes make revisions aligned with the review discussion above.

docs/content/v0.20/release.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Deliberately feature-incomplete releases
+
+We plan a few deliberately feature-incomplete releases. They will be regular releases as far as the technology here is concerned. They will be announced only to selected users who acknowledge that they are getting something that is incomplete. In GitHub, these will be marked as "pre-releases". The status of these releases will be made clear in their documentation (which currently appears in [the release notes](release-notes.md).
+
+## Step-by-Step
+
+Making a new release requires a contributor to do the following things. Here `$version` is the semver identifier for the release (e.g., `1.2.3-rc2`).
+
+- Edit [the source for the KCM PCH](../../../config/postcreate-hooks/kubestellar.yaml) and update the tag in the reference to the KCM container image (it appears in the last object, a `Job`).
```
ezrasilvera, line 65: 1. What is "KCM" ? 
2. Maybe more specific pointer, e.g., "Update the --version  argument"

MikeSpreitzer: 1. Is answered twice in this document already

MikeSpreitzer:
The force-push to b31edf326 splits responsibility for identifying the latest stable release from the responsibility for describing how to use a given release. This leaves examples.md describing how to use its own release, creating another self-reference problem.

docs/content/v0.20/release.md
pdettori: I am in the process of testing a solution for https://github.com/kubestellar/kubeflex/issues/215, which would eliminate the need to create an additional helm chart.

PR #1761, opened 2024-02-14
✨ Second step to rename two Kinds in our API

### LINKED ISSUE #1729: feature: Rename two Kinds in our API

waltforme:
## Summary
This PR takes step 2/2 of https://github.com/kubestellar/kubestellar/issues/1729.

## Related issue(s)

Closes #1729


config/crd/bases/control.kubestellar.io_bindings.yaml
```diff
@@ -20,9 +20,9 @@ spec:
   - name: v1alpha1
     schema:
       openAPIV3Schema:
-        description: Binding is mapped 1:1 to a single BindingPolicy object. The decision
-          resource reflects the resolution of the BindingPolicy's selectors, and explicitly
-          reflects which objects should go to what destinations.
+        description: Binding is mapped 1:1 to a single BindingPolicy object. It reflects
```
MikeSpreitzer, line 23: In English, "it" refers to the most recent noun phrase. In the current revision, that would be that BindingPolicy object, which is not what you are trying to say.

api/control/v1alpha1/types.go
```diff
@@ -169,7 +169,7 @@ type BindingPolicyList struct {
 }
 
 // Binding is mapped 1:1 to a single BindingPolicy object.
-// The decision resource reflects the resolution of the BindingPolicy's selectors,
+// It reflects the resolution of the BindingPolicy's selectors,
```
MikeSpreitzer, line 172: "it" again.

config/crd/bases/control.kubestellar.io_bindings.yaml
waltforme: I didn't find an authoritative source where pronoun references are clearly defined. However ChatGPT agrees with my feeling:
image

But I will also respect a native speaker's suggestion and follow up in another PR.

Issue #1734, opened 2024-02-08
feature: settle debate about the semantics of singleton status return


MikeSpreitzer:
### Feature Description

The semantics now are different than they were in release-0.14. The semantics now require something (you might call it a "scheduler", whatever it is) below the Placement (BindingPolicy) interface to make choices about distribution. This is a camel's nose in the tent that not every one is happy about.

### Proposed Solution

Discuss and reach agreement.

### Want to contribute?

- [ ] I would like to work on this issue.

ezrasilvera:
Let's use this issue for proposals.  The discussion is is on the definition and not on how we can enforce it. 
My proposal is:
**Singleton**:
Singleton is defined as flag in the placement and the interpretation is  that each objects included in this placement should be deployed into a single WEC.  If there are two different placements with overlapping object(s) that are deployed to different WECs , this violates the singleton pattern. 
In practice the singleton is controlled on the deployed object level by using a label on the objects (that was generated automatically) 

**Singleton status**
Status of an singleton object is returned as a WorkStatus object and then copied into the original object in the WDS.  
Current State: if there was a violation of the singleton pattern and such an object was deployed into two WECs , the status addon on each WEC will generate a WorkStatus , resulting in multiple WorkStatuse objects. The controller today copy one of these statuses into the object in WDS. This behavior should be changed. Either enforcing the singleton so there are only a single object instance  with such a label. or produce an error and don't update the WDS object.

ezrasilvera:
cc: @nirrozenbaum @pdettori @MikeSpreitzer

nirrozenbaum:
I suggest to leave aside the current status cause once we agree on the definition we can do the required work to implement it. 

My thinking about this topic is a bit different. 

**_observation_** - singleton status is a private case of status summarization. explanation:
if we had status summarization feature, would it be valid to define distribution to multiple clusters (e.g, `cluster1`, `cluster2`, .. `clusterN`) and define a policy to collect full status only from `cluster1`? 
sounds to me like a totally valid scenario.
for the long term, we need to consider how would we settle this debate if we had status summarization feature since this is where we're heading. in the long term I think we all agree we want status summarization in KubeStellar.

considering the above, I do not think we should couple distribution with status collection. these are two **DIFFERENT and TOTALLY SEPARATED** features of KubeStellar.

so my proposal is 
[...]

Issue #1754, opened 2024-02-13
bug: BindingPolicy is not explicit about scope-creating API objects


MikeSpreitzer:
### Describe the bug

In Kubernetes API machinery there are three kinds of API objects that create some sort of "scope" in which other objects exist. The current BindingPolicy API is not explicit, but should be, about which (if any) of these scope-defining API objects are implicitly selected.

1. A `Namespace` object defines a namesapce.
2. A `CustomResourceDefinition` object defines a Kind/resource.
3. An `APIService` object imports a whole API group.

### Steps To Reproduce

See above.

### Expected Behavior

A reader of the API knows what to expect.

PR #1765, opened 2024-02-15
✨ Switch to using pluggable transport (disables old transport-in-placement code)

### LINKED ISSUE #1686: feature: Document pluggable-transport architecture
### Feature Description

need to document pluggable-transport architecture and APIs.

### Proposed Solution

document.

### Want to contribute?

### Additional Context

_No response_

### LINKED ISSUE #1695: bug: placement controller may do concurrent work on one wrapped object
### Describe the bug

From review of the architecture document, I noticed that the placement controller may have concurrent goroutines working on the same ManifestWork --- and not necessarily with the same versions of their inputs. Also, the controller makes too many assumptions about the current state of a given ManifestWork object rather than reading it from the apiserver (or a local cache of that).

### Steps To Reproduce

A reliable procedure to demonstrate the problem would require more fine-grained and precise control over timing than the technology offers.

### Expected Behavior

Only one goroutine should work on a given output object at a time, using the latest inputs.

### Additional Context

_No response_

### LINKED ISSUE #1730: feature: disable (old) transport-in-placement logic
### Feature Description

After #1681 was merged, now is the time to disable all old transport object-wrapping paths and the logic around it.
As a side-job in this task, I will also clean up unrelated, unused code.

This updates the placement controller's responsibilities to the single one of maintaining PlacementDecision (soon Binding) objects.

This task is expected to delete 600-700 lines of code.

### Proposed Solution

Update code

### Want to contribute?

### Additional Context

_No response_

### LINKED ISSUE #1731: feature: finalize documentation, build and testing for transport controller
### Feature Description

in order to run transport controller it should use a specific transport plugin implementation.
- we need to document how to run the ocm based transport controller and what it is (how is it working, what it does), 
- how to build the images of ocm based transport controller. 
- change all flows (including tests) to use the transport controller 

### Proposed Solution

do the above

### Want to contribute?

### Additional Context

_No response_

vMaroon:
## Summary
Deleted transport-in-placement out of binding-policy controller, and have done the required to compensate for the deleted functionality, of which the most significant changes are:
- Resolutions now also track whether a binding requires singleton reported state for selected object
  - At the moment, this assumes correctness. This assumption is hidden behind the resolver's new function.
- Object reconciliation guarantees that a relevant label is present if the object is selected by a singleton-status reporting binding, and not present otherwise.
- BindingPolicy deletion queues up only selected objects in order to trigger the above.

**After merging this PR, pluggable transport will be working end to end.**

## Related issue(s)

Fixes #1730 
Fixes #1686
Fixes #1695 
Fixes #1731


vMaroon:
Note: e2e tests were executed locally.

Multi-cluster test works e2e, singleton status test's binding-transport parts work correctly, pending @andreyod's work on status-addon.

api/control/v1alpha1/types.go
```diff
@@ -49,8 +49,8 @@ type BindingPolicySpec struct {
 	// WantSingletonReportedState indicates that (a) the number of selected locations is intended
 	// to be 1 and (b) the reported state of each downsynced object should be returned back to
 	// the object in this space.
-	// When multiple BindingPolicy objects match the same workload object,
-	// the OR of these booleans rules.
+	// For this field to take effect, it is required that no other binding policy with this field
+	// set to true selects a shared object.
```
MikeSpreitzer, line 53: "a" here is not very clear. I think perhaps you mean something like the following.

> .. no other BindingPolicy with `WantSingletonReportedState == true` selects any of the same workload objects.

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -113,6 +121,19 @@ func (resolution *bindingPolicyResolution) setDestinations(destinations sets.Set
 	resolution.destinations = destinations
 }
 
+// getObjectKeys returns the Keys of the objects in the resolution.
+func (resolution *bindingPolicyResolution) getObjectKeys() []*util.Key {
```
MikeSpreitzer, line 125: Note to self: why not `Set[util.Key]` ?

test/e2e/multi-cluster-deployment/use-kubestellar.sh
```diff
@@ -73,9 +73,9 @@ EOF
 :
 : -------------------------------------------------------------------------
 : "Verify that manifestworks wrapping the objects have been created in the mailbox namespaces"
-: Expect twice: one header line, one for the nginx namespace, one for the nginx deployment, one for the status agent Deployment
+: Expect twice: one header line, one for the nginx-bindingpolicy Binding\'s workload, one for the status agent Deployment
```
MikeSpreitzer, line 76: (A) "workload" is vague here, particularly while issue #1754 is not resolved.

(B) Why should the namespace no longer appear in these listings?

test/e2e/singleton-status/use-kubestellar.sh
```diff
@@ -73,9 +73,9 @@ EOF
 :
 : -------------------------------------------------------------------------
 : "Verify that manifestworks wrapping the objects have been created in the mailbox namespaces"
-: Expect: one header line, one for the nginx namespace, one for the nginx deployment, one for the status agent Deployment
+: Expect: one header line, one for the nginx-singleton-bindingpolicy Binding\'s workload, one for the status agent Deployment
```
MikeSpreitzer, line 76: (A) "workload" is vague here, particularly while issue #1754 is not resolved.

(B) Why should the namespace no longer appear in these listings?

pkg/binding/bindingpolicy.go
```diff
@@ -81,24 +76,45 @@ func (c *Controller) handleBindingPolicy(ctx context.Context, obj runtime.Object
...
 
+	// we delete finalizer only after cleaning up in the case above
 	if err := c.handleBindingPolicyFinalizer(ctx, bindingPolicy); err != nil {
-		return err
+		return fmt.Errorf("failed to handle finalizer for BindingPolicy %s: %w", bindingPolicy.Name, err)
 	}
 
-	return c.cleanUpObjectsNoLongerMatching(bindingPolicy)
+	c.bindingPolicyResolver.DeleteResolution(bindingPolicy.GetName())
+	c.logger.Info("Deleted resolution for bindingpolic", "name", bindingPolicy.Name)
```
MikeSpreitzer, line 115: The "y" is not silent.

pkg/binding/bindingpolicy.go
```diff
@@ -81,24 +76,45 @@ func (c *Controller) handleBindingPolicy(ctx context.Context, obj runtime.Object
...
 	}
 
+	// we delete finalizer only after cleaning up in the case above
 	if err := c.handleBindingPolicyFinalizer(ctx, bindingPolicy); err != nil {
-		return err
+		return fmt.Errorf("failed to handle finalizer for BindingPolicy %s: %w", bindingPolicy.Name, err)
 	}
 
-	return c.cleanUpObjectsNoLongerMatching(bindingPolicy)
+	c.bindingPolicyResolver.DeleteResolution(bindingPolicy.GetName())
```
MikeSpreitzer, line 114: Note to self: why was this not needed before?

pkg/binding/bindingpolicy.go
```diff
@@ -113,16 +129,33 @@ func (c *Controller) requeueForBindingPolicyChanges() error {
...
+		obj, err := c.getObjectFromKey(*key)
+		if err != nil {
+			return fmt.Errorf("failed to get object from key: %w", err)
+		}
+
+		objs = append(objs, obj)
+	}
+
+	for _, obj := range objs {
+		c.enqueueObject(obj, true)
```
MikeSpreitzer, line 155: Why two separate loops over the same set of objects? Why not fuse the two loops?

pkg/binding/controller.go
```diff
@@ -222,9 +221,6 @@ func (c *Controller) run(ctx context.Context, workers int) error {
 						c.handleObject(new)
 					},
 					DeleteFunc: func(obj interface{}) {
-						if shouldSkipDelete(obj) {
-							return
-						}
 						c.handleObject(obj)
```
MikeSpreitzer, line 224: What happens if obj is a `cache.DeletedFinalStateUnknown` ?

pkg/binding/bindingpolicy-resolution.go
vMaroon: That indeed should happen, along with the keys rework.

For the moment this is very convenient (and aligned with the resolver).

pkg/binding/bindingpolicy.go
vMaroon: It was moved, not added (in this func).

pkg/binding/bindingpolicy.go
vMaroon: Because in that case, it is probable that a failing object will keep failing for a number of retries. All objects iterated over before it would keep getting queued up.

Therefore I chose to enqueue all workload objects or none.

pkg/binding/controller.go
```diff
@@ -509,38 +503,7 @@ func (c *Controller) reconcile(ctx context.Context, key util.Key) error {
...
-		c.logger.Info("Deleting", "object", util.RefToRuntimeObj(obj),
-			"from clusters", clusterSet)
-		deleteObjectOnManagedClusters(c.logger, c.ocmClient, *key.DeletedObject, clusterSet)
-		return nil
-	}
-
-	c.logger.Info("Delivering", "object", util.RefToRuntimeObj(obj),
-		"to clusters", clusterSet)
-	return c.deliverObjectToManagedClusters(obj, clusterSet, managedByBindingPolicies, singletonStatus)
+	return c.updateResolutions(ctx, obj)
```
MikeSpreitzer, line 506: Why is it right to remove the error wrapping?

MikeSpreitzer:
This PR makes the architecture document a liar. Please include the corresponding updates to the architecture document.

pkg/binding/selectors.go
```diff
@@ -117,17 +74,104 @@ func (c *Controller) updateDecisions(obj runtime.Object) error {
...
+		return nil
+	}
+
+	// label found, should delete it and update object
+	delete(labels, util.BindingPolicyLabelSingletonStatusKey)
+	unstructuredObj.SetLabels(labels)
+
+	gvr, found := c.gvkGvrMapper.GetGvr(unstructuredObj.GetObjectKind().GroupVersionKind())
+	if !found {
+		return fmt.Errorf("failed to get GVR for object %v", util.RefToRuntimeObj(obj))
```
MikeSpreitzer, line 134: Not necessarily an error, this could happen if this bit of code runs after the object and its GVR have been deleted.

pkg/binding/selectors.go
```diff
@@ -117,17 +74,104 @@ func (c *Controller) updateDecisions(obj runtime.Object) error {
...
+	}
+
+	if unstructuredObj.GetNamespace() == metav1.NamespaceNone {
+		_, err := c.dynamicClient.Resource(gvr).Update(ctx, unstructuredObj, metav1.UpdateOptions{})
+		return err
+	}
+
+	_, err := c.dynamicClient.Resource(gvr).Namespace(unstructuredObj.GetNamespace()).Update(ctx, unstructuredObj, metav1.UpdateOptions{})
+	return err
```
MikeSpreitzer, line 143: FYI, there is some common code here that runs before, and some common code here that runs after, the choice of whether to invoke `.Namespace(..)` or not.

pkg/binding/selectors.go
```diff
@@ -117,17 +74,104 @@ func (c *Controller) updateDecisions(obj runtime.Object) error {
...
+	unstructuredObj.SetLabels(labels)
+
+	gvr, found := c.gvkGvrMapper.GetGvr(unstructuredObj.GetObjectKind().GroupVersionKind())
+	if !found {
+		return fmt.Errorf("failed to get GVR for object %v", util.RefToRuntimeObj(obj))
+	}
+
+	if unstructuredObj.GetNamespace() == metav1.NamespaceNone {
+		_, err := c.dynamicClient.Resource(gvr).Update(ctx, unstructuredObj, metav1.UpdateOptions{})
+		return err
```
MikeSpreitzer, line 139: Not an error if the object is deleted by now.

pkg/binding/selectors.go
```diff
@@ -117,17 +74,104 @@ func (c *Controller) updateDecisions(obj runtime.Object) error {
...
+
+	_, err := c.dynamicClient.Resource(gvr).Namespace(unstructuredObj.GetNamespace()).Update(ctx, unstructuredObj, metav1.UpdateOptions{})
+	return err
+}
+
+func (c *Controller) addSingletonLabel(ctx context.Context, obj runtime.Object, bindingPolicyName string) error {
+	unstructuredObj, ok := obj.(*unstructured.Unstructured)
+	if !ok {
+		return fmt.Errorf("failed to convert runtime.Object to unstructured.Unstructured")
+	}
```
MikeSpreitzer, line 150: FYI: We know at development time that `ok` will always be `true`, right? If so then it would be legitimate to replace this with simply

```suggestion
	unstructuredObj := obj.(*unstructured.Unstructured)
```

pkg/binding/selectors.go
```diff
@@ -117,17 +74,104 @@ func (c *Controller) updateDecisions(obj runtime.Object) error {
...
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+
+	// skip if label already exists
+	if _, found := labels[util.BindingPolicyLabelSingletonStatusKey]; found {
+		return nil
+	}
+
+	labels[util.BindingPolicyLabelSingletonStatusKey] = bindingPolicyName
```
MikeSpreitzer, line 162: FYI: if the label is already set to a different policy's name, then I would issue a warning message here and an `Event`.

pkg/binding/selectors.go
```diff
@@ -117,17 +74,104 @@ func (c *Controller) updateDecisions(obj runtime.Object) error {
...
+	unstructuredObj.SetLabels(labels)
+
+	gvr, found := c.gvkGvrMapper.GetGvr(unstructuredObj.GetObjectKind().GroupVersionKind())
+	if !found {
+		return fmt.Errorf("failed to get GVR for object %v", util.RefToRuntimeObj(obj))
+	}
+
+	if unstructuredObj.GetNamespace() == metav1.NamespaceNone {
+		_, err := c.dynamicClient.Resource(gvr).Update(ctx, unstructuredObj, metav1.UpdateOptions{})
```
MikeSpreitzer, line 171: Not an error if the object and its GVR are already deleted.

pkg/binding/selectors.go
```diff
@@ -117,17 +74,104 @@ func (c *Controller) updateDecisions(obj runtime.Object) error {
...
+	if _, found := labels[util.BindingPolicyLabelSingletonStatusKey]; found {
+		return nil
+	}
+
+	labels[util.BindingPolicyLabelSingletonStatusKey] = bindingPolicyName
+	unstructuredObj.SetLabels(labels)
+
+	gvr, found := c.gvkGvrMapper.GetGvr(unstructuredObj.GetObjectKind().GroupVersionKind())
+	if !found {
+		return fmt.Errorf("failed to get GVR for object %v", util.RefToRuntimeObj(obj))
```
MikeSpreitzer, line 167: Not an error if the object and its GVR are already deleted.

pkg/binding/bindingpolicy-resolution.go
MikeSpreitzer: More importantly, the ultimate consumer does not really care, it is a loop in `func (c *Controller) requeueSelectedWorkloadObjects`.

pkg/binding/bindingpolicy.go
```diff
@@ -113,16 +129,33 @@ func (c *Controller) requeueForBindingPolicyChanges() error {
...
+	}
+
+	objs := make([]runtime.Object, 0, len(objectKeys))
+	for _, key := range objectKeys {
+		obj, err := c.getObjectFromKey(*key)
+		if err != nil {
+			return fmt.Errorf("failed to get object from key: %w", err)
+		}
+
+		objs = append(objs, obj)
```
MikeSpreitzer, line 151: What if `obj == nil` (which c.getObjectFromKey can return with a `nil` error)?

test/e2e/multi-cluster-deployment/use-kubestellar.sh
ezrasilvera: agree about "workload" We need some agreed terminology. Workload is not really defined in KS docs.

ezrasilvera:
@vMaroon @nirrozenbaum now that we switch to actually using the pluggable transport and new bindings, we need to make sure our documents are aligned (architecture doc, examples, etc  ).  I think that these doc updates  dhould  be added to this   PR , It will help with the review as well. Thanks

test/e2e/multi-cluster-deployment/use-kubestellar.sh
MikeSpreitzer: Ah, we since we have a change from wrapping individual objects to wrapping everything matching a BindingPolicy, that explains the change in expected count. Why is the ManifestWork for the status agent `Deployment` separate? Because it is generated by the status controller, not the transport controller

nirrozenbaum:
> @vMaroon @nirrozenbaum now that we switch to actually using the pluggable transport and new bindings, we need to make sure our documents are aligned (architecture doc, examples, etc ). I think that these doc updates dhould be added to this PR , It will help with the review as well. Thanks

@ezrasilvera agreed, we'll make sure it's part of the PR.

test/e2e/multi-cluster-deployment/use-kubestellar.sh
nirrozenbaum: maybe we add the term "workload" to the architecture doc?
We actually used this term in the Binding CRD to define the list of objects (cluster-scope + namespace-scope) that are propagated from center to WEC and as a result of a specific BindingPoilcy.
workload use:
https://github.com/kubestellar/kubestellar/blob/main/api/control/v1alpha1/types.go#L196 
the fact that it's mapped to a specific BindingPolicy:
https://github.com/kubestellar/kubestellar/blob/main/api/control/v1alpha1/types.go#L171-#L173

api/control/v1alpha1/types.go
vMaroon: Done. What you stated is right, the intention was to specify the "wanted" behavior, which would've been implemented later.

pkg/binding/controller.go
vMaroon: This was previously only checking if the object is a BindingPolicy. I think this should be issued as a bug and handled separately.

pkg/binding/controller.go
vMaroon: Do you refer to the removed log line? There is logging in cases of errors in updateResolutions and its call paths, and if everything goes well, controller::processNextWorkItem has the following:
>		
            // Finally, if no error occurs we Forget this item so it does not
		// get queued again until another change happens.
		c.workqueue.Forget(obj)
		c.logger.V(2).Info("Successfully synced", "object", obj)

pkg/binding/selectors.go
vMaroon: Added an isBeingDeleted check to the conditions of the call.

pkg/binding/selectors.go
vMaroon: I wasn't very sure of what you mean here, but I unified addSingletonLabel and removeSingletonLabel into a single function.

pkg/binding/selectors.go
vMaroon: Leaving this until validation is properly handled.

pkg/binding/bindingpolicy.go
vMaroon: Right I missed that. Do we want to keep getObjectFromKey that way? for now I added an `|| obj == nil` to the check.

vMaroon:
@MikeSpreitzer thank you for reviewing, answered all comments - expect for updating docs, which I'll do next.

docs/content/v0.20/architecture.md
```diff
@@ -309,40 +310,20 @@ follows:
...
-         (if more than one label is present, as other binding-policies may “own” the object).
-       - For each cluster:
-         -  Sets the manifest namespace == name of the cluster
-         -  Uses the client connected to the ITS to do a server-side
-            apply patch of the manifest into the namespace.
-         - At this time there is only one field manager name for the server-side
-           apply (`kubestellar`) but the name should include also the WDS
-           name to allow detecting conflicts when two different field managers
-           try to patch a manifest with the same name.   
+        - Whether the object is a match or not, the worker notes the object as a match for the in-memory representation
```
MikeSpreitzer, line 313: "the object is ... not" clashes with "notes the object as a match". I think you are trying to say something like the following?

> The worker notes whether or not the object is a match, in an internal representation of the matching set.

docs/content/v0.20/architecture.md
```diff
@@ -309,40 +310,20 @@ follows:
...
+         `Binding`, the worker enqueues the latter for syncing.
+        - If a matched `BindingPolicy` has `WantSingletonReportedState` set to true (**see note below), the object is 
+       labeled with a special label in order to be able to track it for status reporting.
+        - If no matching `BindingPolicy` has `WantSingletonReportedState` set to true, the worker removes the label if
+       exists.
     - Worker returns and is ready to pick other keys from the queue.
 
+**WantSingletonReportedState**:
+currently it is the user's responsibility to make sure that a binding-policy that sets `WantSingletonReportedState` to
+true is not in conflict with other binding-policies that do the same, and that the binding-policy selects only a single
```
MikeSpreitzer, line 324: Is "that do the same" an accurate limitation here? Will the implementation do the right thing if two BindingPolicy objects match a given workload object, where one BindingPolicy has `WantSingletonReportedState == true` and the other has it `false`?

docs/content/v0.20/architecture.md
```diff
@@ -68,15 +68,20 @@ and updates *WorkStatus* objects in the ITS namespace associated with the WEC.
 
 ## KubeStellar Controller Manager
 
-This module manages binding controller and status controller. The binding controller watches `BindingPolicy` and workload objects on the Workload Definition Space (WDS) and and maintains in the Inventory and Transport Space (ITS) a wrapped object per workload object to be delivered. The status controller watches for *WorkStatus* objects on the ITS and updates the
+This module manages binding controller, transport controller, and status controller.
```
MikeSpreitzer, line 71: How does this transport controller get into this controller manager without creating a dependency cycle among go modules?

pkg/binding/bindingpolicy.go
MikeSpreitzer: It seems a little treacherous to me the way it is now, as demonstrated by the fact that you missed this. Does not have to be addressed in this PR.

pkg/binding/selectors.go
```diff
@@ -117,17 +77,89 @@ func (c *Controller) updateDecisions(obj runtime.Object) error {
...
+		labels = make(map[string]string)
+	}
+
+	_, found := labels[util.BindingPolicyLabelSingletonStatusKey]
+
+	if bindingPolicyName == emptyBindingPolicyName {
+		if !found {
+			return nil
+		}
+		delete(labels, util.BindingPolicyLabelSingletonStatusKey)
```
MikeSpreitzer, line 134: Is it safe to side-effect the given `obj`? I see nothing in the comments, naming, or type system that answers the question. This is bad.

test/e2e/multi-cluster-deployment/use-kubestellar.sh
MikeSpreitzer: (A) We used to have a glossary. We should do that again.

(B) We have been designing to allow use cases with overlapping "workloads". Or, put another way: using "workload" simply to mean the user's objects that KubeStellar downsyncs and upsyncs, without supposing the "workloads" are disjoint or that a "workload" has any identifier or other realization in KubeStellar.

pkg/binding/controller.go
MikeSpreitzer: Then go ahead and create an Issue.

pkg/binding/controller.go
MikeSpreitzer: No, I was referring to a change that is no longer visible. The change that I commented on replaced a return of a wrapped error with a return of an error without wrapping.

The question now is: are the errors returned by `c.updateResolutions` sufficiently self-explanator? I suspect so, given that it works on only one thing, the same object that the caller is working on.

pkg/binding/selectors.go
MikeSpreitzer: What I meant is that there is a lot of code shared between

```
c.dynamicClient.Resource(gvr).Update(ctx, unstructuredObj, metav1.UpdateOptions{})
```

and

```
c.dynamicClient.Resource(gvr).Namespace(unstructuredObj.GetNamespace()).Update(ctx, unstructuredObj, metav1.UpdateOptions{})
```

The only difference is whether `Namespace` is in the middle of tha chain of calls.

This is not a big deal, certainly not a blocker.

docs/content/v0.20/architecture.md
vMaroon: In that case, the one that has `WantSingletonReportedState` set to true will lead to singleton status reporting for the workload object (assuming it does not conflict with any other, and it selects a single cluster) - and the other one will have no effect on status, but will lead to getting the object delivered to all clusters it selects.

docs/content/v0.20/architecture.md
vMaroon: I think this question might be biased by the knowledge of this problem. It's a technical matter that isn't necessarily related to the architecture specified here - in my opinion. 

BTW, I think the state of this doc is not very satisfying, and perhaps it requires a full re-write.

docs/content/v0.20/architecture.md
nirrozenbaum: @MikeSpreitzer the question of "will the implementation do the right thing" is problematic, cause we didn't define what is the right thing.
btw, if you ask me, the singleton status is a field for status collection and should not affect the distribution. 
so in that case I believe Maroon implemented it "the right way".

pkg/binding/bindingpolicy.go
vMaroon: Opened issue #1790

pkg/binding/selectors.go
vMaroon: Opened issue #1791

pkg/binding/controller.go
vMaroon: Opened issue #1792

vMaroon:
@MikeSpreitzer addressed all comments.

docs/content/v0.20/architecture.md
MikeSpreitzer: No, what I am saying here is that the architecture document is just plain wrong. The transport controller is a distinct process in a distinct pod from a distinct `Deployment` object (I suspect) from a distinct GitHub repo. This is not a tiny detail to omit; it is a major point of the architecture.

docs/content/v0.20/architecture.md
MikeSpreitzer: @nirrozenbaum : my comment here is on documentation of the user's responsibility, my concern is that it is not an accurate reflection of the limitation of the current implementation.

If we are lucky, we will resolve #1734 soon and fix the implementation and documentation soon.

pkg/binding/selectors.go
```diff
@@ -117,17 +77,92 @@ func (c *Controller) updateDecisions(obj runtime.Object) error {
...
 	return nil
 }
+
+// handleSingletonLabel adds or removes the singleton label from the object in the cluster.
+// If a bindingPolicyName is provided, the singleton label is added to the object. If the bindingPolicyName is empty,
+// the singleton label is removed.
+//
+// The method parameter `obj` is not mutated by this function.
+func (c *Controller) handleSingletonLabel(ctx context.Context, obj runtime.Object, bindingPolicyName string) error {
+	unstructuredObj, ok := obj.DeepCopyObject().(*unstructured.Unstructured)
```
MikeSpreitzer, line 121: FYI: this tends to be expensive; this func could do the copy only when it has decided to make a change.

pkg/binding/selectors.go
```diff
@@ -117,17 +77,92 @@ func (c *Controller) updateDecisions(obj runtime.Object) error {
...
-				"object in its resolution", "binding", bindingpolicy.GetName(),
+				"object in its resolution", "binding", bindingPolicy.GetName(),
 				"object", util.RefToRuntimeObj(obj))
-			c.enqueueBinding(bindingpolicy.GetName())
+			c.enqueueBinding(bindingPolicy.GetName())
+		}
+
+		// make sure object has singleton status if needed
+		if !isBeingDeleted(obj) &&
+			c.bindingPolicyResolver.ResolutionRequiresSingletonReportedState(bindingPolicy.GetName()) {
```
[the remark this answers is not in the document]
vMaroon: Due to the assumption of the user abiding by the API requirements. 
I did consider adding checks and loggings, but these are half measures. This should be properly handled in requirement validation code.

pkg/binding/selectors.go
vMaroon: I first went for that, but then this was a bit cleaner in code. But since you commented this I'll do it.

pkg/binding/selectors.go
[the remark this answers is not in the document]
MikeSpreitzer: Oh, right, the positive case should be triggered only once even if considered per BindingPolicy.

docs/content/v0.20/architecture.md
vMaroon: @MikeSpreitzer you're right, the API documentation was changed previously during the review, I've updated it again to be aligned:

>
	// WantSingletonReportedState indicates that (a) the number of selected locations is intended
	// to be 1 and (b) the reported state of each downsynced object should be returned back to
	// the object in this space.
	// When this field is true, there should be no other BindingPolicy with `WantSingletonReportedState`
	// == true that matches any of the same workload objects.
	//
	// At the moment, it is the user's responsibility to comply to the above.

docs/content/v0.20/architecture.md
```diff
@@ -41,7 +41,7 @@ Examples of users interaction with KubeStellar are illustrated in the
 
 The KubeStellar architecture has these main modules:
```
MikeSpreitzer, line 42: The transport controller does not appear in this list. I think it is a big enough deal that it should.

MikeSpreitzer: FYI: (this PR did not introduce this problem, but makes it worse) it would be very helpful if this list made it clear what comes from the KubeStellar org vs is imported from OCM.

docs/content/v0.20/architecture.md
```diff
@@ -239,20 +240,20 @@ as part of the decoupling of transport from binding as described above, the foll
...
+objects and references to the concrete list of destinations that were selected by the policy.
+
 
-The architecture and the event flow for **CURRENT** code for create/update object events is
+The architecture and the event flow of the code for create/update object events is
 illustrated in Figure 3 (some details might be omitted to make the flow easier
-to understand). Once the planned change that is describe in this document is integrated we will update this figure accordingly.
+to understand). 
 
 <figure>
```
MikeSpreitzer, line 258: FYI: this PR did not introduce this problem, but carried on with it. This figure depicts only  one of the "flows" in the binding controller, the one responding to a notification about a workload object. There are others.

docs/content/v0.20/architecture.md
vMaroon: The expected restriction scope is over the singleton-requiring bindingpolicies. I don't see a problem in the case of a binding-policy requiring singleton status reported over a set of objects, while selecting a single cluster, and a different binding-policy matching the same set of objects for distribution without requiring status reporting.

The problem is when there are multiple sources for **singleton** status.

vMaroon: @MikeSpreitzer done

docs/content/v0.20/architecture.md
vMaroon: Lets only do mandatory changes in this PR, and perhaps rework the architecture docs after.

MikeSpreitzer:
The changes to architecture.md are good enough now.

In double-checking everything, I remembered one more problem. The opening comment used to --- rightfully --- say that this should not be merged without updating the instructions for how to run KubeStellar. We now have these instructions in two places: docs/content/v0.20/examples.md and in the E2E test bash scripts. #1780 updated the latter, nothing has updated the former.

I am going to accept this as is, on the grounds that updated instructions are in there _somewhere_. I think the other copy of the instructions needs to be updated ASAP.

/lgtm
/approve
Thanks!

PR #1767, opened 2024-02-16
🌱 Refactor and rename common part of transport controller

### LINKED ISSUE #1750: bug: confusing content in cmd/

MikeSpreitzer:
## Summary
This PR moves the generic transport controller code that was in `cmd/` to `pkg/transport/cmd/`.

This PR also updates the architecture document to be a little more explicit about the structure here.

## Related issue(s)

Fixes #1750

### DIFF
--- a/docs/content/v0.20/architecture.md
+++ b/docs/content/v0.20/architecture.md
@@ -486,7 +486,7 @@ status updates for one object do not require updates of a whole bundle.
 ### Transport Controller - PLANNED TO BE ADDED
 
 The transport controller is pluggable and allows the option to plug different
-implementations of the transport interface. In order to implement transport plugin, the plugin should supply the following:
+implementations of the transport interface. The interface between the plugin and the generic code is [a Go language interface](../../pkg/transport/transport.go) that the plugin has to implement. This interface requires the following from the plugin.
 - Upon registration of a new WEC, plugin should create a namespace for the WEC in the ITS and delete the namespace once the WEC registration goes away (mailbox namespace per WEC);
 - Plugin must be able to wrap any number of objects into a single wrapped object;
 - Have an agent that can be used to pull the wrapped objects from the mailbox namespace and apply them to the WEC. A single example for such an agent is an agent that runs on the WEC and watches the wrapped object in the corresponding namespace in the central hub and is able to unwrap it and apply the objects to the WEC. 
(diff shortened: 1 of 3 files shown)


nirrozenbaum:
@MikeSpreitzer I'm ok with the idea in the PR but would like to request a few changes in the structure:
1. no need to keep the deprecated files under cmd. no one is using this (other than me) and I will fix the pointer in that ocm-plugin as soon as this PR will be merged. keeping deprecated is even more confusing than what it was before. 
let's just rename and move to pkg as you suggested. you don't break anything with this change cause nothing is running transport atm. we don't even break ocm plugin repo cause it's pointing to specific version of KS in go.mod (version before this PR).
2. the separation to `pkg/transport` and `pkg/transportcontroller` is strange. please let's put everything under `pkg/transport`. 
3. once everything is under `pkg/transport` just change the functions in `generic_transport_controller` to lowercase so it's not exposed, e.g., `func Run(...) -> func run(...)` and `func NewTransportController(...) -> func newTransportController(...)`. Then we can keep using function `Run(...)` in `base_controller`. calling it from outside as suggested in this PR looks like - `transportcontroller.MainForTransport()` - using transport twice is redundant, it's in the package name.
after this change it will be called `transport.Run(..)` or alternatively `transport.Main(..)`.   
I strongly prefer calling it `Run`, since as you stated, this is not executable under cmd and therefore it's not main and this also might be confusing.

MikeSpreitzer:
I deleted the deprecated stuff.

I changed NewTransportController to newTransportController.

I changed MainForTransport to GenericMain. There is still a problem with Run, in that a controller typically has a Run func --- and this one is no exception (genericTransportController::Run). Really we are using the word "controller" with two meanings here, one is the whole executable and one is the specific Go language object constructed by newTransportController. I think having the word "main" in the command func helps to direct the reader's thinking.

MikeSpreitzer:
I merged package transportcontroller into transport.

nirrozenbaum:
ok. there is a point in the the fact that a typical controller has Run function.
I can also live with GenericMain as the function name.
What if we put what was previously under `cmd/transport` into `pkg/transport/cmd`, would that work? 
it creates a clear separation between the "cmd-alike" part and the pkg classic part. wdyt?

MikeSpreitzer:
That package structure works for me too.

MikeSpreitzer:
The force-pushes make the latest package structure change and squash.

docs/content/v0.20/architecture.md
```diff
@@ -486,16 +486,16 @@ status updates for one object do not require updates of a whole bundle.
...
 - Plugin must be able to wrap any number of objects into a single wrapped object;
 - Have an agent that can be used to pull the wrapped objects from the mailbox namespace and apply them to the WEC. A single example for such an agent is an agent that runs on the WEC and watches the wrapped object in the corresponding namespace in the central hub and is able to unwrap it and apply the objects to the WEC. 
 - Have inventory representation for the clusters.
 
 The above list is required in order to comply with [<u>SIG Multi-Cluster Work API</u>](https://multicluster.sigs.k8s.io/concepts/work-api/).
 
-In order to implement a KubeStellar transport plugin, one has to implement code that supplies a function to wrap any number of objects into a single wrapped objects. 
-The transport interface that should be implemented can be seen [here](https://github.com/kubestellar/kubestellar/blob/main/pkg/transport/transport.go).  
+Each plugin has an executable with a `main` func that calls [the generic code](../../pkg/transport/cmd/generic-main.go), passing the plugin object that implements the plugin interface.
```
nirrozenbaum, line 497: broken link to the previous “transportcontroller”

nirrozenbaum:
it didn’t refresh so ignore the last comment.
/lgtm
Thanks!

PR #1768, opened 2024-02-16
✨ Change the explicitly coded API group for WorkStatus

### LINKED ISSUE #1701: Remember to update the API group (in KS/KS) for WorkStatus

waltforme:
This PR changes the explicitly coded API group for WorkStatus to `control.kubestellar.io`.

Also updates the helm chart for status addon (in the e2e tests) which incorporated the 'original' WorkStatus API group change.

Fixes #1701.

### DIFF
--- a/pkg/util/unstructured.go
+++ b/pkg/util/unstructured.go
@@ -39,7 +39,7 @@ const (
 	BindingPolicyResource                = "bindingpolicies"
 	BindingKind                          = "Binding"
 	BindingResource                      = "bindings"
-	WorkStatusGroup                      = "edge.kubestellar.io" // TODO: update/import after status-addon is rehomed and its api group for WorkStatus is updated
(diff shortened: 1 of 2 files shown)


MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #1773, opened 2024-02-17
🐛 Put -v in kustomize YAML for the Helm chart

MikeSpreitzer:
## Summary
This PR fixes the kustomization file that gets invoked by `make chart`, to keep putting the `-v=2` among the command line arguments of the kubestellar controller-manager.

This PR also updates the _other_ file that looks like a customization file for the same Deployment to point the reader to the kustomize input that matters.

## Related issue(s)

Fixes #


pdettori (approved):
/lgtm

PR #1774, opened 2024-02-18
🌱 Fix a typo (and replace it) for the binding controller

waltforme:
## Summary
It was overlooked during the renaming because it was not spelled correctly thus not returned by searches. Now let's fix this.


MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #1775, opened 2024-02-18
✨ Start unit test of binding controller, using fake clientsets

MikeSpreitzer:
## Summary
This PR attempts to add a unit test for the binding controller.
This PR takes the approach of using fake clientsets.

## Related issue(s)

Fixes #


MikeSpreitzer:
One problem currently here is the way that the fake clientsets respond to discovery requests. For some kinds of requests (including the one that this controller uses), the response is a hard-coded "I got nothing". For others the test can prescribe the response.

ezrasilvera:
Thanks. 
/lgtm
/approve

MikeSpreitzer:
Oh, dang, sorry I was not clear enough about the status of this. It is not ready to be merged. https://github.com/kubernetes/client-go/blob/v0.28.2/discovery/fake/discovery.go#L84-L88 makes it useless.

PR #1776, opened 2024-02-18
🌱 Refactor binding controller to make it testable

MikeSpreitzer:
## Summary
This PR refactors the binding controller for testability, in two ways. One is remove the dependency on the controller-runtime `Manager` object (which was only used in the most trivial way). The other is to separate out the step of getting the CRDs installed.

This PR also adds some comments on fields in the `Controller` struct, identifying which client is use for what.

## Related issue(s)

Fixes #


nirrozenbaum:
/lgtm
Thanks!

PR #1777, opened 2024-02-18
🌱 Made API registration work for plain as well as controller-runtime

MikeSpreitzer:
## Summary
This PR updates the way that the API types are registered with one or more `Scheme` objects so that it works with plain kube code as well as controller-runtime based code.

## Related issue(s)

Fixes #


api/control/v1alpha1/register.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package v1alpha1
 
 import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sruntime "k8s.io/apimachinery/pkg/runtime"
```
nirrozenbaum, line 21: why do we need `k8sruntime` alias?

api/control/v1alpha1/groupversion_info.go
```diff
@@ -21,17 +21,16 @@ limitations under the License.
 package v1alpha1
 
 import (
+	k8sruntime "k8s.io/apimachinery/pkg/runtime"
```
nirrozenbaum, line 24: same question.. you can use just `runtime.NewSchemeBuilder`

api/control/v1alpha1/register.go
MikeSpreitzer: It's not the only "runtime" in our world.

MikeSpreitzer:
The force-push to 490bb4514 makes the changes that @nirrozenbaum suggested above.

nirrozenbaum (approved):
/lgtm
Thanks!

PR #1780, opened 2024-02-19
🌱 run transport controller in e2e tests

nirrozenbaum:
## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/pr-test-multicluster.yml
+++ b/.github/workflows/pr-test-multicluster.yml
@@ -38,12 +38,16 @@ jobs:
           wget https://github.com/kubestellar/kubeflex/releases/download/v0.4.2/kubeflex_0.4.2_linux_amd64.tar.gz
           tar -xvf kubeflex_0.4.2_linux_amd64.tar.gz bin/kflex
           mv bin/kflex /usr/local/bin
-          rm -fr bin kubeflex_0.4.2_linux_amd64.tar.gz 
+          rm -fr bin kubeflex_0.4.2_linux_amd64.tar.gz
 
       - name: Run test
         run: |
           cd test/e2e/multi-cluster-deployment 
           KFLEX_DISABLE_CHATTY=true ./run-test.sh 
+
+      - name: show transport controller logs ## temp solution until we get transport controller to run as pod
+        if: always()
+        run: cat transport.log
 
       - name: show kubeconfig contexts
         if: always()
--- a/test/e2e/common/cleanup.sh
+++ b/test/e2e/common/cleanup.sh
@@ -15,6 +15,7 @@
 
 set -x # echo so that users can understand what is happening
 set -e # exit on error
+SRC_DIR="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)"
 
 :
 : -------------------------------------------------------------------------
@@ -25,3 +26,13 @@ kind delete cluster --name cluster2
 kind delete cluster --name kubeflex
 kubectl config delete-context cluster1 || true
 kubectl config delete-context cluster2 || true
+
+## as temp solution transport runs as executable. cleanup of the created artifacts.
+cd ${SRC_DIR}/../../.. ## go up KubeStellar directory
+rm -f transport.log
+ocm_transport_plugin_process="$(pgrep ocm-transport-plugin || true)"
+if [[ ! -z "$ocm_transport_plugin_process" ]]; 
+then 
+    kill ${ocm_transport_plugin_process}
+fi
+rm -f ocm-transport-plugin
--- a/test/e2e/common/setup-kubestellar.sh
+++ b/test/e2e/common/setup-kubestellar.sh
@@ -67,6 +67,30 @@ echo "wds1 created."
 
 :
 : -------------------------------------------------------------------------
+: Run OCM transport controller executable
+:
+cd "${SRC_DIR}/../../.." ## go up to KubeStellar directory
+KUBESTELLAR_DIR="$(pwd)"
+## this is a temp solution - run as executble process. this should be replaced to run as pod, ideally using helm
+wget https://github.com/kubestellar/ocm-transport-plugin/archive/refs/tags/v0.1.0-rc1.tar.gz
+tar -xf v0.1.0-rc1.tar.gz
+rm -rf v0.1.0-rc1.tar.gz
+cd ocm-transport-plugin-0.1.0-rc1
+OCM_TRANSPORT_PLUGIN_DIR="$(pwd)"
+pwd
+echo "replace github.com/kubestellar/kubestellar => ${KUBESTELLAR_DIR}/" >> go.mod
+make build
+mv ./bin/ocm-transport-plugin ${KUBESTELLAR_DIR}/ocm-transport-plugin
+cd "${KUBESTELLAR_DIR}"
+pwd
+rm -rf ${OCM_TRANSPORT_PLUGIN_DIR}
+echo "running ocm transport plugin..."
+./ocm-transport-plugin --transport-context imbs1 --wds-context wds1 --wds-name wds1 &> transport.log &
+
+echo "transport controller is running as background process."
+
+:
+: -------------------------------------------------------------------------
 : Create clusters and register with OCM
 :
 function create_cluster() {
--- a/test/e2e/common/setup-kubestellar.sh
+++ b/test/e2e/common/setup-kubestellar.sh
@@ -72,10 +72,11 @@ echo "wds1 created."
 cd "${SRC_DIR}/../../.." ## go up to KubeStellar directory
 KUBESTELLAR_DIR="$(pwd)"
 ## this is a temp solution - run as executble process. this should be replaced to run as pod, ideally using helm
-wget https://github.com/kubestellar/ocm-transport-plugin/archive/refs/tags/v0.1.0-rc1.tar.gz
-tar -xf v0.1.0-rc1.tar.gz
-rm -rf v0.1.0-rc1.tar.gz
-cd ocm-transport-plugin-0.1.0-rc1
+OCM_TRANSPORT_PLUGIN_RELEASE="0.1.0-rc2"
+wget https://github.com/kubestellar/ocm-transport-plugin/archive/refs/tags/v${OCM_TRANSPORT_PLUGIN_RELEASE}.tar.gz
+tar -xf v${OCM_TRANSPORT_PLUGIN_RELEASE}.tar.gz
+rm -rf v${OCM_TRANSPORT_PLUGIN_RELEASE}.tar.gz
+cd ocm-transport-plugin-${OCM_TRANSPORT_PLUGIN_RELEASE}
 OCM_TRANSPORT_PLUGIN_DIR="$(pwd)"
 pwd
 echo "replace github.com/kubestellar/kubestellar => ${KUBESTELLAR_DIR}/" >> go.mod
--- a/.github/workflows/pr-test-singleton-status.yml
+++ b/.github/workflows/pr-test-singleton-status.yml
@@ -45,6 +45,10 @@ jobs:
           cd test/e2e/singleton-status
           KFLEX_DISABLE_CHATTY=true ./run-test.sh 
 
+      - name: show transport controller logs ## temp solution until we get transport controller to run as pod
+        if: always()
+        run: cat transport.log
+
       - name: show kubeconfig contexts
         if: always()
         run: kubectl config get-contexts


test/e2e/common/setup-kubestellar.sh
```diff
@@ -66,6 +65,22 @@ make install-local-chart KUBE_CONTEXT=kind-kubeflex
...
 
+:
+: -------------------------------------------------------------------------
+: Run OCM transport controller executable
+:
+cd "${SRC_DIR}/../../.." ## go up to KubeStellar directory
+KUBESTELLAR_DIR="$(pwd)"
+cd ocm-transport-plugin-0.1.0-rc1
+echo "replace github.com/kubestellar/kubestellar => ${KUBESTELLAR_DIR}/" >> go.mod
+make build
```
ezrasilvera, line 76: If you want to use exec for now, let's not build them. You should have it already in the release tar

test/e2e/common/setup-kubestellar.sh
```diff
@@ -66,6 +65,22 @@ make install-local-chart KUBE_CONTEXT=kind-kubeflex
...
+./bin/ocm-transport-plugin --transport-context imbs1 --wds-context wds1 --wds-name wds1 >/dev/null &
```
ezrasilvera, line 77: Do you implicitly assume a KUBECONFIG is set and points to KF ?

test/e2e/common/setup-kubestellar.sh
nirrozenbaum: that's not possible.
an executable is pointing to a specific version of KubeStellar.
assuming current PR changes transport behavior we want to test *the NEW* behavior. 
for this we must get the source of ocm plugin, make sure it points to *the NEW* code that is currently being tested and build the executable/image.

a pre-build executable or image will not test the new changes of KubeStellar transport.

.github/workflows/pr-test-multicluster.yml
```diff
@@ -38,7 +38,10 @@ jobs:
           wget https://github.com/kubestellar/kubeflex/releases/download/v0.4.2/kubeflex_0.4.2_linux_amd64.tar.gz
           tar -xvf kubeflex_0.4.2_linux_amd64.tar.gz bin/kflex
           mv bin/kflex /usr/local/bin
-          rm -fr bin kubeflex_0.4.2_linux_amd64.tar.gz 
+          rm -fr bin kubeflex_0.4.2_linux_amd64.tar.gz
+          wget https://github.com/kubestellar/ocm-transport-plugin/archive/refs/tags/v0.1.0-rc1.tar.gz
```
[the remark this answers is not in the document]
nirrozenbaum: see previous comment

test/e2e/common/setup-kubestellar.sh
nirrozenbaum: this is not something I've introduced. this is how the test is written today in main.
it assumes a single KUBECONFIG and uses context to switch between different clusters.
see here for example - https://github.com/kubestellar/kubestellar/blob/b2382485dc3da5c4ca5d8692940c470bf3d5ca00/test/e2e/common/setup-kubestellar.sh#L51

I just re-used whatever exists in current tests

.github/workflows/pr-test-multicluster.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       
       - uses: actions/setup-go@v5
         with:
-          go-version: v1.19
+          go-version: v1.20
```
MikeSpreitzer, line 26: This has been done in #1782

test/e2e/common/setup-kubestellar.sh
MikeSpreitzer: @ezrasilvera: The assumption is that there is a kubeconfig file that standard kube clients will use (for the standard reasons: either being named in `$KUBECONFIG` or located at `~/.kube/config`) and it has two specific _contexts_ within it for the ITS and WDS.

nirrozenbaum:
@MikeSpreitzer I would be happy to get your input about the following:
what do you think about having initial version of the e2e tests working with transport as executable process?
I know this is not optimal and not the way we want. I do have `ko-build-local` in makefile that builds a local image, but I get the impression that running the containerized version might take me some time, could be few days.

the reason I'm asking is because I want to unblock other PRs and issues and keep working on changing this from executable to container (the agile way). this will allow Maroon to make progress on his end and will allow Effi to validate open bugs are fixed.
if you agree, of course I will change the run mode not to garbage the output but redirect transport controller to write to a log file. at the end we can print it and then delete it (same as we print today pod's logs).

.github/workflows/pr-test-multicluster.yml
nirrozenbaum: removed from this PR

MikeSpreitzer:
@nirrozenbaum : Yes, I also want to unblock work. I think the essence of the problem is the dependency cycle. So let's urgently pursue that discussion. I will create an Issue for it, while recognizing that it affects the release process in both (if we stay with two) repos.

test/e2e/multi-cluster-deployment/run-test.sh
```diff
@@ -27,3 +27,7 @@ source "${COMMON_SRCS}/setup-shell.sh"
 "${COMMON_SRCS}/setup-kubestellar.sh"
 "${SRC_DIR}/use-kubestellar.sh"
 "${SRC_DIR}/update-delete.sh"
+
+## temp - print transport controller log
```
ezrasilvera, line 31: Please remove and re-push

test/e2e/common/setup-kubestellar.sh
ezrasilvera: That's the  cycle dependency   . We need to resolve it :(

.github/workflows/pr-test-multicluster.yml
```diff
@@ -38,12 +38,15 @@ jobs:
           wget https://github.com/kubestellar/kubeflex/releases/download/v0.4.2/kubeflex_0.4.2_linux_amd64.tar.gz
           tar -xvf kubeflex_0.4.2_linux_amd64.tar.gz bin/kflex
           mv bin/kflex /usr/local/bin
-          rm -fr bin kubeflex_0.4.2_linux_amd64.tar.gz 
+          rm -fr bin kubeflex_0.4.2_linux_amd64.tar.gz
+          wget https://github.com/kubestellar/ocm-transport-plugin/archive/refs/tags/v0.1.0-rc1.tar.gz
```
ezrasilvera, line 42: Please move that to setup-kubestellar.  As you build there that is your  actual install.  You can then also delete the source directory after the build

test/e2e/common/setup-kubestellar.sh
ezrasilvera: That's a risky requirements. So you are basically saying that all transport plugins will need to be re-built?  I believe this should happen only when we change interfaces and the interfaces should be defined in a different repo.

.github/workflows/pr-test-multicluster.yml
MikeSpreitzer: I see that https://github.com/kubestellar/ocm-transport-plugin/releases/tag/v0.1.0-rc1 also has archives of built stuff. Do we really need to have contributors and CI repeat the build?

ezrasilvera: @MikeSpreitzer That was my comment above (https://github.com/kubestellar/kubestellar/pull/1780#discussion_r1494625925) 
@nirrozenbaum point was that he wanted to be able to test new changes in the base transport so he needs to rebuild it. 
I think this is a problematic dependency and it's not a reasonable requirement.    I'm OK with postponing  this to follow-up PR

nirrozenbaum: right. an executable is pointing to a specific version of KubeStellar.
assuming current PR changes transport behavior we want to test the NEW behavior.
for this we must get the source of ocm plugin, make sure it points to the NEW code that is currently being tested and build the executable/image.

a pre-build executable or image will not test the new changes of KubeStellar transport.
this is indeed problematic but let's solve this in a follow up PR (depends on the decision of how we solve the cycle) to unblock other PRs and issues

nirrozenbaum: /unhold

MikeSpreitzer: (A) @nirrozenbaum : I think the purpose of the two existing E2E tests written in bash is twofold: (1) test that the local copy of the ks/ks repo works, and (2) show contributors how to exercise the `main` branch of ks/ks. My understanding of our plans is that we intend to make a release (@pdettori is working on that today) of ks/ks that does NOT use pluggable transport. If this is correct, then that should be completed before a change like this PR should be merged.

(B) I think we do need to settle the issue of how to handle the dependency cycle. As I mentioned in my discussion in Slack, I think that hacking around with both `main`s in absence of a principled approach can get us stuck.

ezrasilvera: @MikeSpreitzer I think your (A) is not relevant here because this code is not executed without bringing up the controller . I don't think it will be problematic for @pdettori to have this (dead) code in the release. 
In any case I don't mind waiting to after 0.20.0

nirrozenbaum: agree with Ezra. since this is an executable and whoever runs 0.20 will use helm to install it, the helm will not install transport controller.

test/e2e/common/cleanup.sh
```diff
@@ -25,3 +26,13 @@ kind delete cluster --name cluster2
...
+
+## as temp solution transport runs as executable. cleanup of the created artifacts.
+cd ${SRC_DIR}/../../.. ## go up KubeStellar directory
+rm -f transport.log
+ocm_transport_plugin_process="$(pgrep ocm-transport-plugin || true)"
+if [[ ! -z "$ocm_transport_plugin_process" ]]; 
+then 
+    kill ${ocm_transport_plugin_process}
+fi
+rm -f ocm-transport-plugin
```
MikeSpreitzer, line 38: FYI: GitHub is less cranky when there is a terminal linebreak

test/e2e/common/setup-kubestellar.sh
```diff
@@ -65,6 +65,30 @@ make install-local-chart KUBE_CONTEXT=kind-kubeflex
...
+rm -rf v0.1.0-rc1.tar.gz
+cd ocm-transport-plugin-0.1.0-rc1
+OCM_TRANSPORT_PLUGIN_DIR="$(pwd)"
+pwd
+echo "replace github.com/kubestellar/kubestellar => ${KUBESTELLAR_DIR}/" >> go.mod
+make build
+mv ./bin/ocm-transport-plugin ${KUBESTELLAR_DIR}/ocm-transport-plugin
+cd "${KUBESTELLAR_DIR}"
+pwd
+rm -rf ${OCM_TRANSPORT_PLUGIN_DIR}
```
MikeSpreitzer, line 86: Doesn't this belong in cleanup?

nirrozenbaum: in the setup I left only the needed files on filesystem. only what is a must.
then the cleanup script removes the things that are a must - e.g., rm log file + kill the process + rm the executable file.
I can move this also to cleanup but this can be removed as soon as I untar the dir. no need to keep it.

.github/workflows/pr-test-multicluster.yml
MikeSpreitzer: @ezrasilvera : I am not sure I understand what you mean by "this code is not executed without bringing up the controller". A contributor reading these E2E tests and doing what is in them (see my A.2) will build and bring up the transport controller, right?

@nirrozenbaum, I am surprised here. I thought you favored the staggered release approach --- which this is not. I expect @pdettori to have ks 0.20.0-rc1 very soon.

MikeSpreitzer: Correction: ks 0.20.0-rc1 is there now.

nirrozenbaum:
squashed to a single commit.

.github/workflows/pr-test-multicluster.yml
nirrozenbaum: np from my side to wait for Paolo's release.

test/e2e/common/setup-kubestellar.sh
MikeSpreitzer: Yes, I see what is currently in the PR.

Note that sometimes things crash in the middle, so a later cleanup step is not reached. And doing `try ... finally ...` is hostile to anyone trying to debug what went wrong. So we have taken a different approach: a test starts with the cleanup. That would even involve removing directories that `tar xf` would later make.

nirrozenbaum: right. if that is not a big concern (which I understand it's not), let's leave as is and remove in the PR that would run a containerized transport controller.

nirrozenbaum: yes it is.  
you mean 0.1.0-rc2 right?
cause officially 0.1.0 still doesn’t exist.. we are still in the process of finalizing and testing 0.1.0.
so I should update otp to point to release 0.20.0-rc1 and create 0.1.0-rc2

MikeSpreitzer: ks/ks now has release 0.20.0, OTP should depend on that now.
Yes, I think it would be best to make an OTP release (candidate) and make this PR use the pre-built stuff from that release.

nirrozenbaum: done

MikeSpreitzer: Maybe there is a misunderstanding here. I still see `make build` above. I will continue the discussion in the thread that @ezrasilvera started.

test/e2e/common/setup-kubestellar.sh
MikeSpreitzer: If we are going to use the staggered release answer to the dependency cycle then that says what to do here. The instructions and tests in release X of ks/ks use a release of ks/OTP that existed before X was made. Yes, that means that we can not advance both repos simultaneously with each depending on a change in the other. That would be synchronized releases.

MikeSpreitzer: Perhaps the problem here is that our discussion of how to handle the dependency cycle has not said enough about testing. If we do staggered releases, how does a contributor to ks/OTP test their work? How does a contributor to ks/ks test their work? If the E2E tests are only in ks/ks then there is no way for a contributor to ks/OTP to do E2E tests of their work. We could move the E2E tests to ks/OTP, but that would just move the problem --- giving a contributor to ks/ks no way to do E2E tests of _their_ work. One, painful, answer would be to have E2E tests in both repos.

nirrozenbaum:
```
If we are going to use the staggered release answer to the dependency cycle then that says what to do here. The instructions and tests in release X of ks/ks use a release of ks/OTP that existed before X was made. Yes, that means that we can not advance both repos simultaneously with each depending on a change in the other. That would be synchronized releases.
```

right. for testing release v0.20.0 of ks/ks we used otp v0.1.0-rc1.
now these testing were over and release v0.20.0 is done.

then we created ks/otp v0.1.0-rc2, this is also done.

now we need to update back to main to use ks/otp 0.1.0-rc2 so that in the next time we release ks/ks we will have the most up to date ks/otp used in the e2e tests.

MikeSpreitzer: I think we still need work on #1786

PR #1769, opened 2024-02-16
✨ Add integration test for binding controller, using test server

MikeSpreitzer:
## Summary
This PR adds a randomized integration test for the binding controller.

## Related issue(s)

Fixes #

PR #1784, opened 2024-02-19
✨ Run the integration test in CI

MikeSpreitzer:
## Summary
This PR adds a Prow job that runs the integration test.

This is built on #1769 , and will be rebased if that merges first.

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/v0.20/contributor.md
+++ b/docs/content/v0.20/contributor.md
@@ -61,7 +61,7 @@ X ArgoCD CLI
 
 ## Integration testing
 
-There is currently one integration test. Contributors can run it. There is also a Prow job that runs this test.
+There is currently one integration test. Contributors can run it. There is also [a GitHub workflow](../../../.github/workflows/pr-test-integration.yml) that runs this test.
 
 This test requires you to already have `etcd` on your `$PATH`.
 See https://github.com/kubernetes/kubernetes/blob/v1.28.2/hack/install-etcd.sh for an example of how to do that.
--- a/docs/content/v0.20/contributor.md
+++ b/docs/content/v0.20/contributor.md
@@ -61,7 +61,7 @@ X ArgoCD CLI
 
 ## Integration testing
 
-There is currently one integration test. Contributors can run it. It is not (yet) run in CI.
+There is currently one integration test. Contributors can run it. There is also a Prow job that runs this test.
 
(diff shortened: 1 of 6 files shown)


MikeSpreitzer:
The force-push to 715621c8c is the needed rebase.

waltforme:
Not a blocker:
I ran the test manually (as suggested by 'Contributors can run it'), and it was successful. The only thing is that the run ends up with a tarball for etcd as well as the extracted contents of the tarball. Maybe we can do some cleanup at the end? Or check whether we need to download etcd at the beginning.
/lgtm

ezrasilvera:
> Not a blocker: I ran the test manually (as suggested by 'Contributors can run it'), and it was successful. The only thing is that the run ends up with a tarball for etcd as well as the extracted contents of the tarball. Maybe we can do some cleanup at the end? Or check whether we need to download etcd at the beginning. /lgtm

For the CICD it doesn't matter as we delete the VM after the test. However I do prefer we clean up.

hack/run-integration-tests.sh
```diff
@@ -0,0 +1,42 @@
...
+if which sha256sum
+then got_sha=$(sha256sum     etcd.tar.gz | awk '{ print $1 }')
+else got_sha=$(shasum -a 256 etcd.tar.gz | awk '{ print $1 }')
+fi
+if [ $expected_sha != "$got_sha" ]; then
+    echo "Got SHA256 $got_sha instead of expected $expected_sha" >& 2
+    exit 1
+fi
+
+tar xzf etcd.tar.gz
```
ezrasilvera, line 39: As @waltforme indicated, we can delete the tar at this point

MikeSpreitzer: done.

docs/content/v0.20/contributor.md
```diff
@@ -61,7 +61,7 @@ X ArgoCD CLI
 
 ## Integration testing
 
-There is currently one integration test. Contributors can run it. It is not (yet) run in CI.
+There is currently one integration test. Contributors can run it. There is also a Prow job that runs this test.
```
ezrasilvera, line 64: This is no longer true (Prow)

MikeSpreitzer: Right. Fixed.

PR #1785, opened 2024-02-20
✨ Update GitHub Workflow, GoReleaser and  examples doc for publishing RC with new process.

pdettori:
## Summary

Update GitHub Workflow, GoReleaser and  examples doc for publishing RC2 with new process.

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/v0.20/examples.md
+++ b/docs/content/v0.20/examples.md
@@ -17,8 +17,6 @@ The following steps establish an initial state used in the examples below.
    kflex init --create-kind
    ```
 
-  If you are installing KubeStellar on an existing OpenShift cluster, just use the command `kflex init`.
-
 2. Update the post-create-hooks in KubeFlex to install kubestellar with the v0.20.0-alpha.1 images:
 
    ```shell
@@ -45,7 +43,7 @@ which installs OCM on it.
    and then install the status add-on:
 
    ```shell
-   helm --kube-context imbs1 upgrade --install status-addon -n open-cluster-management oci://ghcr.io/kubestellar/ocm-status-addon-chart --version v0.2.0-rc2
+   helm --kube-context imbs1 upgrade --install status-addon -n open-cluster-management oci://ghcr.io/kubestellar/ocm-status-addon-chart --version v0.2.0-rc1
    ```
 
    see [here](https://github.ibm.com/dettori/status-addon) for more details on the add-on.
@@ -67,22 +65,6 @@ statefulset created in the imbs1-system namespace.
    ```shell
    kubectl --context kind-kubeflex get deployments,statefulsets --all-namespaces
    ```
-The output should looks something like the following:
-
-```shell
-NAMESPACE            NAME                                             READY   UP-TO-DATE   AVAILABLE   AGE
-ingress-nginx        deployment.apps/ingress-nginx-controller         1/1     1            1           22h
-kube-system          deployment.apps/coredns                          2/2     2            2           22h
-kubeflex-system      deployment.apps/kubeflex-controller-manager      1/1     1            1           22h
-local-path-storage   deployment.apps/local-path-provisioner           1/1     1            1           22h
-wds1-system          deployment.apps/kube-apiserver                   1/1     1            1           22m
-wds1-system          deployment.apps/kube-controller-manager          1/1     1            1           22m
-wds1-system          deployment.apps/kubestellar-controller-manager   1/1     1            1           21m
-
-NAMESPACE         NAME                                   READY   AGE
-imbs1-system      statefulset.apps/vcluster              1/1     11h
-kubeflex-system   statefulset.apps/postgres-postgresql   1/1     22h
-```
 
 ## Scenario 1 - multi-cluster workload deployment with kubectl
 
@@ -94,14 +76,14 @@ Check for available clusters with label `location-group=edge`
 kubectl --context imbs1 get managedclusters -l location-group=edge
 ```
 
-Create a BindingPolicy to deliver an app to all clusters in wds1:
+Create a placement to deliver an app to all clusters in wds1:
 
 ```shell
 kubectl --context wds1 apply -f - <<EOF
-apiVersion: control.kubestellar.io/v1alpha1
-kind: BindingPolicy
+apiVersion: edge.kubestellar.io/v1alpha1
+kind: Placement
 metadata:
-  name: nginx-bpolicy
+  name: nginx-placement
 spec:
   clusterSelectors:
   - matchLabels: {"location-group":"edge"}
@@ -111,7 +93,7 @@ spec:
 EOF
 ```
 
-This BindingPolicy configuration determines **where** to deploy the workload by using
+This placement configuration determines **where** to deploy the workload by using
 the label selector expressions found in *clusterSelectors*. It also specifies **what**
 to deploy through the downsync.labelSelectors expressions.
 Each matchLabels expression is a criterion for selecting a set of objects based on
@@ -173,9 +155,9 @@ kubectl --context cluster2 get deployments -n nginx
 ```
 
 Please note, in line with Kubernetes’ best practices, the order in which you apply
-a BindingPolicy and the objects doesn’t affect the outcome. You can apply the BindingPolicy
+a placement and the objects doesn’t affect the outcome. You can apply the placement
 first followed by the objects, or vice versa. The result remains consistent because
-the binding controller identifies any changes in either the BindingPolicy or the objects,
+the placement controller identifies any changes in either the placement or the objects,
 triggering the start of the reconciliation loop.
 
 ## Scenario 2 - using the hosting cluster as WDS to deploy a custom resource
@@ -217,42 +199,11 @@ EOF
 To create a second WDS based on the hosting cluster, run the command:
 
 ```shell
-kflex create wds2 -t host
+kflex create wds2 -t host -p kubestellar
 ```
 
 where the `-t host` option specifies a control plane of type `host`.
 You can only create on control plane of type `host`.
-
-In this example, we use the helm chart method to install the kubestellat controller manager for the
-hosting cluster so that we can pass additional startup options.
-
-Label the `wds2` control plane as type `wds`:
-
-```shell 
-kubectl label cp wds2 kflex.kubestellar.io/cptype=wds
-```
-
-For this example, we use the `AppWrapper` custom resource defined in the
-[multi cluster app dispatcher](https://github.com/project-codeflare/multi-cluster-app-dispatcher)
-project.
-
-Install the AppWrapper CRD in the WDS and the WECs. Note that due to 
-[this issue](https://github.com/kubestellar/kubestellar/issues/1705) CRDs must be pre-installed 
-on the WDS and on the WECs when using API group filtering. For KubeStellar `0.20.0-rc1` this is required 
-when using a hosting cluster as WDS with a large number of API resources such as OpenShift.
-
-```shell
-clusters=(wds2 cluster1 cluster2);
-  for cluster in "${clusters[@]}"; do
-  kubectl --context ${cluster} apply -f https://raw.githubusercontent.com/project-codeflare/multi-cluster-app-dispatcher/v1.39.0/config/crd/bases/workload.codeflare.dev_appwrappers.yaml
-done
-```
-
-Apply the kubestellar controller-manager helm chart with the option to allow only delivery of objects with api group `workload.codeflare.dev`
-
-```shell
-helm upgrade --install -n wds2-system kubestellar oci://ghcr.io/kubestellar/kubestellar/controller-manager-chart --version v0.20.0-alpha.1 --set ControlPlaneName=wds2 --set APIGroups=workload.codeflare.dev
-```
 
 Check that the kubestellar controller for wds2 is started:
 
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -1,31 +1,39 @@
...
+  ldflags:
+  - "{{ .Env.LDFLAGS }}"
   goos:
   - linux
-  - darwin
   goarch:
   - amd64
   - arm64
-  - s390x
+  ignore:
```
MikeSpreitzer, line 12: Why do we need to "ignore" something not included in the outer product above?

.goreleaser.yaml
```diff
@@ -1,31 +1,39 @@
...
+    goarch: ppc64le
+  env:
+  - CGO_ENABLED=0
+kos:           
+  - repository: ghcr.io/kubestellar/kubestellar/controller-manager
+    main: ./cmd/controller-manager
+    build: controller-manager
+    tags:
+    - '{{.Version}}'
+    - latest
```
MikeSpreitzer, line 23: Why tag with `latest`?

.goreleaser.yaml
```diff
@@ -1,31 +1,39 @@
...
+    bare: true
+    preserve_import_paths: false
+    ldflags:
+    - "{{ .Env.LDFLAGS }}"
+    platforms:
+    - linux/amd64
+    - linux/arm64
+    env:
+    - CGO_ENABLED=0
```
MikeSpreitzer, line 32: Didn't we just say this above?

.goreleaser.yaml
```diff
@@ -1,31 +1,39 @@
...
+release:
+  disable: true # disable release while testing publishing
```
MikeSpreitzer, line 34: For my information: what is this disabling?

MikeSpreitzer, line 34: Is it true that the real reason that we disable publishing GitHub "release artifacts" is that we have none to publish now?

Makefile
```diff
@@ -14,7 +14,7 @@
 
 # Image repo/tag to use all building/pushing image targets
 DOCKER_REGISTRY ?= ghcr.io/kubestellar/kubestellar
-IMAGE_TAG ?= 0.20.0-alpha.1
+IMAGE_TAG ?= 0.20.0-rc1
```
MikeSpreitzer, line 17: Why does this PR even care about this `make` variable? Isn't #1779 the only one that should care?

chart/templates/controller-manager.yaml
```diff
@@ -202,7 +202,8 @@ spec:
         - --leader-elect
         - --wds-name={{.Values.ControlPlaneName}}
         - --api-groups={{.Values.APIGroups}}
-        image: ghcr.io/kubestellar/kubestellar/controller-manager:0.20.0-alpha.1
+        - -v=2
+        image: ghcr.io/kubestellar/kubestellar/controller-manager:0.20.0-rc1
```
MikeSpreitzer, line 206: The tag here is just a placeholder (replaced by the release workflow and the local test automation), right? If so then it should explicitly say that it is a placeholder. Perhaps a tag like `will-be-replaced`.

config/manager/kustomization.yaml
```diff
@@ -6,4 +6,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: ghcr.io/kubestellar/kubestellar/controller-manager
-  newTag: 0.20.0-alpha.1
+  newTag: 0.20.0-rc1
```
MikeSpreitzer, line 9: Shouldn't this be a placeholder?

config/postcreate-hooks/kubestellar.yaml
```diff
@@ -184,7 +184,7 @@ spec:
               - kubestellar
               - oci://ghcr.io/kubestellar/kubestellar/controller-manager-chart
               - --version
-              - "0.20.0-alpha.1"
+              - "v0.20.0-rc1"
```
MikeSpreitzer, line 187: Shouldn't this be a placeholder?

docs/content/v0.20/example-wecs.md
```diff
@@ -4,29 +4,23 @@ The following steps show how to create new clusters and
 register them with the hub as descibed in the
 [official open cluster management docs](https://open-cluster-management.io/getting-started/installation/start-the-control-plane/).
 
-1. Run the following set of commands for creating a new kind cluster with name "cluster1" and registering it with the
+1. Run the following set of commands for creating two kind clusters with name `cluster1` and `cluster2` and registering it with the
```
MikeSpreitzer, line 7: "name" is singluar but we are talking plural here.
I would s/with name/named/

docs/content/v0.20/example-wecs.md
```diff
@@ -4,29 +4,23 @@ The following steps show how to create new clusters and
...
+   clusters=(cluster1 cluster2);
+   for cluster in "${clusters[@]}"; do
+      kind create cluster --name ${cluster}
+      kubectl config rename-context kind-${cluster} ${cluster}
+      clusteradm --context imbs1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk '{print $0 " --context '${cluster}' '${flags}'"}' | sh
+   done   
    ```
 
-   The last line grabs a token from the hub (`imbs1` context), and constructs the command to apply on the new cluster
+   The last line grabs a token from the hub (`imbs1` context), and constructs the command to apply on each cluster
```
MikeSpreitzer, line 20: The "each" part is coming from the loop, not the last line. I think the original text "the new cluster" was good here.

docs/content/v0.20/examples.md
```diff
@@ -1,7 +1,7 @@
 # KubeStellar Usage Examples
 
 **NOTE**: This is unmaintained material that has only been observed to work for the commit tagged as
-`v0.20.0-alpha.1`. CI regularly tests variants of scenarios 1 and 4 that exercise the copy of the repo that they are embedded in (rather than the copy tagged `v0.20.0-alpha.1`), and contributors can run these tests too; see [the e2e tests](../../../test/e2e).
+`v0.20.0-rc1`. CI regularly tests variants of scenarios 1 and 4 that exercise the copy of the repo that they are embedded in (rather than the copy tagged `v0.20.0-rc1`), and contributors can run these tests too; see [the e2e tests](../../../test/e2e).
```
MikeSpreitzer, line 4: Making this edit now is time-traveling. Line 3 says this is a report of testing done in the past. The plan for this text is to keep it a report of the last successfully tested release. It gets updated after that release is created and passes testing.

.goreleaser.yaml
pdettori: Good catch, fixed.

.goreleaser.yaml
pdettori: Probably not helpful. Removed.

.goreleaser.yaml
pdettori: Note also that when we use a tag with a pattern `v${major}.${minor}.${patch}-{something}` goreleaser creates a pre-release, and if the tag has the pattern `v${major}.${minor}.${patch}` then go release creates a release.  That is the result of using  `prerelease: auto` - see https://goreleaser.com/customization/release/#github

Makefile
pdettori: You are right, it should not care. This should be addressed by https://github.com/kubestellar/kubestellar/pull/1779 so I will just revert this to `0.20.0-alpha.1`

chart/templates/controller-manager.yaml
pdettori: ok, done

config/manager/kustomization.yaml
pdettori: ok, done. Used `will-be-replaced` here as well

docs/content/v0.20/example-wecs.md
pdettori: ok, done

docs/content/v0.20/example-wecs.md
pdettori: ok, updated.

docs/content/v0.20/examples.md
pdettori: ok, done

.github/workflows/goreleaser.yml
```diff
@@ -11,73 +11,54 @@ permissions:
...
       with:
         registry: ${{ env.REGISTRY }}
         username: ${{ github.actor }}
-        password: ${{ secrets.GH_ALL_PROJECT_TOKEN }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Package and push chart
+      run: |
+        make chart IMG=${{ env.REGISTRY }}/${{ env.CONTROLLER_MANAGER_IMAGE }}:${{github.ref_name}}
+        helm package ${{ env.CHART_PATH }} --destination . --version ${{github.ref_name}} --app-version ${{github.ref_name}}
```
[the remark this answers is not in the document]
MikeSpreitzer: My question was specifically about the Helm chart, where Chart.yaml identifies its "version" and "appVersion". I think those typically exclude the "v", right?

.goreleaser.yaml
MikeSpreitzer: If you mean the example at https://goreleaser.com/customization/ko/#example , that does not show that CGO_ENABLED MUST only appear in one place. Is there some doc that says specifying it in two places is forbidden?

Anyway, it looks like we are allowed to specify it in only one place and that is what you have now, so that's good.

config/postcreate-hooks/kubestellar.yaml
MikeSpreitzer: Oops, sorry, my original comment here was wrong. #1758 currently says this will be manually managed, not a placeholder.

.github/workflows/goreleaser.yml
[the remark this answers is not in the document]
MikeSpreitzer: Examples:

```
mspreitz@mjs13 kubestellar % helm show chart oci://registry-1.docker.io/bitnamicharts/kafka
...
appVersion: 3.6.1
...
version: 26.10.0
```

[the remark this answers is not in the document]
MikeSpreitzer: `helm search hub wordpress` shows a bunch of chart versions without leading "v"; appVersion mostly lacks it but there are exceptions.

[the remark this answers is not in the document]
MikeSpreitzer: I see that oci://ghcr.io/kubestellar/kubeflex/chart/kubeflex-operator has version and appVersion _with_ the leading "v". So I think that demonstrates that we _can_ do it either way.

MikeSpreitzer:
The `examples.md` file is backward-looking, not forward-looking. It reports on a tested release from the past. This is where we tell users how to use a release. It should not be updated until we are confident that we have a good quality new release.

config/manager/kustomization.yaml
pdettori: using `will-be-replaced` here and in the chart broke the e2e tests, so I reverted to `0.20.0-alpha.1`. This is going to be addressed by #1779 so I am not trying to solve this in this PR.

.github/workflows/goreleaser.yml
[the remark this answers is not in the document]
pdettori: Ok, but I am convinced now that even if the leading "v" can be used, the norm is to not have it, so I updated the action to remove it for the chart version and app version.

MikeSpreitzer:
I still see changes to examples.md here. Didn't you just agree that it is not time to change that yet?

config/postcreate-hooks/kubestellar.yaml
ezrasilvera, line 187: Are we going to manually change this on every release ?  Can't we `sed` it or something through the Make ?

chart/templates/controller-manager.yaml
```diff
@@ -202,6 +202,7 @@ spec:
         - --leader-elect
         - --wds-name={{.Values.ControlPlaneName}}
         - --api-groups={{.Values.APIGroups}}
+        - -v=2
```
ezrasilvera, line 205: Maybe also allow the verbose level to come from values?  Will be convenient for debugging

docs/content/v0.20/examples.md
```diff
@@ -17,6 +17,8 @@ The following steps establish an initial state used in the examples below.
    kflex init --create-kind
    ```
 
+  If you are installing KubeStellar on an existing OpenShift cluster, just use the command `kflex init`.
```
ezrasilvera, line 20: Is this true only for OpenShift or for any "existing cluster" ?

docs/content/v0.20/examples.md
```diff
@@ -17,6 +17,8 @@ The following steps establish an initial state used in the examples below.
...
+
 2. Update the post-create-hooks in KubeFlex to install kubestellar with the v0.20.0-alpha.1 images:
```
ezrasilvera, line 22: Why still alpha.1 ?

pdettori:
> I still see changes to examples.md here. Didn't you just agree that it is not time to change that yet?

@MikeSpreitzer I reverted examples.md back to what is in main. Will update after testing the rc.

chart/templates/controller-manager.yaml
pdettori: I suggest we leave this to another PR as there are changes introduced by #1779 I think we should merge first.

config/postcreate-hooks/kubestellar.yaml
pdettori: @ezrasilvera we do not have a process for that for now. We discussed several approaches (including publishing a helm chart with the release). But for the time being it needs to have the same tag of the release being published so it can be referenced later in the examples after testing the RC and updating the examples. Since this is a larger issue with PCHs I suggest to leave this for successive PRs.

docs/content/v0.20/examples.md
MikeSpreitzer: About the remark about how to cope with OpenShift:

(A) Belongs in a different place than a report on a past successful test.

(B) I am OK with just having a binary distinction between ordinary Kubernetes clusters and OpenShift clusters. OpenShift is by far the leading example of clusters that defy expectations, and we will not soon have enough experience with any others.

PR #1787, opened 2024-02-20
📖 Note dependency cycle between ks/ks and ks/OTP

MikeSpreitzer:
## Summary
This PR updates packaging.md to be accurate about the current status of the ocm-transport-plugin and its dependency cycle with ks/ks.

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/v0.20/packaging.md
+++ b/docs/content/v0.20/packaging.md
@@ -2,16 +2,23 @@
 
 ## Outline of GitHub repositories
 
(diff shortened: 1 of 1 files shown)


ezrasilvera:
@MikeSpreitzer not related to your changes - there are few inaccuracies  in the doc for example the OCM status addon image is not "is built and published using make"  but rather goreleaser.  Do we merge this PR and update the doc in following PRs ? (cc: @pdettori )

MikeSpreitzer:
@ezrasilvera : yes, let's handle other inaccuracies in a separate PR. BTW, can you list them or characterize them (e.g., are they about recent changes in ocm-status-addon)?

PR #1789, opened 2024-02-21
✨ Add some more error wrapping

MikeSpreitzer:
## Summary
This PR adds more error wrapping in the binding controller.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/binding/bindingpolicy.go
+++ b/pkg/binding/bindingpolicy.go
@@ -67,7 +67,7 @@ const (
 func (c *Controller) handleBindingPolicy(ctx context.Context, obj runtime.Object) error {
 	bindingPolicy, err := runtimeObjectToBindingPolicy(obj)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to convert runtime.Object to BindingPolicy: %w", err)
 	}
 
 	// handle requeing for changes in bindingpolicy, excluding deletion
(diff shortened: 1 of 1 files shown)


pkg/binding/bindingpolicy.go
```diff
@@ -67,15 +67,15 @@ const (
 func (c *Controller) handleBindingPolicy(ctx context.Context, obj runtime.Object) error {
 	bindingPolicy, err := runtimeObjectToBindingPolicy(obj)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to convert runtime.Object to BindingPolicy: %s", err)
```
nirrozenbaum, line 70: %w instead of %s

nirrozenbaum: As of Go 1.13 (or earlier if you use [golang.org/x/xerrors](https://godoc.org/golang.org/x/xerrors)), you can use the %w verb which wraps the error such that it can later be unwrapped with [errors.Unwrap](https://golang.org/pkg/errors/#Unwrap), and so that it can be considered with [errors.Is](https://golang.org/pkg/errors/#Is) and [errors.As](https://golang.org/pkg/errors/#As).

nirrozenbaum: from chatgpt:

image

MikeSpreitzer: Yeah, too sleepy. Fixing.

MikeSpreitzer:
The force-push to 0820fc8ca switches to using `%w` instead of `%s`.

PR #1793, opened 2024-02-21
:bug: binding controller: make sure deleted objects are detectable

vMaroon:
## Summary
During tests in #1765 and by @effi-ofer, it was found that deleted objects that go through controller::enqueueObject are later not detectable in ::isBeingDeleted due to losing the deletiontimestamp.

This would bug out whoever expects `isBeingDeleted` to return true on deleted objects.

## Related issue(s)

Fixes #


effi-ofer:
Do we need similar logic in the reconcile?  https://github.com/kubestellar/kubestellar/blob/edc3241ea2f204e7877268b6b584fd893d5d0f20/pkg/binding/controller.go#L497

PR #1797, opened 2024-02-22
📖 update examples.md for v0.20.0-rc2

pdettori:
## Summary

- Updates references to release 
- minor updates to example using the  --resource filter with "host" type CP

### DIFF
--- a/docs/content/v0.20/example-wecs.md
+++ b/docs/content/v0.20/example-wecs.md
@@ -8,7 +8,7 @@ register them with the hub as descibed in the
 OCM hub. This cluster will act as a workload cluster.
 
    ```shell
-   # set flags to "" if you have installed KubeStellar on an OpenShift cluster
+   : set flags to "" if you have installed KubeStellar on an OpenShift cluster
    flags="--force-internal-endpoint-lookup"
    clusters=(cluster1 cluster2);
    for cluster in "${clusters[@]}"; do
--- a/docs/content/v0.20/example-wecs.md
+++ b/docs/content/v0.20/example-wecs.md
@@ -8,7 +8,8 @@ register them with the hub as descibed in the
 OCM hub. This cluster will act as a workload cluster.
 
    ```shell
-   flags="--force-internal-endpoint-lookup" # set this to "" if you have installed KubeStellar on an OpenShift cluster
+   # set flags to "" if you have installed KubeStellar on an OpenShift cluster
+   flags="--force-internal-endpoint-lookup"
    clusters=(cluster1 cluster2);
    for cluster in "${clusters[@]}"; do
       kind create cluster --name ${cluster}
--- a/docs/content/v0.20/example-wecs.md
+++ b/docs/content/v0.20/example-wecs.md
@@ -4,8 +4,7 @@ The following steps show how to create new clusters and
 register them with the hub as descibed in the
 [official open cluster management docs](https://open-cluster-management.io/getting-started/installation/start-the-control-plane/).
 
-1. Run the following set of commands for creating two kind clusters named `cluster1` and `cluster2` and registering them with the
-OCM hub. This cluster will act as a workload cluster.
+1. Execute the following commands to create two kind clusters, named `cluster1` and `cluster2`, and register them with the OCM hub. These clusters will serve as workload clusters. If you have previously executed these commands, you might already have contexts named `cluster1` and `cluster2`. If so, you can remove these contexts using the commands `kubectl config delete-context cluster1` and `kubectl config delete-context cluster2`.
 
    ```shell
    : set flags to "" if you have installed KubeStellar on an OpenShift cluster
--- a/docs/content/v0.20/examples.md
+++ b/docs/content/v0.20/examples.md
@@ -1,7 +1,7 @@
 # KubeStellar Usage Examples
 
 **NOTE**: This is unmaintained material that has only been observed to work for the commit tagged as
-`v0.20.0`. CI regularly tests variants of scenarios 1 and 4 that exercise the copy of the repo that they are embedded in (rather than the copy tagged `v0.20.0`), and contributors can run these tests too; see [the e2e tests](../../../test/e2e).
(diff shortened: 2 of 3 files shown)


pdettori:
NOTE: this should be merged after v0.20.0 is published

docs/content/v0.20/examples.md
```diff
@@ -17,11 +17,13 @@ The following steps establish an initial state used in the examples below.
    kflex init --create-kind
    ```
 
-2. Update the post-create-hooks in KubeFlex to install kubestellar with the v0.20.0-alpha.1 images:
+  If you are installing KubeStellar on an existing OpenShift cluster, just use the command `kflex init`.
```
MikeSpreitzer, line 20: This needs to be indented three spaces in order for markdown to handle it properly.
The attached screen snapshot shows what I currently see, which is this line not visually nested under "1.".

Screenshot 2024-02-21 at 11 16 49 PM

pdettori: ok, done

docs/content/v0.20/examples.md
```diff
@@ -65,6 +66,22 @@ statefulset created in the imbs1-system namespace.
...
+kubeflex-system      deployment.apps/kubeflex-controller-manager      1/1     1            1           22h
+local-path-storage   deployment.apps/local-path-provisioner           1/1     1            1           22h
+wds1-system          deployment.apps/kube-apiserver                   1/1     1            1           22m
+wds1-system          deployment.apps/kube-controller-manager          1/1     1            1           22m
+wds1-system          deployment.apps/kubestellar-controller-manager   1/1     1            1           21m
+
+NAMESPACE         NAME                                   READY   AGE
+imbs1-system      statefulset.apps/vcluster              1/1     11h
+kubeflex-system   statefulset.apps/postgres-postgresql   1/1     22h
+```
```
MikeSpreitzer, line 84: Indent three spaces.

docs/content/v0.20/examples.md
```diff
@@ -199,12 +216,43 @@ EOF
...
+clusters=(wds2 cluster1 cluster2);
+  for cluster in "${clusters[@]}"; do
+  kubectl --context ${cluster} apply -f https://raw.githubusercontent.com/project-codeflare/multi-cluster-app-dispatcher/v1.39.0/config/crd/bases/workload.codeflare.dev_appwrappers.yaml
+done
+```
+
+Apply the kubestellar controller-manager helm chart with the option to allow only delivery of objects with api group `workload.codeflare.dev`
+
+```shell
+helm upgrade --install -n wds2-system kubestellar oci://ghcr.io/kubestellar/kubestellar/controller-manager-chart --version 0.20.0-rc1 --set ControlPlaneName=wds2 --set APIGroups=workload.codeflare.dev
```
MikeSpreitzer, line 253: I do not see an explicit kubeconfig context specification on this command line. I think we should stick with making that explicit. For the `helm` command this would be specified with the `--kube-context` command line flag.

docs/content/v0.20/examples.md
pdettori: fixed

docs/content/v0.20/examples.md
pdettori: fixed

docs/content/v0.20/example-wecs.md
```diff
@@ -8,7 +8,8 @@ register them with the hub as descibed in the
 OCM hub. This cluster will act as a workload cluster.
 
    ```shell
-   flags="--force-internal-endpoint-lookup" # set this to "" if you have installed KubeStellar on an OpenShift cluster
+   # set flags to "" if you have installed KubeStellar on an OpenShift cluster
```
MikeSpreitzer, line 11: `#` is not a command in zsh.
`:` is a command in zsh, sh, and bash.

pdettori: you mean a comment, right ?

docs/content/v0.20/examples.md
```diff
@@ -59,29 +59,29 @@ manager which connects to the `wds1` front-end and the `imbs1` OCM control plane
 
 6. Follow the steps to [create and register two clusters with OCM](example-wecs.md).
 
-7. (optional) Check all deployments and statefulsets running in the hosting cluster. Expect to
-see the wds1 kubestellar-controller-manager created in the wds1-system namespace and the imbs1
-statefulset created in the imbs1-system namespace.
+7. (optional) Check relevant deployments and statefulsets running in the hosting cluster. Expect to
+see the `kubestellar-controller-manager` in the `wds1-system` namespace and the 
+statefulset `vcluster` in the `imbs1-system` namespace, both in a `READY` state.
```
MikeSpreitzer, line 64: The wording here is off. It implies that tables have a STATE column that might or might not hold a value "READY". I would reword, something like

> ... both fully ready

or

> ... both fully ready (number of available replicas equals number of desired replicas).

MikeSpreitzer:
This is progress. Some other issues that could be addressed, do not need to be addressed in this PR:

- examples.md and example-wecs.md do not warn users who are trying them for a second time about deleting the kubeconfig contexts `cluster1` and `cluster2`.
- thirdparties.md is for Linux on amd64, no variability, no warning to users that are using something else.

docs/content/v0.20/example-wecs.md
pdettori: fixed

docs/content/v0.20/examples.md
pdettori: fixed

docs/content/v0.20/example-wecs.md
MikeSpreitzer: Technically, ':' does not mark a comment; rather, it is a command that does nothing. There are important differences. As a command, the usual command-line parsing and processing is done (rather than special skipping for comments).

MikeSpreitzer (commented):
The title could be adjusted for -rc2. Other than that,
/lgtm
Thanks!

PR #1799, opened 2024-02-22
📖 Make examples.md reference its own release instead of a past one

MikeSpreitzer:
## Summary
This PR changes the tense of the examples document, to reference its own release (adding self-reference problems) rather than a past release.

## Related issue(s)

Fixes #


pdettori:
This looks good, should we also update `config/postcreate-hooks/kubestellar.yaml` as part of the same PR ?

MikeSpreitzer:
The PCH does not have to be updated in the same PR. I would leave it separate.

pdettori (approved):
/lgtm

Issue #1451, opened 2023-12-11
feature: add denaturing for Service in controllers


pdettori:
### Feature Description

This is a feature that is required when switching from kcp to KubeFlex-managed API servers. 

A regular Kube API server sets cluster networking-dependent fields that should not be propagated down to the WECs.

This can be tested in a KubeFlex-backed space:

Apply a service:
```
apiVersion: v1
kind: Service
metadata:
  name: hello-world
  labels:
    app: hello-world
spec:
  selector:
    app: hello-world
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
  type: NodePort
 ``` 
After applying the service, several fields are filled in with values that are specific to the network config for the API server in the WDS.

```
apiVersion: v1
kind: Service
metadata:
  labels:
    app: hello-world
  name: hello-world
  namespace: default
spec:
  clusterIP: 10.96.51.158
  clusterIPs:
  - 10.96.51.158
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - nodePort: 31429
    port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: hello-world
  sessionAffinity: None
  type: NodePort
 ``` 
 
 The values on these fields are not valid in most cases for the target WEC.  This behavior cannot be controlled by disabling kube controllers on the WDS space.

### Proposed Solution

Need to nullify those fields in placement translator when using non-KCP based partially denatured spaces such as KubeFlex.

### Want to contribute?

- [ ] I would like to work on this issue.

pdettori:
This is taken care of in new 0.20 code - see https://github.com/kubestellar/kubestellar/blob/971986355b75adec42f602e27f73a6393ad828b9/pkg/ocm/ocm.go#L87-L91

@nirrozenbaum  @vMaroon  : We just have to make sure this is also preserved in the new transport code

nirrozenbaum:
@pdettori noted to verify we don't miss this.
I think the best thing to do in order to validate it's preserved is to add e2e test that covers this.
shouldn't be a big effort considering we already have some e2e tests.. just need to distribute a service as above.
@ezrasilvera @effi-ofer wdyt? 

@pdettori as a side question - what happens if we don't nullify these fields? is the agent failing to apply the service on the WEC? do you know if there is an error and what is the error?

nirrozenbaum:
[...]

PR #1803, opened 2024-02-23
✨ Add filtering of Job.spec.template.metadata.labels

### LINKED ISSUE #1766: feature:  add denature for Jobs in the WDS
### Feature Description

This is a feature that is required when creating a job inside a WDS

A regular Kube API server sets some labels in the job object that should not be propagated down to the WECs.

This can be tested in a WDS: 

Apply a job:

```
apiVersion: batch/v1
kind: Job
metadata:
  name: pi
  namespace: test-kfp
  labels:
    app.kubernetes.io/name: nginx-singleton
spec:
  template:
    spec:
      containers:
      - name: pi
        image: perl:5.34.0
        command: ["perl",  "-Mbignum=bpi", "-wle", "print bpi(2000)"]
      restartPolicy: Never
  backoffLimit: 4
```

After applying the job, several labels are added by the API server in the WDS.

```
apiVersion: batch/v1
kind: Job
metadata:
  annotations:
    batch.kubernetes.io/job-tracking: ""
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"batch/v1","kind":"Job","metadata":{"annotations":{},"labels":{"app.kubernetes.io/name":"nginx-singleton"},"name":"pi","namespace":"test-kfp"},"spec":{"backoffLimit":4,"template":{"spec":{"containers":[{"command":["perl","-Mbignum=bpi","-wle","print bpi(2000)"],"image":"perl:5.34.0","name":"pi"}],"restartPolicy":"Never"}}}}
  creationTimestamp: "2024-02-15T20:57:52Z"
  generation: 1
  labels:
    app.kubernetes.io/name: nginx-singleton
  name: pi
  namespace: test-kfp
  resourceVersion: "1281831"
  uid: 1515c04b-7406-40b5-b799-46a536258999
spec:
  backoffLimit: 4
  completionMode: NonIndexed
  completions: 1
  parallelism: 1
  selector:
    matchLabels:
      batch.kubernetes.io/controller-uid: 1515c04b-7406-40b5-b799-46a536258999
  suspend: false
  template:
    metadata:
      creationTimestamp: null
      labels:
        batch.kubernetes.io/controller-uid: 1515c04b-7406-40b5-b799-46a536258999
        batch.kubernetes.io/job-name: pi
        controller-uid: 1515c04b-7406-40b5-b799-46a536258999
        job-name: pi
    spec:
      containers:
      - command:
        - perl
        - -Mbignum=bpi
        - -wle
        - print bpi(2000)
        image: perl:5.34.0
        imagePullPolicy: IfNotPresent
        name: pi
        resources: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      restartPolicy: Never
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
status: {}
```

The Manifestwork shows the following error: 

```
 resourceStatus:
    manifests:
    - conditions:
      - lastTransitionTime: "2024-02-15T20:57:52Z"
        message: 'Failed to apply manifest: Job.batch "pi" is invalid: [spec.template.metadata.labels[controller-uid]:
          Invalid value: map[string]string{"batch.kubernetes.io/controller-uid":"1515c04b-7406-40b5-b799-46a536258999",
          "batch.kubernetes.io/job-name":"pi", "controller-uid":"1515c04b-7406-40b5-b799-46a536258999",
          "job-name":"pi"}: must be ''466083b2-5542-44bf-8452-3d8c22c6a162'', spec.template.metadata.labels[batch.kubernetes.io/controller-uid]:
          Invalid value: map[string]string{"batch.kubernetes.io/controller-uid":"1515c04b-7406-40b5-b799-46a536258999",
          "batch.kubernetes.io/job-name":"pi", "controller-uid":"1515c04b-7406-40b5-b799-46a536258999",
          "job-name":"pi"}: must be ''466083b2-5542-44bf-8452-3d8c22c6a162'', spec.selector:
          Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"batch.kubernetes.io/controller-uid":"1515c04b-7406-40b5-b799-46a536258999"},
          MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: `selector` not auto-generated]'
        reason: AppliedManifestFailed
        status: "False"
        type: Applied
      - lastTransitionTime: "2024-02-15T20:57:52Z"
        message: Resource is not available
        reason: ResourceNotAvailable
        status: "False"
        type: Available
      resourceMeta:
        group: batch
        kind: Job
        name: pi
        namespace: test-kfp
        ordinal: 0
        resource: jobs
        version: v1
      statusFeedback: {}

```

### Proposed Solution

 The added labels cannot be re-used in another cluster and need to be zero-ed

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR updates the generic transport code to filter out of a JobSpec's template.metadata.labels the ones that the kube apiserver manages.

This PR does not update the old wrapping code that is about to be removed in the switch to pluggable transport.

## Related issue(s)

Fixes #1766

### DIFF
--- a/pkg/transport/generic_transport_controller.go
+++ b/pkg/transport/generic_transport_controller.go
@@ -19,6 +19,7 @@ package transport
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -371,6 +372,7 @@ func (c *genericTransportController) updateWrappedObjectsAndFinalizer(ctx contex
 }
 
 func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, binding *v1alpha1.Binding) ([]*unstructured.Unstructured, error) {
+	logger := klog.FromContext(ctx)
 	objectsToPropagate := make([]*unstructured.Unstructured, 0)
 	// add cluster-scoped objects to the 'objectsToPropagate' slice
 	for _, clusterScopedObject := range binding.Spec.Workload.ClusterScope {
@@ -384,7 +386,7 @@ func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, bind
 			if err != nil {
 				return nil, fmt.Errorf("failed to get required cluster-scoped object '%s' with gvr %s from WDS - %w", objectName, gvr, err)
 			}
-			objectsToPropagate = append(objectsToPropagate, cleanObject(object))
+			objectsToPropagate = append(objectsToPropagate, cleanObject(logger, object))
 		}
 	}
 	// add namespace-scoped objects to the 'objectsToPropagate' slice
@@ -401,7 +403,7 @@ func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, bind
 					return nil, fmt.Errorf("failed to get required namespace-scoped object '%s' in namespace '%s' with gvr '%s' from WDS - %w", objectName,
 						objectsByNamespace.Namespace, gvr, err)
 				}
-				objectsToPropagate = append(objectsToPropagate, cleanObject(object))
+				objectsToPropagate = append(objectsToPropagate, cleanObject(logger, object))
 			}
 		}
 	}
@@ -594,7 +596,7 @@ func setLabel(object metav1.Object, key string, value any) {
 }
 
 // cleanObject is a function to clean object before adding it to a wrapped object. these fields shouldn't be propagated to WEC.
-func cleanObject(object *unstructured.Unstructured) *unstructured.Unstructured {
+func cleanObject(logger logr.Logger, object *unstructured.Unstructured) *unstructured.Unstructured {
 	objectCopy := object.DeepCopy() // don't modify object directly. create a copy before zeroing fields
 	objectCopy.SetManagedFields(nil)
 	objectCopy.SetFinalizers(nil)
@@ -608,6 +610,25 @@ func cleanObject(object *unstructured.Unstructured) *unstructured.Unstructured {
 	delete(annotations, "kubectl.kubernetes.io/last-applied-configuration")
 	objectCopy.SetAnnotations(annotations)
 
-	return objectCopy
+	gvk := objectCopy.GroupVersionKind()
+	if gvk.Group != "apps" || gvk.Kind != "Job" || !strings.HasPrefix(gvk.Version, "v1") {
+		return objectCopy
+	}
 
+	objectU := objectCopy.UnstructuredContent()
+	podLabels, found, err := unstructured.NestedMap(objectU, "spec", "template", "metadata", "labels")
+	if err != nil || !found {
+		logger.V(3).Error(nil, "Job lacks expected template labels", "namespace", objectCopy.GetNamespace(), "name", objectCopy.GetName(), "found", found, "err", err)
+		return objectCopy
+	}
+	delete(podLabels, "batch.kubernetes.io/controller-uid")
+	delete(podLabels, "batch.kubernetes.io/job-name")
+	delete(podLabels, "controller-uid")
+	delete(podLabels, "job-name")
+	err = unstructured.SetNestedMap(objectU, podLabels, "spec", "template", "metadata", "labels")
+	if err != nil {
+		logger.Error(err, "Inconceivable! Failed to update existing pod labels in Job", "namespace", objectCopy.GetNamespace(), "name", objectCopy.GetName())
+	}
+	objectCopy.SetUnstructuredContent(objectU)
+	return objectCopy
 }


pkg/transport/generic_transport_controller.go
```diff
@@ -608,6 +610,25 @@ func cleanObject(object *unstructured.Unstructured) *unstructured.Unstructured {
 	delete(annotations, "kubectl.kubernetes.io/last-applied-configuration")
 	objectCopy.SetAnnotations(annotations)
 
-	return objectCopy
+	gvk := objectCopy.GroupVersionKind()
+	if gvk.Group != "apps" || gvk.Kind != "Job" || !strings.HasPrefix(gvk.Version, "v1") {
```
pdettori, line 614: Note that we have also another known case where we need to clean an object for a Kubernetes service - see https://github.com/kubestellar/kubestellar/issues/1451

nirrozenbaum:
please do not merge this PR.
we clearly specified which issues should be handle for PT release and this has moved to later release. this is doing changes to transport controller which are not mandatory for the PT release.

in addition, not related to previous comment - this is not the way I wanted to handle denaturing of resources. I was planning to write this code outside of transport controller and inject denaturing for different objects with some function.
I really think we don’t want to have dedicated code for each specific resource denaturing inside transport controller, that would make the controller code really”dirty”.

MikeSpreitzer:
I would not even call this "denaturing". I think of "denaturing" as removing behavior around the object. But the problem here is that the behavior is coming from the kube apiserver and the approach here is not removing that behavior. Instead it is injecting compensating behavior in the distribution path.

@nirrozenbaum : do please share the alternative approach that you are thinking of.

ezrasilvera:
@pdettori @MikeSpreitzer @nirrozenbaum I think we should have a high level discussion first on the roles of KF and space providers VS KS.
For me it's up to the space provider to supply a WDS that supports X,Y,Z  . KS should take the resource deployed to that WDS without any concern of side effects of that resource on the WDS & its provider. 
For example KF-k8s-cp  advertises that is supports deployment of Pods, Deployments, etc.. and should list `Jobs`  as unsupported .  This needs to be fixed at the provider level/side.  
I think the same is true for resources that a controller on the WDS provider changes fields we want to maintain unchanged (e.g., Services). 
My opinion is tat KS shouldn't be aware of the details and capabilities of a specific space provider and therefore shouldn't be part of any "denaturing"  , "filtering" , or "injecting compensation behavior" , this will all require KS to treat WDS coming from different provider differently . 
E.g., if we have a provider that handles Jobs and Services - than KS doesn't need to do anything for those, right ?  If users decide to deploy Pods into the Vcluster space type - do we expect KS to "fix" the side effect of Pod deployed into a VCluster ?  
So to summarize - I think the above issue(s) should be treated at the KF level.

pdettori:
@ezrasilvera what you say makes sense in principle but then we do not have other viable solutions at this time for this issue. Remember that in the current architecture the space provider only manages lifecycle of API servers does not "gets in the middle" between a client (e.g. the KS controllers) and a managed API server.  At the same time, we want to avoid creating and managing forks of the Kubernetes API server. I don't think we can just say to our users "you can deploy any Kubernetes workload natively w/o wrapping" (which is a big part of the value proposition) but then you cannot use services and jobs, which would be a big limitation for many use cases we know about. It might be that in the future there would be the ability to compose API servers based on the API server library KEP, and there might be a customized server which would fit our needs, but we need a solution sooner than that. We discussed also about the "denaturing proxy" - but I think that has also its own challenges. Remember that the "denatured" Kube Api Server should not have a completely neutral behavior. There are actual certain behaviors that are needed - for example GC. That is why a minimum set of controllers need still to be active when we run the k8s control plane.

MikeSpreitzer:
@ezrasilvera : you outlined a simple story that would work but it is not very practical. I agree with @pdettori, we do not want to get into the business of maintaining a fork of kube-apiserver. I do not even see a practical way that we could maintain an alternative server built on the emerging kubernetes control-plane library --- it would still be a lot of stuff that 95% the same as kube-apiserver.

So I think that we are left with two practical possibilities: put scrubbing in the distribution path or introduce a denaturing proxy. I think the latter is the cleaner approach but it would take more work and may have limitations. I do not think the generic object GC has to be one of them --- that involves generic object features (ObjectMeta and TypeMeta) that every object has and a denaturing proxy could conceivably transform in a way that preserves GC behavior. I would like us to eventually get there. But I think that we should start with the easier solution, scrubbing in the distribution path. Later we can resume work on the denaturing proxy.

ezrasilvera:
@pdettori I understand the difficulty, but let's discuss if we can have fthis iltering outside the ks controlers (maybe not) and how can we avoid being  space type dependent and also not end up doing this for many resources making our code much more complex

MikeSpreitzer:
There is no need to make the scrubbing dependent on the space type. All the space types use the kube-apiserver. The only scrubbing-in-distribution that needs to be done is for the behavior that is built into the kube-apiserver.

nirrozenbaum:
while what @ezrasilvera was saying about KF responsibility might be correct, practically I agree with @pdettori and @MikeSpreitzer that this should be done in KS.

there is no doubt this is needed and at least for now, there is no other way around it other than implementing it in KS. 

my initial intention was just to allow me to show the alternative (not implementing this logic inside transport controller).
please give me few days and I'll link here a reference to code with my suggestion.

MikeSpreitzer:
See #1829 instead.

PR #1804, opened 2024-02-23
🐛 Exclude groups rather than groupVersions to watch

### LINKED ISSUE #1781: bug:  update api exclusion list in binding controller
### Describe the bug

additional logging revealed some additional API versions to exclude - see https://github.com/kubestellar/kubestellar/pull/1778#issuecomment-1951829088

### Steps To Reproduce

Look at https://github.com/kubestellar/kubestellar/pull/1778#issuecomment-1951829088

### Expected Behavior

Should exclude all versions for that API group.

### Additional Context

_No response_

waltforme:
## Summary
This PR excludes API groups, rather than groupVersions, when start informers in the binding controller.

I doubt there is a use case where in an API group, some versions should be watched and some other versions should not. So I think the group level granularity is sufficient.

Closes #1781.

### DIFF
--- a/pkg/binding/controller.go
+++ b/pkg/binding/controller.go
@@ -53,14 +53,13 @@ import (
 
 const controllerName = "Binding"
 
-// Resources group versions to exclude for watchers as they should not delivered to other clusters
-var excludedGroupVersions = map[string]bool{
-	"flowcontrol.apiserver.k8s.io/v1beta3": true,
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
This looks good to me but we are in code freeze right now.
/hold

MikeSpreitzer:
code freeze is over.
/unhold

MikeSpreitzer (approved):
/approve
Thanks!

PR #1805, opened 2024-02-23
🌱 Factor discovery out of the generic transport controller

MikeSpreitzer:
## Summary
This PR factors the generic transport controller to allow the caller to take responsibility for determining the wrapper GVR. This enables unit testing with fakes that do not sufficiently support discovery.

This PR also fixes a bug regarding worker ID.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/transport/generic_transport_controller.go
+++ b/pkg/transport/generic_transport_controller.go
@@ -53,7 +53,10 @@ const (
 	originOwnerGenerationAnnotation = "transport.kubestellar.io/originOwnerReferenceBindingGeneration"
 )
 
-// NewTransportController returns a new transport controller
+// NewTransportController returns a new transport controller.
+// This func is like NewTransportControllerForWrappedObjectGVR but first uses
+// the given transport and transportClientset to discover the GVR of wrapped objects.
+// The given transportDynamicClient is used to access the ITS.
 func NewTransportController(ctx context.Context, bindingInformer controlv1alpha1informers.BindingInformer, transport Transport,
 	wdsClientset ksclientset.Interface, wdsDynamicClient dynamic.Interface, transportClientset kubernetes.Interface,
 	transportDynamicClient dynamic.Interface, wdsName string) (*genericTransportController, error) {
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
/hold
for code freeze

pkg/transport/generic_transport_controller.go
```diff
@@ -203,6 +209,7 @@ func (c *genericTransportController) Run(ctx context.Context, workersCount int)
 	c.logger.Info("starting workers", "count", workersCount)
 	// Launch workers to process Binding
 	for i := 1; i <= workersCount; i++ {
+		i := i // gotta love go loop variables
```
nirrozenbaum, line 212: can you name the variable workerId instead of `i := i`? 
that is very confusing for code readers.

pkg/transport/generic_transport_controller.go
```diff
@@ -62,7 +62,13 @@ func NewTransportController(ctx context.Context, bindingInformer controlv1alpha1
 	if err != nil {
 		return nil, fmt.Errorf("failed to get wrapped object GVR - %w", err)
 	}
+	return NewTransportControllerForGVR(ctx, bindingInformer, transport, wdsClientset, wdsDynamicClient, transportClientset, transportDynamicClient, wdsName, wrappedObjectGVR), nil
+}
 
+// NewTransportController returns a new transport controller
+func NewTransportControllerForGVR(ctx context.Context, bindingInformer controlv1alpha1informers.BindingInformer, transport Transport,
```
nirrozenbaum, line 69: can we put this function private? (start with lower case)
unit test file should still be able to call it cause it is the same package right?

in addition, the name `ForGVR` is not so clear (GVR of what?). 
can we call it `ForWrappedObjectGVR`?

pkg/transport/generic_transport_controller.go
MikeSpreitzer: Code readers that do not know about go loop variables and do not want to learn.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: I changed the variable name and put a real explanation in the comment.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: I changed the variable name and put a real explanation in the comment.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: I changed the func name as suggested.
I think this func is something that is reasonable for non-test clients to use.
I added some explanation in the func comments.

MikeSpreitzer:
The force-push to 609b96eb2 makes changes in response to the review above.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: The second func is also simpler in that it can not return an error.

MikeSpreitzer: A caller of the second func is less complex --- no need to construct the k8s clientset, no need to handle an error.

MikeSpreitzer:
Closing in favor of the larger PR #1806, as requested by @ezrasilvera .

Issue #1786, opened 2024-02-20
Dependency cycle between kubestellar and ocm-transport-plugin repos


MikeSpreitzer:
Currently there is a cycle of dependencies. Although packaging.md does not represent it, the kubestellar repo depends on the ocm-transport-plugin (OTP) repo in order to have a functioning system. The kubestellar repo wants to have user instructions that reference the OTP repo, wants to have tests (for use by contributors and for use in CI) that reference the OTP repo; these may involve YAML that references the OTP repo.

Four approaches have been suggested so far.

1: Split the ks/ks repo into two repos, one for a lower layer and one for an upper layer. OTP would depend on only the lower layer repo, and the upper layer repo would depend on both the lower layer repo and the OTP repo.

2: Merge the OCM transport plugin into the kubestellar repo. Most likely the OTP part would appear concentrated under one subdirectory of the kubestellar repo and have a distinct go module. Having multiple go modules in one git repo is possible but tricky; we have observed other projects having trouble with this, but are not aware (at this time) of a clear show-stopper for our use of this idea.

3: Synchronized releases. Have a 1:1 relation between ks/ks releases and ks/OTP releases. This would mean that preparing a release pair involves manual editing of both repos to update them to reference the number of the coming release of the other (as well as themselves, sigh).

4: Staggered releases. This relies on the idea that release `X.Y.Z` is the capstone that appears after a series of test releases `X.Y.Z-rc1`, `X.Y.Z-rc2`, ... `X.Y.Z-rc$N`. Each OTP release is made after a ks release and references it. For example, OTP release `0.1.0-rc1` might be made after ks release `0.20.0-rc2` and reference that. Suppose that the last test release of OTP before `0.1.0` is `0.1.0-rc$M`. This references ks/ks release `X.Y.Z-$N`. By definition of `M` and `N`, testing of ks `X.Y.Z-rc$N` and testing of OTP `0.1.0-rc$M` succeeds. Then we make the plain releases ks `X.Y.Z` and OTP `0.1.0`; each of these is left referencing the successful test release of the other.

MikeSpreitzer:
We clearly have not converged on the thinking about testing in the staggered release approach. I disagree with what was done about https://github.com/kubestellar/kubestellar/pull/1780#discussion_r1501842654

MikeSpreitzer:
Let me elaborate on testing in the staggered release approach.

[...]

PR #1810, opened 2024-02-26
🐛 Add `go mod tidy` when building possibly-new integration

MikeSpreitzer:
## Summary
This PR adds a `go mod tidy` into the recipe for building ks/OTP using the local version of ks/ks. This may be needed when the local version of ks/ks has different dependencies than the version that ks/OTP normally builds with. I needed it while working on #1760.

I am not a fan of this building, see https://github.com/kubestellar/kubestellar/issues/1786#issuecomment-1964802834 ; as long as we are doing this building, this extra step is sometimes needed.

## Related issue(s)

Fixes #


ezrasilvera:
This seems risky.  Do we expect such cases ?

MikeSpreitzer:
@ezrasilvera : not sure what you mean by "this". Let's keep the #1786 discussion in 
#1786. As for this PR, I do not think it is risky, it is a standard bit of normalization, and yes, as I noted in the opening comment, I already needed it.

PR #1812, opened 2024-02-26
✨ added object denaturing extensible mechanism

### LINKED ISSUE #1451: feature: add denaturing for Service in controllers

nirrozenbaum:
This PR adds object denaturing extensible mechanism that is used in transport controller.

## Summary

This PR suggests an alternative approach to #1803. 
In short - This PR introduces a simple and extensible mechanism to denature objects before adding them to a wrapped object. this is done using a map from GVK to a "cleanObjectSpecifics" function with almost zero changes to the transport controller. the most important part is that even if we end up with tens of objects that have to be denatured, this would still be outside of the transport controller code, which keeps the controller code clean, easy to read and to maintain.

Ideally each different GVK would be in its own file (under `pkg/transport/denaturing`), which is also self documented.
it allows us to easily understand which objects require denaturing, and for each what is the specific denaturing logic.

In this PR I also implemented the Service denaturing using the suggested mechanism.
Intentionally, I did *NOT* implement the Job denaturing and left it for someone else to do, to make sure the mechanism is simple enough and easy for understanding also by others. 

## Related issue(s)

Fixes #1451

### DIFF
--- a/pkg/transport/denaturing/object_denaturing_map.go
+++ b/pkg/transport/denaturing/object_denaturing_map.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2024 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package denaturing
+
+import (
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// cleanObjectSpecificsFunction is a function for cleaning fields from a specific object.
+// The function cleans the specific fields in place (object is modified).
+// If the object was retrieved using a lister, it's the caller responsibility
+// to do a DeepCopy before calling this function.
+type cleanObjectSpecificsFunction func(object *unstructured.Unstructured)
+
+func NewObjectDenaturingMap() *ObjectDenaturingMap {
+	serviceTypeMeta := v1.TypeMeta{Kind: "Service", APIVersion: "v1"}
+
+	denaturingMap := map[string]cleanObjectSpecificsFunction{
+		serviceTypeMeta.GroupVersionKind().String(): cleanServiceFields,
+	}
+
+	return &ObjectDenaturingMap{
+		gvkToDenaturingFunc: denaturingMap,
+	}
+}
+
+type ObjectDenaturingMap struct {
+	gvkToDenaturingFunc map[string]cleanObjectSpecificsFunction // map from GVK as string to clean object function
+}
+
+func (denaturingMap *ObjectDenaturingMap) CleanObjectSpecifics(object *unstructured.Unstructured) {
+	gvkAsString := object.GetObjectKind().GroupVersionKind().String()
+	denaturingFunction, found := denaturingMap.gvkToDenaturingFunc[gvkAsString]
+	if !found {
+		return // if no denaturing function was defined for this gvk, do not clean any field
+	}
+	// otherwise, need to clean specific fields from this object
+	denaturingFunction(object)
+}
--- a/pkg/transport/generic_transport_controller.go
+++ b/pkg/transport/generic_transport_controller.go
@@ -43,6 +43,7 @@ import (
 	ksclientset "github.com/kubestellar/kubestellar/pkg/generated/clientset/versioned"
 	controlv1alpha1informers "github.com/kubestellar/kubestellar/pkg/generated/informers/externalversions/control/v1alpha1"
 	controlv1alpha1listers "github.com/kubestellar/kubestellar/pkg/generated/listers/control/v1alpha1"
+	"github.com/kubestellar/kubestellar/pkg/transport/denaturing"
 )
 
 const (
@@ -88,6 +89,7 @@ func NewTransportControllerForWrappedObjectGVR(ctx context.Context, bindingInfor
 		wdsClientset:                wdsClientset,
 		wdsDynamicClient:            wdsDynamicClient,
 		wdsName:                     wdsName,
+		objectDenaturingMap:         denaturing.NewObjectDenaturingMap(),
 	}
 
 	transportController.logger.Info("Setting up event handlers")
@@ -156,6 +158,8 @@ type genericTransportController struct {
 	wdsClientset     ksclientset.Interface
 	wdsDynamicClient dynamic.Interface
 	wdsName          string
+
+	objectDenaturingMap *denaturing.ObjectDenaturingMap
 }
 
 // enqueueBinding takes an Binding resource and
@@ -395,7 +399,7 @@ func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, bind
 			if err != nil {
 				return nil, fmt.Errorf("failed to get required cluster-scoped object '%s' with gvr %s from WDS - %w", objectName, gvr, err)
 			}
-			objectsToPropagate = append(objectsToPropagate, cleanObject(object))
+			objectsToPropagate = append(objectsToPropagate, c.cleanObject(object))
 		}
 	}
 	// add namespace-scoped objects to the 'objectsToPropagate' slice
@@ -412,7 +416,7 @@ func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, bind
 					return nil, fmt.Errorf("failed to get required namespace-scoped object '%s' in namespace '%s' with gvr '%s' from WDS - %w", objectName,
 						objectsByNamespace.Namespace, gvr, err)
 				}
-				objectsToPropagate = append(objectsToPropagate, cleanObject(object))
+				objectsToPropagate = append(objectsToPropagate, c.cleanObject(object))
 			}
 		}
 	}
@@ -605,7 +609,7 @@ func setLabel(object metav1.Object, key string, value any) {
 }
 
 // cleanObject is a function to clean object before adding it to a wrapped object. these fields shouldn't be propagated to WEC.
-func cleanObject(object *unstructured.Unstructured) *unstructured.Unstructured {
+func (c *genericTransportController) cleanObject(object *unstructured.Unstructured) *unstructured.Unstructured {
 	objectCopy := object.DeepCopy() // don't modify object directly. create a copy before zeroing fields
 	objectCopy.SetManagedFields(nil)
 	objectCopy.SetFinalizers(nil)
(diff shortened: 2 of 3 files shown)


pkg/transport/denaturing/object_denaturing_map.go
```diff
@@ -0,0 +1,54 @@
...
+		serviceTypeMeta.GroupVersionKind().String(): cleanServiceFields,
+	}
+
+	return &ObjectDenaturingMap{
+		gvkToDenaturingFunc: denaturingMap,
+	}
+}
+
+type ObjectDenaturingMap struct {
+	gvkToDenaturingFunc map[string]cleanObjectSpecificsFunction // map from GVK as string to clean object function
```
MikeSpreitzer, line 43: I would use GVK instead of a stringified version here, to make the domain of the map more checkable at compile time.

MikeSpreitzer:
This would work.
For my taste, putting it in a distinct package is overkill.
I would just put the following the transport package. Separate files is OK, also not necessary as far as I am concerned.

```go
var scrubbers = map[metav1.GroupVersionKind]ScrubFunc{
    k8scoreapi.GroupVersion.WithKind("Secret"): scrubSecret,
    ]
```

MikeSpreitzer:
I dislike using the term "denaturing" for this feature because this feature is not changing how the objects appear or are treated in the WDS. This feature is about a transformation applied in transport.

MikeSpreitzer:
Also, FYI. This is a mistake we are making everywhere, but I will call it out so that we do not forget it forever. In Kubernetes API machinery, there is a concept like "major version". The v2 of an API group has disjoint objects from the v1. OTOH, v1, v1beta1, and v1alpha1 are all different views of the same objects.

nirrozenbaum:
> I dislike using the term "denaturing" for this feature because this feature is not changing how the objects appear or are treated in the WDS. Denaturing was defined as removing associated behavior (controllers). This feature is about a transformation applied in transport.

@MikeSpreitzer what about `filtering`? `cleaning`?

nirrozenbaum:
> This would work.
> 
> FYI, for my taste, putting it in a distinct package is overkill. I would just put the following the transport package. Separate files, `init` funcs in those files, is OK; also not necessary as far as I am concerned.
> 
> ```go
> var scrubbers = map[metav1.GroupVersionKind]ScrubFunc{
>     k8scoreapi.GroupVersion.WithKind("Secret"): scrubSecret,
>     }
> ```

personally I like the separation (a lot) which makes is easy for code readers/contributors to navigate to the place they are looking for. if that is not a concern, I prefer keeping the object filtering separated.

pkg/transport/denaturing/object_denaturing_map.go
nirrozenbaum: that would work. I will update the PR.

nirrozenbaum:
> Also, FYI. This is a mistake we are making everywhere, but I will call it out so that we do not forget it forever. In Kubernetes API machinery, there is a concept like "major version". The v2 of an API group has disjoint objects from the v1. OTOH, v1, v1beta1, and v1alpha1 are all different views of the same objects.
> 
> In the subject matter of this PR, scrubbers, it is appropriate to distinguish among minor versions, because they can have different representations (not _very_ different, but different enough to require different scrubbing). At least, in principle. I am not aware of any such important difference for Service/v1 and Job/v1 objects.

how do you suggest to handle this? 
obviously we can put two entries in the map, one key using v1 and another v2. 
but I got the impression that this was not your intention, so I'm asking.

nirrozenbaum:
manual e2e tests are working as expected. CICD fails due to unit-test failure.
need to investigate, could be that the test has to be updated according to this change.
until unit test is fixed, hold with this PR

/hold

MikeSpreitzer:
"filtering" or "cleaning" make sense to me.

MikeSpreitzer:
Regarding my comment about versioning in kube api-machinery, this PR is doing it right. It is other code that is doing it wrong.

MikeSpreitzer:
I do not expect the list of cleaners to get long.
I am OK with the high level of structure, just remarking that I would not be sad to see it go.

nirrozenbaum:
@MikeSpreitzer so to summarize the comments into actionable items:
- I will update the key of the map to be GVK instead of the stringified version of the GVK and validate it's working as expected.
- the term `denaturing` will be renamed to `filtering`.
- I need to check why unit test fails. btw, on this topic - the output of the unit test is very hard for understanding anything out of it (at least when it fails). I think this is a very important point. 

other than the above, I think we're ok with keeping as is.
let me know if I missed something.

nirrozenbaum:
attaching a screenshot of how it looks like when transport unit-test fails:
image (6)

MikeSpreitzer:
@nirrozenbaum : we are agreed.

I am looking into the unit test failure too. You can use your favorite debugger on unit tests.

`apiequality.Semantic.DeepEqual` is saying unequal when I think it should be equal, or I am overlooking something. I see one difference, `creationTimestamp` being `nil` vs. omitted from the `map[string]any`. But Semantic.DeepEqual is supposed to consider those equal.

MikeSpreitzer:
"github.com/google/go-cmp/cmp"
 has a Diff func that could be helpful

nirrozenbaum:
is it related to this line which I’ve added in this PR?
https://github.com/kubestellar/kubestellar/blob/eec4b47f38a3b9a111289852b6d78f574cd65f23/pkg/transport/generic_transport_controller.go#L621

It was added just to be on the safe side, but I think it can be omitted

MikeSpreitzer:
`ObjectMeta.CreationTimestamp` is a pointer. Best to set it to `nil`.

MikeSpreitzer:
I still do not understand.

```go
func (u *Unstructured) SetCreationTimestamp(timestamp metav1.Time) {
	ts, _ := timestamp.MarshalQueryParameter()
	if len(ts) == 0 || timestamp.Time.IsZero() {
		RemoveNestedField(u.Object, "metadata", "creationTimestamp")
		return
	}
	u.setNestedField(ts, "metadata", "creationTimestamp")
}
```

MikeSpreitzer:
Ah, it is the expectedObj that has the creationTimestamp set.

Screenshot 2024-02-26 at 5 59 03 PM

MikeSpreitzer:
See #1815

MikeSpreitzer:
Actually, see #1816

pkg/transport/generic_transport_controller.go
```diff
@@ -605,7 +609,7 @@ func setLabel(object metav1.Object, key string, value any) {
 }
 
 // cleanObject is a function to clean object before adding it to a wrapped object. these fields shouldn't be propagated to WEC.
-func cleanObject(object *unstructured.Unstructured) *unstructured.Unstructured {
+func (c *genericTransportController) cleanObject(object *unstructured.Unstructured) *unstructured.Unstructured {
```
MikeSpreitzer, line 612: Changing this from a free-standing func to a method on a controller is an unnecessary entanglement; it thwarted the approach I was taking in #1816 of cleaning the test objects as they were generated (which happened before the controller is created). I see that there is no meaningful difference among instances of ObjectDenaturingMap, no need to hold one in a field of the controller rather than hold one in a global variable; these things have no state that varies after construction.

nirrozenbaum:
> is it related to this line which I’ve added in this PR?
> 
> https://github.com/kubestellar/kubestellar/blob/eec4b47f38a3b9a111289852b6d78f574cd65f23/pkg/transport/generic_transport_controller.go#L621
> 
> It was added just to be on the safe side, but I think it can be omitted

@MikeSpreitzer does the test fails because of this line?

MikeSpreitzer:
@nirrozenbaum : yes, `objectCopy.SetCreationTimestamp(metav1.Time{})` causes the `creationTimestamp` entry to be removed from the `map[string]any` representing the ObjectMeta.

MikeSpreitzer:
I think that it is reasonable for the cleaning function to clear all the server-managed fields.

nirrozenbaum:
ok. then the fix is eady. no need for this line in the controller. 
let me try to remove it and see if it passed

MikeSpreitzer:
I think that it is reasonable for the cleaning function to clear all the server-managed fields.

MikeSpreitzer:
But if the test passes without that, then I suppose that is OK too.

nirrozenbaum:
yes, I agree. 
but let’s split into two different PRs.
I will remove this line and we’ll see if test passes.
then we can open a new issue and fix in separate PR.

pkg/transport/denaturing/object_denaturing_map.go
MikeSpreitzer: Thanks, looking forward to it.

pkg/transport/filtering/object_filtering_map.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+// cleanObjectSpecificsFunction is a function for cleaning fields from a specific object.
+// The function cleans the specific fields in place (object is modified).
+// If the object was retrieved using a lister, it's the caller responsibility
+// to do a DeepCopy before calling this function.
+type cleanObjectSpecificsFunction func(object *unstructured.Unstructured)
+
+func NewObjectFilteringMap() *ObjectFilteringMap {
+	serviceTypeMeta := v1.TypeMeta{Kind: "Service", APIVersion: "v1"}
```
MikeSpreitzer, line 32: FYI: Instead of a literal "v1" here, we could use the result of calling the `String` method of https://github.com/kubernetes/api/blob/v0.28.2/core/v1/register.go#L29

MikeSpreitzer (commented):
This is good enough to merge.
/lgtm
Thanks!

pkg/transport/filtering/object_filtering_map.go
nirrozenbaum: good comment. I changed to use directly `corev1.SchemeGroupVersion.WithKind("Service")` in the map as key.

nirrozenbaum:
@ezrasilvera I've added e2e test for validating this PR in #1824.
in #1824, we can clearly see that the e2e failed cause we didn't clean the fields from the service. see it here:
https://github.com/kubestellar/kubestellar/actions/runs/8078461596/job/22070833973?pr=1824#step:17:302

let's merge this PR and re-run 1824 to validate it's fixed.
if from any reason we see issues (I already tested it manually so I don't expect issues), I will fix on 1824

PR #1815, opened 2024-02-26
🐛 Ignore server-managed fields when testing objects

MikeSpreitzer:
## Summary
This PR updates the generic transport controller test so that the object equality test ignores fields managed by the server.

## Related issue(s)

Fixes #


MikeSpreitzer:
Actually, see #1816 instead.

PR #1816, opened 2024-02-26
🐛 Expect cleand objects

MikeSpreitzer:
## Summary
This PR updates the generic transport controller test to clean the expected objects.

This is an alternative to #1815, which probably will not work because the comparison is of unstructured objects.

## Related issue(s)

Fixes #


MikeSpreitzer:
This is not going to work either, now we need a controller in order to clean :-(.

PR #1818, opened 2024-02-27
✨ Add TestMain in pkg/transport

MikeSpreitzer:
## Summary
This PR adds a TestMain to pkg/transport, so that users of unit tests there can control the logging verbosity.

## Related issue(s)

Fixes #


nirrozenbaum:
why not in unit-test file?

PR #1822, opened 2024-02-28
🌱 Remove reticence and add debug logging in binding controller

MikeSpreitzer:
## Summary
This PR removes the suppression of enqueuing workload object references because it was a bug: nothing guarantees that the object reference is enqueud later.

This PR also makes some debug logging improvements, primarily adding some debug logging. Also some switching from the controller's logger to the worker's logger, to help investigate possible concurrency bugs.

## Related issue(s)

Fixes #


MikeSpreitzer:
See #1823 instead.

PR #1828, opened 2024-02-28
✨ Handle CRDs at scale

### LINKED ISSUE #1705: bug:  client-side throttling when starting dynamic informers for (many) new CRDs

waltforme:
## Summary
Large number of CRDs (e.g. in an OpenShift cluster) may cause issues like #1705 . This PR makes several changes to handle large number or CRDs.

### What's done in this PR
Three things are done in this PR:
1. Change the way that the binding controller handles CRD objects. Previously, for each CRD object, the controller computes the symmetric difference of existing resources and watched resources, in a holistic manner. That requires repeated requests to apiserver for discovery. With this change, the controller focuses on the CRD object only, in a incremental manner, which saves the requests to the apiserver and is more efficient.
2. Make the client-side rate limiter adaptive to: the number of GVRs in the apiserver when the binding controller starts. Together with 1, the client-side throttling is eliminated (see the 'Result' section below for details).
3. Write tests for the binding controller to test the changed behavior.

### Result
Before the change, I can observe the client-side throttling at the start of the binding controller.
- For a cluster which supports aggregated discovery, the throttling is primarily triggered when the reflectors are doing list and watch.
- For a cluster which doesn't support aggregated discovery, additional throttling (also longer waits) is trigger by the initial, intensive, reconciliation cycles for CRD objects.
```
I0228 14:55:43.122143  761795 request.go:629] Waited for 8.152398561s due to client-side throttling, not priority and fairness, request: GET:https://wds1.localtest.me:9443/apis/synthetic-crd.com/v1alpha1/cr11s?limit=500&resourceVersion=0
I0228 14:55:43.322485  761795 request.go:629] Waited for 8.352711234s due to client-side throttling, not priority and fairness, request: GET:https://wds1.localtest.me:9443/apis/synthetic-crd.com/v1alpha1/cr30s?limit=500&resourceVersion=0
I0228 14:55:43.522831  761795 request.go:629] Waited for 8.552861282s due to client-side throttling, not priority and fairness, request: GET:https://wds1.localtest.me:9443/apis/synthetic-crd.com/v1alpha1/cr72s?limit=500&resourceVersion=0
I0228 14:55:43.522865  761795 request.go:697] Waited for 8.552861282s due to client-side throttling, not priority and fairness, request: GET:https://wds1.localtest.me:9443/apis/synthetic-crd.com/v1alpha1/cr72s?limit=500&resourceVersion=0
I0228 14:55:43.722340  761795 request.go:629] Waited for 8.752317832s due to client-side throttling, not priority and fairness, request: GET:https://wds1.localtest.me:9443/apis/synthetic-crd.com/v1alpha1/cr25s?limit=500&resourceVersion=0
I0228 14:55:43.922619  761795 request.go:629] Waited for 8.952537192s due to client-side throttling, not priority and fairness, request: GET:https://wds1.localtest.me:9443/apis/synthetic-crd.com/v1alpha1/cr65s?limit=500&resourceVersion=0
I0228 14:55:44.122122  761795 request.go:629] Waited for 9.152017818s due to client-side throttling, not priority and fairness, request: GET:https://wds1.localtest.me:9443/apis/synthetic-crd.com/v1alpha1/cr21s?limit=500&resourceVersion=0
I0228 14:55:44.322700  761795 request.go:629] Waited for 9.352440908s due to client-side throttling, not priority and fairness, request: GET:https://wds1.localtest.me:9443/apis/synthetic-crd.com/v1alpha1/cr67s?limit=500&resourceVersion=0
I0228 14:55:44.521954  761795 request.go:629] Waited for 9.551632207s due to client-side throttling, not priority and fairness, request: GET:https://wds1.localtest.me:9443/apis/synthetic-crd.com/v1alpha1/cr39s?limit=500&resourceVersion=0
I0228 14:55:44.722205  761795 request.go:629] Waited for 9.750876057s due to client-side throttling, not priority and fairness, request: GET:https://wds1.localtest.me:9443/apis/synthetic-crd.com/v1alpha1/cr73s?limit=500&resourceVersion=0
I0228 14:55:44.722233  761795 request.go:697] Waited for 9.750876057s due to client-side throttling, not priority and fairness, request: GET:https://wds1.localtest.me:9443/apis/synthetic-crd.com/v1alpha1/cr73s?limit=500&resourceVersion=0
I0228 14:55:44.922686  761795 request.go:629] Waited for 9.951253423s due to client-side throttling, not priority and fairness, request: GET:https://wds1.localtest.me:9443/apis/synthetic-crd.com/v1alpha1/cr44s?limit=500&resourceVersion=0
I0228 14:55:45.122104  761795 request.go:629] Waited for 10.150663768s due to client-side throttling, not priority and fairness, request: GET:https://wds1.localtest.me:9443/apis/synthetic-crd.com/v1alpha1/cr76s?limit=500&resourceVersion=0
```
I synthesized the CRDs in group `synthetic-crd.com` above. I populated roughly 100 CRDs into the WDS.

After the changes, the client-side throttling is gone.

Closes #1705

### DIFF
--- a/pkg/binding/controller.go
+++ b/pkg/binding/controller.go
@@ -99,17 +99,26 @@ type Controller struct {
 func NewController(parentLogger logr.Logger, wdsRestConfig *rest.Config, imbsRestConfig *rest.Config,
 	wdsName string, allowedGroupsSet sets.Set[string]) (*Controller, error) {
 
-	dynamicClient, err := dynamic.NewForConfig(wdsRestConfig)
-	if err != nil {
-		return nil, err
-	}
-
 	kubernetesClient, err := kubernetes.NewForConfig(wdsRestConfig)
 	if err != nil {
 		return nil, err
 	}
 
 	extClient, err := apiextensionsclientset.NewForConfig(wdsRestConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	burst, err := countGVRsViaAggregatedDiscovery(wdsRestConfig) // use # of GVRs as burst is tested to be working well
+	if err != nil {
+		parentLogger.Error(err, "Not able to count GVRs via aggregated discovery") // but we can still continue
+	}
+	wdsRestConfig.Burst = adjustBurstIfNecessary(burst)
+
+	// dynamicClient needs higher burst because dynamicClient is repeatedly used to create informers
+	// for each of the GVRs, all at the beginning of the controller run
+	parentLogger.V(3).Info("Setting parameter of the client's token bucket rate limiter", "Burst", wdsRestConfig.Burst)
+	dynamicClient, err := dynamic.NewForConfig(wdsRestConfig)
 	if err != nil {
 		return nil, err
 	}
@@ -122,6 +131,39 @@ func NewController(parentLogger logr.Logger, wdsRestConfig *rest.Config, imbsRes
 	ocmClient := *ocm.GetOCMClient(imbsRestConfig)
 
 	return makeController(parentLogger, dynamicClient, kubernetesClient, extClient, ocmClientset, ocmClient, wdsName, allowedGroupsSet)
+}
+
+func countGVRsViaAggregatedDiscovery(config *rest.Config) (int, error) {
+	countClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return 0, fmt.Errorf("error creating client to count number of GVRs: %w", err)
+	}
+	if countClient.DiscoveryClient.UseLegacyDiscovery {
+		return 0, fmt.Errorf("client using aggregated discovery")
+	}
+	apiResourceLists, err := countClient.Discovery().ServerPreferredResources()
+	if err != nil {
+		return 0, fmt.Errorf("error listing server perferred resources: %w", err)
+	}
+	n := 0
+	for _, list := range apiResourceLists {
+		n += len(list.APIResources)
+	}
+	return n, nil
+}
+
+func adjustBurstIfNecessary(burst int) int {
+	// in case too small
+	if burst < rest.DefaultBurst {
+		return rest.DefaultBurst
+	}
+	// in case too large
+	// https://github.com/kubernetes/kubernetes/pull/105520/files
+	// https://github.com/kubernetes/kubernetes/blob/5d527dcf1265d7fcd0e6c8ec511ce16cc6a40699/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go#L477
+	if burst > 300 {
+		return 300
+	}
+	return burst
 }
 
 func makeController(parentLogger logr.Logger,
--- a/pkg/binding/controller.go
+++ b/pkg/binding/controller.go
@@ -572,7 +572,7 @@ func (c *Controller) reconcile(ctx context.Context, key util.Key) error {
 		logger.Info("Handled bindingpolicy", "object", util.RefToRuntimeObj(obj))
 		return nil
 	} else if util.IsCRD(obj) {
-		if err := c.handleCRD(obj); err != nil {
+		if err := c.handleCRD(ctx, obj); err != nil {
 			return fmt.Errorf("failed to handle CRD: %w", err) // error logging after this call
 			// will add name.
 		}
--- a/pkg/binding/controller.go
+++ b/pkg/binding/controller.go
@@ -35,6 +35,7 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/kubernetes"
@@ -109,12 +110,13 @@ func NewController(parentLogger logr.Logger, wdsRestConfig *rest.Config, imbsRes
 		return nil, err
 	}
 
-	burst, err := countGVRsViaAggregatedDiscovery(wdsRestConfig) // use # of GVRs as burst is tested to be working well
+	nGVRs, err := countGVRsViaAggregatedDiscovery(kubernetesClient)
 	if err != nil {
 		parentLogger.Error(err, "Not able to count GVRs via aggregated discovery") // but we can still continue
 	}
-	wdsRestConfig.Burst = adjustBurstIfNecessary(burst)
-	wdsRestConfig.QPS = adjustQPSIfNecessary(float32(burst) / 4)
+	// tuning the rate limiter based on the number of GVRs is tested to be working well
+	wdsRestConfig.Burst = computeBurstFromNumGVRs(nGVRs)
+	wdsRestConfig.QPS = computeQPSFromNumGVRs(nGVRs)
 	parentLogger.V(1).Info("Parameters of the client's token bucket rate limiter", "burst", wdsRestConfig.Burst, "qps", wdsRestConfig.QPS)
 
 	// dynamicClient needs higher rate than its default because dynamicClient is repeatedly used by the
@@ -134,15 +136,12 @@ func NewController(parentLogger logr.Logger, wdsRestConfig *rest.Config, imbsRes
 	return makeController(parentLogger, dynamicClient, kubernetesClient, extClient, ocmClientset, ocmClient, wdsName, allowedGroupsSet)
 }
 
-func countGVRsViaAggregatedDiscovery(config *rest.Config) (int, error) {
-	countClient, err := kubernetes.NewForConfig(config)
-	if err != nil {
-		return 0, fmt.Errorf("error creating client to count number of GVRs: %w", err)
+func countGVRsViaAggregatedDiscovery(countClient kubernetes.Interface) (int, error) {
+	dc := countClient.Discovery().(*discovery.DiscoveryClient)
+	if dc.UseLegacyDiscovery { // by default it should be false already, just double check
+		dc.UseLegacyDiscovery = false
 	}
-	if countClient.DiscoveryClient.UseLegacyDiscovery {
-		return 0, fmt.Errorf("client using aggregated discovery")
-	}
-	apiResourceLists, err := countClient.Discovery().ServerPreferredResources()
+	apiResourceLists, err := dc.ServerPreferredResources()
 	if err != nil {
 		return 0, fmt.Errorf("error listing server perferred resources: %w", err)
 	}
@@ -153,12 +152,13 @@ func countGVRsViaAggregatedDiscovery(config *rest.Config) (int, error) {
 	return n, nil
 }
 
-func adjustBurstIfNecessary(burst int) int {
-	// in case too small
+func computeBurstFromNumGVRs(nGVRs int) int {
+	burst := nGVRs
+	// in case too small, fall back to default
 	if burst < rest.DefaultBurst {
 		return rest.DefaultBurst
 	}
-	// in case too large
+	// in case too large, look at some value for reference
 	// https://github.com/kubernetes/kubernetes/blob/5d527dcf1265d7fcd0e6c8ec511ce16cc6a40699/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go#L477
 	if burst > 300 {
 		return 300
@@ -166,12 +166,13 @@ func adjustBurstIfNecessary(burst int) int {
 	return burst
 }
 
-func adjustQPSIfNecessary(qps float32) float32 {
-	// in case too small
+func computeQPSFromNumGVRs(nGVRs int) float32 {
+	qps := float32(nGVRs) / 4
+	// in case too small, fall back to default
 	if qps < rest.DefaultQPS {
 		return rest.DefaultQPS
 	}
-	// in case too large
+	// in case too large, look at some value for reference
 	// https://github.com/kubernetes/kubernetes/pull/105520/files
 	if qps > 50.0 {
 		return 50.0
@@ -268,12 +269,13 @@ func (c *Controller) run(ctx context.Context, workers int) error {
 			logger.V(1).Info("Ignoring APIResourceList", "groupVersion", list.GroupVersion)
 			continue
 		}
+		if !util.IsAPIGroupAllowed(gv.Group, c.allowedGroupsSet) {
+			logger.V(1).Info("No need to watch per user input", "groupVersion", list.GroupVersion)
+			continue
+		}
 		logger.V(1).Info("Working on APIResourceList", "groupVersion", list.GroupVersion, "numResources", len(list.APIResources))
 		for _, resource := range list.APIResources {
 			if _, excluded := excludedResourceNames[resource.Name]; excluded {
-				continue
-			}
-			if !util.IsAPIGroupAllowed(gv.Group, c.allowedGroupsSet) {
 				continue
 			}
 			informable := verbsSupportInformers(resource.Verbs)
--- a/pkg/binding/controller.go
+++ b/pkg/binding/controller.go
@@ -114,10 +114,11 @@ func NewController(parentLogger logr.Logger, wdsRestConfig *rest.Config, imbsRes
 		parentLogger.Error(err, "Not able to count GVRs via aggregated discovery") // but we can still continue
 	}
 	wdsRestConfig.Burst = adjustBurstIfNecessary(burst)
+	wdsRestConfig.QPS = adjustQPSIfNecessary(float32(burst) / 4)
+	parentLogger.V(1).Info("Parameters of the client's token bucket rate limiter", "burst", wdsRestConfig.Burst, "qps", wdsRestConfig.QPS)
 
-	// dynamicClient needs higher burst because dynamicClient is repeatedly used to create informers
-	// for each of the GVRs, all at the beginning of the controller run
-	parentLogger.V(3).Info("Setting parameter of the client's token bucket rate limiter", "Burst", wdsRestConfig.Burst)
+	// dynamicClient needs higher rate than its default because dynamicClient is repeatedly used by the
+	// reflectors for each of the GVRs, all at the beginning of the controller run
 	dynamicClient, err := dynamic.NewForConfig(wdsRestConfig)
 	if err != nil {
 		return nil, err
@@ -158,12 +159,24 @@ func adjustBurstIfNecessary(burst int) int {
 		return rest.DefaultBurst
 	}
 	// in case too large
-	// https://github.com/kubernetes/kubernetes/pull/105520/files
 	// https://github.com/kubernetes/kubernetes/blob/5d527dcf1265d7fcd0e6c8ec511ce16cc6a40699/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go#L477
 	if burst > 300 {
 		return 300
 	}
 	return burst
+}
+
+func adjustQPSIfNecessary(qps float32) float32 {
+	// in case too small
+	if qps < rest.DefaultQPS {
+		return rest.DefaultQPS
+	}
+	// in case too large
+	// https://github.com/kubernetes/kubernetes/pull/105520/files
+	if qps > 50.0 {
+		return 50.0
+	}
+	return qps
 }
 
 func makeController(parentLogger logr.Logger,
--- a/pkg/binding/crd.go
+++ b/pkg/binding/crd.go
@@ -19,7 +19,6 @@ package binding
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -58,15 +57,13 @@ func (c *Controller) handleCRD(obj runtime.Object) error {
 
 	key := util.KeyForGroupVersionKind(toStart.groupVersion.Group, toStart.groupVersion.Version, toStart.resource.Kind)
 
-	// toStartList, toStopList, err := c.checkAPIResourcesForUpdates()
-	// if err != nil {
-	// 	return err
-	// }
 	toStartList, toStopList := []APIResource{}, []string{}
 	if isBeingDeleted(obj) {
 		toStopList = append(toStopList, key)
 	} else {
-		toStartList = append(toStartList, toStart)
+		if c.shouldWatch(toStart) {
+			toStartList = append(toStartList, toStart)
+		}
 	}
 
 	go c.startInformersForNewAPIResources(toStartList)
@@ -87,69 +84,22 @@ func (c *Controller) handleCRD(obj runtime.Object) error {
 	return nil
 }
 
-// checks what APIs need starting new informers or stopping informers.
-// Returns a list of APIResources for informers to start and a list of keys for infomers to stop
-func (c *Controller) checkAPIResourcesForUpdates() ([]APIResource, []string, error) {
-	toStart := []APIResource{}
-	toStop := []string{}
-
-	// tracking keys are used to detect what API resources have been removed
-	trackingKeys := map[string]bool{}
-	for k := range c.informers {
-		trackingKeys[k] = true
+func (c *Controller) shouldWatch(r APIResource) bool {
+	if _, excluded := excludedGroups[r.groupVersion.Group]; excluded {
+		return false
 	}
-
-	// Get all the api resources in the cluster
-	apiResources, err := c.kubernetesClient.Discovery().ServerPreferredResources()
-	if err != nil {
-		// ignore the error caused by a stale API service
-		if !strings.Contains(err.Error(), util.UnableToRetrieveCompleteAPIListError) {
-			return nil, nil, err
-		}
+	if !util.IsAPIGroupAllowed(r.groupVersion.Group, c.allowedGroupsSet) {
+		return false
 	}
-
-	// Loop through the api resources and create informers and listers for each of them
-	for _, list := range apiResources {
-		gv, err := schema.ParseGroupVersion(list.GroupVersion)
-		if err != nil {
-			c.logger.Error(err, "Failed to parse a GroupVersion", "groupVersion", list.GroupVersion)
-			continue
-		}
-		if _, excluded := excludedGroups[gv.Group]; excluded {
-			continue
-		}
-		for _, resource := range list.APIResources {
-			if _, excluded := excludedResourceNames[resource.Name]; excluded {
-				continue
-			}
-			if !util.IsAPIGroupAllowed(gv.Group, c.allowedGroupsSet) {
-				continue
-			}
-			informable := verbsSupportInformers(resource.Verbs)
-			if informable {
-				key := util.KeyForGroupVersionKind(gv.Group, gv.Version, resource.Kind)
-				if _, ok := c.informers[key]; !ok {
-					toStart = append(toStart, APIResource{
-						groupVersion: gv,
-						resource:     resource,
-					})
-				}
-				// remove the key from tracking keys, what is left in the map at the end are
-				// keys to the informers that need to be stopped.
-				delete(trackingKeys, key)
-			}
-		}
+	if _, excluded := excludedResourceNames[r.resource.Name]; excluded {
+		return false
 	}
-
-	for k := range trackingKeys {
-		toStop = append(toStop, k)
-	}
-	return toStart, toStop, nil
+	return true
 }
 
 func (c *Controller) startInformersForNewAPIResources(toStartList []APIResource) {
 	for _, toStart := range toStartList {
-		c.logger.Info("New API added. Starting informer for:", "group", toStart.groupVersion.Group,
+		c.logger.Info("Ensuring informer for:", "group", toStart.groupVersion.Group,
 			"version", toStart.groupVersion, "kind", toStart.resource.Kind)
 
 		gvr := schema.GroupVersionResource{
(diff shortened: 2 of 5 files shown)


MikeSpreitzer:
With this change, does the controller-manager succeed without `--api-groups` when an OpenShift cluster is playing the WDS role?

pkg/binding/controller.go
```diff
@@ -99,17 +99,27 @@ type Controller struct {
...
 	}
 
-	kubernetesClient, err := kubernetes.NewForConfig(wdsRestConfig)
+	extClient, err := apiextensionsclientset.NewForConfig(wdsRestConfig)
 	if err != nil {
 		return nil, err
 	}
 
-	extClient, err := apiextensionsclientset.NewForConfig(wdsRestConfig)
+	burst, err := countGVRsViaAggregatedDiscovery(rest.CopyConfig(wdsRestConfig)) // use # of GVRs as burst is tested to be working well
```
MikeSpreitzer, line 112: We could pass `kubernetesClient` here instead of `wdsRestConfig`, right? Passing the client would save a bit of source code and runtime cost; most importantly, it would make the counting func more easily testable.

pkg/binding/crd.go
```diff
@@ -35,19 +37,43 @@ type APIResource struct {
...
+	uObj := obj.(*unstructured.Unstructured)
+	var crdObj *apiextensionsv1.CustomResourceDefinition
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(uObj.UnstructuredContent(), &crdObj); err != nil {
+		return fmt.Errorf("failed to convert Unstructured to CRD: %w", err)
+	}
+
+	toStart := APIResource{
+		groupVersion: schema.GroupVersion{
+			Group:   crdObj.Spec.Group,
+			Version: crdObj.Spec.Versions[0].Name,
```
MikeSpreitzer, line 50: I wonder if it is safe to look at `[0]` and no other. I thought that `v1` and `v2` have distinct populations of objects. Also, I looked at https://github.com/kubernetes/apiextensions-apiserver/blob/v0.28.2/pkg/apis/apiextensions/v1/types.go ; could `[0]` have `Served==false`?

MikeSpreitzer:
Did you consider using https://github.com/kubernetes/client-go/blob/v0.28.2/discovery/cached/memory/memcache.go#L326 with invalidation on create/update/delete of a CRD?

ezrasilvera:
@waltforme Thanks for the PR.  I'm curious do you know which change made the main impact ? E.g., If we would have just changed the client-side rate limiter would that solve the issue without the other changes?

ezrasilvera:
@waltforme Does this suppose to fix #1705 ?  If so can you please add `Fixes`  statement so they are linked .

waltforme:
Force-push to https://github.com/kubestellar/kubestellar/pull/1828/commits/c1aa6c0bc2656c2209108008cecb00243ed7dd73 includes some changes in response to comments, squashing the changes into appropriate commits, and rebasing on main.

waltforme:
> With this change, does the controller-manager succeed without `--api-groups` when an OpenShift cluster is playing the WDS role?

I wish I could.

At the beginning of this work, I tried to run the controller against an OpenShift cluster. But I got 
```
Get "https://imbs1-imbs1-system.edgeplatform1-9ca4d14d48413d18ce61b80811ba4308-0000.us-south.containers.appdomain.cloud:443/api": dial tcp 169.61.129.125:443: i/o timeout
```
I guess I was missing some ingress/route, so I tried [Create an nginx Ingress controller with SSL passthrough](https://raw.githubusercontent.com/kubestellar/kubestellar/main/example/kind-nginx-ingress-with-SSL-passthrough.yaml), which doesn't seem to work out-of-box.

I saw the log in #1705 is from some OpenShift cluster. So @pdettori Would you share how did you run the controller against an OpenShift cluster? Ideally not as a pod, but as a bare process so that weI can debug and make changes efficiently.

pkg/binding/controller.go
```diff
@@ -99,17 +99,27 @@ type Controller struct {
...
+	if err != nil {
+		parentLogger.Error(err, "Not able to count GVRs via aggregated discovery") // but we can still continue
+	}
+	wdsRestConfig.Burst = adjustBurstIfNecessary(burst)
+	wdsRestConfig.QPS = adjustQPSIfNecessary(float32(burst) / 4)
```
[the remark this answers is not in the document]
waltforme: I rewrote this section of code, and renamed the functions. Does it look better now? BTW what is the exact 'side-effect' you mentioned above?

pkg/binding/controller.go
waltforme: Good point. Changed as suggested.

pkg/binding/crd.go
waltforme: Added in https://github.com/kubestellar/kubestellar/pull/1828/commits/c1aa6c0bc2656c2209108008cecb00243ed7dd73 the functionality to manage the watches of multiple versions of a CRD.

waltforme:
> @waltforme Thanks for the PR. I'm curious do you know which change made the main impact ? E.g., If we would have just changed the client-side rate limiter would that solve the issue without the other changes?

Yes, I would say primarily adapting the relate limiter solved the problem.

Before the current changes in this PR, I can observe two bursts of requests, both at the start of the controller, and both may trigger the client-side throttling.
Burst 1: Requests from the reflectors' ListAndWatch, with longer 'Waited for' time.
Burst 2. Requests from `handleCRD` during the initial intensive reconciliation cycles of CRD objects, with shorter 'Waited for' time than Burst 1. These requests trigger client-side throttling *only* if I intentionally set the limit of the rate to some low values in my experiments.

So primarily, adapting the relate limiter solved the problem. But another change (change the handling to an incremental manner), actually eliminates *any* requests from `handleCRD` during the reconciliation cycles, thus eradicates the possibility of throttling due to Burst 2.

P.S. Yes, it will fix #1705 so I linked to it.

waltforme:
> Did you consider using https://github.com/kubernetes/client-go/blob/v0.28.2/discovery/cached/memory/memcache.go#L326 with invalidation on create/update/delete of a CRD?

I did. Because I saw the [discussions](https://github.com/kubestellar/kubestellar/issues/1705#issuecomment-1927490445) in #1705. But I realized that using a cached discovery client may not help us.

With the cached discovery client, we still need to make requests to the apiserver from time to time during the reconciliation cycles --- as you mentioned, on create/update/delete of a CRD. That will be not much different from the behavior before the changes in this PR.

OTOH, the changes in this PR eliminates any requests from handleCRD during the reconciliation cycles. (I put more details regarding 'Burst 2' in [the comment above](https://github.com/kubestellar/kubestellar/pull/1828#issuecomment-1971722249).)

ezrasilvera:
@waltforme Thanks.  From my side it's  LGTM .  I'll let @MikeSpreitzer put the official lgtm label.

pdettori:
> Would you share how did you run the controller against an OpenShift cluster? 

@waltforme - I was able to run on an existing OCP cluster from my local desktop by scaling down the KS controller to 0 replicas and then using `make run` with the kubeconfig context set to the kubeconfig of the hosting cluster. The controller connects to the hosting cluster and uses the CPs to get the secrets with the kubeconfigs (it should use the external kubeconfig and then connects). In your case the issue seems that your controller does not communicate with the imbs API server. Did you have your VPN to IBM Cloud connected during the test ?

pdettori (commented):
/lgtm 
Thanks @waltforme ! will wait for @MikeSpreitzer to see if has more comments on his side before doing an approve

waltforme:
@pdettori Thanks! This is really helpful.

> @waltforme - I was able to run on an existing OCP cluster from my local desktop by scaling down the KS controller to 0 replicas and then using `make run` with the kubeconfig context set to the kubeconfig of the hosting cluster. The controller connects to the hosting cluster and uses the CPs to get the secrets with the kubeconfigs (it should use the external kubeconfig and then connects). In your case the issue seems that your controller does not communicate with the imbs API server. Did you have your VPN to IBM Cloud connected during the test ?

waltforme:
I ran the controller against an OpenShift cluster (i.e. let the OpenShift cluster serve as a KubeFlex ControlPlane of type `host`, and apply `kflex.kubestellar.io/cptype=wds` label to the ControlPlane). Then I collected logs of the binding controller, both before the changes in this PR and after the changes in this PR, as attached.

Before the changes, we can see a lot of throttling messages, and the controller starts very slowly, I quit in the middle so the log is incomplete.

After the changes, the controller starts in seconds.
[before_1828_too_long_incomplete.txt](https://github.com/kubestellar/kubestellar/files/14455861/before_1828_too_long_incomplete.txt)
[after_1828_complete.txt](https://github.com/kubestellar/kubestellar/files/14455862/after_1828_complete.txt)

pdettori:
this looks great @waltforme - did you run with or without the --api-groups flag to select a subset of resources ?

pkg/util/resources.go
```diff
@@ -39,8 +39,8 @@ func ParseAPIGroupsString(apiGroups string) sets.Set[string] {
 	return groupsSet
 }
 
-// IsResourceGroupAllowed checks if a API group is allowed
-// an empty or nil allowedResources slice is equivalent to allow all,
+// IsResourceGroupAllowed checks if a API group is explicitly allowed by user,
+// an empty or nil allowedResources slice is equivalent to allow all.
```
pdettori, line 43: can you please re-enable  `allowedResourceGroups.Insert(apiextensions.GroupName)` as part of this PR? it was disabled exactly because of the throttling issue.

waltforme: Done in https://github.com/kubestellar/kubestellar/pull/1828/commits/215a066fa16f731a4cfc6595239a5834cba6b0f3.

waltforme: Reverted because re-enabling calls for some unit test cases to be updated. Will look at the test cases tmr.

pkg/binding/crd.go
```diff
@@ -35,19 +37,50 @@ type APIResource struct {
...
+				Kind: crdObj.Spec.Names.Kind,
+			},
+		}
+		key := util.KeyForGroupVersionKind(gvr.groupVersion.Group, gvr.groupVersion.Version, gvr.resource.Kind)
+		if isBeingDeleted(obj) {
+			toStopList = append(toStopList, key)
+			continue
+		}
+		if !c.includedToWatch(gvr) {
+			toStopList = append(toStopList, key)
```
MikeSpreitzer, line 66: If `!c.includedToWatch(gvr)`, aren't we guaranteed that it has not been started? `c.includedToWatch` does not vary its behavior over time.

pkg/binding/crd.go
```diff
@@ -35,19 +37,50 @@ type APIResource struct {
...
 
 	go c.startInformersForNewAPIResources(toStartList)
 
 	for _, key := range toStopList {
-		c.logger.Info("API removed, stopping informer.", "key", key)
-		stopper := c.stoppers[key]
-		// close channel
-		close(stopper)
+		c.logger.Info("API should be removed, stopping informer.", "key", key)
+		if stopper, ok := c.stoppers[key]; ok {
```
MikeSpreitzer, line 80: Why might `ok` be `false`? Is it only because of including things in `toStopList` that never pass `c.includedToWatch`?

pkg/binding/crd.go
```diff
@@ -35,19 +37,50 @@ type APIResource struct {
...
 	}
 
 	go c.startInformersForNewAPIResources(toStartList)
 
 	for _, key := range toStopList {
-		c.logger.Info("API removed, stopping informer.", "key", key)
-		stopper := c.stoppers[key]
-		// close channel
-		close(stopper)
+		c.logger.Info("API should be removed, stopping informer.", "key", key)
```
MikeSpreitzer, line 79: FYI, I would let the log message carry the information from `ok` below.

pkg/binding/crd.go
```diff
@@ -35,19 +37,50 @@ type APIResource struct {
...
+			continue
+		}
+		if !ver.Served {
+			toStopList = append(toStopList, key)
+			continue
+		}
+		toStartList = append(toStartList, gvr)
 	}
 
 	go c.startInformersForNewAPIResources(toStartList)
```
MikeSpreitzer, line 76: What if some members of `toStartList` are already being informed on?

pkg/util/resources.go
MikeSpreitzer: `controller_test.go` is junk, for the reason not clearly stated in https://github.com/kubestellar/kubestellar/pull/1775#issuecomment-1951422945 . The fakes do not support the sort of discovery that this controller uses, so the test can not exercise much. We should either delete this junk test or change it to use better fakes or change the controller to use the other kind of discovery.

MikeSpreitzer: @pdettori : can you please elaborate on what you are asking to be re-enabled?

MikeSpreitzer:
In the opening comment, in item 2, I would clarify that the rate limiter --- which is constructed just once at controller startup --- is sensitive to the number of GVRs reported by discovery _at the time of controller startup_.

test/integration/controller-manager/crd-handling_test.go
```diff
@@ -0,0 +1,294 @@
...
+	k8sjson "k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2/ktesting"
+	kastesting "k8s.io/kubernetes/cmd/kube-apiserver/app/testing"
+	"k8s.io/kubernetes/test/integration/framework"
+
+	"github.com/kubestellar/kubestellar/pkg/binding"
+)
+
+// An integration test for the binding controller.
```
MikeSpreitzer, line 37: Integration tests should go in the `test/integration` directory.

pkg/binding/controller.go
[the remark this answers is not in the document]
MikeSpreitzer: The side-effects are still there. I am talking about modifying the struct that the pointer parameter `wdsRestConfig` points to.

pkg/binding/crd.go
```diff
@@ -35,19 +37,50 @@ type APIResource struct {
...
+		}
+		if !c.includedToWatch(gvr) {
+			toStopList = append(toStopList, key)
+			continue
+		}
+		if !ver.Served {
+			toStopList = append(toStopList, key)
+			continue
+		}
+		toStartList = append(toStartList, gvr)
```
MikeSpreitzer, line 73: Welcome to the world of CRD dynamics. Please review https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/types.go#L308 before proceeding further.

pkg/binding/controller.go
```diff
@@ -99,17 +100,28 @@ type Controller struct {
...
 
-	extClient, err := apiextensionsclientset.NewForConfig(wdsRestConfig)
+	nGVRs, err := countGVRsViaAggregatedDiscovery(kubernetesClient)
+	if err != nil {
+		parentLogger.Error(err, "Not able to count GVRs via aggregated discovery") // but we can still continue
+	}
+	// tuning the rate limiter based on the number of GVRs is tested to be working well
+	wdsRestConfig.Burst = computeBurstFromNumGVRs(nGVRs)
+	wdsRestConfig.QPS = computeQPSFromNumGVRs(nGVRs)
+	parentLogger.V(1).Info("Parameters of the client's token bucket rate limiter", "burst", wdsRestConfig.Burst, "qps", wdsRestConfig.QPS)
```
MikeSpreitzer, line 120: This logging should be done via the controller's logger rather than the parent logger.

pkg/binding/controller.go
```diff
@@ -124,6 +136,50 @@ func NewController(parentLogger logr.Logger, wdsRestConfig *rest.Config, imbsRes
 	return makeController(parentLogger, dynamicClient, kubernetesClient, extClient, ocmClientset, ocmClient, wdsName, allowedGroupsSet)
 }
 
+func countGVRsViaAggregatedDiscovery(countClient kubernetes.Interface) (int, error) {
+	dc := countClient.Discovery().(*discovery.DiscoveryClient)
+	if dc.UseLegacyDiscovery { // by default it should be false already, just double check
+		dc.UseLegacyDiscovery = false
+	}
+	apiResourceLists, err := dc.ServerPreferredResources()
```
MikeSpreitzer, line 144: This can be expensive! Wouldn't it be better to find a way to do it only once?

pkg/binding/controller.go
```diff
@@ -124,6 +136,50 @@ func NewController(parentLogger logr.Logger, wdsRestConfig *rest.Config, imbsRes
...
+	if err != nil {
+		return 0, fmt.Errorf("error listing server perferred resources: %w", err)
+	}
+	n := 0
+	for _, list := range apiResourceLists {
+		n += len(list.APIResources)
```
MikeSpreitzer, line 150: Why not apply `util.IsAPIGroupAllowed` here?

pkg/binding/crd.go
```diff
@@ -35,19 +38,55 @@ type APIResource struct {
...
+	if len(toStartList) > 0 {
+		go c.startInformersForNewAPIResources(ctx, toStartList)
+	}
 
 	for _, key := range toStopList {
-		c.logger.Info("API removed, stopping informer.", "key", key)
-		stopper := c.stoppers[key]
-		// close channel
-		close(stopper)
+		logger.Info("API should not be watched, stopping informer.", "key", key)
```
MikeSpreitzer, line 85: This log message is misleading when `ok` in the next statement is `false`.

pkg/binding/controller.go
MikeSpreitzer: Or `includedToWatch`

pkg/binding/crd.go
```diff
@@ -58,69 +97,32 @@ func (c *Controller) handleCRD(_ runtime.Object) error {
...
+	return false
 }
 
-func (c *Controller) startInformersForNewAPIResources(toStartList []APIResource) {
+func (c *Controller) startInformersForNewAPIResources(ctx context.Context, toStartList []APIResource) {
+	logger := klog.FromContext(ctx)
 	for _, toStart := range toStartList {
-		c.logger.Info("New API added. Starting informer for:", "group", toStart.groupVersion.Group,
+		logger.Info("Ensuring informer for:", "group", toStart.groupVersion.Group,
 			"version", toStart.groupVersion, "kind", toStart.resource.Kind)
```
MikeSpreitzer, line 127: I see a lot of work below that could be conditional on whether `key` is already in `c.informers`.

pkg/util/resources.go
MikeSpreitzer: Now I see what @pdettori is talking about.

waltforme: Redid in https://github.com/kubestellar/kubestellar/pull/1828/commits/185029311709dc9ececfac041a272c8009dd5c2d together updated unit tests.

pkg/binding/crd.go
waltforme: Changed in latest push.

pkg/binding/crd.go
waltforme: Added in latest push.

pkg/binding/controller.go
waltforme: Controller, as the golang struct, doesn't have access to the values, right?

pkg/binding/controller.go
waltforme: Not sure I understand the question. This PR is doing exactly reducing the number of calls to `ServerPreferredResources()` (during CRD handling) to only one. And the only one is right here.

pkg/binding/controller.go
waltforme: I think the estimation doesn't have to be accurate. Actually we want to tune the limiter in favor of the performance. Also, the parameters of the limiter is bounded by `computeBurstFromNumGVRs` and `computeQPSFromNumGVRs` anyway.

pkg/binding/crd.go
waltforme: Revised.

pkg/binding/crd.go
waltforme: Done in latest push.

pkg/binding/controller.go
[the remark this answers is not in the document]
waltforme: Still not following. Which line modifies the struct `wdsRestConfig`, as a pointer parameter (or in other words as 'an input data structure')? I thought the modifications are explicit and not modifying as any input data structure in current lines 118 and 119.

pkg/binding/crd.go
MikeSpreitzer: I do not see that change in https://github.com/kubestellar/kubestellar/blob/48f75f51b37bd325ad4c075aa15ad0c6b2158d6b/pkg/binding/crd.go#L84-L85

MikeSpreitzer: FYI: I was trying to suggest that there would be just one log message in both cases (`ok` and `!ok`), with the log messages somehow letting the reader know whether `ok` was true or not. For example, have two different log statements (in the two arms of the `if`), which would not need an additional log statement up front.

pkg/binding/controller.go
MikeSpreitzer: I am not sure I understand. In the current version of this PR (https://github.com/kubestellar/kubestellar/blob/48f75f51b37bd325ad4c075aa15ad0c6b2158d6b/pkg/binding/controller.go), the controller does not even exist at the point in code that we are discussing. To use the controller's logger here would require moving the construction of that logger from inside `makeController` to somewhere earlier than the line we are discussing.

pkg/binding/controller.go
MikeSpreitzer: I see that func also being called at https://github.com/kubestellar/kubestellar/blob/48f75f51b37bd325ad4c075aa15ad0c6b2158d6b/pkg/binding/controller.go#L249 and https://github.com/kubestellar/kubestellar/blob/48f75f51b37bd325ad4c075aa15ad0c6b2158d6b/pkg/crd/crds.go#L94

pkg/binding/crd.go
MikeSpreitzer: I see the early-out now. I am surprised by the retention of the complex code later for updating `c.listers`, `c.informers`, and `c.stoppers`. Aren't they coordinated? For a given key, either all three have an entry or all three do not have an entry, right? BTW, this would be more clearly handled by one map from key to a pointer to a struct holding the informer, lister, and stopper.

pkg/binding/controller.go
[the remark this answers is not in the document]
waltforme: I finally see the point (it would have been helpful in the first comment to say the input data structure **of `NewController`**). But I do have a different view on this issue and would like to discuss as follows.

My view is that if a caller passes a pointer to a callee, the caller should be 'prepared' that the pointed content may be modified. If the caller doesn't want the possible modification, than pass a copy or value instead. In other words, if a pointer is passed, than permission of modification is implicit.

Is there some established convention (pls do let me know if there is)? Or this view can vary from person to person?

pkg/binding/controller.go
waltforme: > the controller does not even exist at the point in code that we are discussing

This is exactly what made me confused. So what you suggested is to construct the controller's log earlier and use it. I get the point. But what is the issue of using the `parentLogger` instead? BTW why is it called `parentLogger`, i.e. who is the child?

MikeSpreitzer: The binding controller is just one of the controllers in the controller-manager process.

pkg/binding/controller.go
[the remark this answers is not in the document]
MikeSpreitzer: I am not aware of a universal convention about this for programming in Go. However, we are talking here about a more specific subject matter: setting up controllers in Kubernetes code. The convention in Kubernetes code is that the consumer of a REST config does not side-effect the copy given unless there is something explicit in the func's name or comments about that. The consumer can make a copy for its own purpose, if necessary.

MikeSpreitzer (approved):
This is good enough to merge. The remaining issues can be addressed in follow-on.
/lgtm
/approve
Thanks!

PR #1829, opened 2024-02-28
🐛 Add cleaning for Job objects

### LINKED ISSUE #1766: feature:  add denature for Jobs in the WDS

MikeSpreitzer:
## Summary
This PR updates the generic transport controller code to remove the apiserver-maintained fields of a JobSpec.

The corresponding test enhancement will be done in #1824 or follow-on.

## Related issue(s)

This is an alternative to #1803, which we decided not to pursue (in favor of #1812).

Fixes #1766

### DIFF
--- a/pkg/transport/filtering/job.go
+++ b/pkg/transport/filtering/job.go
@@ -1,0 +1,41 @@
+/*
+Copyright 2024 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package filtering
+
+import (
+	"github.com/go-logr/logr"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func cleanJob(logger logr.Logger, object *unstructured.Unstructured) {
+	objectU := object.UnstructuredContent()
+	podLabels, found, err := unstructured.NestedMap(objectU, "spec", "template", "metadata", "labels")
+	if err != nil || !found {
+		logger.V(3).Error(nil, "Job lacks expected template labels", "namespace", object.GetNamespace(), "name", object.GetName(), "found", found, "err", err)
+		return
+	}
+	delete(podLabels, "batch.kubernetes.io/controller-uid")
+	delete(podLabels, "batch.kubernetes.io/job-name")
+	delete(podLabels, "controller-uid")
+	delete(podLabels, "job-name")
+	err = unstructured.SetNestedMap(objectU, podLabels, "spec", "template", "metadata", "labels")
+	if err != nil {
+		logger.Error(err, "Inconceivable! Failed to update existing pod labels in Job", "namespace", object.GetNamespace(), "name", object.GetName())
+	}
+	object.SetUnstructuredContent(objectU)
+}
--- a/pkg/transport/filtering/job.go
+++ b/pkg/transport/filtering/job.go
@@ -17,28 +17,19 @@ limitations under the License.
 package filtering
 
 import (
-	"github.com/go-logr/logr"
-
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-func cleanJob(logger logr.Logger, object *unstructured.Unstructured) {
+func cleanJob(object *unstructured.Unstructured) {
 	objectU := object.UnstructuredContent()
-	podLabels, found, err := unstructured.NestedMap(objectU, "spec", "template", "metadata", "labels")
+	podLabels, found, _ := unstructured.NestedMap(objectU, "spec", "template", "metadata", "labels")
 	if !found {
-		return
-	}
-	if err != nil {
-		logger.V(3).Error(nil, "Job object does not have expected struture, no spec.template.metadata.labels", "namespace", object.GetNamespace(), "name", object.GetName(), "err", err)
 		return
 	}
 	delete(podLabels, "batch.kubernetes.io/controller-uid")
 	delete(podLabels, "batch.kubernetes.io/job-name")
 	delete(podLabels, "controller-uid")
 	delete(podLabels, "job-name")
-	err = unstructured.SetNestedMap(objectU, podLabels, "spec", "template", "metadata", "labels")
-	if err != nil { // that condition can never be true
-		panic(err)
-	}
+	_ = unstructured.SetNestedMap(objectU, podLabels, "spec", "template", "metadata", "labels")
 	object.SetUnstructuredContent(objectU)
 }
--- a/pkg/transport/filtering/job.go
+++ b/pkg/transport/filtering/job.go
@@ -20,16 +20,55 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
+func badJobKey(key string) bool {
+	return key == "controller-uid" || key == "batch.kubernetes.io/controller-uid"
+}
+
 func cleanJob(object *unstructured.Unstructured) {
 	objectU := object.UnstructuredContent()
-	podLabels, found, _ := unstructured.NestedMap(objectU, "spec", "template", "metadata", "labels")
-	if !found {
-		return
+	changed := false
+	selector, foundSelector, _ := unstructured.NestedMap(objectU, "spec", "selector")
+	if foundSelector {
+		if matchLabels, found, _ := unstructured.NestedMap(selector, "matchLabels"); found {
+			if cleanLabels(matchLabels) {
+				_ = unstructured.SetNestedMap(objectU, matchLabels, "spec", "selector", "matchLabels")
+				changed = true
+			}
+		}
+		if matchExpressions, found, _ := unstructured.NestedSlice(selector, "matchExpressions"); found {
+			cleanedExpressions := make([]any, 0, len(matchExpressions))
+			for _, expr := range matchExpressions {
+				exprM := expr.(map[string]any)
+				key, _, _ := unstructured.NestedString(exprM, "key")
+				if !badJobKey(key) {
+					cleanedExpressions = append(cleanedExpressions, expr)
+				}
+			}
+			if len(cleanedExpressions) != len(matchExpressions) {
+				unstructured.SetNestedSlice(objectU, cleanedExpressions, "spec", "selector", "matchExpressions")
+				changed = true
+			}
+		}
 	}
-	delete(podLabels, "batch.kubernetes.io/controller-uid")
-	delete(podLabels, "batch.kubernetes.io/job-name")
-	delete(podLabels, "controller-uid")
-	delete(podLabels, "job-name")
-	_ = unstructured.SetNestedMap(objectU, podLabels, "spec", "template", "metadata", "labels")
-	object.SetUnstructuredContent(objectU)
+	podLabels, foundlabels, _ := unstructured.NestedMap(objectU, "spec", "template", "metadata", "labels")
+	if foundlabels {
+		if cleanLabels(podLabels) {
+			_ = unstructured.SetNestedMap(objectU, podLabels, "spec", "template", "metadata", "labels")
+			changed = true
+		}
+	}
+	if changed {
+		object.SetUnstructuredContent(objectU)
+	}
+}
+
+func cleanLabels(labels map[string]any) bool {
+	changed := false
+	for key := range labels {
+		if badJobKey(key) {
+			delete(labels, key)
+			changed = true
+		}
+	}
+	return changed
 }
--- a/pkg/transport/filtering/job.go
+++ b/pkg/transport/filtering/job.go
@@ -25,8 +25,11 @@ import (
 func cleanJob(logger logr.Logger, object *unstructured.Unstructured) {
 	objectU := object.UnstructuredContent()
 	podLabels, found, err := unstructured.NestedMap(objectU, "spec", "template", "metadata", "labels")
-	if err != nil || !found {
-		logger.V(3).Error(nil, "Job lacks expected template labels", "namespace", object.GetNamespace(), "name", object.GetName(), "found", found, "err", err)
+	if !found {
+		return
+	}
+	if err != nil {
+		logger.V(3).Error(nil, "Job object does not have expected struture, no spec.template.metadata.labels", "namespace", object.GetNamespace(), "name", object.GetName(), "err", err)
 		return
 	}
 	delete(podLabels, "batch.kubernetes.io/controller-uid")
@@ -34,8 +37,8 @@ func cleanJob(logger logr.Logger, object *unstructured.Unstructured) {
 	delete(podLabels, "controller-uid")
 	delete(podLabels, "job-name")
 	err = unstructured.SetNestedMap(objectU, podLabels, "spec", "template", "metadata", "labels")
-	if err != nil {
-		logger.Error(err, "Inconceivable! Failed to update existing pod labels in Job", "namespace", object.GetNamespace(), "name", object.GetName())
+	if err != nil { // that condition can never be true
+		panic(err)
 	}
 	object.SetUnstructuredContent(objectU)
 }
--- a/pkg/transport/filtering/object_filtering_map.go
+++ b/pkg/transport/filtering/object_filtering_map.go
@@ -17,8 +17,6 @@ limitations under the License.
 package filtering
 
 import (
-	"github.com/go-logr/logr"
-
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -28,7 +26,7 @@ import (
 // The function cleans the specific fields in place (object is modified).
 // If the object was retrieved using a lister, it's the caller responsibility
 // to do a DeepCopy before calling this function.
-type cleanObjectSpecificsFunction func(logger logr.Logger, object *unstructured.Unstructured)
+type cleanObjectSpecificsFunction func(object *unstructured.Unstructured)
 
 func NewObjectFilteringMap() *ObjectFilteringMap {
 	filteringMap := map[schema.GroupVersionKind]cleanObjectSpecificsFunction{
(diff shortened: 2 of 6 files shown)


pkg/transport/filtering/job.go
```diff
@@ -0,0 +1,41 @@
...
+		return
+	}
+	delete(podLabels, "batch.kubernetes.io/controller-uid")
+	delete(podLabels, "batch.kubernetes.io/job-name")
+	delete(podLabels, "controller-uid")
+	delete(podLabels, "job-name")
+	err = unstructured.SetNestedMap(objectU, podLabels, "spec", "template", "metadata", "labels")
+	if err != nil {
+		logger.Error(err, "Inconceivable! Failed to update existing pod labels in Job", "namespace", object.GetNamespace(), "name", object.GetName())
+	}
```
nirrozenbaum, line 39: I don't think the logger is needed here at all.
what is the purpose of this clean function?
to clean several labels from `spec.template.metadata.labels`.

if `spec.template.metadata.labels` is not found on the object, then just need to return, there's nothing to clean.

this log is adding noise without any need for it.

nirrozenbaum: in line 37 this doesn't make sense even more. you log an error but then continue the handling.
if anyway you continue processing, then this log is not needed.

pkg/transport/filtering/job.go
```diff
@@ -0,0 +1,41 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func cleanJob(logger logr.Logger, object *unstructured.Unstructured) {
+	objectU := object.UnstructuredContent()
+	podLabels, found, err := unstructured.NestedMap(objectU, "spec", "template", "metadata", "labels")
+	if err != nil || !found {
+		logger.V(3).Error(nil, "Job lacks expected template labels", "namespace", object.GetNamespace(), "name", object.GetName(), "found", found, "err", err)
+		return
+	}
```
nirrozenbaum, line 31: I would write these lines :
```
podLabels, found, _ := unstructured.NestedMap(objectU, "spec", "template", "metadata", "labels")
	!found {
		return
	}
```

pkg/transport/filtering/job.go
```diff
@@ -0,0 +1,41 @@
...
+	delete(podLabels, "batch.kubernetes.io/controller-uid")
+	delete(podLabels, "batch.kubernetes.io/job-name")
+	delete(podLabels, "controller-uid")
+	delete(podLabels, "job-name")
+	err = unstructured.SetNestedMap(objectU, podLabels, "spec", "template", "metadata", "labels")
```
nirrozenbaum, line 36: this error check is redundat. from `SetNestedMap documentation:
```
SetNestedMap sets the map[string]interface{} value of a nested field. Returns an error if value cannot be set because one of the nesting levels is not a map[string]interface{}.
```
you just got the labels from the object and deleted from of them. of course it's a valid map.

pkg/transport/filtering/job.go
MikeSpreitzer: The logging is there because we think the apiserver-managed fields will certainly be present. If they are not, it is a surprise to us. The logging is a basic attempt to make this surprise visible to us.

pkg/transport/filtering/job.go
MikeSpreitzer: Yeah, that's why the log message is "Inconceivable!".

nirrozenbaum:
other than the logger which seems not needed and even confusing (logs error but continue processing) everything looks good to me.
if we can remove the logging I'm happy to lgtm the PR

pkg/transport/filtering/job.go
```diff
@@ -0,0 +1,41 @@
...
+package filtering
+
+import (
+	"github.com/go-logr/logr"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func cleanJob(logger logr.Logger, object *unstructured.Unstructured) {
+	objectU := object.UnstructuredContent()
```
nirrozenbaum, line 26: if you use in line 27 object.Object instead of objectU, then this line is redundant.

pkg/transport/filtering/job.go
nirrozenbaum: so do you agree this is a redundant error check?
if this is inconceivable, we don't need to check for error and add logs that this is inconceivable.
this is just adding noise with no reason

nirrozenbaum, line 36: you can use here object.Object as well

pkg/transport/filtering/job.go
MikeSpreitzer: On second thought, I agree that complaining when the pod spec has no labels is being unnecessarily prickly. I have removed that.

OTOH, if the structure navigation fails, that is a more serious surprise. I have left logging in that case.

MikeSpreitzer: I changed the "Inconceivable!" log message into a panic.

pkg/transport/filtering/job.go
MikeSpreitzer: Yes. That leaves `err != nil`, which is a more serious surprise.

pkg/transport/filtering/job.go
MikeSpreitzer: I changed this log statement into a panic.

pkg/transport/filtering/job.go
MikeSpreitzer: I prefer to use the procedural interface rather that directly accessing internal structure; the former is easier on evolution scenarios.

pkg/transport/filtering/job.go
nirrozenbaum: that change doesn't make it better.
we still have a check for a condition that can never happen.

there is no point in adding noise without any need.

In general I'm in favor of reliable code and adding as many checks as needed, when needed.
here the logging is just not needed.
you wrote it yourself  - `that condition can never be true`. then why are we checking it?

pkg/transport/filtering/job.go
```diff
@@ -0,0 +1,44 @@
...
+	podLabels, found, err := unstructured.NestedMap(objectU, "spec", "template", "metadata", "labels")
+	if !found {
+		return
+	}
+	if err != nil {
+		logger.V(3).Error(nil, "Job object does not have expected struture, no spec.template.metadata.labels", "namespace", object.GetNamespace(), "name", object.GetName(), "err", err)
+		return
```
nirrozenbaum, line 33: same here.. see documentation of unstructured:
```
NestedMap returns a deep copy of map[string]interface{} value of a nested field. Returns false if value is not found and an error if not a map[string]interface{}.
```
if you get false in `found` then it's not found. otherwise, only if the field is found but it's not a map - you'll get an error.
this is VERY redundant check that can never happen

ezrasilvera:
@MikeSpreitzer Any chance you can  also add a Job related test into the existing e2e test for service ?  We don't even need to clean between them just test deploying a Job just after the service .

pkg/transport/filtering/job.go
MikeSpreitzer: For safety as the code is edited in the future. If we simply remove all mention of the possibility of an error here then it is easier for a future change to introduce the possibility of an error here without noticing.

pkg/transport/filtering/job.go
MikeSpreitzer: The type mismatch can happen at any point along the path from the root of the object.

Having this check makes this func less reliant on other parts of our codebase and Kubernetes. This func takes an unstructured object, whose schema is not guaranteed a priori ; it is only when coupled with a larger analysis that shows this func being associated with v1 of Job that we start to get confidence. Even then, full confidence is relying on a lot more Kubernetes code.

nirrozenbaum: if your concern is type mismatch then you try to cast the object to job in the first line of the function. if cast fails it’s not a job and the function can return, cause this is cleaner function for jobs only.
again, for this no needed for any logger

pkg/transport/filtering/job.go
nirrozenbaum: not sure I understand your point here. 
this is v1. it must be backwards compatible. 
any future change in k8s job should keep the existing ux

nirrozenbaum: if in the future there will be a case we need to check, we can always add log in the future. 
I’m not in favor of adding now “for the future”

pkg/transport/filtering/job.go
MikeSpreitzer: Do you mean "cast" as in a [Go type assertion](https://go.dev/ref/spec#Type_assertions)? I would not expect that to succeed here because the `runtime.Object` is a `*unstructured.Unstructured` not a `*appsv1.Job`.

My point is about the complexity of understanding the code. With error checking in the func, it has a reasonable reaction of code outside the func (e.g., in how it gets registered and called) changes.

This debate is less important than getting something merged. I give up.

MikeSpreitzer: I removed the error checking.

nirrozenbaum:
/lgtm
Thanks! 🙏🏼

MikeSpreitzer:
Yes, @ezrasilvera , I am working on that.
/hold

nirrozenbaum:
for understanding what's the issue, one can see the dump of the e2e test.
more specifically, you can see the manifestswork part, which contains the particular error that causes this test not to pass.
```
message: 'Failed to apply manifest: Job.batch "pi" is invalid: [spec.selector:
            Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"batch.kubernetes.io/controller-uid":"a5eb20af-aaf8-4b9f-ada1-63bef08fadf7"},
            MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: `selector` not auto-generated,
            spec.template.metadata.labels: Invalid value: map[string]string{"batch.kubernetes.io/controller-uid":"122ee598-7824-4fd3-aca7-18c7b7d3cff0",
            "batch.kubernetes.io/job-name":"pi", "controller-uid":"122ee598-7824-4fd3-aca7-18c7b7d3cff0",
            "job-name":"pi"}: `selector` does not match template `labels`]'
```

you are missing the cleaning of the label  `batch.kubernetes.io/controller-uid` from `spec.selector.matchLabels`.

MikeSpreitzer:
@nirrozenbaum : yes, I saw that. The current code cleans both the selector and the pod template's labels.

nirrozenbaum:
@MikeSpreitzer maybe I’m missing something, where does the label batch.kubernetes.io/controller-uid gets cleaned from spec.selector.matchLabels?

MikeSpreitzer:
The force-pushes above abandon the sophisticated approach of trimming selector and pod label content not mentioned in `managedFields` because I was not seeing _any_ `managedFields` content in the objects. The current draft of this PR does something simpler and oblivious: it removes selector and pod label content for "controller-uid".

pkg/transport/filtering/object_filtering_map.go
```diff
@@ -43,9 +46,13 @@ type ObjectFilteringMap struct {
...
 func (filteringMap *ObjectFilteringMap) CleanObjectSpecifics(object *unstructured.Unstructured) {
-	filteringFunction, found := filteringMap.gvkToFilteringFunc[object.GetObjectKind().GroupVersionKind()]
+	gvk := object.GetObjectKind().GroupVersionKind()
+	filteringFunction, found := filteringMap.gvkToFilteringFunc[gvk]
 	if !found {
+		klog.V(4).InfoS("No specifics", "gvk", gvk)
 		return // if no filtering function was defined for this gvk, do not clean any field
+	} else {
+		klog.V(4).InfoS("Found specifics", "gvk", gvk)
 	}
```
nirrozenbaum, line 56: can we remove the log from this function?
this will thrash the controller log on every object that is being wrapped, make it harder to follow the log.

nirrozenbaum:
@MikeSpreitzer left a single minor comment about the log in `CleanObjectSpecifics` func. 
other than that looks good.

test/e2e/object-fields-filtering/use-kubestellar.sh
```diff
@@ -18,7 +18,7 @@ set -e # exit on error
 
 :
 : -------------------------------------------------------------------------
-: "Create a bindingpolicy in wds1 to deliver a service to one WEC."
+: "Create a bindingpolicy in wds1 to deliver difficult API objects to one WEC."
```
ezrasilvera, line 21: Why can't we just say to deliver Job and Service API objects.  Users may not know what difficult means here

pkg/transport/filtering/object_filtering_map.go
```diff
@@ -43,9 +46,13 @@ type ObjectFilteringMap struct {
 }
 
 func (filteringMap *ObjectFilteringMap) CleanObjectSpecifics(object *unstructured.Unstructured) {
-	filteringFunction, found := filteringMap.gvkToFilteringFunc[object.GetObjectKind().GroupVersionKind()]
+	gvk := object.GetObjectKind().GroupVersionKind()
+	filteringFunction, found := filteringMap.gvkToFilteringFunc[gvk]
 	if !found {
+		klog.V(4).InfoS("No specifics", "gvk", gvk)
 		return // if no filtering function was defined for this gvk, do not clean any field
+	} else {
```
ezrasilvera, line 54: why do we need the "else" here ?

pkg/transport/filtering/object_filtering_map.go
```diff
@@ -31,6 +33,7 @@ type cleanObjectSpecificsFunction func(object *unstructured.Unstructured)
 func NewObjectFilteringMap() *ObjectFilteringMap {
 	filteringMap := map[schema.GroupVersionKind]cleanObjectSpecificsFunction{
 		corev1.SchemeGroupVersion.WithKind("Service"): cleanServiceFields,
+		batchv1.SchemeGroupVersion.WithKind("Job"):    cleanJob,
```
ezrasilvera, line 36: nit:  cleanJobFields or change also to  cleanService  .  BTW, why not cleanJobObject and cleanServiceObject  ?

pkg/transport/filtering/object_filtering_map.go
MikeSpreitzer: That was an oversight. Fixed.

test/e2e/object-fields-filtering/use-kubestellar.sh
MikeSpreitzer: reworded.

pkg/transport/filtering/object_filtering_map.go
MikeSpreitzer: Now that the unintended logging is removed, there is no longer an "else".

pkg/transport/filtering/object_filtering_map.go
MikeSpreitzer: Made the Service one shorter too.

pkg/transport/filtering/object_filtering_map.go
ezrasilvera: My point is that  you don't need it if the if has a return  inside

ezrasilvera: but as it's gone in any case I'm closing. thanks

ezrasilvera:
Thanks @MikeSpreitzer  It LGTM  I'll let @nirrozenbaum complete his review

PR #1831, opened 2024-02-28
:sparkles: reworked util.Key and the flow of the Binding controller

### LINKED ISSUE #1703: feature: rework the Key struct in the placement code (or equivalent for PlacementResolver)
### Feature Description

The use of the Key struct in the PlacementResolver to identify objects is adding a lot of GVK <-> GVR conversion overhead, which:
- Complicates code
- Creates instances of erroring where can be avoided
- Introduces validation issues when these mappings go out of sync

Initially, the use of Key was for being as close as possible to the existing code (under some efficiency constraints), and a rework was planned for after initial merge.
During the review of #1681, @MikeSpreitzer repeatedly raised the point that things can be better if the resolver worked with resource instead of kind.

### Proposed Solution

Rework placement's Key or introduce a new one for the new PlacementResolver.

### Want to contribute?

### Additional Context

_No response_

### LINKED ISSUE #1792: binding controller informers setup should check cache.DeletedFinalStateUnknown
What happens if obj is a `cache.DeletedFinalStateUnknown` ?

_Originally posted by @MikeSpreitzer in https://github.com/kubestellar/kubestellar/pull/1765#discussion_r1491685185_

vMaroon:
## Summary
The goal of the previous pluggable-transport work was to switch to pluggable-transport given the existing code-base as grounding, with minimal acceptable changes. During this work, some of the binding workflow and object referencing (util.Key) became suboptimal and redundant for the new flow.

This PR:

- Changes the structure (and name) of util.Key, keeping only necessary information to reference an object - that is without a copy of the deleted object when previously needed
- Changes the flow of the binding controller: the reconciliation function does not try to retrieve the object being reconciled, but instead passes the new Key (ObjectIdentifier) to syncer functions, which retrieve the object and handle the case of the object being deleted as they should under the new logic
  - If a bindingpolicy is deleted, we update internal information
  - If a bindingpolicy is being deleted, we delete finalizer and update internal information
  - If a workload object is deleted, it is removed from all resolutions that contain it through a new function
  - If a workload object is being deleted, the previous (pluggable transport) flow is kept
- Changes the resolver so that it maintains sets of ObjectIdentifiers, and no longer needs to use the GvkGvrMapper

This PR also:
- Refactors informers/stoppers maps in binding and status controllers to use GVK struct as key
- Refactors loggers following @MikeSpreitzer recent updates (not to use Controller::logger)
- Removes unused code

Fixes #1703
Fixes #1792


vMaroon:
/hold

GvkGvrMapper's usage can be further refactored out of the code, which will mostly affect pkg/bindingpolicy.go and the interface/implementation of the mapper. The code can still be reviewed.

vMaroon:
/unhold

@MikeSpreitzer pls review

pkg/util/object-identifier.go
```diff
@@ -0,0 +1,99 @@
+/*
+Copyright 2023 The KubeStellar Authors.
```
MikeSpreitzer, line 2: FYI, it's a new year.

pkg/binding/controller.go
```diff
@@ -85,7 +84,7 @@ type Controller struct {
 
 	extClient             apiextensionsclientset.Interface // used for CRD
 	listers               map[string]cache.GenericLister
-	gvkGvrMapper          util.GvkGvrMapper
+	gvkToGvrMapper        util.GVKToGVRMapper
 	informers             map[string]cache.SharedIndexInformer
 	stoppers              map[string]chan struct{}
```
MikeSpreitzer, line 89: FYI: You can now replace these strings with GroupVersionKind or GroupVersionResource. This would make the code more reliably checkable at compile time.

pkg/util/object-identifier.go
```diff
@@ -0,0 +1,99 @@
...
+
+	return fmt.Sprintf("%s/%s/%s", identifier.GVK.Group, identifier.GVK.Version, identifier.GVK.Kind)
+}
+
+// RuntimeObjectToObjectIdentifier creates an ObjectIdentifier given an object
+// that implements runtime.Object. The gvkGvrMapper is used to retrieve the
+// resource name for the given runtime.Object.
+func RuntimeObjectToObjectIdentifier(rObj runtime.Object, gvkToGvrMapper GVKToGVRMapper) (ObjectIdentifier, error) {
+	ok := rObj.GetObjectKind()
+	gvk := ok.GroupVersionKind()
```
MikeSpreitzer, line 52: FYI:

```suggestion
	gvk := rObj.GetObjectKind().GroupVersionKind()
```

pkg/util/object-identifier.go
```diff
@@ -0,0 +1,99 @@
...
+
+	namespacedName, err := cache.ObjectToName(rObj)
+	if err != nil {
```
MikeSpreitzer, line 55: If the input parameter were a `mrObject` instead of a `runtime.Object` then we could call https://github.com/kubernetes/client-go/blob/v0.28.2/tools/cache/store.go#L130 and not have to cope with the possibility of an error.

test/integration/controller-manager/matching_test.go
```diff
@@ -171,16 +171,16 @@ func TestMatching(t *testing.T) {
 	}
 	expectation := map[gvrnn]any{}
 	for _, obj := range append(nsORs, objsCreated...) {
-		key, err := util.KeyForGroupVersionKindNamespaceName(obj.MRObject)
+		identifier, err := util.RuntimeObjectToObjectIdentifier(obj.MRObject, ctlr.GetGvkToGvrMapper())
 		if err != nil {
 			t.Fatalf("Failed to extract Key from %#v: %s", obj.MRObject, err)
 		}
-		k2 := k2r(key, obj.Resource)
+		id2 := id2r(identifier, obj.Resource)
```
MikeSpreitzer, line 178: FYI: `identifier` already carries the resource, no need for this func to take a second argument.

MikeSpreitzer:
I think the code would get even simpler if all the keying were based on GVR instead of GVK or GVK+R. Most of the Kubernetes libraries deal in GVR; some report both. While `kubectl` has the proble of user input that only has GVK, this controller does not. This controller is driven by discovery. The only time this controller has a GVK directly at hand without knowing the Resource is when processing an API object retrieved using the dynamic client. But that dynamic client has to be given the GVR before you get to any methods that deliver objects --- so the Resource is, in fact, already known.

pkg/binding/controller.go
```diff
@@ -477,67 +458,41 @@ func (c *Controller) processNextWorkItem(ctx context.Context) bool {
...
-		return nil
-	}
-
-	return c.updateResolutions(ctx, obj)
+	return c.updateResolutions(ctx, objIdentifier)
 }
 
-func (c *Controller) getObjectFromKey(key util.Key) (runtime.Object, error) {
-	lister := c.listers[key.GvkKey()]
+func (c *Controller) getObjectFromObjectIdentifier(objIdentifier util.ObjectIdentifier) (runtime.Object, error) {
```
MikeSpreitzer, line 488: FYI: I think the name of this func needs to say "object" twice; I would call it "getObjectFromIdentifier" or maybe even just "getObject".

pkg/util/object-identifier.go
```diff
@@ -0,0 +1,99 @@
...
+		return fmt.Sprintf("%s/%s", identifier.GVK.Version, identifier.GVK.Kind)
+	}
+
+	return fmt.Sprintf("%s/%s/%s", identifier.GVK.Group, identifier.GVK.Version, identifier.GVK.Kind)
+}
+
+// RuntimeObjectToObjectIdentifier creates an ObjectIdentifier given an object
+// that implements runtime.Object. The gvkGvrMapper is used to retrieve the
+// resource name for the given runtime.Object.
+func RuntimeObjectToObjectIdentifier(rObj runtime.Object, gvkToGvrMapper GVKToGVRMapper) (ObjectIdentifier, error) {
```
MikeSpreitzer, line 50: FYI, I do not think the name of this func needs to say "object" twice. I would name it something like "ObjectToIdentifier" or "IdentifierForObject".

pkg/binding/binding.go
```diff
@@ -32,24 +32,26 @@ import (
...
 
 		return nil
 	}
 
-	obj, err := c.getObjectFromKey(key)
+	obj, err := c.getObjectFromObjectIdentifier(objIdentifier)
 	if errors.IsNotFound(err) {
-		unstructuredObj = util.EmptyUnstructuredObjectFromKey(key)
+		// a resolution exists and the object is not found, therefore it is deleted and should be created
+		unstructuredObj = util.EmptyUnstructuredObjectFromIdentifier(objIdentifier)
```
MikeSpreitzer, line 50: FYI: I do not think the name of that func needs to say "object" twice.

MikeSpreitzer, line 50: FYI: Since this func is defined in the same package as the type of its first argument, this func could be a method on that type of object instead.

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -149,19 +121,17 @@ func (resolution *bindingPolicyResolution) toBindingSpec(gvkGvrMapper util.GvkGv
 	// iterate over all objects and build workload efficiently. No (GVR, namespace, name) tuple is
 	// duplicated in the objectIdentifierToKey map, due to the uniqueness of the Key. Therefore, whenever an object is about to
 	// be appended to an objects slice, we simply append.
-	for _, key := range resolution.objectIdentifierToKey {
-		gvr, found := gvkGvrMapper.GetGvr(key.GVK)
-		if !found {
-			return nil, fmt.Errorf("failed to get GVR for GVK %s", key.GvkKey())
-		}
+	for objIdentifier := range resolution.getObjectIdentifiers() {
+		gvr := schema.GroupVersionResource{Group: objIdentifier.GVK.Group, Version: objIdentifier.GVK.Version,
```
MikeSpreitzer, line 125: FYI, I would define a `GVR` method on `ObjectIdentifier`.

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -297,41 +267,43 @@ func workloadMatchesBindingSpec(bindingSpecWorkload *v1alpha1.DownsyncObjectRefe
...
+				objIdentifier.Resource != clusterScopeDownsyncObjects.GroupVersionResource.Resource {
+				continue
+			}
 
-			if _, found := objectIdentifierToKeyMap[formattedKey]; !found {
+			// check if object is in the objectNamesSet
+			if !objectNamesSet.Has(objIdentifier.ObjectName.Name) {
 				return false
 			}
 		}
```
MikeSpreitzer, line 309: This is much more complicated than necessary. Just convert each member of `clusterScopeDownsyncObjects.ObjectNames` to an ObjectIdentifier and test whether `objectIdentifiers.Has` it. Similar comment applies in the namespaced case.

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -185,13 +184,14 @@ func (resolver *bindingPolicyResolver) NoteBindingPolicy(bindingpolicy *v1alpha1
...
-// the resolution if exists.
+// AddObjectIdentifier adds an object's identifier to the resolution for
+// the given bindingpolicy key. If the object is already in the resolution,
+// it is not added again.
 //
 // The returned bool indicates whether the bindingpolicy resolution was changed.
-func (resolver *bindingPolicyResolver) NoteObject(bindingPolicyKey string,
-	obj runtime.Object) (bool, error) {
+// If no resolution is associated with the given key, an error is returned.
+func (resolver *bindingPolicyResolver) AddObjectIdentifier(bindingPolicyKey string,
```
MikeSpreitzer, line 193: FYI: "Add" is not quite the right word here, since we are explicitly contemplating the possibility that the ObjectIdentifier is already included. I would use something like "Ensure".

pkg/binding/controller.go
```diff
@@ -443,29 +424,29 @@ func (c *Controller) processNextWorkItem(ctx context.Context) bool {
...
 			// Put the item back on the workqueue to handle any transient errors.
 			c.workqueue.AddRateLimited(obj)
-			return fmt.Errorf("error syncing key '%#v': %s, requeuing", obj, err.Error())
+			return fmt.Errorf("error syncing object (identifier: %#v): %s, requeuing", objIdentifier, err.Error())
 		}
 		// Finally, if no error occurs we Forget this item so it does not
 		// get queued again until another change happens.
 		c.workqueue.Forget(obj)
-		logger.V(2).Info("Successfully synced", "object", obj)
+		logger.V(2).Info("Successfully synced", "object identifier", objIdentifier)
```
MikeSpreitzer, line 449: I have seen loggers object to keys with spaces in them.

pkg/binding/bindingpolicy.go
```diff
@@ -96,27 +103,26 @@ func (c *Controller) handleBindingPolicy(ctx context.Context, obj runtime.Object
...
+	if err = c.handleBindingPolicyFinalizer(ctx, bindingPolicy); err != nil {
+		return fmt.Errorf("failed to handle finalizer for bindingPolicy %s: %w", bindingPolicy.Name, err)
+	}
+
+	return c.updateResolutionForDeletedBindingPolicy(ctx, objIdentifier.ObjectName.Name)
 }
 
-func (c *Controller) handleBindingPolicyDeletion(ctx context.Context, bindingPolicy *v1alpha1.BindingPolicy) error {
-	if c.bindingPolicyResolver.ResolutionRequiresSingletonReportedState(bindingPolicy.GetName()) {
+func (c *Controller) updateResolutionForDeletedBindingPolicy(ctx context.Context, bindingPolicyName string) error {
```
MikeSpreitzer, line 114: FYI: The word "Deleted" in this method's name is not exactly right --- the object has not actually been deleted yet, right?

pkg/binding/bindingpolicy.go
```diff
@@ -96,27 +103,26 @@ func (c *Controller) handleBindingPolicy(ctx context.Context, obj runtime.Object
...
+	if c.bindingPolicyResolver.ResolutionRequiresSingletonReportedState(bindingPolicyName) {
 		// if the bindingpolicy required a singleton status, all selected objects should
 		// be requeued in order to remove the label
-		if err := c.requeueSelectedWorkloadObjects(ctx, bindingPolicy.GetName()); err != nil {
+		if err := c.requeueSelectedWorkloadObjects(ctx, bindingPolicyName); err != nil {
```
MikeSpreitzer, line 118: FYI: There can be multiple updates to a BindingPolicy while it is in the process of being removed.

pkg/binding/selectors.go
```diff
@@ -35,70 +36,82 @@ const emptyBindingPolicyName = ""
...
+	bindingPolicyRuntimeObjects, err := c.listBindingPolicies()
 	if err != nil {
 		return err
 	}
+
 	logger := klog.FromContext(ctx)
 
+	obj, err := c.getObjectFromObjectIdentifier(objIdentifier)
+	if errors.IsNotFound(err) {
+		// object is deleted, delete from all resolutions it exists in
```
MikeSpreitzer, line 49: .. and enqueue references to those Binding objects, right?

pkg/util/object-identifier.go
```diff
@@ -0,0 +1,99 @@
...
+
+	obj.SetGroupVersionKind(objIdentifier.GVK)
+	obj.SetNamespace(objIdentifier.ObjectName.Namespace)
+	obj.SetName(objIdentifier.ObjectName.Name)
+
+	return obj
+}
+
+func ObjIdentifierIsForCRD(objIdentifier ObjectIdentifier) bool { // CRDs might have different versions. therefore, using "any" in CRD version
+	return gvkMatches(objIdentifier.GVK, apiextensions.GroupName, AnyVersion, CRDKind)
```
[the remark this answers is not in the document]
vMaroon: This one has AnyVersion for version, I'll update the rest.

pkg/binding/bindingpolicy.go
vMaroon: After the first one goes through, the resolution would be deleted, and this block would not be entered.

pkg/util/object-identifier.go
MikeSpreitzer: We do not have to update all our files every year. OTOH, it is customary to write the current year in a new file.

pkg/binding/bindingpolicy.go
MikeSpreitzer: Ah, right.

pkg/binding/bindingpolicy-resolution.go
vMaroon: The problem is that the Kind is missing from the references in the Binding. To deal with that, I either have to:
1. Have the resolver work on a subtype of ObjectIdentifier that does not contain the Kind (or a new struct type)
2. Implement comparable for ObjectIdentifier to ignore Kind
3. Build sets for both with a type that contains only GVR and ObjectName

I think given that these are the alternatives, I think the implementation is acceptable. If you think otherwise, let me know.

pkg/util/object-identifier.go
vMaroon: Understandable, done.

vMaroon:
@MikeSpreitzer addressed comments.

pkg/binding/controller.go
MikeSpreitzer: Dang, I knew I should have been more specific. I wasn't, because I do not remember exactly _what_ was complaining. But the lesson I learned is that an Info key should look like a private identifier in Go: use only alphanumerics and maybe underscore, but lowerCamelCase is preferred.

pkg/binding/bindingpolicy-resolution.go
MikeSpreitzer: Oh, right, you can not use `Set::Has`.

pkg/binding/controller.go
vMaroon: Updated.

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -297,41 +267,43 @@ func workloadMatchesBindingSpec(bindingSpecWorkload *v1alpha1.DownsyncObjectRefe
...
+			if objIdentifier.GVK.Group != clusterScopeDownsyncObjects.GroupVersionResource.Group ||
+				objIdentifier.GVK.Version != clusterScopeDownsyncObjects.GroupVersionResource.Version ||
+				objIdentifier.Resource != clusterScopeDownsyncObjects.GroupVersionResource.Resource {
+				continue
+			}
 
-			if _, found := objectIdentifierToKeyMap[formattedKey]; !found {
+			// check if object is in the objectNamesSet
+			if !objectNamesSet.Has(objIdentifier.ObjectName.Name) {
 				return false
```
MikeSpreitzer, line 307: The logic in new lines 289 through here is backwards. These lines check whether the object names in clusterScopeDownsyncObjects are a _superset_ of those in objectIdentifiers with the right GVR. This func needs to check for subset rather than superset.

vMaroon: Aren't the directions equivalent under the guarantee of uniqueness of entries in both groups (resolution's object identifiers, and namespaced + cluster scoped identifiers) and the equality of sizes?

The groups match if and only if every object in either group exists in the other, given equality of sizes.

MikeSpreitzer: Ah, right.

pkg/binding/controller.go
```diff
@@ -359,57 +360,36 @@ func verbsSupportInformers(verbs []string) bool {
...
-
-				key.DeletedObject = &deletedObj
 
-				c.workqueue.Add(key)
-			}
-			return
-		}
-	}
-	c.workqueue.Add(key)
+	c.workqueue.Add(objIdentifier)
```
MikeSpreitzer, line 379: An Objectidentifier includes `ResourceVersion string`, which is not appropriate to include in an object reference in a workqueue.

vMaroon: @MikeSpreitzer this was accidentally left here, I make no use of this field because I agree - it should not be here.

This is why I didn't want to continue working tonight 😢.

Feel free to remove it in the rebase PR.

vMaroon: Removed.

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -54,4 +54,4 @@ spec:
         - "--leader-elect"
         - "--wds-name={{.Values.ControlPlaneName}}"
         - "--api-groups={{.Values.APIGroups}}"
-        - -v=2
+        - -v=5
```
MikeSpreitzer, line 57: I made this change for debugging; we probably do not want to run this high normally.

pkg/util/object-identifier.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/kubestellar/kubestellar/api/control/v1alpha1"
+)
+
+// ObjectIdentifier struct is used to add items to the workqueue.
+// The ObjectIdentifier contains all the necessary information to retrieve an object.
+type ObjectIdentifier struct {
+	GVK        schema.GroupVersionKind
+	Resource   string
+	ObjectName cache.ObjectName
+}
```
MikeSpreitzer, line 36: This has the same contents as the type SourceRef in the same package. Do we really need two types?

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -66,26 +65,29 @@ type BindingPolicyResolver interface {
 	// singleton status reporting requirement in the resolution.
 	NoteBindingPolicy(bindingpolicy *v1alpha1.BindingPolicy)
```
MikeSpreitzer, line 66: FYI: Sorry I did not notice this before, but this method makes a BindingPolicyResolver care what those bindingPolicyKey strings are --- this method uses a BindingPolicy's Name as he bindingPolicyKey. So the comment on this interface is inaccurate. Does not have to be fixed in this PR.

MikeSpreitzer (approved):
I think that this is good enough to merge and continue working in follow-on PRs.

/lgtm
/approve
Thanks!

pkg/util/object-identifier.go
vMaroon: I think at the moment we can live with that - soon (my) focus will move to the status controller, which will involve redesigning some things, and aligning it - code-wise - to the binding controller.

If it can wait until the latter, we will continue with only the ObjectIdentifier, if not, I can refactor the `SourceRef` out.

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -200,58 +201,55 @@ func (resolver *bindingPolicyResolver) NoteObject(bindingPolicyKey string,
...
 }
 
-// SetDestinations updates the maintained bindingpolicy's
-// destinations resolution for the given bindingpolicy key.
-// The given destinations set is expected not to be mutated after this call.
-func (resolver *bindingPolicyResolver) SetDestinations(bindingPolicyKey string,
-	destinations sets.Set[string]) {
+// GetObjectIdentifiers returns a copy of the object identifiers associated
+// with the given bindingpolicy key.
+// If no resolution is associated with the given key, an error is returned.
```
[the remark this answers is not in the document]
vMaroon: I think you are referring to the single case this is done, which is when checking if the resolver's `EnsureObjectIdentifierWithVersion` returned an error prefixed with `bindingPolicyResolutionNotFoundErrorPrefix`, during the call to controller::updateResolutions for a given object identifier.

In this case, `EnsureObjectIdentifierWithVersion` does have to return this error, but in the special scenario noted here:
```
resolutionUpdated, err := c.bindingPolicyResolver.EnsureObjectIdentifierWithVersion(bindingPolicy.GetName(),
			objIdentifier, objMR.GetResourceVersion())
		if err != nil {
			if errorIsBindingPolicyResolutionNotFound(err) {
				// this case can occur if a bindingpolicy resolution was deleted AFTER
				// starting this iteration and BEFORE getting to the NoteObject function,
				// which occurs if a bindingpolicy was deleted in this time-window.
				logger.V(4).Info("skipped EnsureObjectIdentifierWithVersion for object-identifier (%v) and "+
					"bindingpolicy %s because the latter was deleted", objIdentifier, bindingPolicy.GetName())
				continue
			}
		...
```

An error should not be reported. Do you think it is fine to keep this single instance of checking the error? Should the check be changed to a more common implementation? Or is this single-case usage fine?

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -66,26 +65,29 @@ type BindingPolicyResolver interface {
...
+	RemoveObjectIdentifier(bindingPolicyKey string, objIdentifier util.ObjectIdentifier) bool
+	// GetObjectIdentifiers returns the object identifiers associated with the
+	// given bindingpolicy key.
+	// If no resolution is associated with the given key, an error is returned.
+	GetObjectIdentifiers(bindingPolicyKey string) (sets.Set[util.ObjectIdentifier], error)
+
 	// SetDestinations updates the maintained bindingpolicy's
 	// destinations resolution for the given bindingpolicy key.
 	// The given destinations set is expected not to be mutated after this call.
 	SetDestinations(bindingPolicyKey string, destinations sets.Set[string])
```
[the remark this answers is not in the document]
vMaroon: You're right, `SetDestinations` does not return an error because it is always called after `NoteBindingPolicy`, but that should not impact the design of the resolver. 

Updated.

pkg/binding/bindingpolicy.go
```diff
@@ -127,17 +133,17 @@ func (c *Controller) requeueSelectedWorkloadObjects(ctx context.Context, binding
...
-		obj, err := c.getObjectFromKey(*key)
+	objs := make([]runtime.Object, 0, len(objectIdentifiers))
+	for objIdentifier := range objectIdentifiers {
+		obj, err := c.getObjectFromIdentifier(objIdentifier)
 		if err != nil || obj == nil {
-			return fmt.Errorf("failed to get object from key: %w", err)
+			return fmt.Errorf("failed to get object from identifier: %w", err)
 		}
 
 		objs = append(objs, obj)
```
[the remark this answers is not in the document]
vMaroon: You're right, and not only that, but this also has a bug in case the controller::getObjectFromIdentifier fails due to an object being deleted, which should not stop the queueing of the rest of the objects.

Updated.

pkg/binding/controller.go
```diff
@@ -416,57 +419,42 @@ func verbsSupportInformers(verbs []string) bool {
...
 
 // Event handler: enqueues the objects to be processed
-// At this time it is very simple, more complex processing might be required here
-func (c *Controller) handleObject(obj any) {
-	rObj := obj.(runtime.Object)
-	c.logger.V(2).Info("Got object event", "obj", util.RefToRuntimeObj(rObj))
-	c.enqueueObject(obj, false)
+// At this time it is very simple, more complex processing might be required
+// here.
+func (c *Controller) handleObject(obj any, eventType string) {
```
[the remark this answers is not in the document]
vMaroon: You're right, but what we need here is the resource, and not the kind.

Updated.

pkg/binding/controller.go
```diff
@@ -416,57 +419,42 @@ func verbsSupportInformers(verbs []string) bool {
...
 
-// enqueueObject converts an object into a key struct which is then put onto the work queue.
-func (c *Controller) enqueueObject(obj interface{}, skipCheckIsDeleted bool) {
-	var key util.Key
-	var err error
-	if key, err = util.KeyForGroupVersionKindNamespaceName(obj); err != nil {
+// enqueueObject converts an object into an ObjectIdentifier struct which is
+// then put onto the work queue.
+func (c *Controller) enqueueObject(obj interface{}) {
+	objIdentifier, err := util.IdentifierForObject(obj.(util.MRObject), c.gvkToGvrMapper)
```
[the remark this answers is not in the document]
vMaroon: Mapper completely deleted.

PR #1835, opened 2024-02-29
:sparkles: Remove legacy workflows

ezrasilvera:
These workflows are used only for the 0.14 release.  In the rare case were we need to fix 0.14 the tests can be done through the Makefile. 
We don't need to maintain 11 workflow files that are never used

## Related issue(s)

Fixes #


pdettori (approved):
/lgtm

Issue #1540, opened 2024-01-12
feature: reduce agent footprint for OCM transport


pdettori:
### Feature Description

the OCM work and registration agent run on the WECs in separate pods currently. It is possible to reduce the agent footprint consolidating them in one pod, as enabled by OCM agent flag.

### Proposed Solution

According to OCM docs:

"To reduce the footprint of agent in the managed cluster, singleton mode is introduced since v0.12.0. In the singleton mode, the work and registration agent will be run as a single pod in the managed cluster."

Need to add flag --singleton in join command. See https://open-cluster-management.io/getting-started/installation/register-a-cluster/

We can add this to our guides/scripts.

### Want to contribute?

- [ ] I would like to work on this issue.

PR #1846, opened 2024-03-03
:sparkles: Reduce agent footprint for OCM transport

andreyod:
## Summary
Add `--singleton` flag in `clusteradm join` command.
According to OCM docs https://open-cluster-management.io/getting-started/installation/register-a-cluster/ :
"To reduce the footprint of agent in the managed cluster, singleton mode is introduced since v0.12.0. In the singleton mode, the work and registration agent will be run as a single pod in the managed cluster."

Without `--singleton` flag:
```
kubectl --context cluster1 get deployments -n open-cluster-management-agent
NAME                            READY   UP-TO-DATE   AVAILABLE   AGE
klusterlet-registration-agent   1/1     1            1           19h
klusterlet-work-agent           1/1     1            1           19h
```
With `--singleton` flag:
```
kubectl --context cluster1 get deployments -n open-cluster-management-agent
NAME               READY   UP-TO-DATE   AVAILABLE   AGE
klusterlet-agent   1/1     1            1           81m
```
## Related issue(s)

Fixes https://github.com/kubestellar/kubestellar/issues/1540

### DIFF
--- a/docs/content/direct/example-wecs.md
+++ b/docs/content/direct/example-wecs.md
@@ -13,7 +13,7 @@ register them with the hub as descibed in the
    for cluster in "${clusters[@]}"; do
       kind create cluster --name ${cluster}
       kubectl config rename-context kind-${cluster} ${cluster}
-      clusteradm --context imbs1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk '{print $0 " --context '${cluster}' '${flags}'"}' | sh
+      clusteradm --context imbs1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk '{print $0 " --context '${cluster}' --singleton '${flags}'"}' | sh
    done   
    ```
 
--- a/test/e2e/common/setup-kubestellar.sh
+++ b/test/e2e/common/setup-kubestellar.sh
@@ -118,7 +118,7 @@ function create_cluster() {
   cluster=$1
   kind create cluster --name $cluster
   kubectl config rename-context kind-${cluster} $cluster
-  clusteradm --context imbs1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk '{print $0 " --context '${cluster}' --force-internal-endpoint-lookup"}' | sh
+  clusteradm --context imbs1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk '{print $0 " --context '${cluster}' --singleton --force-internal-endpoint-lookup"}' | sh
 }
 
 create_cluster cluster1


nirrozenbaum:
looks good to me. 
I will let @pdettori and @MikeSpreitzer comment as well before lgtm.

docs/content/direct/example-wecs.md
```diff
@@ -7,13 +7,13 @@ register them with the hub as descibed in the
...
-   : set flags to "" if you have installed KubeStellar on an OpenShift cluster
-   flags="--force-internal-endpoint-lookup"
+   : set flags to "--singleton" if you have installed KubeStellar on an OpenShift cluster
+   flags="--singleton --force-internal-endpoint-lookup"
    clusters=(cluster1 cluster2);
    for cluster in "${clusters[@]}"; do
       kind create cluster --name ${cluster}
       kubectl config rename-context kind-${cluster} ${cluster}
-      clusteradm --context imbs1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk '{print $0 " --context '${cluster}' '${flags}'"}' | sh
+      clusteradm --context imbs1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk -v fl="$flags" '{print $0 " --context '${cluster}' " fl}' | sh
```
MikeSpreitzer, line 16: Yikes the injection of `$flags` is getting complicated. Couldn't we do it more simply, as follows?

```shell
      clusteradm --context imbs1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk '{print $0 " --context ${cluster} ${flags}"}' | sh
```

Also: does this PR really need a change here?

andreyod: It looks like `awk` have an issue with environment variables so when I added the second flag it didn't work. The `-v` makes the variable native to the `awk` so this is the recommended way.
I tried to reverse the quotes but this didn't work either.
We already have `$flags` and it's in the user-guide, and `--singleton` is a flag (just like `--force-internal-endpoint-lookup`) so it felt natural to add it to `$flags`

MikeSpreitzer: The original version of this command is not even asking `awk` to deal with the shell variable `flags` --- the syntax was asking the shell to interpolate the value of `flags` into the command line.

MikeSpreitzer: I think we do not have to be too rigid about the word "flags". This is a variable that we created to solve a problem. We do not have to make the solution more complicated than necessary. Also, see that `--context` flag that is already not in `flags`? I think the simplest solution is to make `--singleton` just as constant and direct and literal as `--context`. Rename the variable if you like.

andreyod:
> There were more edits originally (see [08a2e7f](https://github.com/kubestellar/kubestellar/commit/08a2e7f85ce507abf363ae3cc272a2d7e5628f6e)) --- doc update, comment update. I thought these other edits were good.

If we don't change the `flags`, those edits are not needed.

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #1854, opened 2024-03-04
✨ Use descriptive placeholder instead of obsolete tags

MikeSpreitzer:
## Summary
This PR replaces a couple of literal tags that get replaced by the publishing workflow (`.github/workflows/goreleaser.yml`) with a statement to that effect.

## Related issue(s)

Fixes #

### DIFF
--- a/chart/templates/controller-manager.yaml
+++ b/chart/templates/controller-manager.yaml
@@ -203,7 +203,7 @@ spec:
         - --wds-name={{.Values.ControlPlaneName}}
         - --api-groups={{.Values.APIGroups}}
         - -v=2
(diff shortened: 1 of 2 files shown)


francostellari:
@MikeSpreitzer I personally think that this is not a good idea because it makes the chart in main unusable with building locally.
I think latest should be the default in main, like for kubeflex, and could be customized to a specific tag during the release process.

MikeSpreitzer:
@francostellari: we have support for testing locally, it replaces the placeholders. You can see it in the Makefile target `local-chart`. The E2E tests use that.

ezrasilvera:
Thanks. 
Approving after the call we had 
/lgtm
/approve

PR #1857, opened 2024-03-05
✨ Introduce Deployment object for transport controller

MikeSpreitzer:
## Summary
This PR changes the E2E test setup to run the transport controller using a Deployment object.

This involves a network hack that only works when the hosting cluster is a kind cluster. It begs the question of what the more general story should be.

Even more fundamentally, this involves a config hack: loading the user's kubeconfig into a Secret object. If the transport controller could use kubeflex to get client configs then neither of these hacks would be needed --- but that would only work with kubeflex.

## Related issue(s)

Fixes #

### DIFF
--- a/test/e2e/common/setup-kubestellar.sh
+++ b/test/e2e/common/setup-kubestellar.sh
@@ -98,15 +98,73 @@ OCM_TRANSPORT_PLUGIN_DIR="$(pwd)"
 pwd
 echo "replace github.com/kubestellar/kubestellar => ${KUBESTELLAR_DIR}/" >> go.mod
 go mod tidy
-make build
-mv ./bin/ocm-transport-plugin ${KUBESTELLAR_DIR}/ocm-transport-plugin
+(
+    if docker version | grep -qi podman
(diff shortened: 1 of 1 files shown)


test/e2e/common/setup-kubestellar.sh
```diff
@@ -98,15 +98,73 @@ OCM_TRANSPORT_PLUGIN_DIR="$(pwd)"
...
+    if docker version | grep -qi podman
+    then export DOCKER_HOST=unix://$HOME/.local/share/containers/podman/machine/qemu/podman.sock
+    fi
+    KO_DOCKER_REPO=ko.local ko build -B ./cmd/transport-controller -t local --platform linux/$(go env GOARCH)
+)
+kind load --name kubeflex docker-image ko.local/transport-controller:local
+
 cd "${KUBESTELLAR_DIR}"
 pwd
-rm -rf ${OCM_TRANSPORT_PLUGIN_DIR}
```
waltforme, line 105: Should we keep doing this `rm`?

MikeSpreitzer:
See #1863 instead.

PR #1858, opened 2024-03-05
🌱 Some follow-up changes after PR1828

waltforme:
## Summary
This PR contains changes to address some of the open comments to #1828.

/cc @MikeSpreitzer

### DIFF
--- a/pkg/binding/controller.go
+++ b/pkg/binding/controller.go
@@ -97,6 +97,7 @@ type Controller struct {
 // Create a new binding controller
 func NewController(parentLogger logr.Logger, wdsRestConfig *rest.Config, imbsRestConfig *rest.Config,
 	wdsName string, allowedGroupsSet sets.Set[string]) (*Controller, error) {
+	logger := parentLogger.WithName(controllerName)
 
 	kubernetesClient, err := kubernetes.NewForConfig(wdsRestConfig)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/binding/controller.go
```diff
@@ -110,16 +111,17 @@ func NewController(parentLogger logr.Logger, wdsRestConfig *rest.Config, imbsRes
...
 	nGVRs, err := countGVRsViaAggregatedDiscovery(kubernetesClient)
 	if err != nil {
-		parentLogger.Error(err, "Not able to count GVRs via aggregated discovery") // but we can still continue
+		logger.Error(err, "Not able to count GVRs via aggregated discovery") // but we can still continue
 	}
 	// tuning the rate limiter based on the number of GVRs is tested to be working well
-	wdsRestConfig.Burst = computeBurstFromNumGVRs(nGVRs)
-	wdsRestConfig.QPS = computeQPSFromNumGVRs(nGVRs)
-	parentLogger.V(1).Info("Parameters of the client's token bucket rate limiter", "burst", wdsRestConfig.Burst, "qps", wdsRestConfig.QPS)
+	wdsRestConfigTuned := rest.CopyConfig(wdsRestConfig)
```
MikeSpreitzer, line 117: Should the status controller also be using tuned client config?

waltforme: Maybe not. Because (1) the status controller doesn't do discovery, (2) the status controller shares binding controller's same set of informers and same set of listers thus doesn't do the per-resource ListAndWatch one more time.

However, when I was revisiting the status controller. I found some points maybe worth discussing.
1. Since an informer for BindingPolicy is already within the set of per-resource informers, is it possible to reuse it, instead of creating a separate one as shown below? https://github.com/kubestellar/kubestellar/blob/272096782b39d2ee7102775bc03d677f3991e004/pkg/status/controller.go#L142 Further, should the status controller even watch bindingpolicies? I actually don't see the separate bindingpolicy informer's event handlers been added.
2. The status controller does use that same set of per-resource listers to access workload objects in WDSes. But I don't see the status controller use them to watch anything. I quickly commented out the informers and the e2e tests seems to be OK.

Maybe we can do some cleaning up here?

MikeSpreitzer (commented):
I am OK with doing further work in follow-on PRs.
/lgtm
Thanks!

pkg/binding/controller.go
ezrasilvera: Good points.  @pdettori Any comment on that ?

pdettori: @waltforme @ezrasilvera those are good points. It is certainly possible that there is some overlap as some code borrowed from the code for the binding (a.k.a. placement controller). @waltforme by all means pls. consider another PR for more clean up.

waltforme: Sure will do. Opened #1862 to track it.

PR #1863, opened 2024-03-07
✨ run transport controller in a pod

### LINKED ISSUE #1809: feature: deploy transport controller in a pod
### Feature Description

it should be possible to deploy transport controller in a pod. ideally using helm

### Proposed Solution

the above.

### Want to contribute?

### Additional Context

_No response_

nirrozenbaum:
## Related issue(s)

Fixes #1809


nirrozenbaum:
@francostellari (cc @pdettori) once this PR is merged, please feel free to look on how to run the same using a helm chart.
need to pay attention that we need the option to run transport controller using both official image (as can be seen in example.md in this PR) and using a local image (as can be seen in e2e setup-kubestellar script).

deploy/transport-controller-deployment.yaml.template
```diff
@@ -0,0 +1,89 @@
...
+      name: transport-controller
+  template:
+    metadata:
+      labels:
+        name: transport-controller
+    spec:
+      serviceAccountName: kubestellar-controller-manager
+      initContainers:
+      - name: setup-wds-kubeconfig
+        image: bitnami/kubectl
```
francostellari, line 51: In the other repos we use our custom quay.io/kubestellar/kubect images because of the bandwidth limitations in using DockerHub images. Can we agree on what to use for all the repos?

nirrozenbaum: good point, sure. 
I was not aware that we have kubectl image under kubestellar.
of course just need to validate it works as expected.
pushed a fix (for both comments).

francostellari: you may have to specify a tab/version since we don't have latest/stable
`quay.io/kubestellar/kubectl:1.27.8`

pdettori: I think there was also some issues with some of the bitnami images not being multi arch, so in any case it is better to use our own

deploy/transport-controller-deployment.yaml.template
```diff
@@ -0,0 +1,89 @@
...
+        volumeMounts:
+        - name: config-volume
+          mountPath: /mnt/config
+        - name: shared-volume
+          mountPath: /mnt/shared
+      - name: setup-imbs-kubeconfig
+        image: quay.io/kubestellar/kubectl
+        imagePullPolicy: Always
+        command: [ "bin/sh", "-c", "bash /mnt/config/get-kubeconfig.sh ${IMBS_NAME} false | base64 -d > /mnt/shared/transport-kubeconfig"]
+        volumeMounts:
```
pdettori, line 63: why the flag for in-cluster is false here ? isn't IMBS also in-cluster ?

deploy/transport-controller-deployment.yaml.template
```diff
@@ -0,0 +1,89 @@
...
+          - --wds-name=${WDS_NAME}
+          - -v=4
+          volumeMounts:
+          - name: shared-volume
+            mountPath: /mnt/shared
+            readOnly: true
+      hostAliases:
+      - ip: ${HOST_IP}
+        hostnames: [ "wds1.localtest.me", "imbs1.localtest.me" ]
+      volumes:
```
pdettori, line 84: it looks like these values are hardcoded ... they should be parametrized, no ? we could use other names for the wds/imbs and there could be multiple wds...

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: from transport point of view - no.
transport runs in cluster wds, while imbs is external

deploy/transport-controller-deployment.yaml.template
francostellari: additionally, this may work for local testing but what about deployment in OCP, is `hostAliases` workable there?

pdettori: I am also wondering ... why those hostnames are needed at all ? The pod should always be using the in-cluster config, unless you intend to run the pod on a cluster different from where wds and imbs are located.

francostellari:
@pdettori @nirrozenbaum can you clarify if this controller needs to run for each `wds`?  does it run in the `wds` cp? or i the cluster running the `wds` cp? Perhaps you have a diagram posted somewhere?

MikeSpreitzer:
Why not squash the 10 commits here together?

deploy/transport-controller-deployment.yaml.template
MikeSpreitzer: I thought we already agreed we are using the "Inventory Transport Space (ITS)" terminology rather than "IMBS". I know some stuff has not been updated, but that does not undo the decision. I would not harden the mistake by hard-coding it in something in git.

MikeSpreitzer:
@francostellari : yes, the transport controller has to be run once for each WDS. It can not run "in" a WDS because a "space" can not necessarily run pods. That is why it runs in the hosting cluster and is connected to the WDS.

.github/workflows/pr-test-multicluster.yml
```diff
@@ -71,6 +67,10 @@ jobs:
         if: always()
         run: kubectl --context kind-kubeflex logs -n wds1-system $(kubectl --context kind-kubeflex get pod -n wds1-system --selector=control-plane=controller-manager -o jsonpath='{.items[0].metadata.name}')
 
+      - name: show transport controller log
+        if: always()
+        run: kubectl --context kind-kubeflex -n wds1-system logs $(kubectl --context kind-kubeflex -n wds1-system get pods -l name=transport-controller -o jsonpath='{.items[*].metadata.name}')
```
MikeSpreitzer, line 72: I think the `*` here is dangerous.

```shell
mspreitz@mjs13 kubestellar % kubectl get pods -A -o jsonpath='{.items[*].metadata.name}'
coredns-76f75df574-f58ph coredns-76f75df574-m7zvk etcd-kind-control-plane kindnet-vhrs7 kube-apiserver-kind-control-plane kube-controller-manager-kind-control-plane kube-proxy-78nn9 kube-scheduler-kind-control-plane local-path-provisioner-7577fdbbfb-6m7cx
```

deploy/transport-controller-deployment.yaml.template
```diff
@@ -0,0 +1,89 @@
...
+  name: transport-controller-config
+  namespace: ${WDS_NAME}-system
+data:
+  get-kubeconfig.sh: |
+    #!/bin/bash
+    # this script receives a ControlPlane name and a parameter
+    # that determines whether to extract the ControlPlane's in-cluster kubeconfig
+    # or the external kubeconfig (if set to "true", the first will be retrieved).
+    # The function returns the requested kubeconfig's content in base64.
+    # it assumes that it runs within a kubeflex context.
```
MikeSpreitzer, line 13: I am not sure what "a kubeflex context" means. Do you mean that `kubectl` is configured (in the usual ways, by `$KUBECONFIG` and `~/.kube/config`) to access the hosting cluster?

deploy/transport-controller-deployment.yaml.template
```diff
@@ -0,0 +1,89 @@
...
+
+    cpname="$1"
+    get_incluster_key="$2"
+
+    key=""
+    if [[ $get_incluster_key == "true" ]]; then
```
MikeSpreitzer, line 19: I would quote the LHS here, to protect against the case where it has spaces in it.

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: looks like quay.io/kubestellar/kubectl image is working different than bitnami. 
probably a minor difference, but it broke the init container setup.
will debug and push a fix soon

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: all e2e tests are currently using the name `imbs`, I was just using the same pattern.
I agree with the comment, but this can be done in a separate PR (after the release) to align name to ITS in all places. 
btw, after introducing inventory, it will be renamed (again) to transport space.
so we may end up with:
wds, transport space, inventory space.
and transport+inventory can run on the same space, so it may become ITS.

nirrozenbaum:
> Why not squash the 10 commits here together?
@MikeSpreitzer 
yes, possible. 
I will squash as soon as I make it pass the tests.

.github/workflows/pr-test-multicluster.yml
nirrozenbaum: will change to `items[0]`

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: the LHS is just `true` or `false`, which represents: `true` for using in cluster, `false` for external kubeconfig.
I can add it if you insist, but don't see much value.

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: seems like `quay.io/kubestellar/kubectl` doesn't have bash.

nirrozenbaum: pushed a fix. will see if e2e works now

.github/workflows/pr-test-multicluster.yml
nirrozenbaum: fixed

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: yes, fixed.

nirrozenbaum:
@MikeSpreitzer I fixed the change which broke the e2e.
if all e2e pass, I will squash all commits.
let me know if there are more comments.

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: @francostellari @pdettori before adding the alias I couldn't run the controller using a local image loaded kubeflex.
this is a crucial part in being able to run e2e for code changes of KS repo.
keep in mind that alias is NOT replacing the name. it's just an alias (another name), which allows us to run in case of local env. If I'm not mistaken, I don't think there should be any problem with running in any cluster.

but you are correct that I missed the parameterization of this (probably when I tested locally). will push a fix.

nirrozenbaum: @pdettori regarding the fact that there are multiple wds - you are correct but it's not relevant to this deployment, since transport is per wds. this specific transport controller deployment is per specific WDS. so we should use the WDS_NAME variable.

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: e2e are working with quay.io/kubestellar/kubectl image. resolving thread.

deploy/transport-controller-deployment.yaml.template
```diff
@@ -0,0 +1,89 @@
...
+        image: bitnami/kubectl
+        imagePullPolicy: Always
+        command: [ "bin/sh", "-c", "bash /mnt/config/get-kubeconfig.sh ${WDS_NAME} true | base64 -d > /mnt/shared/wds-kubeconfig"]
+        volumeMounts:
+        - name: config-volume
+          mountPath: /mnt/config
+        - name: shared-volume
+          mountPath: /mnt/shared
+      - name: setup-imbs-kubeconfig
+        image: bitnami/kubectl
```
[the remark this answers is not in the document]
nirrozenbaum: fixed

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: fixed to use parameters instead of hardcoded.

nirrozenbaum:
@MikeSpreitzer squashed. left one commit of Maroon (not to remove commit of someone else).

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: wds:
transport controller is deployed in the hosting cluster in the wds1-system namespace (since it's per wds).
for this I think we can easily understand we need to use in cluster config.

imbs:
transport controller is running in wds, not in imbs (or ITS). 
therefore, the transport controller is external to imbs.
the transport controller pod is running in a pod in WDS, while imbs is a different space. 
we can even run imbs not through KubeFlex (just a regular kind cluster) and use it the same with the external access.
@MikeSpreitzer maybe you have a better explanation than the above?

ezrasilvera: @nirrozenbaum  I think that both spaces should be IN-Cluster

deploy/transport-controller-deployment.yaml.template
ezrasilvera: I don't think we use `imbs`  as a name. I agree that the current controller has assumption of a single imbs, but I think it searches KF for a space with imbs label so there is flexibility in the name

deploy/transport-controller-deployment.yaml.template
ezrasilvera: @nirrozenbaum the transport controller is not running in a Pod in WDS as WDS can't run any pod. The TC is running on the hosting cluster and is watching/interacting with both WDS and IMBS .  Both of those spaces/CP are hosted in the same cluster as the TC Pod so you should access the in-cluster. 
When you were running the controller as EXE you were indeed external to the cluster (and then you were external to BOTH spaces) .

ezrasilvera: @pdettori ^^ ??

deploy/transport-controller-deployment.yaml.template
```diff
@@ -0,0 +1,90 @@
...
+  selector:
+    matchLabels:
+      name: transport-controller
+  template:
+    metadata:
+      labels:
+        name: transport-controller
+    spec:
+      serviceAccountName: kubestellar-controller-manager
+      initContainers:
```
ezrasilvera, line 49: Maybe I'm missing something. Why do we need two different init containers and can't generate both files using a single one ?

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: I just validated and ran it locally using incluster for both wds+imbs and it works.
maybe since imbs is vcluster, it also works using the external access.

in general, I think using external access is the "more correct" way, cause this allows us to have it working in any kind of setup, with imbs in KF or without KF, in a pod or in process.
on the other hand, using the incluster is specifically for running everything in KubeFlex. 
it will work in current setup cause we use KF for all.

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: it's possible.
I did it just for separation of concerns, for example - If I want to run IMBS in kind cluster (just ocm hub in kind) I can easily remove the imbs-setup init container and mount imbs kubeconfig in a secret instead.
it can be done in a single init container but I like more the separation since the way we get WDS kubeconfig is not necessarily identical to how we get the IMBS kubeconfig.

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: I mean "more correct" in theory. in practice, here we always use KF.

deploy/transport-controller-deployment.yaml.template
MikeSpreitzer: While the second arg to this script _should_ be `true` or `false`, it is generally better if a script is prepared for the caller to goof up.

deploy/transport-controller-deployment.yaml.template
pdettori: @nirrozenbaum remember that although you have expressed a desire to use also an ITS (a.k.a IMBS) as an external cluster you are getting secrets for vcluster that are created in the kubeflex hosting cluster, and kubeflex creates that in-cluster key. I do not know how the external address based on `localtest.me` would work for a process running in a pod because it resolves to 127.0.0.1 and a pod would see that as its local network namespace.

nirrozenbaum: my intention was only "conceptually". we can conceptually run ITS in a different cluster and use the external.
but I have no problem changing to incluster, this deployment is used when everything is in KF.
if we want in the future to run ITS in a different cluster outside of KF we will take of that then (anyway current deployment reads from KF, as you mentioned @pdettori).

I will change to use incluster for both.

nirrozenbaum: fixed

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: fixed

deploy/transport-controller-deployment.yaml.template
ezrasilvera: @nirrozenbaum the running in "wds-namespace-system"  is not relevant at all here. We have a mix-up in the definitions. 
You have an api-server that is exposed as the WDS space. When you deploy resources into WDS you basically interact with that api-server and the etcd server it uses. 
Your Transport Controller is NOT deployed to WDS at all , it deployed to a regular cluster (the hosting cluster in your case) , you supply that controller with config information to access other "clusters"  in your case the api-servers of WDS and IMBS.  
Because all the Pods involved - the transport controller the WDS-api-server and the IMBS-api-server  reside ion the same cluster you can and should use the in-cluster access. If you use the external access the Pod won't be able to resolve the cluster IP.
We need to be carful when we say "run in a cluster"  VS "interact with cluster" .

ezrasilvera: @pdettori It reminds me that I think the KS controller has a bug on how it test if to access in-cluster Vs external.  I think that the current code will choose "in-cluster"  as long as the controller is running in a cluster (even if that cluster is different than that of KF). I guess the assumption is that KS controller pod and KF both run in the same cluster

deploy/transport-controller-deployment.yaml.template
nirrozenbaum: @ezrasilvera not if two different setups require using different images (e.g, image that contains some tool). 
if there is no issue with keeping two separate init containers, I prefer keeping the two

docs/content/direct/examples.md
```diff
@@ -60,20 +60,14 @@ manager which connects to the `wds1` front-end and the `imbs1` OCM control plane
...
    ```shell
-   OS=$(go env GOOS)
-   ARCH=$(go env GOARCH)
-   OCM_TRANSPORT_PLUGIN_RELEASE="0.1.0-rc2"
-   OCM_TRANSPORT_PLUGIN_TAR=ocm-transport-plugin_${OCM_TRANSPORT_PLUGIN_RELEASE}_${OS}_${ARCH}.tar.gz
-   wget https://github.com/kubestellar/ocm-transport-plugin/releases/download/v${OCM_TRANSPORT_PLUGIN_RELEASE}/${OCM_TRANSPORT_PLUGIN_TAR}
-   tar -xvf ${OCM_TRANSPORT_PLUGIN_TAR} transport-controller
-   rm -fr bin ${OCM_TRANSPORT_PLUGIN_TAR}
-   ./transport-controller --transport-context imbs1 --wds-context wds1 --wds-name wds1 &> transport.log &
+   ./hack/deploy-transport-controller.sh wds1 imbs1 ghcr.io/kubestellar/ocm-transport-plugin/transport-controller:0.1.0-rc4
```
MikeSpreitzer, line 70: We have never told the user what the current working directory should be. We have not, before this PR, required a user to `git clone` this directory. This doc does not mention the need to `git clone`.

hack/deploy-transport-controller.sh
```diff
@@ -0,0 +1,25 @@
...
+# Transport deployment script assumes it runs within a kubeflex context.
+
+export WDS_NAME="$1" ## first argument is WDS name
+export IMBS_NAME="$2" ## second argument is IMBS name
+export TRANSPORT_CONTROLLER_IMAGE="$3" ## third argument is transport controller image
+export HOST_IP=$(ifconfig | grep -w inet | awk '{ print $2 }' | grep -v 127.0.0.1 | head -1)
+export IMAGE_PULL_POLICY="${IMAGE_PULL_POLICY:=Always}"
+
+SRC_DIR="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)"
+cpname='$cpname' get_incluster_key='$get_incluster_key' secret_name='$secret_name' secret_namespace='$secret_namespace' key='$key' envsubst < "${SRC_DIR}/../deploy/transport-controller-deployment.yaml.template"  | kubectl apply -f -
```
MikeSpreitzer, line 25: The `cpname='$cpname' get_incluster_key='$get_incluster_key' secret_name='$secret_name' secret_namespace='$secret_namespace' key='$key'` part is awkward and fragile. It requires maintainers to keep it consistent with the script in the ConfigMap. I would use two `kubectl apply` commands, one using `envsubst` and one not. Or...

Wouldn't it be simpler to inline the content of that transport-controller-deployment.yaml.template file, making this script say

```shell
kubectl apply -f - <<EOF
(template goes here, using full ordinary shell syntax -- including \$ )
EOF
```

test/e2e/common/setup-kubestellar.sh
```diff
@@ -86,27 +86,29 @@ echo "wds1 created."
...
 KUBESTELLAR_DIR="$(pwd)"
-## this is a temp solution - run as executble process. this should be replaced to run as pod, ideally using helm
-OCM_TRANSPORT_PLUGIN_RELEASE="0.1.0-rc2"
+OCM_TRANSPORT_PLUGIN_RELEASE="0.1.0-rc4"
 curl -sL https://github.com/kubestellar/ocm-transport-plugin/archive/refs/tags/v${OCM_TRANSPORT_PLUGIN_RELEASE}.tar.gz | tar xz
 cd ocm-transport-plugin-${OCM_TRANSPORT_PLUGIN_RELEASE}
 OCM_TRANSPORT_PLUGIN_DIR="$(pwd)"
 pwd
 echo "replace github.com/kubestellar/kubestellar => ${KUBESTELLAR_DIR}/" >> go.mod
-go mod tidy
```
MikeSpreitzer, line 100: I do not think this PR removes the need for `go mod tidy`.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -86,27 +86,29 @@ echo "wds1 created."
...
 cd "${KUBESTELLAR_DIR}"
 pwd
 rm -rf ${OCM_TRANSPORT_PLUGIN_DIR}
 echo "running ocm transport plugin..."
-./ocm-transport-plugin --transport-context imbs1 --wds-context wds1 --wds-name wds1 -v=4 &> transport.log &
+kubectl config use-context kind-kubeflex ## transport deployment script assumes it runs within kubeflex context
+IMAGE_PULL_POLICY=Never ./hack/deploy-transport-controller.sh wds1 imbs1 ko.local/transport-controller:${OCM_TRANSPORT_PLUGIN_RELEASE}
 
-echo "transport controller is running as background process."
+kubectl -n wds1-system wait --for=condition=Ready pod/$(kubectl -n wds1-system get pods -l name=transport-controller -o jsonpath='{.items[*].metadata.name}')
```
MikeSpreitzer, line 108: The `*` here is dangerous.

`kubectl wait` will immediately fail if the relevant object does not exist yet. That is why the bash scripts have their own wait loop that allows the object to not exist for a while.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -86,27 +86,29 @@ echo "wds1 created."
...
+kubectl -n wds1-system get pods $(kubectl -n wds1-system get pods -l name=transport-controller -o jsonpath='{.items[*].metadata.name}')
```
MikeSpreitzer, line 109: What does this command accomplish?

test/e2e/common/setup-kubestellar.sh
```diff
@@ -138,7 +140,7 @@ kubectl --context imbs1 label managedcluster cluster2 location-group=edge name=c
 : Expect to see the wds1 kubestellar-controller-manager created in the wds1-system 
 : namespace and the imbs1 statefulset created in the imbs1-system namespace.
```
MikeSpreitzer, line 141: This list should be expanded to include the transport controller.

docs/content/direct/examples.md
nirrozenbaum: @MikeSpreitzer do you prefer we invoke it using curl with url of the script?
doing that is a self reference isn’t it?

do you have a suggestion?

please keep in mind that this is (again) temp state until @francostellari completes the helm chart (which can be done only AFTER this PR is merged)

test/e2e/common/setup-kubestellar.sh
nirrozenbaum: go mod tidy is done internally in “make ko-local-build” command.

deploy/transport-controller-deployment.yaml.template
MikeSpreitzer: Having the init in two containers does not bother me.

docs/content/direct/examples.md
MikeSpreitzer: The design here requires the user to get a copy of the repo, using either `git clone` or fetch an archive and unpack. Since it is required, it should be clearly said.

Unfortunately, I think this **does** introduce another self-reference --- if we follow our usual style. Many repos avoid literal self references here by instead telling the user to pick a release and then describing how to use the chosen release without literally including the release identifier in those instructions. We have been congratulating ourselves on having instructions that can be sped through with just a few copy-and-pasted lines, which is not quite consistent with that. I am not opposed to switching to that other style, but I think it should be done with a larger consensus than just two people. I know @pdettori also had short copy-able instructions for KSLight. I know that @clubanderson has opinions.

test/e2e/common/setup-kubestellar.sh
MikeSpreitzer: I do not see that in https://github.com/nirrozenbaum/kubestellar/blob/kflex-config/Makefile

nirrozenbaum: see here:
https://github.com/kubestellar/ocm-transport-plugin/blob/main/Makefile#L39

MikeSpreitzer: Oh, sorry, I was looking at the wrong repo's Makefile.

MikeSpreitzer: Perhaps the `go mod tidy` needs to precede the `go vet`.

docs/content/direct/examples.md
pdettori: so far we have tried to avoid requiring users to clone the repo. It adds some burden to users that just want to quickly try to run the examples. I understand that the helm chart would come later, but I am wondering if in the meantime we could do something like curl the raw script from GitHub and pipe it into sh. If this is not acceptable for some folks or too complex at this stage I am not opposed to move ahead with the current approach but with the understanding that we after this release is out we should really focus on improving UX, that is, publish a helm chart and use that (perhaps in addition to pch) as the only method of installation for a release.

nirrozenbaum: and it keeps the pattern of a copy paste simple step

MikeSpreitzer: I can live with curl-to-bash too.
If this PR introduces another self-reference then it should include an update to release.md.

nirrozenbaum: ok. I actually understand now that's a bit problematic.
this file doesn't exist yet in git (only in my fork) so I can't put curl to to bash pointing to this file.
but ideally I would like to do that.
@MikeSpreitzer is it ok to change this in a followup PR?

nirrozenbaum: after we merge this PR, the file will exist..

test/e2e/common/setup-kubestellar.sh
nirrozenbaum: I will switch the order in ocm plugin. but that will take place only when we use the next ocm plugin rc

test/e2e/common/setup-kubestellar.sh
nirrozenbaum: no need. removed it

test/e2e/common/setup-kubestellar.sh
nirrozenbaum: fixed

test/e2e/common/setup-kubestellar.sh
nirrozenbaum: fixed

hack/deploy-transport-controller.sh
nirrozenbaum: I pushed a fix. 
apparently `envsubst` supports the notion of specifying a list of env vars to replace and it leaves all the others. 
so I just added the 5 env vars that are set in this script.
let me know if it makes sense.

test/e2e/common/setup-kubestellar.sh
MikeSpreitzer: I see that one of the two things I raised got fixed. I still see the `*`.

test/e2e/common/setup-kubestellar.sh
MikeSpreitzer: I see the change in https://github.com/kubestellar/ocm-transport-plugin/blob/main/Makefile . Until kubestellar can use it, I think we need the `go mod tidy` to remain here.

hack/deploy-transport-controller.sh
MikeSpreitzer: That is good enough.

docs/content/direct/examples.md
MikeSpreitzer: I think you could make all the changes in this PR. As you say, once this PR merges, a file that it introduces will exist. We do not test examples.md in CI.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -86,27 +86,28 @@ echo "wds1 created."
...
 pwd
 echo "replace github.com/kubestellar/kubestellar => ${KUBESTELLAR_DIR}/" >> go.mod
-go mod tidy
-make build
-mv ./bin/ocm-transport-plugin ${KUBESTELLAR_DIR}/ocm-transport-plugin
+IMAGE_TAG=${OCM_TRANSPORT_PLUGIN_RELEASE} make ko-build-local
+kind load --name kubeflex docker-image ko.local/transport-controller:${OCM_TRANSPORT_PLUGIN_RELEASE} # load local image to kubeflex
 cd "${KUBESTELLAR_DIR}"
 pwd
 rm -rf ${OCM_TRANSPORT_PLUGIN_DIR}
```
MikeSpreitzer, line 103: FYI: this is really painful when something goes wrong. How about moving this `rm` to the end of each test, where it is only done on success?

ezrasilvera:
In general I'm OK with going forward in this direction as we want to cut a release.  However I find this whole approach too complex and I think it can be done in a simpler way.
Why can't we simply pushed a deployment shell script into the image, that script will first retrieve the configs from KF and generates the needed files on the container FS (with similar commands as this PR is doing in the init container), and then it will launch the controller deployment. 
The only parameter that is needed is the kubeconfig to access KF, which can be easily passed. At this point I would also pass a fixed `in-cluster=true`   , as for now we are officially supporting a single hub cluster for all our control components.

nirrozenbaum:
> @nirrozenbaum In general I'm OK with going forward in this direction as we want to have a release ASAP. However I find this whole approach too complex and I think it can be done in a much simpler way. Why can't we simply pushed a deployment shell script into the image, that script will first retrieve the configs from KF and generates the needed files on the container FS (with similar commands as this PR is doing in the init container), and then it will launch the controller deployment. This way you'll be able to avoid the init containers, CM to store the secrets, and the k8s volume mounts for the files. The only parameter that is needed is the kubeconfig to access KF, which can be easily passed. At this point I would also pass a fixed `in-cluster=true` , as for now we are officially supporting a single hub cluster for all our control components.

This suggestion is doing a mix between transport and KubeFlex. this is exact what I wanted to avoid.
Today there is tight coupling between KF and binding controller. we have an open issue to decouple them. the right approach which all team agree on is separating space manager from the controllers (and images).

the solution I came up with keeps the clean separation of transport and KF. other the the deployment itself, transport is completely decoupled from KF. for example, for testing transport locally, I created a kind cluster and ran transport without KF.

your suggestion goes in the opposite direction and involves adding KF specifics script into transport image, that’s bad.

ezrasilvera:
> > @nirrozenbaum In general I'm OK with going forward in this direction as we want to have a release ASAP. However I find this whole approach too complex and I think it can be done in a much simpler way. Why can't we simply pushed a deployment shell script into the image, that script will first retrieve the configs from KF and generates the needed files on the container FS (with similar commands as this PR is doing in the init container), and then it will launch the controller deployment. This way you'll be able to avoid the init containers, CM to store the secrets, and the k8s volume mounts for the files. The only parameter that is needed is the kubeconfig to access KF, which can be easily passed. At this point I would also pass a fixed `in-cluster=true` , as for now we are officially supporting a single hub cluster for all our control components.
> 
> This suggestion is doing a mix between transport and KubeFlex. this is exact what I wanted to avoid. Today there is tight coupling between KF and binding controller. we have an open issue to decouple them. the right approach which all team agree on is separating space manager from the controllers (and images).
> 
> the solution I came up with keeps the clean separation of transport and KF. other the the deployment itself, transport is completely decoupled from KF. for example, for testing transport locally, I created a kind cluster and ran transport without KF.
> 
> your suggestion goes in the opposite direction and involves adding KF specifics script into transport image, that’s bad.

I don't agree. You have the same coupling. The fact that your dependency is in a script outside the image doesn't change much.  And once you go to Helm it will be part of the helm as well. I don't see much difference that worth this complexity. 
And we agreed that KF as space manger is something we decided to support in any case due to simpler user  experience. We do want to add direct config parameters but also allow KF usage in our controllers. (for people who prefer that).

nirrozenbaum:
> > > @nirrozenbaum In general I'm OK with going forward in this direction as we want to have a release ASAP. However I find this whole approach too complex and I think it can be done in a much simpler way. Why can't we simply pushed a deployment shell script into the image, that script will first retrieve the configs from KF and generates the needed files on the container FS (with similar commands as this PR is doing in the init container), and then it will launch the controller deployment. This way you'll be able to avoid the init containers, CM to store the secrets, and the k8s volume mounts for the files. The only parameter that is needed is the kubeconfig to access KF, which can be easily passed. At this point I would also pass a fixed `in-cluster=true` , as for now we are officially supporting a single hub cluster for all our control components.
> > 
> > 
> > This suggestion is doing a mix between transport and KubeFlex. this is exact what I wanted to avoid. Today there is tight coupling between KF and binding controller. we have an open issue to decouple them. the right approach which all team agree on is separating space manager from the controllers (and images).
> > the solution I came up with keeps the clean separation of transport and KF. other the the deployment itself, transport is completely decoupled from KF. for example, for testing transport locally, I created a kind cluster and ran transport without KF.
> > your suggestion goes in the opposite direction and involves adding KF specifics script into transport image, that’s bad.
> 
> I don't agree. You have the same coupling. The fact that your dependency is in a script outside the image doesn't change much. And once you go to Helm it will be part of the helm as well. I don't see much difference that worth this complexity. And we agreed that KF as space manger is something we decided to support in any case due to simpler user experience. We do want to add direct config parameters but also allow KF usage in our controllers. (for people who prefer that).

this solution supports using kubeFlex 🙂.
this is what we do in e2e. 
and user experience is very simple - run a single script. 
what’s created internally is implementation details and not related to user experience.

I think the fact that it’s outside the image changes a LOT. that’s exactly coupling vs decoupling. the fact I can run transport image without anything that’s related to KF is complete decoupling, which is exactly what we need.

pushing KF specifics into transport image is in fact creating coupling between transport and KF, there is no way around it.

regarding helm, if that would be written correctly, we shouldn’t have any coupling between transport and KF there as well. exactly for this we have code reviews to guarantee that will happen.

ezrasilvera:
> > > > @nirrozenbaum In general I'm OK with going forward in this direction as we want to have a release ASAP. However I find this whole approach too complex and I think it can be done in a much simpler way. Why can't we simply pushed a deployment shell script into the image, that script will first retrieve the configs from KF and generates the needed files on the container FS (with similar commands as this PR is doing in the init container), and then it will launch the controller deployment. This way you'll be able to avoid the init containers, CM to store the secrets, and the k8s volume mounts for the files. The only parameter that is needed is the kubeconfig to access KF, which can be easily passed. At this point I would also pass a fixed `in-cluster=true` , as for now we are officially supporting a single hub cluster for all our control components.
> > > 
> > > 
> > > This suggestion is doing a mix between transport and KubeFlex. this is exact what I wanted to avoid. Today there is tight coupling between KF and binding controller. we have an open issue to decouple them. the right approach which all team agree on is separating space manager from the controllers (and images).
> > > the solution I came up with keeps the clean separation of transport and KF. other the the deployment itself, transport is completely decoupled from KF. for example, for testing transport locally, I created a kind cluster and ran transport without KF.
> > > your suggestion goes in the opposite direction and involves adding KF specifics script into transport image, that’s bad.
> > 
> > 
> > I don't agree. You have the same coupling. The fact that your dependency is in a script outside the image doesn't change much. And once you go to Helm it will be part of the helm as well. I don't see much difference that worth this complexity. And we agreed that KF as space manger is something we decided to support in any case due to simpler user experience. We do want to add direct config parameters but also allow KF usage in our controllers. (for people who prefer that).
> 
> this solution supports using kubeFlex 🙂. this is what we do in e2e. and user experience is very simple - run a single script. what’s created internally is implementation details and not related to user experience.
> 
> I think the fact that it’s outside the image changes a LOT. that’s exactly coupling vs decoupling. the fact I can run transport image without anything that’s related to KF is complete decoupling, which is exactly what we need.
> 
> pushing KF specifics into transport image is in fact creating coupling between transport and KF, there is no way around it.
> 
> regarding helm, if that would be written correctly, we shouldn’t have any coupling between transport and KF there as well. exactly for this we have code reviews to guarantee that will happen.

Please note that there is no "KF dependency" on what I suggested.  We simply say that the script that deploy this controller expect to find the secrets in a certain place. You can go ahead and put them yourself without KF. So it doesn't matter if the script that fetch those secrets is inside or outside the image. As a matter of fact you could even push that logic into the controller itself and publish that you controller can wither get configs on the command line, or secrets in predefine name/ns, or secretRef object name, or context names, or .....
It's completely the same and it's only semantics.  
Also your image can have multiple entry point and your controller can start with kubeconfigs if supplied,  with secrets that stored in predefined locations (if preferred) , or with interacting with KF.  None of this makes the image " dependent on KF" .  But again, this is a semantic discussion and if you supply script+image you should treat the two as  a bundle that expects the input info to be coming from secrets at specific location, I'm just suggesting a much simpler and cleaner way to do it.

docs/content/direct/examples.md
```diff
@@ -60,20 +60,14 @@ manager which connects to the `wds1` front-end and the `imbs1` OCM control plane
...
    ```shell
-   OS=$(go env GOOS)
-   ARCH=$(go env GOARCH)
-   OCM_TRANSPORT_PLUGIN_RELEASE="0.1.0-rc2"
-   OCM_TRANSPORT_PLUGIN_TAR=ocm-transport-plugin_${OCM_TRANSPORT_PLUGIN_RELEASE}_${OS}_${ARCH}.tar.gz
-   wget https://github.com/kubestellar/ocm-transport-plugin/releases/download/v${OCM_TRANSPORT_PLUGIN_RELEASE}/${OCM_TRANSPORT_PLUGIN_TAR}
-   tar -xvf ${OCM_TRANSPORT_PLUGIN_TAR} transport-controller
-   rm -fr bin ${OCM_TRANSPORT_PLUGIN_TAR}
-   ./transport-controller --transport-context imbs1 --wds-context wds1 --wds-name wds1 &> transport.log &
+   bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/main/hack/deploy-transport-controller.sh) wds1 imbs1 ghcr.io/kubestellar/ocm-transport-plugin/transport-controller:0.1.0-rc4
```
MikeSpreitzer, line 70: Instead of `main` this should reference the next release, which will be `0.21.0-rc1`. You are introducing another self-reference here.

MikeSpreitzer:
@ezrasilvera: I imagine that this will not be the only way to run the transport controller in a Pod. We will want another way that does not rely on KubeFlex.

docs/content/direct/examples.md
MikeSpreitzer: The main problem in (B) is that curl will fetch only one file, deploy-transport-controller.sh; the user does not have a local copy of the file referenced from deploy-transport-controller.sh.

MikeSpreitzer: The thing that seems odd to me about requiring a user of a particular release of ks/ks to specify the release of ks/OTP to use is that there is already an answer to "which OCP release?" given if we are following the staggered release approach to coping with the dependency cycle (the synchronized release approach also gives an answer).

MikeSpreitzer: I was not thinking of _adding_ another self-reference, I was thinking of _moving_ the one new one from examples.md to deploy-transport-controller.sh.

nirrozenbaum: > The main problem in (B) is that curl will fetch only one file, deploy-transport-controller.sh; the user does not have a local copy of the file referenced from deploy-transport-controller.sh.

I get your point. that's not so trivial to do the same logic in a single file.
I fixed the other comment but haven't pushed it yet cause I'm thinking about a simple way to answer your comment.

MikeSpreitzer: I am not sure I understand the difficulty. I would just make deploy-transport-controller.sh do `kubectl apply -f -<<EOF` and line the template.

nirrozenbaum: the difficulty is the config map which uses `"$1`, `$2`, etc.. if that's is done in the same file, the bash replaces these placeholders with the first arg of the script (WDS_NAME, IMBS_NAME) and let empty the cpname, etc.. 
that breaks the config map behavior.

in the deployment itself there is no problem.
I'm trying to work around the configmap.

nirrozenbaum: try to this locally and see what happens:
```
cat <<EOF
apiVersion: v1
kind: ConfigMap
metadata:
  name: transport-controller-config
data:
  get-kubeconfig.sh: |
    #!/bin/bash
    # this script receives a ControlPlane name and a parameter
    # that determines whether to extract the ControlPlane's in-cluster kubeconfig
    # or the external kubeconfig (if set to "true", the first will be retrieved).
    # The function returns the requested kubeconfig's content in base64.
    # it assumes the kubeconfig context is set to access the hosting cluster.

    cpname="$1"
    get_incluster_key="$2"

    key=""
    if [[ "$get_incluster_key" == "true" ]]; then
      key=$(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.inClusterKey}');
    else
      key=$(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.key}');
    fi

    # get secret details
    secret_name=$(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.name}')
    secret_namespace=$(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.namespace}')

    # get the kubeconfig in base64
    kubectl get secret $secret_name -n $secret_namespace -o=jsonpath="{.data.$key}"
EOF
```

nirrozenbaum: fixed with escaping.
at first I confused the escaping character `/` with `\` so it didn't work. 
I guess I was too tired and had to sleep in order to get this fixed :).

MikeSpreitzer (commented):
/lgtm
Thanks!

PR #1869, opened 2024-03-11
:bug: fix race-conditions over Binding controller field maps

### LINKED ISSUE #1860: bug: binding controller has race-conditions over member maps
### Describe the bug

The binding controller has 3 maps:
- listers
- informers
- stoppers

These maps are concurrently written to when handling CRDs.

### Steps To Reproduce

Read code

### Expected Behavior

Maps should be RWLock protected

### Additional Context

_No response_

vMaroon:
## Summary
This PR fixes race conditions over controller field maps that may occur due to concurrent read/writes across goroutines.

The maps are:
```
listers               map[schema.GroupVersionResource]cache.GenericLister
informers             map[schema.GroupVersionResource]cache.SharedIndexInformer
stoppers              map[schema.GroupVersionResource]chan struct{}
```

When a CRD is being handled by a worker, the following call may be made:
```
if len(toStartList) > 0 {
	go c.startInformersForNewAPIResources(ctx, toStartList)
}
```

Where `c.startInformersForNewAPIResources` contains writes to these maps. 
Which means that there are 2 levels of concurrency here: different workers may handle different CRD objects, and each handling may contain executions of the above inner block.

Therefore, this PR introduces a new thread-safe interface `ConcurrentMap` which is implemented with a RWLock protected map.

### Notes 
- This change alone is sufficient since the now proper use of the controller/workqueue pattern makes sure that no two goroutines will work on the same keys in these maps.
- While these race conditions were not observed in practice (due to the lack of high-concurrency tests over CRD objects), they are apparent in the code.

## Related issue(s)

Fixes #1860

### DIFF
--- a/pkg/util/concurrent-map.go
+++ b/pkg/util/concurrent-map.go
@@ -1,8 +1,25 @@
+/*
+Copyright 2024 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package util
 
 import (
-	"k8s.io/apimachinery/pkg/util/sets"
 	"sync"
+
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // ConcurrentMap is a thread-safe map.
--- a/pkg/util/concurrent-map.go
+++ b/pkg/util/concurrent-map.go
@@ -84,7 +84,6 @@ func (mm *rwMutexMap[K, V]) Get(key K) (V, bool) {
 
 // Iterator iterates over the map and calls the given function for each
 // key/value pair sequentially.
-// If the given function returns false, the iteration is stopped.
 // If the given function returns an error, the iteration is stopped and
 // the error is returned.
 // During the iteration, the map must not be mutated by the given function.
--- a/pkg/util/concurrent-map.go
+++ b/pkg/util/concurrent-map.go
@@ -18,8 +18,6 @@ package util
 
 import (
 	"sync"
-
-	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // ConcurrentMap is a thread-safe map.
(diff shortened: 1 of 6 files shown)


pkg/util/concurrent-map.go
```diff
@@ -0,0 +1,96 @@
...
+	Set(key K, value V)
+	// Remove removes the value for the given key.
+	Remove(key K)
+	// Get gets the value for the given key.
+	// The second return value is true if the key exists in the map, otherwise false.
+	// Getting a key does not guarantee that no other goroutine will also work with it.
+	Get(key K) (V, bool)
+	// Keys returns a copy of all keys in the map at the time of the call.
+	// Nothing guarantees that when used, these keys are still in the map.
+	Keys() sets.Set[K]
```
MikeSpreitzer, line 37: This is more expensive than this PR needs. As far as I can tell, this PR only needs an iterator method that holds the read lock --- the body of the iteration does not modify the map. I recommend introducing an interator method that anticpates https://go.dev/wiki/RangefuncExperiment .

vMaroon: @MikeSpreitzer which compute expense is the dealbreaker for you? Having to work with a copy of the keys (the cloning) or the loss of performance optimizations that we get from using `range` on a map directly?

If it is the first, and your suggestion in private (giving back a copy of the map for the user to work with) is good enough, I prefer going this route for now.

Because other than this, making the user work with a disconnected copy of the map, is sort of equivalent to locking the map in order to iterate and blocking writes throughout the latter. What do you think?

MikeSpreitzer: The expense that concerns me is making a copy of the keys. The current version of this PR _is_ doing a `range` on a map (different map, but I suppose the cost of the `range` is not much different).

Privately I suggested a few different things. Many of them leave the user iterating over a set of keys, if not also values, that is disconnected from the evolution of the ConcurrentMap. This troubles me, as it is different from the semantics of a primitive `map` in Go. And it is different in a way that is particularly easy to overlook. Nothing shows it syntactically, and testing only uncovers it if the test is very stressful of concurent behavior.

I think that one reasonable approach would be to simply disallow mutation of the ConcurrentMap while an iteration is going on. That is simple and efficient. It is hard to miss during testing --- if the body of the iteration tries a mutation then something will immediately go wrong. This approach is good enough for this PR.

I would prefer to see something that anticipates https://go.dev/wiki/RangefuncExperiment . That can be done while disallowing mutation of the ConcurrentMap while an iteration is going on. I think a viable alternative is to encapsulate the pattern that I see repeated in this PR into the iteration function. That is, have the iteration function start by making a copy of the keys while holding the lock, and prepare for each call on `yield` by doing a `Get` and skipping the key if it is no longer in the map. Again, I like this less because it is subtly different from the behavior for a built-in `map`, but at least it would anticipate the new Go feature and be more DRY.

MikeSpreitzer: I definitely sympathize with the concern for the error cases! In fact, that was my focus when I did this in the legacy system; see https://github.com/kubestellar/kubestellar/blob/v0.14.2/pkg/placement/set-types.go#L45 (and https://github.com/kubestellar/kubestellar/blob/v0.14.2/pkg/placement/map-types.go#L35 shows why you only need to look at the first place).

I presume the reasoning behind the choice made in https://go.dev/wiki/RangefuncExperiment is that it is a better match for the semantics of Go for loops. Whereas Python has `try ... finally ...`, Go has no built-in consideration for errors. A Go for loop only has `continue` and `break`, so allowing `yield` to only return `true` or `false` aligns with the functionality of Go `for ... range ...`.

vMaroon: @MikeSpreitzer thanks, do you have any further requests on the implementation? I'm not about what you meant with `why you only need to look at the first place`.

MikeSpreitzer: By "why you only need to look at the first place" I was thinking you would ask "why did Mike tell me to look at an interface for Set rather than Map?" The answer is that both are specializations of a more basic interface that was called Visitable.

I recommend that we follow the lead of the rangefunc experiment and make `yield` return only a boolean. As you noted and they noted (https://go.dev/wiki/RangefuncExperiment#how-are-more-complicated-loops-implemented), it is _sufficient_ to have `yield` return only a boolean. If you really want `yield` to be able to return an `error` then I suggest you consider the simpler approach of having `yield` return _only_ an error. In the legacy KubeStellar I found that quite workable.

vMaroon: @MikeSpreitzer thanks for clarifying, I went for the latter.

MikeSpreitzer (commented):
/lgtm
But I think we do not want to merge until after the 0.21.0 release.

MikeSpreitzer (approved):
No reason to hold this any longer.
/approve
Thanks!

PR #1870, opened 2024-03-11
:bug: undo the assumption of individual delta events in binding controller

### LINKED ISSUE #1841: bug: Assumption of individual deltas in binding controller
### Describe the bug

https://github.com/kubestellar/kubestellar/blob/f11571b28b512f451faa47e4ac41a0fc1e9a67f7/pkg/binding/controller.go#L398 assumes that each individual change (whatever would that even be?) gets its own notification. This is not true. One update notification can have multiple differences between the old and new objects.

### Steps To Reproduce

This is from code inspection.

### Expected Behavior

Controllers cope with the fact that notifications are only eventually consistent.

### Additional Context

_No response_

vMaroon:
## Summary
In `shouldSkipUpdate` which is called in informers update events handling, the code was assuming that individual deltas get individual events. Then based on this assumption, a binding-policy would get skipped if the update involved increasing the size of the object finalizers list.

This can cause the skipping of the handling of binding-policies that satisfy the latter, while containing other changes.

## Related issue(s)
This issue was found and opened by @MikeSpreitzer 

Fixes #1841

### DIFF
--- a/pkg/binding/controller.go
+++ b/pkg/binding/controller.go
@@ -388,10 +388,7 @@ func shouldSkipUpdate(old, new interface{}) bool {
 	if newMObj.GetResourceVersion() == oldMObj.GetResourceVersion() {
 		return true
 	}
-	// avoid enqueing events when adding finalizers to bindingpolicy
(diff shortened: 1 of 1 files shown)


nirrozenbaum:
looks good to me.

Issue #1882, opened 2024-03-11
bug: add-helm-code.sh does not abort when `yq` is not found


MikeSpreitzer:
### Describe the bug

@francostellari reported that when he did not have `yq` on his `$PATH` and invoked `test/e2e/multi-cluster-deployment/run-test.sh`, the script blundered on.

```
cp -R chart local-chart
cd config/manager && /home/vagrant/kubestellar/bin/kustomize edit set image controller=ghcr.io/kubestellar/kubestellar/controller-manager:fd72570c8
/home/vagrant/kubestellar/bin/kustomize build config/default > local-chart/templates/controller-manager.yaml
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
scripts/add-helm-code.sh --dir /home/vagrant/kubestellar/local-chart add
scripts/add-helm-code.sh: line 77: yq: command not found
scripts/add-helm-code.sh: line 78: yq: command not found
scripts/add-helm-code.sh: line 77: yq: command not found
scripts/add-helm-code.sh: line 78: yq: command not found
scripts/add-helm-code.sh: line 77: yq: command not found
scripts/add-helm-code.sh: line 78: yq: command not found
scripts/add-helm-code.sh: line 77: yq: command not found
scripts/add-helm-code.sh: line 78: yq: command not found
scripts/add-helm-code.sh: line 77: yq: command not found
scripts/add-helm-code.sh: line 78: yq: command not found
scripts/add-helm-code.sh: line 77: yq: command not found
scripts/add-helm-code.sh: line 78: yq: command not found
scripts/add-helm-code.sh: line 77: yq: command not found
scripts/add-helm-code.sh: line 78: yq: command not found
git checkout -- config/manager/kustomization.yaml
```

### Steps To Reproduce

You can see the problem in the source. There is no `set -e` or anything like that.

### Expected Behavior

A missing prerequisite causes an immediate and obvious failure.

MikeSpreitzer:
@pdettori

PR #1883, opened 2024-03-11
🐛 Add pre req check in some e2e scripts

francostellari:
## Summary

Add pre req check in some e2e scripts.

## Related issue(s)

Fixes #
#1882

### DIFF
--- a/hack/run-integration-tests.sh
+++ b/hack/run-integration-tests.sh
@@ -24,39 +24,41 @@
 set -x
 set -e
 
-etcd_version=3.5.12
-platform="$(go env GOOS)-$(go env GOARCH)"
+if ! which etcd; then
+    etcd_version=3.5.12
+    platform="$(go env GOOS)-$(go env GOARCH)"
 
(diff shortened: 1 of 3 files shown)


nirrozenbaum:
/hold

This PR changes files that transport containerization PR changes. 
the transport PR is highest priority, please merge this one only after transport is merged to avoid having to rebase and cause additional delays.

scripts/add-helm-code.sh
```diff
@@ -72,6 +72,7 @@ if [[ $CMD == "add" ]]; then
   done < "$op_file"
 
   # Loop over all yaml files in the directory
+  "${HOME_DIR}/hack/check_pre_req.sh" --assert yq
```
MikeSpreitzer, line 75: Since there is no `set -e` in this script, a failed prereq check here will not stop the script from blundering on. Don't we want `set -e` in this script?

francostellari: ok, done

PR #1884, opened 2024-03-12
:bug: fix Binding cluster selection implementation

### LINKED ISSUE #1655: feature: Add more automated tests to CI/CD
### Feature Description

Currently we have two automated tests (multi-cluster-deployment and singleton-status) that test core features of KubeStellar. This is a good start but we need to test also other usage scenarios. 

### Proposed Solution

We can think of other tests as variations built as extensions of multi-cluster-deployment. These tests are designed to exercise different parts of the code:

0. Test update of an object - deliver a deployment, increase the number of replicas from 1 to 2, verify they are updated on the WEC.

1. Create object 1 with label A and label B, object 2 with label B, and a Placement that matches labels A AND B, and verify that only object 1 matches and is delivered.

2. Create object 1 with label A, object 2 with label B, and a Placement with 2 object selectors, one with label A and one with label B, and verify both objects get delivered.

3. Similar to 2, but test also specifying other criteria for selection of objects: namespace, api group, resource, name

4. Start from `multi-cluster-deployment` scenario, delete the placement and verify that the object is deleted from the WECs.

5. Start from `multi-cluster-deployment` scenario, change the placement objectSelector to no longer match and verify that the object is deleted from the WECs.

6. Start from  `multi-cluster-deployment` scenario, delete the object and verify it is deleted from the WECs.

7. Create an object and two placements that match the object (overlapping placements). Verify that by deleting one of the placements the object stays in the WEC, but when the second placement is deleted the object is deleted from the WEC.

8. Create an object and a placement that does not match,  change the placement to match, then verify that the object gets delivered.

9. Write a test for https://github.com/kubestellar/kubestellar/issues/1541 (which has been fixed)

### Want to contribute?

### Additional Context

_No response_

### LINKED ISSUE #1699: bug: convertLabelSelectorsToMap is broken two ways
### Describe the bug

`pkg/ocm/ocm.go:convertLabelSelectorsToMap` is broken in three ways.

1. The return type can not represent the OR functionality that is the semantics in https://github.com/kubestellar/kubestellar/blob/ad2f86d60cf5b27a88a16839d419d152b346ff5a/api/edge/v1alpha1/types.go#L26-L27
2. https://github.com/kubestellar/kubestellar/blob/ad2f86d60cf5b27a88a16839d419d152b346ff5a/pkg/ocm/ocm.go#L190 will fail for a selector that is not simply an AND of key=value pairs. There are indeed more complex label selectors. See https://v1-28.docs.kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors .

### Steps To Reproduce

This bug report comes from code inspection.

### Expected Behavior

Any valid user input can be handled.

### Additional Context

_No response_

vMaroon:
## Summary
This PR fixes the implementation of ocm::FindClustersBySelectors to support OR in BindingPolicy spec's `clusterSelectors`, as well as support complex label selectors such as [set-based selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement).

## Related issue(s)

Fixes #1699
Fixes #1655

### DIFF
--- a/test/e2e/multi-cluster-deployment/use-kubestellar.sh
+++ b/test/e2e/multi-cluster-deployment/use-kubestellar.sh
@@ -28,7 +28,9 @@ metadata:
   name: nginx-bindingpolicy
 spec:
   clusterSelectors:
-  - matchLabels: {"location-group":"edge"}
+  - matchLabels: {"name":"cluster1"}
+  - matchExpressions:
+    - { key: name, operator: In, values: [cluster2, cluster3] }
   downsync:
   - objectSelectors:
     - matchLabels: {"app.kubernetes.io/name":"nginx"}
--- a/test/e2e/multi-cluster-deployment/use-kubestellar.sh
+++ b/test/e2e/multi-cluster-deployment/use-kubestellar.sh
@@ -30,7 +30,7 @@ spec:
   clusterSelectors:
   - matchLabels: {"name":"cluster1"}
   - matchExpressions:
-    - { key: name, operator: In, values: [cluster2, cluster3] }
+    - { key: name, operator: In, values: [cluster2] }
   downsync:
   - objectSelectors:
     - matchLabels: {"app.kubernetes.io/name":"nginx"}
--- a/test/e2e/multi-cluster-deployment/use-kubestellar.sh
+++ b/test/e2e/multi-cluster-deployment/use-kubestellar.sh
@@ -28,9 +28,7 @@ metadata:
   name: nginx-bindingpolicy
 spec:
   clusterSelectors:
-  - matchLabels: {"name":"cluster1"}
-  - matchExpressions:
-    - { key: name, operator: In, values: [cluster2] }
+  - matchLabels: {"location-group":"edge"}
   downsync:
   - objectSelectors:
     - matchLabels: {"app.kubernetes.io/name":"nginx"}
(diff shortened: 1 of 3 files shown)


test/e2e/multi-cluster-deployment/use-kubestellar.sh
```diff
@@ -28,7 +28,9 @@ metadata:
   name: nginx-bindingpolicy
 spec:
   clusterSelectors:
-  - matchLabels: {"location-group":"edge"}
+  - matchLabels: {"name":"cluster1"}
+  - matchExpressions:
+    - { key: name, operator: In, values: [cluster2, cluster3] }
```
nirrozenbaum, line 33: I understand the reason for changing this file (you want to show an example how to use it and also verify this is tested), but not sure I would do this change.
few reasons:
- there is no `cluster3` in the e2e test.
- this change is testing the introduced change, but removes the test that matchLabels can select more than one cluster.
- it might be better to put this example in a README file. we can add a section that shows examples for how to select clusters.

vMaroon: I agree on cluster3, removed while updating on slack.

As for the 2nd point, what do you think about testing both here? As in:
- Revert the test to how it was
- After success: change the BP to select no clusters
- Wait until update is reflected in cluster
- Change the BP to select the two clusters with the selectors I wrote
- Test

vMaroon:
@MikeSpreitzer if you prefer a lister is used instead of client calls already in this PR, pls let me know. Otherwise we can issue it. The intention is for this fix to be a part of the PT release.

effi-ofer:
I translated the test case and added it to PR #1713.

PR #1888, opened 2024-03-12
✨ Improve the Status Controller

### LINKED ISSUE #1862: feature: Clean up the status controller
### Feature Description

We had some discussion about the status controller and all agreed that it can be simplified by doing some cleaning up. Detailed [here](https://github.com/kubestellar/kubestellar/pull/1858#discussion_r1513714628).

### Proposed Solution

Do it.

### Want to contribute?

### Additional Context

_No response_

waltforme:
## Summary
This PR improves the status controller in two aspects.
1. Idiomatic synchronization between the status controller and the binding controller using a channel. So that the status controller doesn't need to check everyone of the per-resource informers for cache sync. Instead it proceeds immediately when the binding controller knows that all caches are synced.
2. Removal of watch to bindingpolicies.

## Related issue(s)
Closes #1862

### DIFF
--- a/pkg/status/controller.go
+++ b/pkg/status/controller.go
@@ -41,27 +41,24 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
-	"github.com/kubestellar/kubestellar/api/control/v1alpha1"
 	"github.com/kubestellar/kubestellar/pkg/util"
 )
 
 const controllerName = "Status"
 
-// Status controller watches workstatues and checks associated bindingpolicies for singleton status. If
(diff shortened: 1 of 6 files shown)


pkg/status/controller.go
```diff
@@ -108,20 +104,18 @@ func NewController(wdsRestConfig *rest.Config, imbsRestConfig *rest.Config,
 		imbsDynClient:  imbsDynClient,
 		imbsKubeClient: imbsKubeClient,
 		workqueue:      workqueue.NewRateLimitingQueue(ratelimiter),
-		listers:        listers,
-		informers:      informers,
```
andreyod, line 112: The parameters `listers` and `informers` are no longer used in this func and should be removed from the signature. Right?

waltforme: Exactly. Removed in latest push.

PR #1889, opened 2024-03-12
📖  prep for release 0.21.0-rc1

nirrozenbaum:
## Summary

prep for release. 
in `examples.md` there are many places to update the same version. this is tedious and error prone.
I've added a step to define the required versions and then use these versions in all other steps.

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/direct/README.md
+++ b/docs/content/direct/README.md
@@ -34,7 +34,8 @@ including OCM Klusterlet for the WECs.
 ## Latest stable release
 
 We do not have one that is proven very good yet.
-The latest release is [0.21.0-rc1](../../../../v0.21.0-rc1).
+The first release using the new architecture is [0.20.0](../../../../v0.20.0); it is feture-incomplete.
+The latest release is [0.21.0-rc1](../../../../v0.21.0-rc1); it is also feature-incomplete.
 See also [the release notes](release-notes.md).
 
 ## Architecture
--- a/docs/content/direct/README.md
+++ b/docs/content/direct/README.md
@@ -34,7 +34,7 @@ including OCM Klusterlet for the WECs.
 ## Latest stable release
 
 We do not have one that is proven very good yet.
-The latest release is [0.20.0](../../../../v0.20.0).
+The latest release is [0.21.0-rc1](../../../../v0.21.0-rc1).
 See also [the release notes](release-notes.md).
 
 ## Architecture
--- a/docs/content/direct/release-notes.md
+++ b/docs/content/direct/release-notes.md
@@ -12,3 +12,7 @@ The following sections list the known issues for each release. The issue list is
    * If a workload object is deleted, or changed to remove some Placements from the matching set, it will not be handled correctly.
    * A Placement update that (a) removes workload objects or clusters from their respective matching sets is not handled correctly.
 
+## Major changes from previous release
+
+* Starting from release v0.21.0-rc1, dynamic changes to WECs **are supported**. Existing placements will be updated when new WECs are added/updated/delete or when labels are added/updated/deleted on existing WECs (not working in previous releases).
+
--- a/docs/content/direct/release-notes.md
+++ b/docs/content/direct/release-notes.md
@@ -2,17 +2,33 @@
 
 The following sections list the known issues for each release. The issue list is not differential (i.e., compared to previous releases) but a full list representing the overall state of the specific release. 
 
-## Every release after 0.15.X
+## 0.21.0-rc1
+
+### Major changes for 0.21.0-rc1
+
+* This release introduces pluggable transport. Currently the only plugin is [the OCM transport plugin](https://github.com/kubestellar/ocm-transport-plugin).
+
+### Bug fixes in 0.21.0-rc1
+
+* dynamic changes to WECs **are supported**. Existing Bindings and ManifestWorks will be updated when new WECs are added/updated/delete or when labels are added/updated/deleted on existing WECs
+* An update to a workload object that removes some BindingPolicies from the matching set _is_ handled correctly.
+* Changes that happen while a controller is down and are handled correctly:
+   * If a workload object is deleted, or changed to remove some BindingPolicies from the matching set;
+   * A BindingPolicy update that removes workload objects or clusters from their respective matching sets.
+
+### Remaining limitations in 0.21.0-rc1
 
 * Removing of WorkStatus objects (on the transport namespace) is not supported and may not result in recreation of that object
-* Singleton: It's the user responsibility to make sure there are no shared objects in two different (singleton) placements that target two different WECs. Currently there is no enforcement on on that. 
+* Singleton status return: It is the user responsibility to make sure that if a BindingPolicy requesting singleton status return matches a given workload object then no other BindingPolicy matches the same object. Currently there is no enforcement of that.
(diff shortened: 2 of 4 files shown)


docs/content/direct/README.md
```diff
@@ -34,7 +34,7 @@ including OCM Klusterlet for the WECs.
 ## Latest stable release
 
 We do not have one that is proven very good yet.
-The latest release is [0.20.0](../../../../v0.20.0).
+The latest release is [0.21.0-rc1](../../../../v0.21.0-rc1).
```
MikeSpreitzer, line 37: Since this is only an RC, maybe also retain mention of 0.20.0?

docs/content/direct/release-notes.md
```diff
@@ -4,7 +4,6 @@ The following sections list the known issues for each release. The issue list is
 
 ## Every release after 0.15.X
 
-* Dynamic changes to WECs are not supported. Existing placements will not be updated when new WECs are added or when labels are added/deleted on existing WECs
```
MikeSpreitzer, line 7: This is still true of 0.20.0, right?

docs/content/direct/README.md
ezrasilvera: Or maybe let's change to :
The latest stable release: 0.20.0
The latest development release candidate is: 0.21.0-rc1

ezrasilvera: @MikeSpreitzer What do you think ?

docs/content/direct/release-notes.md
ezrasilvera: Agree. We can a list of issues per release or just add a "fixed since previous version" section

nirrozenbaum: why not having each release tags specify its issues?
for example, if you switch to tag 0.20, you will see the list relevant for that release.
why do we need to maintain in main list of issues for every release we ever made?

ezrasilvera: I don't mind. I just remember that most "release notes" docs I saw list the history of all previous release. But what ever both you and @MikeSpreitzer  think is fine with me..

docs/content/direct/README.md
MikeSpreitzer: Since 0.20.0 is deliberately feature-incomplete, I would be clear about that.  How about something like the following?

> The first release using the new architecture is [0.20.0](../../../../v0.20.0); it is feture-incomplete.
> The latest release is [0.21.0-rc1](../../../../v0.21.0-rc1); it is also feature-incomplete.

docs/content/direct/release-notes.md
MikeSpreitzer: Perhaps my language was misleading. I did not mean to say that the release notes have to be a series of deltas. I agree that it makes sense for each regular release's description to be self-contained. By "regular" I mean non-RC. I am less sure about how to handle release candidates. I would be OK with a very simple uniform approach, let the description of every release (RC or regular) be self-contained.

That said, I think it would also be helpful to mention the major delta(s) from the previous regular release. Not as the primary description, but as an aid to someone who has some knowledge of previous release(s).

docs/content/direct/README.md
nirrozenbaum: fixed

docs/content/direct/release-notes.md
nirrozenbaum: added section in release notes

MikeSpreitzer: I see that this PR still removes "Dynamic changes to WECs are not supported" from the "Every release after 0.15.X" while introducing a section "Major changes from previous release" that says "dynamic changes to WECs **are supported**" --- which reads awkward since the reader has not previously seen a statement that they are _not_ supported. I also do not see a mention of pluggable transport.

I am surprised, I thought @nirrozenbaum was favoring having the primary content of each release be self contained. I think that would look something like have the following section.

## Release 0.20.0 and its release candidates

(the content currently under "Every release after 0.15.X")

## Release 0.21.0-rc1

* A positive statement about pluggable transport
* A positive statement along the lines of asserting complete coverage of various scenarios with concurrency and updates and deletes of various kinds of objects.
* (the negative content under "Every release after 0.15.X" but NOT including "Dynamic changes to WECs are not supported")

MikeSpreitzer: But in the better order, of course: most recent release first.

MikeSpreitzer: Actually, reviewing the negative statements about 0.20.0, I think more of them no longer apply; they are bugs that have been fixed. Following are these statements.

* An update to a workload object that removes some Placements from the matching set is not handled correctly.
* Some operations are not handled correctly while the controller is down:
   * If a workload object is deleted, or changed to remove some Placements from the matching set, it will not be handled correctly.
   * A Placement update that (a) removes workload objects or clusters from their respective matching sets is not handled correctly.

MikeSpreitzer: And the statement about singleton status return is too narrow (see my review comments on every other copy of that statement).

MikeSpreitzer: So I think something like the following would make sense.

## 0.21.0-rc1

### Major changes for 0.21.0-rc1

* This release introduces pluggable transport. Currently the only plugin is [the OCM transport plugin](https://github.com/kubestellar/ocm-transport-plugin).

### Bug fixes in 0.21.0-rc1

* dynamic changes to WECs **are supported**. Existing Bindings and ManifestWorks will be updated when new WECs are added/updated/delete or when labels are added/updated/deleted on existing WECs
* An update to a workload object that removes some BindingPolicies from the matching set _is_ handled correctly.
* Changes that happen while a controller is down and are handled correctly:
   * If a workload object is deleted, or changed to remove some BindingPolicies from the matching set;
   * A BindingPolicy update that removes workload objects or clusters from their respective matching sets.

### Remaining limitations in 0.21.0-rc1

* Removing of WorkStatus objects (on the transport namespace) is not supported and may not result in recreation of that object
* Singleton status return: It is the user responsibility to make sure that if a BindingPolicy requesting singleton status return matches a given workload object then no other BindingPolicy matches the same object. Currently there is no enforcement of that.
* Objects on two different WDSs shouldn't have the exact same identifier (same group, version, kind, name and namespace). Such a conflict is currently not identified.

## 0.20.0 and its release candidates

* Dynamic changes to WECs are not supported. Existing ManifestWorks will not be updated when new WECs are added or when labels are added/deleted on existing WECs
* Removing of WorkStatus objects (on the transport namespace) is not supported and may not result in recreation of that object
* Singleton status return: It is the user responsibility to make sure that if a BindingPolicy requesting singleton status return matches a given workload object then no other BindingPolicy matches the same object. Currently there is no enforcement of that.
* Objects on two different WDSs shouldn't have the exact same identifier (same group, version, kind, name and namespace). Such a conflict is currently not identified.
* An update to a workload object that removes some BindingPolicies from the matching set is not handled correctly.
* Some operations are not handled correctly while the controller is down:
   * If a workload object is deleted, or changed to remove some BindingPolicies from the matching set, it will not be handled correctly.
   * A BindingPolicy update that removes workload objects or clusters from their respective matching sets is not handled correctly.

nirrozenbaum: fixed. changes as you requested.

PR #1891, opened 2024-03-13
📖 Update some docs

waltforme:
## Summary
This PR updates docs for integration tests, corrects some small mistakes in docs for e2e tests, and fixes some broken links.


test/e2e/object-fields-filtering/run-test.sh
```diff
@@ -13,8 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This is an end to end test of multi cluster deployement.  
-# For readable instructions, please visit ../../../docs/content/v0.20
+# This is an end to end test of object fields filtering.
```
MikeSpreitzer, line 16: We prefer "cleaning" to "fields filtering". Labels and annotations are not fields. "filtering" is currently accurate, but (a) seems a little more precise than necessary, might become inaccurate in the future, and (b) a reader seeing "object filtering" will most likely think we mean removal of entire objects. "object content filtering" would remove objection (b) but not (a). "object cleaning" avoids both.

MikeSpreitzer (commented):
This is good enough to merge. I left one comment suggesting further improvement.

/lgtm
Thanks!

test/e2e/object-fields-filtering/run-test.sh
waltforme: Got the point, will address in another PR.

PR #1898, opened 2024-03-14
🌱 Remove irrelevant hostAliases for transport-controller pod

MikeSpreitzer:
## Summary
This PR simplifies `scripts/deploy-transport-controller.sh` by removing the provision of hostAliases for `wds1.localtest.me` and `imbs1.localtest.me`. These host aliases are not needed. The init containers in the transport-controller Deployment extract the _in-cluster_ kubeconfigs for wds1 and imbs1, and these use Kubernetes Service domain names to reach the relevant apiserver. It is the _external_ kubeconfigs that use `<something>.localtest.me`. But the transport-controller belongs in the kubeflex hosting cluster (as long as we are using kubeflex at all, and those init containers are specific to using kubeflex), so the external kubeconfigs are irrelevant.

## Related issue(s)

Fixes #


MikeSpreitzer:
/cc @pdettori 
/cc @nirrozenbaum 
@francostellari 
@ezrasilvera

nirrozenbaum:
/lgtm
Thanks! 🙏🏼

PR #1899, opened 2024-03-14
📖 Fix doc bugs just found

MikeSpreitzer:
## Summary
This PR fixes various problems I found reviewing 0.21.0-rc2 and trying to run through examples.md.  I did not get past the setup.

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/direct/examples.md
+++ b/docs/content/direct/examples.md
@@ -14,6 +14,10 @@ See [pre-reqs](pre-reqs.md).
 
 The following steps establish an initial state used in the examples below.
 
+1. You may want to `set -x` in your shell so that any failures in the setup or usage scenarios are not lost.
+
+1. If you ran through these scenarios previously then you will need to do a bit of cleanup first. See how it is done in [the cleanup script for our E2E tests](../../../test/e2e/common/cleanup.sh).
+
 1. Set environment variables to hold KubeStellar and OCM-status-addon desired versions:
 
    ```shell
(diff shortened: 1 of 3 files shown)


docs/content/direct/examples.md
```diff
@@ -14,6 +14,15 @@ See [pre-reqs](pre-reqs.md).
...
 
+1. You may want to `set -x` in your shell so that any failures in the setup or usage scenarios are not lost.
+
+1. If you ran through these scenarios previously then you will need to do a bit of cleanup first. Remove your kubeconfig contexts that were left behind and will get in the way of context rename commands to come.
+
+   ```shell
+   kubectl config delete-context cluster1
+   kubectl config delete-context cluster2
+   ```
+
```
nirrozenbaum, line 25: looking in e2e tests cleanup, it seems that it cleans more than these two contexts. it also removed the kind clusters - https://github.com/kubestellar/kubestellar/blob/main/test/e2e/common/cleanup.sh#L24-#L28 
(a) don't we want to clean the clusters as well?
(b) do we want to point to the cleanup script instead of writing the same here?

ezrasilvera: I'm not sure we need to do any cleaning on the examples. For the user it should be enough to say "you should start from a clean environment"  , and maybe suggest to delete his whole environment. 
I agree with @nirrozenbaum - we can also point to our cleanup script as a reference.

docs/content/direct/examples.md
```diff
@@ -70,12 +79,14 @@ manager which connects to the `wds1` front-end and the `imbs1` OCM control plane
 1. Run the OCM based transport controller in a pod.  
 **NOTE**: This is work in progress, in the future the controller will be deployed through a Helm chart.
 
-   Run [transport deployment script](../../../scripts/deploy-transport-controller.sh).  
-   This script expects to get three arguments - (1) wds name; (2) imbs name; and (3) transport controller image.  
+   Run [transport deployment script](../../../scripts/deploy-transport-controller.sh), as follows.
+   This script requires that the user's current kubeconfig context be for the kubeflex hosting cluster.
+   This script expects to get two or three arguments - (1) wds name; (2) imbs name; and (3) transport controller image.
```
ezrasilvera, line 84: Why do we need the lengthy description why not put the command itself + explanation below of the parameters (like a help output of a command)

docs/content/direct/examples.md
```diff
@@ -70,12 +79,14 @@ manager which connects to the `wds1` front-end and the `imbs1` OCM control plane
...
-   This script expects to get three arguments - (1) wds name; (2) imbs name; and (3) transport controller image.  
+   Run [transport deployment script](../../../scripts/deploy-transport-controller.sh), as follows.
+   This script requires that the user's current kubeconfig context be for the kubeflex hosting cluster.
+   This script expects to get two or three arguments - (1) wds name; (2) imbs name; and (3) transport controller image.  
    While the first and second arguments are mandatory, the third one is optional.
    The transport controller image argument can be specified to a specific image, or, if omitted, it defaults to the OCM transport plugin release that preceded the KubeStellar release being used.
-   For example, one can deploy transport controller using the following command:
+   For example, one can deploy transport controller using the following commands:
    ```shell
+   kubectl config use-context kind-kubeflex
```
francostellari, line 89: The above will work for kind deployment but not OCP, where the kubeflex context has a different name.

Perhaps a more generic approach is to use `kflex ctx`

docs/content/direct/examples.md
MikeSpreitzer: Good idea. Done.

docs/content/direct/examples.md
MikeSpreitzer: I think it is better to lead with an explanation of what is going on. People often execute displayed commands as soon as they are seen, so I think it is helpful to give the understanding first.

docs/content/direct/examples.md
MikeSpreitzer: Good point. Done.

waltforme:
Like the to-be-merged branch name. Happy pi day to everyone.

Issue #1838, opened 2024-02-29
bug: singleton status does not change when wantSingletonReportedState changes


effi-ofer:
### Describe the bug

To set the singleton status, the wantSingletonReportedState flag needs to be set to true in the spec of the BindingPolicy.  When this flag changes from true to false, the singleton status does not changes and the status is still reported.

### Steps To Reproduce

Create a singleton deployment and bindpolicy.  This can be done by running the e2e singleton test.  Then patch the wantSingletonReportedState to false as follows:
`kubectl --context wds1 patch bindingpolicy nginx-singleton --type=merge -p '{"spec": {"wantSingletonReportedState": false}}'`

Here's what I do:
```
2216  kubectl --context wds1 create -f bp.yaml
 2217  kubectl --context wds1 create -f d.yaml
 2218  kubectl --context wds1 get deployment -n nginx nginx-singleton -o yaml
 2219  kubectl --context wds1 get bindingpolicies  nginx-singleton -o yaml
 2220  kubectl --context wds1 patch bindingpolicy nginx-singleton --type=merge -p '{"spec": {"wantSingletonReportedState": false}}'
 2221  kubectl --context wds1 get bindingpolicies  nginx-singleton -o yaml
 2222  kubectl --context wds1 get deployment -n nginx nginx-singleton -o yaml
```
Where bp.yaml is
```
apiVersion: control.kubestellar.io/v1alpha1
kind: BindingPolicy
metadata:
  creationTimestamp: "2024-02-29T17:12:47Z"
  finalizers:
  - bindingpolicy.kubestellar.io/kscontroller
  generation: 2
  name: nginx-singleton
  resourceVersion: "5006"
  uid: 2be0aad0-0cc4-4985-a6c7-0cbd6244b2f4
spec:
  clusterSelectors:
  - matchLabels:
      name: cluster1
  downsync:
  - objectSelectors:
    - matchLabels:
        app.kubernetes.io/name: nginx-singleton
  wantSingletonReportedState: true
```
And d.yaml is
```
apiVersion: control.kubestellar.io/v1alpha1
kind: BindingPolicy
metadata:
  creationTimestamp: "2024-02-29T17:12:47Z"
  finalizers:
  - bindingpolicy.kubestellar.io/kscontroller
  generation: 2
  name: nginx-singleton
  resourceVersion: "5006"
  uid: 2be0aad0-0cc4-4985-a6c7-0cbd6244b2f4
spec:
  clusterSelectors:
[...]

Issue #1856, opened 2024-03-05
bug: When a label removed from an object on WDS, it is not removed on a WEC


andreyod:
### Describe the bug

When a label is removed from an object on WDS, transport controller removes that label in the ManifestWork, but it is not removed from the object on the WEC. 
Note that add label and update the label value works as expected.

### Steps To Reproduce

1. Setup KubeStellar, create a deployment and bindpolicy. This can be done by running the e2e multi-cluster-deployment test.
2. Patch the deployment with a label:
    `kubectl --context wds1 label deployments nginx-deployment -n nginx test=test`
3. Verify the label is in the Manifest and on the deployment in the WEC 
    `kubectl --context imbs1 get manifestworks -n cluster1 nginx-bindingpolicy-wds1 -o jsonpath="{.spec.workload.manifests}" | jq`
    `kubectl --context cluster1 get deployments nginx-deployment -n nginx -o jsonpath="{.metadata.labels}" | jq`
4. Remove the label:
    `kubectl --context wds1 label deployments nginx-deployment -n nginx test-`
5. Verify the label was removed from the Manifest:
    `kubectl --context imbs1 get manifestworks -n cluster1 nginx-bindingpolicy-wds1 -o jsonpath="{.spec.workload.manifests}" | jq`
6. Check the label on the WEC
```
kubectl --context cluster1 get deployments nginx-deployment -n nginx -o jsonpath="{.metadata.labels}" | jq
{
  "app.kubernetes.io/name": "nginx",
  "test": "test"
}
```

### Expected Behavior

In step 6 the `test` label should not be in the deployment on the WEC.

andreyod:
Same issue when using OCM hub directly.
Can this be OCM configuration issue? Or maybe OCM bug.

andreyod:
> sounds like a bug rather than a configuration. can you open a bug in ocm? you can tag me once the bug is there and I'll try to push for having a fix (or you can fix it if you want)

In our case OCM is installed by Kubeflex. We need to understand how it's installed and what version before we open a bug in OCM.

nirrozenbaum:
you can also install ocm as a standalone using kind (just follow ocm local standalone) and test it.
this way it's bare ocm, not related at all to kflex.
https://open-cluster-management.io/getting-started/quick-start/

nirrozenbaum:
2 minutes setup

andreyod:
Bug in OCM: https://github.com/open-cluster-management-io/ocm/issues/367

[...]

PR #1900, opened 2024-03-14
:sparkles: Set the singleton status label value to be true or false

andreyod:
## Summary
Today the singleton status label value is set to the `BindingPolicy` name and this value is unused. This PR propose to set it to true/false and then the status controller will remove the status from the workload object when singleton status label value is set to `false` in the `WorkStatus`, and will add the status if it's `true`.

## Related issue(s)
https://github.com/kubestellar/kubestellar/issues/1838
https://github.com/kubestellar/kubestellar/issues/1856

Fixes #
Together with https://github.com/kubestellar/ocm-status-addon/pull/42, should fix https://github.com/kubestellar/kubestellar/issues/1838

### DIFF
--- a/pkg/binding/selectors.go
+++ b/pkg/binding/selectors.go
@@ -30,8 +30,6 @@ import (
 	"github.com/kubestellar/kubestellar/pkg/util"
 )
 
-const emptyBindingPolicyName = ""
-
 // when an object is updated, we iterate over all bindingpolicies and update
 // resolutions that are affected by the update. Every changed resolution leads
 // to queueing its relevant binding for syncing.
@@ -122,9 +120,8 @@ func (c *Controller) updateResolutions(ctx context.Context, objIdentifier util.O
 		}
 
 		// make sure object has singleton status if needed
-		if !objBeingDeleted &&
-			c.bindingPolicyResolver.ResolutionRequiresSingletonReportedState(bindingPolicy.GetName()) {
-			if err := c.handleSingletonLabel(ctx, obj, objGVR, bindingPolicy.GetName()); err != nil {
+		if !objBeingDeleted && c.bindingPolicyResolver.ResolutionRequiresSingletonReportedState(bindingPolicy.GetName()) {
+			if err := c.handleSingletonLabel(ctx, obj, objGVR, util.BindingPolicyLabelSingletonStatusValueSet); err != nil {
 				return fmt.Errorf("failed to add singleton label to object: %w", err)
 			}
 
@@ -137,7 +134,7 @@ func (c *Controller) updateResolutions(ctx context.Context, objIdentifier util.O
 	// NOTE that this takes care of the case where the object was previously selected by a singleton binding
 	// and is no longer selected by any binding.
 	if !objBeingDeleted && !isSelectedBySingletonBinding {
-		if err := c.handleSingletonLabel(ctx, obj, objGVR, emptyBindingPolicyName); err != nil {
+		if err := c.handleSingletonLabel(ctx, obj, objGVR, util.BindingPolicyLabelSingletonStatusValueUnset); err != nil {
 			return fmt.Errorf("failed to remove singleton label from object: %w", err)
 		}
 	}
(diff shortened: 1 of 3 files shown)


vMaroon:
_(Unrelated to reviewing code)_

The title and implementation suggest that this is simply refactoring out an un-used value, but the issues closed by this PR are more complex than that. Can you explain?

Also, is it possible to achieve the same result while still setting the BP name as value? Only because this provides more information that may (but also may not) be useful later.

andreyod:
> _(Unrelated to reviewing code)_
> 
> The title suggests that this is simply refactoring out an un-used value, but the issues closed by this PR are more complex than that. Can you explain whether the setting to true or false is required for the solving these issues or is it the first?
> 
> Depending on that, is it possible to achieve the same result while still setting the BP name as value? Only because this provides more information that may (but also may not) be useful later.

The setting to true or false is required for solving the issue #1838.
Because `WorkStatus` is managed by ocm-status-addon on the WEC we need an indication when we should remove the status from the workload object in WDS.  Because of the OCM bug explained in #1856 we can't remove the label, so the proposed solution is to set the value to false.

vMaroon:
@andreyod is it reasonable to have the policy name act as the true value, and some other set value to indicate that the label was removed? I think it's also essential to explain that there is an external bug that backs this unset value, otherwise the keeping of the label can be questioned when reading the code.

pkg/util/labels.go
```diff
@@ -28,9 +28,12 @@ import (
 )
 
 const (
-	// BindingPolicyLabelSingletonStatusKey is the key for the singleton status reporting
-	// requirement. The value should be the bindingpolicy that is selecting the labeled object.
+	// BindingPolicyLabelSingletonStatusKey is the key for the singleton status reporting requirement.
 	BindingPolicyLabelSingletonStatusKey = "managed-by.kubestellar.io/singletonstatus"
+	// BindingPolicyLabelSingletonStatusValueSet is the value when the status reporting is required.
```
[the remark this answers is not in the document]
andreyod: For the constants vs function: please note that we also use the values in selectors.go , not just verifying it in the status controller.
For the names: I'm fine with BindingPolicySingletonStatusRequired , what should be the name for the second constant?

andreyod:
> @andreyod is it reasonable to have the policy name act as the true value, and some other set value to indicate "removal"? I think it's also essential to explain that there is an external bug that backs this unset value, otherwise the keeping of the label can be questioned when reading the code.

Fine with me, but note that we don't control the policy name(it's given by the user) so what would be the string value to indicate "removal"?

pdettori:
I agree with @vMaroon, can we use the name of the policy as value and some other value to indicate disabling/removal?

MikeSpreitzer:
The syntax and character set for label values is documented in https://go.dev/play/p/c6KI_I4bbOx . Where is the corresponding documentation for BindingPolicy names?

MikeSpreitzer:
FYI, a Kubernetes ObjectMeta.UID is always a valid label value.

pkg/binding/selectors.go
```diff
@@ -137,21 +134,20 @@ func (c *Controller) updateResolutions(ctx context.Context, objIdentifier util.O
 	// NOTE that this takes care of the case where the object was previously selected by a singleton binding
 	// and is no longer selected by any binding.
 	if !objBeingDeleted && !isSelectedBySingletonBinding {
-		if err := c.handleSingletonLabel(ctx, obj, objGVR, emptyBindingPolicyName); err != nil {
+		if err := c.handleSingletonLabel(ctx, obj, objGVR, util.BindingPolicyLabelSingletonStatusValueUnset); err != nil {
 			return fmt.Errorf("failed to remove singleton label from object: %w", err)
```
vMaroon, line 138: This error message should be updated since the label no longer gets removed.

andreyod:
The label value is only used to indicate the status add/remove, so to solve the current issue we can use true/false and if needed we can change this in the future.

pkg/binding/selectors.go
vMaroon: Perhaps both this and the error message after the other call can be changed to directly reflect controller::handleSingletonLabel failing (this was not introduced in this PR)

PR #1920, opened 2024-03-19
✨Separate CRDs in the Helm chart

francostellari:
## Summary

Move CRDs to the chart `crds/` folder to let Helm install them before applying the template

## Related issue(s)

Fixes #

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -15,7 +15,6 @@ namePrefix: kubestellar-
 #    someName: someValue
 
 resources:
-# - ../crd
 - ../rbac
 - ../manager
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in


config/default/kustomization.yaml
```diff
@@ -15,7 +15,7 @@ namePrefix: kubestellar-
 #    someName: someValue
 
 resources:
-- ../crd
+# - ../crd
```
MikeSpreitzer, line 18: What was this kustomization accomplishing?

If we are no longer going to use that, perhaps it should be deleted?

francostellari: Removing this line prevents the CRDs to be injected into the chart template yaml.
When doing the same in kubefkex and otp, we commented the line instead of completely removing it.
I can remove it if you prefer

MikeSpreitzer: In my own testing and analysis, it appears that this particular line was accomplishing nothing. It is a bit of generic infrastructure inserted by kubebuilder, but it never made a difference in this repo.

Removing this line makes all of config/crd/ unreferenced junk.

francostellari: ok done

MikeSpreitzer: If I understand correctly, https://github.com/kubestellar/kubestellar/blob/821fb7a49c875b5aa433832b38429be457d2e86a/Makefile#L140 is also saying --- and relying on the fact that --- this repo is not using any kustomize functionality for its CRDs. @pdettori , have I got this right?

pdettori: @MikeSpreitzer correct, at least for the CRD there is currently no kustomization done.

MikeSpreitzer (commented):
/lgtm

PR #1931, opened 2024-03-20
🐛 Clean metadata.generateName in workload objects before wrapping

### LINKED ISSUE #1852: bug: objects with "generateName" set are created in the WDS but not in the WEC
### Describe the bug

Objects with `generateName` set are created in the WDS but not in the WEC. The manifestwork for an object with `generateName` set is created, but the status conditions show an error from the OCM work agent:

```
"Error delivering object to mailbox" err="admission webhook \"manifestworkvalidators.admission.work.open-cluster-management.io\" denied the request: generateName must not be set in manifest" logger="Binding"
``` 

### Steps To Reproduce

1. Create an object with `generateName` set and match it with a binding policy for delivery to a WEC. The object that initially caused the issue is the following:
```yaml
apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
  generateName: openshift-nfd-
  generation: 1
  labels:
    app.kubernetes.io/part-of: wec-nfd
  name: openshift-nfd
  namespace: openshift-nfd
spec:
  targetNamespaces:
  - openshift-nfd
  upgradeStrategy: Default
```
2. Check the manifest work for the object in imbs1 and verify it is not delivered to the web.

### Expected Behavior

Since OCM does not deliver such objects, user should be informed or prevented from trying to deliver objects with `generateName` set.

### Additional Context

There are several options here on how to address this:
- just document the limitation
- propagate the error from the OCM work agent to the binding policy status
- run an admission WebHook on the WDS that does the check and rejects objects with `generateName` set

waltforme:
## Summary
The background of this PR is documented in #1852. In the discussions of #1852, we agreed with an approach to (a) let the names be generated in WDSes, (b) clean the `.metadata.generateName` field in workload objects; (3) wrap the objects with name already generated for transport.

This PR implements the approach.

## Related issue(s)
Fixes #1852

### DIFF
--- a/pkg/transport/generic_transport_controller.go
+++ b/pkg/transport/generic_transport_controller.go
@@ -604,6 +604,7 @@ func cleanObject(object *unstructured.Unstructured) *unstructured.Unstructured {
 	objectCopy.SetSelfLink("")
 	objectCopy.SetResourceVersion("")
 	objectCopy.SetUID("")
+	objectCopy.SetGenerateName("")
 
 	annotations := objectCopy.GetAnnotations()
 	delete(annotations, "kubectl.kubernetes.io/last-applied-configuration")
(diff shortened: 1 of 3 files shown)


pdettori (commented):
/lgtm

PR #1934, opened 2024-03-20
🐛 Demonstrate list item numbering problem

MikeSpreitzer:
## Summary
This problem attempts to fix the problem with list item numbering on the website.
The fix does not work, as you can verify by previewing according to the instructions in https://github.com/kubestellar/kubestellar/tree/main/docs#serving-up-documents-locally

## Related issue(s)

Fixes #


MikeSpreitzer:
/cc @pdettori 
/cc @clubanderson 

/hold
.. because this does not help.

Issue #1894, opened 2024-03-13
bug:  resources with status not defined as subresource are not correctly handled


pdettori:
### Describe the bug

Not all resources follow the best practice of defining status as a subresource. For example, the Argo Workflow project has a "workflows" resource that does not have the status defined as subresource. See https://github.com/argoproj/argo-workflows/blob/5b3909b9567a5c6fc4fef81ac0785cfddbffb291/manifests/base/crds/minimal/argoproj.io_workflows.yaml#L55.

As a result, there are two problems:

1. When down syncing a resource that has a status which is NOT a subresource, the OCM transport does not filter out the status and that is synced downstream. This creates problems in some use cases (e.g. Argo Workflows) where the WDS is of type "host" and we want to suspend a workflow on the WDS (where there are active controllers for workflows) and delegate the execution on other clusters (via KubeStellar), as the up sync status overrides the locally generated status downstream. Note that this is NOT an issue when status IS a subresource, as the OCM transport does NOT sync downstream a status subresource. This can be verified on a OCM Hub + Managed Cluster by applying [foo-crd.yaml.txt](https://github.com/kubestellar/kubestellar/files/14588890/foo-crd.yaml.txt) in the managed cluster and 
[foo-manifest.yaml.txt](https://github.com/kubestellar/kubestellar/files/14588895/foo-manifest.yaml.txt) in the hub. The status is present but since it is a subresource it is not propagated.

2. The status controller is assuming status is a subresource, so it does not update the status of the corresponding object in the WDS when using singleton status. See https://github.com/kubestellar/kubestellar/blob/5554d0a3a7fe88ba64a647485c48944dbddf27e9/pkg/status/controller.go#L374-L377

### Steps To Reproduce

1. Setup a kubestellar environment
2. Follow steps in [scenario 2](https://github.com/kubestellar/kubestellar/blob/main/docs/content/direct/examples.md#scenario-2---using-the-hosting-cluster-as-wds-to-deploy-a-custom-resource) to create a WDS of "host" type
3. Install argo workflows on both hosting cluster and managed clusters (see https://argo-workflows.readthedocs.io/en/latest/quick-start/#install-argo-workflows )
4. Write binding policy to match workflows with wantSingletonReportedStatus to deliver to one WEC a workflow object
5. Create a workflow with "spec.suspend=true" 
6. Observe the issue

### Expected Behavior

1. Status from a WDS object should not be propagated downstream
2. Status should be correctly updated in the WDS object

[...]

PR #1937, opened 2024-03-21
:bug: Filter and update status if not subresource

andreyod:
## Summary
1. Filter out the status when down syncing a resource. The OCM transport does not filter out the status which is not a subresource(for example Argo Workflow resource).

2. The status controller should update the status regardless to subresource or not.

## Related issue(s)

Fixes https://github.com/kubestellar/kubestellar/issues/1894

### DIFF
--- a/pkg/status/controller.go
+++ b/pkg/status/controller.go
@@ -328,8 +328,16 @@ func updateObjectStatus(ctx context.Context, objRef *util.SourceRef, status map[
 
 	if objRef.Namespace == "" {
 		_, err = wdsDynClient.Resource(gvr).UpdateStatus(ctx, unstrObj, metav1.UpdateOptions{})
+		// if resource not found it may mean no status subresource - try to update the whole resource
+		if errors.IsNotFound(err) {
+			_, err = wdsDynClient.Resource(gvr).Update(ctx, unstrObj, metav1.UpdateOptions{})
+		}
 	} else {
 		_, err = wdsDynClient.Resource(gvr).Namespace(objRef.Namespace).UpdateStatus(ctx, unstrObj, metav1.UpdateOptions{})
+		// if resource not found it may mean no status subresource - try to update the whole resource
+		if errors.IsNotFound(err) {
+			_, err = wdsDynClient.Resource(gvr).Namespace(objRef.Namespace).Update(ctx, unstrObj, metav1.UpdateOptions{})
+		}
 	}
 	if err != nil {
 		return fmt.Errorf("failed to update status: %w", err)
--- a/pkg/status/controller.go
+++ b/pkg/status/controller.go
@@ -328,18 +328,14 @@ func updateObjectStatus(ctx context.Context, objRef *util.SourceRef, status map[
 
 	if objRef.Namespace == "" {
 		_, err = wdsDynClient.Resource(gvr).UpdateStatus(ctx, unstrObj, metav1.UpdateOptions{})
-		// if resource not found it may mean no status subresource - try to update the whole resource
-		if errors.IsNotFound(err) {
-			_, err = wdsDynClient.Resource(gvr).Update(ctx, unstrObj, metav1.UpdateOptions{})
-		}
 	} else {
 		_, err = wdsDynClient.Resource(gvr).Namespace(objRef.Namespace).UpdateStatus(ctx, unstrObj, metav1.UpdateOptions{})
-		// if resource not found it may mean no status subresource - try to update the whole resource
-		if errors.IsNotFound(err) {
-			_, err = wdsDynClient.Resource(gvr).Namespace(objRef.Namespace).Update(ctx, unstrObj, metav1.UpdateOptions{})
-		}
 	}
 	if err != nil {
+		// if resource not found it may mean no status subresource - try to patch the status
+		if errors.IsNotFound(err) {
+			return util.PatchStatus(ctx, unstrObj, status, objRef.Namespace, gvr, wdsDynClient)
+		}
 		return fmt.Errorf("failed to update status: %w", err)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/status/controller.go
```diff
@@ -328,8 +328,16 @@ func updateObjectStatus(ctx context.Context, objRef *util.SourceRef, status map[
 
 	if objRef.Namespace == "" {
 		_, err = wdsDynClient.Resource(gvr).UpdateStatus(ctx, unstrObj, metav1.UpdateOptions{})
+		// if resource not found it may mean no status subresource - try to update the whole resource
+		if errors.IsNotFound(err) {
+			_, err = wdsDynClient.Resource(gvr).Update(ctx, unstrObj, metav1.UpdateOptions{})
+		}
```
pdettori, line 334: This works, but may not be efficient as for non-subresources statuses there are alway two calls for each status update, the first of which triggers the not found error, and considering that status can be updated quite often may not be optimal. I would suggest to think about a solution that can lookup if status is a sub-resource or not and then does the correct thing. This was done previously in legacy KS - please have a look at https://github.com/kubestellar/kubestellar/pull/1228

pkg/status/controller.go
```diff
@@ -328,8 +328,16 @@ func updateObjectStatus(ctx context.Context, objRef *util.SourceRef, status map[
...
 	} else {
 		_, err = wdsDynClient.Resource(gvr).Namespace(objRef.Namespace).UpdateStatus(ctx, unstrObj, metav1.UpdateOptions{})
+		// if resource not found it may mean no status subresource - try to update the whole resource
+		if errors.IsNotFound(err) {
+			_, err = wdsDynClient.Resource(gvr).Namespace(objRef.Namespace).Update(ctx, unstrObj, metav1.UpdateOptions{})
```
pdettori, line 339: In addition to previous comment, it may be more efficient just to update the status section with a patch rather than an update of the whole object. For example, I was experimenting with:

```
patchBytes, _ := json.Marshal(statusPatch.Object)
			_, err = wdsDynClient.Resource(gvr).Namespace(objRef.Namespace).Patch(ctx, statusPatch.GetName(), types.MergePatchType, patchBytes, metav1.PatchOptions{})
```

where:

```
func createStatusPatch(unstrObj *unstructured.Unstructured, status map[string]interface{}) *unstructured.Unstructured {
	patchedObj := &unstructured.Unstructured{}
	patchedObj.SetAPIVersion(unstrObj.GetAPIVersion())
	patchedObj.SetKind(unstrObj.GetKind())
	patchedObj.SetName(unstrObj.GetName())
	patchedObj.SetNamespace(unstrObj.GetNamespace())
	patchedObj.Object["status"] = status
	return patchedObj
}
```

andreyod: Is `patch` API call more efficient than `update` API call?

pkg/transport/generic_transport_controller.go
```diff
@@ -610,6 +610,9 @@ func cleanObject(object *unstructured.Unstructured) *unstructured.Unstructured {
 	delete(annotations, "kubectl.kubernetes.io/last-applied-configuration")
 	objectCopy.SetAnnotations(annotations)
 
+	// OCM removes the status only if subresource. We remove it in all cases.
```
nirrozenbaum, line 613: can we remove the OCM comment?
generic transport code doesn’t depend on ocm and shouldn’t depend.
it also shouldn’t care which transport is used, so specifying here ocm in the comments is confusing

pkg/status/controller.go
pdettori: The difference here is that using the update method the whole object (spec and status) is retrieved from the lister, status is updated on that object and then the update updates the whole object (spec and status). The approach I proposed above only sets API version, kind, name and namespace and the status to make the patch. So I expect it to be more efficient as less content is sent over the wire, and it seems also cleaner to me. On the other hand, this was not an issue for the status subresource as in that case only status is being updated.

pkg/status/controller.go
andreyod: After looking at https://github.com/kubestellar/kubestellar/pull/1228, I don't think this will be more efficient, because the subresource lookup also has an API call under the hood in the `restmapper.GetAPIGroupResources(discoveryClient)` to get the `groupResources`. And we need to get the `groupResources` list for each status update, because if we do this once, then any CRD added after this point in time will not be on the list.
WDYT?

pdettori: @andreyod you are correct, if the restmapper.GetAPIGroupResources(discoveryClient) is called every time this would be even worse (especially if there are many api resources such as in OCP). I was suggesting that since we do have code in the binding controller where we detect new CRDs we could at that point update the rest mapper and/or related structs that would keep the flag about status subresource. However this might be a further optimization I am ok to defer, so for now I am fine with the current approach you implemented since it is simple and works for the current purpose.

pkg/status/controller.go
andreyod: Switched to use `patch` instead of `update`.

pkg/transport/generic_transport_controller.go
andreyod: Changed the comment.

pkg/util/unstructured.go
```diff
@@ -182,3 +187,32 @@ func CheckAPIisPresent(dc *discovery.DiscoveryClient, gvr schema.GroupVersionRes
...
+	patchBytes, err := json.Marshal(CreateStatusPatch(unstrObj, status))
+	if err != nil {
+		return err
+	}
+
+	if namespace != "" {
+		_, err = dynamicClient.Resource(gvr).Namespace(namespace).Patch(ctx, unstrObj.GetName(), types.MergePatchType, patchBytes, metav1.PatchOptions{})
+	} else {
+		_, err = dynamicClient.Resource(gvr).Patch(ctx, unstrObj.GetName(), types.MergePatchType, patchBytes, metav1.PatchOptions{})
+	}
```
nirrozenbaum, line 215: why not using in function `UpdateStatus`?

MikeSpreitzer: When this func is called, it is already known that the generated UpdateStatus method failed.

PR #1940, opened 2024-03-21
:book: Create deploy-on-k3d.md

eaepstein:
Use K3D instead of Kind for improved performance.

## Summary
Document steps for running KS v0.21.0 on K3D.  The procedure could be simplified to more closely match running on Kind.

## Related issue(s)

Fixes #


eaepstein:
Fix commit and improve document

Issue #1866, opened 2024-03-07
feature: deploy transport controller through a helm chart


nirrozenbaum:
### Feature Description

up until now pluggable transport controller was running as an executable. 
we've made some work to run it in a pod and deploy it using a script.
we would like to have transport controller deployed in a pod through a helm chart.

### Proposed Solution

either add transport to the existing kubestellar helm chart (the preffered approach)
or create new helm chart for transport.

### Want to contribute?

- [ ] I would like to work on this issue.

francostellari:
Kubestellar documentation will be updated when OTP provides the first release of the new helm chart

francostellari:
@nirrozenbaum can you work on a new release with the OTP chart , when you have time?

nirrozenbaum:
@francostellari created release v0.1.0

PR #1942, opened 2024-03-21
📖 Update docs to use OTP Helm Chart

francostellari:
## Summary

Update docs to use OTP Helm Chart

Pending OTP chart 0.1.1 release

## Related issue(s)

Fixes #
#1866


test/e2e/common/setup-kubestellar.sh
```diff
@@ -90,23 +90,13 @@ echo "wds1 created."
...
-echo "running ocm transport plugin..."
-kubectl config use-context kind-kubeflex ## transport deployment script assumes it runs within kubeflex context
-IMAGE_PULL_POLICY=Never ./scripts/deploy-transport-controller.sh wds1 imbs1 ko.local/transport-controller:${OCM_TRANSPORT_PLUGIN_RELEASE}
+OCM_TRANSPORT_PLUGIN_RELEASE="0.1.1"
+
+kflex ctx # root
+
+helm upgrade --install ocm-transport-plugin oci://ghcr.io/kubestellar/ocm-transport-plugin/chart/ocm-transport-plugin --version ${OCM_TRANSPORT_PLUGIN} \
+--set transport_cp_name=imbs1 \
+--set wds_cp_name=wds1
```
nirrozenbaum, line 99: This will not work.
the e2e setup should build local image from the OTP repo.
explanation:
OTP uses ks/ks functions. if we now change any of the ks/ks transport related functions, we need to build locally the OTP image. otherwise we don't test the latest change.

scripts/deploy-transport-controller.sh
```diff
@@ -1,112 +0,0 @@
-#!/usr/bin/env bash
-# Copyright 2024 The KubeStellar Authors.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
```
nirrozenbaum, line 6: are we sure we want to delete this file?
until there's a different way to deploy a locally built image (e.g., for e2e purpose or just for testing my changes), I wouldn't remove it.

test/e2e/common/setup-kubestellar.sh
francostellari: Since OTP builds its own image and chart, should those be separately tested in the OTP repo.
Kubestellar is importing an existing OTP chart for use.
I believe this is not different than KS using kflex chart... and not building kflex locally.
Or not build OCM locally, etc etc
Since we are advocating for separation, a KS test should not require a OTP build, in my opinion.
May be we can discus this in one of the meetings.

nirrozenbaum: I agree with you on the general idea. 
I agree that KS test should **NOT** require OTP local build. 
but in reality OTP currently has some issues in terms of the dependencies between KS and OTP. things have to be improved and separated better. 
unfortunately, until that happens there is no good way to test KS without OTP local build.

ezrasilvera: @francostellari Currently our e2e is aimed for the developer - it builds the local KS , while using exiting packages of the other components. 
While OTP resides on different repo and is has it's own image - it's not really an "external" package and the design today is that it needs to be build against a specific KS version.  So due to this dependency, for the testing of latest KS you do need to build OTP against that code which means you can't use pre-built image

ezrasilvera: BTW, we are planning to add e2e for releases, for that e2e we will use release images for both KS and OTP

ezrasilvera: In the future we are discussing with @nirrozenbaum how to change this OTP <--> KS dependency so you don't need to build OTP against the KS code.

francostellari: @ezrasilvera so, in conclusion, I should change the user facing instructions to use Helm but the e2e will remain as is testing an alternative process. I will revert the changes with the exception of examples.md. Right?

ezrasilvera: > @ezrasilvera so, in conclusion, I should change the user facing instructions to use Helm but the e2e will remain as is testing an alternative process. I will revert the changes with the exception of examples.md. Right?

@francostellari Yes.  BTW, @effi-ofer is developing e2e that tests releases, those tests can use helm.

francostellari:
@nirrozenbaum @ezrasilvera I have reverted back the changes to the e2e. I left only the documentation change.
I think this is still pending the new chart release of 0.1.1.

docs/content/direct/examples.md
```diff
@@ -71,24 +72,18 @@ manager which connects to the `wds1` front-end and the `imbs1` OCM control plane
    kflex create wds1 -p kubestellar
    ```
 
-1. Run the OCM based transport controller in a pod.  
-**NOTE**: This is work in progress, in the future the controller will be deployed through a Helm chart.
+1. Run the OCM based transport controller in a pod.
```
pdettori, line 76: Everything we deploy now runs in a pod, I would just say "Deploy the OCM based transport controller"

docs/content/direct/examples.md
```diff
@@ -257,17 +255,17 @@ hosting cluster so that we can pass additional startup options.
...
 [multi cluster app dispatcher](https://github.com/project-codeflare/multi-cluster-app-dispatcher)
 project.
 
-Install the AppWrapper CRD in the WDS and the WECs. Note that due to 
-[this issue](https://github.com/kubestellar/kubestellar/issues/1705) CRDs must be pre-installed 
-on the WDS and on the WECs when using API group filtering. For this release of KubeStellar, that pre-installation is required 
+Install the AppWrapper CRD in the WDS and the WECs. Note that due to
+[this issue](https://github.com/kubestellar/kubestellar/issues/1705) CRDs must be pre-installed
+on the WDS and on the WECs when using API group filtering. For this release of KubeStellar, that pre-installation is required
 when a WDS has a large number of API resources (such as would be found in a hosting cluster that is OpenShift).
```
pdettori, line 269: The whole "Note that... " sentence can be removed, that issue has been fixed.

docs/content/direct/examples.md
francostellari: ok, done

docs/content/direct/examples.md
francostellari: ok, removed

francostellari:
I messed up the rebase as usual

PR #1944, opened 2024-03-21
 :sparkles: New release - 0.21.1

ezrasilvera:
## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/direct/release-notes.md
+++ b/docs/content/direct/release-notes.md
@@ -4,6 +4,9 @@ The following sections list the known issues for each release. The issue list is
 
 ## 0.21.0 and its release candidates
 
+### Patch release  0.21.1 
+This patch release mainly updates the documentation exposed under kubestellar.io . In all other aspects it is identical to 0.21.0
+
 ### Major changes for 0.21.0 and its release candidates
 
 * This release introduces pluggable transport. Currently the only plugin is [the OCM transport plugin](https://github.com/kubestellar/ocm-transport-plugin).
--- a/docs/content/direct/release-notes.md
+++ b/docs/content/direct/release-notes.md
@@ -2,10 +2,11 @@
 
 The following sections list the known issues for each release. The issue list is not differential (i.e., compared to previous releases) but a full list representing the overall state of the specific release. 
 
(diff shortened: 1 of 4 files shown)


docs/content/direct/release-notes.md
```diff
@@ -4,6 +4,9 @@ The following sections list the known issues for each release. The issue list is
 
 ## 0.21.0 and its release candidates
 
+### Patch release  0.21.1 
+This patch release mainly updates the documentation exposed under kubestellar.io . In all other aspects it is identical to 0.21.0
```
MikeSpreitzer, line 8: > In all other aspects it is identical to 0.21.0

No, this is not true. There have been a few bug fixes.

I do not think that it is necessary to list them all here; GitHub automatically collects a changelog (see https://github.com/kubestellar/kubestellar/releases/tag/v0.21.0 for an example).  I might simply remove the inaccurate sentence.

docs/content/direct/release-notes.md
```diff
@@ -4,6 +4,9 @@ The following sections list the known issues for each release. The issue list is
 
 ## 0.21.0 and its release candidates
 
+### Patch release  0.21.1
```
MikeSpreitzer, line 7: "Patch release  0.21.1 " does not logically fit under "0.21.0 and its release candidates". BTW, even for a patch release we will have release candidates. I think the best approach is to simply have "## 0.21.1 and its release candidates".

MikeSpreitzer:
This is not all that we need to do in order to prepare for a release.  See https://github.com/kubestellar/kubestellar/blob/main/docs/content/direct/release.md

docs/content/direct/release-notes.md
ezrasilvera: Done

MikeSpreitzer (commented):
/lgtm
Thanks!

MikeSpreitzer:
In double-checking, I did find this: https://github.com/kubestellar/kubestellar/blob/3df25e469cb8b6e45334ba7e3f40d488a79a9800/README.md?plain=1#L13

PR #1947, opened 2024-03-21
✨ Script to build kubectl image

francostellari:
## Summary

Script to build kubectl image

## Related issue(s)

Fixes #

### DIFF
--- a/hack/Dockerfile.kubectl
+++ b/hack/Dockerfile.kubectl
@@ -1,0 +1,43 @@
+# Copyright YEAR The KubeStellar Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal
+
+ARG TARGETOS
+ARG TARGETARCH
+ARG VERSION
+
+RUN [ "$VERSION" == "" ] && VERSION=$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt) ; \
+    curl \
+	-L https://storage.googleapis.com/kubernetes-release/release/v${VERSION#v}/bin/${TARGETOS}/${TARGETARCH}/kubectl \
+	-o /usr/bin/kubectl && chmod +x /usr/bin/kubectl && \
+    curl \
+	-L https://github.com/stedolan/jq/releases/latest/download/jq-${TARGETOS}-${TARGETARCH} \
+	-o /usr/bin/jq && \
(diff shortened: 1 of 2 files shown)


hack/Dockerfile.kubectl
```diff
@@ -0,0 +1,43 @@
+# Copyright YEAR The KubeStellar Authors.
```
waltforme, line 1: Sounds like any `YEAR`?

hack/Dockerfile.kubectl
```diff
@@ -0,0 +1,43 @@
...
+# RUN microdnf install -y findutils
+
+USER 1001
+
+ENTRYPOINT [ "kubectl" ]
+CMD [ "--help" ]
+
+
+# https://github.com/AICoE/kubectl-container/blob/master/Dockerfile
+# https://gist.github.com/rponte/fdc0724dd984088606b0
```
waltforme, line 43: Some comment about this red symbol: https://github.com/kubestellar/kubestellar/pull/1967#discussion_r1541741908.

hack/build-kubectl-image.sh
```diff
@@ -0,0 +1,83 @@
...
+
+docker buildx create --name kubestellar --platform ${platform} --use
+
+docker buildx build . \
+    --push \
+    --platform ${platform} \
+    --tag quay.io/kubestellar/kubectl:${version#v} \
+    --build-arg version=${version#v} \
+    --tag quay.io/kubestellar/kubectl:${version#v} \
+    -f /vagrant/Dockerfile.kubectl
```
waltforme, line 81: I don't understand the `vagrant` here.

hack/Dockerfile.kubectl
francostellari: @waltforme thank you for catching it. Fixed in the commit.

hack/Dockerfile.kubectl
francostellari: Removed, I used as abase

hack/Dockerfile.kubectl
```diff
@@ -0,0 +1,43 @@
...
```
waltforme, line 43: BTW, are these two references intentionally kept in this dockerfile?

hack/Dockerfile.kubectl
```diff
@@ -0,0 +1,43 @@
...
+    curl \
+	-L https://github.com/stedolan/jq/releases/latest/download/jq-${TARGETOS}-${TARGETARCH} \
+	-o /usr/bin/jq && \
+	chmod +x /usr/bin/jq && \
+    curl \
+	-L https://github.com/mikefarah/yq/releases/latest/download/yq_${TARGETOS}_${TARGETARCH} \
+	-o /usr/bin/yq && \
+	chmod +x /usr/bin/yq
+
+# RUN microdnf install -y findutils
```
waltforme, line 34: Maybe we just remove this if we don't need it?

hack/build-kubectl-image.sh
francostellari: leftover from testing, fixed

hack/Dockerfile.kubectl
francostellari: removed

hack/Dockerfile.kubectl
francostellari: yes, I commented to check if we needed it and forgot to remove it
removed it now

PR #1950, opened 2024-03-22
✨ Add customization by go template expansion

### LINKED ISSUE #1897: feature:  Workload customization per WEC
### Feature Description
In some cases when a workload is deployed to multiple WECs users may want to personalized the workload parameters for specific WEC(s)

### Current state description and problems involved
In general a user can achieve the above by:
1. Create a different workload object for each WEC
2. Create a binding policy that will send the workload to a specific WEC

While this may be good for some cases this is not a practical solution for several reason:
1. If we want to personalize cluster scope resources we need to use different WDS for each instance
2. Scalability issues: we basically duplicate the object on the WDS(s) which become impractical for large number of WECs
3. Manageability:  The user need to manage all those copies + the different binding policies. 

### Assumptions
1. For phase 1, we will be focusing on Hub based customization 
2. For phase 1 we will start with simple parameter substitution 

### Description
The user will specify a template with placeholders for fields that need to be personalized.  Once the target WEC is known KS will populate the placeholders with the WEC specific values.  

### Additional requirements
1. Some common fields should be pre-defined by KS (e.g., "wec-name" )
3. The should be able to add workload specific fields  (e.g., "debug-level")
4. Simple solution - if possible use existing mechanism or know syntax (e.g., Go templating) 
5. Very simple UX - user should easily define new fields and/or use existing pre-defined system fields.

### LINKED ISSUE #1964: bug: docs/content/direct/README.md has broken link to user-guide.md
### Describe the bug

https://github.com/kubestellar/kubestellar/blob/main/docs/content/direct/README.md refers to a file that does not exist, `user-guide.md`. PR #1962 renamed that file but did not update the reference here. The re-org done in that PR was exclusively focused on the view available through the website and ignored docs/content/direct/README.md.

### Steps To Reproduce

See above.

### Expected Behavior

No docs broken. The docs/content/direct/ content is viewable both through the website and through GitHub's direct view.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR adds template expansion, a form of parametric customization.

This PR expands the schema of a Binding, by adding a status subresource that the transport controller uses to report user errors.

This is based on #1760, which introduced a different form of parametric customization --- parameter expansion. This PR adds template expansion _instead of_ parameter expansion.

## Related issue(s)

Fixes #1897 
Fixes #1964


api/control/v1alpha1/types.go
```diff
@@ -20,6 +20,25 @@ import (
...
+// name to parameter value.
+//
+// A destination is described by an inventory object in an Inventory and Transport Space,
+// and its labels and annotations provide parameter names and values (with labels taking priority over annotations).
+//
+// Note that this sort of customization has limited applicability.  It can only be used where
+// the un-expanded string passes the validation conditions of the relevant object type.
+// For more broadly applicable customization, see Customizer objects.
+
+const TemplateExpansionAnnotationKey string = "control.kubestellar.io/expand-templates"
```
nirrozenbaum, line 40: can we name this annotation with a more "user-friendly" name?
I would expect something like `control.kubestellar.io/customization: true`. 
customization is a feature of KS which will be described in KS docs (both website and git readme) up completion. 
I think users of the KS may understand this much better with this name after they read the docs.

api/control/v1alpha1/types.go
```diff
@@ -245,6 +267,11 @@ type Destination struct {
 	ClusterId string `json:"clusterId,omitempty"`
 }
 
+type BindingStatus struct {
+	ObservedGeneration int64    `json:"observedGeneration"`
+	Errors             []string `json:"errors,omitempty"`
```
nirrozenbaum, line 272: I see your point about adding errors field in `Binding` status. 
Having said that, the user creates `BindingPolicy`, not `Binding`. 
I assume a typical K8s user would look for the error in the object he created (meaning in `BindingPolicy`).

docs/content/direct/user-guide.md
```diff
@@ -31,3 +31,59 @@ Additionally, the recipe for registering a WEC with the ITS needs to be modified
 ## When everythig is not on the same machine
 
 Thus far we can only say how to handle this when the hosting cluster is OpenShift. The problem is getting URLs that work from everywhere. OpenShift is a hosted product, your clusters have domain names that are resolvable from everywhere. In other words, if you use an OpenShift cluster as your hosting cluster then this problem is already solved.
+
+## Rule-based customization
+
+KubeStellar can distribute one workload object to multiple WECs, and it is not uncommon for users to need some customization to each WEC. By _rule based_ we mean that the customization is not expressed via one or more literal expressions but rather can refer to _properties_ of each WEC by property name. As KubeStellar distributes or transports a workload object from WDS to a WEC, the object can be transformed in a way that depends on those properties.
```
nirrozenbaum, line 37: can we use positive description instead of using negative twice in a row?
e.g.,  instead of: 
`and it is not uncommon for users to... `
use:
`and is it common for users.. `

docs/content/direct/user-guide.md
```diff
@@ -31,3 +31,59 @@ Additionally, the recipe for registering a WEC with the ITS needs to be modified
...
+
+KubeStellar currently has a simple but limited way to specify rule-based customization, called "template expansion". This works with templates as defined in the "text/template" standard package of Go, and is applied to each leaf string of an object independently. More expressive customization will be developed in the future.
```
nirrozenbaum, line 39: I would remove the sentence `More expressive customization will be developed in the future.`
it doesn't add any information. if and when extending the customization we can describe it (as any other feature).

docs/content/direct/user-guide.md
```diff
@@ -31,3 +31,59 @@ Additionally, the recipe for registering a WEC with the ITS needs to be modified
...
+
+### Template Expansion
+
+Template expansion is an optional feature that a user can request on an object-by-object basis. The way to request this feature on an object is to put the following annotation on the object.
+
+```yaml
+    control.kubestellar.io/expand-templates: "true"
+```
+
+The customization that template expansion does when distributing an object from a WDS to a WEC is applied independently to each leaf string of the object. The string is first parsed as a template, and then replaced with the result of expanding the template (using the `Execute` method of a `*Template`) using a `map` holding key-value pairs conveying properties of the WEC. For the purposes of template expansion, the "properties" of a WEC are defined by the labels and annotations of the inventory object (`ManagedCluster`) representing that WEC in the inventory. Labels take precedence over annotations. Errors from this process are reported in the status field of the Binding object involved.
```
nirrozenbaum, line 49: this file is called `user-guide` but includes way too much implementation details.
a user-guide typically describes for a user what is the feature and how to use it. 
her the description includes which function of which package is used internally. 
```
using the `Execute` method of a `*Template`
```
a user-guide usually doesn't include implementation details. I would maybe mention that the the customization is based on the Go templating `text/template` module and that's it.

docs/content/direct/user-guide.md
```diff
@@ -31,3 +31,59 @@ Additionally, the recipe for registering a WEC with the ITS needs to be modified
...
+  name: virgo
+...
+```
+
+When distributed to the virgo WEC, that ClusterLogForwarded would say the following.
+
+```yaml
+...
+      url: "https://my.loki.server.com/virgo"
+...
```
nirrozenbaum, line 88: this example is excellent for user-guide.
it shows the user concrete example for how to use the feature.
no need for much more than that. 
short textual description and concrete example.

api/control/v1alpha1/types.go
MikeSpreitzer: I am open to other names. But template expansion is not the only form of customization that we contemplate, so I think that the name here should be more specific than simply "customization".

api/control/v1alpha1/types.go
MikeSpreitzer: Indeed. My thinking is that we can add propagation back to the BindingPolicy in a follow-on PR.

docs/content/direct/user-guide.md
MikeSpreitzer: Done.

docs/content/direct/user-guide.md
MikeSpreitzer: Reworded.

docs/content/direct/user-guide.md
MikeSpreitzer: I reworded to simplify.

pkg/transport/generic_transport_controller.go
```diff
@@ -19,16 +19,21 @@ package transport
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
+	clusterinformers "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	clusterlisters "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	clusterapi "open-cluster-management.io/api/cluster/v1"
```
nirrozenbaum, line 28: I really think using the clusters objects for storing customization parameters is a wrong decision.

clusters objects may have a lot of events (e.g, for updating periodic liveness status) that most of them are not related to customization and in general it's not very clear why would one use labels/annotations and the precedence order you defined.
labels are typically used for group selection not for configuration values. plus, any change in any cluster labels will trigger re-evaluation of all binding policies (through the binding controller). 

I think that using MC to store the parameters for customization is kind of a mix between non-related points that looks very awkward to me. 

IMO a much more native decision is to implement a very straight forward thin controller that runs in ITS, that watches clusters objects and upon creation of MC creates a matching ConfigMap in its dedicated ns with the same name (we can even define the MC as OwnerReference of the ConfigMap). this controller can ignore update events of MC (cause that's not interesting) and also ignore delete events (ownerReference can delete the configmap).
we can also define that the configmap includes default fields that KS provides like key `name` and value `cluster_name`. (e.g., what Andy wanted for Loki).

customization key value pairs are configuration values and as such, It feels much more native to store it in ConfigMap + it keeps the clear separation between Inventory and Customization parameters.

nirrozenbaum: @ezrasilvera @pdettori would be helpful to hear also your thoughts on this point.

pkg/transport/generic_transport_controller.go
```diff
@@ -61,27 +70,32 @@ var objectsFilter = filtering.NewObjectFilteringMap()
...
 	wdsClientset ksclientset.Interface, wdsDynamicClient dynamic.Interface,
 	transportDynamicClient dynamic.Interface, wdsName string, wrappedObjectGVR schema.GroupVersionResource) *genericTransportController {
 	dynamicInformerFactory := dynamicinformer.NewDynamicSharedInformerFactory(transportDynamicClient, 0)
 	wrappedObjectGenericInformer := dynamicInformerFactory.ForResource(wrappedObjectGVR)
 
 	transportController := &genericTransportController{
 		logger:                      klog.FromContext(ctx),
+		inventoryInformerSynced:     inventoryPreInformer.Informer().HasSynced,
+		inventoryLister:             inventoryPreInformer.Lister(),
+		bindingClient:               bindingClient,
```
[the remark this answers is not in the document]
nirrozenbaum: now I see that the only use of WdsClientset is to update Binding.
so even better, to replace WdsClientset with bindingClient (no need for both).

pkg/transport/generic_transport_controller.go
```diff
@@ -410,44 +465,174 @@ func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, bind
...
+	if cares == nil {
+		delete(c.bindingCares, bindingName)
+	} else {
+		c.bindingCares[bindingName] = cares
+	}
 }
 
-func (c *genericTransportController) propagateWrappedObjectToClusters(ctx context.Context, desiredWrappedObject *unstructured.Unstructured,
+func (c *genericTransportController) enqueueBindingsForCluster(clusterAny any) {
+	cluster := clusterAny.(*clusterapi.ManagedCluster)
```
nirrozenbaum, line 596: this is coming back to using MC to store customization parameters comment. 
so now we need to cast to OCM MC. that's a bad coupling :(

pkg/transport/cmd/generic-main.go
```diff
@@ -100,15 +102,31 @@ func GenericMain(transportImplementation transport.Transport) {
...
+		logger.Error(err, "failed to create OCM clientset for transport space")
+		klog.FlushAndExit(klog.ExitFlushTimeout, 1)
+	}
+
+	wdsKsInformerFactory := ksinformers.NewSharedInformerFactoryWithOptions(wdsClientset, defaultResyncPeriod)
+
+	ocmInformerFactory := clusterinformers.NewSharedInformerFactory(ocmClientset, defaultResyncPeriod)
 
-	factory := ksinformers.NewSharedInformerFactoryWithOptions(wdsClientset, defaultResyncPeriod)
+	inventoryPreInformer := ocmInformerFactory.Cluster().V1().ManagedClusters()
```
nirrozenbaum, line 115: I mentioned in other comment explanation about removing this, but anyway mentioning the following - 
usually an informer variable is called with the type of the object.
e.g., informer for pods is called `podInformer`, informer for binding is called `bindingInformer`, and informer for ManagedCluster should be called `managedClusterInformer` or just `clusterInformer`. those are the naming conventions in k8s. 
if you agree to the other comments about using configmap, then this comment is not relevant to this file but it will be relevant to the new thin controller I mentioned.

pkg/transport/generic_transport_controller.go
```diff
@@ -92,6 +106,7 @@ func NewTransportControllerForWrappedObjectGVR(ctx context.Context, bindingInfor
 		wdsClientset:                wdsClientset,
 		wdsDynamicClient:            wdsDynamicClient,
 		wdsName:                     wdsName,
+		bindingCares:                make(map[string]sets.Set[v1alpha1.Destination]),
```
nirrozenbaum, line 109: wouldn't it be easier to hold a map from destination to set of Binding names? (the opposite direction)
seems to me like a simpler approach since an event that represents a change in customization parameters is for a single destination (and then we have set of binding names right away without any loop/calculation).

nirrozenbaum, line 109: from reading the name `bindingCares`, it is not very clear that it's related to customization.
can we name this variable with a more meaningful name? something with customization?

pkg/transport/generic_transport_controller.go
```diff
@@ -615,3 +802,9 @@ func cleanObject(object *unstructured.Unstructured) *unstructured.Unstructured {
 
 	return objectCopy
 }
+
+func MapCopyInto[Key comparable, Val any](dest, src map[Key]Val) {
+	for key, val := range src {
+		dest[key] = val
+	}
+}
```
nirrozenbaum, line 810: if we switch to using configmap, this is not needed.
data field of configmap is a map[string]string.

pkg/transport/generic_transport_controller.go
```diff
@@ -366,13 +396,22 @@ func (c *genericTransportController) updateWrappedObjectsAndFinalizer(ctx contex
...
+	}
+	bindingCopy := binding.DeepCopy()
+	bindingCopy.Status = v1alpha1.BindingStatus{
+		ObservedGeneration: binding.Generation,
+		Errors:             bindingErrors,
+	}
+	_, err = c.bindingClient.UpdateStatus(ctx, bindingCopy, metav1.UpdateOptions{FieldManager: ControllerName})
 	if err != nil {
-		return fmt.Errorf("failed to build wrapped object from Binding '%s' - %w", binding.GetName(), err)
+		return err
```
nirrozenbaum, line 410: wrap err - 
```
return fmt.Errorf("failed to update status of Binding '%s' - %w", binding.GetName(), err)
```

pkg/transport/generic_transport_controller.go
```diff
@@ -385,6 +424,22 @@ func (c *genericTransportController) updateWrappedObjectsAndFinalizer(ctx contex
...
 
+func (c *genericTransportController) valuesForDestination(bindingName string, dest v1alpha1.Destination) map[string]string {
+	c.addBindingCare(bindingName, dest) // ensure any change notification received after this causes reconsideration
+	cluster, err := c.inventoryLister.Get(dest.ClusterId)
+	if err != nil {
+		// This should be a transient condition that is already going to be remedied,
+		// because this means that the caller is working on old information.
+		// Just return anything in this case.
+		c.logger.V(2).Info("Failed to get inventory object", "name", dest.ClusterId, "err", err)
+		return nil
```
nirrozenbaum, line 435: this assumption is wrong. if you return nil it will not be re-queued for re-evaluation.

pkg/customize/customize.go
```diff
@@ -0,0 +1,98 @@
...
+
+// Expander is something that can do template expansion on unmarshaled JSON data.
+type Expander struct {
+	// Errors is the `.Error()` of the errors encountered
+	Errors []string
+
+	// ChangedSome reports whether parameter expansion made any changes to the data.
+	ChangedSome bool
+
+	defs map[string]string
```
nirrozenbaum, line 34: I don't know what does the name `defs` mean. this is not very clear.
I assume by reading the code you meant it's the customization key-value parameters (of a specific WEC).
so:
(a) I would name the variable to a meaningful name, like `customizationParameters`.
(b) looks like a bad idea to initialize a new `Expander` for every Binding and for every destination.
why not creating one customizer object that transport holds and change the signature of the `customize` function to receive the `customizationParameters` and return as return values the errors? 
the customizationParameters don't change frequently (or at least I don't expect it to change frequently).
(c) getting back to one of the previous comments - I think this should be called `Customizer`. and the function should be called `customize`. these names are very unclear.
I think a reader can easily understand that a `Customizer` can `customize` an object using `customizationParameters`.
on the other hand, I think it's not very clear what is `ExpandTemplates` and `Expander`.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: I went with the simplest thing I could think of.

pkg/transport/generic_transport_controller.go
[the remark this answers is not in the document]
MikeSpreitzer: I agree. Passing the clientset rather than the client config makes the controller more testable.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: The direction taken is easier for the code that emits this relation; the opposite direction is easier for the code that consumes this relation.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: Right, we need this func because the current definition involves merging two maps.

pkg/transport/generic_transport_controller.go
nirrozenbaum: I think using ConfigMap would be much simpler :).
both in terms of code readability, maintenance and clear separation of concerns and in terms of UX which I think would much easier for users. users of k8s know that configmaps are used for storing key-value configuration pairs.

api/control/v1alpha1/types.go
nirrozenbaum: we should specify in our docs that KS customization is based on go templating. then we should be able to use just `customization` as the annotation name. whoever wants to customize an object should read the docs and understand it's based on go templating (sounds very reasonable to me).
we should do any customization that go templating does, not necessarily string substitution. 
I think you're trying to be too specific.

pkg/transport/generic_transport_controller.go
```diff
@@ -494,6 +679,7 @@ func (c *genericTransportController) createOrUpdateWrappedObject(ctx context.Con
 		if err != nil {
 			return fmt.Errorf("failed to create wrapped object '%s' in destination WEC mailbox namespace '%s' - %w", wrappedObject.GetName(), namespace, err)
 		}
+		klog.FromContext(ctx).V(3).Info("Created wrapped object to ITS", "namespace", namespace, "objectName", wrappedObject.GetName(), "wrappedObject", wrappedObject)
```
nirrozenbaum, line 682: can we remove this log line? 
I assume this was for debugging purposes locally? 
logging every create/update will flood the log with those messages and will make it harder to see errors.

pkg/transport/generic_transport_controller.go
```diff
@@ -504,6 +690,7 @@ func (c *genericTransportController) createOrUpdateWrappedObject(ctx context.Con
 	if err != nil {
 		return fmt.Errorf("failed to update wrapped object '%s' in destination WEC mailbox namespace '%s' - %w", wrappedObject.GetName(), namespace, err)
 	}
+	klog.FromContext(ctx).V(3).Info("Updated wrapped object to ITS", "namespace", namespace, "objectName", wrappedObject.GetName(), "wrappedObject", wrappedObject)
```
nirrozenbaum, line 693: same comment as the create

api/control/v1alpha1/types.go
MikeSpreitzer: I am remembering our other conversations about other styles of customization. I call one of them "replace-this-with-that". In Karmada, they have such a concept and call it "overrides". As noted in this PR, template expansion is very limited in its expressiveness. I think it is reasonably likely that we will introduce one or more additional forms of customization in the future. That is why I do not want to equate "customization" with "template expansion".

pkg/transport/generic_transport_controller.go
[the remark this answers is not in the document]
MikeSpreitzer: I misunderstood.
I now have made the suggested change.

pkg/transport/generic_transport_controller.go
```diff
@@ -308,6 +337,7 @@ func (c *genericTransportController) syncHandler(ctx context.Context, objectName
 	}
 
 	if isObjectBeingDeleted(binding) {
+		c.setBindingCares(binding.Name, nil)
```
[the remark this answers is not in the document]
MikeSpreitzer: > we shouldn't delete this entry before the next line succeeds (deleteWrappedObjectsAndFinalizer)

At this point it is known that the Binding object is being deleted. Even if the following line returns an error, there is no longer a need to respond to a change in a matching inventory object by enqueuing a reference to the Binding object.

> in addition, it's much more readable to have delete(c.bindingCares, bindingName) rather than calling set function with nil (which internally deletes the entry).

That seems strange to me. All the other manipulations of c.bindingCares are through funcs. It seems nautral to me for this one to also be through a func. It is like there is a little internal abstraction for this data structure about this relation.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: Sure, this can be renamed (along with the methods that manipulate it). The idea here is a little more primitive that customization. It is just about the relation between Binding objects and inventory objects. How about `bindingSensitivities`?

pkg/transport/generic_transport_controller.go
MikeSpreitzer: Does that wrap really add anything? This is already in the context of working on a Binding object. If the context's logger does not already have the Binding reference in its values, it should.

nirrozenbaum: yes, it actually does. it shows errors from all layers which helps in debugging the problem.

pkg/transport/generic_transport_controller.go
nirrozenbaum: do you store all bindings in this map or only ones that contain customization? (meaning at least one workload object has to be customized) 
if the answer is all binding objects - why do we need it?
it seems to me that this map is relevant only for binding objects that contain at least one workload object that required customization, is that correct?.

I would name it in the most self explaining name I could find. `bindingSensitivities` is very similar to `bindingCares` in a sense that it's not explaining the purpose. so a reader (in this case it's me) has to keep thinking what is the purpose of this field.

How about `bindingToDestinations`? and if it stores only customization related bindings then we can add it in a comment.

pkg/transport/generic_transport_controller.go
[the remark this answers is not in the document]
nirrozenbaum: 1. you're correct.
2. I agree with the explanation.

pkg/transport/generic_transport_controller.go
nirrozenbaum: do you have a reference for this point from sig-multicluster?
the ideal would be to get aligned with sig-multicluster.
 
intuitively, I think sig-multicluster wouldn't go into this direction of using both labels and annotations with a precedence order. I would be very surprised if that happens.

can we agree on checking how sig-multicluster defining cluster properties and get aligned with that?
It would also help us later in the inventory work for easier alignment with sig-multicluster.

api/control/v1alpha1/types.go
nirrozenbaum: so in your vision, let's assume in the future we introduce one more form of customization (let's call it XXX).
would it be valid to define in a single object two types of customization? both string substitution (for some fields) and XXX (for other fields)?
sounds to me like a valid scenario, isn't it?
so now do you expect users to put multiple annotations on object? one for every type of customization?

to make a long story short - we're not there yet. and I have some doubts we will be there in the near future.
until that happens, I think documenting that string subst is the only customization option and just use customization as annotation name is much easier for understanding. 
trying to be too specific for future maybes might get very confusing. 
for example - even though I know exactly what we're doing, this name confuse me.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: I am working on a revision that includes expanding this comment to reference the bigger picture, hopefully making it clearer why re-queuing is not needed in this situation.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: I have increased the V level for these debug logging statements.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: I did not mean to suggest that sig-multicluster was going to propagate cluster self-description properties into inventory object labels and annotations; I think it would be into some modeled part of the inventory object. I got this in a verbal discussion in a meeting. There might be a recording. There might be something written down.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: A self-explaining name would be `bindingToDestinationsWhosePropertiesMatterToThatBinding`

pkg/customize/customize.go
MikeSpreitzer: I revised `pkg/customize` along the lines you suggested.

pkg/transport/cmd/generic-main.go
MikeSpreitzer: There is an unfortunate collision of names here. The type defined at https://github.com/open-cluster-management-io/api/blob/v0.12.0/client/cluster/informers/externalversions/cluster/v1/managedcluster.go#L21 is called "ManagedClusterInformer" and yet has a method that is named "Informer" and returns something that is _also_ called an informer and, in particular, is an informer of ManagedCluster objects. I have tended to call the one that precedes the other a "PreInformer", to be clearer about which is which. When looking at code like the line that we are discussing here, nothing else explicitly says this.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: I have tried shortening that to something manageable, `bindingSensitiveDestinations`.

api/control/v1alpha1/types.go
MikeSpreitzer: For example: in release-0.14 and earlier of KubeStellar, you can see in https://github.com/kubestellar/kubestellar/blob/v0.14.0/pkg/apis/edge/v2alpha1/customize.go that we had two forms of customization and they did involve two annotations but the value of one of them was not true/false but rather a reference to an object.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: Revised to not be specific to MangedCluster.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: Revised to wrap.

pkg/transport/generic_transport_controller.go
```diff
@@ -410,44 +465,174 @@ func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, bind
...
+
+	bindingErrors := []string{}
+	destToDefs := map[v1alpha1.Destination]map[string]string{}
+
+	// This will become non-nil if any object to propagate needs customization
+	var destToCustomizedObjects map[v1alpha1.Destination][]*unstructured.Unstructured
+
+	// Look through the objects to propagate to see if any needs customization.
+	// If any needs customization then catch up destToCustomizedObjects and proceed from there.
+	for objIdx, objToPropagate := range objectsToPropagate {
```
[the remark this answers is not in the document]
MikeSpreitzer: The current draft has a different form of simplicity: the distinction between objects that request template expansion and those that do not is concentrated in a narrower part of the code, just the internals of one method (initializeWrappedObject).

The current draft has a different form of efficiency. Iterating over the slice of objects twice instead of once makes a smaller difference than doing unnecessary template expansions. The current draft realizes the opportunity to recognize that an object that requests template expansion does not actually use it. I think that this can happen for any of three reasons. Remember that when KubeStellar requires its user to put an annotation or label on an object coming from a third party, this is a problem. One way to solve this would be to add a boolean in BindingPolicySpec that says to request template expansion on all matching objects. Another would be introduce another control object that has rules identifying the workload objects to be subject to template expansion. Another would be to use a feature of the facility importing the workload objects to add an annotation to all imported objects. In all of these, it is likely that the request paint would be applied with a broad brush.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: You can see the revision now.

pkg/transport/generic_transport_controller.go
nirrozenbaum: it was understood from your words that SIG-MC are doing that or going to do that.
if that was the case I'd say we can treat SIG-MC as a standard and align with that.

but after reading the last message I understand that is not the case.
I was trying to google quickly (didn't invest more than 5 minutes) and haven't found any docs about how SIG-MC are going to define properties of clusters. so I don't think at this point there is a standard for this thing and therefore we should do what's right for KS.

btw, on the other hand, the standard in kubernetes in general (not specifically in SIG-MC) for defining key-value configuration pairs is using configmaps. nothing prevents us from using labels + annotations but the standard is definitely configmaps.

I also don't agree with you previous comment:
> I was trying to follow Occam's Razor ("Entities must not be multiplied beyond necessity").
> 
> The ManagedCluster describes the WEC. I think that it is entirely reasonable to look for WEC properties in the labels and annotations of the ManagedCluster. Labels and annotations are open for any use, as far as the Kube metamodel is concerned. Also in this vein: my understanding of the direction that sig-multicluster is going in will see the self-describing cluster properties inside the WEC propagated to the core to appear in the inventory object.

Eventually we need to define key-value properties pairs somewhere. 
one option that you suggested is to use labels/annotations.
another option that I suggest is to use configmaps.
the properties are not multiplied. the fact that it's in a different object doesn't mean we multiply beyond necessity. there is a necessity to separate the things cause using inventory for defining customization parameters feels like a spaghetti code. 
separation of concerns is widely used in software engineering and it doesn't mean we multiply beyond necessity.

ezrasilvera: @MikeSpreitzer @nirrozenbaum My 2 cents:  My original vision was to have a "chain of config inputs" , similar to what is usually done with application configuration. When a key is found in a level the search is aborted and that value is used. 
This allows the possibility to override values, specify values for different criteria, etc..

We can think of multiple such levels also in our case:
1. CM matching a label
2. CM per BP
3. CM per MC
4. etc..
At this time I agree we can just have parameters per MC - but I agree I would personally use a different object and not the MC object itself, for several reasons:
1. Allows flexibility/override between WDSs - E.g., special patter for cluster identifier 
2. Security - we may not allow in the future setting fields on the MC object itself as it is managed by the inventory and may risk the whole environment
3. Clutter: What happen if we want to use many different keys?  (cluster-name, cluster-ip, etc..)

pkg/transport/generic_transport_controller.go
MikeSpreitzer: On second thought, while it is not incorrect to proceed with an arbitrary map, I think that it would indeed be better to abort the current work (since, as noted, it is already known to be moot).

pkg/transport/generic_transport_controller.go
MikeSpreitzer: @ezrasilvera: A layered or chained approach explicitly contemplates multiple layers and can start with having only one simple layer.

On further thought, I am less enthused about the user experience of using inventory object labels and annotations as the source of parameters. Now that we are using Go language templates, the parameter names have to be valid as identifiers in the Go language. Label and annotation names (keys) commonly have characters (`-`, `.`, `/`) that are not allowed in a Go identifier. We _could_ try to define mapping from labels + annotations to a data structure using only valid Go identifiers. E.g., `foo.com/bar=baz` maps to `map[string]map[string]map[string]string{"foo": map[string]map[string]string{"com": map[string]string{"bar": "baz"}}}`. That would be sort of doable. What would `foo=xyz,foo.bar=abc` map to?

Another possible approach would be to define a mapping of the three problematic characters to ones that are allowed in a Go identifier. Something like mapping `foo-bar.com/baz` to `foo🄼bar🄳com🅂bar`.

Both of these approaches mean that any change to any label or annotation of an inventory object would require re-creation of every ManifestWork that is sensitive to that inventory object's labels and annotations.

MikeSpreitzer:
The force-push to 983e3092a is a rebase onto main and adds a commit that switches to taking customization parameters from a ConfigMap and a default.

MikeSpreitzer:
The force-push to b79834139 does some doc fixup/cleanup.

api/control/v1alpha1/types.go
```diff
@@ -20,6 +20,34 @@ import (
...
+//
+// Note that this sort of customization has limited applicability.  It can only be used where
+// the un-expanded string passes the validation conditions of the relevant object type.
+// For more broadly applicable customization, see Customizer objects.
+
+const TemplateExpansionAnnotationKey string = "control.kubestellar.io/expand-templates"
+
+// ParameterConfigMapNamespace is the namespace in the ITS that holds ConfigMap objects that provide
+// parameter values for template expansion.
+const ParameterConfigMapNamespace = "customization-parameters"
```
nirrozenbaum, line 49: I didn't understand this definition. 
do you plan to use a single namespace to hold customization configmaps for all WECs? why?
if user A has permissions to customize WEC1 and user B has permissions to customize WEC2 - now both users can access both maps.
why aren't you using the WEC namespace?
each WEC namespace can hold inside it's customization configmap...

pkg/customize/customize.go
```diff
@@ -0,0 +1,104 @@
...
+		return typed
+	}
+}
+
+// expandString does parameter expansion on one string
+func (exp *expander) expandString(path, input string) string {
+	if !strings.Contains(input, "{{") {
+		return input
+	}
+	exp.wantedChange = true
```
nirrozenbaum, line 83: the documentation line says - 
`// wantedChange reports whether parameter expansion made any changes to the data`
do we know that it made changes to the data at this point? what happens if there was an error in parse/execute?

pkg/transport/generic_transport_controller.go
```diff
@@ -618,3 +913,24 @@ func cleanObject(object *unstructured.Unstructured) *unstructured.Unstructured {
...
+		return false
+	}
+	for key, val1 := range map1 {
+		val2, have := map2[key]
+		if val1 != val2 || !have {
+			return false
+		}
+	}
+	return true
+}
```
nirrozenbaum, line 936: seems like these functions are not used anywhere.. can we remove it?

pkg/transport/generic_transport_controller.go
```diff
@@ -410,44 +506,241 @@ func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, bind
...
+	if !visitedClusterName && oldProps["clusterName"] != destName {
+		return false
+	}
+	return newNumProperties == len(oldProps) && !mismatch
+}
+
+// customizeForDest customizes the given object for the given destination,
+// if any customization is called for. The returned boolean indicates whether
+// any customization was called for.
+func (c *genericTransportController) customizeForDest(object *unstructured.Unstructured, dest string, defs map[string]string) (*unstructured.Unstructured, []string, bool) {
```
nirrozenbaum, line 719: can you please use full names and not shortcuts? this is very confusing.
dest, defs..

api/control/v1alpha1/types.go
MikeSpreitzer: The mailbox namespaces are a private implementation detail, not part of the user interface.

pkg/transport/cmd/generic-main.go
```diff
@@ -101,14 +102,24 @@ func GenericMain(transportImplementation transport.Transport) {
...
+		logger.Error(err, "failed to construct transport controller")
+		klog.FlushAndExit(klog.ExitFlushTimeout, 1)
+	}
+
+	itsK8sInformerFactory.Start(ctx.Done())
 
 	// notice that there is no need to run Start method in a separate goroutine.
 	// Start method is non-blocking and runs each of the factory's informers in its own dedicated goroutine.
-	factory.Start(ctx.Done())
+	wdsKsInformerFactory.Start(ctx.Done())
```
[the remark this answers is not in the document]
MikeSpreitzer: I do not understand the question. There are now two informer factories, and there is a Start call for each of them.

pkg/customize/customize.go
MikeSpreitzer: Fixed the comment on that field.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: Yes, done.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: Made names longer.

I also made the terminology more consistent. I was previously using "properties" in some places, "parameters" in others. I have now tried to consistently use "properties".

ezrasilvera:
@MikeSpreitzer is it replacing 1760?  Should we close it?

MikeSpreitzer:
Yes, this is replacing 1760 if we continue to like all the fall-out of switching from simple parameter expansion to the ever more complicated template expansion.

MikeSpreitzer:
The force-push to 539227a7a is a rebase onto `main`.

docs/content/direct/user-guide.md
```diff
@@ -0,0 +1,75 @@
...
+
+Status of this document: it is the barest of a start. Much more needs to be written.
+
+This document is for users of a release. Examples of using the latest stable release are in [the examples document](examples.md). This document adds information not conveyed in the examples.
+
+## Using a pre-existing cluster as the hosting cluster
+
+See [Using an existing hosting cluster](./hosting-cluster.md)
+
+## When everything is not on the same machine
```
ezrasilvera, line 11: Do you mean same machine or same cluster ? 
What do you mean when you say "Everything"  ? Is it just spaces? what about having KF running on one cluster and KS on another?  How about the OTP plugin can it run on a different cluster? 
I think we should drop that section completely at this point and put Under Construction

docs/content/direct/user-guide.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Template Expansion
+
+Template expansion is an optional feature that a user can request on an object-by-object basis. The way to request this feature on an object is to put the following annotation on the object.
+
+```yaml
+    control.kubestellar.io/expand-templates: "true"
+```
+
+The customization that template expansion does when distributing an object from a WDS to a WEC is applied independently to each leaf string of the object and is based on the "text/template" standard pacakge of Go. The string is parsed as a template and then replaced with the result of expanding the template. Errors from this process are reported in the status field of the Binding object involved.
```
ezrasilvera, line 29: Do errors in customization fail the deployment? If so maybe we should explicitly mention it ?

docs/content/direct/user-guide.md
```diff
@@ -0,0 +1,75 @@
...
+
+The data used when expanding the template are parameters describing the WEC. These come from two sources.
```
ezrasilvera, line 31: maybe "parameter values" To emphasis (as oppose to the keys )

docs/content/direct/user-guide.md
```diff
@@ -0,0 +1,75 @@
...
+
+1. The ConfigMap object, if any, that is in the namespace named "customization-properties" in the ITS and has the same name as the inventory object for the WEC. In particular, the ConfigMap string and binary data items whose name is valid as a [Go language identifier](https://go.dev/ref/spec#Identifiers) supply parameters.
+1. If not otherwise defined, there is a parameter whose name is "clusterName" and whose value is the name of the inventory item for the WEC.
```
ezrasilvera, line 34: This is not clear to me. Only the "clusterName" parameter is supported?  I though every parameter that is not in the CM and does exist in the MC will be taken from there . 
It's also not clear where is it taken from, which object ?

pkg/transport/generic_transport_controller.go
```diff
@@ -228,6 +287,37 @@ func (c *genericTransportController) Run(ctx context.Context, workersCount int)
...
+func (c *genericTransportController) ensurePropertyNamespace(ctx context.Context) {
+	logger := klog.FromContext(ctx)
+	for {
+		_, err := c.itsNSClient.Get(ctx, v1alpha1.PropertyConfigMapNamespace, metav1.GetOptions{})
+		if err == nil {
+			logger.Info("Found property namespace already exists")
+			return
+		}
+		if !errors.IsNotFound(err) {
+			logger.Info("Failed to Get the property namespace", "err", err)
```
ezrasilvera, line 299: Why don't we return here ?

test/e2e/multi-cluster-deployment/use-kubestellar.sh
```diff
@@ -51,6 +51,9 @@ kind: Deployment
 metadata:
   name: nginx-deployment
   namespace: nginx
+  annotations:
```
ezrasilvera, line 54: Why did we decide that customization deserves to go into the shell testing?  Why not just add a ginkgo test for customization?

test/e2e/common/setup-kubestellar.sh
```diff
@@ -112,6 +112,8 @@ wait-for-cmd '(kubectl -n wds1-system wait --for=condition=Ready pod/$(kubectl -
 
 echo "transport controller is running."
 
+wait-for-cmd 'kubectl --context imbs1 get ns customization-properties'
```
ezrasilvera, line 115: Why did we decide that customization deserves to go into the shell testing? Why not just add a ginkgo test for customization?  
I think this customization should be removed from this test and moved to ginkgo

pkg/transport/generic_transport_controller.go
MikeSpreitzer: Because the retry loop is in this func.

docs/content/direct/user-guide.md
MikeSpreitzer: What do you mean by "every parameter that ... does exist in the MC"? Do you mean the labels and annotations?

Are you saying you are not sure what object is meant by the text "the inventory item for the WEC" ?  Would it help if that were expanded to "the inventory item for the WEC (i.e., the ManagedCluster object)"?

ezrasilvera: 1. Yes adding ManagedCluster object is perfect. Thanks
2. yes - I thought any "key" you use in the template , you first search for it in the CM and if not there you search over the labels and annotations in the ManagedCluster

docs/content/direct/user-guide.md
ezrasilvera: I meant if the customization fail, then the object is not deployed to the WEC(s) ?

docs/content/direct/user-guide.md
MikeSpreitzer: I mean same machine. By "everything" I mean the components in the KubeStellar architecture, everything in scope for this document.

Already in the examples based on `kind` clusters, there are KubeStellar components in different clusters: there is a hosting cluster, spaces created by KubeFlex running in that hosting cluster, and two WECs that are outside of the hosting cluster. This section is not about changing the architecture (e.g., the fact that the transport controller runs in the hosting cluster), it is about choices for using the existing architecture.

The existing architecture does not say that all the WECs have to run on the same machine as the hosting cluster. However, at the current level of development, it is possible to beyond a one-machine system only if the hosting cluster is OpenShift. This is worth explaining, even though it is disappointing. So I do not think this material should be deleted.

test/e2e/multi-cluster-deployment/use-kubestellar.sh
MikeSpreitzer: I chose to put it there because I expect this will be commonly needed.

MikeSpreitzer:
The force-push to a0df3d569 makes a couple of changes in response to @ezrasilvera 's review above.

docs/content/direct/user-guide.md
MikeSpreitzer: (A) oops, I see the old "parameter" language is still here. I changed it to "property" for consistency.

(B) I added a parenthetic remark to clarify. @nirrozenbaum will not like it because we expect generalization in the future. We will have to address this then.

docs/content/direct/user-guide.md
MikeSpreitzer: What sort of keys were you thinking of? I do not find that any sort of key comes to mind when reading a sentence about "parameters". I just changed that to "properties", for consistency.

docs/content/direct/user-guide.md
MikeSpreitzer: I added an explanation that I hope will answer this question.

docs/mkdocs.yml
```diff
@@ -36,6 +36,7 @@ nav:
       - Landing: index.md 
   - Overview:
       - Readme: readme.md
+      - User guide: user-guide.md
```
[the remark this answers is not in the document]
ezrasilvera: Fine with me for now.

test/e2e/multi-cluster-deployment/use-kubestellar.sh
ezrasilvera: I think having a specific test for that is better. I do agree we can add explicit section on customization in the user-guide which I think is what you did. 
I don't mind moving it out in future PR if you prefer not to spend time on that now, if so I'll open an issue for that.

docs/content/direct/user-guide.md
ezrasilvera: My question was a process one, I understand we update the status, do we even wrap an object if there was an error in customization ? (just to understand)

docs/content/direct/user-guide.md
ezrasilvera: I'm not sure all components can be indeed separated to different clusters (not by design but due to bugs) , but will not stop this PR due to this section as it's not really related to it.

docs/content/direct/user-guide.md
MikeSpreitzer: The broken YAML will go into the ManifestWork. The klusterlet will get an error from the apiserver when trying to create the broken object.

test/e2e/multi-cluster-deployment/use-kubestellar.sh
MikeSpreitzer: I think that adding a separate test just for template expansion is more costly to us contributors and more costly at runtime. I think it is better to fold that coverage into another test.

docs/content/direct/user-guide.md
MikeSpreitzer: (C) I added a commit that adds inventory object labels and annotations as sources of WEC properties.

docs/content/direct/user-guide.md
nirrozenbaum: IMO if we are aware of a failure before distribution, it's better to update the error in the status field without distributing, it's a waste of time and resources

docs/content/direct/user-guide.md
nirrozenbaum: @ezrasilvera what we agreed on when we discussed with the team was the start with config maps, without labels/annotations. even you agreed on this :)
I remember we even mentioned that hierarchy can be done using multiple config maps but as a starting point we start with a single configmap.
I understand the idea of hierarchy, but I have no doubt it complicates things. not only the implementation but also the users (not sure how intuitive is this direction for users). therefore the agreement was to start simple with only one configmap per WEC, and keep thinking about the hierarchy idea.

docs/content/direct/user-guide.md
MikeSpreitzer: I just pushed a revision that suppresses outward propagation when there are template expansion errors. See the update to the user guide, and/or to the code, for a precise statement.

MikeSpreitzer: However, I have realized that my answer about broken YAML was mostly misleading. Remember that template expansion is applied independently to each leaf string. So the higher level structure is certainly unchanged; at worst, the output from a troubled expansion will fail validation for that particular string field. While possible, I think that is practically very unlikely --- remember that the un-expanded template string passed validation.

docs/content/direct/user-guide.md
ezrasilvera: @nirrozenbaum You are correct. I have no problem with that .   For some reason I was under the wrong assumption that we were adding CM and also reading from MC inventory objects. 
I agree it's probably also not good to count on MC that may be changed in the future when we introduce new inventory.  
I still think it would be good (maybe future PR ? ) if KS can create the initial CM with some default properties (e.g., ClusterName)

ezrasilvera: @nirrozenbaum @MikeSpreitzer Sorry for the mess :(   That's what happen when someone tries to review without full awareness of the already resolved comments/discussions

MikeSpreitzer:
The recent force-pushes are a rebase onto `main` and a fix of a locking bug (holding only the read lock while doing a write).

MikeSpreitzer:
The force-push to f7d6cdf22 is a rebase onto `main`.

pkg/transport/generic_transport_controller.go
```diff
@@ -146,40 +211,60 @@ func getGvrFromWrappedObject(clientset kubernetes.Interface, wrappedObject runti
...
-	wdsClientset     ksclientset.Interface
 	wdsDynamicClient dynamic.Interface
 	wdsName          string
+
+	propsMutex sync.Mutex
+
+	// bindingSensitiveDestinations maps Binding name to the set of destinations whose properties the Binding is senstive to.
+	// Access to both the map and the Sets it holds is controlled by the RWMutex.
+	// The sets are mutable with the RWMutex held.
+	bindingSensitiveDestinations map[string]sets.Set[v1alpha1.Destination]
```
vMaroon, line 251: Why not use utils.ConcurrentMap?

pkg/transport/generic_transport_controller.go
```diff
@@ -146,40 +211,60 @@ func getGvrFromWrappedObject(clientset kubernetes.Interface, wrappedObject runti
...
+
+	// destinationProperties maps a destination to the properties to use for it in template expansion.
+	// Access only while holding RWMutex and keep consistent with bindingSensitiveDestinations.
+	// An entry is removed from this map when this controller is notified of
+	// deletion of the destination's property ConfigMap.
+	// Every `clusterProperties` that appears here is immutable from the time that it arrived.
+	destinationProperties map[v1alpha1.Destination]clusterProperties
```
vMaroon, line 258: Same as above, don't have to answer both.

pkg/transport/generic_transport_controller.go
```diff
@@ -416,46 +580,225 @@ func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, bind
 	return objectsToPropagate, nil
 }
 
-func (c *genericTransportController) initializeWrappedObject(ctx context.Context, binding *v1alpha1.Binding) (*unstructured.Unstructured, error) {
+func (c *genericTransportController) initializeWrappedObject(ctx context.Context, binding *v1alpha1.Binding) (func(v1alpha1.Destination) (*unstructured.Unstructured, bool), []string, error) {
```
vMaroon, line 583: This function is too packed and hard to trace.
Can you outline its functional flow in compact description, while splitting it into smaller functions that separate concerns? I think this would not only make this function clearer, but as it seems to be a central function, its increased clarity would boost understanding the rest of the code.

For example, if I didn't miss any functionality in the flow:
1. `getObjectsToPropagate`: retrieves the objects from WDS based on the binding
2. `processObjectsForCustomization`: examines each object to determine if it needs customization based on its annotations, and processes accordingly
3. (already in code) `customizeForDestination `: performs the customization for an individual object
4. `appendObjectToDestination`: appends an object to the corresponding destination in the destToCustomizedObjects map while ensuring that objects are correctly grouped by their destination
5. `updateBindingSensitivities`: updates the binding sensitivities based on the destinations involved in the customization
6. `prepareWrappedObjects`: prepares the final wrapped objects for each destination
7. finally, `wrapObjects`: converts objects into their unstructured form

This is the flow I could pick up from the function. Anything that achieves separation of concerns and makes the function easier to process is ok.

MikeSpreitzer:
The recent force-pushes are squashes and rebase onto `main`.

docs/content/direct/user-guide.md
```diff
@@ -0,0 +1,81 @@
...
+1. The labels of the inventory item for the WEC supply properties if the label's name (AKA key) is valid as a Go language identifier.
+1. There is a pre-defined property whose name is "clusterName" and whose value is the name of the inventory item (i.e., the ManagedCluster object) for the WEC.
+
+A Binding object's `status` section has a field holding a slice of error message strings reporting user errors that arose the last time the transport controller processed that Binding, along with the `observedGeneration` reporting the `metadata.generation` that was processed. For each workload object that the Binding references: if template expansion reports errors for any destinations, the errors reported for the first such destination are included in the Binding object's status.
+
+Any failure in any template expansion for a given Binding suppresses propagation of desired state from that Binding; the previosly propagated desired state from that Binding, if any, remains in place in the WEC.
+
+Template expansion can only be applied when and where the un-expanded leaf strings pass the validation that the WDS applies, and can only epxress substring replacements.
```
ezrasilvera, line 42: Not very clear. Also do you define what is a "leaf" before this place ?

pkg/abstract/slices.go
```diff
@@ -25,3 +25,23 @@ func SliceDelete[Elt any](slice *[]Elt, index int) {
 	}
 	*slice = (*slice)[:lastIndex]
 }
+
+// SliceCopy copies a given slice into new storage
```
ezrasilvera, line 29: Isn't this actual a CreateFrom rather than copy ?

pkg/abstract/slices.go
```diff
@@ -25,3 +25,23 @@ func SliceDelete[Elt any](slice *[]Elt, index int) {
...
+func SliceCopy[Elt any](input []Elt) []Elt {
```
ezrasilvera, line 30: Isn't this actual a CreateFrom rather than copy ?

pkg/abstract/slices.go
```diff
@@ -25,3 +25,23 @@ func SliceDelete[Elt any](slice *[]Elt, index int) {
...
+	if input == nil {
+		return nil
+	}
+	return append(make([]Elt, 0, len(input)), input...)
+}
+
+func SliceEqual[Elt comparable](slice1, slice2 []Elt) bool {
```
ezrasilvera, line 37: Why do we need the Slice prefix?  Isn't IsEqual  while the parameters are slices good enough ?

ezrasilvera:
@vMaroon   This PR seems OK to me (I have some additional comments about the docs but I can delay those and simply push PRs to change ).   Please review @MikeSpreitzer answers when you have time so we can converge her. Thanks

ezrasilvera:
@MikeSpreitzer Not blocker for the PR - But it would be very helpful if you can add a diagram to the arch doc illustrating the details of the customization including the relations between the data structures you use here. 
In general we agreed to have a design doc with low level details before we push a PR for such features, so we could have just copied from there. However for this PR we didn't follow those rules, so at least, if possible let's  add it with/after  the PR

pkg/abstract/slices.go
MikeSpreitzer: We have both "SliceEqual" and "PrimitiveMapEqual".

pkg/abstract/slices.go
MikeSpreitzer: In plain English, "copy" connotes making a new thing. When I copy a sculpture, I make a new one that looks like an existing one. It is true that Unix & C introduced things like "memcpy", which writes the contents of one block of memory into another pre-existing block of memory, and Go has a built-in "copy" that does something similar. But SliceCopy here is unlike those, in that it does not take two slices as input, it takes just one and returns a new slice (new memory). So the English word "copy" is quite appropriate here.

docs/content/direct/user-guide.md
MikeSpreitzer: I rely on the general understanding of "leaf" as the things in a tree that have no subdivisions in the tree structure. I think readers can be relied on to understand that YAML and JSON documents have tree structure.

docs/content/direct/user-guide.md
```diff
@@ -0,0 +1,81 @@
...
+
+## When everything is not on the same machine
+
+Thus far we can only say how to handle this when the hosting cluster is OpenShift. The problem is getting URLs that work from everywhere. OpenShift is a hosted product, your clusters have domain names that are resolvable from everywhere. In other words, if you use an OpenShift cluster as your hosting cluster then this problem is already solved.
+
+## Rule-based customization
+
+KubeStellar can distribute one workload object to multiple WECs, and it is common for users to need some customization to each WEC. By _rule based_ we mean that the customization is not expressed via one or more literal expressions but rather can refer to _properties_ of each WEC by property name. As KubeStellar distributes or transports a workload object from WDS to a WEC, the object can be transformed in a way that depends on those properties.
+
+At its current level of development, KubeStellar has a simple but limited way to specify rule-based customization, called "template expansion".
```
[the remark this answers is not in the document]
MikeSpreitzer: This **is** a statement of what we support, along with some adjectives that call out aspects of that thing that are important for the reader to be clearly aware of.

MikeSpreitzer:
@ezrasilvera, regarding architecture documentation: OK, I will keep that in mind when I make the PR that updates that doc.

pkg/transport/generic_transport_controller.go
vMaroon: It's an option to expand the interface implementation with a map that shares a mutex with another. In all cases this is not a blocker as it is a personal preference, I can just PR it later if you approve.

docs/content/direct/user-guide.md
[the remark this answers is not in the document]
MikeSpreitzer: For example, a reader needs to clearly understand that the Go template expansion in KubeStellar is very different from the Go template expansion in a Helm chart. In the Helm case, the Go template is the whole YAML source, while in KubeStellar, the templates are the leaf strings.

Issue #1951, opened 2024-03-22
bug: test/e2e/ginkgo issues


MikeSpreitzer:
### Describe the bug

(1) This test is flaky in CI. Sometimes it fails, for no obvious reason, and upon retry it succeeds.

(2) The README gives no clue that setting KUBECONFIG is _necessary_.

(3) Setting KUBECONFIG _is necessary_.

(4) The workflow has too little debug output to debug problems.

### Steps To Reproduce

For (1) and (4), look at a workflow run that failed.

For (2), read the README.

For (3), try invoking ginkgo _without_ setting KUBECONFIG.

### Expected Behavior

Non-flaky tests.
README does not waste the user's time rediscovering a known mystery.
It should not be necessary to set an environment variable to its default value.
It should be possible to debug failed runs.

MikeSpreitzer:
I wonder if a cause of the flakiness is the fact that https://github.com/kubestellar/kubestellar/blob/ee686b085ace011ac9c7c5ad8a9c387c880dcc99/test/util/util.go#L390-L401 _starts_ the process of a Pod being removed but does not wait for the completion of that process.

MikeSpreitzer:
There are definitely problems with the resilience tests not recognizing what recovery looks like. With multiple resilience tests in series, later ones are failing because the system has not fully recovered from earlier ones.

MikeSpreitzer:
Or, put another way, every bit of test code that sends a request should actually use a retry loop.

MikeSpreitzer:
This test is still flaky.  See the failure in try 1 at https://github.com/kubestellar/kubestellar/actions/runs/8448139484/attempts/1 followed by a successful retry.

effi-ofer:
> This test is still flaky. See the failure in try 1 at https://github.com/kubestellar/kubestellar/actions/runs/8448139484/attempts/1 followed by a successful retry.

@MikeSpreitzer Are we waiting long enough?  The logging of the failure seems to indicate we are only waiting 1min.  
```
/home/runner/work/kubestellar/kubestellar/test/e2e/ginkgo/multiple_cluster_deployment_test.go:35 @ 03/27/24 **07:40:33.1**
  [FAILED] Unexpected error:
      <*errors.StatusError | 0xc000030000>: 
      the server was unable to return a response in the time allotted, but may still be processing the request (get deployments.apps)
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
[...]

PR #1957, opened 2024-03-25
🐛 Fixup resiliency testing

MikeSpreitzer:
## Summary
The PR makes various improvements and fixes in resilience testing.

## Related issue(s)

This is intended to causes of flakiness that I know about now. The flakiness is part of #1951 .

### DIFF
--- a/.github/workflows/pr-test-e2e.yml
+++ b/.github/workflows/pr-test-e2e.yml
@@ -69,8 +69,9 @@ jobs:
       - name: show pods in hosting cluster
         if: always()
         run: |
+          date
           kubectl --context kind-kubeflex get pods -A
-          kubectl --context kind-kubeflex get pods -A | grep -vw Running | grep -vw Completed | grep -v NAME | while read ns name rest; do echo; kubectl --context kind-kubeflex describe pod -n $ns $name; kubectl --context kind-kubeflex logs -n $ns $name || true; done
+          kubectl --context kind-kubeflex get pods -A | grep -vw Running | grep -vw Completed | grep -v NAME | while read ns name rest; do echo; kubectl --context kind-kubeflex describe pod -n $ns $name; echo; kubectl --context kind-kubeflex logs -p -n $ns $name || true; echo; kubectl --context kind-kubeflex logs -n $ns $name || true; done
 
       - name: show Deployment objects in hosting cluster
         if: always()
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-test-e2e.yml
```diff
@@ -68,7 +68,14 @@ jobs:
...
         if: always()
-        run: kubectl --context kind-kubeflex get pods -A
+        run: |
+          date
+          kubectl --context kind-kubeflex get pods -A
+          kubectl --context kind-kubeflex get pods -A | grep -vw Running | grep -vw Completed | grep -v NAME | while read ns name rest; do echo; kubectl --context kind-kubeflex describe pod -n $ns $name; echo; kubectl --context kind-kubeflex logs -p -n $ns $name || true; echo; kubectl --context kind-kubeflex logs -n $ns $name || true; done
+
+      - name: show Deployment objects in hosting cluster
+        if: always()
+        run: kubectl --context kind-kubeflex get deployments -A
```
effi-ofer, line 78: Won't the logs of all the pods here make redundant the logs we do in subsequent sections, like ' show kubestellar controller manager log' ?

MikeSpreitzer: Right now we have an overlap, which is better than a gap. Neither set covers the other.
We have some logging for every pod that is neither Running nor Completed.
We have some logging for a few pods regardless of their phase.

effi-ofer:
> > util.DeletePods(ctx, coreCluster, "kubeflex-system", "")
> 
> This line all by itself has inherent flakiness: there is no guarantee of the order in which the pods are deleted, and if the postgress pod is deleted first then the other deletions may fail.
> 
> But I do not think that I have observed this failure mode.

Hey @MikeSpreitzer I am going to LGTM this PR, but I have a question about the above comment.  Does the order in which pods are killed makes a difference?  How?  In real life one pod can be killed before another and if this somehow breaks us, then this is something we should resolve?  Or do you mean the test itself is broken somehow?

effi-ofer (approved):

MikeSpreitzer:
@effi-ofer, regarding my comment about `util.DeletePods(ctx, coreCluster, "kubeflex-system", "")`: The problem was not the mere fact that multiple pods are deleted, nor the fact that the order of deletion is not guaranteed, but the combination of those with not using a retry loop. This PR fixed that.

PR #1959, opened 2024-03-25
:book: Describe how to deploy KubeStellar on K3D-based hub and wec clusters

eaepstein:
## Summary
Initial version. Would be a bit simpler with work on other components.
## Related issue(s)

Fixes #


ezrasilvera:
@eaepstein Let's merge it now as a separate file, and then later on add a link once we put a "integration examples"

PR #1969, opened 2024-03-27
🐛 remove CRDs from Helm chart

### LINKED ISSUE #1965: bug: e2e tests fail when running on the released v0.21.1
### Describe the bug

Running the e2e tests results in the following failure:
```
kubectl --context kind-kubeflex logs -n wds1-system kubestellar-rbqz7
Pulled: ghcr.io/kubestellar/kubestellar/controller-manager-chart:0.21.1
Digest: sha256:079918338a24c7863d887b2b205ad906416e747ddb4c39c59e29b463eb6b9ee9
Error: failed to install CRD crds/control.kubestellar.io_bindingpolicies.yaml: 1 error occurred:
        * customresourcedefinitions.apiextensions.k8s.io is forbidden: User "system:serviceaccount:wds1-system:default" cannot create resource "customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope
```

The very first visible failure is:
```
+ : Run OCM transport controller in a pod
  + :
  + cd /home/effi/edge-mc/test/e2e/common/../../..
  ++ pwd
  + KUBESTELLAR_DIR=/home/effi/edge-mc
  + OCM_TRANSPORT_PLUGIN_RELEASE=0.1.0
  + curl -sL https://github.com/kubestellar/ocm-transport-plugin/archive/refs/tags/v0.1.0.tar.gz
  + tar xz
  + cd ocm-transport-plugin-0.1.0
  ++ pwd
  + OCM_TRANSPORT_PLUGIN_DIR=/home/effi/edge-mc/ocm-transport-plugin-0.1.0
  + pwd
  + echo 'replace github.com/kubestellar/kubestellar => /home/effi/edge-mc/'
  + go mod tidy
  + IMAGE_TAG=0.1.0
  + make ko-build-local
  2024/03/27 13:35:45 Using base cgr.dev/chainguard/static:latest@sha256:739aaf25ce9c6ba75c3752d7fde4a94de386c6f44eba27239d2b98e91752e3ff for github.com/kubestellar/ocm-transport-plugin/cmd/transport-controller
  2024/03/27 13:35:46 Building github.com/kubestellar/ocm-transport-plugin/cmd/transport-controller for linux/amd64
  2024/03/27 13:35:49 Loading ko.local/transport-controller:9647b62ca785663b88d61390218bda6a5c870702389611b8fdc1c89024458728
  2024/03/27 13:35:50 Loaded ko.local/transport-controller:9647b62ca785663b88d61390218bda6a5c870702389611b8fdc1c89024458728
  2024/03/27 13:35:50 Adding tag 0.1.0
  2024/03/27 13:35:50 Added tag 0.1.0
  + kind load --name kubeflex docker-image ko.local/transport-controller:0.1.0
  Image: "" with ID "sha256:58e00f9434b4c52adbbb9b44bf3e0a2f3dd8b45b897b9e0b3c55184ad27dcbef" not yet present on node "kubeflex-control-plane", loading...
  + cd /home/effi/edge-mc
  + pwd
  + rm -rf /home/effi/edge-mc/ocm-transport-plugin-0.1.0
  + echo 'running ocm transport plugin...'
  + kubectl config use-context kind-kubeflex
  + IMAGE_PULL_POLICY=Never
  + ./scripts/deploy-transport-controller.sh wds1 imbs1 ko.local/transport-controller:0.1.0
  + wait-for-cmd '(kubectl -n wds1-system wait --for=condition=Ready pod/$(kubectl -n wds1-system get pods -l name=transport-controller -o jsonpath={.items[0].metadata.name}))'
  + cmd='(kubectl -n wds1-system wait --for=condition=Ready pod/$(kubectl -n wds1-system get pods -l name=transport-controller -o jsonpath={.items[0].metadata.name}))'
  + wait_counter=0
  + eval '(kubectl -n wds1-system wait --for=condition=Ready pod/$(kubectl -n wds1-system get pods -l name=transport-controller -o jsonpath={.items[0].metadata.name}))'
  +++ kubectl -n wds1-system get pods -l name=transport-controller -o 'jsonpath={.items[0].metadata.name}'
  error: error executing jsonpath "{.items[0].metadata.name}": Error executing template: array index out of bounds: index 0, length 0. Printing more information for debugging the template:
        template was:
                {.items[0].metadata.name}
        object given to jsonpath engine was:
                map[string]interface {}{"apiVersion":"v1", "items":[]interface {}{}, "kind":"List", "metadata":map[string]interface {}{"resourceVersion":""}}
```

### Steps To Reproduce

From the test/e2e/ginkgo directory, issue: ginkgo -v -- -released

### Expected Behavior

Tests pass.

### Additional Context

_No response_

francostellari:
## Summary

Fix regression introduced by #1920  by removing the CRDs from the Helm chart

## Related issue(s)

Fixes #1965

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./cmd/controller-manager/main.go $(ARGS)
 
 .PHONY: ko-build-local
-ko-build-local: ## Build local container image with ko 
+ko-build-local: ## Build local container image with ko
 	$(shell (docker version | { ! grep -qi podman; } ) || echo "DOCKER_HOST=unix://$$HOME/.local/share/containers/podman/machine/qemu/podman.sock ") KO_DOCKER_REPO=ko.local ko build -B ./cmd/${CMD_NAME} -t ${IMAGE_TAG} --platform linux/${ARCH}
 	docker tag ko.local/${CMD_NAME}:${IMAGE_TAG} ${IMAGE_REPO}:${IMAGE_TAG}
 
@@ -184,8 +184,6 @@ chart: manifests kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(shell echo ${IMG} | sed 's/\(:.*\)v/\1/')
 	$(KUSTOMIZE) build config/default > chart/templates/controller-manager.yaml
 	scripts/add-helm-code.sh add
-	@mkdir -p chart/crds
-	@cp config/crd/bases/*.yaml chart/crds/
 
 
 .PHONY: local-chart
@@ -198,8 +196,6 @@ endif
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(shell echo ${IMG} | sed 's/\(:.*\)v/\1/')
 	$(KUSTOMIZE) build config/default > local-chart/templates/controller-manager.yaml
 	scripts/add-helm-code.sh --dir ${PWD}/local-chart add
-	@mkdir -p local-chart/crds
-	@cp config/crd/bases/*.yaml local-chart/crds/
 	git checkout -- config/manager/kustomization.yaml
 
 ##@ Deployment


config/default/kustomization.yaml
```diff
@@ -15,6 +15,7 @@ namePrefix: kubestellar-
 #    someName: someValue
 
 resources:
+- ../crd
```
pdettori, line 18: To avoid more confusion in the future I think we should remove that line 18 (the crds are not incorporated in the crd-level kustomization file, but it should be easier to understand if we disable from here)

francostellari: ok, done

MikeSpreitzer: but not pushed

MikeSpreitzer:
Where (besides a long and private and ephemeral Slack discussion) is the description of what was going wrong and why this is a fix and why this fix does not itself break something?

MikeSpreitzer:
From @pdettori:
> The reason CRD [should not be] installed by the helm chart is that when we use the WDS of type k8s, the CRD needs to be installed in the WDS, not in the hosting cluster. But the helm chart installs the deployment and RBAC in the hosting cluster, and the same helm chart cannot install something in the hosting cluster and something else in the WDS. Thus it is the ks controller manager that installs the CRDs. CRDs are copied in https://github.com/kubestellar/kubestellar/tree/release-0.21.0/pkg/crd/files and installed by go code running in the ksc controller manager in  the WDS to which is connected to.

pdettori:
@MikeSpreitzer the change that @francostellari did and that is being reverted here was to copy the CRDs into the chart. Before we did not have the CRDs in the chart, because they are installed at startup by the WDS controller manager in the associated WDS. The RBAC associated with the helm chart job defined in the PCH for kubestellar does not have rules for giving the job privileges to install the helm charts in the hosting cluster, thus the helm install fails.

config/default/kustomization.yaml
pdettori: @MikeSpreitzer it was already excluded, this PR was putting it back, but since @francostellari took it out now it does not show in the changed files - see https://github.com/kubestellar/kubestellar/blob/a7d1e58feebd3d661811c3a1dc93f46fa923e5fc/config/default/kustomization.yaml

MikeSpreitzer: Thanks for the explanation.
/unhold

PR #1971, opened 2024-03-27
:sparkles:  A simple helper script to update the version number

ezrasilvera:
## Related issue(s)

Fixes #

### DIFF
--- a/scripts/update-release.sh
+++ b/scripts/update-release.sh
@@ -23,8 +23,8 @@
 # Input version should be X.Y.Z
 input_version="$1"
 
-if ! [[ "$1" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-    echo "Error: Invalid version format. Please provide a version in the format X.Y.Z (all numbers)." >&2
+if ! [[ "$1" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-rc[0-9]+)?$ ]]; then
+    echo "Error: Invalid version format. Please provide a version in the format X.Y.Z[-RCn] (X, Y, Z, n are all numbers). " >&2
     exit 1
 fi
 
--- a/scripts/update-release.sh
+++ b/scripts/update-release.sh
@@ -1,5 +1,25 @@
 #!/bin/bash
 
+# Copyright 2023 The KubeStellar Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


pdettori (commented):
/lgtm

scripts/update-release.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Input version should be X.Y.Z
```
MikeSpreitzer, line 3: No support for X.Y.Z-rc1?

scripts/update-release.sh
```diff
@@ -0,0 +1,17 @@
...
+    exit 1
+fi
+
+# Replace version string using sed
+
+sed -i "/- --version/{n;s/- \"[0-9]\+\.[0-9]\+\.[0-9]\+\"/- \"$input_version\"/}" config/postcreate-hooks/kubestellar.yaml
+
+sed -i "s/export KUBESTELLAR_VERSION=[0-9]\+\.[0-9]\+\.[0-9]\+/export KUBESTELLAR_VERSION=$input_version/" docs/content/direct/examples.md
+
+sed -i "/The latest release is/s/[0-9]\+\.[0-9]\+\.[0-9]\+/$input_version/g" docs/content/direct/README.md
```
MikeSpreitzer, line 17: The wording being edited here is very interim. The plan is to focus this doc on saying what is the latest stable release. But we can not do that until we have one.

MikeSpreitzer:
The good news is that this is not a doomsday machine, so keeping it a secret is not so bad.

scripts/update-release.sh
ezrasilvera: Good point.

ezrasilvera: @MikeSpreitzer do we need that for RC?  docs are not uploaded for RC and also the readme is not changed (as the RC is not the "latest version" we want to point to

scripts/update-release.sh
ezrasilvera: I used what we have now.

scripts/update-release.sh
MikeSpreitzer: https://github.com/kubestellar/kubestellar/blob/main/docs/content/direct/release.md#step-by-step makes no exception for test releases. The editing of the README remark about latest release needs some nuanced handling.

ezrasilvera: Done. -rcX is supported as well

effi-ofer: You should update the comment text to reflect this change.  Can be done separately.

effi-ofer (approved):

PR #1974, opened 2024-03-27
🐛 Make CleanupWDS use retry loops

MikeSpreitzer:
## Summary
This PR revises the CleanupWDS func used in the ginkgo-based E2E test, to use retry loops around all calls to the apiserver. This is good because it brings the test code up to par with normal kube client conventions.

## Related issue(s)

The addresses one source of the flakiness noted in #1951

### DIFF
--- a/.github/workflows/pr-test-e2e.yml
+++ b/.github/workflows/pr-test-e2e.yml
@@ -71,7 +71,13 @@ jobs:
         run: |
           date
           kubectl --context kind-kubeflex get pods -A
-          kubectl --context kind-kubeflex get pods -A | grep -vw Running | grep -vw Completed | grep -v NAME | while read ns name rest; do echo; kubectl --context kind-kubeflex describe pod -n $ns $name; echo; kubectl --context kind-kubeflex logs -p -n $ns $name || true; echo; kubectl --context kind-kubeflex logs -n $ns $name || true; done
+          kubectl --context kind-kubeflex get pods -A | grep -vw Running | grep -vw Completed | grep -v NAME | while read ns name rest; do echo; kubectl --context kind-kubeflex describe pod -n $ns $name; echo; kubectl --context kind-kubeflex logs -n $ns $name || true; done
+
+      - name: show previous logs in hosting cluster
+        if: always()
+        run: |
+          date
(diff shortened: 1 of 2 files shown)


ezrasilvera:
@MikeSpreitzer wouldn't it be easier (and cleaner) to just remove the WDS and re-create it (instead of trying to clean up the resources inside it)  ?  Creating WDS should be relatively fast. I also see that you clean pre-define list of things, what happen if future tests uses new/other resource types in WDS?

MikeSpreitzer:
@ezrasilvera : I presume the reason that the original author of the resilience testing chose to take down pods is that pod deletion and replacement is something that can really happen and we want to know that the code reacts to this properly.

MikeSpreitzer:
/cc @effi-ofer 
/cc @pdettori 
/cc @ezrasilvera 
/cc @waltforme 

I think that I have got the test to be robust. However, after the test is done the state dumping consistently shows the kube-controller-manager to be troubled. But I think that is a separate issue. Not necessarily an unimportant one! But separate.

@pdettori , look at try number 2 in https://github.com/kubestellar/kubestellar/actions/runs/8458927359/job/23180745898 . The kube-controller-manager is reported to have failed. You can see the timing of two runs, and the log from the last one (I do not understand why the log from the previous run was uavailable). The evidence shows only the last incarnation of the postgres pod, which is very young. The kube-controller-manager crashed right around the time of the last start of the postgres pod.

ezrasilvera:
> @ezrasilvera : I presume the reason that the original author of the resilience testing chose to take down pods is that pod deletion and replacement is something that can really happen and we want to know that the system reacts to this properly.

@MikeSpreitzer The CleanupWDS  is called in the BeforeEach  of all e2e so it's not just resiliency.   We have a special test for resiliency where individual Pods are removed. I think that for e2e tests it's essential we start from guaranteed clean environment and this is better achieved by simply re-create the WDS.

ezrasilvera:
> @ezrasilvera, regarding your remark "I think that for e2e tests it's essential we start from guaranteed clean environment and this is better achieved by simply re-create the WDS": I am not convinced that is essential. I _am_ convinced that it would make the test series take a lot longer to run. Can we pursue this in follow-on work?

1. It shouldn't take more time as the k8s-WDS suppose to be very fast
2. I'm OK with delaying such changes to future work

PR #1988, opened 2024-03-29
🐛 Fix triggering problems in doc workflows

### LINKED ISSUE #1986: bug: problems with broken links crawler workflow
### Describe the bug

(A) I tried to apply it to the current contents of the branch named `main`, and nothing happened. See https://github.com/kubestellar/kubestellar/actions/runs/8484373434 . Perhaps it is because the condition at https://github.com/kubestellar/kubestellar/blob/fc7cebf2f21ffa640064e021b9604a5226af97bf/.github/workflows/broken-links-crawler.yml#L20 fails when the workflow is manually dispatched.

(B) Currently the latest run is https://github.com/kubestellar/kubestellar/actions/runs/8484036554, which I think happened after the conclusion of the latest run of the build-and-push workflow. But the crawler issued complaints that seem invalid. For example, the crawler complained:

> Error: ClientResponseError: 404 - https://docs.kubestellar.io/main/direct/examples/user-guide.md found by navigating through: https://docs.kubestellar.io/main/ -> https://docs.kubestellar.io/main/direct/examples/

but I can not find a link to "anything/user-guide.md" when I view the source of https://docs.kubestellar.io/main/direct/examples/ . For example:

```
mspreitz@mjs13 kubestellar % curl https://docs.kubestellar.io/main/direct/examples/ | grep user-guide
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  111k  100  111k    0     0   952k      0 --:--:-- --:--:-- --:--:--  960k
mspreitz@mjs13 kubestellar % echo $?
1
```

### Steps To Reproduce

See above

### Expected Behavior

(A) The workflow can be manually dispatched.

(B) The right errors are reported.

### Additional Context

_No response_

### LINKED ISSUE #1987: bug: "Generate and push docs" workflow runs uselessly on PRs
### Describe the bug

https://github.com/kubestellar/kubestellar/blob/fc7cebf2f21ffa640064e021b9604a5226af97bf/.github/workflows/docs-gen-and-push.yml triggers on PR activity, and has no effect (indeed it would be wrong to update the website based on creating or updating a PR). It _would_ be desirable if we had a workflow that created a preview somewhere when there is PR activity, but as far as I know our current website technology can not do that.

You can see in https://github.com/kubestellar/kubestellar/actions/workflows/docs-gen-and-push.yml many runs that are based on PR create or update.  For example, https://github.com/kubestellar/kubestellar/actions/runs/8483842486

### Steps To Reproduce

See above.

### Expected Behavior

No useless runs.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR makes the broken links crawler do something when invoked manually. This PR also stops the docs-gen-and-push workflow from being triggered by mere PR activity.

I have tested the former using `main` in my fork. I am unable to test the latter because it has a condition that restricts it to only running on the shared repo.

## Related issue(s)

Fixes #1986 
Fixes #1987

### DIFF
--- a/.github/workflows/broken-links-crawler.yml
+++ b/.github/workflows/broken-links-crawler.yml
@@ -14,10 +14,17 @@ permissions:
   contents: write
 
 jobs:
+  debug-event:
+    name: debug-event-contents
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "event name is:" ${{ github.event_name }} 
+      - run: echo "event type is:" ${{ github.event.action }} 
+
   broken-links-crawler:
     name: broken-links-crawler
     runs-on: ubuntu-latest
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
     steps:
       - name: get workflow_dispatch branch name
         shell: bash
--- a/.github/workflows/docs-gen-and-push.yml
+++ b/.github/workflows/docs-gen-and-push.yml
@@ -5,14 +5,6 @@ on:
   # So we can trigger manually if needed
   workflow_dispatch:
   # To confirm any changes to docs build successfully, without deploying them
-  pull_request_target:
-    branches:
-      - main
-      - "release-*"
-    paths:
-      - "docs/**"
-      - "*.md"
-      - ".github/workflows/docs-gen-and-push.yaml"
   push:
     branches:
       - main


.github/workflows/broken-links-crawler.yml
```diff
@@ -14,10 +14,17 @@ permissions:
   contents: write
 
 jobs:
+  debug-event:
```
ezrasilvera, line 17: Do you want to leave it here, or was it just  used during some debugging ?

ezrasilvera:
@MikeSpreitzer while you are at these scripts maybe we can re-visit the push action as well. 
I think that due to the path filter we may not generate docs for a new release in case not file was changed in those directories. 

  push:
    branches:
      - main
      - "release-*"
    paths:
      - "docs/**"
      - "*.md"
      - ".github/workflows/docs-gen-and-push.yaml"

.github/workflows/broken-links-crawler.yml
MikeSpreitzer: I would like to leave it there for a while, to gather evidence, eventually remove.

MikeSpreitzer:
@ezrasilvera: can you illustrate a scenario in which the paths filter prevents a desired run?

MikeSpreitzer:
@ezrasilvera: I tried a test of a normal scenario, using the `main` branch of my fork, and the triggering did not go wrong. You can look at https://github.com/MikeSpreitzer/kcp-edge-mc/tree/main now, I am leaving my `main` in this unusual state so that you can see it. You will see, in order from older to later:
- The commit in this PR.
- A commit that removes the restriction that the docs-gen-and-push.yml workflow only runs on the shared repo.
- A merge of a PR (https://github.com/MikeSpreitzer/kcp-edge-mc/pull/4) that has two commits: an older one that makes a doc change and a newer one that makes a code change.

You will see that the merge of PR 4 _did_ trigger the docs-gen-and-push.yml workflow (which failed as expected, my fork is not set up to do those privileged things).

ezrasilvera:
> @ezrasilvera: can you illustrate a scenario in which the paths filter prevents a desired run?

Are we allowing to amend a release branch ?  (push commits to existing release branch? ). 
If we don't allow it - then I guess the paths are not effecting the trigger for a release branch as the initial push always contains the docs as well

ezrasilvera:
> The idea illustrated here is that merging a PR makes a change to all the files affected by that PR, not just the last commit in that PR. The merge commit is a combination of all the commits being contributed.

That's strange.  The paths should explicitly restrict the trigger to PRs containing files in those paths.  Are you saying that the PR was triggered for PRs without such files?

MikeSpreitzer:
I would not be opposed to amending a release branch to improve the documentation of that release.

>  The paths should explicitly restrict the trigger to PRs containing files in those paths.  Are you saying that the PR was triggered for PRs without such files?

I am not sure I follow what you mean. A PR refers to a branch, which points to a commit, and a commit holds a snapshot of all the files in the repo. A PR is a request to merge some changes: the changes in the PR's branch (`fix-doc-workflows` in my fork, for the PR at hand) since the latest common ancestor (commit fc7cebf2f21ffa640064e021b9604a5226af97bf in this case) with the PR's base branch (`main` in the shared repo, in this case).

In the case of https://github.com/MikeSpreitzer/kcp-edge-mc/pull/4, that PR requested to merge branch `path-test` or my fork into branch `main` of my fork, and their latest common ancestor is commit d4ff3bff5af83cb227c4bb4c47c035b8e950213f . So that PR requested the merge of the changes from commit d4ff3bff5af83cb227c4bb4c47c035b8e950213f to commit 1e160911ef15062d4f586b5933f623af40e46d37 (the one that my branch `push-test` points to). That is, both the change made by commit fcf5d69d202e282215a836cb8fe094ce30bd3304 and the change made by commit 1e160911ef15062d4f586b5933f623af40e46d37 .

The `paths` filtering in a workflow trigger is about what files are being _changed_.  You can see that in the documents at https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files and https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore .

So PR 4 in my fork passed the paths test because one of the files changed by the merge commit is a doc file.

ezrasilvera:
> I would not be opposed to amending a release branch to improve the documentation of that release.

Let me try explain again what I meant - If we amend a branch "release-"  with files that don't match the paths filtering statement, then the workflow will not be triggered. I'm not sure this is what we want for the release branch.

Issue #1913, opened 2024-03-18
bug:  When WDS and ITS are co-located in hosting cluster, KS controller fights with Status Add-on on presence of singleton status label


pdettori:
### Describe the bug

The workstatus is created but never updated:

```shell
$ k --context imbs1 get workstatuses.control.kubestellar.io -n cluster1 213674cb-7830-4ea8-ad2c-21f5043667c0-argoproj.iov1alpha1-workflow-argo-awesome-poochenheimer -o yaml
apiVersion: control.kubestellar.io/v1alpha1
kind: WorkStatus
metadata:
  creationTimestamp: "2024-03-15T02:34:43Z"
  generation: 1
  labels:
    transport.kubestellar.io/originOwnerReferenceBindingKey: workflows
    transport.kubestellar.io/originWdsName: wds0
  name: 213674cb-7830-4ea8-ad2c-21f5043667c0-argoproj.iov1alpha1-workflow-argo-awesome-poochenheimer
  namespace: cluster1
  ownerReferences:
  - apiVersion: work.open-cluster-management.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: ManifestWork
    name: workflows-wds0
    uid: 36fce4a3-d995-454f-adbe-bbe296ded9d4
  resourceVersion: "8044"
  uid: 18e5cb09-bb80-4520-aea5-6069fce2f62a
spec:
  sourceRef:
    group: argoproj.io
    kind: Workflow
    name: awesome-poochenheimer
    namespace: argo
    resource: workflows
    version: v1alpha1
```    

The log shows a large number of conflicts that never get resolved, even after hours:

```
E0315 02:28:54.216968       1 agent.go:277] error syncing key 'util.Key{GvkKey:"argoproj.io/v1alpha1/Workflow", NamespaceNameKey:"argo/awesome-poochenheimer", DeletedObject:(*runtime.Object)(nil)}': Operation cannot be fulfilled on workstatuses.control.kubestellar.io "213674cb-7830-4ea8-ad2c-21f5043667c0-argoproj.iov1alpha1-workflow-argo-awesome-poochenheimer": the object has been modified; please apply your changes to the latest version and try again, requeuing
2024-03-15T02:28:54Z    INFO    going to update status: {"object": "[argo] workflow.argoproj.io/awesome-poochenheimer"}
[...]

PR #1999, opened 2024-04-01
:bug: Do not watch workstatuses in binding controller

andreyod:
## Summary
The binding controller does not need to reconcile workstatuses and in some setups this can cause a race condition with the status addon. For details see https://github.com/kubestellar/kubestellar/issues/1913

## Related issue(s)

Fixes https://github.com/kubestellar/kubestellar/issues/1913


MikeSpreitzer (commented):
/lgtm
Thanks!

Issue #1936, opened 2024-03-21
bug:  job filtering is not working as expected


pdettori:
### Describe the bug

The `batch.kubernetes.io/controller-uid` and `controller-uid` labels are still present in a job wrapped in the ManifestWork. This causes the error shown in the ManifestWork:

```
 message: 'Failed to apply manifest: Job.batch "pi" is invalid: [spec.template.metadata.labels[controller-uid]:
          Required value: must be ''3840adbb-55d5-46f3-992c-51b0eccc0abc'', spec.template.metadata.labels[batch.kubernetes.io/controller-uid]:
          Required value: must be ''3840adbb-55d5-46f3-992c-51b0eccc0abc'', spec.selector:
          Invalid value: v1.LabelSelector{MatchLabels:map[string]string{}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}:
          field is immutable]'
```          

### Steps To Reproduce

1. Run release 0.21.0
2. Apply a test job and related binding policy:
```
kubectl apply -f - <<EOF
apiVersion: batch/v1
kind: Job
metadata:
  name: pi
spec:
  suspend: true
  template:
    spec:
      containers:
      - name: pi
        image: perl:5.34.0
        command: ["perl",  "-Mbignum=bpi", "-wle", "print bpi(2000)"]
      restartPolicy: Never
  backoffLimit: 4
---
apiVersion: control.kubestellar.io/v1alpha1
kind: BindingPolicy
metadata:
  name: jobs
spec:
  wantSingletonReportedState: true
  clusterSelectors:
  - matchLabels:
      name: cluster1
  downsync:
  - objectSelectors:
    - matchLabels: 
        batch.kubernetes.io/job-name: pi
EOF  
``` 
3. Check ManifestWork:
```
$ k get manifestwork -n cluster1 jobs-wds0 -o yaml
apiVersion: work.open-cluster-management.io/v1
kind: ManifestWork
metadata:
  annotations:
    transport.kubestellar.io/originOwnerReferenceBindingGeneration: "2"
  creationTimestamp: "2024-03-21T01:11:10Z"
  finalizers:
  - cluster.open-cluster-management.io/manifest-work-cleanup
  generation: 2
  labels:
    transport.kubestellar.io/originOwnerReferenceBindingKey: jobs
    transport.kubestellar.io/originWdsName: wds0
  name: jobs-wds0
  namespace: cluster1
  resourceVersion: "80085"
[...]

PR #2009, opened 2024-04-03
🐛 More better Job cleaning

MikeSpreitzer:
## Summary
This PR makes the cleaning of Job objects more extensive. As reported in #1936 , in my testing it got the Job objects to the WEC but also received error messages in the ManifestWorkStatus.

## Related issue(s)

This is intended to address #1936 but is not entirely satisfactory (yet).

### DIFF
--- a/pkg/transport/filtering/job.go
+++ b/pkg/transport/filtering/job.go
@@ -27,42 +27,51 @@ func badJobKey(key string) bool {
 func cleanJob(object *unstructured.Unstructured) {
 	objectU := object.UnstructuredContent()
 	changed := false
-	selector, foundSelector, _ := unstructured.NestedMap(objectU, "spec", "selector")
+	if jobAnnotations, found, _ := unstructured.NestedMap(objectU, "metadata", "annotations"); found {
+		if cleanJobAnnotations(jobAnnotations) {
+			_ = unstructured.SetNestedMap(objectU, jobAnnotations, "metadata", "annotations")
+			changed = true
+		}
+	}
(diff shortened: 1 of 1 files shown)


pkg/transport/filtering/job.go
```diff
@@ -27,42 +27,51 @@ func badJobKey(key string) bool {
...
+		unstructured.RemoveNestedField(objectU, "status")
+		changed = true
+	}
 	if changed {
 		object.SetUnstructuredContent(objectU)
 	}
 }
 
-func cleanLabels(labels map[string]any) bool {
+func cleanJobAnnotations(annotations map[string]any) bool {
```
effi-ofer, line 63: I am ok with merging this as is.  But long term this is a bit ad hoc.  It would be nice to have a filtering function that takes a map[string]any and a list of strings to filter and then does the filtering.  Likely we will see additional resources and fields that will need to be filtered so long term we also need a way to indicate what needs to be filtered and where for a given resource type.  I think Ezra already opened an issue for something along these lines last week.

effi-ofer (approved):

pkg/transport/filtering/job.go
MikeSpreitzer: I would not object to defining and using a func that filtered a `map[string]string` by removing keys in a given `Set[string]`.

MikeSpreitzer:
Note that this will _not_ fully solve the problem, see the end of https://github.com/kubestellar/kubestellar/issues/1936#issuecomment-2024471573 .

Issue #1976, opened 2024-03-28
feature: Custom  WEC-oblivious object transformation


ezrasilvera:
## Description of current state
Currently KS supports two modes of "manipulating" the objects that are deployed into the WEC:
1. Filtering:
    * Is done per resource types 
    * Is done only for pre-defined Kube native resources (E.g., services)
    * Execute the exact pre-defined operation on the field (E.g., remove, set specific value, etc..)
2. Parameter expansion: Simple parameter substitution
    * Currently is done per WEC retrieving the same value for all objects deployed to that WEC
   
## Requirements
In several use cases the above mechanisms are not flexible enough. Some examples:
1. Filtering for CRDs - some workloads require filtering for their CRDs, supporting that with the current approach is not practical.  (E.g., Argo Workflow)
2. Different filtering for objects of the same KIND (e.g., per selection criteria) 
3. Different operations/filtering depending on the workload deployed (even for Kube resources)
4. Be able to easily define complex operations (replace fields, set values, remove fields, etc..) 

## Proposal
Staged approach to gradually cover the above requirements:

### Phase 1
1. Start with something that binds to workload objects only in terms of the "kind" or "resource" of the object, not more fine-grained.
   - Allow for backwards-compatible extension in the future to more fine-grained binding. Here "backwards-compatible" means that old clients get what they expect even when using the extended API and new implementation.
2. Start with something that only describes removal of parts of the workload object.
   - Allow for backwards-compatible extension in the future to allow other transformations on the workload object.

### Future phases
1. Allow the user to perform other operations than just removal (e.g., adding and updating fields)  
2. Allow the user to define selection criteria on which objects the operation will apply and not just according to KIND

pdettori:
Example of specific requirements for the Argo Workflows and Kubeflow Pipelines use case:

- Filtering (or changing value) for the `.spec.suspend` field
- Filtering for specific labels:
  `workflows.argoproj.io/completed`
  `workflows.argoproj.io/phase`
  `pipeline/runid`

MikeSpreitzer:
Terminology: "filtering" is a poor choice when you mean something more general (which indeed you have asserted). For that reason we have taken to saying "cleaning" rather than "filtering".

While I support the more general terminology, so that we do not have to do a bunch of rewriting if and when we want to do something more general, I have not yet seen a demonstrated need for anything other than removing content.

I also have not seen a demonstrated need to treat different objects of a given kind differently, in a given WDS.

[...]

PR #2012, opened 2024-04-03
✨ Remove spec.suspend when cleaning a Job

MikeSpreitzer:
## Summary
This PR continues the campaign for cleaning Job objects, by removing `spec.suspend` if it exists.

This is unsatisfying to me. This change **assumes** that `suspend: true` is present (when it is found) because the scenario is one where the WDS is _not_ denatured wrt Job execution. That is not the only possible scenario. It is also possible that the WDS _is_ denatured wrt Job execution and the user is trying to exert remote control over when the job is executed in the WEC.

## Related issue(s)

This addresses part of #1936

### DIFF
--- a/pkg/transport/filtering/job.go
+++ b/pkg/transport/filtering/job.go
@@ -44,6 +44,10 @@ func cleanJob(object *unstructured.Unstructured) {
 		unstructured.RemoveNestedField(objectU, "spec", "selector")
 		changed = true
 	}
+	if _, found, _ := unstructured.NestedFieldNoCopy(objectU, "spec", "suspend"); found {
+		unstructured.RemoveNestedField(objectU, "spec", "suspend")
+		changed = true
+	}
 	podLabels, foundlabels, _ := unstructured.NestedMap(objectU, "spec", "template", "metadata", "labels")
 	if foundlabels {
 		if cleanJobLabels(podLabels) {


effi-ofer:
Why is this needed?  We set suspend=true when the host is the wds?  I am asking because if I check jobs on my current setup (a seperate kind hosted wds) I see suspend=false.

vMaroon:
I can understand this being a temporary fix (although that is not stated), this practically disables the ability of dispatching Jobs with this field set to true while wanting it to remain true.

I think this is more of a customization matter than a filtering one, do you agree? I'm not sure what we concluded on customization requirements due to the length of the discussion in the relevant issue, but I would imagine customization to support both parameter substitution and resource object fields customization operations, such as replacing values and more.

vMaroon:
So what about jobs that want the field to remain true on the WECs? Do we want this pushed as a temporary fix then address it properly later? Or what's the plan?

MikeSpreitzer:
I am OK with merging this as a temporary expedient. I have added a note (https://github.com/kubestellar/kubestellar/issues/1976#issuecomment-2040502804) about potential more general handling in the future.

pdettori:
IMHO this cleaning is useful for the use case @cwiklik and I have been looking into for the integration of Kueue and KubeStellar. The use case is based on running on a WDS of type "host". The job is kept in suspend state in the WDS (where there is an active job controller)  and delivered with  to a WEC, where the job should run (thus the suspend should be removed or set to false). Having this cleaning would help to unblock @cwiklik until we have a more general solution such as https://github.com/kubestellar/kubestellar/issues/1976

PR #2026, opened 2024-04-05
✨ Enable kubestellar-controller-manager to be told ITS name

### LINKED ISSUE #2025: feature: Support more than one ITS
### Feature Description

Currently the kubestellar-controller-manager can not be told which ITS to use, it insists that there is only one and uses it. This is an unnecessary restriction.

### Proposed Solution

Add a command line flag for specifying which ITS to use.

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR generalizes the kubestellar controller-manager so that it can be told which ITS to use. This enables more than one ITS in the system!

## Related issue(s)

This addresses part of #2025 . Fully addressing it would involve adding testing and making any needed doc updates.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -246,7 +246,7 @@ CONTROLLER_TOOLS_VERSION ?= v0.13.0
 .PHONY: kustomize
 kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
 $(KUSTOMIZE): $(LOCALBIN)
-	test -s $(LOCALBIN)/kustomize || GOBIN=$(LOCALBIN) go install sigs.k8s.io/kustomize/kustomize/v5@latest
+	test -s $(LOCALBIN)/kustomize || GOBIN=$(LOCALBIN) go install sigs.k8s.io/kustomize/kustomize/v5@v5.3.0
 
 .PHONY: controller-gen
 controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
--- a/chart/templates/controller-manager.yaml
+++ b/chart/templates/controller-manager.yaml
@@ -200,6 +200,7 @@ spec:
         - --health-probe-bind-address=:8081
         - --metrics-bind-address=127.0.0.1:8080
         - --leader-elect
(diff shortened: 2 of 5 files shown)


MikeSpreitzer:
@francostellari 
@clubanderson

francostellari:
@MikeSpreitzer I would like to suggest that if an explicit ITS name is not provided or ="" , then the controller should default to it's original behavior of looking through the CPs of type imbs and choosing the first one of the list. This would make backward compatible and simplify the life in basic cases where only on ITS is used.

francostellari:
@MikeSpreitzer I would suggest to extend the PR to include the necessary changes to the Helm chart to add the argument to the https://github.com/kubestellar/kubestellar/blob/233f8dde6250e7ff0e10fd778a7d5cbaa9850896/chart/templates/controller-manager.yaml#L199 and the value to the `values.yaml`

MikeSpreitzer:
@francostellari, re:
> I would like to suggest that if an explicit ITS name is not provided or ="" , then the controller should default to its original behavior of looking through the CPs of type imbs and choosing the first one of the list.

The original behavior is to emit an error message if there is more than one ITS, not use the first. See #2027 and the code.

What makes you think this PR is not doing what you suggest?  It says:

```go
	flag.StringVar(&itsName, "its-name", "", "name of the Inventory and Transport Space to connect to (empty string means to use the only one)")
```

MikeSpreitzer:
@francostellari : I have generalized the kubestellar controller-manager chart as you suggested.

francostellari:
> @francostellari, re:
> 
> > I would like to suggest that if an explicit ITS name is not provided or ="" , then the controller should default to its original behavior of looking through the CPs of type imbs and choosing the first one of the list.
> 
> The original behavior is to emit an error message if there is more than one ITS, not use the first. See #2027 and the code.
> 
> What makes you think this PR is not making the empty string mean the original behavior? It says:
> 
> ```go
> 	flag.StringVar(&itsName, "its-name", "", "name of the Inventory and Transport Space to connect to (empty string means to use the only one)")
> ```

@MikeSpreitzer sorry, I missed the text.

PR #2030, opened 2024-04-05
🐛 Allow a 'singleton' bindingpolicy to select no clusters

### LINKED ISSUE #1945: bug:  controller manager crashes on OpenShift within a few minutes of start
[cm2.log](https://github.com/kubestellar/kubestellar/files/14706234/cm2.log)
### Describe the bug

Running kubestellar v0.21.0 on OpenShift testing scenario#2 which uses WDS of type host. This cluster has many crds.

```
kubectl api-resources | wc                                                                                                                                                                                                                      ─╯
     409    1763   69102
```
 
After the KS install I deployed mcad appwrapper on the host and noticed that it was not downsyncing to the WEC. Checked the controller manager log and noticed a crash. Check the attached log details. The reason for the crash is:` panic: runtime error: index out of range [0] with length 0`

This crash happens soon after the controller manager starts.

I verified that my WECs have joined and all is well there:

```
kubectl --context imbs1 get managedclusters                                                                          

NAME  HUB ACCEPTED  MANAGED CLUSTER URLS  JOINED  AVAILABLE  AGE
wec4  true                 True   True    155m
wec3  true                 True   True    155m
```

I am running with private image built by Paolo which fixes a timeout problem in controller manager. The image is 

[quay.io/pdettori/controller-manager:1710901385](http://quay.io/pdettori/controller-manager:1710901385)


### Steps To Reproduce

Follow kubestellar install steps documented in examples for scenario#2 
[](https://docs.kubestellar.io/main/direct/examples/#scenario-2-using-the-hosting-cluster-as-wds-to-deploy-a-custom-resource)

After installing controller manager release level v0.21.0 (as instructed by KS documentation), edit controller manager deployment and replace the cm image with [quay.io/pdettori/controller-manager:1710901385](url)



### Expected Behavior

The controller manager should not crash 

### Additional Context

_No response_

waltforme:
## Summary
This PR fixes #1945. More details are documented at https://github.com/kubestellar/kubestellar/issues/1945#issuecomment-2034897896.

Fixes #1945

### DIFF
--- a/pkg/binding/bindingpolicy.go
+++ b/pkg/binding/bindingpolicy.go
@@ -403,7 +403,8 @@ func SliceContains[Elt comparable](slice []Elt, seek Elt) bool {
 	return false
 }
 
-// sort by name and pick first cluster so that the choice is deterministic based on names
+// pickSingleDestination sorts clusters by name and picks first cluster so that the choice is deterministic based on names
+// pickSingleDestination expects a non-empty clusterSet
 func pickSingleDestination(clusterSet sets.Set[string]) sets.Set[string] {
 	return sets.New(sets.List(clusterSet)[0])
 }
--- a/pkg/binding/bindingpolicy.go
+++ b/pkg/binding/bindingpolicy.go
@@ -86,12 +86,17 @@ func (c *Controller) handleBindingPolicy(ctx context.Context, objIdentifier util
 		if err != nil {
 			return fmt.Errorf("failed to ocm.FindClustersBySelectors: %w", err)
 		}
+		if len(clusterSet) == 0 {
+			logger.Info("No clusters are selected by BindingPolicy", "name", bindingPolicy.Name)
+		}
 
 		if bindingPolicy.Spec.WantSingletonReportedState {
 			// if the bindingpolicy requires a singleton status, then we should only
 			// have one destination
 			// TODO: this should be removed once we have proper enforcement or error reporting for this
-			clusterSet = pickSingleDestination(clusterSet)
+			if len(clusterSet) > 1 {
+				clusterSet = pickSingleDestination(clusterSet)
+			}
 		}
 
 		// set destinations and enqueue binding for syncing
--- a/test/e2e/ginkgo/multiple_cluster_deployment_test.go
+++ b/test/e2e/ginkgo/multiple_cluster_deployment_test.go
@@ -214,6 +214,29 @@ var _ = ginkgo.Describe("end to end testing", func() {
 	})
 
 	ginkgo.Context("singleton status testing", func() {
+		ginkgo.It("no singleton status when a singleton bindingpolicy/deployment is created but no matching WEC(s)", func() {
+			util.DeleteDeployment(ctx, wds, ns, "nginx")
+			util.CreateDeployment(ctx, wds, ns, "nginx-singleton",
+				map[string]string{
+					"app.kubernetes.io/name": "nginx-singleton",
+				})
+			util.CreateBindingPolicy(ctx, ksWds, "nginx-singleton",
+				[]metav1.LabelSelector{
+					{MatchLabels: map[string]string{"name": "CelestialNexus"}},
+				},
+				[]ksapi.DownsyncObjectTest{
+					{ObjectSelectors: []metav1.LabelSelector{
+						{MatchLabels: map[string]string{"app.kubernetes.io/name": "nginx-singleton"}},
+					}}})
+			patch := []byte(`{"spec":{"wantSingletonReportedState": true}}`)
+			_, err := ksWds.ControlV1alpha1().BindingPolicies().Patch(
+				ctx, "nginx-singleton", types.MergePatchType, patch, metav1.PatchOptions{})
+			gomega.ExpectWithOffset(1, err).NotTo(gomega.HaveOccurred())
+			util.ValidateNumDeployments(ctx, wec1, ns, 0)
+			util.ValidateNumDeployments(ctx, wec2, ns, 0)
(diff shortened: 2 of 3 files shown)


pkg/binding/bindingpolicy.go
```diff
@@ -91,7 +91,9 @@ func (c *Controller) handleBindingPolicy(ctx context.Context, objIdentifier util
 			// if the bindingpolicy requires a singleton status, then we should only
 			// have one destination
 			// TODO: this should be removed once we have proper enforcement or error reporting for this
-			clusterSet = pickSingleDestination(clusterSet)
+			if len(clusterSet) > 1 {
```
MikeSpreitzer, line 94: If `len(clusterSet) == 0` then this violates what the user is expected to do; I would log an error at this point. #1950 introduces a better way to report user errors.

waltforme: `len(clusterSet) == 0` may be a transient and legal state during which a user is managing the selection of WEC(s), including change the binding policies' cluster selectors, or changing labels of the WEC(s), or adding/removing WEC(s).

Say the user wants a cluster to report the status. He/she may create the binding policy as the first step, then onboard or label the cluster as the second step. We shouldn't and can't enforce the order of the two steps. Should we consider this as an error?

vMaroon: Regardless of how this conversation resolves, there is a bug in `pickSingleDestination` since it would attempt to access the first element in an empty `sets.List`.

@waltforme it's not related to this PR, but if `pickSingleDestination` won't be changed, we should add function documentation on expecting that the parameter `clusterSet` is a none-empty set.

waltforme: @vMaroon Thanks! Not related to this PR but why not just do something here, so I did 7c0c47f3701514e5d5b4d7d942d103eb89a978ea.

vMaroon: @waltforme thanks, I meant to handle it in this PR indeed.

MikeSpreitzer: Indeed the user inconsistency may be transient. Or it may not. At the point in the code that we are discussing here, it is impossible to know. That is why I am dissatisfied with silently ignoring it. #1950 introduces a state-based way to report user errors; that will naturally mean that a transient error will not persistently be reported. Until then, I think the best that can be done is to log a warning.

waltforme: I added logging for the `len(clusterSet) == 0` case in the latest push. The `handleBindingPolicy` func uses non-nil error to signal requeuing, but I think we don't need to requeue for this case. So, I'm not logging an error. I hope there is a `logger.Warning` func but there is no such thing. I am using a `Info` instead.

waltforme:
Force-push to https://github.com/kubestellar/kubestellar/commit/df6719e7335b4d776e7adf430c8380745872e948 adds logging for `len(clusterSet) == 0`, then squash that logging to appropriate commit.

pkg/binding/bindingpolicy.go
pdettori: I agree that just logging is fine until we have a more general solution as @MikeSpreitzer pointed out.

pdettori (commented):
/lgtm

pkg/binding/bindingpolicy.go
MikeSpreitzer: @waltforme: I still do not see that logging. I looked in https://github.com/waltforme/kubestellar/blob/singleton/pkg/binding/bindingpolicy.go and in https://github.com/kubestellar/kubestellar/commit/df6719e7335b4d776e7adf430c8380745872e948

waltforme: The logging is squashed into https://github.com/kubestellar/kubestellar/pull/2030/commits/5689c99b8d730e85710d1eaa1e9f2c4db648d108. @MikeSpreitzer

MikeSpreitzer: Oh, sorry, the new logging *is* in the commits where I looked, but not at the place in the logic where I expected it.

pkg/binding/bindingpolicy.go
```diff
@@ -86,12 +86,17 @@ func (c *Controller) handleBindingPolicy(ctx context.Context, objIdentifier util
 		if err != nil {
 			return fmt.Errorf("failed to ocm.FindClustersBySelectors: %w", err)
 		}
+		if len(clusterSet) == 0 {
+			logger.Info("No clusters are selected by BindingPolicy", "name", bindingPolicy.Name)
```
MikeSpreitzer, line 90: This is a user error _if_ `bindingPolicy.Spec.WantSingletonReportedState`, otherwise it is not a problem. The user always gets feedback on `len(clusterSet)` by counting the listed clusters in the Binding object, so logging it in non-problematic cases seems spurious to me. This can be improved in follow-on work.

/approve

waltforme: IMO, `bindingPolicy.Spec.WantSingletonReportedState` doesn't affect whether `len(clusterSet)==0` is a problem or not.

If it's 'impossible to know whether or not the error is transient' for `WantSingletonReportedState`==true, then it's also 'impossible to know whether or not the error is transient' for `WantSingletonReportedState`==false.

MikeSpreitzer: `len(clusterSet)==0 && bindingPolicy.Spec.WantSingletonReportedState` is an inconsistent state. The problem is the conjunction. The problem may go away in the future. Now that #2044 has merged we can make progress on state-based error reporting.

PR #2034, opened 2024-04-07
✨ Add support for testing with deployment on K3D

eaepstein:
## Summary
The following steps will deploy/use K3D by setting env-var "USE_K3D".  e.g. USE_K3D=1, or yes, or anything:
setup-kubestellar.sh, run-test.sh, check_pre_req.sh, cleanup.sh, Makefile, ginkgo_suite_test.go

Tested on Ubuntu 22.04. 

## Related issue(s)

Fixes #

### DIFF
--- a/test/e2e/common/cleanup.sh
+++ b/test/e2e/common/cleanup.sh
@@ -20,8 +20,17 @@ set -e # exit on error
 : -------------------------------------------------------------------------
 : "Cleaning up from previous run of an e2e test"
 
-kind delete cluster --name cluster1
-kind delete cluster --name cluster2
-kind delete cluster --name kubeflex
+if [ -z "$USE_K3D" ]; then
+   kind delete cluster --name cluster1
+   kind delete cluster --name cluster2
+   kind delete cluster --name kubeflex
+else
+   k3d cluster delete cluster1
+   k3d cluster delete cluster2
+   k3d cluster delete kubeflex
+fi
 kubectl config delete-context cluster1 || true
 kubectl config delete-context cluster2 || true
+kubectl config delete-context kubeflex || true
+kubectl config delete-context imbs1 || true
+kubectl config delete-context wds1 || true
--- a/test/e2e/common/setup-kubestellar.sh
+++ b/test/e2e/common/setup-kubestellar.sh
@@ -41,10 +41,53 @@ source "$COMMON_SRCS/setup-shell.sh"
 
 :
 : -------------------------------------------------------------------------
-: Create a Kind hosting cluster with nginx ingress controller and KubeFlex controller-manager
+: Create a hosting cluster with nginx ingress controller and KubeFlex controller-manager
 :
-kflex init --create-kind $disable_chatty_status
-: Kubeflex kind cluster created.
+if [ -z "$USE_K3D" ]; then
+   kflex init --create-kind $disable_chatty_status
+   kubectl config rename-context kind-kubeflex kubeflex
+   : Kubeflex kind cluster created.
+else
+   k3d cluster create -p "9443:443@loadbalancer" --k3s-arg "--disable=traefik@server:*" kubeflex
+   helm install ingress-nginx ingress-nginx --repo https://kubernetes.github.io/ingress-nginx --version 4.6.1 --namespace ingress-nginx --create-namespace
+   kubectl config rename-context k3d-kubeflex kubeflex
+   :
+   : Kubeflex k3d cluster created. Rename docker container until kubeflex feature implemented
+   docker stop k3d-kubeflex-server-0
+   docker rename k3d-kubeflex-server-0 kubeflex-control-plane
+   docker start kubeflex-control-plane
+   :
+   : Wait 60 seconds for pods to restart
+   sleep 60
+   :
+   : modify ingress deployment and install kflex controller
+   tmpfile=$(mktemp)
+   cat > "tmpfile" <<EOF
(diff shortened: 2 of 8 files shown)


effi-ofer:
Looks good @eaepstein but the pr-test-e2e workflow fail because it tries to access the old kind-kubeflex context.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -41,10 +41,53 @@ source "$COMMON_SRCS/setup-shell.sh"
 
 :
 : -------------------------------------------------------------------------
-: Create a Kind hosting cluster with nginx ingress controller and KubeFlex controller-manager
+: Create a hosting cluster with nginx ingress controller and KubeFlex controller-manager
 :
-kflex init --create-kind $disable_chatty_status
-: Kubeflex kind cluster created.
+if [ -z "$USE_K3D" ]; then
```
ezrasilvera, line 46: @eaepstein With such a big change + we may support other platforms in the future, I would recommend we separate the platform dependent part into separate scripts and have a different setup for each (setup-kind setup-ocp setup-k3d, ect..) 
We of course  should try to leave all common parts in a single script

test/e2e/common/cleanup.sh
```diff
@@ -20,8 +20,17 @@ set -e # exit on error
 : -------------------------------------------------------------------------
 : "Cleaning up from previous run of an e2e test"
 
-kind delete cluster --name cluster1
-kind delete cluster --name cluster2
-kind delete cluster --name kubeflex
+if [ -z "$USE_K3D" ]; then
```
ezrasilvera, line 23: See the other comment - I think we need separate scripts for different envs

test/e2e/common/setup-kubestellar.sh
```diff
@@ -41,10 +41,53 @@ source "$COMMON_SRCS/setup-shell.sh"
...
+   kflex init --create-kind $disable_chatty_status
+   kubectl config rename-context kind-kubeflex kubeflex
```
ezrasilvera, line 48: maybe kf-mgt or kflexmgt or something that shows that this is the management cluster where KF is running.

test/e2e/multi-cluster-deployment/run-test.sh
```diff
@@ -34,7 +34,11 @@ SRC_DIR="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)"
 COMMON_SRCS="${SRC_DIR}/../common"
 HACK_DIR="${SRC_DIR}/../../../hack"
 
-"${HACK_DIR}/check_pre_req.sh" --assert --verbose kubectl docker kind make go ko yq helm kflex ocm
+if [ -z "$USE_K3D" ]; then
```
ezrasilvera, line 37: I would move that to  the setup-ks  as you in any case have different code for each platform there , so we can maintain the run-test agnostic to platform.

ezrasilvera:
@eaepstein few general comments:
1. Not sure we agreed we should run all our e2e tests on k3d and not just the main 2 examples (multi cluster & singleton) 
2. Maybe it would be better to try and separate the break the setup script and create a different platform dependent one for each platform. In the future we may want to support several other platforms and I suspect we don't want to go into a complex if/else statement  all over our testing scripts
3. Maybe we can have some generic format that a platform setup script need to comply with, this way we can have our flow agnostic to the platform itself e.g., 
   - createCluster
   - deleteCluster
   - renameCluster
   - installKubeFlex
   - etc...

test/e2e/common/setup-kubestellar.sh
eaepstein: There is really only one big change: creating the kubeflex cluster. All that code should be folded into kubeflex so that 
`kflex init --create-kind $disable_chatty_status`
becomes
`kflex init --create-k3d $disable_chatty_status`

All the other places are one or two line changes which could be externalized into one or more extensible scripts.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -41,10 +41,53 @@ source "$COMMON_SRCS/setup-shell.sh"
...
+   : Kubeflex kind cluster created.
+else
+   k3d cluster create -p "9443:443@loadbalancer" --k3s-arg "--disable=traefik@server:*" kubeflex
```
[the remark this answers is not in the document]
eaepstein: A more flexible setup-kubestellar could be useful beyond testing. An alternative would be one or more helm charts to get KS and managed clusters going in different environments.

test/e2e/multi-cluster-deployment/run-test.sh
eaepstein: Duplicating small routines like run-test.sh into multiple scripts differing by just one lines may not be the best answer.

test/e2e/common/setup-kubestellar.sh
ezrasilvera: So maybe we should just start with that ?  and add a PR to KF and only then to here ?

test/e2e/multi-cluster-deployment/run-test.sh
ezrasilvera: I agree , however multiple if(s)  are also problematic. Think of 5 different platforms :(

eaepstein:
Dropping the idea of idea of making the basic test script work with k3d in addition to kind until there is a good reason to do so.

Issue #1756, opened 2024-02-13
bug: scripts/add-helm-code.sh issues


MikeSpreitzer:
### Describe the bug

(1) The script named "add-helm-code.sh" in `scripts/` is badly named. It does not add code; it tweaks some names in `operator.yaml`.

(2) The comments suggest that it takes one command line argument that can be either "add" or "remove". But it only does something if that argument is "add".

A better name would be something like `specialize-names.sh`.

(3) This script is designed to ONLY be used in one place --- the Makefile recipe for target `chart` --- but does not say so.

(4) A script for development-time use belongs in `hack/` not `scripts/` (the latter is for users not contributors).

(5) The script does some very fragile text search-and-replace operations. Since it is really an amendment to `$(KUSTOMIZE) build config/default`, it would be more robust to adjust the kustomization to also do this work.

### Steps To Reproduce

See above.

### Expected Behavior

See above.

ezrasilvera:
@francostellari is this still a valid issue?  Are you working on it ?

francostellari:
> @francostellari is this still a valid issue? Are you working on it ?

@ezrasilvera (/cc @MikeSpreitzer ) indeed, I was learning a bit more about YQ to implement a solution with it.
Please see #2036

MikeSpreitzer:
I think that this issue has been resolved by #2036 .

PR #2036, opened 2024-04-08
✨ Replace add-helm-code.sh with YQ commands

francostellari:
## Summary

A combination of YQ commands is used to remove the need of using script `add-helm-code.sh` as described in #1756 

## Related issue(s)

Fixes #
#1756


ezrasilvera:
@francostellari the previous script also had cleanup-tags.  Isn't it being used anymore ?

ezrasilvera:
@francostellari one high level comment. 
1. In general I think we should reduce the usage of Makefile phony targets with complex logic in them and move instead to use utility scripts (i.e., remove the target completely). This PR actually goes in the opposite direction.  (cc: @MikeSpreitzer ) .
2. It seems strange to me that 2 lines of yq replaced the whole script. Doesn't it mean that the script itself could have been just a couple of lines as well ?

francostellari:
> @francostellari the previous script also had cleanup-tags. Isn't it being used anymore ?

I could not determine any impact of the cleanup in the final yaml.
@pdettori can you help me understand what it needs to do?

francostellari:
> @francostellari one high level comment.
> 
> 1. In general I think we should reduce the usage of Makefile phony targets with complex logic in them and move instead to use utility scripts (i.e., remove the target completely). This PR actually goes in the opposite direction.  (cc: @MikeSpreitzer ) .
> 2. It seems strange to me that 2 lines of yq replaced the whole script. Doesn't it mean that the script itself could have been just a couple of lines as well ?

On the other hand putting 2 lines into a script, seems a bit excessive.
Let me know what you agree on, no difference from my point of view.

MikeSpreitzer:
Yes, the old script could have been just two invocations of yq. I get the impression that the script we see today is the result of some evolution in the author's thinking, and he has not had time to fully tidy it up; I think that the litany of complaints in #1756 is evidence of this. I think that the bash function named `cleanup_helm_tags` in the old script is a vestige of old thinking. I think @francostellari's replacement here is much simpler and has no need of that bash function.

Since I last ranted about phony bash targets I have realized one complication in getting rid of them. When a phony target depends on a non-phony target, replacing that phony target with a script means that either the script invokes `make` on the non-phony dependency or invokes a script that has make-like functionality. Neither is entirely pleasing. In any case, this PR is not introducing any phony `make` targets, so I do not think we need to hold up this PR for that general concern.

MikeSpreitzer (commented):
/lgtm
/approve
Thanks!

Makefile
```diff
@@ -182,9 +182,10 @@ kind-load-image:
 .PHONY: chart
 chart: manifests kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(shell echo ${IMG} | sed 's/\(:.*\)v/\1/')
-	$(KUSTOMIZE) build config/default > chart/templates/controller-manager.yaml
-	scripts/add-helm-code.sh add
-
+	$(KUSTOMIZE) build config/default \
+		| yq '. | select(.kind == "ClusterRole*").metadata.name |= "{{.Values.ControlPlaneName}}-" + .' \
```
MikeSpreitzer, line 186: Is `. |` a no-op in `yq`?

MikeSpreitzer: Yes, it is a no-op. This is not a blocker.
/lgtm

PR #2041, opened 2024-04-10
✨ Introduce some control over controller verbosity

MikeSpreitzer:
## Summary
This PR introduces some control over the logging verbosity of the kubestellar controller-manager and the transport controller. The controllers themselves already had command line flags for this, but the higher level constructs passed constant values. This PR generalizes the kubestellar controller-manager Helm chart so that it can be told what verbosity to pass in.  This PR generalizes the scripts/deploy-transport-controller.sh script. This PR also generalizes the test/e2e/common/setup-kubestellar.sh and test/e2e/multi-cluster-deployment/run-test.sh scripts.

This PR does NOT change the kubestellar PostCreateHook, even though PostCreateHooks can have user-defined parameters in release 0.5.0 and later of KubeFlex. One reason is that KubeStellar currently only requires release 0.4.2 of KubeFlex. The other reason is that a PostCreateHook parameter can not have a default value; every usage of the PCH has to specify the value to use for that parameter.

## Related issue(s)

This addresses part of #2042


scripts/deploy-transport-controller.sh
```diff
@@ -15,10 +15,66 @@
...
+          fi;;
+        (--image-pull-policy)
+          if (( $# > 1 )); then
+            IMAGE_PULL_POLICY="$2"
+            shift
+          else
+            echo "Missing image-pull-policy value" >&2
+            exit 1;
+          fi;;
+        (--controller-verbosity)
```
ezrasilvera, line 46: Maybe "controller"  is redundant here ?  This script is only for controller.  BTW, can we also add -v which is common behavior ?

scripts/deploy-transport-controller.sh
```diff
@@ -15,10 +15,66 @@
...
+          fi;;
+        (-*) echo "$0: unrecognized flag '$1'" >&2
+            exit 1;;
+        (*) args[${#args[*]}]="$1"
+    esac
+    shift
+done
+
+if (( ${#args[*]} < 2 || ${#args[*]} > 3 )); then
+    echo "$0: expecting two or three positional arguments" >&2
```
ezrasilvera, line 62: I know it's more complicated to test for the exact acceptable options, but this error doesn't give much valuable info

test/e2e/multi-cluster-deployment/README.md
```diff
@@ -2,6 +2,10 @@
 
 This test is an executable variant of the "multi-cluster workload deployment with kubectl" scenario in [the examples doc](../../../docs/content/direct/examples.md). This test has the same prerequisites as the cited one. This test can test either (a) the  local copy of the repo or (b) the release identified in the kubestellar PostCreateHook (which will be the last release created, regardless of quality, except for that brief time when it identifies the release about to be made). Testing the local copy is the default behavior; to test the release identified in the PostCreateHook, pass `--released` on the command line of `run-test.sh`.
 
+The kubestellar controller-manager will be invoked with `-v=2` unless otherwise specified on the command line with `--kubestellar-controller-manager-verbosity $number`. This verbosity can not be set to a value other than 2 when using `--released`.
+
+The transport controller will be invoked with `-v=4` unless othewise specified on the command line with `--transport-controller-verbosity $number`.
```
ezrasilvera, line 7: Seems like a very high level. Developers usually push a lot of printouts into V4 (@nirrozenbaum  ? )

scripts/deploy-transport-controller.sh
MikeSpreitzer: The Deployment has three containers, only one of which is the controller.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -15,17 +15,42 @@
...
+TRANSPORT_CONTROLLER_VERBOSITY=4
+
+while [ $# != 0 ]; do
+    case "$1" in
+        (-h|--help) echo "$0 usage: (--released | --kubestellar-controller-manager-verbosity \$num | --transport-controller-verbosity \$num)*"
+                    exit;;
+        (--released)
+            wds_extra="-p kubestellar"
+            use_release=true;;
+        (--kubestellar-controller-manager-verbosity)
```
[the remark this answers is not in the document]
MikeSpreitzer: The problem here is that this setup covers all the system components. "manager" is not very clear, and while it is true that today there is only one pod that is called a 'controller manager', that is a fragile detail.

chart/templates/controller-manager.yaml
```diff
@@ -1,237 +1 @@
...
-        securityContext:
-          allowPrivilegeEscalation: false
-          capabilities:
-            drop:
-            - ALL
-      securityContext:
-        runAsNonRoot: true
-      serviceAccountName: kubestellar-controller-manager
-      terminationGracePeriodSeconds: 10
+This file is overwritten by `make chart` (whose results are not kept in Git) and ignored by `make local-chart`.
```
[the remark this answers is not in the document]
MikeSpreitzer: I considered that, and thought that leaving an explanation in place would be more helpful.

scripts/deploy-transport-controller.sh
MikeSpreitzer: Yes. Because either two or three is acceptable, a decent explanation would basically recapitulate most of the usage message. I have pushded a revision that adds a reference to the usage.

chart/templates/controller-manager.yaml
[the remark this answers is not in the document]
ezrasilvera: This seems strange. It's as if we will leave such a place holder for all our generated files that are not in git. I don't see the value in this

test/e2e/multi-cluster-deployment/README.md
ezrasilvera: I know, this is why I pointed to @nirrozenbaum I thought 4 is too high for our release default.

chart/templates/controller-manager.yaml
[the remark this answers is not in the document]
MikeSpreitzer: I think that this one is a fairly special case. First, note that currently this file **is** under Git control. More importantly, it is part of a suite of files --- the ones that make up the Helm chart definition. The rest of the suite is in the Git-controlled contents, and the reader will would reasonably wonder about it if this were not here. Finally, note that this is not the only Git-controller place where we have a note saying that it gets replaced.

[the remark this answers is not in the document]
ezrasilvera: I still strongly believe we should just remove it  and instead put the comment that the file is generate in the docs and/or  in the helm definition. 
However, I will not block the PR because of this. I leave it up to you ...

test/e2e/multi-cluster-deployment/README.md
ezrasilvera: @MikeSpreitzer if level 4 seems like a reasonable output , I'm OK with it.

ezrasilvera:
We can delete the file in the future once we agree how to document it
/lgtm

PR #2045, opened 2024-04-10
✨ Define API for custom cleaning

MikeSpreitzer:
## Summary
This PR proposes an API for custom cleaning.

## Related issue(s)

This is the start of addressing #1976 . In particular it covers Phase 1 in the proposal section


ezrasilvera:
@MikeSpreitzer (@pdettori @eaepstein ) I don't understand how we "start addressing" #1976 before we have agreement on what are the requirements. 
As we agreed in the call, all items that are marked as "major feature"  should follow the procedure of:
1. Design document that clearly state the requirements
2. Review and agreement on the requirements
3. High level design (proposal) through goggle doc (or detailed slides)
   - Review of the design through comments 
4. Coding PR

You started directly with (5).  We didn't even have any discussion what's so ever  other than couple of general comment in #1976   on the actual requirements and what is the problem we are trying to solve here.

api/control/v1alpha1/types.go
```diff
@@ -257,3 +257,46 @@ type BindingList struct {
...
+// The name of this object is the `resource.apiGroup` of the resource
+// that it is about; for example, `appwrappers.workload.codeflare.dev`.
+// The cleaning specified here is in addition to whatever cleaning
+// is built into KubeStellar for that resource.
+//
+// +genclient
+// +genclient:nonNamespaced
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Cluster,shortName={cc},categories={all}
+type CustomClean struct {
```
ezrasilvera, line 272: Why only Cleaning?    Why can't we add or change fields ?

api/control/v1alpha1/types.go
```diff
@@ -257,3 +257,46 @@ type BindingList struct {
...
+type CustomCleanSpec struct {
+
+	// `remove` is a list of JSONPath expressions (https://goessner.net/articles/JsonPath/)
+	// that identify part of the object to remove if present.
+	// Only a subset of JSONPath is supported.
+	// The expression used in a filter must be a conjunction of field == literal tests.
+	// Examples:
+	// - "$.spec.resources.GenericItems[*].generictemplate.metadata.resourceVersion"
+	// - "$.store.book[?(@.author == 'Kilgore Trout' && @.category == 'fiction'].price"
+	Remove []string `json:"remove,omitempty"`
```
ezrasilvera, line 289: I guess if we only "clean" then "remove is OK, but I would expect a general json patch string here

nirrozenbaum: I also think we may do remove/update/add. 
it would be good to have kind of `operator` (add/update/remove).

nirrozenbaum:
I wrote one main comment about leveraging the same mechanism not only for cleaning but for add/update/remove. 
other than that the api looks good to me.

config/crd/bases/control.kubestellar.io_customcleans.yaml
```diff
@@ -0,0 +1,58 @@
...
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: CustomCleanSpec describes how to clean an object.
+            properties:
```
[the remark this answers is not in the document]
ezrasilvera: @nirrozenbaum I agree - but as I pointed out, this goes back to the requirements.  I agree the solution should support more than just removing fields.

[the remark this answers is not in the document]
MikeSpreitzer: In the future, it will be easier for me to follow if comments about the API go in the API file rather than derived files like this CRD YAML.

MikeSpreitzer:
The force-push to c5aa4680b responds to the recent requirements and naming discussion.

api/control/v1alpha1/types.go
```diff
@@ -257,3 +257,46 @@ type BindingList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []Binding `json:"items"`
 }
+
+// CustomClean describes how to clean a resource.
+// That is, how to clean objects of a given resource.
```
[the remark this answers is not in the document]
ezrasilvera: I think we agreed to have a more flexible way of mapping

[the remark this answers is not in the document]
MikeSpreitzer: You should be seeing that now in the latest revision.

api/control/v1alpha1/types.go
MikeSpreitzer: you should be seeing that the latest revision uses a different name.

ezrasilvera: Thanks

pdettori:
I think this proposal reflects well what we agreed on in terms of simplicity, addressing only the use case at hand as a starting point and supporting future evolution to cover more complex use cases.

api/control/v1alpha1/types.go
```diff
@@ -257,3 +257,53 @@ type BindingList struct {
...
+// +kubebuilder:resource:scope=Cluster,shortName={ct},categories={all}
+type CustomTransform struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec CustomTransformSpec `json:"spec,omitempty"`
+}
+
+// CustomTransformSpec selects some objects and describes how to transform them.
+// The selected objects are those that match the `apiGroup` and `resource` fields.
```
ezrasilvera, line 279: Do we want to add a comment about future extension that will add more than just remove ?   (here or near the Remove array

MikeSpreitzer: I have been beaten up about including remarks about potential future development. Even in documentation. if anything, let's preserve such thoughts in the Issue, #1976 .

Issue #2037, opened 2024-04-08
feature: Easy release testing for KS on OCP


ezrasilvera:
1. Configure an OCP environment for KCP (hub + 2 WECs)
2. Script(s) based procedure for release testing
   - Download KS release sources to get the testing scripts 
   - Download all release artifacts (KS, OTP, etc..)
   - Install KS environment (KS components, WECs, agents , etc.._)
   - Run the KS e2e tests (deployment t multiple clusters + ginkgo)
  
Goal:  Have a very simple procedure, preferably using 1-2 scripts that automatically will perform the needed configuration and launch the tests.

effi-ofer:
Thanks Braulio for delivering #2046.  As discussed, this issue will also contain the method of access to the OCP cluster.

dumb0002:
A simple procedure to automatically run our e2e tests in a OCP env has be implemented in this PR: https://github.com/kubestellar/kubestellar/pull/2046

PR #2046, opened 2024-04-11
✨   KS release test OCP: deployment in multiple clusters

dumb0002:
## Summary

This PR adds an easy release testing for KS on OCP for deployment in multiple clusters

## Related issue(s)
https://github.com/kubestellar/kubestellar/issues/2037

Fixes #

### DIFF
--- a/test/e2e/common/cleanup.sh
+++ b/test/e2e/common/cleanup.sh
@@ -16,12 +16,57 @@
 set -x # echo so that users can understand what is happening
 set -e # exit on error
 
+
+env="kind"
+
+if [ "$1" == "--env" ]; then
+    env="$2"
+    shift 2
+fi
+
 :
 : -------------------------------------------------------------------------
 : "Cleaning up from previous run of an e2e test"
 
-kind delete cluster --name cluster1
-kind delete cluster --name cluster2
-kind delete cluster --name kubeflex
-kubectl config delete-context cluster1 || true
-kubectl config delete-context cluster2 || true
+if [ $env == "kind" ];then
+    kind delete cluster --name cluster1
+    kind delete cluster --name cluster2
+    kind delete cluster --name kubeflex
+    kubectl config delete-context cluster1 || true
+    kubectl config delete-context cluster2 || true
+
+elif [ $env == "ocp" ];then
+    # Unregister the managed clusters
+    function unregister_cluster() {
+        cluster=$1
+        kubectl --context $cluster delete ns nginx
+        clusteradm unjoin --cluster-name $cluster
+        kubectl --context $cluster delete ns open-cluster-management open-cluster-management-agent open-cluster-management-agent-addon
+    }
+
+    unregister_cluster cluster1
+    unregister_cluster cluster2
+
+    # To uninstall KubeFlex, first ensure you remove all you control planes:
+    kubectl config use-context kscore
+    kubectl delete cps --all
+    helm delete -n kubeflex-system kubeflex-operator
+    helm delete -n kubeflex-system postgres
+    kubectl -n kubeflex-system delete pvc data-postgres-postgresql-0
+    kubectl delete ns kubeflex-system
+
+    # Unset the kubeconfig contexts
+    kubectl config unset contexts.its1
+    kubectl config unset contexts.wds1
+    kubectl config unset contexts.wds2
+else
+   echo "$0: unknown flag option" >&2 ;
+   echo "Usage: $0 [--env kind | ocp]" >& 2
+   exit 1
+fi
+
+
+
+
+
+
--- a/test/e2e/multi-cluster-deployment/run-test.sh
+++ b/test/e2e/multi-cluster-deployment/run-test.sh
@@ -18,9 +18,11 @@
 
 set -x # so users can see what is going on
 
+env="kind"
+
 while [ $# != 0 ]; do
     case "$1" in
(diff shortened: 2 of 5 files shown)


test/e2e/common/cleanup.sh
```diff
@@ -16,12 +16,53 @@
 set -x # echo so that users can understand what is happening
 set -e # exit on error
 
+
+env="kind"
+
+if [ "$1" == "--env" ]; then
+    env="$2"
+    shift
```
MikeSpreitzer, line 24: Consuming two command line arguments would normally be done with _two_ `shift` commands.
This is not a fatal problem at the moment, because nothing later looks at the command line. But this is a fragile comfort.

test/e2e/common/cleanup.sh
```diff
@@ -16,12 +16,53 @@
 set -x # echo so that users can understand what is happening
 set -e # exit on error
 
+
+env="kind"
+
+if [ "$1" == "--env" ]; then
```
MikeSpreitzer, line 22: What should happen if there is no `--env` on the command line? I think that a silent "nothing" is not the best answer.

test/e2e/multi-cluster-deployment/run-test.sh
```diff
@@ -18,25 +18,43 @@
...
     setup_flags="$1"
     shift
 fi
 
-if [ "$#" != 0 ]; then
-    echo "Usage: $0 [--released]" >& 2
-    exit 1
+if [ "$1" == "--env" ]; then
+    env="$2"
+    shift
```
MikeSpreitzer, line 30: Consuming `--env $something` requires _two_ `shift` commands.

test/e2e/multi-cluster-deployment/run-test.sh
```diff
@@ -18,25 +18,43 @@
...
-"${HACK_DIR}/check_pre_req.sh" --assert --verbose kubectl docker kind make go ko yq helm kflex ocm
+    "${COMMON_SRCS}/cleanup.sh"
+    source "${COMMON_SRCS}/setup-shell.sh"
+    "${COMMON_SRCS}/setup-kubestellar.sh" $setup_flags
+    "${SRC_DIR}/use-kubestellar.sh"
+fi
+
+
+if [ $env == "ocp" ];then
+   bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/release-$KUBESTELLAR_VERSION/test/e2e/common/cleanup.sh) --env ocp
```
MikeSpreitzer, line 55: Why the change to fetching the cleanup script from github? In the `kind` case, the local copy is used. These E2E tests are specifically designed to be able to test the local copy of the repo.

test/e2e/multi-cluster-deployment/run-test.sh
```diff
@@ -18,25 +18,43 @@
...
+   source <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/release-$KUBESTELLAR_VERSION/test/e2e/common/setup-shell.sh)
+   bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/release-$KUBESTELLAR_VERSION/test/e2e/common/setup-kubestellar-ocp.sh)
+   bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/release-$KUBESTELLAR_VERSION/test/e2e/multi-cluster-deployment/use-kubestellar.sh) --env ocp
+fi
```
MikeSpreitzer, line 60: What should happen if `$env` is neither "kind" nor "ocp"? I think that a silent nothing is not a good answer. I think that an error message would be better.

test/e2e/common/cleanup.sh
dumb0002: Fixed

test/e2e/common/cleanup.sh
dumb0002: if the flag --env is missing, it would default to use its default parameter "kind". The e2e test for kind is already using the same script without providing any flags

test/e2e/multi-cluster-deployment/run-test.sh
dumb0002: Fixed

test/e2e/multi-cluster-deployment/run-test.sh
dumb0002: Fixed - added an error message

test/e2e/multi-cluster-deployment/run-test.sh
dumb0002: This e2e test is focused to provide support to test KS releases on OCP - not a local copy.

ezrasilvera: @dumb0002 - correct. The testing that supports the on going development (i.e., tests that run on latest main for PR validation) are planed to be running only through the automatic github actions. 
Once we have the OCP tests hooked into the automatic CICD and decide we want to have OCP validation also for the ongoing development we can improve the tests to supprt that as well.

ezrasilvera:
Thanks @dumb0002 . I put the LGTM.  I think you address all comments.  I will let you decide if/when to add the `approve`
/lgtm

test/e2e/multi-cluster-deployment/run-test.sh
MikeSpreitzer: @dumb0002 : by "local copy" I mean the code copy, not where the containers run. The E2E tests have always been able to test a contributor's modified version of KubeStellar. This PR breaks that.

PR #2049, opened 2024-04-12
✨ Add additional validations for KubeStellar CRDs

### LINKED ISSUE #1840: bug:  no validation is present in bindingpolicy CRDs for label selectors
### Describe the bug

It is currently possible to create bindingpolicies with invalid labels in label selector. That can be detected only looking at the kubestellar-controller-manager logs:

```
E0301 13:34:57.225776 1 controller.go:424] error matching selectors: values[0][name]: Invalid value: "edgeplatform-test-wec1, edgeplatform-test-wec2": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue', or 'my_value', or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

### Steps To Reproduce

1. create a binding policy as follows:

```
kind: Placement
metadata:
  name: nvidia-gpu-placement
spec:
  clusterSelectors:
  - matchLabels:
      name: edgeplatform-test-wec1, edgeplatform-test-wec2
  downsync:
  - objectSelectors:
    - matchLabels:
        app.kubernetes.io/part-of: wec-gpu
  wantSingletonReportedState: true
 ``` 
 
 2. look at kubestellar-controller-manager logs

### Expected Behavior

Should validate and prevent user to create such binding policy.

### Additional Context

A similar experiment with a deployment shows that there is validation for label selectors:

Apply a deployment as follows:

```
piVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx, nginx2
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80

(⎈|kind-kubeflex:N/A)$ k apply -f ~/Desktop/test1.yml 
The Deployment "nginx-deployment" is invalid: 
* spec.selector.matchLabels: Invalid value: "nginx, nginx2": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
* spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app":"nginx, nginx2"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: invalid label selector
```

### PR DESCRIPTION
**Pull request by waltforme**
## Summary
This PR adds more validations into the KubeStellar CRDs' `openAPIV3Schema`, in addition to those validations that created by controller-gen. Details are being discussed in #1840.

## Related issue(s)

Fixes #1840

### CODE REVIEW

**Comment by pdettori** [comment]
@waltforme the proposed solution looks good to me. I will wait to see if @MikeSpreitzer has additional comments.

File: config/crd/components/all/more_validations_in_bindingpolicies.yaml [hunk 1]
```diff
@@ -0,0 +1,12 @@
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/clusterSelectors/items/properties/matchLabels/additionalProperties/pattern
+  value: "^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$"
```
MikeSpreitzer, line 3: This does not express the length restriction.

config/crd/components/all/more_validations_in_bindingpolicies.yaml
```diff
@@ -0,0 +1,12 @@
...
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/clusterSelectors/items/properties/matchExpressions/items/properties/key/pattern
+  value: "^([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]$"
```
MikeSpreitzer, line 6: This should be the restriction for label names, not label values.

pkg/crd/crds_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+var bp_good_matchLabels = `
+apiVersion: control.kubestellar.io/v1alpha1
+kind: BindingPolicy
+metadata:
+  name: nvidia-gpu-placement
+spec:
+  clusterSelectors:
+  - matchLabels:
+      name: edgeplatform-test-wecs
```
MikeSpreitzer, line 154: The "good" cases should be more challenging. Reviewing https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set, I suggest something like the following.

```
  - matchLabels: {"sub.domain.test/0.9-5_Xy": "This-is_1.test"}
```

MikeSpreitzer:
This feels more complicated than necessary. There are so may layers of tweaking here that it is hard to follow.

I think that it would all get much simpler if pkg/crd could read the (combined) output of `kustomize build` rather than require the Makefile and kustomize files to produce individual CRD files.

The subtractive approach is buggy, now that we have three CRDs but only one is subtracted. It could be fixed by subtracting two, but I think it would be simpler going forward to use three positives rather than three all-subtract-two.

MikeSpreitzer:
Also, it is not obvious to me why we need `config/crd/overlays/all` when the functionality of `config/crd` is already to yield all of them.

waltforme:
The latest force-push (to https://github.com/kubestellar/kubestellar/commit/91f47b4b8fb844e2419ec4067d37e186910f92bf)
- switches to an approach to kustomize and parse single yaml file (for all CRDs), instead of multiple yaml files (each file for one CRD);
- reverts Kustomization's `components` and `overlays`;
- rebases onto latest main and squashes commits.

pkg/crd/crds_test.go
waltforme: Changed as suggested.

config/crd/components/all/more_validations_in_bindingpolicies.yaml
waltforme: Yes. Label names are validated here. And they are validated using the same regex as the one used by the apiserver.
 
```
The Deployment "nginx-deployment" is invalid: 
* spec.selector.matchExpressions[0].key: Invalid value: "-tier": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')
* spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app":"nginx"}, MatchExpressions:[]v1.LabelSelectorRequirement{v1.LabelSelectorRequirement{Key:"-tier", Operator:"In", Values:[]string{"cache", "local"}}}}: invalid label selector
```

config/crd/components/all/more_validations_in_bindingpolicies.yaml
waltforme: I don’t see a way to accurately check the length of the value using regex.
 
Also, this regex here is the same one used by the apiserver to validate label values, and the apiserver checks the length before and in addition to using the regex: https://github.com/kubernetes/kubernetes/blob/ebc1a7b7fb52f30021b7e66f77070dac8e6b839b/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L166-L176

waltforme:
> This feels more complicated than necessary. There are so may layers of tweaking here that it is hard to follow.
> 
> I think that it would all get much simpler if pkg/crd could read the (combined) output of `kustomize build` rather than require the Makefile and kustomize files to produce individual CRD files.
> 
> The subtractive approach is buggy, now that we have three CRDs but only one is subtracted. It could be fixed by subtracting two, but I think it would be simpler going forward to use three positives rather than three all-subtract-two.

Changed as suggested.

waltforme:
> Also, it is not obvious to me why we need `config/crd/overlays/all` when the functionality of `config/crd` is already to yield all of them.

Removed the Kustomize `overlays` and `components`.

config/crd/components/all/more_validations_in_bindingpolicies.yaml
MikeSpreitzer: O crud, I see that Go regex does not have an "AND" constructor. So indeed you can not write a regex that checks length AND syntax.

config/crd/kustomization.yaml
```diff
@@ -17,5 +17,13 @@ patchesStrategicMerge:
 #+kubebuilder:scaffold:crdkustomizecainjectionpatch
 
 # the following config is for teaching kustomize how to do kustomization for CRDs.
-configurations:
-- kustomizeconfig.yaml
+# configurations:
+# - kustomizeconfig.yaml
```
MikeSpreitzer, line 21: Don't we still need to "teaching kustomize how to do kustomization for CRDs" ?

pkg/crd/crds_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+var bp_good_matchExpressions = `
+apiVersion: control.kubestellar.io/v1alpha1
+kind: BindingPolicy
+metadata:
+  name: nvidia-gpu-placement
+spec:
+  clusterSelectors:
+  - matchExpressions:
+    - key: model
```
MikeSpreitzer, line 228: I recommend a more challenging key. See what https://v1-28.docs.kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set says about allowable label names.

MikeSpreitzer:
Yes, this looks simpler. You might want to squash, to simplify the Git history here too.

config/crd/components/all/more_validations_in_bindingpolicies.yaml
waltforme: Turns out that the key's _prefix_ should also be validated. Added this to the regex. Added/modified unit test cases for the key's prefix.

config/crd/kustomization.yaml
waltforme: Reverted the changes here.

(I doubt we still need this or not, so I was experimenting and saw no diff with or without it.)

pkg/crd/crds_test.go
waltforme: Done.

config/crd/patches/more_validations_in_bindingpolicies.yaml
```diff
@@ -0,0 +1,12 @@
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/clusterSelectors/items/properties/matchLabels/additionalProperties/pattern
+  value: "^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$"
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/clusterSelectors/items/properties/matchExpressions/items/properties/key/pattern
+  value: ^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]$
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/clusterSelectors/items/properties/matchExpressions/items/properties/operator/pattern
+  value: "^(In|NotIn|Exists|DoesNotExist)$"
```
MikeSpreitzer, line 9: This could be done more simply and directly with https://json-schema.org/understanding-json-schema/reference/enum instead. Not a blocker for this PR.

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

Issue #2017, opened 2024-04-04
feature: Combined status from multiple WECs


ezrasilvera:
Support getting workload status when deploying to multiple WECs.  The user should be able to define what should be returned and if/how to summarize it.

There are few modes that seems important:
1. Direct copy: Copy the statuses from all wecs into WDS. This is probably reasonable only for a small number of WECs
2. Selected fields: Copy the statues from all WECs but only for selected filed(s) 
3. Apply some summarization operator over the statuses from all WECs

(Summarization is a subitem of this feature)

A design proposal from @vMaroon :
https://docs.google.com/document/d/1kkJsdX7hfUzLl88i5t-i_Tcn30wDjOeM

A design proposal from @MikeSpreitzer : #2055

vMaroon:
@ezrasilvera can we move this to high priority?

pdettori:
This feature might be helpful for the Argo CD integration. A summary object may be shown in the application graph to get aggregate information about status of deployments on the WECs. I expect also this to be valuable for getting aggregate status of the deployment of the stack in the data pre-processing use case that @clubanderson has been working on.

shmuelk:
The "admin web app" I mentioned is part of an application that we are working on that uses KS. It makes it easier for users to work with our application.

ezrasilvera:
Moving to high priority due to @shmuelk input.

MikeSpreitzer:
See #567 for a proposal for the legacy system.

MikeSpreitzer:
I do not want to forget about scale. I think that what gets returned into the WDS needs to be something that "fits" into the WDS. There are a few things that can reduce status volume to fit into the WDS:

- use aggregation functions with constant-size outputs (like the ones in SQL). For a non-example: set union is technically an aggregation function, but its output is not constant sized.
- use what is called "projection" in relational algebra (and appears after the keyword "SELECT" in SQL) to filter the content of each workload object.
- cap the size of accumulated lists/sets. This can be practical when the concern is looking for objects "in trouble" because of bugs; in this usage it suffices to find a few WECs that exhibit the bug(s).

I think that if a user (client) wants to see the full reported state of a given object from every WEC AND the number of WEC may be large then we need to have an interface that does not require making a copy of all that data in the WDS.

pdettori:
Since there are many possible features we could offer here, it would be helpful to have more specific requirements for the known use cases - @shmuelk would you have an example of how a summarized status would look like for your use case ?

[...]

PR #2055, opened 2024-04-15
✨ Start drafting a proposal for combining reported state

MikeSpreitzer:
## Summary
This is one proposal for an API for addressing #2017. See the changes in `types.go`, and the new doc file (which should be folded into the user guide, which should exist).


ezrasilvera:
This should be included and be done as part of the overall design of 2017.  
As we agreed, for all such features, we first review and agree on the requirements, we then create a design proposal (@vMaroon is doing that for this  2017) , and finally we have an implementation PR. 
@MikeSpreitzer You are welcomed to add comments and proposals to the design doc @vMaroon created and we can/should discuss it there.  
There is no point in splitting the design into two proposals.

docs/content/direct/combined-status.md
```diff
@@ -0,0 +1,148 @@
...
+
+1. The specification of how to combine reported state appears in the BindingPolicy objects right next to the criteria for selecting workload objects. This saves users the trouble of having to write selection criteria twice.
+
+1. The combined reported state appears in a new kind of object, one per (workload object, BindingPolicy object) pair.
+
+1. A user can request a list without aggregation, possibly after filtering, but certainly with a limit on list length. The expectation is that such a list makes sense only if the length of the list will be modest. For users that want access to the full reported state from each WEC for a large number of WECs, KubeStellar should have an abstraction that gives the users access --- in a functional way, not by making another copy --- to that state (which is already in the mailbox namespaces).
```
effi-ofer, line 17: Is the mailbox still a concept now that we replaced IMBS with ITS?

MikeSpreitzer: We still have mailbox namespaces in the ITS. Changing the term from IMBS to ITS does not change that. Also, it does not change the fact that the mailbox namespaces are an internal implementation detail.

MikeSpreitzer:
@ezrasilvera: take this in the spirit of a design document that includes some Go code. See how there is markdown here too?

vMaroon:
@MikeSpreitzer I think @ezrasilvera's point is that I had already started working on this issue, and have asked to work on it in a live meeting before the bug-fixing break, and again after ezra made the features doc.

Are we going to both work on this in parallel and have the team vote? Are we going to merge into a single solution? Do you reject the proposal I made or did you make this independently? 

I think we should sync.

MikeSpreitzer:
@vMaroon: I assumed that it was OK for other people to also submit design proposals. As you can see, I have read your design proposal and made some comments on it. I think that we are in conversation. This proposal addresses comments that I made on the other one and adds a few ideas of its own. I think that the clearest way to show those ideas is with an API and examples.

MikeSpreitzer:
I have marked this PR as a "draft", trying to make the intent clearer. This is meant as a part of a design discussion, showing ideas of what the API might look like.

MikeSpreitzer:
Apparently GitHub 'draft' status means "not for review". That seems not quite right, I am asking people to consider the ideas here. I will mark it as "hold" because I am not asking this to be merged now.

/hold

vMaroon:
@MikeSpreitzer matching SQL SELECT functionality is smart and is a capability I wanted to achieve in iterations, going for it directly is sensible and is a very powerful capability that can cover pretty much any use-case or user requirement.
However, I believe the current API might be overly complex and too narrowly focused on direct SQL analogies—such as the use of terms like 'rows' and 'columns'. Simplifying this API to make it more intuitive and less reliant on SQL expertise would be beneficial.

The flexibility I wanted to achieve in my proposal can be started as groundwork provided by a non-invasive extension to your proposal:
`StatusReturn` could be streamlined as suggested, then repurposed into a standalone API object that users can define and reference by name. Additionally, we would permit the direct inlining of this API within BindingPolicies. This dual capability would satisfy both simple and complex use-cases.
This way, status collection intents can be declared once and reused in different BindingPolicies, for different object selectors, while specific intents can be inlined.

By enabling the latter, we allow an environment where common-use templates can be provided out-of-the-box, developed by the community and developed by users in their specific uses. This also opens up the path for future higher-level APIs that could apply status collection intents to multiple BindingPolicies, as I tried to initiate in the StatusPolicy proposal.

I think it makes sense to proceed with this "third" proposal, focusing initially on developing a user-friendly API for StatusReturn. What do you think?

docs/content/direct/combined-status.md
```diff
@@ -0,0 +1,148 @@
...
+### Relation with SQL
+
+To a given workload object, the user can bind one or more named "status combiners". Each status combiner is analogous to an SQL SELECT statement that does the following things.
+
+1. The SELECT statement has one input table, and that has a row per WEC that the object is prescribed to propagate to.
+
+1. The SELECT statement can have a WHERE clause that filters out some of the rows.
+
+1. The SELECT statmement either does aggregation or does not. In the case of not doing aggregation, the SELECT statement simply has a collection of named expressions defining the columns of its output.
+
```
pdettori, line 30: s/statmement/statement/

clubanderson:
This design PR has been open as WIP for 21 months. To keep the repository clean while preserving the discussion, this has been converted to a GitHub Discussion.

**Continue the conversation here:** https://github.com/kubestellar/kubestellar/discussions/3620

Thank you @MikeSpreitzer for the design work. The discussion can be revisited when there's renewed interest in implementing this feature.

PR #2056, opened 2024-04-16
🐛 Clearer logging of sync outcome

### LINKED ISSUE #2053: bug: transport controller log message about transient problem is too alarming
### Describe the bug

When the transport controller runs into a transient problem working on an item from the work queue, an **error** is logged --- with no clue that this is a transient problem that will be retried. This confuses users, see #2050 for an example.

### Steps To Reproduce

See #2050

### Expected Behavior

The log message indicates that the problem is transient and that the operation will be retried. Users are not confused into thinking there is a fatal problem here.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR updates the way that the generic transport controller code logs the outcome of syncing one object reference. Now there is distinct logging for an unrecoverable error vs. one that causes a retry.

## Related issue(s)

Fixes #2053

### DIFF
--- a/pkg/transport/generic_transport_controller.go
+++ b/pkg/transport/generic_transport_controller.go
@@ -244,16 +244,30 @@ func (c *genericTransportController) processNextWorkItem(ctx context.Context, wo
 		return false
 	}
 
-	if err := c.process(ctx, obj); err != nil {
-		c.logger.Error(err, "failed to process Binding object", "objectKey", obj, "workerID", workerID)
-	} else {
-		c.logger.Info("synced Binding object successfully.", "objectKey", obj, "workerID", workerID)
-	}
-
+	var err error
+	var retry bool
+	defer func() {
+		if err == nil {
+			// If no error occurs we Forget this item so it does not
+			// get queued again until another change happens.
+			c.workqueue.Forget(obj)
+			c.logger.Info("Synced Binding object successfully.", "objectKey", obj, "workerID", workerID)
+		} else if retry {
+			c.workqueue.AddRateLimited(obj)
+			c.logger.Info("Encountered transient error while processing Binding object, will retry later", "objectKey", obj, "workerID", workerID, "err", err)
+		} else {
+			c.workqueue.Forget(obj)
+			c.logger.Error(err, "Failed to process Binding object", "objectKey", obj, "workerID", workerID)
+		}
+	}()
+	err, retry = c.process(ctx, obj)
 	return true
 }
 
-func (c *genericTransportController) process(ctx context.Context, obj interface{}) error {
+// process works on one object reference from the work queue.
+// If the returned error is not nil then the returned boolean indicates
+// whether to retry.
+func (c *genericTransportController) process(ctx context.Context, obj interface{}) (error, bool) {
 	// We call Done here so the workqueue knows we have finished processing this item.
 	// We also must remember to call Forget if we do not want this work item being re-queued.
 	// For example, we do not call Forget if a transient error occurs, instead the item is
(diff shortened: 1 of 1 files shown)


pkg/transport/generic_transport_controller.go
```diff
@@ -244,16 +244,30 @@ func (c *genericTransportController) processNextWorkItem(ctx context.Context, wo
...
+	var retry bool
+	defer func() {
+		if err == nil {
+			// If no error occurs we Forget this item so it does not
+			// get queued again until another change happens.
+			c.workqueue.Forget(obj)
+			c.logger.Info("Synced Binding object successfully.", "objectKey", obj, "workerID", workerID)
+		} else if retry {
+			c.workqueue.AddRateLimited(obj)
+			c.logger.Info("Encountered transient error while processing Binding object, will retry later", "objectKey", obj, "workerID", workerID, "err", err)
```
ezrasilvera, line 257: Maybe move to higher log level ?   We normally don't care about these transient errors.

MikeSpreitzer: I think that there are a lot of logging level adjustments to be made. I think that it will be best to do that after I finish all the plumbing of control over the logging level.

At the current stage of development I recommend consistency with the level of the other Info message about the outcome of `c.process(ctx, obj)`.

ezrasilvera: I'm not sure regular user seeing such error will not still be alarmed, it doesn't directly implied here that it is an expected behavior.

nirrozenbaum:
@MikeSpreitzer not specifically in this PR but in general - I think you're adding too many logging that most of them are not relevant. nobody cares that an event is transient or that a retry will be done later. k8s everything is eventual consistency, so if it should be retried to bring the system to the desired state it will be retried later. this is obvious for a k8s user.
so much logging make the logs very noisy and very hard to follow.
 
personally, when I run tests to make sure a change is working, I'm usually putting a log running in a window and following what's happening. but with so much logging that's almost impossible to follow what happens.
in addition, it also makes the code very long and much less readable.

MikeSpreitzer:
@nirrozenbaum , @ezrasilvera: I think your comments are in opposite directions regarding what to expect from a log reader: @nirrozenbaum says it is OK to assume understanding of how Kubernetes controllers generally work, @ezrasilvera's comment seems (to me) to suggest the opposite. I have made a revision that tries to make you both happy, raising the level of that message and making it more verbose about its meaning.

ezrasilvera:
> The whole point of leveled logging is to give a way to resolve the tension between wanting to be able to debug and wanting to not overwhelm the reader normally. We have lacked control over the log level, and so the code has evolved with little attention to log level and being forced to sit between a rock and a hard place. In separate PRs I am trying to plumb control over the log level all the way through, and my expectation is that after that there will be a round of revision of log statements that will raise the level of many of them.
> 
> @nirrozenbaum , @ezrasilvera: I think your comments are in opposite directions regarding what to expect from a log reader: @nirrozenbaum says it is OK to assume understanding of how Kubernetes controllers generally work, @ezrasilvera's comment seems (to me) to suggest the opposite.
> 
> I have made a revision that tries to make you both happy, raising the level of that message and making it more verbose about its meaning.

I think we said the same , I simply wasn't brave enough to suggest removing it completely :-)

ezrasilvera:
@MikeSpreitzer What's the default info log level  we use today ?   If it's (4)  then the fix will not have real impact to suppress messages .

MikeSpreitzer:
@ezrasilvera: Currently the default for the transport controller is `-v=4`. I think that both you and @nirrozenbaum are arguing that this is too high, and I do not disagree. I want to focus on fully plumbing control over the `-v` setting. I think that `V(4)` is a good setting for the particular log statement that we are discussing.

MikeSpreitzer:
It would also be good to have a discussion and general rough agreement on how to choose the level for a given log statement. But that is a separate matter, and it can not be fruitfully pursued until we have all the needed control over `-v`.

ezrasilvera:
> It would also be good to have a discussion and general rough agreement on how to choose the level for a given log statement. But that is a separate matter, and it can not be fruitfully pursued until we have all the needed control over `-v`.

I agree that an overall review of log level definition across our code is needed. However in this specific case, I think the "log level" discussion was technical details. The main point was - both @nirrozenbaum and I don't think this messages should be displayed in the log (Nir say "Never"  while I say "By default")
The reason I mentioned log level was that I don't mind leaving the line there for rare cases someone would need this info and currently the log level is our only way to leave that line while making sure it is not printed in normal/default running.

PR #2057, opened 2024-04-16
✨ Start restoring/updating JSONPath package

MikeSpreitzer:
## Summary
This PR adds an implementation of JSONPath that can support removing content identified by a JSONPath.

It is common to use JSONPath to _read_ a fragment of a document. It is less common to use JSONPath to _write_ to a fragment, even less common to use a JSONPath to _delete_ a fragment. I did not find an existing package that was well maintained and much help with deleting fragments and reasonably efficient (i.e., not require serializing and deserializing the whole document for every single deletion).

## Related issue(s)

This is part of implementing #1976


pdettori:
@MikeSpreitzer have you considered using [sjson](https://github.com/tidwall/sjson) ? it seems to provide all the features we need, including delete.

MikeSpreitzer:
@pdettori: The legacy KubeStellar uses sjson. The problem with sjson is that its input and output is a string (in JavaScript Object Notation). For our purposes we want to work with data (as produced by encoding/json.Unmarshal) rather than whole-document-as-string.

MikeSpreitzer:
The transport controller is already using `unstructured.Unstructured` to represent a workload object. This holds unmarshaled data, not a single JSON-encoded string.

MikeSpreitzer:
Also, sjson does not handle filter queries (the `[? <test> ]` syntax), which is necessary to cope with the Kubernetes love of representing a map by a list of records.

pdettori:
Perhaps https://github.com/evanphx/json-patch is another option - it allows to do things such as remove (or replace) fields based on a JSONPath - see https://github.com/evanphx/json-patch?tab=readme-ov-file#create-and-apply-a-json-patch. It looks also that json-patch is widely adopted  in lots of projects, including Kubernetes - see https://github.com/search?q=github.com%2Fevanphx%2Fjson-patch&type=code



Paolo Dettori
  [< 1 minute ago](https://ibm-research.slack.com/archives/C031BHTG2LB/p1713387242407579?thread_ts=1713368869.884059&cid=C031BHTG2LB)
BTW, json-patch is used in a lot of projects, including Kubernetes - see https://github.com/search?q=github.com%2Fevanphx%2Fjson-patch&type=code

MikeSpreitzer:
JSON Patch is overkill and underkill. It does more than we need. It can not express picking the list element with a certain key.

See #2058 for a simpler alternative to this PR.

pdettori:
Since #2058 seems to be a simplified version of this PR, do we still need this one ?

MikeSpreitzer:
Closing this one because we seem to be preferring #2058

PR #2058, opened 2024-04-17
✨ Start restoring/updating JSONPath package, simplified

MikeSpreitzer:
## Summary
This PR adds a very stripped down version of JSONPath. This is an alternative to #2057 .

## Related issue(s)

This is part of addressing #1976


pkg/jsonpath/eval.go
```diff
@@ -0,0 +1,82 @@
...
+// The order in which result nodes are produced is not consistent with the RFC.
+
+// JSONValue is something that can be produced by encoding/json.Unmarshal into a pointer
+// to a nil `any`.
+// That is: `bool`, `float64`, `string`, `nil`, `[]any`, or `map[string]any` --- where those
+// nested `any` have the same restriction.
+type JSONValue = any
+
+// Node is a JSON document node.
+// The other methods MUST NOT be invoked after Remove is invoked.
```
ezrasilvera, line 31: You mean there is no  checking for the `nil`   value?  What happen if we do call it after Remove ?

pkg/jsonpath/eval.go
```diff
@@ -0,0 +1,82 @@
...
+limitations under the License.
+*/
+
+package jsonpath
+
+// This file implements JSONPath querying almost as described in RFC 9535.
+// This file does not implement the function extensions.
+// The only comparisons implemented here are equality and inequality.
+// The order in which result nodes are produced is not consistent with the RFC.
+
```
pdettori, line 23: could this lead to unexpected behavior for the users expecting RFC-compliant ordering ?

pkg/jsonpath/eval.go
```diff
@@ -0,0 +1,82 @@
...
+func (fn FieldNode) Remove() {
+	delete(fn.Object, fn.Key)
+}
+
+// QueryValue applies `query` to `node`, invoking `yield` on each
+// of the nodes that the query produces, in a context where the document
+// root is `root`.
+func QueryValue(query Query, node Node, yield func(Node)) {
+	for _, fieldName := range query {
```
pdettori, line 74: It may be useful to have a validation of the query - what happens for a malformed query ?

pkg/jsonpath/eval.go
```diff
@@ -0,0 +1,82 @@
...
+		obj, ok := node.Get().(map[string]any)
+		if !ok {
+			return
+		}
+		node = FieldNode{obj, fieldName}
+	}
+	yield(node)
+}
```
pdettori, line 82: More general comment - can we expect scenarios where multiple goroutines were to call Get or Remove on shared nodes?  it could lead to race conditions

pkg/jsonpath/lexer.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	err := lxr.advance()
+	return lxr, err
+}
+
+func (lxr *Lexer) ParseQuery() (Query, error) {
+	query := Query{}
+	if lxr.chr != '$' {
+		return query, fmt.Errorf("syntax error at %d: missing root identifier (dollar sign)", lxr.chrPos)
+	}
```
pdettori, line 74: It does not  seem to handle queries  with an empty string).  Should probably check for that.

pkg/jsonpath/eval.go
MikeSpreitzer: These Query values can not be malformed.

pkg/jsonpath/eval.go
MikeSpreitzer: For simplicity, this level of development does not address that use case.

pkg/jsonpath/lexer.go
MikeSpreitzer: The empty string is not a valid JSONPath query, and the code will reject that.

pkg/jsonpath/eval.go
MikeSpreitzer: Oops, forgot to update that comment. Fixed in the latest push.

pkg/jsonpath/lexer.go
MikeSpreitzer: The latest revision includes a test for this case.

ezrasilvera:
@MikeSpreitzer (cc: @pdettori ) Although @pdettori resolved the comment about concurrency, I'm a little concerned about future usage.  We never know who and when someone decide to use these functions and valuate the  concurrency assumption. It's relatively hard sometime to find such bugs.
Can we add disclaimers on the file header with the limitations/assumptions?

MikeSpreitzer:
@ezrasilvera, @pdettori: the latest revision has disclaimers about concurrent usage.

PR #2059, opened 2024-04-18
✨ Add CustomTransformStatus and printer columns

MikeSpreitzer:
## Summary
This PR fixes an oversight in the CustomTransform object type, adding a status section. The spec has strings that can have syntax errors, so we need a place to report them; the status section provides that.

This PR also defines custom printer columns for CustomTransform, whose absence was another oversight.

## Related issue(s)

Fixes #


api/control/v1alpha1/types.go
```diff
@@ -268,11 +268,15 @@ type BindingList struct {
 // +genclient:nonNamespaced
 // +kubebuilder:object:root=true
 // +kubebuilder:resource:scope=Cluster,shortName={ct},categories={all}
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="WGROUP",type="string",JSONPath=".spec.apiGroup"
```
ezrasilvera, line 272: What's the W stands for ?

MikeSpreitzer: "Workload". Something is needed to distinguish from the object's own attributes.

pdettori: And, can we use RESOURCE instead of WRSC ? We do not have a lot of columns and perhaps clarity is better than saving columns space.

MikeSpreitzer: I think it is confusing to have columns labeled "GROUP" and "RESOURCE" that do **NOT** refer to the object's own GROUP and RESOURCE. Remember that these objects do indeed have their own GROUP and RESOURCE. That is why I think we need some distinguishing information in the column headers.

MikeSpreitzer: What do you think of making these columns say "SUBJECT_GROUP" and "SUBJECT_RESOURCE"?

MikeSpreitzer: or "WORKLOAD_GROUP" and "WORKLOAD_RESOURCE"?

api/control/v1alpha1/types.go
```diff
@@ -268,11 +268,15 @@ type BindingList struct {
...
+// +kubebuilder:printcolumn:name="WRSC",type="string",JSONPath=".spec.resource"
 type CustomTransform struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec CustomTransformSpec `json:"spec,omitempty"`
+	Spec   CustomTransformSpec   `json:"spec,omitempty"`
```
vMaroon, line 278: Sorry I was out the week this was added. 3 points:

1. What do you think about making this API's name consistent with other APIs we have, and with similar APIs in other projects, by adding a Policy postfix?
2. Why are we deviating from the resource selectors present in other places? I think it would be ideal to have a uniform API for resource selection, across API resources that require the capability.
3. I understand that the focus on providing only a `Remove` option is because there is already a user requirement for it, but it's strange that we don't refer to adding or updating anywhere. Do you think it can be added to the API and implemented later? Or at least have the intention documented.

MikeSpreitzer: 1. For "BindingPolicy" vs "Binding", I see a functional need for putting "Policy" in the name of one of the types. For CustomTransform, I do not see a functional need. I also think that just those two words convey the meaning pretty well. In mathematics, for example, we have things like "Fourier Transform". You may think of it as a policy for changing mathematical objects, but it is not necessary to say "Fourier Transform Policy".
2. We considered that earlier and decided to start simple and add complexity later if we found the need.
3. Yes, we explicitly decided to start simple with something that can be incrementally extended later if the need arises.

MikeSpreitzer:
The force-push to ec45e986b renames those printer columns.

api/control/v1alpha1/types.go
pdettori: I would prefer "SUBJECT_GROUP" and "SUBJECT_RESOURCE". WORKLODAD_ is fine but we may get stuck debating if an object to deliver is a workload or a configuration.

MikeSpreitzer: @vMaroon : can you live with "SUBJECT_GROUP" and "SUBJECT_RESOURCE"?

vMaroon: Yes.

api/control/v1alpha1/types.go
ezrasilvera: (1) @vMaroon I think Policy is not a good fit here. @MikeSpreitzer Why not ResourceTransform to indicate what do we transforming ?   Do we have a non "Custom" transformation somewhere that this needs to be distinctly different ?  
(3) I mentioned before, I think we do need to add comment on that in our docs when we document this new feature. We all agree that for the release we are going to have this feature documented.

MikeSpreitzer: @ezrasilvera: thank you for your LGTM on #2045, which defined the API here. I hope we do not have to rehash the "Kind" defined there now.

Yes, I think we are all agreed that documentation is a must-have.

PR #2060, opened 2024-04-18
✨ Implement custom transforms

MikeSpreitzer:
## Summary
This PR adds the behavior intended for CustomTransform objects.

This PR fixes an oversight in #2045, which neglected to update the API object type registration func to include the new object types.

## Related issue(s)

This finishes the implementation part of #1976 . The documentation and some testing remain.

### DIFF
--- a/pkg/transport/custom_transform_collection.go
+++ b/pkg/transport/custom_transform_collection.go
@@ -1,0 +1,216 @@
+/*
+Copyright 2024 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package transport
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+
+	"github.com/kubestellar/kubestellar/api/control/v1alpha1"
+	controlclient "github.com/kubestellar/kubestellar/pkg/generated/clientset/versioned/typed/control/v1alpha1"
+	"github.com/kubestellar/kubestellar/pkg/jsonpath"
+)
+
+// customTransformCollection maintains the digested custom transformation instructions
+// for each metav1.GroupResource.
+type customTransformCollection struct {
+	// client is here for updating status in a CustomTransform
+	client controlclient.CustomTransformInterface
+
+	// getTransformObjects is the part of the CustomTransform informer's cache.Indexer behavior
+	// that is needed here, for using the index named in `customTransformDomainIndexName`.
+	getTransformObjects func(indexName, indexedValue string) ([]any, error)
+
+	// enqueue is used to add a reference to a Binding that needs to be re-processed because
+	// of a change to a CustomTransform that the Binding is sensitive to.
+	enqueue func(any)
+
+	// mutex must be locked while accessing the following fields or their contents
+	mutex sync.Mutex
+
+	// grToTransformData has an entry for every GroupResource that some Binding cares about
+	// (i.e., lists an object of that GroupResource), and no more entries.
+	grToTransformData map[metav1.GroupResource]*groupResourceTransformData
+
+	// ctNameToSpec holds the Spec last processed for each CustomTransform
+	ctNameToSpec map[string]v1alpha1.CustomTransformSpec
+
+	// bindingNameToGroupResources tracks the set of GroupResource that each Binding
+	// references. This is so that when the set for a given Binding changes,
+	// for the GroupResources that are no longer in the set, the Binding's Name can
+	// be removed from groupResourceTransformData.bindingsThatCare.
+	bindingNameToGroupResources map[string]sets.Set[metav1.GroupResource]
+}
+
+// groupResourceTransformData is the ingested custom transforms for a given GroupResource
+type groupResourceTransformData struct {
+	bindingsThatCare sets.Set[string /*Binding name*/] // not empty
+	removes          []jsonpath.Query
+}
+
+func newCustomTransformCollection(client controlclient.CustomTransformInterface, getTransformObjects func(indexName, indexedValue string) ([]any, error), enqueue func(any)) *customTransformCollection {
+	return &customTransformCollection{
+		client:                      client,
+		getTransformObjects:         getTransformObjects,
+		enqueue:                     enqueue,
+		grToTransformData:           make(map[metav1.GroupResource]*groupResourceTransformData),
+		ctNameToSpec:                make(map[string]v1alpha1.CustomTransformSpec),
+		bindingNameToGroupResources: make(map[string]sets.Set[metav1.GroupResource]),
+	}
+}
+
+// Call this to process a new version, or deletion, of a CustomTransform.
+func (ctc *customTransformCollection) noteCustomTransform(ctx context.Context, name string, ct *v1alpha1.CustomTransform) {
+	ctc.mutex.Lock()
+	defer ctc.mutex.Unlock()
+	oldSpec, hadSpec := ctc.ctNameToSpec[name]
+	if ct == nil && !hadSpec {
+		return // was absent and is absent; no change
+	}
+	var oldGroupResource, newGroupResource metav1.GroupResource
+	if hadSpec {
+		oldGroupResource = ctSpecGroupResource(oldSpec)
+	}
+	if ct != nil {
+		newGroupResource = ctSpecGroupResource(ct.Spec)
+	}
+	if ct != nil && hadSpec &&
+		oldGroupResource == newGroupResource &&
+		sets.New(oldSpec.Remove...).Equal(sets.New(ct.Spec.Remove...)) {
+		return // unchanged
+	}
+	// GroupResource or Spec.Remove changed
+	logger := klog.FromContext(ctx)
+	if hadSpec {
+		// Invalidate the cached analysis for the relevant GroupResource, if there is one
+		if oldGRTD, hadGRTD := ctc.grToTransformData[oldGroupResource]; hadGRTD {
+			delete(ctc.grToTransformData, oldGroupResource)
+			for bindingName := range oldGRTD.bindingsThatCare {
+				logger.V(5).Info("Enqueuing reference to Binding because of change to CustomTranform", "bindingName", bindingName, "customTransformName", name)
+				ctc.enqueue(bindingName)
+			}
+		}
+	}
+	if ct != nil {
+		ctc.ctNameToSpec[name] = ct.Spec
+	} else {
+		delete(ctc.ctNameToSpec, name)
+	}
+}
+
+// getCustomTransformData returns the groupResourceTransformData to use
+// for the given GroupResource on behalf of the named Binding.
+func (ctc *customTransformCollection) getCustomTransformData(ctx context.Context, groupResource metav1.GroupResource, bindingName string) *groupResourceTransformData {
+	logger := klog.FromContext(ctx)
+	ctc.mutex.Lock()
+	defer ctc.mutex.Unlock()
+	grtd, ok := ctc.grToTransformData[groupResource]
+	if !ok {
+		grtd = &groupResourceTransformData{
+			bindingsThatCare: sets.New(bindingName),
+		}
+		ctKey := customTransformDomainKey(groupResource.Group, groupResource.Resource)
+		ctAnys, err := ctc.getTransformObjects(customTransformDomainIndexName, ctKey)
+		if err != nil {
+			// This only happens if the index is not defined;
+			// that is, it never happens.
+			// If it does, retry will not help.
+			logger.Error(err, "Failed to get objects from CustomTransform domain index", "key", ctKey)
+		}
+		ctNames := []string{}
+		for _, ctAny := range ctAnys {
+			ct := ctAny.(*v1alpha1.CustomTransform)
+			removes := ctc.parseRemovesAndUpdateStatus(ctx, ct)
+			grtd.removes = append(grtd.removes, removes...)
+			oldSpec, had := ctc.ctNameToSpec[ct.Name]
+			if had {
+				oldGroupResource := ctSpecGroupResource(oldSpec)
+				if oldGroupResource != groupResource {
+					oldGRTD := ctc.grToTransformData[oldGroupResource]
+					if oldGRTD != nil {
+						for _, oldBindingName := range oldGRTD.bindingsThatCare {
+							logger.V(5).Info("Enqueuing reference to Binding because CustomTransform changed its GroupResource", "customTransformName", ct.Name, "bindingName", oldBindingName, "oldGroupResource", oldGroupResource, "newGroupResource", groupResource)
+							ctc.enqueue(oldBindingName)
+						}
+					}
+				}
+			}
+			ctc.ctNameToSpec[ct.Name] = ct.Spec
+			ctNames = append(ctNames, ct.Name)
+		}
+		if len(ctAnys) > 1 {
+			logger.Error(nil, "Multiple CustomTransform objects apply to one GroupResource", "groupResource", groupResource, "names", ctNames)
+		}
+		ctc.grToTransformData[groupResource] = grtd
+	} else {
+		grtd.bindingsThatCare.Insert(bindingName)
+	}
+	return grtd
+}
+
+func (ctc *customTransformCollection) parseRemovesAndUpdateStatus(ctx context.Context, ct *v1alpha1.CustomTransform) (removes []jsonpath.Query) {
+	logger := klog.FromContext(ctx)
+	ctCopy := ct.DeepCopy()
+	ctCopy.Status = v1alpha1.CustomTransformStatus{ObservedGeneration: ct.Generation}
+	for idx, queryS := range ct.Spec.Remove {
+		query, err := jsonpath.ParseQuery(queryS)
+		if err != nil {
+			ctCopy.Status.Errors = append(ctCopy.Status.Errors, fmt.Sprintf("Error in spec.remove[%d]: %s", idx, err.Error()))
+		} else if len(query) == 0 {
+			ctCopy.Status.Errors = append(ctCopy.Status.Errors, fmt.Sprintf("Invalid spec.remove[%d]: it identifies the whole object", idx))
+		} else {
+			removes = append(removes, query)
+		}
+	}
+	ctEcho, err := ctc.client.UpdateStatus(ctx, ctCopy, metav1.UpdateOptions{FieldManager: ControllerName})
+	if err != nil {
+		logger.Error(err, "Failed to write status of CustomTransform", "name", ct.Name, "resourceVersion", ct.ResourceVersion, "status", ctCopy.Status)
+	} else {
+		logger.V(4).Info("Wrote status of CustomTransform", "name", ct.Name, "resourceVersion", ctEcho.ResourceVersion, "observedGeneration", ctCopy.Status.ObservedGeneration)
+	}
+	return
+}
+
+// Call this when the new full set is known, so that obsolete relationships can be deleted
+func (ctc *customTransformCollection) setBindingGroupResources(bindingName string, newGroupResources sets.Set[metav1.GroupResource]) {
+	ctc.mutex.Lock()
+	defer ctc.mutex.Unlock()
+	oldGroupResources := ctc.bindingNameToGroupResources[bindingName]
+	for groupResource := range oldGroupResources {
+		if newGroupResources.Has(groupResource) {
+			continue
+		}
+		// This one is being removed
+		if grtd, ok := ctc.grToTransformData[groupResource]; ok {
+			grtd.bindingsThatCare.Delete(bindingName)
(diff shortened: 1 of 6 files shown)


pkg/transport/generic_transport_controller.go
```diff
@@ -40,9 +43,12 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/kubestellar/kubestellar/api/control/v1alpha1"
+	a "github.com/kubestellar/kubestellar/pkg/abstract"
```
ezrasilvera, line 46: nit: Maybe a better name?

MikeSpreitzer:
The recent force-pushes are a rebase onto `main` and removal of a debugging statement.

pkg/transport/custom_transform_collection.go
```diff
@@ -0,0 +1,216 @@
...
+	if ct != nil {
+		ctc.ctNameToSpec[name] = ct.Spec
+	} else {
+		delete(ctc.ctNameToSpec, name)
+	}
+}
+
+// getCustomTransformData returns the groupResourceTransformData to use
+// for the given GroupResource on behalf of the named Binding.
+func (ctc *customTransformCollection) getCustomTransformData(ctx context.Context, groupResource metav1.GroupResource, bindingName string) *groupResourceTransformData {
```
vMaroon, line 122: Understanding this and other functions would be significantly easier (personally) if different logical blocks are preceded with brief descriptive comments. 
Though, in some cases a comment is overkill (such as when the logical block is small and self-explanatory) - but a new line would be ideal in helping with identifying it.

This practice, which I could not find a consistent name/documentation for, can be found everywhere in the binding controller for example.

pkg/transport/custom_transform_collection.go
```diff
@@ -0,0 +1,216 @@
...
+	// getTransformObjects is the part of the CustomTransform informer's cache.Indexer behavior
+	// that is needed here, for using the index named in `customTransformDomainIndexName`.
+	getTransformObjects func(indexName, indexedValue string) ([]any, error)
+
+	// enqueue is used to add a reference to a Binding that needs to be re-processed because
+	// of a change to a CustomTransform that the Binding is sensitive to.
+	enqueue func(any)
+
+	// mutex must be locked while accessing the following fields or their contents
+	mutex sync.Mutex
```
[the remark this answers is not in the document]
MikeSpreitzer: I think that the methods of customTransformCollection need their Lock and Unlock statements where they are, covering more than individual map actions. If the map actions _also_ tried to Lock then this would be bad.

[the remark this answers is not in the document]
MikeSpreitzer: It looks to me like all the methods of customTransformCollection that need any lock, need a write lock.

pkg/transport/custom_transform_collection.go
```diff
@@ -0,0 +1,216 @@
...
+		client:                      client,
+		getTransformObjects:         getTransformObjects,
+		enqueue:                     enqueue,
+		grToTransformData:           make(map[metav1.GroupResource]*groupResourceTransformData),
+		ctNameToSpec:                make(map[string]v1alpha1.CustomTransformSpec),
+		bindingNameToGroupResources: make(map[string]sets.Set[metav1.GroupResource]),
+	}
+}
+
+// Call this to process a new version, or deletion, of a CustomTransform.
```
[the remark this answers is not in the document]
MikeSpreitzer: I have expanded the comments and factored the biggest function.

pkg/transport/custom_transform_collection.go
MikeSpreitzer: I have revised and extended the comments and factored the biggest function.

pkg/transport/custom_transform_collection.go
```diff
@@ -0,0 +1,244 @@
...
+// GetCustomTransformData returns the groupResourceTransformData to use
+// for the given GroupResource on behalf of the named Binding.
+// This method returns a cached answer if one is available, otherwise
+// digests the relevant CustomTransform object(s) and caches the result.
+// Always records the fact that the given binding depends on the answer.
+func (ctc *customTransformCollection) GetCustomTransformData(ctx context.Context, groupResource metav1.GroupResource, bindingName string) *groupResourceTransformData {
+	logger := klog.FromContext(ctx)
+	ctc.mutex.Lock()
+	defer ctc.mutex.Unlock()
+	grtd, ok := ctc.grToTransformData[groupResource]
```
vMaroon, line 94: `grtd` constantly requires translation in head while reading, which makes building understanding harder - what do you think about `grTransformData`? Other variable names seem to follow similar structures.

MikeSpreitzer: OK, renamed.

pkg/transport/generic_transport_controller.go
```diff
@@ -975,7 +1056,7 @@ func setLabel(object metav1.Object, key string, value any) {
 }
 
 // cleanObject is a function to clean object before adding it to a wrapped object. these fields shouldn't be propagated to WEC.
-func cleanObject(object *unstructured.Unstructured) *unstructured.Unstructured {
+func (ctc *customTransformCollection) cleanObject(ctx context.Context, groupResource metav1.GroupResource, object *unstructured.Unstructured, bindingName string) *unstructured.Unstructured {
```
vMaroon, line 1059: I think we should differentiate between cleaning objects and transforming them.
The transformation should occur in a new function, and the order of first cleaning and then transforming should be preserved (as it is below).

pkg/transport/custom_transform_collection.go
```diff
@@ -0,0 +1,243 @@
...
+		// Digest each relevant CustomTransform, accumulating remove instructions in groupResourceTransforData.removes.
+		// Invalidate cache entry for each CustomTransform that changed its Spec's .Group or .Resource.
+		for _, ctAny := range ctAnys {
+			ct := ctAny.(*v1alpha1.CustomTransform)
+			removes := ctc.digestCustomTransformLocked(ctx, groupResource, bindingName, ct)
+			grTransformData.removes = append(grTransformData.removes, removes...)
+			ctNames = append(ctNames, ct.Name)
+		}
+		if len(ctAnys) > 1 { // This is not recommended
+			logger.Error(nil, "Multiple CustomTransform objects apply to one GroupResource", "groupResource", groupResource, "names", ctNames)
```
vMaroon, line 118: This does not follow the use of Error across the rest of the codebase. 

Additionally, perhaps this is something we should think of. E.g., for the following: https://github.com/kubestellar/kubestellar/issues/2064#issuecomment-2067992923

Karmada has a similar feature as well, but they implement it through 2 different resources.

pkg/transport/generic_transport_controller.go
```diff
@@ -452,6 +509,23 @@ func (c *genericTransportController) syncProperties(ctx context.Context, invName
...
+		ct = nil
+	} else if err != nil { // This never happens; retry will not help
+		logger.Error(err, "Failed to Get from CustomTransformLister", "name", name)
+		ct = nil
+	}
+	c.customTransformCollection.NoteCustomTransform(ctx, name, ct)
+	return nil
+}
+
+func ctSpecGroupResource(spec v1alpha1.CustomTransformSpec) metav1.GroupResource {
```
vMaroon, line 525: This function is declared here, but used only within `custom_transform_collection.go`. Perhaps it should be moved to there?

pkg/transport/custom_transform_collection.go
```diff
@@ -0,0 +1,243 @@
...
+		ctAnys, err := ctc.getTransformObjects(customTransformDomainIndexName, ctKey)
+		if err != nil {
+			// This only happens if the index is not defined;
+			// that is, it never happens.
+			// If it does, retry will not help.
+			logger.Error(err, "Failed to get objects from CustomTransform domain index", "key", ctKey)
+		}
+		ctNames := []string{}
+
+		// Digest each relevant CustomTransform, accumulating remove instructions in groupResourceTransforData.removes.
```
vMaroon, line 109: s/groupResourceTransforData/groupResourceTransferData/

pkg/transport/custom_transform_collection.go
```diff
@@ -0,0 +1,243 @@
...
+	if oldGRTransformData, had := ctc.grToTransformData[oldGroupResource]; had {
+		delete(ctc.grToTransformData, oldGroupResource)
+		for bindingName := range oldGRTransformData.bindingsThatCare {
+			if bindingName == triggerBindingName {
+				continue
+			}
+			logger.V(5).Info("Enqueuing reference to Binding because "+reason, append(extraLogArgs, "bindingName", bindingName, "customTransformName", ctName, "oldGroupResource", oldGroupResource))
+			ctc.enqueue(bindingName)
+		}
+	} else {
```
vMaroon, line 186: While this is purely a code styling preference, nesting across the code can be reduced in this and several other functions and sub code-blocks by following the guard clause pattern, as followed in the binding controller, and was followed in the transport controller previously.

pkg/transport/generic_transport_controller.go
vMaroon: BTW, in pkg code, only test files use snake case names, while other code file names use lisp case. Both `generic_transport_controller.go` and `custom_transform_collection.go` should use lisp case, or the rest of the files should be changed to use snake case.

vMaroon:
Added some comments. Overall I think the implementation is nice.

I understand that we don't want to make updating docs a blocker, but if you think that it can be done quickly, it would be ideal (or if it really follows quickly after this PR then it's understandable). 
I think the architecture docs **diagrams** should also be updated/extended to reflect the cache implemented here, and how it is mutated.

pkg/transport/generic_transport_controller.go
```diff
@@ -394,6 +443,8 @@ func (c *genericTransportController) processNextWorkItem(ctx context.Context) bo
 // to which that matters. The string is the name of the inventory object.
 type recollectProperties string
 
+type customTransformReference string // the name of the cluster-scoped object
```
vMaroon, line 446: FYI, of all the items the workqueue supports, only the binding's key remains an unaliased string type, do you think that it should remain like that?

pkg/transport/custom_transform_collection.go
MikeSpreitzer: @vMaroon:

I am not sure I get your meaning. If your point about this Error log statement is that it is unusual to use a log statement to tell the user that they are doing something not recommended, then I agree. This message should appear in an API object. Since it is not a fatal error, I did not make the controller put it in the `status.errors` of the Binding or CustomTransform. I think that what we should do is introduce `Warnings []` in CustomTransformStatus and report it there. I was planning to get to that after this PR merges. I also wanted to have some discussion about whether we are really agreed that this is not recommended (and would say so in the to-be-written doc if agreed).

pkg/transport/generic_transport_controller.go
MikeSpreitzer: I do not think that `pkg` is or should be special in this regard. I have seen enough of both dash and underscore that I did not feel the need to try to enforce some consistency. It would feel a bit awkward to me to enforce consistent usage of the opposite of what the code generators create.

pkg/transport/custom_transform_collection.go
MikeSpreitzer: This particular bit of code is not highly nested, so I did not stress over it. Using the `if vars := expr; cond` form saves having to have a separate line to declare `oldGRTransformData`.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: I noticed that too. I left it as is, for the sake of not changing more than necessary. I think we can address this (and some other things) after this PR merges.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: We have had some discussion of terminology, perhaps not all in your presence. @ezrasilvera was strongly of the opinion that the feature should be called "transformation", and others agreed. We also agreed that we do not currently have a use case for doing more than removing content, and this is a subset of what is called "transformation".

MikeSpreitzer: The force-push to ecaf86fcc updates the name of this method and the comments on it.

pkg/transport/custom_transform_collection.go
MikeSpreitzer: Thanks, fixed.

MikeSpreitzer:
Regarding documentation updates: yes, I plan for that to quickly follow this PR. Testing updates too.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: BTW, this method should ultimately be in the other file; I left it here for now in order to spare reviewers from more verbose `diff` output than necessary.

pkg/transport/custom_transform_collection.go
vMaroon: The point was this use is unusual, yes. I guess it can be left until the API is more mature to contain errors. Do we really want warnings in the API? What qualifies a warning? A state that is fine but can go wrong? Or what?

The 2nd part of my comment refers to `if len(ctAnys) > 1 { // This is not recommended` - the whole matter. Perhaps now is not the time for it, but I think it's not insensible for use-cases of chaining transformations (in order) to appear, such as what's mentioned in the comment. Therefore if possible, the solution should not be fragile to the idea (should not support it, but should not be fragile - which it's not).

pkg/transport/generic_transport_controller.go
vMaroon: I think we should follow one pattern. If you think we should follow what the generators create, then we will. But not being uniform in this matter looks bad I believe (and personally bothers me). You don't have to do this of course, I can make a followup PR.

pkg/transport/custom_transform_collection.go
vMaroon: I thank you for your patience in reading through this detailed message beforehand:

I agree that this one in particular is simple, but my intention was to highlight that the pattern is not used. This is more bothersome in more complex functions. While the guard clause pattern may require expanding the form you wrote, its benefits overweigh this downside:

- Decreased nesting that better abides by the _Single Level of Abstraction_ principle mentioned in Clean Code, this boosts readability and understanding by avoiding mental grouping

- Consistency throughout the project: while we're multiple developers with different styles and tastes, I think it's crucial to be consistent over this pattern, which would ease the onboarding of new developers and reduce cognitive load tax

- Getting done with the short case(s) first: usually these are the fail cases, but I think a generalization of this pattern is to first go through the shorter cases, which decreases (human) memory overhead that increases in every step into a new block. When doing this, as you read the function, when you're done with the short ones you can fully focus on what's left, and assume that everything else is fine

- Easier error handling, flow control and testability (unit tests that target clear each - ideally single level - branch)

I felt the need to emphasize on the benefits of this practice since it can fall down to a matter of opinion, and obviously you are a much more seasoned developer than me. While the benefits may not seem huge in a single code iteration (~PR), if such clarity practices are not followed, time after time we risk getting into a complex, hard to read codebase.

Therefore if possible, it would be great if you can implement this in the functions added in this PR, or allow me to do it.

pkg/transport/custom_transform_collection.go
MikeSpreitzer: So the first question is whether we want to recommend that restriction.  I have opened Issue #2067 about that.

pkg/transport/custom_transform_collection.go
MikeSpreitzer: I will do that now.

MikeSpreitzer:
The force-push to 7dc41d76f re-organizes some function internal structure as discussed in review above.

pkg/transport/custom_transform_collection.go
MikeSpreitzer: Done in force-push to 7dc41d76f

pkg/transport/custom_transform_collection.go
```diff
@@ -0,0 +1,260 @@
...
+// `shouldHave` asserts that ctc.ctNameToSpec has an entry for this name.
+// triggerBindingName, if not empty, is the name of the Binding being processed when this change was noticed.
+// reason and extraLogArgs go into the debug log statements.
+// Caller asserts that the ctc's mutex is locked.
+func (ctc *customTransformCollection) invalidateCacheEntryLocked(ctx context.Context, shouldHave bool, ctName, triggerBindingName string, oldGroupResource metav1.GroupResource, reason string, extraLogArgs ...any) {
+	logger := klog.FromContext(ctx)
+	oldGRTransformData, had := ctc.grToTransformData[oldGroupResource]
+	if !had {
+		if shouldHave {
+			logger.Error(nil, "Impossible condition: ctc.ctNameToSpec has an entry for a CustomTransform but ctc.grToTransformData has no entry for the GroupResource", "customTransformName", ctName, "groupResource", oldGroupResource)
```
vMaroon, line 184: If I understand the intention, the current state of the code should never get here, but if someone updates the code and does not properly maintain the maps, this can happen.
**If this is the case, then why do we say it's impossible? It would be a misuse of maps/mutex - applies to the other instance of similar logging too.** 

--
Generally in both this and the other instance of an "impossible" error being logged, I think maintaining expected correctness as an invariant through controlling the mutability of the maps by design is better, such as through not exposing the maps/mutex to developers in this struct, but protecting them through an interface and providing functionality that does not break the invariance. Though is not a necessity or a blocker.

MikeSpreitzer: Again, it is not just the maps, the consistency invariants involve the values in the maps too.

Not expose to "developers"? This is code that developers can modify; it is unavoidably exposed to developers. I suspect you mean to suggest moving `customTranformCollection` to its own new package. That could be done. I did not do it here because it seemed heavier weight than necessary to me. I agree that we can consider this in follow-on work.

vMaroon: Technically everything is exposed to KS developers, yes.

What I meant is protecting the maps and direct mutation through moving their implementation somewhere else (can still be the same package), then abstracting the implementation through an interface.

Then the developer of custom transforms imp can use this new data structure as a black-box interface, leaving its content and actual implementation "protected".

This is the closest we can get in Go to implementing a class with private fields then using the class in the same package. As the user of the class (but still a developer in the package) - you practically can modify the class, but you leave it be and only use its exposed fields/methods because the point behind it is the abstraction of some logically related functionality.

MikeSpreitzer: If we do not trust developers of the `pkg/transform` package then introducing a Go interface will not achieve isolation --- the goofy developer can still do stuff like `.(*customTransformCollectionImpl)`.

pkg/transport/generic_transport_controller.go
MikeSpreitzer: On further reflection and doc writing, I have realized that this method is a client of the instructions cache. I re-framed it as such in today's pushes.

pkg/transport/custom_transform_collection.go
MikeSpreitzer: I have introduced such an internal interface.

pkg/transport/custom_transform_collection.go
```diff
@@ -0,0 +1,268 @@
...
+		client:                      client,
+		getTransformObjects:         getTransformObjects,
+		enqueue:                     enqueue,
+		grToTransformData:           make(map[metav1.GroupResource]*groupResourceTransformData),
+		ctNameToSpec:                make(map[string]v1alpha1.CustomTransformSpec),
+		bindingNameToGroupResources: make(map[string]sets.Set[metav1.GroupResource]),
+	}
+}
+
+// GetCustomTransformData returns the groupResourceTransformData to use
```
vMaroon, line 90: This documentation needs updating.

vMaroon (commented):
The last point I have on this implementation is that the coupling between CustomTransform dependent functions and binding-name dependent functionality can be confusing. 
I understand that the coupling leads to compaction of code, but it does confuse me. I also understand that this coupling may not be avoidable due to the consistency claimed over lock usage. 
At the very least, function names and documentation would better reflect the need and use of bindingNames, because `on behalf of binding` is not informative. If it's possible to break the coupling/mix of CT consuming functionality and binding-name consuming functionality it would be even better.

It would also be good if another reviewer comments on this point **while reviewing the code regardless** since this is my last before LGTM'ing (cc: @pdettori @ezrasilvera).

pkg/transport/custom_transform_collection.go
MikeSpreitzer: Oh crud, I forgot to push the last change. Pushed just now.

MikeSpreitzer: I pushed another commit, hopefully making the comments clearer.

The relation with Bindings is a critical part of the customTransformCollection functionality. It is only by tracking the set of relevant Bindings that the collection knows when to remove an entry for a particular GroupResource. I hope the documentation #2066 makes this clear: see the inclusion of the USES relation in the definition of the cache. At this point in time I would not recommend creating an abstraction whose only concrete realization is Binding; we can address that later if the need arises.

pkg/transport/custom_transform_collection.go
```diff
@@ -0,0 +1,279 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+
+	"github.com/kubestellar/kubestellar/api/control/v1alpha1"
+	controlclient "github.com/kubestellar/kubestellar/pkg/generated/clientset/versioned/typed/control/v1alpha1"
+	"github.com/kubestellar/kubestellar/pkg/jsonpath"
+)
+
+type customTransformCollection interface {
```
vMaroon, line 33: The (optional) comment touching on using interfaces to hide dev-mistakes-error-prone fields that require delicate handling was referring to internals of the implementation.

This is ok too - but this interface lacks documentation (docs on this interface are what would appear to the developer during use in the transport code). Once fixed I'll LGTM.

MikeSpreitzer: The force-push to c31f8e154 adds comment on the new internal interface.

vMaroon (approved):
/lgtm

PR #2066, opened 2024-04-22
📖 Document CustomTransform, fix arch doc of template expansion

MikeSpreitzer:
## Summary
This PR updates the user-guide and architecture documents, adding WEC-independent transforms. This PR also fixes some inaccuracies in the architecture document's description of how template expansion is implemented.

## Related issue(s)

Fixes #


docs/content/direct/user-guide.md
```diff
@@ -12,6 +12,62 @@ See [Using an existing hosting cluster](./hosting-cluster.md)
...
+
+1. In `metadata`, the annotation named `batch.kubernetes.io/job-tracking`
+
+1. In `metadata` _and_ in `spec.template.metadata`, the labels named `controller-uid` or `batch.kubernetes.io/controller-uid`.
+
+### Configured transformation of workload objects
+
+The user can configure additional transformations of workload objects by putting `CustomTransform` (in the `control.kubestellar.io` API group) objects in the WDS. Each `CustomTransform` object binds to certain workload objects and specifies certain transformations.
+
+Currently the binding is simply by naming the workload object's API group and "resource" name in the `Customtransform`'s `spec`. The transformations from all of the bound `CustomTransform` objects are applied to the workload object. There should be at most one `CustomTransform` object that specifies a given API group and resource.
```
vMaroon, line 53: s/Customtransform/CustomTransform/

docs/content/direct/architecture.md
```diff
@@ -476,8 +476,10 @@ When processing a reference to a `Binding` object that still exists, the transpo
...
+
+The transport controller maintains a cached set of customization properties for each destination, and an association between `Binding` and the set of destinations that it references. When a relevant informer delivers an event about an inventory object (either a `ManagedCluster` object or a `ConfigMap` object that adds properties for that destination) the controller enqueues a work item of type `recollectProperties`. This work item carries the name of the inventory object. Processing that work item starts by re-computing the full map of properties for that destination. If the cache has an entry for that destination and the cached properties differ from the ones freshly computed, the controller updates that cache entry and enqueues a reference to every `Binding` object that depends on the properties of that destination.
```
vMaroon, line 485: This paragraph touches back on parameter expansion after two unrelated paragraphs, and specifically after the one and only that mentions CustomTransforms. I understand the grouping since the one above is technical over caching, but it might make more sense to place this one after the paragraph that describes parameter expansion. I say this because I had to pause to understand that we now are back to a different customization feature.

docs/content/direct/architecture.md
```diff
@@ -476,8 +476,10 @@ When processing a reference to a `Binding` object that still exists, the transpo
...
+To construct the function from destination to customized wrapped object, the transport controller reads the `Binding`'s list of references to workload objects. The controller reads those objects from the WDS using a Kubernetes "dynamic" client. Immediately upon reading each workload object, the controller applies the WEC-independent transforms. After doing that for all the listed workload objects, the controller goes through those objects one-by-one and applies template expansion for each destination if the object requests template expansion. If any of those objects requests template expansion and has a string that actually involves template expansion: the controller accumulates a map from destination to slice of customized objects and then invokes the transport plugin on each of those slices, to ultimately produce the function from destination to wrapped object. If none of the selected workload objects actually involved any template expansion then the controller wraps the slice of workload objects to get one wrapped object and produces a constant function from destination to that one wrapped object.
```
vMaroon, line 479: Optional: s/WEC-independent transforms/WEC-independent transforms (CustomTransforms)/

or any other variation that mentions that you refer to the transformations derived from CustomTransforms.

docs/content/direct/architecture.md
```diff
@@ -476,8 +476,10 @@ When processing a reference to a `Binding` object that still exists, the transpo
...
... of the `CustomTransform` objects, the transport controller maintains a cache that maps a `metav1.GroupResource` to the result of digesting all the relevant `CustomTransform` objects. The data type `customTransformCollection` holds the cache. This cache is populated as needed in the course of transforming workload objects. To support eventually removing an entry from that cache when no longer needed, each entry also tracks the set of `Binding` objects that depend on that cache entry; when that set becomes empty, the entry is removed from the cache. In order to know when a `Binding` has changed from dependent to not dependent, the cache also tracks the set of `metav1.GroupResource` relevant to each `Binding`. Also: in order to know what to update when the Spec of a `CustomTransform` changes, the cache also tracks the spec of each `CustomTransform` that contributed to an entry in the cache. When such a `CustomTransformSpec` changes in any way, the controller invalidates the cache entry for the `GroupResource` of the last known version of that spec. Additionally, if the change affected the `GroupResource` then the controller also invlidates the cache entry for the old `GroupResource`.
```
vMaroon, line 483: While this paragraph is technically correct, I think it can greatly benefit from a diagram that shows the flow visually.

docs/content/direct/architecture.md
MikeSpreitzer: The paragraph about the caching for WEC-independent transforms precedes the paragraph about the caching for customizing to WECs because that is the order in which the two sorts of transformations are done and the order in which they are mentioned in the overall function paragraph.

We could swap the two paragraphs about caching, but with either order the second paragraph about caching is separated from the earlier words about what it is caching for.

docs/content/direct/user-guide.md
MikeSpreitzer: Thanks, fixed.

docs/content/direct/architecture.md
MikeSpreitzer: updated

docs/content/direct/architecture.md
MikeSpreitzer: I have revised to add, hopefully, more better explanation.

Rather than trying to diagram the logic in each function, I started with an explanation of the data structure that they are maintaining. I gave a relational exposition, since it is simple and that is a widely known conceptual framework. With the relations and invariants shown, and the way that the data structure represents the relations shown, the explanations of the public methods of the cache can lean on that other material to be shorter and more comprehensible (I hope).

docs/content/direct/user-guide.md
```diff
@@ -12,6 +12,62 @@ See [Using an existing hosting cluster](./hosting-cluster.md)
...
+
+The following are applied to every workload object.
+
+1. Remove the following fields from `metadata`: `managedFields`, `finalizers`, `generation`, `ownerReferences`, `selfLink`, `resourceVersion`, `UID`, `generateName`.
+1. Remove the annotation named `kubectl.kubernetes.io/last-applied-configuration`.
+1. Remove the `status`.
+
+## Built-in transformations of specific kinds of workload object
+
+In a `Service` (core API group) object, remove the following fields from `spec`: `clusterIP`, `clusterIPs`, `ipFamilies`, `externalTrafficPolicy`, `internalTrafficPolicy`, `ipFamilyPolicy`, `sessionAffinity`. Also remove the `nodePort` field from every port unless the annotation `kubestellar.io/annotations/preserve=nodeport` is present.
```
MikeSpreitzer, line 37: #2070 changes this. The second PR among these two to merged should first be updated accordingly.

PR #2068, opened 2024-04-22
✨ Add CustomTransform test

MikeSpreitzer:
## Summary
This PR updates the ginkgo-based end-to-end test suite to exercise CustomTransform functionality.

This PR is built on #2060, adding one commit. If that PR merges first then I will rebase this one.

## Related issue(s)

This is part of addressing #1976


MikeSpreitzer:
The force-push to 6dc4bcc5d is a rebase on to the latest edition of #2060

test/e2e/ginkgo/multiple_cluster_deployment_test.go
```diff
@@ -50,9 +52,16 @@ var _ = ginkgo.Describe("end to end testing", func() {
 	})
 
 	ginkgo.Context("multiple WECs", func() {
-		ginkgo.It("propagates deployment to the WECs", func() {
-			util.ValidateNumDeployments(ctx, wec1, ns, 1)
-			util.ValidateNumDeployments(ctx, wec2, ns, 1)
+		ginkgo.It("propagates deployment to the WECs while applying CustomTransform", func() {
```
ezrasilvera, line 55: I really think this should be a different It test , i.e., one with transformation and one without

MikeSpreitzer: This test already takes an annoyingly long time to run. I think it is OK for this one `It` to test multiple aspects of functionality. In fact, it is unavoidable to test multiple aspects, since the end-to-end behavior is the result of many bits of functionality working together. I do not think that the extra time to run another whole end-to-end scenario is worth any benefit of separating out the test of CustomTransform.

ezrasilvera: I wouldn't block the PR for that, but I'm very surprised to hear it. I was under the impression that the tests  themselves take tenth of seconds each, and that most of the time is the initial setup...

MikeSpreitzer: You can see the timings in what little logging _is_ produced. For example, the run just done on this PR, https://github.com/kubestellar/kubestellar/actions/runs/8836725617/job/24263965498?pr=2068 . In that case the first `It` took 2 1/2 seconds.

pdettori: I share @ezrasilvera's concern about always testing with the transformation, but I can see @MikeSpreitzer's point, so I think it's ok to move this forward.

pdettori (commented):
/lgtm

PR #2072, opened 2024-04-23
📖 Fix some wording in e2e/multi-cluster-deployment README for OCP

MikeSpreitzer:
## Summary
This PR makes a couple of small fixes to the wording in `test/e2e/multi-cluster-deployment/README.md`.

I am not entirely sure I got this right. Does this test really use three OCP clusters, or one somehow do double duty?

## Related issue(s)

Fixes #

### DIFF
--- a/test/e2e/multi-cluster-deployment/README.md
+++ b/test/e2e/multi-cluster-deployment/README.md
@@ -6,7 +6,7 @@ The kubestellar controller-manager will be invoked with `-v=2` unless otherwise 
 
 The transport controller will be invoked with `-v=4` unless othewise specified on the command line with `--transport-controller-verbosity $number`.
 
-## Running the test using a script
+## Running the test on three new Kind clusters
 
 Starting from a local directory containing the git repo, do the following.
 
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
@clubanderson 
@effi-ofer 
@KPRoche

test/e2e/multi-cluster-deployment/README.md
```diff
@@ -15,12 +15,10 @@ cd test/e2e/multi-cluster-deployment
 ./run-test.sh
 ```
 
-## Running the test in OCP cluster
+## Running the test in three OCP clusters
```
ezrasilvera, line 18: Not sure "three" is needed. We don't mention this explicitly for KIND or any other env. 
"Running the test using OCP clusters"

ezrasilvera:
Not sure why we need a different run instructions for OCP (other than the config) if the run-test.sh  includes an "if ocp"  block

test/e2e/multi-cluster-deployment/README.md
MikeSpreitzer: I pushed a revision that makes the structure of the two sub-headings parallel. I think it is important to be very clear about the infrastructure aspect.

MikeSpreitzer:
> Not sure why we need a different run instructions for OCP (other than the config) if the run-test.sh includes an "if ocp" block

I agree, but that is a different issue. I started to address it, but realized that I was unable to test on OCP. In any case, this is a matter for a different PR.

PR #2076, opened 2024-04-23
✨ Make the status controller handle the deletion of workstatuses

### LINKED ISSUE #2020: bug: No deletion of workload object status on deletion of workstatus
### Describe the bug

Status of workload objects in WDSes are not deleted, when corresponding WorkStatus objects are deleted due to BindingPolicy selection changes on singleton status report WEC(s).

I ran into this bug when I'm writing tests to my code changes for #1945.

### Steps To Reproduce

1. Run Kubestellar into a state where a WorkStatus object exists and the singleton status is reported to the corresponding workload object. One way to do that is leveraging the "singleton status testing" part of the ginkgo e2e tests.
2. Delete the WorkStatus object by changing the WEC selection. To do this, change the bindingpolicy's `clusterSelectors`, so that the `clusterSelectors` doesn't cover any WECs. Another way is to change the labels of the WECs so that they are not selected by the bindingpolicy any more.
3. Observe the deletion of the the WorkStatus object and the unexpected, continuous existence of the workload object status.

### Expected Behavior

Status of workload objects in WDSes should be deleted when corresponding WorkStatus objects are deleted.

### Additional Context

Currently the status controller just do nothing on deletion of WorkStatus object.
https://github.com/kubestellar/kubestellar/blob/189107031c2c2792aebad35128c0fd7b2abd2191/pkg/status/controller.go#L280-L283

More importantly, a _reference_ to WorkStatus object, instead of a full object, gets queued in the workqueue of the status controller. So when a WorkStatus object is deleted, the information to track the corresponding workload object (in .spec) is not visible to the workers.

Therefore, to fix this one, I think we need to change the current structure of the status controller quite a bit.

waltforme:
## Summary
This PR adds logic to the status controller so that it handles the deletion of WorkStatus objects.

This PR also completes a pending e2e test, tracked by [this comment](https://github.com/kubestellar/kubestellar/issues/2020#issuecomment-2037539110). The test was blocked by #2020, and it's now unblocked by this PR.

## Related issue(s)

Closes #2020

### DIFF
--- a/pkg/status/controller.go
+++ b/pkg/status/controller.go
@@ -170,7 +170,7 @@ func (c *Controller) runWorkStatusInformer(ctx context.Context) {
 			if shouldSkipDelete(obj) {
 				return
 			}
-			c.handleObject(obj)
+			c.handleObjectDeletion(obj)
 		},
 	})
 
@@ -204,6 +204,14 @@ func (c *Controller) handleObject(obj any) {
 	c.enqueueObject(obj)
 }
 
+// handleObjectDeletion is like handleObject except that handleObjectDeletion calls
+// enqueueObjectDeletion instead of enqueueObject
+func (c *Controller) handleObjectDeletion(obj any) {
+	rObj := obj.(runtime.Object)
+	c.logger.V(4).Info("Got object deletion event", "obj", util.RefToRuntimeObj(rObj))
+	c.enqueueObjectDeletion(obj)
+}
+
 // enqueueObject generates key and put it onto the work queue.
 func (c *Controller) enqueueObject(obj interface{}) {
 	ref, err := cache.ObjectToName(obj)
@@ -212,6 +220,12 @@ func (c *Controller) enqueueObject(obj interface{}) {
 		return
 	}
 	c.workqueue.Add(ref)
+}
+
+// enqueueObjectDeletion is like enqueueObject except that enqueueObjectDeletion
+// puts the full object instead of its reference into the work queue.
+func (c *Controller) enqueueObjectDeletion(obj interface{}) {
+	c.workqueue.Add(obj)
 }
 
 // runWorker is a long-running function that will continually call the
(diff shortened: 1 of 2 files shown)


pkg/status/controller.go
```diff
@@ -204,6 +204,14 @@ func (c *Controller) handleObject(obj any) {
 	c.enqueueObject(obj)
 }
 
+// handleObjectDeletion is like handleObject except that handleObjectDeletion calls
+// enqueueObjectDeletion instead of enqueueObject
+func (c *Controller) handleObjectDeletion(obj any) {
+	rObj := obj.(runtime.Object)
```
MikeSpreitzer, line 210: This could fail; the object could be a https://github.com/kubernetes/client-go/blob/v0.28.4/tools/cache/delta_fifo.go#L800

BTW: If you stop enqueuing whole objects then you will be able to use https://github.com/kubernetes/client-go/blob/v0.30.0/tools/cache/controller.go#L342 when using Kube release 1.30 or later.

pkg/status/controller.go
```diff
@@ -204,6 +204,14 @@ func (c *Controller) handleObject(obj any) {
...
+	c.logger.V(4).Info("Got object deletion event", "obj", util.RefToRuntimeObj(rObj))
+	c.enqueueObjectDeletion(obj)
```
MikeSpreitzer, line 212: The approach of enqueuing an object rather than an object reference can go wrong. The workqueue suppresses duplicates, as identified by Go's primitive `==`. The object is not `==` its https://github.com/kubernetes/client-go/blob/v0.28.4/tools/cache/object-names.go#L24 . The FIFO part of the workqueue is ordered, but retries spend some time in the delay part before going back in the FIFO. So the following sequence of events can get handled incorrectly.

1. User creates object.
2. Controller is notified of create and enqueues ObjectName.
3. Controller processes ObjectName, seeing new object.
4. User deletes object.
5. Controller is notified of delete and enqueues whole object.
6. User re-creates object.
7. Controller is notified of create and enqueues ObjectName.
8. Controller processes whole old object but gets an error, puts whole object back for retry later.
9. Controller processes ObjectName, seeing new version of object. Does right stuff?
10. Controller processes whole old object, undoes stuff just done.
11. Nothing else happens, user is sad.

MikeSpreitzer:
Using the status controller's memory of the association between workload object and WorkStatus object can not work because the WorkStatus object could be deleted while the status controller is not running.

If we _could_ rely on the controller's memory, then it would not be necessary for the controller to enqueue whole WorkStatus objects. Instead the controller could maintain a map from WorkStatus name to the `util.SourceRef` from that WorkStatus.

One approach would be to rely on the status controller's memory, improving that by having the controller maintain a finalizer on the WorkStatus object. This is technically imperfect because Kubernetes does not have a perfect failure detector; it is barely possible that two concurrent copies of the controller could be running, each thinking that it is the sole manager of the finalizer and related state.

Another approach would be to maintain the association information in the workload object. Define a label name such as `internal.kubestellar.io/work-status-ref` to hold a reference to the WorkStatus that is the source of the workload object's status information.

MikeSpreitzer:
The approach of relying on _the_ WorkStatus object to indicate which workload object it applies to is fragile: because Kubernetes has no multi-object atomic transactions, and because Kubernetes is based on eventual consistency with no coherence between notifications about different objects, it is possible that the controller may see multiple WorkStatus objects referring to the same workload object for a while.

waltforme:
I thought about the proposed solutions, and also reflected on the current designs of the status controller.

When the status controller handles singleton statuses, it solely relies on the changes of WorkStatus object. In other words, the controller doesn't have a desired state clearly defined. This seems to be 'edge-driven' instead of (recommended) 'level-driven'. More details are available [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/controllers.md#guidelines). As a consequence, we have Issue #2020. We may probably have more issues in the future.

To improve, I'm proposing some redesigns in [this doc](https://docs.google.com/document/d/1KD0S79cxqDp2mD55FQ4BGeWtxEN2TM6g9uiiSa-LO80/edit?usp=sharing). I started by (a) introducing a desired state, and (b) switching to the established 'edge-triggered, level-driven' pattern.

PR #2079, opened 2024-04-24
📖 Mostly update the architecture doc of status add-on agent

### LINKED ISSUE #2078: bug: description of status add-on agent is confusingly split/duplicated in arch doc
### Describe the bug

The OCM Status Add-On Agent is described in two places in the architecture document.
1. https://github.com/kubestellar/kubestellar/blob/99d6a53d97bcec6393c6a673297a1c39c9908dac/docs/content/direct/architecture.md#ocm-status-add-on-agent
2. In the last paragraph of the section on the "status controller" (https://github.com/kubestellar/kubestellar/blob/99d6a53d97bcec6393c6a673297a1c39c9908dac/docs/content/direct/architecture.md#ocm-status-add-on-agent).

The second one is confusing, since it is buried in the description of a different controller. This makes it confusing for the reader, who thinks that they are reading a description of the status controller when they are actually looking at a description of the OCM Status Add-On Agent.

The second one is also largely duplicative of the first.

### Steps To Reproduce

See above.

### Expected Behavior

No great confusion or duplication.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR updates the architecture document, mainly regarding the OCM Status Add-On Agent. This PR also makes some references to OCM object types into hyperlinks to their souce (which is helpful for developers). This PR also changes several references to Go datatypes to use code-quoting instead of emphasis.

## Related issue(s)

Fixes #2078

### DIFF
--- a/docs/content/direct/architecture.md
+++ b/docs/content/direct/architecture.md
@@ -50,11 +50,11 @@ that are put in a mailbox namespace are delivered to the matching WEC.
 on all WECs and sets the RBAC permissions for it using the OCM add-on framework.
 
 - *OCM Agent*: This module registers the WEC to the OCM Hub, watches for 
-ManifestWorks and unwraps and syncs the objects into the WEC.
+[ManifestWork.v1.work.open-cluster-management.io](https://github.com/open-cluster-management-io/api/blob/v0.12.0/work/v1/types.go#L17) objects and unwraps and syncs the objects into the WEC.
 
-- *OCM Status Add-On Agent*: This module watches *AppliedManifestWorks* 
+- *OCM Status Add-On Agent*: This module watches [AppliedManifestWork.v1.work.open-cluster-management.io](https://github.com/open-cluster-management-io/api/blob/v0.12.0/work/v1/types.go#L528) objects 
 to find objects that are synced by the OCM agent, gets their status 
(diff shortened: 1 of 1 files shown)


docs/content/direct/architecture.md
```diff
@@ -50,11 +50,11 @@ that are put in a mailbox namespace are delivered to the matching WEC.
 on all WECs and sets the RBAC permissions for it using the OCM add-on framework.
 
 - *OCM Agent*: This module registers the WEC to the OCM Hub, watches for 
-ManifestWorks and unwraps and syncs the objects into the WEC.
+[ManifestWork.v1.work.open-cluster-management.io](https://github.com/open-cluster-management-io/api/blob/v0.11.0/work/v1/types.go#L17) objects and unwraps and syncs the objects into the WEC.
```
waltforme, line 53: I see that we are using v0.12.0 for the OCM APIs. Do we want to make the version consistent if we put a link to the OCM source, or it doesn't matter? (There seems to be no diff though between 0.11.0 and 0.12.0).

MikeSpreitzer: Thanks for caching that. I have pushed a rev that corrects the versions.

PR #2086, opened 2024-04-26
✨ Update PCHs to include ocm status-addon and transport-plugin

francostellari:
## Summary

Update PCHs to include ocm status-addon and transport-plugin:

1. update ocm.yaml PCH to include a second container installing the OCM status-addon along OCM
2. update kubestellar.yaml PCH to include a second container installing the OCM transport-plugin along with the KubeStellar plugin
3. update the example.md to remove the corresponding 2 steps

## Related issue(s)

Fixes #

### DIFF
--- a/config/postcreate-hooks/kubestellar.yaml
+++ b/config/postcreate-hooks/kubestellar.yaml
@@ -74,8 +74,8 @@ spec:
       verbs:
       - get
       - list
-      - watch      
-  - apiVersion: rbac.authorization.k8s.io/v1  
+      - watch
+  - apiVersion: rbac.authorization.k8s.io/v1
     kind: ClusterRoleBinding
     metadata:
       name: "{{.Namespace}}-{{.HookName}}"
@@ -86,7 +86,7 @@ spec:
     subjects:
     - kind: ServiceAccount
       name: default
-      namespace: "{{.Namespace}}"       
+      namespace: "{{.Namespace}}"
   - apiVersion: rbac.authorization.k8s.io/v1
     kind: Role
     metadata:
@@ -138,7 +138,7 @@ spec:
       resources:
       - services
       verbs:
-      - get      
+      - get
     - apiGroups:
       - apps
       resources:
@@ -152,8 +152,8 @@ spec:
       - roles
       - rolebindings
       verbs:
-      - create                
-  - apiVersion: rbac.authorization.k8s.io/v1  
+      - create
+  - apiVersion: rbac.authorization.k8s.io/v1
     kind: RoleBinding
     metadata:
       name: "{{.HookName}}"
@@ -164,7 +164,7 @@ spec:
     subjects:
     - kind: ServiceAccount
       name: default
-      namespace: "{{.Namespace}}"   
+      namespace: "{{.Namespace}}"
   - apiVersion: batch/v1
     kind: Job
     metadata:
@@ -179,7 +179,7 @@ spec:
             args:
               - upgrade
               - --install
-              - -n
+              - --namespace
               - "{{.Namespace}}"
               - kubestellar
               - oci://ghcr.io/kubestellar/kubestellar/controller-manager-chart
@@ -189,6 +189,23 @@ spec:
               - "ControlPlaneName={{.ControlPlaneName}}"
             env:
             - name: XDG_CACHE_HOME
-              value: /tmp/helm/.cache  
+              value: /tmp/helm/.cache
+          - name: "{{.HookName}}-otp"
+            image: quay.io/kubestellar/helm:3.14.0
+            imagePullPolicy: IfNotPresent
+            args:
+              - upgrade
+              - --install
+              - --namespace
+              - "{{.Namespace}}"
+              - ocm-transport-plugin
+              - oci://ghcr.io/kubestellar/ocm-transport-plugin/chart/ocm-transport-plugin
+              - --version
+              - "0.1.7"
+              - --set
+              - "wds_cp_name={{.ControlPlaneName}}"
+            env:
+            - name: XDG_CACHE_HOME
+              value: /tmp/helm/.cache
           restartPolicy: Never
       backoffLimit: 1
--- a/config/postcreate-hooks/ocm.yaml
+++ b/config/postcreate-hooks/ocm.yaml
@@ -14,17 +14,40 @@ spec:
       template:
         spec:
           containers:
-          - name: "{{.HookName}}"
+          - name: "{{.HookName}}-ocm"
             image: quay.io/kubestellar/clusteradm:0.7.2
             args:
             - init
             env:
             - name: KUBECONFIG
-              value: "/etc/kube/config-incluster"    
+              value: "/etc/kube/config-incluster"
             volumeMounts:
             - name: kubeconfig
               mountPath: "/etc/kube"
-              readOnly: true  
+              readOnly: true
+          - name: "{{.HookName}}-statusaddon"
+            image: quay.io/kubestellar/helm:3.14.0
+            args:
+              - upgrade
+              - --install
+              - status-addon
+              - oci://ghcr.io/kubestellar/ocm-status-addon-chart
+              - --version
+              - v0.2.0-rc8
+              - --namespace
+              - open-cluster-management
+              - --create-namespace
+            env:
+            - name: HELM_CONFIG_HOME
+              value: "/tmp"
+            - name: HELM_CACHE_HOME
+              value: "/tmp"
+            - name: KUBECONFIG
+              value: "/etc/kube/config-incluster"
+            volumeMounts:
+            - name: kubeconfig
+              mountPath: "/etc/kube"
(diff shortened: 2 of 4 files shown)


francostellari:
/unhold
@MikeSpreitzer This PR is now working in Kind, passing `run.sh --released`, and rebased
Please take another look

francostellari:
@clubanderson please take a look, this should remove 2 steps in the common setup and simplify your installation

francostellari:
@MikeSpreitzer I rebased on ks 0.22.0 and retested with `--released`. This PR is ready to review again. Thank you.

config/postcreate-hooks/kubestellar.yaml
```diff
@@ -189,6 +189,23 @@ spec:
...
+            imagePullPolicy: IfNotPresent
+            args:
+              - upgrade
+              - --install
+              - --namespace
+              - "{{.Namespace}}"
+              - ocm-transport-plugin
+              - oci://ghcr.io/kubestellar/ocm-transport-plugin/chart/ocm-transport-plugin
+              - --version
+              - "0.1.7"
```
pdettori, line 204: Could we use the PCH support for user-defined objects (or vars) to avoid having these versions numbers hardcoded in the PCH ? See https://github.com/kubestellar/kubeflex/blob/main/docs/users.md#user-provided-objects

config/postcreate-hooks/ocm.yaml
```diff
@@ -14,17 +14,40 @@ spec:
...
+              readOnly: true
+          - name: "{{.HookName}}-statusaddon"
+            image: quay.io/kubestellar/helm:3.14.0
+            args:
+              - upgrade
+              - --install
+              - status-addon
+              - oci://ghcr.io/kubestellar/ocm-status-addon-chart
+              - --version
+              - v0.2.0-rc8
```
pdettori, line 36: Same comment as above for user-defined objects

config/postcreate-hooks/kubestellar.yaml
francostellari: @pdettori I could do that way, and I'm fine changing it. However, somewhat, I think we want to hide the specific dependencies to the general user similar to what you do for postgress.
The current way, one specify the version of KS and the PCH already use the correct dependencies.
In your proposed way, one would have to specify the version of KS PCH, then specify the version of status addon and otp on the kflex command line... more flexible, I guess but more prone to errors.

@MikeSpreitzer @ezrasilvera do you also have an opinion?
I'm fine either way

BTW, this will be superseded by the hub-chart where my understanding of the feedback was to hard code the dependencies in the yaml and not expose them in the `values.yaml` .

pdettori: I am aware of the desire of hiding deps and I agree. I was more concerned with how in many places we need to update dependencies when we make a new release. It would be an improvement to find ways to define all versions only in one single place rather than scattered all over our code.

MikeSpreitzer: The anticipated hub chart will be a convenient all-in-one bundle but not the only way to create a WDS. I do not think that we should require that all the WDSes are created up-front. I would not want to make the user creating a WDS some time after the initial setup have to specify the ks/OTP release to use.

config/postcreate-hooks/kubestellar.yaml
```diff
@@ -189,6 +189,23 @@ spec:
...
+              - --set
+              - "wds_cp_name={{.ControlPlaneName}}"
```
MikeSpreitzer, line 206: Shouldn't this also `--set transport_cp_name=` something?

docs/content/direct/examples.md
```diff
@@ -42,45 +40,21 @@ The following steps establish an initial state used in the examples below.
 
 1. Create an inventory & mailbox space of type `vcluster` running *OCM* (Open Cluster Management)
 in KubeFlex. Note that `-p ocm` runs a post-create hook on the *vcluster* control plane
-which installs OCM on it.
+which installs OCM on it. This step includes the status add-on described [here](./architecture.md#ocm-status-add-on-agent) for more details on the add-on.
```
MikeSpreitzer, line 43: The architecture document describes two status add-on thingies: the "Status Add-On Controller" and the "OCM Status Add-On Agent". I recommend that the text here refer specifically to the controller.

MikeSpreitzer:
Successfully tested `test/e2e/multi-cluster-deployment/run-test.sh` (**no** `--released`).

config/postcreate-hooks/kubestellar.yaml
francostellari: @MikeSpreitzer Both ks and otp plugin can use an ITS parameter. In the current ks PCH deploying only the kc plugin, it has been decided not to pass explicitly an its parameter and let the plugin choose the its cp. I have adopted the same behavior when adding the otp plugin. You may remember that  a [PR](https://github.com/kubestellar/ocm-transport-plugin/pull/12) for OTP was used to allow for the automatic selection of the ITS if none is passed.

docs/content/direct/examples.md
```diff
@@ -22,8 +22,6 @@ The following steps establish an initial state used in the examples below.
 
     ```shell
     export KUBESTELLAR_VERSION=0.22.0
-    export OCM_STATUS_ADDON_VERSION=0.2.0-rc8
-    export OCM_TRANSPORT_PLUGIN=0.1.7
```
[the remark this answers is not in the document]
francostellari: @MikeSpreitzer can you clarify your comment?

docs/content/direct/examples.md
```diff
@@ -42,45 +40,21 @@ The following steps establish an initial state used in the examples below.
...
     kflex create its1 --type vcluster -p ocm
     ```
 
-1. Install status add-on on its1:
-
-    Wait until the `managedclusteraddons` resource shows up on `its1`. You can check on that with the command:
-
-    ```shell
-    kubectl --context its1 api-resources | grep managedclusteraddons || true
-    ```
```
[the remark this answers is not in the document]
francostellari: Yes

docs/content/direct/examples.md
francostellari: ok, reworded as you suggested

docs/content/direct/examples.md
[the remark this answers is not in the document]
MikeSpreitzer: I mean that I like this change.

config/postcreate-hooks/kubestellar.yaml
MikeSpreitzer: .. and when a user later adds a WDS, we should not make them specify the OTP version. That is why I approve of specifying the OTP version here.

config/postcreate-hooks/kubestellar.yaml
MikeSpreitzer: > Both ks and otp plugin can use an ITS parameter

I suppose that by "can" you mean "could be changed to".

I am aware of the other facts that you recited. The problem is that we are trying to support a general number of ITSes and general choice of ITS for each WDS. Forcing a guess for every WDS does not do that. This is why I am advocating adding an ITS name parameter to the kubestellar PCH.

francostellari: @MikeSpreitzer there is no technical issue preventing it. the kflex command line will get longer and more verbose, even in the simple cases since there are no defaults, e.g.:

```shell
kflex create wds1 -p kubestellar --set ITSName=its1
```
There is also no easy way to document the set values since there "hidden" in the PCH yaml... perhaps we can put some comments at the header of the PCH.

I think in the short term we will move to use ks hub chart for the common setup and wih that it will easy, in my opinion, to achieve this and document it. Perhaps we can leave the current approach for a little longer and wait for the hub.
You can take a look to some initial drafting already enabling it [here](https://github.com/francostellari/kubestellar/tree/hub-chart/hub-chart)... please ignore the patches and test code to work around to pending new chart releases, this is not the final product

MikeSpreitzer: I think that we can pursue the ITS-specification issue in follow-on PRs.

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

config/postcreate-hooks/kubestellar.yaml
MikeSpreitzer: @pdettori, I think that @francostellari has identified some important issues for PostCreateHooks:

1. Default values
2. Easy-to-find declaration of all the user-defined parameters

pdettori: I see your point @MikeSpreitzer. I am aware of the issue for default values  https://github.com/kubestellar/kubeflex/issues/233, I think it's a useful feature to have, we could also have some place for declarations. For the last point on helm, yes, that is a good point and I suppose it's true in general, so it seems like helm as a front-end for handling values and defaults makes a lot of sense.

PR #2088, opened 2024-04-26
🐛 Document and update dependency of ks/ks on ks/OSA

### LINKED ISSUE #2087: bug: incorrect new self-reference
### Describe the bug

https://github.com/kubestellar/kubestellar/blob/v0.22.0-rc1/docs/content/direct/deploy-on-k3d.md?plain=1#L24 refers to the wrong version of ks/ks

### Steps To Reproduce

Look at it.  Look at the URL.

### Expected Behavior

Consistency.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR updates the references to ks/OSA to use the latest version, and updates the release procedure document to mention this.

This PR also updates the `deploy-on-k3d.md` document to something that works for me, where "works" means that it runs successfully and can be followed by use of examples.md. This involved retreating from some changes that anticipated some sensible renaming of things in other files (such as examples.md). I think that the ambition is laudible but belongs in a separate PR.

## Related issue(s)

Fixes #2087

### DIFF
--- a/docs/content/direct/release.md
+++ b/docs/content/direct/release.md
@@ -21,6 +21,15 @@ Between each release of [ks/OTP](https://github.com/kubestellar/ocm-transport-pl
 
 - Edit `test/e2e/common/setup-kubestellar.sh`: update the setting of `OCM_TRANSPORT_PLUGIN_RELEASE` to the latest.
 
+### Reacting to a new ocm-status-addon release
+
+Update the references to the ocm-status-addon release in the following files.
+
+- `docs/content/direct/examples.md`
+- `docs/content/direct/deploy-on-k3d.md`
+- `test/e2e/multi-cluster-deployment/README.md`
+- `test/e2e/common/setup-kubestellar.sh`
+
 ### Making a new kubestellar release
 
 Making a new kubestellar release requires a contributor to do the following things. Here `$version` is the semver identifier for the release (e.g., `1.2.3-rc2`).
@@ -28,6 +37,8 @@ Making a new kubestellar release requires a contributor to do the following thin
 - Edit the source for the KCM PCH (in `config/postcreate-hooks/kubestellar.yaml`) and update the tag in the reference to the KCM container image (it appears in the last object, a `Job`).
 
 - Edit [the examples document](examples.md) to update the self-references for the coming release.
+
+- Edit [the k3d deploy instructions](deploy-on-k3d.md) to update the self-reference.
 
 - Until we have our first stable release, edit [the README](README.md#latest-stable-release) where it wishes it could cite a stable release but instead cites the latest release, to rever to the coming release.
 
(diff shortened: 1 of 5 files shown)


MikeSpreitzer:
/cc @pdettori 
/cc @ezrasilvera 
/cc @eaepstein 
/cc @nirrozenbaum 
@KPRoche

ezrasilvera:
I don't think we ever decided that k3d is a platform we should test on each release. As it is not tested, we shouldn't update the release docs to include updating that example.

docs/content/direct/release.md
```diff
@@ -29,6 +38,8 @@ Making a new kubestellar release requires a contributor to do the following thin
 
 - Edit [the examples document](examples.md) to update the self-references for the coming release.
 
+- Edit [the k3d deploy instructions](deploy-on-k3d.md) to update the self-reference.
+
```
ezrasilvera, line 42: k3d is just an example and not part of every release update

docs/content/direct/release.md
```diff
@@ -21,6 +21,15 @@ Between each release of [ks/OTP](https://github.com/kubestellar/ocm-transport-pl
 
 - Edit `test/e2e/common/setup-kubestellar.sh`: update the setting of `OCM_TRANSPORT_PLUGIN_RELEASE` to the latest.
 
+### Reacting to a new ocm-status-addon release
+
+Update the references to the ocm-status-addon release in the following files.
+
+- `docs/content/direct/examples.md`
+- `docs/content/direct/deploy-on-k3d.md`
```
ezrasilvera, line 29: same comment for k3d

docs/content/direct/release.md
MikeSpreitzer: We need to maintain this document or mark it as a historical point-in-time.

My understanding is that we are working on supporting users who use k3d. @eaepstein ?

ezrasilvera: We haven't decided that. For now we support it through point in time example. We can't support continuously all our examples, we would need to add them into the release test, and validat8each release. This is time consuming

eaepstein: If and when the e2e test using k3d is integrated it can be part of the automated test and the write-up significantly simplified. Until then we can add something to the document like "tested on release x.y" and validating with the latest release can be optional.

ezrasilvera: @eaepstein Exactly.  So it should not be part of our release documents (i.e., the need to update the links in the example for every release).  As all the other non-e2e examples we need to include a specific release version on which it was tested.

MikeSpreitzer: As far as I can tell, currently `deploy-on-k3d.md` does not work as written and produces a state from which the user can not proceed to `examples.md`. This is mainly because of anticipation of laudible changes in naming conventions throughout this project, to remove its overwhelming specialization to infrastructure that comes from `kind`.

I have extended this PR with commits that retreat from than ambition, modifying `deploy-on-k3d.md` to something that _does_ work and produces a state from which a user _can_ proceed to `examples.md`. Yes, with the unsightly fact that the kubeflex hosting context is named "kind-kubeflex" --- because that is baked into `examples.md`. I think that changing the rest of the project contents to remove the dedication to `kind` is a good thing to do but belongs in a separate PR.

MikeSpreitzer:
@francostellari: this PR is also updating dependencies

Issue #2019, opened 2024-04-04
bug: weird stuff with latest and stable release pointers in the website


MikeSpreitzer:
### Describe the bug

- https://docs.kubestellar.io/release-0.21.0/direct/examples/ tells the reader to "You're not viewing the stable release. Click here to go to the stable release" (link to https://docs.kubestellar.io/release-0.21.0/). This refers to itself.
- Similar for every release.
- https://docs.kubestellar.io/release-0.21.1/ tells the reader "You're not viewing the stable release. Click here to go to the stable release" (link to https://docs.kubestellar.io/latest/). Note wording mismatch: "stable" in the words, "latest" in the link.
- We seem to be treating "main" as the "latest" release, which is wrong because it is not something that users can use (it is not actually a release).

### Steps To Reproduce

See above.

### Expected Behavior

Consistent behavior

MikeSpreitzer:
What triggers the warning shown here to appear?

<img width="606" alt="Screenshot 2024-04-04 at 1 38 35 PM" src="https://github.com/kubestellar/kubestellar/assets/14296719/b1b8c021-f937-42f8-8e36-fe00ce4dea55">

MikeSpreitzer:
And why does this one page look different from all the others (of a given release)?

<img width="591" alt="Screenshot 2024-04-04 at 1 43 18 PM" src="https://github.com/kubestellar/kubestellar/assets/14296719/1820e6c4-5e9c-4eb1-9ed2-f084694673f6">

MikeSpreitzer:
Regarding the "version warning", there is some documentation at https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/?h=#version-warning and https://github.com/jimporter/mike .

MikeSpreitzer:
Regarding every version pointing to itself as "latest" and "stable" in the warning banner, I figured out the pathology.

Take https://docs.kubestellar.io/release-0.22.0-rc1/readme/ as an example. The HTML at that URL has the following for the warning banner:

```
You're not viewing the latest release.
  <a href="[../latest/..](view-source:https://docs.kubestellar.io/release-0.22.0-rc1/)">
    <strong>Click here to go to the latest release</strong>
  </a>
  Also, FYI, 
  <a href="[../stable/..](view-source:https://docs.kubestellar.io/release-0.22.0-rc1/)">
    <strong>Click here to go to the latest stable release</strong>
  </a>
```

[...]

PR #2107, opened 2024-05-03
📖 Fix version warning for release-0.22.0

MikeSpreitzer:
## Summary
This PR fixes the functionality of the "version warning" banner (see https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/?h=#setting-a-default-version), so that the links correctly lead to the "stable" and "latest" aliases.

## Related issue(s)

This addresses the first problem in Issue #2019 --- but only for release-0.22.0 .

### DIFF
--- a/docs/overrides/main.html
+++ b/docs/overrides/main.html
@@ -60,13 +60,13 @@
 {% endblock %}
 
 {% block outdated %}
-  You're not viewing the latest release.
-  <a href="{{ '../latest/' ~ base_url }}">
-    <strong>Click here to go to the latest release</strong>
+  You're not viewing the latest stable release.
+  <a href="{{ '../' ~ base_url ~ '/stable' }}">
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
BTW, I first updated the "latest" alias manually. Here is the key command that I ran:

```shell
mike alias -r upstream --update-aliases release-0.22.0 latest
```

I manually pushed that upstream after that.

MikeSpreitzer:
I tested this locally with `mike deploy` sans `--push` and then `mike serve`.

ezrasilvera (changes requested):
As all our regular releases are considered as "stable" - please change the need for an explicit definition of "stable release version"

MikeSpreitzer:
@ezrasilvera: I have narrowed this PR so that it _only_ fixes the version warning banner behavior. We are pursuing the alias policy issues in other PRs.

PR #2108, opened 2024-05-03
📖 Fix version warning banner in release-0.22.0-rc3

MikeSpreitzer:
## Summary
This PR fixes the version navigation in the version warning banner, in branch release-0.22.0-rc3.

## Related issue(s)

This addresses the first problem in Issue #2019  --- for the "release-0.22.0-rc3" version of the source of the website.

PR #2109, opened 2024-05-03
✨ Parameterize deploy docs

MikeSpreitzer:
## Summary
This PR changes `.github/workflows/docs-gen-and-push.yml` from fixed aliasing to parameterized aliasing. Before this change, there was a fixed policy:

- version "main" gets aliased to "latest" and "unstable";
- version "release-0.21.1" (as defined by a line in `docs/scripts/deploy-docs.sh`) gets aliased to "stable".

With this change: there is still a fixed policy of aliasing "main" to "unstable", but the aliases "latest" and "stable" are not automatically applied. Instead the workflow takes inputs that tell it whether to apply the "latest" alias and whether to apply the "stable" alias. When project maintainers decide to advance the "latest" and/or "stable" aliases, they explicitly run the workflow with appropriate parameter values. Here is an example of what that looked like when I was testing:

Screenshot 2024-05-03 at 1 18 15 AM

Note also that `mike` gives control over where a visitor to the site initially lands. In `mike` jargon this is called the "default" version. Our website currently has that set to "stable". We could change it to "latest" or "unstable" if we wanted to. See https://github.com/jimporter/mike?tab=readme-ov-file#setting-the-default-version .

`mike` has a `list` command that lists all the website versions and their aliases. Here is the current list for our website:

```
release-0.22.0 [latest]
release-0.22.0-rc3
release-0.22.0-rc2
release-0.22.0-rc1
release-0.21.2 [stable]
release-0.21.2-rc1
release-0.21.1
release-0.21.0
release-0.15
release-0.14
release-0.13
release-0.12
release-0.11
release-0.10
release-0.9
release-0.8
release-0.7
release-0.6
release-0.5
release-0.4
release-0.3
release-0.2
main [unstable]
```

This PR also removes an unnecessary step from the workflow: setup for Go. The doc tools are written in Python.

## Related issue(s)

This addresses part of #2019 .

### DIFF
--- a/docs/scripts/deploy-docs.sh
+++ b/docs/scripts/deploy-docs.sh
@@ -22,8 +22,6 @@ set -o xtrace
 REPO_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)
 cd "$REPO_ROOT/docs"
 
-STABLE_RELEASE=release-0.21.1
-
 if [[ "${GITHUB_EVENT_NAME:-}" == "pull_request" || "${GITHUB_EVENT_NAME:-}" == "pull_request_target" ]]; then
   # For PRs, we don't want to use GITHUB_REF_NAME, which will be something like merge/1234; instead, we want to use
   # the branch the PR is targeting, such as main or release-0.11
@@ -44,7 +42,22 @@ fi
 
 MIKE_OPTIONS=()
 ALIAS_OPTIONS=()
-ALIAS_OPTIONS_LATEST=()
+ALIAS_LIST=()
+
+while (( $# > 0 )); do
+    case "$1" in
+	(--alias-to=latest)
+	    ALIAS_OPTIONS=(--update-aliases)
+	    ALIAS_LIST+=(latest);;
+	(--alias-to=stable)
+	    ALIAS_OPTIONS=(--update-aliases)
+	    ALIAS_LIST+=(stable);;
+	(*)
+	    echo "$0 usage: [--alias-to={latest,stable}]*" >&2
+	    exit 1;;
+    esac
+    shift
+done
 
 if [[ -n "${REMOTE:-}" ]]; then
   MIKE_OPTIONS+=(--remote "$REMOTE")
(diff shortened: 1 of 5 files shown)


MikeSpreitzer:
I successfully tested this by pushing to my fork. I pushed branches named "release-test" and "main", containing the same contents as this PR plus one more commit that stops the suppression of the workflow on forks and the attempt to run with elevated credentials.

docs/scripts/deploy-docs.sh
```diff
@@ -88,13 +83,9 @@ if [[ -n "${CI:-}" ]]; then
...
   mike deploy "${MIKE_OPTIONS[@]}" $VERSION
   exit
 fi
 
-mike deploy "${MIKE_OPTIONS[@]}" "${ALIAS_OPTIONS[@]}"
-
-if [ ${#ALIAS_OPTIONS_LATEST[@]} -gt 0 ]; then
-    mike deploy "${MIKE_OPTIONS[@]}" "${ALIAS_OPTIONS_LATEST[@]}"
-fi
+mike deploy "${MIKE_OPTIONS[@]}" "${ALIAS_OPTIONS[@]}" "$VERSION" "${ALIAS_LIST[@]}"
```
MikeSpreitzer, line 91: FYI, I simplified from multiple runs of `mike deploy` to just one. A single invocation of `mike deploy` can be given multiple aliases. Note the dot-dot-dot after "[alias]" in https://pypi.org/project/mike/0.3.4/#building-your-docs . For an example of a successful test, see the deploy-docs invocation at https://github.com/MikeSpreitzer/kcp-edge-mc/actions/runs/8934350045/job/24541156589#step:5:176 and the corresponding `mike deploy` invocation at https://github.com/MikeSpreitzer/kcp-edge-mc/actions/runs/8934350045/job/24541156589#step:5:213 .

ezrasilvera:
@MikeSpreitzer (cc: @pdettori ) I think that we all agree we consider all our regular releases as "stable"  (not the RCs ).  I don't think the user/admin need to explicitly decide and put a "stable" label.  
I agree that the "main" can be considered unstable.

ezrasilvera (changes requested):
Please remove the mandatory need for user input and make the default as "latest" and "stable"  . 
"main" should always be defined as "unstable"

MikeSpreitzer:
Automatically identifying the _latest_ stable is part of the question. As you see in #2108 , for example, the regular release branch that was most recently merged into is not necessarily the latest release (we can make changes in older release branches).

Do we simply want "latest" =def= "stable" and RCs are never that, or do we want a release candidate to be "latest" when it is higher than the last regular release?

ezrasilvera:
> Automatically identifying the _latest_ stable is part of the question. As you see in #2108 , for example, the regular release branch that was most recently merged into is not necessarily the latest release (we can make changes in older release branches).
> 
> Do we simply want "latest" =def= "stable" and RCs are never that, or do we want a release candidate to be "latest" when it is higher than the last regular release?

I think that this is indeed what we want.  latest==stable +  any regular release ==stable (it's implied by the first statement as each regular release was "latest" when it was created)

ezrasilvera:
> Automatically identifying the _latest_ stable is part of the question. As you see in #2108 , for example, the regular release branch that was most recently merged into is not necessarily the latest release (we can make changes in older release branches).
> 

For me latest== highest version number   ...    I don't know how difficult it is to create such a rule, as you are correct it's not equal to the release that go the most recent change/merge

MikeSpreitzer:
@ezrasilvera : just to be clear, you are proposing that an RC release is _never_ "latest", right?

docs/scripts/deploy-docs.sh
ezrasilvera: Thanks @MikeSpreitzer this is indeed how I use `mike`  from the command line.  
I think we also have some `Voodoo`  steps in the mike process that are not needed but we can push in a follow up PRs

ezrasilvera:
> @ezrasilvera : just to be clear, you are proposing that an RC release is _never_ "latest", right?

They are certainly "no stable" .  So I personally think they shouldn't be "latest" as we  defined "latest stable"  , Also at least for now we assume RCs are trainset releases that live for a very short time and are only a route for the "regular release"

MikeSpreitzer:
I could not easily find documentation to this effect, but experiment shows that `mike list` seems to be ordering things consistently with semver. I added `release-0.21.3`, and after that `mike list` produced the following list.

```
release-0.22.0 [latest]
release-0.22.0-rc3
release-0.21.3
release-0.21.2 [stable]
release-0.21.2-rc1
release-0.21.1
release-0.21.0
release-0.14
main [unstable]
```

MikeSpreitzer:
We are using the `mike` concept of "alias". An alias points to only one version.

MikeSpreitzer:
@ezrasilvera: I think you are asserting the following concepts and policy.

We have "regular" releases and "candidate" releases (RCs). We have no concept of "stable" release. The latest regular release is what we advocate and advertise.

The website has "versions". There is a version for every release and a version for "main".

The website displays the version warning banner whenever the reader is looking at a version other than "main" or the latest regular release. The version warning banner will say "you are not viewing the latest release. Click here to go to the latest release."

Someone testing a candidate release can find it on the website but is not what the website defaults to showing. Such a tester has to explicitly navigate to the candidate release that they want, and will get the version warning banner.

ezrasilvera:
@MikeSpreitzer I think we are +- on the same page.  Let's merge this and we can fine tune the behavior  in future PRs.

PR #2116, opened 2024-05-05
📖 Describe the accepted JSONPath subset more carefully

### LINKED ISSUE #2113: bug: user guide should be more clear about limitations of CustomTransform JSONPath syntax
### Describe the bug

The user guide should have enough information to explain how to deal with an array of values; for example it is not clear from the docs if rules below:

``` 
kind: CustomTransform
metadata:
  name: workflow
spec:
  apiGroup: argoproj.io
  resource: workflows
  remove:
  - $.spec.suspend
  - $.metadata.labels["workflows.argoproj.io/completed"]
  - $.metadata.labels["workflows.argoproj.io/phase"]
  - $.metadata.labels["pipeline/runid"]
```  

can also be written as:

```
kind: CustomTransform
metadata:
  name: workflow
spec:
  apiGroup: argoproj.io
  resource: workflows
  remove:
  - $.spec.suspend
  - $.metadata.labels["workflows.argoproj.io/completed","workflows.argoproj.io/phase","pipeline/runid" ]
```

(the answer I have got from the maintainers is NO, but I could not infer that from the documentation)  

### Steps To Reproduce

Look at the docs

### Expected Behavior

Should be more clear about the array syntax expression.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR makes the description of the subset of JSONPath accepted in a CustomTransform object more precise.

## Related issue(s)

Fixes #2113

### DIFF
--- a/docs/content/direct/user-guide.md
+++ b/docs/content/direct/user-guide.md
@@ -58,7 +58,7 @@ The user can configure additional transformations of workload objects by putting
 
 Currently the binding is simply by naming the workload object's API group and "resource" name in the `CustomTransform`'s `spec`. The transformations from all of the bound `CustomTransform` objects are applied to the workload object. There should be at most one `CustomTransform` object that specifies a given API group and resource.
 
-Currently the only available transformations are removals of specified content. The content to be removed is identified by a small subset of JSONPath. In this subset, the only available operators are `.someName` and `["some string"]`. The allowed names and strings are as specified in [RFC 9535](https://www.rfc-editor.org/rfc/rfc9535.html), except that only double-quoted strings are allowed.
+Currently the only available transformations are removals of specified content. The content to be removed is identified by a small subset of JSONPath (which was originally and somewhat loosely defined in [an article by Stefan Goessner](https://goessner.net/articles/JsonPath/) and later defined more carefully in [RFC 9535](https://datatracker.ietf.org/doc/rfc9535/)). In the subset accepted here: the root node identifier (`$`) must be followed by a positive number of segments, where each segment is either (a) `.` and a name (a `member-name-shorthand`, in the grammar of the RFC) or (b) `[`, a string literal, and `]`; no more of the grammar is allowed, not even whitespace. The allowed names and string literalss are as specified in RFC 9535, except that only double-quoted strings are allowed.
 
 For example, the following `CustomTransform` object says to remove the `spec` field named `suspend` from `Job` objects (in the API group `batch`).
 


docs/content/direct/user-guide.md
```diff
@@ -58,7 +58,7 @@ The user can configure additional transformations of workload objects by putting
...
 
-Currently the only available transformations are removals of specified content. The content to be removed is identified by a small subset of JSONPath. In this subset, the only available operators are `.someName` and `["some string"]`. The allowed names and strings are as specified in [RFC 9535](https://www.rfc-editor.org/rfc/rfc9535.html), except that only double-quoted strings are allowed.
+Currently the only available transformations are removals of specified content. The content to be removed is identified by a small subset of JSONPath (which was originally and somewhat loosely defined in [an article by Stefan Goessner](https://goessner.net/articles/JsonPath/) and later defined more carefully in [RFC 9535](https://datatracker.ietf.org/doc/rfc9535/)). In the subset accepted here: the root node identifier (`$`) must be followed by a positive number of segments, where each segment is either (a) `.` and a name (a `member-name-shorthand`, in the grammar of the RFC) or (b) `[`, a string literal, and `]`; no more of the grammar is allowed, not even whitespace. The allowed names and string literalss are as specified in RFC 9535, except that only double-quoted strings are allowed.
```
ezrasilvera, line 62: I think the following cn be removed.  " (which was originally and somewhat loosely defined in [an article by Stefan Goessner](https://goessner.net/articles/JsonPath/) and later defined more carefully in [RFC 9535](https://datatracker.ietf.org/doc/rfc9535/))."   In any case we basically say we don't fully support what is there ...

MikeSpreitzer: There is a reference to the RFC, for the definitions of the name and string literal syntax. So we need to keep at least a reference to that.

Is it really a problem to keep a reference to the original article, which is more widely known?

I suspect that we may expand the subset that is allowed here in the future. Having a reference for the background for that seems useful to me.

ezrasilvera: I think that people wants a very short & focused description in a user-guide , I don't see the purpose of the pointer to the article and the history statement, I'm not against pointing to the RFC.  I will not block the PR over it. 
@pdettori - as you requested this explanation I'll let you LGTM it if you are fine with it.

MikeSpreitzer: Some questions that the existing draft answers, that a reader might ask if we remove the reference to the Goessner article:

- "I never heard of this RFC, but I know about something called 'JSONPath' from an article by Stefan Goessner; are we talking about the same thing"?
- "I see that you did not reference the Goessner article; is that deliberate or did you not know about it? I am concerned that if you are not aware of the Goessner article then your future developments may unnecessarily diverge from the larger JSONPath community."
- "The Goessner article is shorter and sweeter; why did you not cite that?"

pdettori: I am fine with the explanation, thank you @MikeSpreitzer

pdettori: /lgtm

PR #2117, opened 2024-05-05
:sparkles: Combined Status API

### LINKED ISSUE #2138: [Combined-Status feature phase 1/5] User-Spec API Declaration
_This is the first of the 5 planned stages for Combined-Status #2017, and the first of the three mandatory ones._

Declare and document the new API - consisting of new API objects and modifications to existing ones.
The user-spec declaration portion should be done through BindingPolicies.

#### Completion Criterias
- API proposal, implementation and review
- Sufficient documentation that covers the API changes

vMaroon:
## Summary
This PR addresses #2017 in a new proposal that is based on @MikeSpreitzer's in #2055, with the flexibility intended to be achieved in my previous outline [here](https://docs.google.com/document/d/1kkJsdX7hfUzLl88i5t-i_Tcn30wDjOeM).
This proposal mainly inherits the essence Mike outlined, while making one significant change:
* StatusReturn, now named StatusCollector, is a separate API resource that can be defined and reused.

The motivation behind this change, as well as letting go of the first proposal documented is going for convenient simplicity. After conducting an analysis research on competitors, I'm convinced that KS wins in multi-cluster K8s if convenience and simplicity remain within top priorities.

Through separating the StatusCollector from being simply inlined in BindingPolicies, KubeStellar can allow the reuse of combined status templates. Users would be able to define them once, per resource of interest, and use them across BindingPolicies. In addition to this benefit, KubeStellar would provide a set of StatusCollectors for the built-in K8s resources, and later host a library of common use collectors. This way, a user can have full multi-cluster experience by simply adding one resource to their general workflow - a BindingPolicy.
* Enabling the use of default collectors is not in the API proposal yet - it will be PR'd along with the default collectors in the future.

## Notes:

- This PR removes parts of the API proposed in #2055 and leaves them in TODOs, such as augmented data that is injected in workload objects
- **An extension to this proposal is to allow for inlined StatusCollectors, as can be seen in the API (types.go)** - which would allow users to stack their own unique-case collectors that do not benefit from re-use, along with ones from KS common library
- The majority of the files/code that appear in the PR are generated
 
## Related issue(s)
- #2017 

## Fixes
Closes #2138


vMaroon:
CC @ezrasilvera @pdettori @MikeSpreitzer @nirrozenbaum

api/control/v1alpha1/types.go
```diff
@@ -56,6 +56,24 @@ const TemplateExpansionAnnotationKey string = "control.kubestellar.io/expand-tem
...
+// +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:resource:scope=Cluster,shortName={bp}
+type BindingPolicy struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   BindingPolicySpec   `json:"spec,omitempty"`
+	Status BindingPolicyStatus `json:"status,omitempty"`
+}
+
```
MikeSpreitzer, line 76: I like the changes that are just re-ordering content in this file. I think it may be better to do them first as a distinct PR.

api/control/v1alpha1/types.go
```diff
@@ -142,6 +136,18 @@ const (
...
+// DownsyncObjectTestAndStatusCollection identifies some objects (by a predicate)
+// and asks for some combined status to be returned from those objects.
+// The latter is dictated through applying a set of StatusCollectors,
+type DownsyncObjectTestAndStatusCollection struct {
+	DownsyncObjectTest `json:",inline"`
+	// statusCollectors is a list of StatusCollectors name references that are applied to the selected objects.
+	StatusCollectors []string `json:"statusCollectors,omitempty"`
+	// inlinedStatusCollectors is a list of StatusCollectorSpec that are inlined in the BindingPolicy.
+	// The StatusCollectors in this list are applied to the selected objects.
+	// InlinedStatusCollectors   []StatusCollectorSpec `json:"inlinedStatusCollectors,omitempty"` TODO
```
MikeSpreitzer, line 148: If we are going to go with the distinct API objects then I would not recommend adding the complexity of also having inlining.

api/control/v1alpha1/types.go
```diff
@@ -302,6 +315,211 @@ type BindingList struct {
...
+// StatusCollectorSpec defines the desired state of StatusCollector.
+type StatusCollectorSpec struct {
+	// `filter`, if given, is applied first.
+	// It must evaluate to a boolean or null (which is treated as false).
+	// This is like the WHERE clause in an SQL SELECT statement.
+	// +optional
+	Filter *Expression `json:"filter,omitempty"`
+
+	// `groupBy` says how to group workload objects for aggregation (if there is any).
+	// Each expression must evaluate to an atomic value.
```
MikeSpreitzer, line 340: I would add a comment that `groupBy` must be empty when `combinedFields` is.

api/control/v1alpha1/types.go
```diff
@@ -302,6 +315,211 @@ type BindingList struct {
...
+	Items           []StatusCollector `json:"items"`
+}
+
+// CombinedStatus holds the combined status from the WECs for one particular (workload object, BindingPolicy) pair.
+// The namespace of the CombinedStatus object is the namespace of the workload object,
+// or "kubestellar-report" if the workload object has no namespace.
+// The name of the CombinedStatus object is the concatenation of:
+// - the NamespacedName the workload object
+// - the string ":"
+// - the Name of the BindingPolicy object.
```
MikeSpreitzer, line 440: (A) NamespacedName is not a string, the reader wonders how it is concatenated into a string.
(B) For any reasonable answer to (A), the prescribed combination might overflow the limit on name length.

MikeSpreitzer, line 440: This combined name might be longer than a name is allowed to be.

api/control/v1alpha1/types.go
```diff
@@ -302,6 +315,211 @@ type BindingList struct {
...
+	// +optional
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []StatusCollector `json:"items"`
+}
+
+// CombinedStatus holds the combined status from the WECs for one particular (workload object, BindingPolicy) pair.
+// The namespace of the CombinedStatus object is the namespace of the workload object,
+// or "kubestellar-report" if the workload object has no namespace.
+// The name of the CombinedStatus object is the concatenation of:
+// - the NamespacedName the workload object
```
MikeSpreitzer, line 438: A NamespacedName is not a string, the reader wonders how it is to be concatenated into a string.

api/control/v1alpha1/types.go
```diff
@@ -302,6 +315,211 @@ type BindingList struct {
...
+// - "status.kubestellar.io/name" holding the name of the workload object;
+// - "status.kubestellar.io/binding-policy" holding the name of the BindingPolicy object.
+//
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:shortName={cs}
+// +kubebuilder:printcolumn:name="GROUP",type="string",JSONPath=".metadata.labels['status\\.kubestellar\\.io/api-group']"
+// +kubebuilder:printcolumn:name="RSC",type="string",JSONPath=".metadata.labels['status\\.kubestellar\\.io/resource']"
+// +kubebuilder:printcolumn:name="NS",type="string",JSONPath=".metadata.labels['status\\.kubestellar\\.io/namespace']"
+// +kubebuilder:printcolumn:name="NAME",type="string",JSONPath=".metadata.labels['status\\.kubestellar\\.io/name']"
```
MikeSpreitzer, line 454: There was some discussion in the context of #2055 about how to name these printer columns. The issue is that a CombinedStatus object already has a NAME, namespace, API group, and resource. For example, isn't there already a non-custom column named NAME? I think that where we landed (although it is not represented in #2055, I forget why) was to call them SUBJECT_GROUP, SUBJECT_RSC, SUBJECT_NS, and SUBJECT_NAME.

api/control/v1alpha1/types.go
```diff
@@ -302,6 +315,211 @@ type BindingList struct {
...
+// +kubebuilder:printcolumn:name="BINDINGPOLICY",type="string",JSONPath=".metadata.labels['status\\.kubestellar\\.io/policy']"
+type CombinedStatus struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// `results` has an entry for every applicable NamedStatusCombiner.
```
MikeSpreitzer, line 460: This text needs updating, "NamedStatusCombiner" is not a thing in this PR.

MikeSpreitzer:
I am OK with making the status combination prescriptions be independent API objects.

I see no value in rebasing this onto #2055.

I _do_ think that some very likely use cases will need the agumentation proposed in #2055. For example, producing a list of WECs where a given Deployment object does not have the prescribed number of available replicas. Producing a list of WECs that have not checked in recently enough. I am OK with adding that in follow-on work. I have been intending to write up some examples. I think that is very much needed.

api/control/v1alpha1/types.go
vMaroon: done.

api/control/v1alpha1/types.go
vMaroon: done.

api/control/v1alpha1/types.go
vMaroon: This can be pushed until it is clearly needed, if ever.

api/control/v1alpha1/types.go
vMaroon: We have a problem here: originally you suggested concatenating the UIDs of the object and the bindingpolicy. This combination overflows the name length limit which is 63 characters. 
A resource UID is a 128-bit number. Which follows the standard UUID format, which is 36 characters long, including 4 hyphens.

I think hashing OBJ_NAME:BP_NAME to 63 characters is good enough for now - what do you think? This is simple and effective enough.
Another option would be generating a name and maintaining it within the Binding or object. I'm fine with both.

api/control/v1alpha1/types.go
vMaroon: This should be handled through the above comment.

api/control/v1alpha1/types.go
vMaroon: done.

api/control/v1alpha1/types.go
```diff
@@ -302,6 +315,211 @@ type BindingList struct {
...
+	AggregatorTypeMin   AggregatorType = "MIN"
+	AggregatorTypeMax   AggregatorType = "MAX"
+)
+
+// Expression is some value to derive from a workload object from a WEC.
+// An Expression is either a JSONPath identifying a value to extract
+// or a boolean combination of comparisons of atomic values.
+// While the expression here in the Go type system admits other values,
+// the controller will accept only the restricted set stated above.
+type Expression struct {
```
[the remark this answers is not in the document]
vMaroon: done.

api/control/v1alpha1/types.go
vMaroon: If that is not a strong opinion, I'll keep them here.

vMaroon:
@MikeSpreitzer thanks, addressed comments. I intended to rebase on your draft since a lot of the code is copied from there. Since you do not mind, I also prefer not rebasing.

The steps that follow this PR are:
1. A PR that implements modifying the Binding API to reflect which collectors should be applied to which resources
    * An alternative to this would be an additional system resource that contains the decision - although at the moment this seems unlikely
1. A PR that implements creating CombinedStatus objects based on (1)
1. Extending the combined-status API to include what you originally proposed in Augmentations

api/control/v1alpha1/types.go
MikeSpreitzer: I am not aware of the limit on name length being 63. Where do you get that?

vMaroon: My bad, that was a google search highlight of the wrong place. This applies to label names.
Reverted back to your original suggestion.

api/control/v1alpha1/types.go
MikeSpreitzer: OK to leave them here.

api/control/v1alpha1/types.go
MikeSpreitzer: Looks like you read a "not" that I did not write.

api/control/v1alpha1/types.go
MikeSpreitzer: FYI, the actual restrictions that apply are not well documented. See https://github.com/kubernetes/website/issues/46262

api/control/v1alpha1/types.go
vMaroon: fixed.

MikeSpreitzer:
Yes, I think that the StatusCollector references should go in the BindingSpec.

api/control/v1alpha1/types.go
vMaroon: Thanks @MikeSpreitzer, is there anything else left to do in this PR? I have the next in line almost ready.

ezrasilvera:
@vMaroon Can you share here an example on how I define the policy in case I want the receive the full status from all WECs?   Can you also explain how the result will look like ? 
Personally I think that for this specific case users will expect to 1) no need to define any selection for fileds/keys , as they want the whole status 2) Get a vector or another collection of original object Status structure

MikeSpreitzer:
@ezrasilvera:

- A useful return of individual status includes the name of the WEC. That is one of the things that  "augmentation" is for.
- The approach to getting a list would be to SELECT augmentation.name, status.
- This is not designed for large lists. For full access to all the reported state, I think that we should have a separate procedural interface rather than inventing some scalable way to make another copy.

MikeSpreitzer:
I think we are all agreed that the plan is that user-facing documentation, CI testing, and contributor testing are included by the end of step 2.

ezrasilvera:
@MikeSpreitzer to clarify - I wasn't asking for docs.  I wanted to understand if this PR and approach will support users that are looking for the use case I mentioned, if the answer is "no - we would need another mechanism"  I'm OK with that, but I'll have to open a new issue to track that and we would need to see if indeed there are users waiting for that. 
So I just wanted to understand from @vMaroon if the feature I'm describing is included or not.

ezrasilvera:
To clarify my question - I'm looking on users that develop automation through informers on the returned status/objects when working on a single cluster, and want to do similar thing when working with KS across (small number) of clusters

vMaroon:
@ezrasilvera the return of a status copy or sub-status copy per WEC as separate objects is not a part of this API. Users can select status to fetch per WEC into a single CombinedStatus object instead. I understand the need for separate copies - but I agree with mike on that optimally we'd want a solution that doesn't just duplicate resources from the ITS to the WDS.
If until a clean pipeline for that case we do bear with duplication - it would be a simple extension to this API.

MikeSpreitzer:
@vMaroon: the design for augmentation allows for return of an object's status section from each WEC.

vMaroon:
@MikeSpreitzer to be sure we are talking about the same thing - Ezra is referring to fetching the status of objects per wec, into a separate containing object each.

As far as I saw on Augmentations, I thought the intention there was to conveniently add information that is not normally a part of the workload object, but one that is artificially added and can be utilized by the user, such as WEC name.

> // Augmentation defines what is implicitly added to every workload object from a WEC,
// for purposes of status collection.

The only actual status containing object in the proposal is CombinedStatus, which is per (workload-object, binding-policy) and not WEC.

ezrasilvera:
@vMaroonThanks for the answer. I agree with this. Let's revisit in case we see concrete request for this requirement (I'm not aware of one, I was just speculating :-) )

PR #2122, opened 2024-05-06
🐛 Implement usage of ITSName argument by the chart

francostellari:
## Summary

KS chart suggests that one could set ITSName: "" to something like its1 and expect the chart to use it and pass it to the controller. However, the value does not seem used in the Helm template.

This PR updates the chart to use it: `--its-name={{.Values.ITSName}}`

## Related issue(s)

Fixes #


MikeSpreitzer (approved):
Apparently we do not have an established test for this Helm chart that contributors can use. That should be fixed, as a separate issue.

/lgtm
/approve
Thanks!

MikeSpreitzer:
So I was wrong, our E2E tests exercise the current version of this chart.

MikeSpreitzer (commented):
/lgtm

PR #2127, opened 2024-05-07
🐛 Make binding controller use typed clients for known types

### LINKED ISSUE #1690: feature: use clientset and explicit informers for placement/placement-decision resources in placement controller
### Feature Description

Since placement and placement-decision are known API resources, their handling should not be generic in the controller code.

### Proposed Solution

Use proper clientset

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR changes the binding controller so that it uses typed clients and informers for the types that are known at development time (ManagedCluster, BindingPolicy, Binding). This fixes an observed problem where the dynamic client would sometimes be broken, depending on startup timing. One example error message was as follows.

```
E0506 20:56:54.575051       1 controller.go:548] error reconciling object (identifier: util.ObjectIdentifier{GVK:schema.GroupVersionKind{Group:"control.kubestellar.io", Version:"v1alpha1", Kind:"BindingPolicy"}, Resource:"bindingpolicies", ObjectName:cache.ObjectName{Namespace:"", Name:"nginx-bpolicy"}}): failed to handle bindingpolicy: failed to ocm.FindClustersBySelectors: error listing clusters: no matches for kind "ManagedCluster" in version "cluster.open-cluster-management.io/v1", requeuing
```

## Related issue(s)

Fixes #1690


pkg/binding/binding.go
```diff
@@ -22,46 +22,36 @@ import (
...
 	"k8s.io/klog/v2"
 
 	"github.com/kubestellar/kubestellar/api/control/v1alpha1"
-	"github.com/kubestellar/kubestellar/pkg/util"
 )
 
 // syncBinding syncs a binding object with what is resolved by the bindingpolicy resolver.
-func (c *Controller) syncBinding(ctx context.Context, objIdentifier util.ObjectIdentifier) error {
+func (c *Controller) syncBinding(ctx context.Context, bindingName string) error {
 	logger := klog.FromContext(ctx)
```
pdettori, line 32: why the change from objectIdentifier to a bindingName string ? Is this part of some other cleanup effort?

pkg/binding/binding.go
```diff
@@ -101,58 +93,23 @@ func (c *Controller) updateOrCreateBinding(ctx context.Context, bdg *v1alpha1.Bi
...
 	bdg.SetOwnerReferences([]metav1.OwnerReference{ownerReference})
 
-	// update or create binding
-	unstructuredBinding, err := bindingToUnstructuredObject(bdg)
-	if err != nil {
-		return fmt.Errorf("failed to convert Binding to Unstructured: %w", err)
-	}
-
 	logger := klog.FromContext(ctx)
+	bdgEcho, err := c.controlClient.Bindings().Update(ctx, bdg, metav1.UpdateOptions{FieldManager: controllerName})
```
pdettori, line 98: We did not use a field manager before, right ? This sounds like a good idea, just wondering if this is something that could create issues with existing bindings (if any) and if yes, if users should be informed about this.

pdettori (commented):
overall looks good to me, I just have couple of questions/comments

vMaroon:
I had previously closed a relevant issue that I thought is never going to get the time, reopened. @MikeSpreitzer 
- #1690

pkg/binding/controller.go
```diff
@@ -260,6 +301,23 @@ func (c *Controller) AppendKSResources(ctx context.Context) error {
 func (c *Controller) Start(parentCtx context.Context, workers int, cListers chan interface{}) error {
 	logger := klog.FromContext(parentCtx).WithName(controllerName)
 	ctx := klog.NewContext(parentCtx, logger)
+
+	// Create informer on managedclusters so we can re-evaluate BindingPolicies.
+	// This informer differ from the previous informers in that it listens on the ocm hub.
```
vMaroon, line 306: It would be more accurate to replace `previous` with `other` since this is the first reference to an informer in this body's call.

It would also be better for the eye if the calls that refer to the other informers are preceded with a new line since otherwise it may look like the comment refers to the entirety of the block formed by the connected lines. This is an opinion and preference on this matter, not a fact.

pkg/binding/binding.go
MikeSpreitzer: I think it is better to make the formal arguments take only the name, not suggest that there is a meaningful namespace input.

I do not think it makes much of a difference whether the conversion from `bindingRef` to string is done inside this method or not. I chose to do it in the logic closer to the workqueue because it seems specific to that.

pkg/binding/binding.go
MikeSpreitzer: As far as I know, we do not support in-place upgrades. If we did then I think it would take some more thorough review, documentation, and testing. And probably some real development work, I expect that issues would turn up.

pkg/binding/controller.go
MikeSpreitzer: The force-push to ddcd3f666 makes these changes.

vMaroon:
Thanks @MikeSpreitzer this is a nice change. Looks good to me - @pdettori can LGTM if the comments are resolved.

pkg/binding/binding.go
ezrasilvera: I agree with @MikeSpreitzer - I think we have an open issue to support upgrades, I also think it is very important, however I believe we have several other "gaps"  we need to close to support it besides this one.

ezrasilvera:
Thanks. Looks good to me, It's always better to use explicit client when possible. 
I'll leave the formal `approve`  to @pdettori  as he had one open comment ...
/lgtm

PR #2128, opened 2024-05-08
:sparkles:  Don't generate and push docs for main to kubestellar.io

ezrasilvera:
## Related issue(s)

Fixes #

PR #2129, opened 2024-05-08
:sparkles: Removed main from gh-pages

ezrasilvera:
This need to be merged after #2128 so that it's not re-genearted

## Related issue(s)

Fixes #


MikeSpreitzer (commented):
/lgtm

until #2128 merges:
/hold

PR #2134, opened 2024-05-09
🐛 Fixes to getRestConfig and callers

### LINKED ISSUE #2131: bug: control-plane waiting logic has gaps
### Describe the bug

https://github.com/kubestellar/kubestellar/blob/223360a38d01b5405c0056573c730ca846ca76d7/pkg/util/kubeconfig.go#L68-L125 will fail if there are some ITS control planes and a specific one is requested but does not exist yet. The func that is repeated until success should include looking for the specifically requested control plane, if there is one.

### Steps To Reproduce

See above.

### Expected Behavior

No spurious failure.

### Additional Context

I noticed this while working on #2127

### LINKED ISSUE #2133: bug: labelKey ignored in GetWDSKubeconfig
### Describe the bug

https://github.com/kubestellar/kubestellar/blob/223360a38d01b5405c0056573c730ca846ca76d7/pkg/util/kubeconfig.go#L73-L156 uses the given `labelKey` in log and error messages but not in the actual query.

### Steps To Reproduce

Look at the code.

### Expected Behavior

Either the parameter makes a difference or does not exist.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR removes the spurious parameter for the name of the label that holds a ControlPlane's "type" for KubeStellar purposes (WDS vs ITS).

This PR also fixes the wait loop in pkg/util.getRestConfig to wait for the right condition, not too much and not too little.

## Related issue(s)

Fixes #2131
Fixes #2133

### DIFF
--- a/pkg/util/kubeconfig.go
+++ b/pkg/util/kubeconfig.go
@@ -43,85 +43,68 @@ const (
 	ControlPlaneTypeWDS   = "wds"
 	// errors
 	ErrNoControlPlane        = "no control plane found. At least one control plane labeled with %s=%s must be present"
-	ErrControlPlaneNotFound  = "control plane with label %s=%s and name %s was not found"
-	ErrMultipleControlPlanes = "more than one control plane with same label %s=%s was found and no name was specified"
+	ErrControlPlaneNotFound  = "control plane with type %s and name %s was not found"
+	ErrMultipleControlPlanes = "more than one control plane of type %s was found and no name was specified"
 )
 
-func GetWDSKubeconfig(logger logr.Logger, wdsName, wdsLabel string) (*rest.Config, string, error) {
-	// default label for WDS
-	label := Label{
-		Key:   ControlPlaneTypeLabel,
-		Value: ControlPlaneTypeWDS,
-	}
-	// use the specified label instead of default label if provided
-	var err error
-	if wdsLabel != "" {
-		label, err = SplitLabelKeyAndValue(wdsLabel)
-		if err != nil {
-			return nil, "", err
-		}
-	}
-	logger.Info("using label", "key", label.Key, "value", label.Value)
-	return getRestConfig(logger, wdsName, label.Key, label.Value)
+func GetWDSKubeconfig(logger logr.Logger, wdsName string) (*rest.Config, string, error) {
+	return getRestConfig(logger, wdsName, ControlPlaneTypeWDS)
 }
 
 func GetITSKubeconfig(logger logr.Logger, itsName string) (*rest.Config, string, error) {
-	return getRestConfig(logger, itsName, ControlPlaneTypeLabel, ControlPlaneTypeITS)
+	return getRestConfig(logger, itsName, ControlPlaneTypeITS)
 }
 
 // get the rest config for a control plane based on labels and name
-func getRestConfig(logger logr.Logger, cpName, labelKey, labelValue string) (*rest.Config, string, error) {
+func getRestConfig(logger logr.Logger, cpName, labelValue string) (*rest.Config, string, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
 	defer cancel()
 
 	kubeClient := *kslclient.GetClient()
-	list := &kfv1aplha1.ControlPlaneList{}
 	labelSelector := labels.SelectorFromSet(labels.Set(map[string]string{
 		ControlPlaneTypeLabel: labelValue,
 	}))
+	var targetCP *kfv1aplha1.ControlPlane
 
 	// wait until there are control planes with supplied labels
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
/cc @pdettori 
/cc @vMaroon 
@francostellari

pkg/util/kubeconfig.go
```diff
@@ -61,7 +61,7 @@ func GetWDSKubeconfig(logger logr.Logger, wdsName, wdsLabel string) (*rest.Confi
 			return nil, "", err
 		}
 	}
-	logger.Info("using label", "key", label.Key, "value", label.Value)
+	logger.Info("Using label", "wdsName", wdsName, "keyJustKidding", label.Key, "value", label.Value)
```
pdettori, line 64: Why the "keyJustKidding" in the log message? It seems confusing. Also, it says "using label" but then it prints the wdsName and value. I would change the log pattern this way:

logger.Info("Using label",  "name", label.Key, "value", label.Value,"wdsName",wdsName)

pkg/util/kubeconfig.go
```diff
@@ -75,53 +75,50 @@ func getRestConfig(logger logr.Logger, cpName, labelKey, labelValue string) (*re
...
 	kubeClient := *kslclient.GetClient()
-	list := &kfv1aplha1.ControlPlaneList{}
 	labelSelector := labels.SelectorFromSet(labels.Set(map[string]string{
 		ControlPlaneTypeLabel: labelValue,
 	}))
+	var targetCP *kfv1aplha1.ControlPlane
 
 	// wait until there are control planes with supplied labels
-	logger.Info("waiting for cp with label", "key", labelKey, "value", labelValue)
+	logger.Info("Waiting for cp with label", "cpName", cpName, "keyJustKidding", labelKey, "value", labelValue)
```
pdettori, line 84: Same comment as above, I suggest changing to 

logger.Info("Waiting for cp with label",  "name", labelKey, "value", labelValue, "cpName", cpName)

pkg/util/kubeconfig.go
```diff
@@ -75,53 +75,50 @@ func getRestConfig(logger logr.Logger, cpName, labelKey, labelValue string) (*re
...
+		if cpName != "" {
+			cp := &kfv1aplha1.ControlPlane{}
+			err := kubeClient.Get(ctx, client.ObjectKey{Name: cpName}, cp, &client.GetOptions{})
+			if err == nil {
+				targetCP = cp
+				return true, nil
+			}
+			return false, nil
+		}
+		list := &kfv1aplha1.ControlPlaneList{}
```
pdettori, line 95: The original intent when this code was initially written was to allow selection of a WDS or ITS either by name or label. If the user does not specify a label, the default label is used. So far, the only usage pattern has been to specify the name for the WDS and just use the default label for the ITS (that is, an ITS name was not required).  Since kubeflex may create also control plane that may not necessarily be used as WDS or ITSs it may still make sense to require to have a label in each to indicate the role. If we relax the restriction on ITSs, we should allow more then one but then we would need to associate WDSs with ITSs.

pkg/util/kubeconfig.go
MikeSpreitzer: The "JustKidding" part is an interim marker that this needs to be fixed as part of fixing #2133 , and a warning to anyone who reads such a log message before the fix.

pkg/util/kubeconfig.go
MikeSpreitzer: > So far, the only usage pattern has been to specify the name for the WDS and just use the default label for the ITS

We have begun changing this, allowing more than one ITS cp in KubeFlex. @francostellari is working right now on an all-in-one chart that allows a general number of WDSes, a general number of ITSes, and a general selection of ITS for each WDS.

> it may still make sense to require to have a label in each to indicate the role

Indeed it might, perhaps for general orderliness or informing something else. But when the user has indicated the name of the ITS control plane to use, I see no reason to refuse based on labels.

pkg/util/kubeconfig.go
MikeSpreitzer: The force-push to 9df116506 changes the order as suggested here.

pkg/util/kubeconfig.go
MikeSpreitzer: The force-push to https://github.com/kubestellar/kubestellar/commit/9df1165065fa60b4c82b04e2c716b51735b26af5 changes the order as suggested here.

pkg/util/kubeconfig.go
MikeSpreitzer: The force-push to 359b169b7 makes this moot.

pdettori (commented):
/lgtm

PR #2137, opened 2024-05-09
🐛 Mostly enable E2E test of current code on OCP

### LINKED ISSUE #2071: bug: test/e2e/multi-cluster-deployment/run-test.sh does not support testing a contributor's work on OCP
### Describe the bug

`test/e2e/multi-cluster-deployment/run-test.sh`, in the OCP case, can only test a release; it can _not_ test a contributor's modified version of the repo. This fails to meet a critical requirement on the E2E tests.

### Steps To Reproduce

Look at the script or read the special case for OCP in the README.md.

### Expected Behavior

As for `kind`, when using OCP this test can test a contributor's modified version of the repo.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR endeavors to make `test/e2e/multi-cluster-deployment` able to test the local repository contents rather than a release.

It falls short in just one way, as far as I know. That is loading the locally built container images (for the kubestellar controller-manager and the transport-controller) into some registry that the hosting cluster can pull from. Is https://docs.openshift.com/container-platform/4.15/registry/index.html the right answer here?

## Related issue(s)

Fixes #2071


scripts/deploy-transport-controller.sh
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Transport deployment script assumes it runs using the kubeflex hosting cluster context.
+# This script installs the transport controller in the current kubeconfig context
```
ezrasilvera, line 16: nit:  Maybe "using the current .."  ?

test/e2e/multi-cluster-deployment/run-test.sh
```diff
@@ -48,27 +48,21 @@ while [ $# != 0 ]; do
...
-    "${HACK_DIR}/check_pre_req.sh" --assert --verbose kubectl docker kind make go ko yq helm kflex ocm
+SRC_DIR="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)"
+COMMON_SRCS="${SRC_DIR}/../common"
+HACK_DIR="${SRC_DIR}/../../../hack"
 
-    "${COMMON_SRCS}/cleanup.sh"
-    source "${COMMON_SRCS}/setup-shell.sh"
-    "${COMMON_SRCS}/setup-kubestellar.sh" $setup_flags
-    "${SRC_DIR}/use-kubestellar.sh"
+"${HACK_DIR}/check_pre_req.sh" --assert --verbose kubectl docker kind make go ko yq helm kflex ocm
```
ezrasilvera, line 63: Don't we run this also for the "release" tests?  In that case I don't think we require the ability to build, so I'm not sure all these packages are mandatory. Enforcing it here may contradict our docs for users.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -109,25 +131,25 @@ echo "wds1 created."
...
   echo "replace github.com/kubestellar/kubestellar => ${KUBESTELLAR_DIR}/" >> go.mod
-  go mod tidy # TODO to be deleted next time we bump ocm transport release (done in ocm transport makefile)
   IMAGE_TAG=${OCM_TRANSPORT_PLUGIN_RELEASE} make ko-build-local
   kind load --name kubeflex docker-image ko.local/transport-controller:${OCM_TRANSPORT_PLUGIN_RELEASE} # load local image to kubeflex
   cd "${KUBESTELLAR_DIR}"
   pwd
   rm -rf ${OCM_TRANSPORT_PLUGIN_DIR}
   echo "running ocm transport plugin..."
-  kubectl config use-context kind-kubeflex ## transport deployment script assumes it runs within kubeflex context
+  kubectl config use-context "$HOSTING_CONTEXT" ## transport deployment script works with current context
```
ezrasilvera, line 148: Wouldn't it be better to remove this and change the deployment script to get a context name parameter, and have that script use the context directly?  
Setting context has always bad side effects , for example an error may leave us pointing to a different context than the one we started with, etc..

test/e2e/multi-cluster-deployment/run-test.sh
MikeSpreitzer: This list of checked prereqs is not being changed in this PR; can we address it in a follow-on PR?

scripts/deploy-transport-controller.sh
MikeSpreitzer: Moot now.

test/e2e/common/setup-kubestellar.sh
MikeSpreitzer: Done.

scripts/deploy-transport-controller.sh
```diff
@@ -13,19 +13,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Transport deployment script assumes it runs using the kubeflex hosting cluster context.
-
-
 DEFAULT_TRANSPORT_CONTROLLER_IMAGE=ghcr.io/kubestellar/ocm-transport-plugin/transport-controller:0.1.7
 
+set_context=""
```
ezrasilvera, line 18: set_context is confusing.  seems like some Boolean flag.   tc-context ?  or even just "context"  ?

MikeSpreitzer: The idea was to not mess with `--context` if not requested.

But it was buggy. So I have changed it to always give `--context` to kubectl.

PR #2145, opened 2024-05-09
🌱 Update kubeflex dependency to release 0.6.1

MikeSpreitzer:
## Summary
This PR notes that the required version of KubeFlex has increased to 0.6.1.

Note that `hack/check_pre_req.sh` does not check whether a given dependency has an adequate version.

## Related issue(s)

Fixes #


pdettori (commented):
/lgtm

PR #2146, opened 2024-05-09
:book: 📖 docs Update docs\README.md with note about navigation definition

KPRoche:
## Summary
After searching for quite some time while working to design better site structure, I finally figured out where the site navigation menu is hiding (in the mkdocs.yml). 

I added a note in the readme pointing out that the site navigation is also defined in the mkdocs.yml


MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

Issue #2104, opened 2024-05-02
feature: single script that activates both bash and ginkgo e2e tests


dumb0002:
### Feature Description

To improve the current e2e tests with a single script that activates both bash and ginkgo tests in both kind and ocp clusters

### Proposed Solution

Extend the [run-test.sh](https://github.com/kubestellar/kubestellar/blob/release-0.22.0-rc2/test/e2e/multi-cluster-deployment/run-test.sh) script to support executing of ginkgo tests for both kind and ocp clusters

### Want to contribute?

- [ ] I would like to work on this issue.

PR #2153, opened 2024-05-10
✨  Single script that activates e2e tests for bash and ginkgo tests

dumb0002:
## Summary

This PR improves the current e2e tests with a single script that activates both bash and ginkgo tests for kind and ocp clusters

## Related issue(s)

https://github.com/kubestellar/kubestellar/issues/2104

Fixes #

### DIFF
--- a/test/e2e/common/cleanup.sh
+++ b/test/e2e/common/cleanup.sh
@@ -41,7 +41,7 @@ elif [ $env == "ocp" ];then
         cluster=$1
 
         kubectl --context $cluster delete ns nginx --ignore-not-found
-        clusteradm unjoin --cluster-name $cluster
+        clusteradm unjoin --cluster-name $cluster  2> /dev/null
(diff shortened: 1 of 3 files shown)


effi-ofer:
Thanks for taking on this task Braulio.  It would be nice to also see the ginkgo driver file supporting the env=ocp parameter, but your solution is good enough for me.  /LGTM.

effi-ofer (approved):

MikeSpreitzer:
I like the idea of a script that invokes both of the existing tests. But note that the two existing tests run different scenarios. Putting the script that invokes both in the directory for one of the two scenarios is confusing.

dumb0002:
@MikeSpreitzer, maybe we could promote the script [use-kubestellar.sh](https://github.com/kubestellar/kubestellar/blob/main/test/e2e/multi-cluster-deployment/use-kubestellar.sh) to a new directory called `/bash` (that will be a better description for the test in the bash scenario)

MikeSpreitzer:
I think that we are pursuing the re-org under #2160

PR #2154, opened 2024-05-11
✨ Create script to upload container image to OCP cluster

MikeSpreitzer:
## Summary
This PR adds a script that uploads a local container image to an OCP cluster.
This is the key missing piece of technology that stopped #2137 from being able to test the local version of the repo on OCP; there remains a little work to integrate this new script into the E2E testing.

## Related issue(s)

Fixes #


MikeSpreitzer:
Successfully tested with the EdgePlatform1 cluster.

scripts/ocp-load-image.sh
```diff
@@ -0,0 +1,93 @@
...
+local_image="$1"
+shift
+remote_name_tag="$1"
+shift
+
+set -x
+set -e
+
+if [ "$context" == "." ]; then
+    context=$(kubectl config current-context)
```
ezrasilvera, line 66: Why not have the context optional and have the default `current-context`  ?

scripts/ocp-load-image.sh
```diff
@@ -0,0 +1,93 @@
...
+regint=$(kubectl --context "$context" get images.config cluster -o 'jsonpath={.status.internalRegistryHostname}')
+
+if [ -z "$regint" ]; then
+    echo "$0: Failed to get internal hostname for OpenShift image registry" >&2
+    exit 20
+fi
+echo "internalRegistryHostname=$regint"
+echo "in-cluster-image-ref=${regint}/openshift/$remote_name_tag"
+
+oc --context "$context" whoami -t | docker login -u kubeadmin --password-stdin "$regext"
```
ezrasilvera, line 90: Seems the script prompts for a password ?  Why don't we give the option to specify password and then pass it to docker login ?
This will allow users to automate if they need to ...

MikeSpreitzer: See how the input to `docker login` is coming from the output of another command? The user of the script is not providing the password input to `docker login`.

Issue #2150, opened 2024-05-10
Combined-Status Feature Integration/Ginko Tests


vMaroon:
Following the merging of the API, the writing of e2e and integration tests that cover the feature is now possible independently of the feature's implementation as outlined in:
- #2143

If a tests-only PR arises out of this issue, it would remain open and be rebased-on in the 3rd implementation step of the feature which would address #2141 , then merged together with the latter.

vMaroon:
/cc @ezrasilvera 
/cc @MikeSpreitzer 
/cc @effi-ofer 

It would be helpful if someone can pick this up.

clubanderson:
Stale issues rot after 30d of inactivity.
Mark the issue as fresh with `/remove-lifecycle rotten`.
Rotten issues close after an additional 30d of inactivity.

If this issue is safe to close now please do so with `/close`.

/lifecycle rotten

clubanderson:
Rotten issues close after 30d of inactivity.
Reopen the issue with `/reopen`.
Mark the issue as fresh with `/remove-lifecycle rotten`.

/close

Issue #2143, opened 2024-05-09
epic: Combined Status from multiple WECs


vMaroon:
## Summary
Originally started in #2017, this feature extends KubeStellar with the capability of allowing users to define, within the mechanism for workload distribution, how to retrieve and aggregate status for workload objects.

The manner in which the latter is done is through an API that is analogous to a simple SQL SELECT statement, in which a user can request a list without aggregation, possibly after filtering, but certainly with a limit on list length.

## Phased Approach Outline
This  feature will be implemented in a phased approach of three  PRs. 

### Essential PRs Phase
Implementing the functionality of the feature is split into three stages:
- Declaring new API resources and API documentation: this part only declares the user-spec portion of the API
    - Issue: #2138
- Implementing the calculation of user-spec API declared in (1)
    - Issue: #2140
- Calculating and generating combined-status (CombinedStatus) objects
    - Issue: #2141

clubanderson:
Stale issues rot after 30d of inactivity.
Mark the issue as fresh with `/remove-lifecycle rotten`.
Rotten issues close after an additional 30d of inactivity.

If this issue is safe to close now please do so with `/close`.

/lifecycle rotten

clubanderson:
Rotten issues close after 30d of inactivity.
Reopen the issue with `/reopen`.
Mark the issue as fresh with `/remove-lifecycle rotten`.

/close

PR #2155, opened 2024-05-11
:sparkles: Combined Status User-Spec-API Result API & Calculation

### LINKED ISSUE #2140: [Combined-Status feature phase 2/3] User-Spec API Result API & Calculation
_This is the second of the 3 mandatory planned PRs for Combined-Status #2143._

Given user-spec combined status API declaration through BindingPolicies, the selection/calculation results should appear in the associated Binding objects.

#### Completion Criteria
- Binding API updates that reflect the status to collect/combine as specified by the new API in BindingPolicies
- Binding controller updates that calculate the above
- Sufficient documentation that covers the API changes
- Testing

vMaroon:
## Summary
This PR is the 2nd of the ones planned in epic #2143. This PR:
- Introduces API updates to Bindings that reflect association with StatusCollectors
- Extends and modifies the Binding controller to calculate these associations

### TODO: documentation

Notes:
- I chose to save a list of all used StatusCollectors at the Binding's spec flat level, then reference the list through indices per selected object (if was selected by a test that has StatusCollectors)
    - I have no strong preference over keeping the API this way or duplicating references. Opinions are welcome. 
- **TESTING**: 
    - I have not extended the current tests because we practice testing the expected end state, and at the moment we only have the intermediate state of calculated Bindings. I have manually covered the following:
        - BindingPolicies selecting objects without StatusCollectors
        - BindingPolicies selecting an object with multiple StatusCollectors
            - Modifications to the latter
        - BindingPolicies selecting two objects with all possible variations on 2 objects x 2 StatusCollectors
    - If @MikeSpreitzer or anyone else believe that tests that check for Binding states are essential at this step, I will execute. Though recall that #2150 should cover automated testing and is a completion criteria of the epic

## Related issue(s)
- #2143 
- #2140

Closes #2140


api/control/v1alpha1/types.go
```diff
@@ -403,7 +411,9 @@ const (
 type Expression struct {
 	Op   ExpressionOperator `json:"op"`
 	Path string             `json:"path,omitempty"`
-	Args []Expression       `json:"args,omitempty"`
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Schemaless
```
MikeSpreitzer, line 415: Why mark this as schemalesss? (https://book.kubebuilder.io/reference/markers/crd-validation)

api/control/v1alpha1/types.go
```diff
@@ -252,6 +252,10 @@ type BindingSpec struct {
 
 	// `destinations` is a list of cluster-identifiers that the objects should be propagated to.
 	Destinations []Destination `json:"destinations,omitempty"`
+
+	// `statusCollectors` is a list of StatusCollectors name references that appear in the associated BindingPolicy
+	// and are applied to one or more of the objects in the `workload` field.
+	StatusCollectors []string `json:"statusCollectors,omitempty"`
 }
 
 // DownsyncObjectReferences defines the objects to be down-synced, grouping them by scope.
```
MikeSpreitzer, line 261: Analogously to what was done for BindingPolicy, I would not pretend that the status collector reference is part of the object reference. These are two distinct concepts. Were they are paired, have a pair type and types for the two individual components.

api/control/v1alpha1/types.go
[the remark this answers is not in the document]
vMaroon: The generated CRD cannot be applied due to validation fails - caused by Expression being a recursive field.

I followed this thread https://github.com/kubernetes-sigs/controller-tools/issues/585 - if you are aware of a better solution, please point me to it.

api/control/v1alpha1/types.go
```diff
@@ -278,6 +282,8 @@ type NamespaceScopeDownsyncObject struct {
 	Name string `json:"name"`
 	// `resourceVersion` is the version of the resource to downsync.
 	ResourceVersion string `json:"resourceVersion"`
+	// `statusCollectorIndices` is a list of indices in spec.statusCollectors that should be applied to the object.
+	StatusCollectorIndices []int `json:"statusCollectorIndices,omitempty"`
```
MikeSpreitzer, line 286: I understand the desire for economy, but note that using an index into a list here adds a dependency and a source of changes: the index depends on the contents and ordering in the list. At the very least, we should do something about the stability of the ordering for the items in BindingSpec.StatusCollectors. An insertion requires changes in NamespaceScopeDownsyncObject values that would not otherwise change. I am thinking about future efficiency improvements, in which the controllers become more incremental and driven by fine-grain changes. If this were a set of StatusCollector names instead of a list of indices into BindingSpec.StatusCollectors, then irrelevant changes would not affect a given NamespaceScopeDownsyncObject[AndStatusCollectors].

Also, if you put a premium on compactness that you might consider a representation pairs a single object reference to a single reference to a set of StatusCollector names, rather than directly with a set (of index or name). For example, the BindingSpec might have what is logically a map from `StatusCollectorSetName string` to Set[StatusColllector name] (but, following Kubernetes design rules, is represented by a slice of `struct {StatusCollectorSetName string, StatusCollectorNames []string}`).

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -66,6 +69,37 @@ func (resolution *bindingPolicyResolution) ensureObjectIdentifierWithVersion(obj
...
+	}
+
+	// if the object identifier does not exist in the statuscollectors map, add it
+	if _, exists := resolution.objectIdentifierToStatusCollectors[objIdentifier]; !exists {
+		resolution.objectIdentifierToStatusCollectors[objIdentifier] = statusCollectors
+		return true
+	}
+
+	// if the object identifier's statuscollectors set is already equal the given set, do nothing
+	if resolution.objectIdentifierToStatusCollectors[objIdentifier].Equal(statusCollectors) {
```
MikeSpreitzer, line 95: This does three hash table lookups but could be changed to do only one.

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -36,8 +37,10 @@ import (
 type bindingPolicyResolution struct {
 	sync.RWMutex
 
-	objectIdentifierToResourceVersion map[util.ObjectIdentifier]string
-	destinations                      sets.Set[string]
+	objectIdentifierToResourceVersion  map[util.ObjectIdentifier]string
+	objectIdentifierToStatusCollectors map[util.ObjectIdentifier]sets.Set[string]
```
MikeSpreitzer, line 41: I think the code would be simpler and more efficient if these two maps were merged into one, something like the following.

```
objectToData map[util.ObjectIdentifier] objectData
```

where

```
type objectData struct {ResourceVersion string; StatusCollectors sets.Set[string]}
```

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -184,38 +254,43 @@ func destinationsMatch(resolvedDestinations sets.Set[string], bindingDestination
 	return true
 }
 
-type objectRefAndVersion struct {
+type objectRef struct {
```
MikeSpreitzer, line 257: Why the renaming? This is still an object reference plus ResourceVersion.

pkg/binding/selectors.go
```diff
@@ -139,6 +116,52 @@ func (c *Controller) updateResolutions(ctx context.Context, objIdentifier util.O
...
+		}
+
+		return fmt.Errorf("failed to update resolution for bindingpolicy %s for object (identifier: %v): %v",
+			objResourceVersion, objIdentifier, err)
+	}
+
+	resolutionStatusCollectorsChanged, err := c.bindingPolicyResolver.SetStatusCollectorsForObject(bindingPolicyName,
+		objIdentifier, matchedStatusCollectorsSet)
+	if err != nil {
+		return fmt.Errorf("failed to set statuscollectors for object (identifier: %v): %v", objIdentifier,
```
MikeSpreitzer, line 141: `errorIsBindingPolicyResolutionNotFound(err)` needs to be checked here too, it could have become true since the last check.
Better yet, merge EnsureObjectIdentifierWithVersion and SetStatusCollectorsForObject.

MikeSpreitzer (commented):
I have finished my first review.

api/control/v1alpha1/types.go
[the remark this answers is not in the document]
MikeSpreitzer: Oh!

The Kubernetes API design rules strongly prefer "structural" schemas. And I already have one criticism of the definition of "structural". Let me review this situation.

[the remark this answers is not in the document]
MikeSpreitzer: BTW, did you notice that `Expression` can not be a literal? Probably an oversight.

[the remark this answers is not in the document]
vMaroon: Yes. I thought that we want it to start like this for simplicity. Allowing literals would make adding more `ExpressionOperator` kinds reasonable. I think we should consider this an enhancement to the API and follow it up later.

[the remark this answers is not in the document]
MikeSpreitzer: FYI, https://github.com/kubernetes/apiextensions-apiserver/blob/v0.30.0/pkg/apis/apiextensions/validation/validation.go#L1841-L1853 . I know this is not exactly the same question, but it is close. It looks to me like this forbids creation of a CRD with `Spec.PreserveUnknownFields` specifying true.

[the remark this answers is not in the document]
MikeSpreitzer: I think that literals are needed for even some pretty basic use cases. Their omission was an oversight.

See #2156 for one way to solve the problem with the recursive type here.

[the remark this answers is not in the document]
MikeSpreitzer: For the problem with the recursive type structure, we could respond in any of several ways.

1. Do not model the structure of an Expression in the OpenAPI schema (which is what the current draft of this PR does). That is allowed but gives up on the usual aspiration of modeling.
2. Remove recursion from Expression by flatting boolean expressions to a normalized form (such as DNF).
3. Just represent an expression in some string syntax, inventing a syntax as expressive as Expression currently is.
4. Represent an expression as a string in CEL syntax, with CEL semantics (#2156 makes the API change for that).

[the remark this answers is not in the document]
vMaroon: I'm ok with (4) although it should be accommodated with the appropriate doc updates (commented on that PR). BTW - this is more relevant to the PR next in line, this issue should not be a blocker for this one.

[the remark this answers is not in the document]
MikeSpreitzer: #2156 made that switch.

api/control/v1alpha1/types.go
vMaroon: done

api/control/v1alpha1/types.go
vMaroon: for simplicity and since the last option can be equivalent to the simpler choice of going for set of names per object, I chose the latter.

pkg/binding/bindingpolicy-resolution.go
vMaroon: done

pkg/binding/selectors.go
vMaroon: done

vMaroon:
@MikeSpreitzer thanks and apologies for the delayed response. Addressed your comments.
The force-push rebases on main.

api/control/v1alpha1/types.go
```diff
@@ -260,11 +260,19 @@ type BindingSpec struct {
...
 	// `namespaceScope` holds a list of namespace-scoped object references with their associated
 	// resource versions to downsync.
-	NamespaceScope []NamespaceScopeDownsyncObject `json:"namespaceScope,omitempty"`
+	NamespaceScope []NamespaceScopeDownsyncObjectAndStatusCollectors `json:"namespaceScope,omitempty"`
+}
+
+// NamespaceScopeDownsyncObjectAndStatusCollectors represents a specific namespace-scoped object to downsync,
+// and the status collectors that should be applied to it.
+type NamespaceScopeDownsyncObjectAndStatusCollectors struct {
+	*NamespaceScopeDownsyncObject `json:",inline"`
```
MikeSpreitzer, line 273: The level of pointer indirection here is surprising/unusual. In a kube API object, pointer indirection is usually associated with a bit of substructure being optional. But that is not what is happening here.

For an example in Kubernetes, consider the slice of Container in a PodSpec (Container is to PodSpec as NamespaceScopeDownsyncObjectAndStatusCollectors is to DownsyncObjectReferences).  A Container has a non-optional field `Resources ResourceRequirements` --- no pointer.

api/control/v1alpha1/types.go
```diff
@@ -260,11 +260,19 @@ type BindingSpec struct {
 type DownsyncObjectReferences struct {
```
MikeSpreitzer, line 260: This type name is still misleading. Something like `DownsyncObjectReferencesWithStatusCollectors` would be more accurate.

api/control/v1alpha1/types.go
```diff
@@ -260,11 +260,19 @@ type BindingSpec struct {
...
-	ClusterScope []ClusterScopeDownsyncObject `json:"clusterScope,omitempty"`
+	ClusterScope []ClusterScopeDownsyncObjectAndStatusCollectors `json:"clusterScope,omitempty"`
 
 	// `namespaceScope` holds a list of namespace-scoped object references with their associated
 	// resource versions to downsync.
-	NamespaceScope []NamespaceScopeDownsyncObject `json:"namespaceScope,omitempty"`
+	NamespaceScope []NamespaceScopeDownsyncObjectAndStatusCollectors `json:"namespaceScope,omitempty"`
+}
+
+// NamespaceScopeDownsyncObjectAndStatusCollectors represents a specific namespace-scoped object to downsync,
```
MikeSpreitzer, line 270: s/represents/references/

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -48,21 +49,41 @@ type bindingPolicyResolution struct {
...
 
-// ensureObjectIdentifierWithVersion ensures that an object identifier exists
-// in the resolution and is associated with the given resource version.
+// objectData stores the resource version and status collectors for an object.
+type objectData struct {
+	ResourceVersion  string
+	StatusCollectors sets.Set[string]
+}
+
+// ensureObjectIdentifier ensures that an object identifier exists
```
MikeSpreitzer, line 58: FYI, I think that the name of this method would be clearer if it talked about more than the identifier for the object. Something like `ensureObjectData` might be better.

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -184,38 +215,45 @@ func destinationsMatch(resolvedDestinations sets.Set[string], bindingDestination
...
+		bindingObjectRefAndDataSet.Insert(objectRefAndData{
 			GroupVersionResource: schema.GroupVersionResource(clusterScopeDownsyncObject.GroupVersionResource),
 			ObjectName: cache.ObjectName{
 				Name:      clusterScopeDownsyncObject.Name,
 				Namespace: metav1.NamespaceNone,
 			},
-			ResourceVersion: clusterScopeDownsyncObject.ResourceVersion,
+			objectData: &objectData{
+				ResourceVersion:  clusterScopeDownsyncObject.ResourceVersion,
+				StatusCollectors: sets.New[string](clusterScopeDownsyncObject.StatusCollectors...),
```
MikeSpreitzer, line 237: FYI, I think that you can s/[string]// here, Go's type inference will figure it out.

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -148,21 +177,23 @@ func (resolution *bindingPolicyResolution) matchesBindingSpec(bindingSpec *v1alp
...
 	}
 
-	objectRefSetFromWorkload := bindingObjectRefAndVersionSetFromWorkload(&bindingSpec.Workload)
+	objRefAndDataSetFromWorkload := bindingObjectRefAndDataSetFromWorkload(&bindingSpec.Workload)
 
-	for objIdentifier, objResourceVersion := range resolution.objectIdentifierToResourceVersion {
-		if !objectRefSetFromWorkload.Has(objectRefAndVersion{
+	for objIdentifier, objData := range resolution.objectIdentifierToData {
+		// check if object ref exists, then check if the object data matches
+		if !objRefAndDataSetFromWorkload.Has(objectRefAndData{
```
MikeSpreitzer, line 189: This is going to compare `objectRefAndData` according to Go `==` --- which is not what you want, because that will compare pointer values for the `*objectData` field rather than doing a more semantic equality test.

MikeSpreitzer (commented):
I have finished another review of this PR.

pkg/binding/bindingpolicy-resolution.go
vMaroon: Apologies, thought I fixed that.

api/control/v1alpha1/types.go
vMaroon: Thanks

vMaroon:
@MikeSpreitzer thanks, addressed all.

api/control/v1alpha1/types.go
MikeSpreitzer: I see that the comment on ClusterScopeDownsyncObjectAndStatusCollectors got changed, but the comment here still says "represents".

MikeSpreitzer (commented):
Looks good, maybe one editing or git push oversight?

api/control/v1alpha1/types.go
vMaroon: @MikeSpreitzer done

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #2156, opened 2024-05-13
✨ Switch StatusCollector expressions to CEL

MikeSpreitzer:
## Summary
This PR changes the StatusCollector API to use CEL rather than a bespoke API for expressions.

I had proposed a bespoke API for the sake of keeping down conceptual and runtime complexity. But I forgot that Kubernetes API machinery does not like recursive types.

We could flatten out the old Expression type by forcing the boolean expressions into a normalized form (such as disjunctive normal form).

We could define our own string syntax for our own form of expressions.

Or we could bite the bullet and switch to using CEL.

Still to do: define the limits that we will impose on cost. I think that we should start by disabling the macros that have non-constant cost. Then perhaps impose a limited based on looking at some examples.

## Related issue(s)

This is one way to remove a problem encountered in #2155


MikeSpreitzer:
This also raises the urgency of answering #1538

ezrasilvera:
@MikeSpreitzer I saw a lot of complaints about performance issues with the CEL processing. I suspect this is not an issue in our case - when do we evaluate/process   the expression ?  Is it only once, when the manifest is processed ?

MikeSpreitzer:
@ezrasilvera: can you please tell us how to find those complaints?

Depending on the use case, there is a trust issue relevant to performance. Specifically, when the admin of the controllers does not trust the authors of the expressions to write something that is not painfully expensive.

It remains to be seen how effectively we can educate users on the costs.

The expressions that apply to a given object have to be re-evaluated every time the summarizing controller processes a new version of that object.

api/control/v1alpha1/types.go
```diff
@@ -394,27 +394,14 @@ const (
...
-
-const (
-	OperatorPath  ExpressionOperator = "Path" // JSONPath string
-	OperatorOr    ExpressionOperator = "Or"
-	OperatorAnd   ExpressionOperator = "And"
-	OperatorNot   ExpressionOperator = "Not"
-	OperatorEqual ExpressionOperator = "Equal"
-)
+// Expression is written in the Common Expression Language.
+// See github.com/google/cel-go/cel for the Go implementatiom used in Kubernetes.
```
vMaroon, line 398: `github.com/google/cel-go/cel` does not exist.

api/control/v1alpha1/types.go
```diff
@@ -394,27 +394,14 @@ const (
...
+// The expression will be type-checked against the schema for the object type at hand,
+// using the Kubernetes library code for converting an OpenAPI schema to a CEL type
+// (e.g., https://github.com/kubernetes/apiserver/blob/v0.28.2/pkg/cel/common/schemas.go#L40).
+// Parsing errors are posted to the status.Errors of the StatusCollector.
+// Type checking errors are posted to the status.Errors of the Binding and BindingPolicy.
+type Expression string
```
vMaroon, line 404: Can you update documentation to reflect the (base of the) expected schema?

api/control/v1alpha1/types.go
MikeSpreitzer: Sorry, I meant github.com/google/cel-go

api/control/v1alpha1/types.go
MikeSpreitzer: What do you mean by "base of the"?

vMaroon: As in the first version of the schema, maybe a redundant phrase. The intention is to have the expected and supported schema documented for users to know. I assumed that it would be equivalent to how `Expression` was defined previously.

MikeSpreitzer: I am still not sure I understand the question. The point of this PR is to change from the previous formulation of Expression to using CEL. See how the comment on this type says that the expression is written in the Common Expression Language?

api/control/v1alpha1/types.go
MikeSpreitzer: The force-push to 46fc894fa fixes this typo.

api/control/v1alpha1/types.go
vMaroon: I confused it with something else. It may be useful to add a link to this - https://kubernetes.io/docs/reference/using-api/cel/  - either here or in the documentation.

Also, line 398: s/implementatiom/implementation

Issue #2091, opened 2024-04-29
bug: `kflex ctx` does not work after that context is renamed


MikeSpreitzer:
### Describe the bug

I am debugging `deploy-on-k3d.md` and having trouble with the kubeflex hosting context after it is renamed. Here is a demonstration of the problem, showing that `kflex ctx` switches the current context to the original hosting context name regardless of any renaming that it has been through since then.

```shell
mspreitz@mjs13 testks % kubectl config get-contexts | fgrep \*
*         kubeflex                                                                               k3d-kubeflex                                     admin@k3d-kubeflex                                                                

mspreitz@mjs13 testks % kubectl config rename-context kubeflex kf
Context "kubeflex" renamed to "kf".

mspreitz@mjs13 testks % kubectl config current-context
kf

mspreitz@mjs13 testks % kflex ctx
✔ Checking for saved initial context...
✔ Switching to initial context...

mspreitz@mjs13 testks % kubectl config current-context
kind-kubeflex

mspreitz@mjs13 testks % kubectl config get-contexts | grep kubeflex
          kf                                                                                     k3d-kubeflex                                     admin@k3d-kubeflex                                                                
```

### Steps To Reproduce

See above.

### Expected Behavior

Not sure.

MikeSpreitzer:
@pdettori 
@eaepstein

pdettori:
For background, here is some info on how the initial context is maintained in the kubeconfig file: https://github.com/kubestellar/kubeflex/blob/main/docs/users.md#initial-context

Since kflex cannot know that that initial context has been renamed by another command (in  this case kubectl), I think that there are couple of possible (non exclusive) approaches around this issue:

1. Document that if the user renames the context for the KubeFlex hosting cluster, will need also to manually edit the Kubeconfig file to set the value of `kflex-initial-ctx-name` to the renamed context name.

2. Add a new subcommand to kflex to allow renaming context so that it automatically does the update of `kflex-initial-ctx-name`. 

wdyt ?

MikeSpreitzer:
I do not like (1) because it is very tedious and very fragile. I do not like (2) because it is fragile. I think that there may be more possibilities. Could we make `kflex ctx` search the kubeconfig and find the right context by looking at attributes of all the contexts? Is there an extensions field in a kubeconfig "cluster"?

pdettori:
[...]

PR #2157, opened 2024-05-14
✨ KubeStellar Core chart

francostellari:
## Summary

This PR proposes a first version of the KubeStellar Core chart that allows one to:
1. install KubeFlex and its requirements
2. install PCHs for ITSes and WDSes
3. install a  user defined list of ITSes and WDSes

This PR attempts to provide an update to the goreleaser to automatically generate the chart upon a new release.

Please see this initial documentation: https://github.com/francostellari/kubestellar/blob/hub-chart/docs/content/direct/hub-chart.md

Not included in this PR:
1. update of the example.md to reflect the availability of this chart (I think this should be done after the first version of the chart is actually released, without advertising, and tested... a chart must be released to be actually used)
2. the new documentation is not connected to the overall documentation

## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -8,11 +8,12 @@ on:
 permissions:
   contents: write
   packages: write
- 
+
 env:
   REGISTRY: ghcr.io
   CONTROLLER_MANAGER_IMAGE: kubestellar/kubestellar/controller-manager
   CHART_PATH: ./chart
+  CORE_CHART_PATH: ./core-chart
 
 jobs:
   goreleaser:
@@ -48,18 +49,21 @@ jobs:
       uses: azure/setup-helm@v4
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    
+
     - name: Login to registry
       uses: docker/login-action@v3
       with:
         registry: ${{ env.REGISTRY }}
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
-      
-    - name: Package and push chart
+
+    - name: Package and push charts
       run: |
         chartVersion=$(echo ${{ github.ref_name }} | cut -c 2-)
         make chart IMG=${{ env.REGISTRY }}/${{ env.CONTROLLER_MANAGER_IMAGE }}:${{github.ref_name}}
         helm package ${{ env.CHART_PATH }} --destination . --version ${chartVersion} --app-version ${chartVersion}
-        helm push ./*.tgz oci://${{ env.REGISTRY }}/kubestellar/kubestellar    
+        helm push ./controller-manager-chart*.tgz oci://${{ env.REGISTRY }}/kubestellar/kubestellar
+        sed -i "s/^KUBESTELLAR_VERSION:.*$/KUBESTELLAR_VERSION: \"${chartVersion}\"/g" ${{ env.CORE_CHART_PATH }}/values.yaml
+        helm package ${{ env.CORE_CHART_PATH }} --destination . --version ${chartVersion} --app-version ${chartVersion} --dependency-update
+        helm push ./core-chart*.tgz oci://${{ env.REGISTRY }}/kubestellar/kubestellar
 
--- a/.gitignore
+++ b/.gitignore
@@ -29,4 +29,6 @@ local-chart/
 # word counts
 /counts
 
-
+# Core Chart generated dependencies
+core-chart/Chart.lock
+core-chart/charts/
--- a/core-chart/templates/NOTES.txt
+++ b/core-chart/templates/NOTES.txt
@@ -1,0 +1,53 @@
+Wait for all Control Planes to be ready:
+
+echo "Waiting for all Control Planes to be ready..."
+for cpname in `kubectl get controlplane -o name`; do
+    cpname=${cpname##*/}
+    while [[ $(kubectl get cp $cpname -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do
+        echo "waiting for Control Plane $cpname..."
+        sleep 5
+    done
+    echo $cpname
+done
+
+Control Planes of type host can be accessed using the host cluster kubeconfig.
+
+Obtain the kubeconfig to access a Control Plane $cpname of another type with the command:
+
+echo "Getting the kubeconfig of Control Planes $cpname ==> ./kubeconfig-$cpname..."
+kubectl get secret $(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.name}') \
+    -n $(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.namespace}') \
+    -o=jsonpath="{.data.$(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.key}')}" \
+    | base64 -d > "./kubeconfig-$cpname"
+
+which will create a file "kubeconfig-$cpname" in the current folder.
+
+Obtain the kubeconfig to access all Control Plane with the command:
+
+echo "Getting the kubeconfig of all Control Planes..."
+for cpname in `kubectl get controlplane -o name`; do
+    cpname=${cpname##*/}
+    echo "Getting the kubeconfig of Control Planes $cpname ==> /tmp/kubeconfig-$cpname..."
+    if [[ "$(kubectl get controlplane $cpname -o=jsonpath='{.spec.type}')" == "host" ]] ; then
+        kubectl config view --minify --flatten > "/tmp/kubeconfig-$cpname"
+    else
+        kubectl get secret $(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.name}') \
+            -n $(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.namespace}') \
+            -o=jsonpath="{.data.$(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.key}')}" \
+            | base64 -d > "/tmp/kubeconfig-$cpname"
+    fi
+    kubectl --kubeconfig "kubeconfig-$cpname" config rename-context $(kubectl --kubeconfig "kubeconfig-$cpname" config current-context) $cpname 2> /dev/null
+done
+
+Afterwards access the content of a Control Plane $cpname can be accessed by specifing its kubeconfig:
+
+kubectl --kubeconfig "kubeconfig-$cpname" ...
+
+The individual kubeconfigs can also be merged as contexts of the current `~/.kube/config` with the following command:
+
+KUBECONFIG=~/.kube/config:$(find . -maxdepth 1 -type f -name '/tmp/kubeconfig-*' | tr '\n' ':') kubectl config view --flatten > /temp/kubeconfig-merged
+cp /temp/kubeconfig-merged ~/.kube/config
+
+Afterwards the content of a Control Plane $cpname can be accessed by specifing its contexts:
+
+kubectl --context "$cpname" ...
--- a/core-chart/templates/kubeflex-clusterrolebinding.yaml
+++ b/core-chart/templates/kubeflex-clusterrolebinding.yaml
@@ -1,0 +1,18 @@
+{{- range $cp := (concat (.Values.ITSes | default list) (.Values.WDSes | default list) ) }}
+{{- if eq $cp.type "host" }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: kubeflex-manager-cluster-admin-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+- kind: ServiceAccount
+  name: kubeflex-controller-manager
+  namespace: kubeflex-system
+{{- break }}
+{{- end }}
+{{- end }}
--- a/core-chart/values.yaml
+++ b/core-chart/values.yaml
@@ -1,0 +1,61 @@
+# Default values for KubeStellar-Core.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+
+# Container/chart image versions. These values are set during the chart release process,
+# please do not change them unless you know what your are doing.
+HELM_VERSION: "3.15.0"
+KUBECTL_VERSION: "1.30.1"
+KUBESTELLAR_VERSION: "0.23.0-alpha.2"
+CLUSTERADM_VERSION: "0.8.2"
+OCM_STATUS_ADDON_VERSION: "0.2.0-rc8"
+OCM_TRANSPORT_PLUGIN_VERSION: "0.1.8"
+
+
+# Control controller log verbosity
+# The "default" verbosity value will be used for all controllers unless a specific controller verbosity override is specified
+verbosity:
+  default: 2
+  # Specific controller verbosity overrides:
+  # kubestellar: 6
+  # clusteradm: 6
+  # status_controller: 6
+  # status_agent: 6
+  # transport: 6
+
+
+# KubeFlex override values
+kubeflex-operator:
+  install: true
+  # installPostgreSQL: true # automatically install the appropriate version of PostgreSQL required by KubeFlex
+  # isOpenShift: false # set this variable to true when installing the chart in an OpenShift cluster
+  # domain: localtest.me # used to define the DNS domain name used from outside the KubeFlex hosting cluster to reach that cluster's Ingress endpoint
+  # externalPort: 9443 # used to set the port to access the Control Planes API
+  # hostContainer: kubeflex-control-plane # used to set the name of the container that runs the KubeFlex hosting cluster
+  # verbosity: 0 $ used to set the verbosity of KubeFlex controller (this is not impacted by the global VERBOSITY value)
+
+
+# Determine if the Post Create Hooks should be installed by the chart
+InstallPCHs: true
+
+
+# List the Inventory and Transport Spaces (ITSes) to be created by the chart
+# Each ITS consists of a mandatory unique name and an optional type, which could be either host or vcluster (default to vcluster, if not specified)
+ITSes: # all the CPs in this list will execute the its.yaml PCH
+  # - name: its1 # name of the control plane
+  #   type: host # optional type of control plane host or vcluster (default to vcluster, if not specified)
+  # - name: its2 # name of the control plane
+  #   type: host # optional type of control plane host or vcluster (default to vcluster, if not specified)
+
+
+# List the Workload Description Spaces (WDSes) to be created by the chart
+# Each WDS consists of a mandatory unique name and several optional parameters:
+# - type: host or k8s (default to k8s, if not specified)
+# - APIGroups: string holding a comma-separated list of APIGroups
+# - ITSName: the name of the ITS control plane to be used by the WDS. Note that the ITSName MUST be specified if more than one ITS exists at the moment the WDS PCH starts.
+WDSes: # all the CPs in this list will execute the wds.yaml PCH
+  # - name: wds1 # name of the control plane
+  #   type: host # optional type of control plane host or k8s (default to k8s, if not specified)
+  #   APIGroups: "" # optional string holding a comma-separated list of APIGroups
+  #   ITSName: its1 # optional name of the ITS control plane, this MUST be specified if more than one ITS exists at the moment the WDS PCH starts
--- a/docs/content/direct/common-setup-core-chart.md
+++ b/docs/content/direct/common-setup-core-chart.md
@@ -1,0 +1,83 @@
+# Common Setup - KubeStellar Core chart
+
+For more details please see [here](./core-chart.md).
+
+This documents explains how to use KubeStellar Core chart to deploy a new instance of KubeStellar
+with a choice of user-defined KubeFlex Control Planes (CPs).
+
+The information provided is specific for the following release:
+
+```shell
+export KUBESTELLAR_VERSION=0.23.0-alpha.2
+```
+
+It may also be a good idea to do a bit of cleanup first. See how it is done in the cleanup script for our E2E tests (in `test/e2e/common/cleanup.sh`).
+
+## Create a Kind cluster
+
+The setup of KubStellar via the Core chart requires the existance of at least one cluster
+to be used for the deployment of the chart.
+
+For convenience, a new local Kind cluster that satisfies the requirements for KubeStellar setup
+and that can be used to exercises the [examples](./examples.md) can be created with the following command:
+
+```shell
+sudo su <<EOF
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${KUBESTELLAR_VERSION}/scripts/create-kind-cluster-with-SSL-passthrough.sh) --name kubeflex --port 9443
+EOF
+```
+
+## Install KubeStellar Core chart
+
+A KubeStellar Core installation compatible with the common setup suitable for Common Setup described in the [examples](https://docs.kubestellar.io/release-0.22.0/direct/examples/) could be achieved with the following command:
+
+```shell
+helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart --version $KUBESTELLAR_VERSION \
+  --set-json='ITSes=[{"name":"its1"}]' \
+  --set-json='WDSes=[{"name":"wds1"}]'
+```
+
+Remember to include the option `--set "kubeflex-operator.isOpenShift=true"`, when deploying in an OpenShift cluster.
+
+## Obtain the kubeconfigs to access the KubeFlex Control Planes
+
+After the KubeStellar Core has fully deployed, before changing the kubeconfig context, `kflex` CLI can be used to retrieve the kubeconfig of all the KubeFlex Control Planes and store them as contexts of the current kubeconfig:
+
+```shell
+kflex ctx its1
+kflex ctx wds1
+kflex ctx # switch back to the initial context
+```
+
+Afterwards the content of a Control Plane `<cpname>` can be accessed by specifing its context:
+
+```shell
+kubectl --context <cpname> ...
+```
+
+## (optional) Check relevant deployments and statefulsets running in the hosting cluster
+
+Expect to see the `kubestellar-controller-manager` in the `wds1-system` namespace and the
+statefulset `vcluster` in the `its1-system` namespace, both fully ready.
+
+```shell
+kubectl get deployments,statefulsets --all-namespaces
+```
+
+The output should look something like the following:
+
+```shell
+NAMESPACE            NAME                                             READY   UP-TO-DATE   AVAILABLE   AGE
+ingress-nginx        deployment.apps/ingress-nginx-controller         1/1     1            1           22h
+kube-system          deployment.apps/coredns                          2/2     2            2           22h
+kubeflex-system      deployment.apps/kubeflex-controller-manager      1/1     1            1           22h
+local-path-storage   deployment.apps/local-path-provisioner           1/1     1            1           22h
+wds1-system          deployment.apps/kube-apiserver                   1/1     1            1           22m
+wds1-system          deployment.apps/kube-controller-manager          1/1     1            1           22m
+wds1-system          deployment.apps/kubestellar-controller-manager   1/1     1            1           21m
+wds1-system          deployment.apps/transport-controller             1/1     1            1           21m
+
+NAMESPACE         NAME                                   READY   AGE
+its1-system       statefulset.apps/vcluster              1/1     11h
+kubeflex-system   statefulset.apps/postgres-postgresql   1/1     22h
+```
--- a/docs/content/direct/common-setup-step-by-step.md
+++ b/docs/content/direct/common-setup-step-by-step.md
@@ -1,0 +1,71 @@
+# Common Setup - step-by-step
+
+The following steps create a deployment of KubeStellar Core
+
+1. You may want to `set -e` in your shell so that any failures in the setup or usage scenarios are not lost.
+
+1. If you ran through these scenarios previously then you will need to do a bit of cleanup first. See how it is done in the cleanup script for our E2E tests (in `test/e2e/common/cleanup.sh`).
+
+1. Set environment variables to hold KubeStellar and OCM-status-addon desired versions:
+
+    ```shell
+    export KUBESTELLAR_VERSION=0.23.0-alpha.2
+    ```
+
+1. Create a Kind hosting cluster with nginx ingress controller and KubeFlex controller-manager installed:
+
+    ```shell
+    kflex init --create-kind
+    ```
+   If you are installing KubeStellar on an existing Kubernetes or OpenShift cluster, just use the command `kflex init`.
+
+1. Update the post-create-hooks in KubeFlex to install kubestellar with the desired images:
+
+    ```shell
+    kubectl apply -f https://raw.githubusercontent.com/kubestellar/kubestellar/v${KUBESTELLAR_VERSION}/config/postcreate-hooks/kubestellar.yaml
+    kubectl apply -f https://raw.githubusercontent.com/kubestellar/kubestellar/v${KUBESTELLAR_VERSION}/config/postcreate-hooks/ocm.yaml
+    ```
+
+1. Create an inventory & mailbox space of type `vcluster` running *OCM* (Open Cluster Management)
+in KubeFlex. Note that `-p ocm` runs a post-create hook on the *vcluster* control plane
+which installs OCM on it.
+This step includes the installation of the status add-on controller.
+See [here](./architecture.md#ocm-status-add-on-agent) for more details on the add-on.
+
+    ```shell
+    kflex create its1 --type vcluster -p ocm
+    ```
+
+1. Create a Workload Description Space `wds1` in KubeFlex. Similarly to before, `-p kubestellar`
+runs a post-create hook on the *k8s* control plane that starts an instance of a KubeStellar controller
+manager which connects to the `wds1` front-end and the `its1` OCM control plane back-end.
+Additionally the OCM based transport controller will be deployed.
+
+    ```shell
+    kflex create wds1 -p kubestellar
+    ```
+
+1. (optional) Check relevant deployments and statefulsets running in the hosting cluster. Expect to
+see the `kubestellar-controller-manager` in the `wds1-system` namespace and the
+statefulset `vcluster` in the `its1-system` namespace, both fully ready.
+
+    ```shell
+    kubectl --context kind-kubeflex get deployments,statefulsets --all-namespaces
+    ```
+   The output should look something like the following:
+
+    ```shell
+    NAMESPACE            NAME                                             READY   UP-TO-DATE   AVAILABLE   AGE
+    ingress-nginx        deployment.apps/ingress-nginx-controller         1/1     1            1           22h
+    kube-system          deployment.apps/coredns                          2/2     2            2           22h
+    kubeflex-system      deployment.apps/kubeflex-controller-manager      1/1     1            1           22h
+    local-path-storage   deployment.apps/local-path-provisioner           1/1     1            1           22h
+    wds1-system          deployment.apps/kube-apiserver                   1/1     1            1           22m
+    wds1-system          deployment.apps/kube-controller-manager          1/1     1            1           22m
+    wds1-system          deployment.apps/kubestellar-controller-manager   1/1     1            1           21m
+    wds1-system          deployment.apps/transport-controller             1/1     1            1           21m
+
+    NAMESPACE         NAME                                   READY   AGE
+    its1-system       statefulset.apps/vcluster              1/1     11h
+    kubeflex-system   statefulset.apps/postgres-postgresql   1/1     22h
+    ```
--- a/docs/content/direct/core-chart.md
+++ b/docs/content/direct/core-chart.md
@@ -1,0 +1,279 @@
+# KubeStellar Core chart usage
+
+This documents explains how to use KubeStellar Core chart to deploy a new instance of KubeStellar
+with a choice of user-defined KubeFlex Control Planes (CPs).
+
+The information provided is specific for the following release:
+
+```shell
+export KUBESTELLAR_VERSION=0.23.0-alpha.2
+```
+
+## Pre-requisites
+
+To install the Helm chart the only requirement is [Helm](https://helm.sh/).
+However, additional executables may be required to create/manage the cluster(s) (_e.g._, Kind and kubectl),
+to join Workload Execution Clusters (WEKs) (_e.g._, clusteradm),
+and to interact with Control Planes (_e.g._, kubectl), _etc_.
+For such purpose, a full list of executable that may be required can be fund [here](./pre-reqs.md).
+
+The setup of KubStellar via the Core chart requires the existence of a KubeFlex hosting cluster.
+
+This can be:
+
+1. A local **Kind** or **k3s** cluster with an ingress with SSL passthrough and a mapping to host port 9443
+
+This option is particularly useful for first time users or users that would like to have a local deployment.
+
+It is important to note that, when the hosting cluster was created by **kind** or **k3s** and its Ingress domain name is left to default to localtest.me, then the name of the container running hosting cluster must be also be referenced during the Helm chart installation by setting `--set "kubeflex-controller.hostContainer=<control-plane-container-name>"`.
+The `<control-plane-container-name>` is the name of the container in which kind or k3d is running the relevant control plane. One may use `docker ps` to find the `<control-plane-container-name>`.
+
+If a host port number different from the expected 9443 is used for the Kind cluster, then the same port number must be specified during the chart installation by adding the following argument `--set "kubeflex-controller.externalPort=<port>"`.
+
+By default the KubeStelalr Core chart uses a test domain `localtest.me`, which is ok for testing on a single host machine. However, scenarios that span more than one machine, it is necessary to set `--set "kubeflex-controller.domain=<domain>"` to a more appropriate `<domain>` that can be reached from Workload Execution CLusters (WECs).
+
+For convenience, a new local **Kind** cluster that satisfies the requirements for KubeStellar setup
+and that can be used to exercises the [examples](./examples.md) can be created with the following command:
+
+```shell
+sudo su <<EOF
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v23.0/scripts/create-kind-cluster-with-SSL-passthrough.sh) --name kubeflex --port 9443
+EOF
+```
+
+Alternatively, a new local **k3s** cluster that satisfies the requirements for KubeStellar setup
+and that can be used to exercises the [examples](./examples.md) can be created with the following command:
+
+```shell
+sudo su <<EOF
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v23.0/scripts/create-k3s-cluster-with-SSL-passthrough.sh) --port 9443
+EOF
+```
+
+3. An **OpenShift** cluster
+
+When using this option, one is required to explicitely set the `isOpenShift` variable to `true` by including `--set "kubeflex-operator.isOpenShift=true"` in the Helm chart installation command.
+
+## KubeStellar Core Chart values
+
+The KubeStellar chart makes available to the user several values that may be used to customize its installation into an existing cluster:
+
+```yaml
+# KubeFlex override values
+kubeflex-operator:
+  install: true # enable/disable the installation of KubeFlex by the chart (default: true)
+  installPostgreSQL: true # enable/disable the installation of the appropriate version of PostgreSQL required by KubeFlex (default: true)
+  isOpenShift: false # set this variable to true when installing the chart in an OpenShift cluster (default: false)
+  # Kind cluster specific settings:
+  domain: localtest.me # used to define the DNS domain name used from outside the KubeFlex hosting cluster to reach that cluster's Ingress endpoint (default: localtest.me)
+  externalPort: 9443 # used to set the port to access the Control Planes API (default: 9443)
+  hostContainer: kubeflex-control-plane # used to set the name of the container that runs the KubeFlex hosting cluster (default: kubeflex-control-plane, which corresponds to a Kind cluster with name kubeflex)
+
+# Determine if the Post Create Hooks should be installed by the chart
+InstallPCHs: true
+
+# List the Inventory and Transport Spaces (ITSes) to be created by the chart
+# Each ITS consists of a mandatory unique name and an optional type, which could be either host or vcluster (default to vcluster, if not specified)
+ITSes: # ==> installs ocm + ocm-status-addon
+
+# List the Workload Description Spaces (WDSes) to be created by the chart
+# Each WDS consists of a mandatory unique name and several optional parameters:
+# - type: host or k8s (default to k8s, if not specified)
+# - APIGroups: a comma separated list of APIGroups
+# - ITSName: the name of the ITS control plane to be used by the WDS. Note that the ITSName MUST be specified if more than one ITS exists.
+WDSes: # ==> installs kubestellar + ocm-transport-plugin
+```
+
+The first section of the `values.yaml` file refers to parameters that are specific to the KubeFlex instllation, see [here](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md) for more information.
+
+In particular:
+- `kubeflex-operator.install` accepts a boolean value to enable/disable the installation of KubeFlex into the cluster by the chart
+- `kubeflex-operator.isOpenShift` must be set to true by the user when installing the chart into a OpenShift cluster
+
+By default, the chart will install the KubeFlex and its PostgreSQL dependency.
+
+The second section allows a user of the chart to determine if Post Create Hooks (PCHs) needed for creating ITSes and WDSes control planes should be instlalled by the chart. By default `InstallPCHs` is set to `true` to enable the instllation of the PCHs, however one may want to set this value to `false` when installing multiple copies of the chart to avoid conflicts. A single copy of the PCHs is required and allowed per cluster.
+
+The third section of the `values.yaml` file allows one to create a list of Inventory and Transport Spaces (ITSes). By default, this list is empty and no ITS will be created by the chart. A list of ITSes can be specified using the following format:
+
+```yaml
+ITSes: # all the CPs in this list will execute the its.yaml PCH
+  - name: <its1>          # mandatory name of the control plane
+    type: <vlcuster|host> # optional type of control plane host or vcluster (default to vcluster, if not specified)
+  - name: <its2>          # mandatory name of the control plane
+    type: <vlcuster|host> # optional type of control plane host or vcluster (default to vcluster, if not specified)
+  ...
+```
+
+where `name` must specify a name unique among all the control planes in that KubeFlex deployment and the optional `type` can be either vlcuster (default) or host, see [here](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md) for more information.
+
+The fourth section of the `values.yaml` file allows one to create a list of Workload Description Spaces (WDSes). By default, this list is empty and no WDS will be created by the chart. A list of WDSes can be specified using the following format:
+
+```yaml
+WDSes: # all the CPs in this list will execute the wds.yaml PCH
+  - name: <wds1>     # mandatory name of the control plane
+    type: <host|k8s> # optional type of control plane host or k8s (default to k8s, if not specified)
+    APIGroups: ""    # optional string holding a comma-separated list of APIGroups
+    ITSName: <its1>  # optional name of the ITS control plane, this MUST be specified if more than one ITS exists at the moment the WDS PCH starts
+  - name: <wds2>     # mandatory name of the control plane
+    type: <host|k8s> # optional type of control plane host or k8s (default to k8s, if not specified)
+    APIGroups: ""    # optional string holding a comma-separated list of APIGroups
+    ITSName: <its2>  # optional name of the ITS control plane, this MUST be specified if more than one ITS exists at the moment the WDS PCH starts
+  ...
+```
+
+where `name` must specify a name unique among all the control planes in that KubeFlex deployment (note that this must be unique among both ITSes and WDSes), the optional `type` can be either k8s (default) or host, see [here](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md) for more information, the optional `APIGroups` provides a list of APIGroups, see [here](https://docs.kubestellar.io/release-0.22.0/direct/examples/#scenario-2-using-the-hosting-cluster-as-wds-to-deploy-a-custom-resource) for more information, and `ITSName` specify the ITS connected to the new WDS being created (this parameter MUST be specified if more that one ITS exists in the cluster, if no value is specified and only one ITS exists in the cluster, then it will be automatically selected).
+
+## KubeStellar Core Chart usage
+
+A specific version of the KubeStellar core chart can be simply installed in an existing cluster using the following command:
+
+```shell
+helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart --version $KUBESTELLAR_VERSION
+```
+
+The above command will install KubeFlex and the Post Create Hooks, but no Control Planes.
+Please remeber to add `--set "kubeflex-operator.isOpenShift=true"`, when installing into an OpenShift cluster.
+
+User defined control planes can be added using additional value files of `--set` arguments, _e.g._:
+
+- add a single ITS named its1 of default vcluster type: `--set-json='ITSes=[{"name":"its1"}]'`
+- add two ITSes named its1 and its2 of of type vlcuster and host, respectevely: `--set-json='ITSes=[{"name":"its1"},{"name":"its2","type":"host"}]'`
+- add a single WDS named wds1 of default k8s type connected to the one and only ITS: `--set-json='WDSes=[{"name":"wds1"}]'`
+
+A KubeStellar Core installation compatible with the common setup suitable for Common Setup described in the [examples](./examples.md) could be achieved with the following command:
+
+```shell
+helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart --version $KUBESTELLAR_VERSION \
+  --set-json='ITSes=[{"name":"its1"}]' \
+  --set-json='WDSes=[{"name":"wds1"}]'
+```
+
+After the initial instllation is completed, there are two main ways to install additional control planes (_e.g._, create a second `wds2` WDS):
+
+1. Upgrade the initial chart. This choice requires to relist the existing control planes, which would otherwise be deleted:
+
+```shell
+helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart --version $KUBESTELLAR_VERSION \
+  --set-json='ITSes=[{"name":"its1"}]' \
+  --set-json='WDSes=[{"name":"wds1"},{"name":"wds2"}]'
+```
+
+2. Install a new chart with a different name. This choice does not requires to relist the existing control planes, but requies to disable the reinstallation of KubeFlex and PCHs:
+
+```shell
+helm upgrade --install add-wds2 oci://ghcr.io/kubestellar/kubestellar/core-chart --version $KUBESTELLAR_VERSION \
+  --set='kubeflex-operator.install=false,InstallPCHs=false' \
+  --set-json='WDSes=[{name":"wds2"}]'
+```
+
+## Obtaining the Control Planes kubeconfigs
+
+The following code can be used to wait for the Ready state of all the KubeFlex Control Planes in the hosting cluster:
+
+```shell
+echo "Waiting for all Control Planes to be ready..."
+for cpname in `kubectl get controlplane -o name`; do
+  cpname=${cpname##*/}
+  while [[ $(kubectl get cp $cpname -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do
+    echo "Waiting for Control Plane $cpname..."
+    sleep 5
+  done
+  echo $cpname
+done
+```
+
+The kubeconfig of the KubeFlex control planes can be obtained both with and without the use of `kflex` CLI.
+
+1. using `kflex` CLI
+
+The kubeconfig of a `$cpname` Control Plane can be added as a context of the existing kubeconfig by using the command:
+
+```shell
+kflex ctx $cpname
+```
+
+Note that the command above will also switch the current context to `$cpname`, use `kflex ctx` to switch back to the initial context.
+
+To automatically add all Control Planes as contexts of the current kubeconfig, one can use the convenience script below:
+
+```shell
+echo "Getting the kubeconfig of all Control Planes..."
+for cpname in `kubectl get controlplane -o name`; do
+  cpname=${cpname##*/}
+  echo "Getting the kubeconfig of Control Planes \"$cpname\"..."
+  kflex ctx $cpname
+done
+kflex ctx # switch back to the initial context
+```
+
+Afterwards the content of a Control Plane `$cpname` can be accessed by specifing its context:
+
+```shell
+kubectl --context "$cpname" ...
+```
+
+2. using plain `kubectl` commands
+
+The following code can be used to obtain the kubeconfig of all the KubeFlex Control Planes in the hosting cluster:
+
+```shell
+echo "Getting the kubeconfig of all Control Planes..."
+for cpname in `kubectl get controlplane -o name`; do
+  cpname=${cpname##*/}
+  echo "Getting the kubeconfig of Control Planes $cpname ==> kubeconfig-$cpname..."
+  if [[ "$(kubectl get controlplane $cpname -o=jsonpath='{.spec.type}')" == "host" ]] ; then
+    kubectl config view --minify --flatten > "kubeconfig-$cpname"
+  else
+    kubectl get secret $(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.name}') \
+      -n $(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.namespace}') \
+      -o=jsonpath="{.data.$(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.key}')}" \
+      | base64 -d > "kubeconfig-$cpname"
+  fi
+  kubectl --kubeconfig "kubeconfig-$cpname" config rename-context $(kubectl --kubeconfig "kubeconfig-$cpname" config current-context) $cpname 2> /dev/null
+done
+```
+
+The code above will saves the kubeconfig of a control plane `$cpname` to a corresponding file `kubeconfig-$cpname` in the local folder.
+The current context will be renamed to `$cpname`.
+The content of a Control Plane `$cpname` can be accessed by specifing its kubeconfig:
+
+```shell
+kubectl --kubeconfig "kubeconfig-$cpname" ...
+```
+
+The individual kubeconfigs can also be merged as contexts of the current `~/.kube/config` with the following command:
+
+```shell
+cp ~/.kube/config ~/.kube/config.bak
+KUBECONFIG=~/.kube/config:$(find . -maxdepth 1 -type f -name 'kubeconfig-*' | tr '\n' ':') kubectl config view --flatten > ~/.kube/kubeconfig-merged
+mv ~/.kube/kubeconfig-merged ~/.kube/config
+```
+
+Afterwards the content of a Control Plane `$cpname` can be accessed by specifing its context:
+
+```shell
+kubectl --context "$cpname" ...
+```
+
+## Uninstalling the KubeStellar Core chart
+
+The chart can be uninsitalled using the command:
+
+```shell
+helm uninstall ks-core
+```
+
+This will remove KubeFlex, PostgreSQL, Post Create Hooks (PCHs), and all KubeFlex Control Planes (_i.e._, ITSes and WDSes) that were created by the chart.
+
+Additionally, if a **Kind** cluster was created with the provide script, it can be deleted with the command:
+
+```shell
+kind delete cluster --name kubeflex
+```
+
+Alternatively, if a **k3s** cluster was created with the provide script, it can be deleted with the command:
+
+```shell
+/usr/local/bin/k3s-uninstall.sh
+```
--- a/scripts/create-k3s-cluster-with-SSL-passthrough.sh
+++ b/scripts/create-k3s-cluster-with-SSL-passthrough.sh
@@ -1,0 +1,100 @@
+#!/usr/bin/env bash
+# Copyright 2024 The KubeStellar Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Create a k3s cluster for KubeStellar deployment
+
+port=9443
+wait=true
+
+display_help() {
+  cat << EOF
+Usage: $0 [--port port] [--nowait][-X]
+
+--port port     map the specified host port to the kind cluster port 443 (default: 9443)
+--nowait        when set to true, the script proceeds without waiting for the nginx ingress patching to complete
+-X              enable verbose execution of the script for debugging
+EOF
+}
+
+while (( $# > 0 )); do
+    case "$1" in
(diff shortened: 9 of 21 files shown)


docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
+# KubeStellar Hub chart usage
+
+This documents explains how to use KubeStellar Hub chart to deploy a new instance of KubeStellar
+with a choise of user-defined Control Plains.
```
MikeSpreitzer, line 4: s/choise/choice/

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
+# KubeStellar Hub chart usage
+
+This documents explains how to use KubeStellar Hub chart to deploy a new instance of KubeStellar
+with a choise of user-defined Control Plains.
+
+The information provided is specific for the following release:
+
+```shell
+export KUBESTELLAR_VERSION=0.23.0-alpha.1
```
MikeSpreitzer, line 9: You're on alpha.2 now.

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+The information provided is specific for the following release:
+
+```shell
+export KUBESTELLAR_VERSION=0.23.0-alpha.1
+```
+
+## Pre-requisites
+
+A list of executable that may be required during the deployment can be fund [here](./pre-reqs.md).
+At minimum [Helm](https://helm.sh/) is required.
```
MikeSpreitzer, line 15: I do not understand what this remark about "at minimum" is trying to say. Aren't all the referenced requirements required?

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+1. A local **Kind** cluster with an ingress with SSL passthrough and a mapping to host port 9443
+
+This option is particularly useful for first time users or users that would like to have a local deployment.
+
+For convenience, a new local Kind clusters that satisfies the requirements for KubeStellar setup
+and that can be used to exercises the [examples](./examples.md) can be created with the following command:
+
+```shell
+sudo su <<EOF
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${KUBESTELLAR_VERSION}/scripts/create-kind-cluster-with-SSL-passthrough.sh) --name kubestellar --port 9443
```
MikeSpreitzer, line 31: If the port has to be 9443, and the script defaults to that value, then why repeat it here?

There was nothing said earlier about the `kind` cluster name. I suspect it has to be `kubestellar`, since I see no variable holding the cluster name. If this is right, then same question as for the port number.

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+EOF
+```
+
+It is important to note that, if a different name is used for the Kind cluster, then its control plane name must be also be referenced during the Helm chart installation by setting `--set "kubeflex-controller.hostContainer=<control-plane-name>".
```
MikeSpreitzer, line 35: This remark is true regardless of whether the `kind` cluster is created using the convenience script shown above or another way, right? If so then this remark should come before the discussion of the convenience script.

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+
+2. A local **k3s** cluster with an ingress with SSL passthrough and a mapping to host port 9443
+
+... Under onstruction ...
+
```
MikeSpreitzer, line 40: In the non-OpenShift cases, it is important that `kubeflex init` has a `--domain` option, which is critical to making scenarios that span more than one machine work.

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+The KubeStellar chart makes available to the user several values that may be used to customize its installation into an existing cluster:
+
+```yaml
+# KubeFlex override values
+kubeflex-operator:
+  install: true # enable/disable the installation of KubeFlex by the chart (default: true)
+  installPostgreSQL: true # enable/disable the installation of the appropriate version of PostgreSQL required by KubeFlex (default: true)
+  isOpenShift: false # set this variable to true when installing the chart in an OpenShift cluster (default: false)
+  # Kind cluster specific settings:
+  domain: localtest.me # used to set the Control Planes DNS in a Kind cluster installation (default: localtest.me)
```
MikeSpreitzer, line 57: This is not just for the case of `kind`, it is for everything that is not OpenShift.

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+  externalPort: 9443 # used to set the port to access the Control Planes API (default: 944)
```
MikeSpreitzer, line 58: Why is "(default: 944)" in the comment on this line?

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+  hostContainer: kubeflex-control-plane # used to set the name of cluster control plane (default: kubeflex-control-plane, which corresponds to a Kind cluster with name kubeflex)
```
MikeSpreitzer, line 59: I do not understand some of the words here. What is "cluster control plane" trying to say? I am really lost trying to understand the parenthetic remark about default.

I am not sure whether this is relevant, but: when I do `kind create cluster --name kubeflex`, this creates a _context_ in my current kubeconfig file and this context is named "kind-kubeflex".

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+
+# Determine if the Post Create Hooks should be installed by the chart
+InstallPCHs: true
+
+# List the Inventory and Transport Spaces (ITSes) to be created by the chart
+# Each ITS consists of a mandatory unique name and an optional type, which could be either host|vcluster (default to vcluster, if not specified)
```
MikeSpreitzer, line 65: Since this is English text, I would replace 'host|vcluster' with '"host" or "vcluster"'.

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+ITSes: # ==> installs ocm + ocm-status-addon
```
MikeSpreitzer, line 66: Is it even valid YAML to follow `foo:` with nothing? I think I have gotten blowback from this before. If you mean that the default is an empty JSON array then I would explicitly write `[]` here.

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+
+# List the Workload Description Spaces (WDSes) to be created by the chart
+# Each WDS consists of a mandatory unique name and several optional parameters:
+# - type: host|k8s (default to k8s, if not specified)
+# - APIGroups: a comma separated list of APIGroups
+# - ITSName: the name of the ITS control plane to be used by the WDS. Note that the ITSName MUST be specified if more than one ITS exists.
+WDSes: # ==> installs kubestellar + ocm-transport-plugin
```
MikeSpreitzer, line 73: Similar comments to the ITS ones apply here.

"kubestellar" is encompasses more than you mean here; I think you mean "the kubestellar controller-manager".

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+The setup of KubStellar via the Hub chart requires the existance of at least one cluster
+to be used for the deployment of the chart.
+
+This can be:
+
+1. A local **Kind** cluster with an ingress with SSL passthrough and a mapping to host port 9443
+
+This option is particularly useful for first time users or users that would like to have a local deployment.
+
+For convenience, a new local Kind clusters that satisfies the requirements for KubeStellar setup
```
MikeSpreitzer, line 26: s/clusters/cluster/

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+
+The first section of the `values.yaml` file refers to parameters that are specific to the KubeFlex instllation, see [here](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md) for more information.
+
+In particular:
+- `kubeflex-operator.install` accepts a boolean value to enable/disable the installation of KubeFlex into the cluster by the chart
+- `kubeflex-operator.isOpenShift` must be set to true by the user when installing the chart into a OpenShift cluster
+
+By default, the chart will install the KubeFlex and its PostgreSQL dependency.
+
+The second section allows a user of the chart to determine if Post Create Hooks (PCHs) needed for creating ITSes and WDSes control planes should be instlalled by the chart. By default `InstallPCHs` is set to `true` to enable the instllation of the PCHs, however one mya want to set this value to `false` when installing multiple copies of the chart to avoid conflicts. A single copy of the PCHs is required and allowed per cluster.
```
MikeSpreitzer, line 84: s/mya/may/

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+
```
MikeSpreitzer, line 85: s/no control plane/no ITS control plane/

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+  - name: <its1>          # mandatory name of the control plane
+    type: <vlcuster|host> # optional type of control plane host|vcluster (default to vcluster, if not specified)
+  - name: <its2>          # mandatory name of the control plane
+    type: <vlcuster|host> # optional type of control plane host|vcluster (default to vcluster, if not specified)
+  ...
+```
+
+where `name` must specify a unique name of the control plane and the optional `type` can be either vlcuster (default) or host, see [here](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md) for more information.
+
+The fourth section of the `values.yaml` file allows one to create a list of Workload Description Spaces (WDSes). By default, this list is empty and no control plane will be created by the chart. A list of WDSes can be specified using the following format:
```
MikeSpreitzer, line 99: s/no control plane/no WDS/

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+
+```yaml
+WDSes: # ==> installs kubestellar + ocm-transport-plugin
+  - name: <wds1>     # mandatory name of the control plane
+    type: <host|k8s> # optional type of control plane host|k8s (default to k8s, if not specified)
+    APIGroups: ""    # optional list of APIGroups
+    ITSName: <its1>  # optional name of the ITS control plane, this MUST be specified if more than one ITS exists
```
MikeSpreitzer, line 106: More than one ITS exists _when_? Since usage of this chart changes the number of ITSes, the reader needs to be told which point in time matters here.

docs/content/direct/hub-chart.md
francostellari: ok, thank you

docs/content/direct/hub-chart.md
francostellari: I will try to elaborate a bit. I guess what I was trying to say is that technically to just deploy the chart one only needs helm. Obviously to create a kind cluster one also need kind, to joing cluster one would need clusteradm.... but they are not strictly necessary to deploy the chart.

docs/content/direct/hub-chart.md
francostellari: 1. The intention was to "document" the options by exposing them to a user that may want to change them.
2. I rewarded the text below to explain that the port may be changed as well... and what needs to be done in that case
3. technically, there is bug in kflex that allows only `kubeflex`... until the next release of the chart... so I renamed back to `kubeflex`

docs/content/direct/hub-chart.md
francostellari: ok, good idea

docs/content/direct/hub-chart.md
francostellari: @MikeSpreitzer I'm not 100% sure I understand... I think you refer to the section before and that I should make a comment about setting the domain to something meaningful?

francostellari: I added a comment on that line

docs/content/direct/hub-chart.md
francostellari: typo

docs/content/direct/hub-chart.md
francostellari: It was not clear to me untill I stumbled into it.
When one does `kind create cluster --name foo`, the the cluster runs in a docker container called `foo-control-plane`
Then one must tell kflex to use that name `hostContainer: foo-control-plane`
In all our examples (expecially when `kflex init -c`) the kind cluster was called kubeflex... which is ok with all defaults.
I stambled into it because I thought that we may want to call the cluster kubestellar instead.

Please note that pending a new release of kflex, there is bug that actually prevent this.

docs/content/direct/hub-chart.md
francostellari: I changed it in the comments

docs/content/direct/hub-chart.md
francostellari: I can tell you that it's is valid, for Helm

The `==>` applies to all but I will change it to be more clear

OCM is removed

docs/content/direct/hub-chart.md
francostellari: yes, I reworded

docs/content/direct/hub-chart.md
francostellari: ok, thank you

docs/content/direct/hub-chart.md
francostellari: ok, thank you

docs/content/direct/hub-chart.md
francostellari: ok

docs/content/direct/hub-chart.md
francostellari: ok

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,157 @@
...
+
+where `name` must specify a unique name of the control plane and the optional `type` can be either vlcuster (default) or host, see [here](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md) for more information.
+
+The fourth section of the `values.yaml` file allows one to create a list of Workload Description Spaces (WDSes). By default, this list is empty and no control plane will be created by the chart. A list of WDSes can be specified using the following format:
+
+```yaml
+WDSes: # ==> installs kubestellar + ocm-transport-plugin
+  - name: <wds1>     # mandatory name of the control plane
+    type: <host|k8s> # optional type of control plane host|k8s (default to k8s, if not specified)
+    APIGroups: ""    # optional list of APIGroups
```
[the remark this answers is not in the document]
francostellari: ok

docs/content/direct/hub-chart.md
francostellari: "at the moment the WDS PCH is executed"

docs/content/direct/hub-chart.md
MikeSpreitzer: Note also that @dumb0002's testing plan involves multiple machines and NOT OpenShift, so he will need something that works in his configuration.

docs/content/direct/hub-chart.md
MikeSpreitzer: That is my point. When NOT using OCP, but using multiple machines, this `domain` parameter is important regardless of whether the clusters are created by `kind` or something else.

docs/content/direct/hub-chart.md
MikeSpreitzer: Why write the default value twice on one line?

docs/content/direct/hub-chart.md
MikeSpreitzer: Execution of a PCH takes place over time. Perhaps you mean when the PCH _starts_ ?

francostellari: ok

docs/content/direct/hub-chart.md
francostellari: I guess it can be removed, in reality in the values.yaml, the above is commented out and it relies on the kubeflex chart default values

docs/content/direct/hub-chart.md
francostellari: I reworked the text to talk a bit more about this parameter

.gitignore
```diff
@@ -29,4 +29,6 @@ local-chart/
 # word counts
 /counts
 
-
+# Hub Chart generated dependencies
+hub-chart/Chart.lock
```
MikeSpreitzer, line 33: What is `hub-chart/Chart.lock` ?

docs/content/direct/common-setup-hub-chart.md
```diff
@@ -0,0 +1,80 @@
...
+## Obtain the kubeconfigs to access the KubeFlex Control Planes
+
+After the KubeStellar Hub has fully deployed,
+the following code can be used to obtain the kubeconfig of all the KubeFlex Control Planes in the hub cluster:
+
+```shell
+echo "Getting the kubeconfig of all Control Planes..."
+for cpname in `kubectl get controlplane -o name`; do
+  cpname=${cpname##*/}
+  echo "Getting the kubeconfig of Control Planes $cpname ==> /tmp/kubeconfig-$cpname..."
```
MikeSpreitzer, line 49: Thankfully, s|/tmp/||
It would be less secure to put these credentials on /tmp.

docs/content/direct/common-setup-hub-chart.md
```diff
@@ -0,0 +1,80 @@
...
+Remember to include the option `--set "kubeflex-operator.isOpenShift=true"`, when deploying in an OpenShift cluster.
+
+## Obtain the kubeconfigs to access the KubeFlex Control Planes
+
+After the KubeStellar Hub has fully deployed,
+the following code can be used to obtain the kubeconfig of all the KubeFlex Control Planes in the hub cluster:
+
+```shell
+echo "Getting the kubeconfig of all Control Planes..."
+for cpname in `kubectl get controlplane -o name`; do
```
MikeSpreitzer, line 47: This makes a big new assumption: that the user does not have any irrelevant control planes, or that they are sufficiently few/inconsequential that this overspray is OK.

docs/content/direct/common-setup-hub-chart.md
```diff
@@ -0,0 +1,80 @@
...
+  echo "Getting the kubeconfig of Control Planes $cpname ==> /tmp/kubeconfig-$cpname..."
+  if [[ "$(kubectl get controlplane $cpname -o=jsonpath='{.spec.type}')" == "host" ]] ; then
+    kubectl config view --minify --flatten > "kubeconfig-$cpname"
+  else
+    kubectl get secret $(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.name}') \
+      -n $(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.namespace}') \
+      -o=jsonpath="{.data.$(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.key}')}" \
+      | base64 -d > "kubeconfig-$cpname"
+  fi
+  kubectl --kubeconfig "kubeconfig-$cpname" config rename-context $(kubectl --kubeconfig "kubeconfig-$cpname" config current-context) $cpname 2> /dev/null
```
MikeSpreitzer, line 58: Note to self: is this renaming necessary? The context is inaccessible unless the kubeconfig is specified; given that the kubeconfig will be specified in every command, why also specify the context?

docs/content/direct/common-setup-step-by-step.md
```diff
@@ -0,0 +1,71 @@
+# Common Setup - step-by-step
+
+The following steps create a deployment of KubeStellar Hub
+
+1. You may want to `set -e` in your shell so that any failures in the setup or usage scenarios are not lost.
+
+1. If you ran through these scenarios previously then you will need to do a bit of cleanup first. See how it is done in the cleanup script for our E2E tests (in `test/e2e/common/cleanup.sh`).
+
+1. Set environment variables to hold KubeStellar and OCM-status-addon desired versions:
```
MikeSpreitzer, line 9: s/and OCM-status-addon//

docs/content/direct/common-setup-step-by-step.md
```diff
@@ -0,0 +1,71 @@
...
+
+1. If you ran through these scenarios previously then you will need to do a bit of cleanup first. See how it is done in the cleanup script for our E2E tests (in `test/e2e/common/cleanup.sh`).
+
+1. Set environment variables to hold KubeStellar and OCM-status-addon desired versions:
+
+    ```shell
+    export KUBESTELLAR_VERSION=0.23.0-alpha.2
+    ```
+
+1. Create a Kind hosting cluster with nginx ingress controller and KubeFlex controller-manager installed:
```
MikeSpreitzer, line 15: s/Kind/`kind`/

docs/content/direct/common-setup-step-by-step.md
```diff
@@ -0,0 +1,71 @@
...
+
+1. You may want to `set -e` in your shell so that any failures in the setup or usage scenarios are not lost.
+
+1. If you ran through these scenarios previously then you will need to do a bit of cleanup first. See how it is done in the cleanup script for our E2E tests (in `test/e2e/common/cleanup.sh`).
+
+1. Set environment variables to hold KubeStellar and OCM-status-addon desired versions:
+
+    ```shell
+    export KUBESTELLAR_VERSION=0.23.0-alpha.2
+    ```
```
MikeSpreitzer, line 13: Don't these steps also belong in the alternative recipe using the hub chart?

Perhaps you could minimize how much this PR changes examples.md by leaving the current content in there but adding remarks like "The following steps will be replaced by [using the hub chart](common-setup-hub-chart.md)" before the to-be-replaced steps and following them with a remark like "Here end the steps to be replaced by [using the hub chart](common-setup-hub-chart.md)".

docs/content/direct/common-setup-hub-chart.md
```diff
@@ -0,0 +1,80 @@
...
+
+```shell
+helm upgrade --install ks-hub oci://ghcr.io/kubestellar/kubestellar/hub-chart --version $KUBESTELLAR_VERSION \
+  --set-json='ITSes=[{"name":"its1"}]' \
+  --set-json='WDSes=[{"name":"wds1"}]'
+```
+
+Remember to include the option `--set "kubeflex-operator.isOpenShift=true"`, when deploying in an OpenShift cluster.
+
+## Obtain the kubeconfigs to access the KubeFlex Control Planes
```
MikeSpreitzer, line 40: Dang, this section is long and fiddly. The addition of this complexity is required because this PR replaces the use of `kflex create` with creating ControlPlane API objects but `kflex create` did more than that (it also modified the current kubeconfig). We switched from `kflex create` with PCH use to Helm chart instantiation, in part because Helm charts are better regarding parameters (i.e., a chart declares all its parameters in a well-defined place and provides default value, while a PCH does neither of those things). We use `kflex` or a published Helm chart in order to avoid requiring the user to fetch KubeStellar scripts. If the user were fetching KubeStellar scripts then the complexity in this section could be nicely wrapped up in a script with a simple interface. And the deficiencies of PCHes wrt parameters would be less important because the user would use scripts instead of directlly using PCHes.

docs/content/direct/common-setup-step-by-step.md
```diff
@@ -0,0 +1,71 @@
...
+1. Create a Workload Description Space `wds1` in KubeFlex. Similarly to before, `-p kubestellar`
+runs a post-create hook on the *k8s* control plane that starts an instance of a KubeStellar controller
+manager which connects to the `wds1` front-end and the `its1` OCM control plane back-end.
+Additionally the OCM based transport controller will be deployed.
+
+    ```shell
+    kflex create wds1 -p kubestellar
+    ```
+
+1. (optional) Check relevant deployments and statefulsets running in the hosting cluster. Expect to
```
MikeSpreitzer, line 48: This step is also not specific to not using the hub chart.

docs/content/direct/common-setup-hub-chart.md
ezrasilvera: And?  What's the problem in that ?  It is clearly said it takes ALL the CPs defined here.  If users want to take this as an example but retrieve a subset of CPs they can do that themselves.

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,229 @@
...
+to join Workload Execution Clusters (WEKs) (_e.g._, clusteradm),
+and to interact with Control Planes (_e.g._, kubectl), _etc_.
+For such purpose, a full list of executable that may be required can be fund [here](./pre-reqs.md).
+
+The setup of KubStellar via the Hub chart requires the existance of at least one cluster
+to be used for the deployment of the chart.
+
+This can be:
+
+1. A local **Kind** cluster with an ingress with SSL passthrough and a mapping to host port 9443
```
MikeSpreitzer, line 25: A lot of the information in this section is not specific to `kind`. It is equally true for k3s, and (I suspect) any other sort of cluster that is not OpenShift. Also, a lot of this information is not specific to the hub chart.

docs/content/direct/common-setup-hub-chart.md
ezrasilvera: I originally thought that using PCH doesn't really gives you major advantage if you are going to automate your deployments through scripts in any case. If you use scripts you can go go ahead and install the exact thing PCH is doing but directly with more control over the process .

hub-chart/values.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+
+# KubeFlex override values
+kubeflex-operator:
+  install: true
+  # installPostgreSQL: true # automatically install the appropriate version of PostgreSQL required by KubeFlex
+  # isOpenShift: false # set this variable to true when installing the chart in an OpenShift cluster
+  # domain: localtest.me # used to set the Control Planes DNS in a Kind cluster installation
+  # externalPort: 9443 # used to set the port to access the Control Planes API
+  # hostContainer: kubeflex-control-plane # set the name of kubeflex control plane
```
MikeSpreitzer, line 34: If I understand correctly, this gets set to the name of a container rather than the KubeFlex name of a control plane. I do not understand the general story (see https://github.com/kubestellar/kubeflex/issues/250), but the wording here should be clarified. I suggest something like "the name of the container that runs the KubeFlex hosting cluster". Ideally this would include a pointer to the KubeFlex documentation for this, but there is none (as far as I can tell, @pdettori).

A little more precisely, this is an opportunity to override the default for a "value" of the kubeflex helm chart. I would explicitly say this.

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,229 @@
...
+
+This option is particularly useful for first time users or users that would like to have a local deployment.
+
+It is important to note that, if a Kind cluster name different from `kubeflex` is used, then its Control Plane name must be also be referenced during the Helm chart installation by setting `--set "kubeflex-controller.hostContainer=<control-plane-name>"`. The Control Plane name can be obtained by using `docker ps` to find the name of the container running the Kind cluster used for deploying KubeStellar Hub chart, _e.g._ `kubeflex-control-plane`.
```
MikeSpreitzer, line 29: It gets set to a container name, not a control plane name, right? See my other comment on this Value.

docs/content/direct/hub-chart.md
MikeSpreitzer: I would make the comment something more like "used to define the DNS domain name used from outside the KubeFlex hosting cluster to reach that cluster's Ingress endpoint".

docs/content/direct/hub-chart.md
MikeSpreitzer: I do not see a change. https://github.com/francostellari/kubestellar/blob/hub-chart/docs/content/direct/hub-chart.md?plain=1#L72 still says "host|vcluster". (That branch currently points at https://github.com/francostellari/kubestellar/commit/005f9bccad9321dcaae35fa578dee9e561132584)

docs/content/direct/hub-chart.md
MikeSpreitzer: I understand now. See my comment on this entry in the values.yaml file.

docs/content/direct/common-setup-hub-chart.md
MikeSpreitzer: There are different "you"s. I distinguish between the creators/maintainers (contributors, if you like) of KubeStellar and the _users_ of KubeStellar. It is unfriendly to make the _users_ of KubeStellar write and maintain scripts that have a lot of detailed knowledge of implementation details that us KubeStellar creators/maintainers are trying to wrap up and hide. I wish that we could provide scripts for users to invoke.

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,229 @@
...
+  - name: <wds2>     # mandatory name of the control plane
+    type: <host|k8s> # optional type of control plane host or k8s (default to k8s, if not specified)
+    APIGroups: ""    # optional string holding a comma-separated list of APIGroups
+    ITSName: <its2>  # optional name of the ITS control plane, this MUST be specified if more than one ITS exists at the moment the WDS PCH starts
+  ...
+```
+
+where `name` must specify a unique name of the control plane (not that this must be unique among both ITSes and WDSes), the optional `type` can be either k8s (default) or host, see [here](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md) for more information, the optional `APIGroups` provides a list of APIGroups, see [here](https://docs.kubestellar.io/release-0.22.0/direct/examples/#scenario-2-using-the-hosting-cluster-as-wds-to-deploy-a-custom-resource) for more information, and `ITSName` specify the ITS connected to the new WDS being created (this parameter MUST specified only if more that one ITS exists in the cluster, if no value is specified and only one ITS exists in the cluster, then it will automatically selected).
```
MikeSpreitzer, line 121: unique among all the user's control planes

MikeSpreitzer, line 121: I think that the wording about specifying the ITS can be improved. "only if X" means, among other things, "not, if not X". Rather than say "MUST be specified only if", I would write "MUST be specified if".

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,229 @@
...
+```
+
+The above command will install KubeFlex and the Post Create Hooks, but no Control Planes.
+Please remeber to add `--set "kubeflex-operator.isOpenShift=true"`, when installing into an OpenShift cluster.
+
+User defined control planes can be added using additional value files of `--set` arguments, _e.g._:
+
+- add a single ITS named its1 of default vcluster type: `--set-json='ITSes=[{"name":"its1"}]'`
+- add two ITSes named its1 and its2 of of type vlcuster and host, respectevely: `--set-json='ITSes=[{"name":"its1"},{"name":"its2","type":"host"}]'`
+- add a single WDS named wds1 of default k8s type: `--set-json='WDSes=[{"name":"wds1"}]'`
```
MikeSpreitzer, line 138: .. using the one and only ITS

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,229 @@
...
+
+A KubeStellar Hub installation compatible with the common setup suitable for Common Setup described in the [examples](./examples.md) could be achieved with the following command:
+
+```shell
+helm upgrade --install ks-hub oci://ghcr.io/kubestellar/kubestellar/hub-chart --version $KUBESTELLAR_VERSION \
+  --set-json='ITSes=[{"name":"its1"}]' \
+  --set-json='WDSes=[{"name":"wds1"}]'
```
MikeSpreitzer, line 145: Hooray! This is much simpler.

docs/content/direct/common-setup-hub-chart.md
MikeSpreitzer: "defined here"? What do you mean by "here"? The reader's attention is focused on the control planes involved in the Helm chart. But surprise! Now we get a command that deals with all the control planes in KubeFlex, regardless of where they came from or what they are for.

There is a non-small amount of bash code in the loop body here; if the user wants to apply it in any other way than being in this loop then there is a lot to copy and adapt.

docs/content/direct/common-setup-hub-chart.md
```diff
@@ -0,0 +1,80 @@
...
+
+After the KubeStellar Hub has fully deployed,
```
MikeSpreitzer, line 42: This glosses over the waiting that is spelled out in the other copy of this procedure.

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,229 @@
...
+  cpname=${cpname##*/}
+  while [[ $(kubectl get cp $cpname -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do
+    echo "Waiting for Control Plane $cpname..."
+    sleep 5
+  done
+  echo $cpname
+done
+```
+
+The following code can be used to obtain the kubeconfig of all the KubeFlex Control Planes in the hub cluster:
```
MikeSpreitzer, line 182: s/hub cluster/hosting cluster/

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,229 @@
...
+
+```shell
+echo "Getting the kubeconfig of all Control Planes..."
+for cpname in `kubectl get controlplane -o name`; do
+  cpname=${cpname##*/}
+  echo "Getting the kubeconfig of Control Planes $cpname ==> /tmp/kubeconfig-$cpname..."
```
MikeSpreitzer, line 188: s|/tmp/||

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,229 @@
...
+    kubectl get secret $(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.name}') \
+      -n $(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.namespace}') \
+      -o=jsonpath="{.data.$(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.key}')}" \
+      | base64 -d > "kubeconfig-$cpname"
+  fi
+  kubectl --kubeconfig "kubeconfig-$cpname" config rename-context $(kubectl --kubeconfig "kubeconfig-$cpname" config current-context) $cpname 2> /dev/null
+done
+```
+
+The code above will saves the kubeconfig of a control plane `$cpname` to a corresponding file `kubeconfig-$cpname` in the local folder.
```
MikeSpreitzer, line 201: You might want to mention, here and in the other copy of this text, the naming of the current context too.

hub-chart/templates/cps/wds.yaml
```diff
@@ -0,0 +1,14 @@
...
+---
+apiVersion: tenancy.kflex.kubestellar.org/v1alpha1
+kind: ControlPlane
+metadata:
+  name: {{ $cp.name }}
+spec:
+  backend: shared
+  type: {{ $cp.type | default "k8s" }}
+  postCreateHook: wds
+  postCreateHookVars:
```
MikeSpreitzer, line 11: Don't we need plumbing here for the logging verbosity of the controllers?

hub-chart/templates/postcreatehooks/wds.yaml
```diff
@@ -0,0 +1,221 @@
...
+      namespace: "{{"{{.Namespace}}"}}"
+  - apiVersion: batch/v1
+    kind: Job
+    metadata:
+      name: "{{"{{.HookName}}"}}"
+    spec:
+      template:
+        spec:
+          containers:
+          - name: "{{"{{.HookName}}-ks"}}"
```
MikeSpreitzer, line 177: I would make the name more specific, indicating that this is the kubestellar controller-manager.

hub-chart/templates/postcreatehooks/wds.yaml
```diff
@@ -0,0 +1,221 @@
...
+              - --version
+              - {{.Values.KUBESTELLAR_VERSION}}
+              - --set
+              - "{{"ControlPlaneName={{.ControlPlaneName}}"}}"
+              - --set
+              - "{{"ITSName={{.ITSName}}"}}"
+              - --set
+              - "{{"APIGroups={{.APIGroups}}"}}"
+              - --set
+              - "ControllerManager.Verbosity={{.Values.KUBESTELLAR_VERBOSITY_OVERRIDE | default .Values.VERBOSITY | default 4 }}"
```
MikeSpreitzer, line 196: I do not see these described in hub-chart/values.yaml or the user doc.

hub-chart/templates/postcreatehooks/wds.yaml
```diff
@@ -0,0 +1,221 @@
...
+            env:
+            - name: XDG_CACHE_HOME
+              value: /tmp/helm/.cache
+          - name: "{{"{{.HookName}}-otp"}}"
```
MikeSpreitzer, line 200: We usually call this the "transport controller".

hub-chart/templates/postcreatehooks/wds.yaml
```diff
@@ -0,0 +1,221 @@
...
+            image: quay.io/kubestellar/helm:{{.Values.HELM_VERSION}}
+            imagePullPolicy: IfNotPresent
+            args:
```
MikeSpreitzer, line 203: https://github.com/kubestellar/ocm-transport-plugin/issues/23

hub-chart/templates/pre-delete-hook.yaml
```diff
@@ -0,0 +1,49 @@
...
+  annotations:
+    # This is what defines this resource as a hook. Without this line, the
+    # job is considered part of the release.
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  template:
+    metadata:
+      name: "{{ .Release.Name }}"
```
MikeSpreitzer, line 21: Maybe add something like "predelete" into this name?

hub-chart/values-cp-only.yaml
```diff
@@ -0,0 +1,6 @@
+# KubeFlex override values
```
MikeSpreitzer, line 1: Nothing references this file. What is it for?

hub-chart/values.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+# Control controller log verbosity
+# The global VERBOSITY value will be sued for all controller unless a specific controller verbosity override is specified
+VERBOSITY: 2 # default global value for all controllers
+# Specific controller verbosity overrides:
+# KUBEFLEX_VERBOSITY_OVERRIDE: 6
+# KUBESTELLAR_VERBOSITY_OVERRIDE: 6
+# CLUSTERADM_VERBOSITY_OVERRIDE: 6
+# OCM_STATUS_ADDON_VERBOSITY_OVERRIDE: 6
+# OCM_TRANSPORT_PLUGIN_VERBOSITY_OVERRIDE: 6
```
MikeSpreitzer, line 24: This is not used anywhere (due to the relevant Helm chart not having the needed parameter).

hub-chart/values.yaml
```diff
@@ -0,0 +1,59 @@
...
+KUBECTL_VERSION: "1.29.3"
+KUBESTELLAR_VERSION: "0.23.0-alpha.2"
+CLUSTERADM_VERSION: "0.8.2"
+OCM_STATUS_ADDON_VERSION: "0.2.0-rc8"
+OCM_TRANSPORT_PLUGIN_VERSION: "0.1.7"
+
+
+# Control controller log verbosity
+# The global VERBOSITY value will be sued for all controller unless a specific controller verbosity override is specified
+VERBOSITY: 2 # default global value for all controllers
```
MikeSpreitzer, line 18: The default for the transport controller is currently 4. (That is not currently something that the transport controller chart exposes, but as soon as it does, changing the default to 2 would be a behavior change --- not a great thing.)

scripts/create-kind-cluster-with-SSL-passthrough.sh
```diff
@@ -0,0 +1,104 @@
...
+fi
+
+echo "Installing an nginx ingress with SSL passthrough enabled..."
+kubectl apply -f https://raw.githubusercontent.com/kubestellar/kubestellar/main/example/kind-nginx-ingress-with-SSL-passthrough.yaml
+
+if [[ "$wait" == "true" ]] ; then
+  echo "Waiting for the SSL passthrough patching to complete..."
+  sleep 30
+  kubectl wait --namespace ingress-nginx \
+    --for=condition=ready pod \
```
MikeSpreitzer, line 95: It is not guaranteed that the pod even exists yet. That is why other scripts that face this problem solve it by a loop in bash, with a `kubectl get` of the pod in each iteration, iterating until finding that the pod exists and is ready.

scripts/create-kind-cluster-with-SSL-passthrough.sh
```diff
@@ -0,0 +1,104 @@
...
+    --selector=app.kubernetes.io/component=controller \
+    --timeout=90s
+fi
+
+if [[ "$setcontext" == "true" ]] ; then
+  echo "Setting context to \"${name}\"..."
+  kubectl config rename-context "kind-${name}" "${name}"
+  kubectl config use-context "${name}"
```
MikeSpreitzer, line 103: This is conflating two very different things.

In the well-trodden path before this PR, the step of creating the hosting cluster would NOT rename the kubeconfig context for the hosting cluster but WOULD make it the current context.

MikeSpreitzer (commented):
I have finished a pass over this PR.

docs/content/direct/hub-chart.md
francostellari: ok

docs/content/direct/hub-chart.md
francostellari: sorry, I missed this one

.gitignore
francostellari: in the repo, hub-chart does not contain the dependency (kflex), which is pulled by go releaser before packaging the chart for the release. during that process 2 things are created:
1. Chart.lock
2. charts/....
During my local testing I obviously need the dependency... and I made the mistake before to push them to the git repo with commits... I added these lines to prevent that mistake.

docs/content/direct/common-setup-hub-chart.md
francostellari: ok, left over, fixed

docs/content/direct/common-setup-hub-chart.md
francostellari: @MikeSpreitzer I cannot cover all possible scenarios, I think.
This is intended to teach a user how to get a CP kubeconfig and provide an easy copy&paste way to do them all.
In you special cases you can choose what to do.
@clubanderson is already doing this in his cluster/setup... we need to get to him to see what he does.

I think the most important decision is to swithc to use separated kubeconfigs or continue merging into a single one and use contexts... this strongly impacts the rewording of the examples

Note that merging is not a good idea in general, for example all its configs have the same vcluster server name... there will be a conflict

docs/content/direct/common-setup-step-by-step.md
francostellari: yes and more further on, that's why I suggested 2 clean readable documents for the 2 processes

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,229 @@
...
+## Pre-requisites
+
+To install the Helm chart the only requirement is [Helm](https://helm.sh/).
+However, additional executables may be required to create/manage the cluster(s) (_e.g._, Kind and kubectl),
+to join Workload Execution Clusters (WEKs) (_e.g._, clusteradm),
+and to interact with Control Planes (_e.g._, kubectl), _etc_.
+For such purpose, a full list of executable that may be required can be fund [here](./pre-reqs.md).
+
+The setup of KubStellar via the Hub chart requires the existance of at least one cluster
+to be used for the deployment of the chart.
```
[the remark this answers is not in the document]
francostellari: I chaged it to hosting cluster

docs/content/direct/hub-chart.md
francostellari: The manual step by step starts with a `kflex init -c`... making most of this information unnecessary

hub-chart/values.yaml
francostellari: ok, done

docs/content/direct/hub-chart.md
francostellari: ok, done

docs/content/direct/hub-chart.md
francostellari: ok, I trust your better English

docs/content/direct/hub-chart.md
francostellari: ok

docs/content/direct/hub-chart.md
francostellari: yes, in general but in reality for a given Cp of known name and type, it's a single command

docs/content/direct/common-setup-hub-chart.md
francostellari: The reason is that waiting for the CP is, I believe, necessary and sufficient for getting the kubeconfigs but not sufficient for the cluster registration/joining operation. There is not a simple way to determine when OCM is ready not kubestellar.

For the moment I purposely was vague.

docs/content/direct/hub-chart.md
francostellari: ok

docs/content/direct/hub-chart.md
francostellari: ok

docs/content/direct/hub-chart.md
francostellari: ok done both

hub-chart/templates/postcreatehooks/wds.yaml
francostellari: ok but there is a length limit... what about kubestellar only

hub-chart/values.yaml
francostellari: I'm undecided... this was a way to provide an alternative to

```shell
helm upgrade --install add-wds2 oci://ghcr.io/kubestellar/kubestellar/hub-chart --version $KUBESTELLAR_VERSION \
  --set='kubeflex-operator.install=false,InstallPCHs=false' \
  --set-json='WDSes=[{name":"wds2"}]'
```

as

```shell
helm upgrade --install add-wds2 oci://ghcr.io/kubestellar/kubestellar/hub-chart --version $KUBESTELLAR_VERSION \
  --values=values-cp-only.yaml \
  --set-json='WDSes=[{name":"wds2"}]'
```

hub-chart/values.yaml
francostellari: yes all controllers have different defaults... can you work it out with @pdettori and @nirrozenbaum 
e.g. kflex has 0 and it is very verbose already

scripts/create-kind-cluster-with-SSL-passthrough.sh
francostellari: we can change it later, I found that it usually take more than 30sec and the wait time give a change to the pod to exist

hub-chart/values.yaml
francostellari: I have implemented them pending the PRs, they will be ignored until then

hub-chart/values.yaml
francostellari: also I realized than kflex, being a dependency, will have to be handled differetly

docs/content/direct/common-setup-hub-chart.md
pdettori: Why this step of obtaining all the kubeconfigs for all CPs and saving them in separate files in the current folder is required and how does it relate to the single chart ?

.gitignore
MikeSpreitzer: Thanks for the explanation.

docs/content/direct/common-setup-hub-chart.md
MikeSpreitzer: This is the future version of the examples.md file content. So the setup has to connect with the usage in the example scenarios.

MikeSpreitzer: @pdettori: This is the future version of the examples.md file content. So the setup has to connect with the usage in the example scenarios.

@francostellari: what about using `kflex` to do the kubeconfig updates?

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Pre-requisites
+
+To install the Helm chart the only requirement is [Helm](https://helm.sh/).
+However, additional executables may be required to create/manage the cluster(s) (_e.g._, Kind and kubectl),
+to join Workload Execution Clusters (WEKs) (_e.g._, clusteradm),
+and to interact with Control Planes (_e.g._, kubectl), _etc_.
+For such purpose, a full list of executable that may be required can be fund [here](./pre-reqs.md).
+
+The setup of KubStellar via the Hub chart requires the existance of at least one hosting cluster.
```
MikeSpreitzer, line 20: The setup does not use multiple hosting clusters. I would say "requires the existence of a KubeFlex hosting cluster".

docs/content/direct/hub-chart.md
MikeSpreitzer: `kflex init -c` is specific to `kind`. A lot of the information in this section is not specific to `kind`. Some of the information here is not specific to the hub chart, it is general considerations that are true independent of the Helm chart: the domain name and ingress port number of the hosting cluster are important configuration facts that have broader significance than just their appearance in this chart.

docs/content/direct/hub-chart.md
MikeSpreitzer: I still see "...hostContainer=\<control-plane-name\>". If I understand correctly, you mean to say "...hostContainer=\<control-plan-container-name\>. The control-plane-container-name is the name of the container in which kind or k3d is running the relevant control plane". And the initial qualification "if a Kind cluster" needs to be clarified to "when the hosting cluster was created by kind or k3d and its Ingress domain name is left to default to localtest.me". See https://github.com/kubestellar/kubeflex/issues/250#issuecomment-2115273144 .

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,262 @@
...
+  - name: <wds2>     # mandatory name of the control plane
+    type: <host|k8s> # optional type of control plane host or k8s (default to k8s, if not specified)
+    APIGroups: ""    # optional string holding a comma-separated list of APIGroups
+    ITSName: <its2>  # optional name of the ITS control plane, this MUST be specified if more than one ITS exists at the moment the WDS PCH starts
+  ...
+```
+
+where `name` must specify a name unique among all the control planes in that KubeFlex deployment (note that this must be unique among both ITSes and WDSes), the optional `type` can be either k8s (default) or host, see [here](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md) for more information, the optional `APIGroups` provides a list of APIGroups, see [here](https://docs.kubestellar.io/release-0.22.0/direct/examples/#scenario-2-using-the-hosting-cluster-as-wds-to-deploy-a-custom-resource) for more information, and `ITSName` specify the ITS connected to the new WDS being created (this parameter MUST be specified if more that one ITS exists in the cluster, if no value is specified and only one ITS exists in the cluster, then it will automatically selected).
```
MikeSpreitzer, line 120: s/will automatically/will be automatically/

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,262 @@
...
+
+1. using `kflex` CLI
+
+The kubeconfig of a `$cpname` Control Plane can be added as a context of the existing kubeconfig by using the command:
+
+```shell
+kflex ctx $cpname
+```
+
+Note that the command above will also switch the current context to `$cpname`, use `kflex ctx` to switch back to the initial context.
```
MikeSpreitzer, line 191: Remember that this does not work if that context was renamed after kflex created it.

hub-chart/templates/postcreatehooks/wds.yaml
```diff
@@ -0,0 +1,223 @@
...
+      kind: Role
+      name: "{{"{{.HookName}}"}}"
+    subjects:
+    - kind: ServiceAccount
+      name: default
+      namespace: "{{"{{.Namespace}}"}}"
+  - apiVersion: batch/v1
+    kind: Job
+    metadata:
+      name: "{{"{{.HookName}}"}}"
```
MikeSpreitzer, line 172: Perhaps add "kscm" into this name, to indicate specifically that it is for the KubeStellar controller-manager. I suggest something short because of the limit on Job name length (https://kubernetes.io/docs/concepts/workloads/controllers/job/#writing-a-job-spec).

hub-chart/templates/postcreatehooks/wds.yaml
```diff
@@ -0,0 +1,223 @@
...
+    spec:
+      template:
+        spec:
+          containers:
+          - name: "{{"{{.HookName}}-kubestellar"}}"
```
MikeSpreitzer, line 177: Why include the HookName in the _container_ name? The Job is already specific to the hook.

hub-chart/templates/pre-delete-hook.yaml
```diff
@@ -0,0 +1,49 @@
...
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+  annotations:
+    # This is what defines this resource as a hook. Without this line, the
+    # job is considered part of the release.
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  template:
+    metadata:
```
MikeSpreitzer, line 20: I notice that none of the example Job objects in https://kubernetes.io/docs/concepts/workloads/controllers/job have a `spec.template.metadata` section. Do we need to/can we/should we specify the pod name? The pod labels?

hub-chart/values-cp-only.yaml
MikeSpreitzer: Do you mean that you want this file to NOT be in git _and_ you want it to be in `.gitignore`? You could do that --- IF you do it in two steps: first one commit that deletes the file, and then a commit that adds it to `.gitignore`. Does not have to be done in this PR.

hub-chart/templates/pre-delete-hook.yaml
francostellari: @MikeSpreitzer is this specific for the pre-delete-hook policy? I can change it if you let me know what you want.

hub-chart/templates/postcreatehooks/wds.yaml
francostellari: I think it was done by some of the hook container before, so I maintained the "style". I guess it's a preference not a mandatory requirement.

hub-chart/templates/postcreatehooks/wds.yaml
francostellari: The hook name will be already `its` ... I think it's appropriate not to be specific of what is in it now and in the future, in my opinion

docs/content/direct/hub-chart.md
francostellari: ok but in this workflow we do not use `kflex init`

docs/content/direct/hub-chart.md
francostellari: ok, thank you

docs/content/direct/hub-chart.md
francostellari: ok

scripts/create-kind-cluster-with-SSL-passthrough.sh
francostellari: ok, I will change it back to old behavior

hub-chart/values-cp-only.yaml
francostellari: @MikeSpreitzer somehow I messed up my response above, the .gitignore response vas actually meant for the `Chart.lock` file.

This file was created because I was thinking of replacing this behavior:

```shell
helm upgrade --install add-wds2 oci://ghcr.io/kubestellar/kubestellar/hub-chart --version $KUBESTELLAR_VERSION \
  --set='kubeflex-operator.install=false,InstallPCHs=false' \
  --set-json='WDSes=[{name":"wds2"}]'
```

with this behavior using the file:

```shell
helm upgrade --install add-wds2 oci://ghcr.io/kubestellar/kubestellar/hub-chart --version $KUBESTELLAR_VERSION \
  --values=values-cp-only.yaml \
  --set-json='WDSes=[{name":"wds2"}]'
```

but I think I will delete the file for the moment.

docs/content/direct/hub-chart.md
francostellari: ok reworded as you suggested

docs/content/direct/hub-chart.md
francostellari: thank you

docs/content/direct/hub-chart.md
MikeSpreitzer: Hmm, now I am wondering now the user's kubeconfig file gets what `kflex ctx` needs.

hub-chart/templates/postcreatehooks/wds.yaml
MikeSpreitzer: This is WDS, not ITS.

This Job is not part of the user interface. I suggest naming this job in a way that helps someone semi-knowledgeable trying to debug something. Remember, for example, that a Job leaves a Pod lying around, some it is not at all uncommon for someone to get a listing of all the Pods. The purpose of this Job is to install the Helm chart for the kubestellar-controller-manager. I think that a name like "install-kubestellar-controller-manager" or "instantiate-kubestellar-controller-manager-chart" would be understandable.

This is not a blocker for this PR.

hub-chart/templates/pre-delete-hook.yaml
MikeSpreitzer: I thought that this is the only Job YAML with `spec.template.metadata`. I am not yet asserting something, I am trying to understand. Is there a specific need to specify the `spec.template.metadata` in this case? I will say that if it is not needed then perhaps it should not be there. In fact, I suspect there may be some confusion here. The Job controller usually takes care of the Pod names.  Following is an example of a Pod from the Job that invokes `clusteradm init`; the Job is named "ocm".

```
mspreitz@mjs13 ocm-status-addon % kubectl --context kind-kubeflex get pods -n its1-system ocm-tlhrp -o yaml
apiVersion: v1
kind: Pod
metadata:
  generateName: ocm-
  labels:
    batch.kubernetes.io/controller-uid: e6edf60f-e1eb-4cbb-88ee-5910cd30245f
    batch.kubernetes.io/job-name: ocm
    controller-uid: e6edf60f-e1eb-4cbb-88ee-5910cd30245f
    job-name: ocm
  name: ocm-tlhrp
  namespace: its1-system
  ownerReferences:
  - apiVersion: batch/v1
    blockOwnerDeletion: true
    controller: true
    kind: Job
    name: ocm
    uid: e6edf60f-e1eb-4cbb-88ee-5910cd30245f
...
```

docs/content/direct/hub-chart.md
francostellari: I'm not sure, it may require looking into the go code

MikeSpreitzer: You might find https://github.com/kubestellar/kubeflex/issues/202 helpful.

hub-chart/templates/pre-delete-hook.yaml
MikeSpreitzer: I wonder if there is some confusion here between the Job's `metadata` and that of the PodSpec inside the Job (`spec.template.metadata`).

scripts/create-k3s-cluster-with-SSL-passthrough.sh
```diff
@@ -0,0 +1,100 @@
...
+    shift
+done
+
+echo "Creating a k3s cluster with SSL passthrougn and ${port} port mapping..."
+
+if which kubectl > /dev/null ; then
+    if test -f ~/.kube/config; then
+        if $(kubectl cluster-info 2> /dev/null | grep -qe "control plane"); then
+            echo kubernetes cluster is already up!
+            kubectl cluster-info | /bin/head -3
```
MikeSpreitzer, line 61: On my mac:

```
mspreitz@mjs13 ocm-status-addon % /bin/head
zsh: no such file or directory: /bin/head
```

docs/content/direct/hub-chart.md
pdettori: @MikeSpreitzer good point. I have proposed a solution in https://github.com/kubestellar/kubestellar/issues/2091#issuecomment-2092951213

MikeSpreitzer: It seems to me that we have a fundamental conflict here between trying to do everything with a Helm chart (which can only affect the contents of a Kubernetes cluster) and KubeFlex keeping important state on the user's machine (which is inaccessible from the Kubernetes cluster).

If we are going to solve this by issuing one `kflex` command, why make it obscure --- why not just make it `kflex init`?

docs/content/direct/common-setup-hub-chart.md
```diff
@@ -0,0 +1,87 @@
...
+
+The setup of KubStellar via the Hub chart requires the existance of at least one cluster
+to be used for the deployment of the chart.
+
+For convenience, a new local Kind cluster that satisfies the requirements for KubeStellar setup
+and that can be used to exercises the [examples](./examples.md) can be created with the following command:
+
+```shell
+sudo su <<EOF
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${KUBESTELLAR_VERSION}/scripts/create-kind-cluster-with-SSL-passthrough.sh) --name kubeflex --port 9443
```
MikeSpreitzer, line 26: FYI, I do not expect the create-kind-cluster-with-SSL-passthrough.sh script to change with every release. I think that this line could have a literal `v0.23.0` rather than `v${KUBESTELLAR_VERSION}` (once v0.23.0 exists). If we ever do change it in a later release then the literal release number can be changed then. Not an issue for this PR, since release 0.23.0 does not exist yet.

scripts/create-k3s-cluster-with-SSL-passthrough.sh
MikeSpreitzer: Why not just use `head` ?

docs/content/direct/common-setup-hub-chart.md
MikeSpreitzer: This is in the context of the example setup, where the set of control planes is fixed. This line could simply be:

```
for cpname in wds1 its1; do
```

docs/content/direct/hub-chart.md
```diff
@@ -0,0 +1,263 @@
...
+```
+
+The kubeconfig of the KubeFlex control planes can be obtained both with and without the use of `kflex` CLI.
+
+1. using `kflex` CLI
+
+The kubeconfig of a `$cpname` Control Plane can be added as a context of the existing kubeconfig by using the command:
+
+```shell
+kflex ctx $cpname
```
MikeSpreitzer, line 189: If using this command, is the wait above necessary first? Does `kflex ctx $cpname` first wait for the named control plane to be ready? If not, could it be changed to do so? (@pdettori ?).

docs/content/direct/hub-chart.md
francostellari: @MikeSpreitzer this was my initial intention when I created the original kflex PR which was voted down. Besides some nomenclature https://github.com/kubestellar/kubeflex/pull/222
In any case I don't think this can be address in this PR.

scripts/create-k3s-cluster-with-SSL-passthrough.sh
francostellari: yes, changed, I was recycling some of Ed code... this is still experimental, I think

docs/content/direct/hub-chart.md
MikeSpreitzer: I think that this has to be addressed in this PR. Without the needed extension in the user's kubeconfig file, this will not work at all.

Perhaps the earlier discussion was not sufficiently informed by this difficulty. I do not remember thinking about this at that time.

MikeSpreitzer: I have realized that my previous statement is not quite right; https://github.com/kubestellar/kubeflex/pull/254 is a relevant doc update about KubeFlex and how it remembers the name of the kubeconfig context used to access the hosting cluster. The key realization is that `kflex ctx $cpname` _will_ work even if the hosting cluster context name is not yet stored in the user's kubeconfig **provided** that the current context is the one for accessing the hosting cluster. So I think that we can proceed in either of two ways.

1. Keep the chart's initial condition at "hosting cluster exists and is accessed by the kubeconfig current-context" and forbidding the user to change the kubeconfig current-context by any other means between (a) instantiating the chart and (b) using `kflex ctx $cpname` or `kflex create`.
2. Chang the chart's initial condition to "the hosting cluster has been kflex-initted (either by `kflex init` or instantiating a Helm chart) AND kflex has stored the name of the kubeconfig context used to access the hosting cluster" --- which will be less fragile if we upgrade `kflex` to store that more eagerly (https://github.com/kubestellar/kubeflex/issues/255).

MikeSpreitzer: After some discussion we decided to continue with the current chart design. To help with the sharp edge of `kflex ctx` being unable to do its job because the kubeconfig extension is missing, we can make the error message clearly explain the situation and how to remedy it. BTW, https://github.com/kubestellar/kubeflex/issues/255 suggests related improvements to `kflex`.

docs/content/direct/hub-chart.md
MikeSpreitzer: https://github.com/kubestellar/kubeflex/issues/258

.gitignore
MikeSpreitzer, line 33: s/hub/core/

Right now, this PR adds the file that you intend to exclude. We could fix this up in follow-on work.

.gitignore
```diff
@@ -29,4 +29,6 @@ local-chart/
...
+hub-chart/charts/
```
MikeSpreitzer, line 34: This PR currently includes core-chart/charts/kubeflex-operator-v0.6.1.tgz . That probably should also be absent from the PR and listed in .gitignore.

core-chart/templates/NOTES.txt
```diff
@@ -0,0 +1,53 @@
+Wait for all Control Planes to be ready:
```
MikeSpreitzer, line 1: See the comments on the other two copies/editions of this text.

core-chart/templates/kubeflex-clusterrolebinding.yaml
```diff
@@ -0,0 +1,18 @@
...
+{{- if eq $cp.type "host" }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: kubeflex-manager-cluster-admin-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
```
MikeSpreitzer, line 11: Is this much really necessary?
Cam be addressed in follow-on.

core-chart/values.yaml
```diff
@@ -0,0 +1,61 @@
...
+# The "default" verbosity value will be used for all controllers unless a specific controller verbosity override is specified
+verbosity:
+  default: 2
+  # Specific controller verbosity overrides:
+  # kubestellar: 6
+  # clusteradm: 6
+  # status_controller: 6
+  # status_agent: 6
+  # transport: 6
+
```
MikeSpreitzer, line 26: FYI, also coming into existence: controllers taking command line flags for request QPS and burst.

docs/content/direct/common-setup-core-chart.md
```diff
@@ -0,0 +1,87 @@
...
+A KubeStellar Core installation compatible with the common setup suitable for Common Setup described in the [examples](https://docs.kubestellar.io/release-0.22.0/direct/examples/) could be achieved with the following command:
+
+```shell
+helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart --version $KUBESTELLAR_VERSION \
+  --set-json='ITSes=[{"name":"its1"}]' \
+  --set-json='WDSes=[{"name":"wds1"}]'
+```
+
+Remember to include the option `--set "kubeflex-operator.isOpenShift=true"`, when deploying in an OpenShift cluster.
+
```
MikeSpreitzer, line 41: As discussed in another comment thread in this PR, right here there should be a bold note about the need to follow up with `kflex ctx $cpname` or `kflex create` before chaning the kubeconfig current-context by other means.

Can be done in follow-on.

docs/content/direct/core-chart.md
```diff
@@ -0,0 +1,279 @@
...
+
+The setup of KubStellar via the Core chart requires the existence of a KubeFlex hosting cluster.
+
+This can be:
+
+1. A local **Kind** or **k3s** cluster with an ingress with SSL passthrough and a mapping to host port 9443
+
+This option is particularly useful for first time users or users that would like to have a local deployment.
+
+It is important to note that, when the hosting cluster was created by **kind** or **k3s** and its Ingress domain name is left to default to localtest.me, then its Control Plane name must be also be referenced during the Helm chart installation by setting `--set "kubeflex-controller.hostContainer=<control-plane-container-name>"`.
```
MikeSpreitzer, line 28: s/its Control Plane name/the name of the container running hosting cluster/

Can be done in follow-on.

docs/content/direct/core-chart.md
```diff
@@ -0,0 +1,279 @@
...
+The `<control-plane-container-name>` is the name of the container in which kind or k3d is running the relevant control plane. One may use `docker ps` to find the `<control-plane-container-name>`.
+
+If a host port number different from the expected 9443 is used for the Kind cluster, then the same port number must be specified during the chart installation by adding the following argument `--set "kubeflex-controller.externalPort=<port>"`.
+
+By default the KubeStelalr Core chart uses a test domanin `localtest.me`, which is ok for testing on a single host machine. However, scenarios that span more than one machine, it may be useful to set `--set "kubeflex-controller.domanin=<domain>"` to a more appropriate `<domain>` that can be reached from Workload Execution CLusters (WECs).
```
MikeSpreitzer, line 33: s/domanin/domain/

Can be done in follow-on.

MikeSpreitzer, line 33: s/may be useful/is necessary/

core-chart/templates/kubeflex-clusterrolebinding.yaml
francostellari: it is necessary for host CPs. it will be installed only if host CP are being created

.gitignore
francostellari: thank you for catching it

docs/content/direct/common-setup-core-chart.md
```diff
@@ -0,0 +1,87 @@
...
+After the KubeStellar Core has fully deployed, `kflex` CLI can be used to retrieve the kubeconfig of all the KubeFlex Control Planes and store them as contexts of the current kubeconfig:
+
+```shell
+echo "Getting the kubeconfig of all Control Planes..."
+for cpname in `kubectl get controlplane -o name`; do
+  cpname=${cpname##*/}
+  echo "Getting the kubeconfig of Control Planes \"$cpname\"..."
+  kflex ctx $cpname
+done
```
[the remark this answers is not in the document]
francostellari: ok

MikeSpreitzer (commented):
I think that this can be merged now and work continue in follow-on PRs.

/lgtm

docs/content/direct/core-chart.md
francostellari: ok

francostellari: ok

docs/content/direct/core-chart.md
francostellari: ok

docs/content/direct/common-setup-core-chart.md
francostellari: I added a comment

.gitignore
francostellari: fixed

.github/workflows/goreleaser.yml
```diff
@@ -48,18 +49,21 @@ jobs:
...
+
     - name: Login to registry
       uses: docker/login-action@v3
       with:
         registry: ${{ env.REGISTRY }}
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
-      
+
     - name: Package and push chart
```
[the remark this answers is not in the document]
francostellari: ok

docs/content/direct/common-setup-hub-chart.md
francostellari: ok

docs/content/direct/common-setup-core-chart.md
```diff
@@ -41,22 +41,18 @@ Remember to include the option `--set "kubeflex-operator.isOpenShift=true"`, whe
...
+kflex ctx wds1
 kflex ctx # switch back to the initial context
 ```
 
-Afterwards the content of a Control Plane `$cpname` can be accessed by specifing its context:
+Afterwards the content of a Control Plane `<cpname>` can be accessed by specifing its context:
 
 ```shell
-kubectl --context "$cpname" ...
+kubectl --context <cpname> ...
```
MikeSpreitzer, line 55: FYI, I am not a fan of using something that is not shell syntax in an exhibited shell command.

MikeSpreitzer (commented):
This is good enough to merge and continue working on in follow-on.

/lgtm

MikeSpreitzer:
Something went wrong with the squash. This PR should now have only 1 commit, but it also includes 6 commits from other (merged) PRs.

MikeSpreitzer (approved):
Looks like the squash was fixed.
/lgtm
/approve
Thanks!

Issue #2160, opened 2024-05-14
E2E test re-org


MikeSpreitzer:
We currently have a confusing organization. The `test/e2e/` directory has two scenario-specific subdirectories (ginkgo, multi-cluster-deployment) plus one (common) for common stuff. Except that is no longer true; `test/e2e/multi-cluster-deployment/run-test.sh` can also run the ginkgo scenario. Also, `test/e2e/ginkgo` directly supports only using freshly-created `kind` clusters, while `test/e2e/multi-cluster-deployment/run-test.sh` allows using pre-existing OCP clusters. In fact, that generality is why that script was further generalized so that it can run the ginkgo scenario.

We should untangle this.

Following is a proposal.

Recognize that `test/e2e/multi-cluster-deployment/run-test.sh` is a general E2E test-runner, able to run either scenario.

Move `test/e2e/multi-cluster-deployment/run-test.sh` and `test/e2e/multi-cluster-deployment/README.md` up one level, to become `test/e2e/run-test.sh` and `test/e2e/README.md`. Adapt the contents of the new `test/e2e/README.md` to be the appropriate readme for all of E2E testing.

Remove the support from `test/e2e/ginkgo` for invoking the cleanup and setup steps; let `test/e2e/ginkgo` rely on the fact that it is run in the context of `test/e2e/run-test.sh` (or anything else that provides the necessary environment).

MikeSpreitzer:
@dumb0002 
@effi-ofer

dumb0002:
@effi-ofer, I agree with @MikeSpreitzer's proposal above. Furthermore, we could create a new directory for the bash tests and have the following directory structure: 

```
 — e2e
     — common
     — ginkgo
     — bash 
     README.md
     run-test.sh
```

The new `bash` directory will host the file: [use-kubestellar.sh](https://github.com/kubestellar/kubestellar/blob/main/test/e2e/multi-cluster-deployment/use-kubestellar.sh)

ezrasilvera:
@dumb0002 @effi-ofer I don't mind a re-org, but I would like to maintain the ability to run each of them separately without depending on a previous "state" . 
Also I want us to limit our platform support to only OCP and KIND at this point (e.g., not include k3d) till we have a better helm/automation that supports multiple platform. I prefer we don't complicate our testing setup script beyond reasonable limit,.

effi-ofer:
[...]

PR #2167, opened 2024-05-20
✨  E2E test re-organization

dumb0002:
## Summary

This PR improves the organization structure for the e2e tests

## Related issue(s)

https://github.com/kubestellar/kubestellar/issues/2160

Fixes #


test/e2e/README.md
```diff
@@ -1,6 +1,59 @@
...
+*         cluster2     <url>:port               <defaul-value>            default
+```
+
+Use the following command to rename the default context name for host cluster and workload execution clusters:
+
+```bash 
+$ kubectl config rename-context <default-wec1-context-name> cluster1
+```
+
+2. Run e2e test in your ocp cluster:
```
MikeSpreitzer, line 47: s/cluster/clusters/

test/e2e/README.md
```diff
@@ -1,6 +1,59 @@
-# End-to-End Tests
+# Kubestellar end to end testing
+
+This test is an executable variant of the "multi-cluster workload deployment with kubectl" scenario in [the examples doc](../../../docs/content/direct/examples.md). This test has the same prerequisites as the cited one. This test can test either (a) the  local copy of the repo or (b) the release identified in the kubestellar PostCreateHook (which will be the last release created, regardless of quality, except for that brief time when it identifies the release about to be made). Testing the local copy is the default behavior; to test the release identified in the PostCreateHook, pass `--released` on the command line of `run-test.sh`.
```
MikeSpreitzer, line 3: "This test"? This is now a generic runner that can run either of two scenarios, on either of two sorts of infrastructure.

test/e2e/bash/README.md
```diff
@@ -0,0 +1,14 @@
...
+This test is an executable variant of the "multi-cluster workload deployment with kubectl" scenario in [the examples doc](../../../docs/content/direct/examples.md). This test has the same prerequisites as the cited one. This test can test either (a) the  local copy of the repo or (b) the release identified in the kubestellar PostCreateHook (which will be the last release created, regardless of quality, except for that brief time when it identifies the release about to be made). Testing the local copy is the default behavior; to test the release identified in the PostCreateHook, pass `--released` on the command line of `run-test.sh`.
+
+## Running the test using a script
+
+Starting from a local directory containing the git repo, do the following.
+
+```
+cd test/e2e/bash
+../run-test.sh
+```
```
MikeSpreitzer, line 14: This is entirely redundant with `test/e2e/README.md`, right?

test/e2e/README.md
dumb0002: Fixed

test/e2e/README.md
dumb0002: Fixed - removed some contents to make the description more generic

test/e2e/bash/README.md
dumb0002: I added that README.md just to simplify the ability to run each test independently (bash & ginkgo).

test/e2e/bash/README.md
```diff
@@ -0,0 +1,14 @@
+#  Kubestellar end to end testing
+
+**PRE-REQ**: All of these tests use three `kind` clusters, so you need to [raise the relevant OS limits](https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files).
```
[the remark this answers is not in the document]
dumb0002: why so? this is a known issue of running more than 2 kind clusters inside a VM.

test/e2e/README.md
```diff
@@ -1,6 +1,59 @@
-# End-to-End Tests
+# Kubestellar end to end testing
+
+This e2e tests can test either (a) the  local copy of the repo or (b) the release identified in the kubestellar PostCreateHook (which will be the last release created, regardless of quality, except for that brief time when it identifies the release about to be made). Testing the local copy is the default behavior; to test the release identified in the PostCreateHook, pass `--released` on the command line of `run-test.sh`.
```
MikeSpreitzer, line 3: This directory does not contain one test. It contains a generic test runner and two subdirectories that each has a distinct test.

test/e2e/README.md
```diff
@@ -1,6 +1,59 @@
...
+
+The kubestellar controller-manager will be invoked with `-v=2` unless otherwise specified on the command line with `--kubestellar-controller-manager-verbosity $number`. This verbosity can not be set to a value other than 2 when using `--released`.
+
+The transport controller will be invoked with `-v=4` unless othewise specified on the command line with `--transport-controller-verbosity $number`.
+
+## Running the test on three new Kind clusters
 
 **PRE-REQ**: All of these tests use three `kind` clusters, so you need to [raise the relevant OS limits](https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files).
```
MikeSpreitzer, line 11: These tests can now be run on OCP clusters, right? The pre-req statement needs to be qualified to say that the OS limit needs to be raised IF the test will be run on `kind` clusters.

dumb0002: Note that the pre-req statement is already inside a section referring to run tests on kind cluster only.

test/e2e/README.md
```diff
@@ -1,6 +1,59 @@
...
 
-- [all the end to end tests](ginkgo)
-- [multi-cluster deployment using shell scripts](multi-cluster-deployment)
+Starting from a local directory containing the git repo, do the following.
+
+a) Bash tests:
+```bash
+cd test/e2e/multi-cluster-deployment
```
MikeSpreitzer, line 17: The `cd` should be just to `test/e2e`, right? Same question for all the other copies of this.

test/e2e/README.md
```diff
@@ -1,6 +1,59 @@
...
+
+```bash
+$ kubectl config get-contexts
+CURRENT   NAME          CLUSTER                   AUTHINFO               NAMESPACE
+          kscore       <url>:port               <defaul-value>            default
+          cluster1     <url>:port               <defaul-value>            default
+*         cluster2     <url>:port               <defaul-value>            default
+```
+
+Use the following command to rename the default context name for host cluster and workload execution clusters:
```
MikeSpreitzer, line 41: I have never been sure I understand the intent of this remark. Is the following an accurate restatement?

> FYI, if you need to rename a kubeconfig context in order to reach the above configuration then you can use the `kubectl config rename-context` command.

test/e2e/bash/README.md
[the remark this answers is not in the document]
MikeSpreitzer: The untrue part is "All of these tests use three `kind` clusters". These tests can use either `kind` or OCP.

test/e2e/bash/README.md
MikeSpreitzer: Even without this section, `test/e2e/README.md` should be telling the user how to run each test independently.

test/e2e/README.md
MikeSpreitzer: Oh, right.

test/e2e/README.md
dumb0002: Yes, that's an accurate restatement.

test/e2e/README.md
dumb0002: Good catch! fixed, thanks!

test/e2e/README.md
dumb0002: ok, I understand your comment better now. Just updated the content to address it.

test/e2e/bash/README.md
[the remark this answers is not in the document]
dumb0002: Note that comment is only included on the section about running the tests on a kind cluster

test/e2e/README.md
```diff
@@ -1,6 +1,59 @@
-# End-to-End Tests
+# Kubestellar end to end testing
+
```
MikeSpreitzer, line 2: This is the introduction to E2E testing. It should start with an explanation of the basic lay of the land. Something like the following. It should explain that this filesystem branch and the run-test.sh script here support the following test matrix.

- Run either one of two scenarios
- Either create three new `kind` clusters or use three existing OpenShift clusters
- Either test the local repo copy or last release before this version of the repo

with the exception that OCP can only be used with a release.

test/e2e/README.md
MikeSpreitzer: Putting an imperative "use the following command ..." at this point in the instructions without temporal qualification is confusing; without being suggested otherwise, the reader supposes that you are telling them something to do between the preceding material and the following material.

test/e2e/bash/README.md
dumb0002: `test/e2e/README.md` is already providing instruction to the user on how to run each test independently. We can remove the additional readme file at `test/e2e/bash/README.md` to avoid any confusion

test/e2e/bash/README.md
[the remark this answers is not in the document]
MikeSpreitzer: This copy is not in such a section.

test/e2e/bash/README.md
MikeSpreitzer: What is needed in `test/e2e/bash/README.md`, if it is not in `test/e2e/README.md`, is a description of the scenario.

test/e2e/bash/README.md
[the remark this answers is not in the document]
dumb0002: I suggest to remove this additional readme file at test/e2e/bash/README.md as it seems to be causing some confusion.

test/e2e/bash/README.md
dumb0002: Which scenario? could you clarify it? thx!

MikeSpreitzer: The scenario that is in this directory's use-kubestellar script. The description is the following text from the above paragraph.

> This test is an executable variant of the "multi-cluster workload deployment with kubectl" scenario in [the examples doc](../../../docs/content/direct/examples.md). This test has the same prerequisites as the cited one.

MikeSpreitzer (approved):
This is progress.

/lgtm
/approve

Thanks!

test/e2e/bash/README.md
dumb0002: Fixed! Added a description for the scenario

test/e2e/bash/README.md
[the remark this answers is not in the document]
dumb0002: Fixed! Added a description for the scenario and followed the suggested recommendation to keep test/e2e/bash/README.md

Issue #2141, opened 2024-05-09
[Combined-Status feature phase 3/3] Combined Status Generation


vMaroon:
_This is the last of the 3 mandatory planned PRs for Combined-Status #2143._

Given the combined-status decisions in the Bindings (after #2140) - the status controller should generate CombinedStatus objects that contain the requested status.

#### Completion Criterias
- Status controller generates CombinedStatus objects that cover the requested state specified in Bindings
- Testing

vMaroon:
Pending work from open PR:

## Misc
- [ ] E2E Tests
- [ ] Documentation

## Functional 
### Mandatory - Heavy
- [ ] [Availability of source object's `.spec` and `.metadata` for CELs](https://github.com/kubestellar/kubestellar/pull/2174#discussion_r1625431727)

### Mandatory - Light
- [ ] [StatusCollector validation](https://github.com/kubestellar/kubestellar/pull/2174#discussion_r1630076164)
- [ ] [WorkStatus extended identity validation](https://github.com/kubestellar/kubestellar/pull/2174#discussion_r1646672420)

### Optional
- [ ] [Updating resolution's workStatusData map values if sensible](https://github.com/kubestellar/kubestellar/pull/2174#discussion_r1625388132)
- [ ] [ReadLocked and WriteLocked function naming methodology explanation](https://github.com/kubestellar/kubestellar/pull/2174#discussion_r1630052874)
- [ ] [Minimize queueing in resolver when noting bindings](https://github.com/kubestellar/kubestellar/pull/2174#discussion_r1643670374)

### Unclassified
The following checklist may not be part of the work that closes this specific issue
- [ ] [CombinedStatus evaluated values type checking](https://github.com/kubestellar/kubestellar/pull/2174#discussion_r1650379839)
- [ ] [Error propagation to CRs](https://github.com/kubestellar/kubestellar/pull/2174#discussion_r1625417093)

MikeSpreitzer:
Does this need some boxes checked to account for recently merged PRs?

vMaroon:
@MikeSpreitzer I'm checking boxes when an associated PR is opened.

clubanderson:
Stale issues rot after 30d of inactivity.
Mark the issue as fresh with `/remove-lifecycle rotten`.
Rotten issues close after an additional 30d of inactivity.

If this issue is safe to close now please do so with `/close`.

/lifecycle rotten

clubanderson:
[...]

PR #2168, opened 2024-05-21
:sparkles: CombinedStatus Calculation (part 1)

vMaroon:
## Summary
This PR is the first in the set that implement #2141. Once the set is completed, Combined-Status would be functional.

In this PR a new binding resolution broker is introduced, which will be used by the to-be-created `CombinedStatusResolver`.

## High-Level Overall Picture
Peek of the overall design and intentions:
- Following this broker, a `WorkStatus` broker will be PR'd. The `WorkStatus` broker will allow callback registering for a finer-granularity notification, where observers will register for changes of a single workload-object (all of its associated work-statuses)
- Using the `binging.ResolutionBroker` and the `status.WorkStatusBroker`, a `CombinedStatusResolver` will maintain:
    - Per (binding) resolution,  and per `StatusCollector`, the information used of every relevant `WorkStatus`
        - When a `WorkStatus` is updated, the `status.WorkStatusBroker` would provide an event to trigger the resolver to fetch the updated information and queue a `CombinedStatus` reference to be updated if necessary. **This provides the necessary benefit of not having to re-evaluate all work-statuses on individual changes

## Related Issues
- #2141

### DIFF
--- a/pkg/binding/bindingpolicy-resolution-broker.go
+++ b/pkg/binding/bindingpolicy-resolution-broker.go
@@ -1,0 +1,129 @@
+/*
+Copyright 2024 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package binding
+
+import (
+	"github.com/kubestellar/kubestellar/api/control/v1alpha1"
+	"github.com/kubestellar/kubestellar/pkg/abstract"
+	"github.com/kubestellar/kubestellar/pkg/util"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sync"
+)
+
+// ResolutionBroker allows for the registration of callback functions that
+// are called when a resolution is updated.
+// It also allows for the retrieval of resolutions for a given bindingPolicyKey.
+type ResolutionBroker interface {
+	// RegisterCallback adds a new callback function that will be called,
+	// on a separate goroutine, when a resolution is updated. The callback
+	// function should accept a bindingPolicyKey.
+	// There is no deduplication of callbacks.
+	RegisterCallback(func(bindingPolicyKey string))
+
+	// GetResolution retrieves the resolution for a given bindingPolicyKey.
+	// If no resolution is associated with the given key, nil is returned.
+	GetResolution(bindingPolicyKey string) *Resolution
+}
+
+// Resolution is a struct that represents the resolution of a binding policy.
+// It contains the destinations and object data for the resolution.
+type Resolution struct {
+	// Destinations is a list of destinations that are the "where" part of the
+	// resolution.
+	Destinations []v1alpha1.Destination
+	// ObjectIdentifierToData is a map of object identifiers to their
+	// corresponding ObjectData.
+	ObjectIdentifierToData map[util.ObjectIdentifier]struct {
+		ResourceVersion  string
+		StatusCollectors []string
+	}
+}
+
+// NewResolutionBroker creates a new ResolutionBroker with the given
+// bindingPolicyResolutionGetter function.
+// The latter function's returned `bindingPolicyResolution` is expected to be
+// thread-safe under the constraint that only its methods are used.
+func NewResolutionBroker(bindingPolicyResolutionGetter func(bindingPolicyKey string) *bindingPolicyResolution) ResolutionBroker {
+	return &resolutionBroker{
+		bindingPolicyResolutionGetter: bindingPolicyResolutionGetter,
+	}
+}
+
+// resolutionBroker implements the ResolutionBroker interface.
+// The broker requires a bindingPolicyResolutionGetter function that retrieves
+// resolutions from a BindingPolicyResolver.
+type resolutionBroker struct {
+	sync.Mutex
+	callbacks []func(bindingPolicyKey string)
+	// bindingPolicyResolutionGetter is a function that retrieves the
+	// resolution for a given bindingPolicyKey. If no resolution is associated
+	// with the given key, nil is returned.
+	// The returned `bindingPolicyResolution` is expected to be thread-safe as
+	// long as only its methods are used.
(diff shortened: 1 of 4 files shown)


pkg/binding/bindingpolicy-resolution-broker.go
```diff
@@ -0,0 +1,129 @@
...
+					StatusCollectors: sets.List(data.StatusCollectors), // members are string copies
+				}
+			}), // while this function breaks the constraint, it maintains its own concurrency safety
+		// by using the PrimitiveMapSafeMap which transforms a map safely using its read-lock.
+	}
+}
+
+// NotifyCallbacks calls all registered callbacks with the given bindingPolicyKey.
+// The callbacks are called on separate goroutines.
+func (broker *resolutionBroker) NotifyCallbacks(bindingPolicyKey string) {
```
vMaroon, line 122: This function is not exported within the interface because it should only be called within the binding module.

pkg/status/controller.go
```diff
@@ -46,16 +45,14 @@ import (
...
+// Controller watches workstatues and checks whether the corresponding
 // workload object asks for the singleton status returning. If yes,
 // the full status will be copied to the workload object in WDS.
 type Controller struct {
 	logger             logr.Logger
 	wdsName            string
 	wdsDynClient       dynamic.Interface
-	wdsKubeClient      kubernetes.Interface
 	itsDynClient       dynamic.Interface
-	itsKubeClient      kubernetes.Interface
```
MikeSpreitzer, line 58: FYI, all the changes in this file are cleanup that can be done in an independent PR.

pkg/binding/bindingpolicy-resolution-broker.go
```diff
@@ -0,0 +1,129 @@
...
+
+// ResolutionBroker allows for the registration of callback functions that
+// are called when a resolution is updated.
+// It also allows for the retrieval of resolutions for a given bindingPolicyKey.
+type ResolutionBroker interface {
+	// RegisterCallback adds a new callback function that will be called,
+	// on a separate goroutine, when a resolution is updated. The callback
+	// function should accept a bindingPolicyKey.
+	// There is no deduplication of callbacks.
+	RegisterCallback(func(bindingPolicyKey string))
```
MikeSpreitzer, line 35: FYI, lesson learned in the legacy system: because go `func` values are not comparable, it is impossible to later add an "Unregister" method. If, OTOH, the thing registered were an interface with a callback method, then "Unregister" would be possible. Not a blocker for this PR.

pkg/binding/bindingpolicy-resolution-broker.go
```diff
@@ -0,0 +1,129 @@
...
+	// bindingPolicyResolutionGetter is a function that retrieves the
+	// resolution for a given bindingPolicyKey. If no resolution is associated
+	// with the given key, nil is returned.
+	// The returned `bindingPolicyResolution` is expected to be thread-safe as
+	// long as only its methods are used.
+	bindingPolicyResolutionGetter func(bindingPolicyKey string) *bindingPolicyResolution
+}
+
+// RegisterCallback adds a new callback function that will be called, on a
+// separate goroutine, when a resolution is updated. The callback function
```
MikeSpreitzer, line 81: Would the caller be simpler if there were also an initial callback?

pkg/binding/bindingpolicy-resolution-broker.go
```diff
@@ -0,0 +1,129 @@
...
+type Resolution struct {
+	// Destinations is a list of destinations that are the "where" part of the
+	// resolution.
+	Destinations []v1alpha1.Destination
+	// ObjectIdentifierToData is a map of object identifiers to their
+	// corresponding ObjectData.
+	ObjectIdentifierToData map[util.ObjectIdentifier]struct {
+		ResourceVersion  string
+		StatusCollectors []string
+	}
```
MikeSpreitzer, line 53: This code will get less verbose if there is a name for this struct type.

pkg/binding/bindingpolicy-resolution-broker.go
```diff
@@ -0,0 +1,129 @@
...
+				StatusCollectors []string
+			} {
+				return struct {
+					ResourceVersion  string
+					StatusCollectors []string
+				}{
+					ResourceVersion:  data.ResourceVersion,
+					StatusCollectors: sets.List(data.StatusCollectors), // members are string copies
+				}
+			}), // while this function breaks the constraint, it maintains its own concurrency safety
```
MikeSpreitzer, line 115: FYI, GetResolution would be more succinct if this func were be a method of `*objectData`.

MikeSpreitzer, line 115: Which constraint is broken here?

pkg/binding/bindingpolicy-resolution-broker.go
MikeSpreitzer: What will call this method?

pkg/binding/bindingpolicy-resolution-broker.go
```diff
@@ -0,0 +1,129 @@
...
+	RegisterCallback(func(bindingPolicyKey string))
+
+	// GetResolution retrieves the resolution for a given bindingPolicyKey.
+	// If no resolution is associated with the given key, nil is returned.
+	GetResolution(bindingPolicyKey string) *Resolution
+}
+
+// Resolution is a struct that represents the resolution of a binding policy.
+// It contains the destinations and object data for the resolution.
+type Resolution struct {
```
MikeSpreitzer, line 44: Is it necessary to make this an immutable data structure? What would go wrong if it were an interface that provides procedural acces? Something like the following?

```
type Resolution interface {
  Destinations(yield(v1alpha1.Destination) bool)
  Workload(yield(util.ObjectIdentifier, struct{ResourceVersion  string; StatusCollectors /*not modified during this call*/ Set[string]}) bool)
}
```

MikeSpreitzer (approved):
I have questions, but they can be addressed in follow-on work.

/lgtm
/approve

Thanks!

MikeSpreitzer:
Looks like `make imports` is needed.

pkg/binding/bindingpolicy-resolution-broker.go
vMaroon: The `BindingPolicyResolver`.

pkg/binding/bindingpolicy-resolution-broker.go
vMaroon: This also makes sense. I thought about registering callbacks before "starting" the resolver - I will go with the simplest of the two when the broker is used.

vMaroon:
@MikeSpreitzer fixed. 
I think once the status feature is merged, we can run a period of refactoring and overall alignment of the status controller with the rest (which I won't be doing this week) - and abstract the resolver if makes sense since it will exist in two forms.

PR #2170, opened 2024-05-22
📖 Brush up E2E READMEs

MikeSpreitzer:
## Summary
This PR updates the `test/e2e/README.md` to be a decent introduction to the whole topic of E2E testing, and trims the READMEs in the scenario subdirectories to focus on describing their scenario.

## Related issue(s)

Fixes #


dumb0002:
@MikeSpreitzer, in addition to trim down description, IMO we should keep the examples of how to run the e2e tests for example: 

a) Bash tests:
```bash
cd test/e2e
./run-test.sh
```

b) Ginkgo tests:
```bash
cd test/e2e
./run-test.sh --test-type ginkgo
```

MikeSpreitzer:
@dumb0002: one of those two is not like the other. `test/e2e/ginkgo/README.md` describes how to run the ginkgo test suite WITHOUT using the generic test runner (as well as pointing to the doc on the generic test runner). For the bash scenario we have nothing analogous; `test/e2e/bash/use-kubestellar.sh` can ONLY be used from the generic runner; there is nothing to add to the description in `test/e2e/README.md` of how to run it.

PR #2179, opened 2024-05-29
📖 Improve code to retrieve the CP kubeconfigs

francostellari:
## Summary

Improve code to retrieve the CP kubeconfigs

## Related issue(s)

Fixes #

### DIFF
--- a/core-chart/templates/NOTES.txt
+++ b/core-chart/templates/NOTES.txt
@@ -1,53 +1,54 @@
-Wait for all Control Planes to be ready:
+Below is a code snippet that allows to wait for all KubeFlex Control Planes to be ready:
 
-echo "Waiting for all Control Planes to be ready..."
+echo "Waiting for all KubeFlex Control Planes to be Ready:"
 for cpname in `kubectl get controlplane -o name`; do
-    cpname=${cpname##*/}
-    while [[ $(kubectl get cp $cpname -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do
-        echo "waiting for Control Plane $cpname..."
-        sleep 5
-    done
-    echo $cpname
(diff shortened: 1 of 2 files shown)


core-chart/templates/NOTES.txt
```diff
@@ -1,53 +1,54 @@
...
-    -o=jsonpath="{.data.$(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.key}')}" \
-    | base64 -d > "./kubeconfig-$cpname"
+  -n $(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.namespace}') \
+  -o=jsonpath="{.data.$(kubectl get controlplane $cpname -o=jsonpath='{.status.secretRef.key}')}" \
+  | base64 -d > "./kubeconfig-$cpname"
 
 which will create a file "kubeconfig-$cpname" in the current folder.
 
-Obtain the kubeconfig to access all Control Plane with the command:
+The following code snippet allows one to obtain the kubeconfig of all KubeFlex Control Plane in the cluster:
```
MikeSpreitzer, line 24: In English it would be "all ... Control Planes" or "every ... Control Plane" or "each ... Control Plane". The last would best work in the context of the rest of the sentence ("the kubeconfig of").

MikeSpreitzer, line 24: More specifcally than "code snippet", it is some shell commands. Which shells does do they work in?

MikeSpreitzer (commented):
The only functional change I see here is removing some problematic "/tmp/" from pathnames, which is good.

I see no reason for the change from backticks to dollar-parens, but it makes no real difference.

I think the wording could use further work, does not have to be done in this PR.

/lgtm
/approve
Thanks!

Issue #2163, opened 2024-05-15
Redesigning the organization of the kubestellar.io website (place for discussion)


KPRoche:
The current organization of the kubestellar.io (and by extension, docs.kubestellar.io) makes it difficult, among other things, for potential new users to actually *try* Kubestellar.

This issue is a start at reviewing and redesigning the outline/information flow of the site to address those issues. 
It will undoubtedly take multiple PRs to implement the changes. 
I have created a fork https://github.com/KPRoche/kubestellar/tree/ksdocs where you can see the source files I'm using locally to render these new version. Everything is in the /docs subtree of the fork. 

# notes on rendering via mkdocs
The site is built from markdown files in the /docs subtree, which means they can be read individually in the github repo, or in their combined website version as rendered by mkdocs. The actual site navigation is laid out in the file _/docs/mkdocs.hml_ 
Each main menu heading, if selected, loads and opens the first linked file in its section and loads its submenu into the left menu area.
The Table of contents menu on the right is generated from the headings in the original markdown file

# proposed new website outline
Note that with a couple of exceptions, the existing content files have not been modified yet. 

## Landing Page
The current landing/welcome page (_docs/content/index.md_) suppresses the regular menu (and release info/selector). These should be restored for the sake of consistent navigation. (I have not figured out exactly where the suppression takes place; this is a question for @clubanderson perhaps)

## Main Navigation
![image](https://github.com/kubestellar/kubestellar/assets/25445603/34ccc85e-9bca-45a0-bbfd-e493ff2f7e73)
The main menu headings have been slightly adjusted to be:
*Welcome | Overview | Using KubeStellar | Contributing | Community | Blog*

## Overview section
![image](https://github.com/kubestellar/kubestellar/assets/25445603/6563c6d4-dafc-43c1-9111-6dcde3487d95)
The user guide has been moved out of this section, and the menu item *Readme* has been relabeled *Basics*

## Using KubeStellar section
_This used to be the Getting Started menu item; note the displayed page is a placeholder_

![image](https://github.com/kubestellar/kubestellar/assets/25445603/d72454d7-d0b1-49ce-a20e-a0210fc33ed2)

_This submenu is the most dramatically changed from the existing site. The content for all the sub-items does not exist yet_

* This section is redesigned to allow for first a simple description of the configuration and installation process, then to work through the flow of an actual setup.  What steps should be called out as menu items (as opposed to being part of the TOC in a single page) is a good topic for discussion
[...]

PR #2183, opened 2024-05-30
:book: Basic website redesign applied to current main

KPRoche:
## Summary
This is an implementation of the basic restructuring of our web docs to be more easily navigable:
It includes added menu items in the landing page
slight reorg (naming) of top level menu items
Significant reorganization of the Using Kubestellar (getting started) section to prune empty pages, and add installation instructions based on the common-setup configuration in our examples.
It also adds a couple of high-level "interstitial" overview pages in those sections.

## Related issue(s)
#2163 
Fixes #


docs/mkdocs.yml
```diff
@@ -34,33 +34,56 @@ site_dir: 'generated'
...
+#      - User guide: direct/user-guide.md
       - Architecture: direct/architecture.md
+      - Related:
+        - KubeFlex: direct/kubeflex-intro.md
+        - KubeStellar Galaxy: direct/galaxy-intro.md
       - Release-notes: direct/release-notes.md
-  - Getting Started: 
-          - Pre-reqs: direct/pre-reqs.md
+  - Using Kubestellar:
+          - Overview: direct/setup-overview.md
```
MikeSpreitzer, line 46: It seems wrong to have two pages titled "Overview". Maybe this second one is "Overview of setting up KubeStellar"? or "Overview of KubeStellar setup"? Maybe "install" or "deploy" instead of "setup"?

docs/mkdocs.yml
```diff
@@ -34,33 +34,56 @@ site_dir: 'generated'
...
+                - Step by Step: direct/common-setup-step-by-step.md
+                - (experimental) Via Helm Chart: direct/core-chart.md
+#                - Prepare the environment: common-subs/placeholder.md
+#                - Initialize Kubeflex: common-subs/placeholder.md
+#                - Install Kubestellar Core components: common-subs/placeholder.md
+#                - Create the Inventory & Transport Space: common-subs/placeholder.md
+#                - Create the Workload Description Space: common-subs/placeholder.md
+              - Set up and Register Workload Execution Clusters (WECs): direct/example-wecs.md
+#              - Register WECS with core: common-subs/placeholder.md
+              - Create Workload Deployments: direct/deploy-workloads.md
```
MikeSpreitzer, line 61: I do not love the "deploy" terminology here.

(A) In some sense, it is KubeStellar's job to "deploy" the workload in each WEC. What this section is trying to talk about, I think, is not that but rather putting the workload desired state into the WDS(es).

(B) What goes into the WDS is not just the workload desired state from the user, but also: the control object desired state from the user, the control object reported state from the KubeStellar implementation, and the workload reported state from the WEC(s) (originally) through KubeStellar.

docs/mkdocs.yml
```diff
@@ -34,33 +34,56 @@ site_dir: 'generated'
...
+#          - Modifying Workloads: common-subs/placeholder.md
+          - Usage & Installation Examples:
```
MikeSpreitzer, line 63: I think that it would be better to order the words temporally: "Installation and usage examples".

docs/mkdocs.yml
KPRoche: I am not wedded to a particular phrasing, but it needs to be only two or three words.
Would "commission workloads" be a better phrase?

docs/mkdocs.yml
KPRoche: Good point!

docs/mkdocs.yml
KPRoche: I was deliberately avoiding "deploy" for some of the same reasons you raise regarding workloads in your other comment.
How about "Basics" as the intro header?

docs/mkdocs.yml
KPRoche: image

Something like this (and this incorporates getting rid of the second overview menu item)

KPRoche: changed deploy to commission

docs/mkdocs.yml
KPRoche: changed overview to basics

docs/mkdocs.yml
MikeSpreitzer: I had expected to follow the lead of the term "Workload _Description_ Space", so say "describe". But what goes in that space is not just the description of the workload but also the control objects.

"commission" is OK. It is more general and vague, can conceivably include creating the control objects as well as the workload desired state.

docs/content/direct/commission-workloads.md
```diff
@@ -0,0 +1,5 @@
+# Deploying Workloads KubeStellar Usage Examples
+
+We are still working on generalized step-by-step instructions for defining, binding and deploying particular workloads.
```
MikeSpreitzer, line 3: I think that the terminology here still needs some work.

I make a distinction between "workload objects" and "control objects". I think that this file is trying to talk about creating the desired state for both sorts of object. The reported state for each does _not_ come from the user.

docs/content/direct/common-setup-kind.md
```diff
@@ -0,0 +1,100 @@
...
+
+1. Make sure you have all the prerequisite software installed [pre-reqs](pre-reqs.md).
+
+1. If you ran through a setup (for one of our example scenarios) previously then you will need to do a bit of cleanup first. See how it is done in the cleanup script for our E2E tests (in `test/e2e/common/cleanup.sh`).
+
+## Setting up the KubeStellar Core components
+
+The following steps establish the setup in our [example scenarios](examples.md) and are a good starting point.
+
+### Prepare the environment ###
```
MikeSpreitzer, line 13: What is the trailing `###` here for?

docs/newcontent/.pages
```diff
@@ -0,0 +1,8 @@
+# nav:
```
MikeSpreitzer, line 1: What is this file for? Did you intend to include it?

docs/newcontent/.pages:Zone.Identifier
```diff
@@ -0,0 +1,3 @@
+[ZoneTransfer]
+ZoneId=3
+ReferrerUrl=C:\Users\KevinRoche\Downloads\kubestellar-ksdocs.zip
```
MikeSpreitzer, line 3: This looks like something that does not belong in git. Same for all the other :Zone.Identifier files.

docs/mkdocs.yml
```diff
@@ -34,33 +34,56 @@ site_dir: 'generated'
...
+      - Related:
+        - KubeFlex: direct/kubeflex-intro.md
+        - KubeStellar Galaxy: direct/galaxy-intro.md
       - Release-notes: direct/release-notes.md
-  - Getting Started: 
-          - Pre-reqs: direct/pre-reqs.md
+  - Using Kubestellar:
+          - Basics: direct/usage-basics.md
+          - Prerequisites: direct/pre-reqs.md
+#              - Pre-reqs: direct/pre-reqs.md
```
MikeSpreitzer, line 48: Why retain this commented-out line?

docs/mkdocs.yml
```diff
@@ -34,33 +34,56 @@ site_dir: 'generated'
...
-      - Security: 
-          - Policy: Contribution guidelines/security/security.md
-          - Contacts: Contribution guidelines/security/security_contacts.md
+#      - License: Contribution guidelines/LICENSE.md
+#      - Governance: Contribution guidelines/governance.md
+#      - Onboarding: Contribution guidelines/onboarding.md
+#      - Website: Contribution guidelines/operations/document-management.md
+#      - Security:
+#          - Policy: Contribution guidelines/security/security.md
+#          - Contacts: Contribution guidelines/security/security_contacts.md
```
MikeSpreitzer, line 86: This is governance stuff that has to stay visible on the website as part of being in the CNCF Sandbox.

docs/mkdocs.yml
```diff
@@ -34,33 +34,56 @@ site_dir: 'generated'
...
+#              - Using existing hosting cluster: direct/hosting-cluster.md
+#              - KubeStellar on KIND: direct/ks-on-kind.md
+              - Deploying KubeStellar on K3D: direct/deploy-on-k3d.md
+#              - KubeStellar on OCP: direct/ks-on-ocp.md
           - Best practices: direct/best-practices.md
-          - Examples: direct/examples.md
+#          - API reference: common-subs/placeholder.md
+          - User Guide: direct/user-guide.md
   - Contributing: 
-      - Overview of contributing: direct/contributor.md
```
MikeSpreitzer, line 52: There is some content in this file that should not disappear from the website.

docs/mkdocs - Copy.yml
```diff
@@ -0,0 +1,213 @@
+site_name: KubeStellar
```
MikeSpreitzer, line 1: Do we want this file in Git?

docs/content/direct/pre-reqs.old.md
```diff
@@ -0,0 +1,99 @@
+# KubeStellar prerequisites
```
MikeSpreitzer, line 1: Do we want this file in Git?

docs/content/direct/common-setup-intro.md
```diff
@@ -0,0 +1,16 @@
...
+
+
+## Common Setup
+
+The following steps set up the hub and establish its initial state
+
+There are two main ways to proceed:
+
+2. [Step-by-step manual setup](common-setup-step-by-step.md)
+1. [Automatically Using KubeStellar Helm chart](core-chart.md)<br/>
```
MikeSpreitzer, line 13: Before this PR, there are two files that serve different purposes:
- one that is the draft of the new contents for examples.md, showing the "getting started" or "quickstart" version of using the core chart; and
- core-chart.md, the user guide contents about the core chart.

I think that it makes sense to not conflate the two. The plan was to move the first file inline into examples.md once @francostellari was sufficiently confident about it; the "step-by-step" alternative is just an interim thing until then --- the plan was that it gets deleted by the time we make the next release.

docs/newcontent/Community/_index.md
```diff
@@ -0,0 +1,34 @@
+
```
MikeSpreitzer, line 1: As far as I can tell, there is no path from `docs/mkdocs.yml` to `docs/newcontent/<anything>`. Did you mean to include this filesystem branch in this PR?

docs/content/direct/common-setup-kind.md
```diff
@@ -0,0 +1,100 @@
+# KubeStellar Core Sample Setup -- Kind with OCM
```
MikeSpreitzer, line 1: Nothing references this file. What is the thinking here?

MikeSpreitzer, line 1: Why is "with OCM" here? As far as I know, we do not have "without OCM" option.

docs/mkdocs.yml
```diff
@@ -34,33 +34,56 @@ site_dir: 'generated'
...
+#                - Prepare the environment: common-subs/placeholder.md
+#                - Initialize Kubeflex: common-subs/placeholder.md
+#                - Install Kubestellar Core components: common-subs/placeholder.md
+#                - Create the Inventory & Transport Space: common-subs/placeholder.md
+#                - Create the Workload Description Space: common-subs/placeholder.md
+              - Set up and Register Workload Execution Clusters (WECs): direct/example-wecs.md
+#              - Register WECS with core: common-subs/placeholder.md
+              - Commission Workloads: direct/commisssion-workloads.md
+#          - Modifying Workloads: common-subs/placeholder.md
+          - Installation & Usage Examples:
```
MikeSpreitzer, line 63: Why is "Installation" here, separate from the previous "Set Up" section? They seem roughly synonymous.

docs/mkdocs.yml
```diff
@@ -34,33 +34,56 @@ site_dir: 'generated'
...
+              - Example Cluster Preparation: direct/hosting-cluster.md
```
MikeSpreitzer, line 64: This particular file is specifically about the usage of a _pre-existing_ hosting cluster, in contrast the other content that _creates_ the hosting cluster.

docs/mkdocs.yml
```diff
@@ -34,33 +34,56 @@ site_dir: 'generated'
...
+                - Step by Step: direct/common-setup-step-by-step.md
+                - (experimental) Via Helm Chart: direct/core-chart.md
+#                - Prepare the environment: common-subs/placeholder.md
+#                - Initialize Kubeflex: common-subs/placeholder.md
+#                - Install Kubestellar Core components: common-subs/placeholder.md
+#                - Create the Inventory & Transport Space: common-subs/placeholder.md
+#                - Create the Workload Description Space: common-subs/placeholder.md
+              - Set up and Register Workload Execution Clusters (WECs): direct/example-wecs.md
+#              - Register WECS with core: common-subs/placeholder.md
+              - Commission Workloads: direct/commisssion-workloads.md
```
MikeSpreitzer, line 61: Typo: too many "s" in "comisssion".

docs/content/direct/common-setup-intro.md
MikeSpreitzer: The introduction of the core Helm chart emphasizes this, but it was already true: setting up KubeStellar does not require creating a hosting cluster --- a pre-existing cluster could be used. The quickstart begins by creating a hosting cluster using `kind` (indirectly through `kflex init --create-kind`), but that is not the only way. Really there are three steps to discern here:

1. Acquire the use of a cluster (create one or get credentials to use an existing one).
2. Initialize it for use as a KubeFlex hosting cluster.
3. Install the KubeStellar core components in this cluster.

docs/content/direct/pre-reqs.md
```diff
@@ -1,15 +1,20 @@
...
+
+
+## Infrastructure (clusters)
+
+Because of its multicluster architecture, KubeStellar requires that you have the necessary privileges and infrastructure access to create and/or configure the necessary clusters. Our current examples use kind or openshift to do this.
 
 ## For Using KubeStellar
 
-- kubeflex version 0.6.1 or higher
+- **kubeflex** version 0.6.1 or higher
```
MikeSpreitzer, line 13: Actually we are making it so that the KubeFlex CLI, `kflex`, in _not_ required.
@francostellari 
Pointing this out can be done in a separate PR.

ezrasilvera:
@KPRoche This PR shows as if everything is completely new.  I'm not sure the expectation here is indeed to review 164 new files.  Are all files indeed new or they were just moved ? If moved, the PR should indicate this so we can focus on reviewing the actual change.

docs/content/direct/common-setup-kind.md
KPRoche: copied from something else. I inferred it was the equivalent of an html header close tag. My mistake

docs/newcontent/.pages:Zone.Identifier
KPRoche: Clearly I screwed up something in my commit.

docs/content/direct/common-setup-kind.md
KPRoche: It was for parallelism in the list of sample setups, going back to a slack conversation a couple months ago.

docs/newcontent/Community/_index.md
KPRoche: no.

docs/mkdocs - Copy.yml
KPRoche: No. I clearly messed up my commit

docs/content/direct/pre-reqs.old.md
KPRoche: no, deleted.

KPRoche:
@MikeSpreitzer and @ezrasilvera -- I've done (I hope) a good cleanup of my commit error from last week. Sorry for the mess.

A general note: There are some MD files still in the docs folders which are not called in my current structure. There are also commented-out nav links in the mkdocs.yaml file. 
These are there because I am aware that the content almost certainly needs to be included *somehow* in the website but didn't know exactly where, and wanted to avoid having to re-add or create filesfrom scratch as opposed to just linking to them after editing. Similarly, the commented out navigation links in mkdocs.yaml are because I am sure it will be evolving more, and I had quite the struggle to get the nav syntax figured out in the first place.

docs/mkdocs.yml
KPRoche: un-commented those lines and re-pushed the file.

docs/mkdocs.yml
```diff
@@ -34,31 +34,54 @@ site_dir: 'generated'
...
+      - Overview: readme.md
+#      - User guide: direct/user-guide.md
       - Architecture: direct/architecture.md
+      - Related:
+        - KubeFlex: direct/kubeflex-intro.md
+        - KubeStellar Galaxy: direct/galaxy-intro.md
       - Release-notes: direct/release-notes.md
-  - Getting Started: 
-          - Pre-reqs: direct/pre-reqs.md
+  - Using Kubestellar:
```
ezrasilvera, line 45: Most project do have "getting started" I think people expecting this.

docs/mkdocs.yml
ezrasilvera: I don't  think any of these steps should be in the mkdocs menu .   I think we want a very simple menu:

- Using Kubestellar:
          - Setup KubeStellar:  "setup-ks.md"  (this should include a pointer to "pre-req")
          - Deploy a workflow to multiple clusters:  "deploy-example.md"
  
 Everything else (if at all needed) should be just linked from within those docs.

MikeSpreitzer: My point here is not about where the tree is clipped for menu inclusion, my point is about the same sort of information not collected in one branch of the tree.

I can see an argument for keeping the discussion of pre-reqs as a peer of installation rather than a part of installation. Someone evaluating or planning will need to know the pre-reqs but not particularly care about the actual installation steps.

MikeSpreitzer:
@KPRoche, about the un-linked files: I think that as we work on re-org, at no point should there be a retreat in the net information content of the website that is not intended to be permanent. In the case of a file that you are not sure where to link, I think that you should link it _somewhere_ with a note about needing more work.

KPRoche:
> @KPRoche, about the un-linked files: I think that as we work on re-org, at no point should there be a retreat in the net information content of the website that is not intended to be permanent. In the case of a file that you are not sure where to link, I think that you should link it _somewhere_ with a note about needing more work.

@MikeSpreitzer that is not a bad idea. let me consider how best to label that in the menu

docs/mkdocs.yml
KPRoche: The tree was expanded based on my *excruciating* experience bouncing back and forth between files linked in the old 'common setup" instructions. If there are multiple files with instructions that have to be done in order, those should be in the left menu tree so they are always visible *in order*. 

I had not come to a resolution of how to organize the installation instructions which were part of the examples section vs  those in a getting started section (since at the moment they are versions of the same process).

docs/mkdocs.yml
KPRoche: oops. fixed it.

docs/mkdocs.yml
KPRoche: I have taken out the "under construction" image at the talk, replacing it with a reminder that we encourage contributors to engage via the github repo interface and our Slack channel, and demoted this file and the 3 others that I had commented out into a "contribution notes" subsection.
I've also started linked other older (not properly integrated files) into a miscellaneous docs section.

KPRoche:
Thanks to lots of very patient coaching by @MikeSpreitzer, the current version of this website design is now viewable at
https://kproche.github.io/kubestellar/newdocs/

It will be much easier to discuss now everyone can *see* the structure. (And yes, it still needs work!)
I'm going to be offline until Tuesday morning June 11.

MikeSpreitzer:
Regarding the top level of the navigation menu:
- When I first encounter a new project, my first question is "what is this thing?", not "how do I get started using it?".
- A random reader is going to think that "Getting Started" either means "quickstart" or "general installation instructions" (which is logically part of the broader category of "user instructions" or "user guide"). If we mean the former, then calling it "quickstart" would be clearer. If we mean the latter, then it should be a section of the user guide rather than a peer topic.

MikeSpreitzer:
Regarding quickstart, examples, and install instructions, I have a few suggestions.

- We only need one quickstart. It does not need three variants for how to run it. Unless and until we hear otherwise, I think we can assume that anyone interested in kicking the tires can use `kind` for that purpose.
- The quickstart can be factored into two parts: setup and examples.
- Use two of the existing examples: Scenarios 1 (a baseline) and 4 (singleton status return). Note that Scenario 4 is described as starting from the same state as 1 starts from, but 4 can also be started from the end state of 1. Rewrite both Scenarios so that they can be done in any order (this might already be true) and explicitly call that out.
- The examples are _also_ useful as a way for a user who has done their own setup to validate that it is working. To this end, make the examples parameterized by some shell variable settings. Make the quickstart have three parts: setup, setting those shell variables, then examples that rely on those variables. Explain the variables and that the examples can be used in any environment with appropriate values for those variables.
- The general setup story is part of the user guide, not the quickstart.
- The general setup story is has, after the warning about tearing down the previous trial, three parts: (1) acquire the usage of a Kubernetes cluster with Ingress controller with SSL passthrough, either by creating it (e.g., with `kind` or k3d or your favorite hosting service on your favorite cloud) or by getting a kubeconfig that allows use of a pre-existing cluster and adding the Ingress controller if it is not already there; (2) initialize that cluster as a KubeFlex hosting cluster; (3) define some ITSes and WDSes. Note that ITSes and WDSes can also be added any time later.

KPRoche:
> Regarding quickstart, examples, and install instructions, I have a few suggestions.
>. 
>.
>.
> * Remember that tearing down the setup is a different thing than tearing down one of the example scenarios.

I've started replying to this in your referencing comment in #2163

PR #2188, opened 2024-06-03
✨ Enable website previews

MikeSpreitzer:
## Summary
This PR modifies the website update workflow to also render in the case of a contributor working on a branch whose name starts with "doc-". The rendering is only done if the actor involved in the github action is the repository owner. The rendering goes to the contributor's own github pages site. For example, this PR caused publication to https://mikespreitzer.github.io/kcp-edge-mc/doc-preview-enable/ .

## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/docs-gen-and-push.yml
+++ b/.github/workflows/docs-gen-and-push.yml
@@ -8,6 +8,7 @@ on:
   push:
     branches:
       - "release-*"
+      - "doc-*"
     paths:
       - "docs/**"
       - "*.md"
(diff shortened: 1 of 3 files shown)


KPRoche:
That looks great! /lgtm

.github/workflows/docs-gen-and-push.yml
```diff
@@ -8,6 +8,7 @@ on:
   push:
     branches:
       - "release-*"
+      - "doc-*"
```
ezrasilvera, line 11: Why do we need to restrict to specific branch name?  We already have paths rule, If someone is changing files under docs/ I expect that user to want to see the generated site even if it's not a "doc-"  release

MikeSpreitzer: The idea here is that this rendering does not happen normally, it only happens when the contributor requests it. This saves the costs of doing renderings that have little value. We have been doing many PRs that involve doc updates without people asking for an automatic rendering.

The contributor requests rendering by naming their branch something that starts with "doc-".

`"doc"- release` is not a thing. I am not sure I understand.

Elsewhere you expressed concern over the costs of the rendering. The control exerted here limits that.

Issue #2158, opened 2024-05-14
epic: Prometheus metrics


MikeSpreitzer:
There are several things that we should do about Prometheus metrics.

- [ ] expose `/metrics` from the kubestellar-controller-manager, connected to at least the metrics from the kubernetes libraries.
- [ ] expose `/metrics` from the generic transport controller, connected to at least the metrics from the kubernetes libraries.
- [ ] expose `/metrics` from the OCM Status Add-On Agent, connected to at least the metrics from the kubernetes libraries.
- [ ] add metrics on apiserver request/responses from kubestellar-controler-manager.
- [ ] add metrics on apiserver request/responses from generic transport controller.
- [ ] add mission-specific metrics (e.g., number of bound objects, WECs per Binding[Policy]) to kubestellar-controller-manager. FYI, Issue #959 had some suggestions for the legacy system.
- [ ] add mission-specific metrics (e.g., sizes of wrapped objects) to generic transport controller.
- [ ] add mission-specific metrics (e.g., sizes of wrapped objects) to the OCM Status Add-On Agent.
- [ ] Find or create documentation on metrics from the OCM pods involved.

MikeSpreitzer:
This is not entirely unrelated to #2094

MikeSpreitzer:
Ugh. Of course controller-runtime does it differently. See https://book.kubebuilder.io/reference/metrics.html for some clues.

MikeSpreitzer:
Also, as always, OpenShift has its own opinion about how to get metrics scraped. https://book.kubebuilder.io/reference/metrics.html#exporting-metrics-for-prometheus talks about the provisions for that in controller-runtime. This is only an FYI, scraping is beyond the scope of this epic.

MikeSpreitzer:
Regarding the OCM Status Add-On: although the framework (https://github.com/open-cluster-management-io/addon-framework/tree/v0.8.1) imports controller-runtime, that is only used in testing. The regular code does not use controller-runtime. Currently ks/OSA's controller and agent offer no metrics.

MikeSpreitzer:
I ran the bash E2E test, using commit 5b6b18e26, and scraped the metrics. I have attached the results.

[kubestellar-controller-manager.metrics.txt](https://github.com/user-attachments/files/15570518/kubestellar-controller-manager.metrics.txt)

[ocm-status-addon-agent.metrics.txt](https://github.com/user-attachments/files/15570519/ocm-status-addon-agent.metrics.txt)

[...]

PR #2192, opened 2024-06-07
✨ Add metrics to transport controller

MikeSpreitzer:
## Summary
This PR adds metrics to the generic transport controller. These added metrics are as follows.

- kubestellar_apiserver_call_latency_seconds: histogram of round trip times for calls on apiservers, broken down by: space, resource, method, and shortened error.
- kubestellar_transport_controller_binding_whats: histogram of the number of workload objects referenced in a Binding, observed whenever a Binding is sync'd.
- kubestellar_transport_controller_binding_wheres: histogram of the number of WECs referenced by a Binding, observed whenever a Binding is sync'd.
- kubestellar_transport_controller_binding_areas: histogram of the product of number of workload objects referenced and number of WECs referenced by a Binding, observed whenever a Binding is sync'd.
- kubestellar_transport_controller_bindings: gauge of the number of Binding objects in the ITS.
- kubestellar_transport_controller_prop_maps: gauge of the number of property ConfigMap objects in the WDS.
- kubestellar_transport_controller_transforms: gauge of the number of CustomTransform objects in the WDS.
- kubestellar_transport_controller_wecs: gauge of the number of inventory objects in the ITS.
- kubestellar_transport_controller_wrapped_objects: gauge of the number of wrapped objects (envelopes, ManifestWork objects) in the ITS.

This PR also adds metrics scrapes at the end of the E2E testing workflow.

## Related issue(s)

This addresses part of #2158

### DIFF
--- a/.github/workflows/pr-test-e2e.yml
+++ b/.github/workflows/pr-test-e2e.yml
@@ -83,6 +83,10 @@ jobs:
         if: always()
         run: kubectl --context kind-kubeflex get deployments -A
 
+      - name: show Deployment objects in its1
+        if: always()
+        run: kubectl --context its1 get deployments -A
+
       - name: show kubestellar controller manager log
         if: always()
         run: kubectl --context kind-kubeflex logs -n wds1-system $(kubectl --context kind-kubeflex get pod -n wds1-system --selector=control-plane=controller-manager -o jsonpath='{.items[0].metadata.name}')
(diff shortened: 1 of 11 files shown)


ezrasilvera:
@MikeSpreitzer  Seems to me like too many metrics than we actually need.  But I don't mind adding them for now, if we see an issue we can always remove.

Issue #2200, opened 2024-06-12
bug: Insufficient debug logging from ginkgo E2E test in CI


MikeSpreitzer:
### Describe the bug

I got a failure of the ginkgo E2E test in CI. See https://github.com/kubestellar/kubestellar/actions/runs/9476682004/job/26109961479?pr=2199 . The run of setup-kubestellar.sh suffered a timeout. The CI test log had no indication of **where** in the running of setup-kubestellar.sh the timeout happened.

Here is what I got in the log:

```
Run cd test/e2e/ginkgo 
  cd test/e2e/ginkgo 
  KFLEX_DISABLE_CHATTY=true ginkgo -vv --trace --no-color $TEST_FLAGS
  shell: /usr/bin/bash -e {0}
  env:
    KO_DOCKER_REPO: ghcr.io/kubestellar/kubestellar
    TEST_FLAGS: 
go: downloading github.com/onsi/gomega v1.29.0

Running Suite: Kubestellar testing - /home/runner/work/kubestellar/kubestellar/test/e2e/ginkgo
==============================================================================================
Random Seed: 1718166228

Will run 17 of 17 specs
------------------------------
[BeforeSuite] 
/home/runner/work/kubestellar/kubestellar/test/e2e/ginkgo/ginkgo_suite_test.go:68
  > Enter [BeforeSuite] TOP-LEVEL - /home/runner/work/kubestellar/kubestellar/test/e2e/ginkgo/ginkgo_suite_test.go:68 @ 06/12/24 04:24:27.163
  STEP: Execing command [../common/setup-kubestellar.sh] - /home/runner/work/kubestellar/kubestellar/test/util/util.go:126 @ 06/12/24 04:24:28.008
  [TIMEDOUT] A suite timeout occurred
  In [BeforeSuite] at: /home/runner/work/kubestellar/kubestellar/test/e2e/ginkgo/ginkgo_suite_test.go:68 @ 06/12/24 05:23:48.889

  Full Stack Trace

  This is the Progress Report generated when the suite timeout occurred:
    In [BeforeSuite] (Node Runtime: 59m21.726s)
      /home/runner/work/kubestellar/kubestellar/test/e2e/ginkgo/ginkgo_suite_test.go:68
      At [By Step] Execing command [../common/setup-kubestellar.sh] (Step Runtime: 59m20.881s)
        /home/runner/work/kubestellar/kubestellar/test/util/util.go:126

      Spec Goroutine
[...]

PR #2202, opened 2024-06-13
🐛 Make ginkgo setup interruptable

MikeSpreitzer:
## Summary
This PR attempts to connect ginkgo timeout to interrupting the `exec.Cmd` that runs `setup-kubestellar.sh`, so that a timeout during that script will cause the logging in SetupKubestellar to take effect. This should give a way to understand what went wrong during `setup-kubestellar.sh` when that suffers a timeout.

## Related issue(s)

This is intended to partially address #2200 .


ezrasilvera:
@MikeSpreitzer Is there a missing piece here ?  How does adding the context helping to propagate the interrupts ?

PR #2206, opened 2024-06-13
✨ Introduce doc vars for releases

MikeSpreitzer:
## Summary
This PR introduces some new documentation variables that align with our current release concepts of "latest regular release" and "latest release". This replaces several self-references with just one (the definition of `ks_latest_release`).

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -101,6 +101,8 @@ There are many global variables defined in the <a href="{{ config.repo_raw_url }
     - edit_uri: {{ config.edit_uri }}
     - ks_branch: {{ config.ks_branch }}
     - ks_tag: {{ config.ks_tag }}
+    - ks_latest_regular_release: {{ config.ks_latest_regular_release }}
+    - ks_latest_release: {{ config.ks_latest_release }}
 
 to use a variables/macro in your documentation reference like this:
 
--- a/docs/content/direct/release.md
+++ b/docs/content/direct/release.md
@@ -38,15 +38,11 @@ Making a new kubestellar release requires a contributor to do the following thin
 
 - If not already in effect, declare a code freeze. There should be nothing but bug fixes and doc improvements while working towards a regular release.
 
+- Edit `docs/mkdocs.yml` and update the definition of `ks_latest_release` to `$version` (e.g., `'0.23.0-rc42'`). If this is a regular release then also update the definition of `ks_latest_regular_release`.
+
 - Edit the source for the KCM PCH (in `config/postcreate-hooks/kubestellar.yaml`) and update the tag in the reference to the KCM container image (it appears in the last object, a `Job`).
 
-- Edit the examples document(`docs/content/direct/examples.md`) to update the self-references for the coming release. In the interim while we have some content split out into `docs/content/direct/common-setup-core-chart.md` and `docs/content/direct/common-setup-step-by-step.md`, edit in those places.
-
 - Update the version in the core chart defaults, `core-chart/values.yaml`.
-
-- Update the version every place it appears in the core chart instructions, `docs/content/direct/core-chart.md`. Eventually we may settle on a regular release for some curl-to-bash invocations.
-
-- Update the microshift directions, `docs/content/direct/microshift.md`.
 
 - Until we have our first stable release, edit the docs README(`docs/content/direct/README.md`, section "latest-stable-release") where it wishes it could cite a stable release but instead cites the latest release, to rever to the coming release.
 
--- a/docs/content/direct/common-setup-core-chart.md
+++ b/docs/content/direct/common-setup-core-chart.md
@@ -8,7 +8,7 @@ with a choice of user-defined KubeFlex Control Planes (CPs).
 The information provided is specific for the following release:
 
 ```shell
-export KUBESTELLAR_VERSION=0.23.0
+export KUBESTELLAR_VERSION={{ config.ks_latest_release }}
 export OCM_TRANSPORT_PLUGIN=0.1.11
 ```
 
--- a/docs/content/direct/common-setup-step-by-step.md
+++ b/docs/content/direct/common-setup-step-by-step.md
@@ -9,7 +9,7 @@ The following steps create a deployment of KubeStellar Core
 1. Set environment variables to hold KubeStellar and OCM-transport-plugin desired versions:
 
     ```shell
-    export KUBESTELLAR_VERSION=0.23.0
+    export KUBESTELLAR_VERSION={{ config.ks_latest_release }}
     export OCM_TRANSPORT_PLUGIN=0.1.11
     ```
 
--- a/docs/content/direct/core-chart.md
+++ b/docs/content/direct/core-chart.md
@@ -6,7 +6,7 @@ with a choice of user-defined KubeFlex Control Planes (CPs).
 The information provided is specific for the following release:
 
 ```shell
-export KUBESTELLAR_VERSION=0.23.0
+export KUBESTELLAR_VERSION={{ config.ks_latest_release }}
 ```
 
 ## Pre-requisites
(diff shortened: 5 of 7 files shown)


MikeSpreitzer:
Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-release-vars/

ezrasilvera:
@MikeSpreitzer I'm a little bothered with the "latest-regular-release"  Vs "latest-release" . Why do we need references for  release candidates ?   It also implies that the "latest-release" is somehow "irregular" or "special" ?  
Maybe we should adopt the "stable"  ?  For us, we always said that for now regular==stable (as we don't have stable tests for "stable" )  - I think that having "latest" Vs "latest-stable"  is better

eaepstein:
I thought KS already has two release types: official releases and RC releases, and latest simply points at the most recent official release. Is that no longer the case? Is latest-regular another name for "latest non-RC release"?

MikeSpreitzer:
@ezrasilvera: @eaepstein guessed the answer. Because candidate releases (RCs) are releases, we have established the term "regular" release to distinguish a release that is not an RC from one that _is_ an RC (or alpha or anything else irregular). Thus: "latest regular release" is the latest release that is not an RC (or alpha or...), while "latest release" is the latest release of any sort (including RC or alpha or ...). This is the terminology that we discussed earlier and have been using.

The self references need to refer to _self_ --- even when that self is an RC or alpha --- so the most commonly used variable is `ks_latest_release`.

eaepstein:
What is the difference between RC and alpha releases? May there could just be "releases" and RC-releases", with "latest" pointing at the most recent "release"?

MikeSpreitzer:
"alpha" is something that @francostellari cajoled me into doing. It is a development snapshot that falls short of the goals for the next regular release. @francostellari's new Helm chart can only use a release, so we needed a release of some sort in order to be able to test and continue developing the Helm chart.

A candidate release is one that, as far as we know, meets the goals for the next regular release.

It will be linguistically confusing if we say that an "RC-release" is _not_ a "release".

Do we really need to re-litigate the terminology right now, as we are trying to get the next regular release out?

MikeSpreitzer:
@eaepstein: The `mike` alias named "latest" points to the latest regular release. And the `mike` version warning banner says "latest release" where it means "latest _regular_ release". This is admittedly inconsistent with the rest of our terminology. This is the result of our historical development path and general hurry.

eaepstein:
>Do we really need to re-litigate the terminology right now, as we are trying to get the next regular release out?
Seems like "alpha" and "regular" are terms introduced without discussion.

MikeSpreitzer:
@eaepstein: I definitely remember discussing "regular". The "alpha" was something that we had earlier, then agreed that we do not need any more, then was re-introduced in a rush with only a few people involved.

ezrasilvera:
I believe we agreed that Latest should always point to a "regular" release.   there is no such thing as "latest RC" or "latest Alpha"

MikeSpreitzer:
@ezrasilvera: you may be referring to the `mike` alias named "latest". We have a terminology snafu here. The `mike` alias "latest" points to the "latest regular release", as you observed. If we were completely consistent then we would rename that `mike` alias from "latest" to "latest_regular". The name of this alias was established long ago, before our current thinking about releases, and nobody has bothered to change it. Also, the `mike` banner warning about version is also outdated, speaking of "latest release" when it means "latest regular release".

I really do not think we want to have to tell people that "a candidate release is not a release". We needed an explicit word to distinguish candidate releases from non-candidate releases. That is why we agreed to call the non-candidate releases "regular". This was when we expected no "alpha" releases. Now that we also have had some "alpha" releases, this still does not change the need for an explicit word to distinguish the regular releases from the other kinds of releases.

The mkdocs variables that we define are not connected to `mike` aliases. I chose to introduce variables consistent with our latest thinking and terminology. If anything, I think that the outdated stuff in `mike` should be updated.

MikeSpreitzer:
In plain English, since we have candidate releases, one of them is always the latest. So we _do_ have a latest candidate release. Same for alpha. In plain English, the latest release would be the latest release of any kind.

MikeSpreitzer:
The force-push to c1c306125 is a rebase onto `main`.

MikeSpreitzer:
The force-push to 533216965 is a rebase onto `main`.

docs/content/direct/release.md
```diff
@@ -38,16 +38,12 @@ Making a new kubestellar release requires a contributor to do the following thin
 
 - If not already in effect, declare a code freeze. There should be nothing but bug fixes and doc improvements while working towards a regular release.
 
-- Edit the source for the KCM PCH (in `config/postcreate-hooks/kubestellar.yaml`) and update the tag in the reference to the KCM container image (it appears in the last object, a `Job`).
+- Edit `docs/mkdocs.yml` and update the definition of `ks_latest_release` to `$version` (e.g., `'0.23.0-rc42'`). If this is a regular release then also update the definition of `ks_latest_regular_release`.
```
ezrasilvera, line 41: As we discussed elsewhere - I think we have a notion of "latest" only for regular releases.

docs/README.md
```diff
@@ -101,6 +101,8 @@ There are many global variables defined in the <a href="{{ config.repo_raw_url }
     - edit_uri: {{ config.edit_uri }}
     - ks_branch: {{ config.ks_branch }}
     - ks_tag: {{ config.ks_tag }}
+    - ks_latest_regular_release: {{ config.ks_latest_regular_release }}
+    - ks_latest_release: {{ config.ks_latest_release }}
```
ezrasilvera, line 105: I suggest we converge to only use one latest (of a regular release) .   All references except one are using the latest of regular release in any case.

MikeSpreitzer: No, most of the uses are for the self-references that need to be to the release being made regardless of what kind of release it is.

KPRoche:
Just a note that the "latest release" message is in the main.html file and therefore easily edited:
(this is from [kproche.github.io/](https://kproche.github.io/kubestellar/doc-new-v2b/) )
image

docs/content/direct/release.md
MikeSpreitzer: To move the discussion into a public forum, created a GitHub "discussion". See https://github.com/kubestellar/kubestellar/discussions/2218 .

docs/README.md
ezrasilvera: @MikeSpreitzer (@eaepstein @pdettori ) I don't agree.  I think that at this stage of the project (it may be changed in the future) we want users to use only "regular release"  . The other releases are point releases aimed for developers for specific task (test a fix, test a new feature, etc..) 
As our web-site is aimed to users - all references to 1latest1 on the web site should refer to the latest `regular release`.   
The few developers who use/test the RCs don't need to use direct references for theses RCs on our web documentation in order to use it. 
BTW, I believe that most of the team actually supported to even  generate the docs only for `regular releases`  , and not for other intermediate releases, but we didn't vote on that and just let it continue :(

eaepstein: >The other releases are point releases aimed for developers for specific task (test a fix, test a new feature, etc..)

Does this translate to mean 0.23.0 is a "regular" release but 0.23.1 is not? And therefor 0.23.1 cannot be labeled as "latest" ?

MikeSpreitzer: @ezrasilvera:

> @MikeSpreitzer (@eaepstein @pdettori ) I don't agree. I think that at this stage of the project (it may be changed in the future) we want users to use only "regular release" .

I agree with that.

The problem is all the self-references that have to be updated when making a release of any kind, including a test (candidate) release. See https://github.com/kubestellar/kubestellar/blob/v0.23.0/docs/content/direct/release.md#making-a-new-kubestellar-release . Most of those self-references are in doc files. I introduced `config.ks_latest_release` to be used in place of those self-references, so that making a release of any kind (including a test release) requires editing only the definition of `config.ks_latest_release` instead of all the places where it appears in the docs.

MikeSpreitzer: @eaepstein: I think that when @ezrasilvera wrote about "point releases" he did not mean "point" in the technical semver sense of major.minor.patch; rather, he meant "point" in the sense of "intended for a narrow purpose". A release identified simply as "major.minor.patch" _is_ a regular release, not a test (candidate) release or alpha or beta or other unusual thing.

MikeSpreitzer: @ezrasilvera: can we separate a change to our policy of which releases appear on the website from this PR?

ezrasilvera:
@MikeSpreitzer Let's merge this and continue the discussion elsewhere 
/lgtm
/approve

PR #2220, opened 2024-06-20
:bug: Update the "new bug" template

ezrasilvera:
Remove the old instructions for specific KS info collection, these instructions are not suitable for the new KS. For now, the general template asking for bug description, logs, and expected behavior is good enough.

## Related issue(s)

Fixes #

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -11,34 +11,6 @@ body:
         
         Here is how you can help us serve you better:
 
-        0. Did you look at issue #920 re: potentially misusing the ‘kubectl ws’ command?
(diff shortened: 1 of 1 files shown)


clubanderson (commented):
/lgtm

JonathanWentaoBao:
This should fix issue #2224

MikeSpreitzer (commented):
I think that we should retain some remarks about collecting the relevant versions.
I think that we should retain some remarks about collecting kubeconfig state. We still have a system that is sensitive to obscure user state: the value of the `KUBECONFIG` envar, the current kubeconfig context.

MikeSpreitzer (approved):
Yes, we can restore the needed content in a follow-on PR.

/lgtm
/approve
Thanks!

Issue #2230, opened 2024-06-20
:bug: old version referenced by k3d instructions


clubanderson:
### Describe the bug

https://docs.kubestellar.io/release-0.23.0/direct/deploy-on-k3d/ references 0.22 of KS

### Steps To Reproduce

1. go to https://docs.kubestellar.io/release-0.23.0/direct/deploy-on-k3d/
2. observe 0.22 ks is referenced

not sure if other references are correct either.

### Expected Behavior

0.23 doc should point to 0.23 ks components

MikeSpreitzer:
That doc says right up front in the introductory material:

> This is a point-in-time statement that worked with KubeStellar release v0.22.0.

Maybe it could be worded more clearly. It is trying to say that that particular document is not about the latest release but rather an older one. It is trying to say that we do not intend to update that document with every release. That was the outcome of a discussion about testing, in which it was asserted with some vehemence that we would not have the capacity to do that much manual testing.

clubanderson:
Ugg. What a mess. Close if you prefer. I don't know what else to do. This has become untenable.

MikeSpreitzer:
I think that the re-org should remove the need for this file. In the meantime I will offer a PR to make the warning more explicit.

clubanderson:
documentation no longer exists. closing

PR #2229, opened 2024-06-20
:book: Implement new website layout

KPRoche:
## This replaces the existing kubestellar.io website with the new structure
This is a replacement for PR #2183. It incorporates many of the suggestions made there, as well as in other discussions

Preview at https://kproche.github.io/kubestellar/doc-new-v2

## Related issue(s)
#2163, #2228, #2221, #2214, #2216, #2222
Fixes #

PR #2233, opened 2024-06-21
📖 [release-0.23.0] Strengthen warning in k3d report

MikeSpreitzer:
## Summary
This PR updates the introductory material of the report of how KubeStellar was deployed using k3d, to emphasize that it is a blast from the past and not referencing the latest regular release.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-warn-harder/direct/deploy-on-k3d/

## Related issue(s)

I hope this will fix issue #2230 in branch `release-0.23.0`, which is the source of the default version of the website currently.

I hope that the greater re-organization work being done in `main` will dissolve the problematic file and put the important pieces elsewhere.

### DIFF
--- a/docs/content/direct/deploy-on-k3d.md
+++ b/docs/content/direct/deploy-on-k3d.md
@@ -1,8 +1,8 @@
 # Deploy KubeStellar on K3D
 
-This document shows how to deploy kubestellar on K3D hub and wec clusters.
+This document shows how KubeStellar release 0.22.0 was set up using feshly created k3d clusters for the hosting cluster and two WECs.
 
-This is a point-in-time statement that worked with KubeStellar release v0.22.0.
+**NOTE WELL:** This is a report of a something was done in the past with an older release of KubeStellar; 0.22.0 is NOT the latest release.
 
 ## Prereqs
 


clubanderson:
I dont see the need for this. If it "worked" before then it should work now. This page creates a requirement for 0.22 to be supported instead of sunsetted. The real solution is to add a test for this workflow. If docs-ecutable were running, that would be easy. Can gingko do this for you?  I think this page is a bad idea if you can't keep it current.

ezrasilvera:
@MikeSpreitzer I think we need another approach for the supported platforms (KIND, k3d, host, etc..)
1. If we only show an example of something we tested on specific release - this should go into `integration examples`  or `ks on k3d tutorial` 
2. If we want a platform to be officially supported, it need to be added to the regular testing. This shouldn't be too difficult once we break the flows to `setup` and then `run test`

MikeSpreitzer:
Trying to put myself in a user's shoes, I wrote what I think the documentation should look like in https://github.com/kubestellar/kubestellar/pull/2229#discussion_r1649126413 . I am focused on users who are developers of a larger system that will include KubeStellar; their needs can not be met with documentation that consists only of three soup-to-nuts recipes. We need to document the general ideas, as well as show concrete examples of individual steps and usage of individual features.

I also agree with the concern that documentation should be regularly tested. That gets a little challenging when the documentation is not just a collection of soup-to-nuts recipes. I think that we will have to settle for coverage of the other documentation by tests whose relationship to the documentation is maintained by hand by us.

MikeSpreitzer:
See #22237 instead.

PR #2234, opened 2024-06-21
📖 [release-0.23.0] Fix broken cross-ref to CONTRIBUTING doc

MikeSpreitzer:
## Summary
This PR replaces a broken absolute link to a document with a working relative link.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-unmain/readme/#contributing

## Related issue(s)

This should fix #2232 in the branch named `release-0.23.0`, which is the one that is currently the source of the default version of the website.

PR #2237, opened 2024-06-21
📖 [release-0.23.0] Remove the report of deploying on k3d

MikeSpreitzer:
## Summary
This PR deletes the report of how to run KubeStellar release 0.22.0 on three new k3d clusters. This is good because nobody wanted to maintain it and nobody wanted to see a non-maintained document. This is bad because it is deleting information that took time to develop and we do not want to lose. The good news is that Git remembers and we have it published at https://docs.kubestellar.io/release-0.22.0/direct/deploy-on-k3d/ , and with all the stress over it, we are not likely to soon forget that we had it.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-delete-k3d-report/readme/

## Related issue(s)

This is one way to resolve issue #2230 in one specific branch, `release-0.23.0` --- which is the one that currently drives the default version of the website. PR #2234 is another. I hope that the website work in `main` will distribute the interesting content from this report into the user guide material about how to set up KubeStellar.


ezrasilvera:
@MikeSpreitzer As I wrote in the other PR - I think this should be presented in all versions, however it should be put as an "example integration"  with specific version pointer.

PR #2247, opened 2024-06-24
📖 Update the outline of KubeStellar usage

MikeSpreitzer:
## Summary
This PR updates the outline of KubeStellar usage, including the re-org of installation and scenario docs that I outlined at https://github.com/kubestellar/kubestellar/pull/2229#discussion_r1649126413 .

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-outline10/direct/usage-basics/ .


KPRoche:
After the conversations earlier today, I wonder if in the "before you begin" section we should flip excerpts from pre-reqs.md and mention the check_prereqs.sh script before the paragraphs about required software?

KPRoche:
> After the conversations earlier today, I wonder if in the "before you begin" section we should flip excerpts from pre-reqs.md and mention the check_prereqs.sh script before the paragraphs about required software?

Or even more specifically, recommend

`check_pre_req.sh docker kubectl kflex ocm helm kind`

 as the command to check for the components required to run the quickstart example.

MikeSpreitzer:
I think the prereqs check script is a little too terse, the explanation needs to come first.

KPRoche:
> I think the prereqs check script is a little too terse, the explanation needs to come first.

Fair enough; we might want to decide just how much to include from the file, or perhaps instead offer the explicit command in the quickstart document and in then case link out to that section of the prereqs md file for more information if the reader wants it (rather than do the current include).

MikeSpreitzer:
Yes, I think that prereq stuff in the quickstart can be tweaked. I would prefer to do it in follow-on work.

KPRoche:
> Yes, I think that prereq stuff in the quickstart can be tweaked. I would prefer to do it in follow-on work.

I agree that should be follow-on

PR #2252, opened 2024-06-25
📖 Add grouping and extension points in usage outline graph

MikeSpreitzer:
## Summary
This PR updates the usage outline doc to include a second graph that adds to the first graph two things: (1) a box showing what the Core Helm chart covers, to emphasize that real usage does not require the user to go through all the steps individually, and (2) explicit call-outs of all the entry points for extending usage, to make it clear that the user has all these choices and why the graph is so articulated.

The base and enhanced graphs are drawn with Mermaid, whose results are not greatly satisfactory to me. I am particular not fond of all the unnecessary line-crossing, because it makes the graph unnecessarily complex visually.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-group-outline/direct/usage-basics/

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/direct/usage-basics.md
+++ b/docs/content/direct/usage-basics.md
@@ -23,9 +23,6 @@ By "maintain" we mean create, read, update, delete, list, and watch as you like,
 There is some flexibility in the ordering of those stage. The following flowchart shows the dependencies.
 
 ```mermaid
----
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
See #2253 instead.

PR #2258, opened 2024-06-26
📖 Simplify quickstart wrt prereqs, scenarios

MikeSpreitzer:
## Summary
This PR tweaks the quickstart in a few ways:

- reduce the volume of prereq material;
- focus on just one usage scenarios;
- include the instruction to configure `kind` for more than two clusters.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-tweak-quickstart/direct/get-started/

## Related issue(s)

None

### DIFF
--- a/docs/content/direct/get-started.md
+++ b/docs/content/direct/get-started.md
@@ -1,34 +1,26 @@
 # KubeStellar Quickstart Setup
 
-This Quick Start outlines step 1 and shows a concrete example of steps 2--7 in the [Installation and Usage outline](usage-basics.md). In this example you will create three new `kind` clusters to serve as your KubeFlex hosting cluster and two WECs.
+This Quick Start outlines step 1, shows a concrete example of steps 2--7 in the [Installation and Usage outline](usage-basics.md), and forwards you to one example of the remaining steps. In this example you will create three new `kind` clusters to serve as your KubeFlex hosting cluster and two WECs.
 
-  1. Before you begin, prepare your system (get the software prerequisites)
-  2. Create the KubeFlex hosting cluster and Kubestellar core components
-  3. Create and register two WECs.
-
----
-## Before You Begin
-
-
-{%
-    include-markdown "pre-reqs.md"
-    rewrite-relative-urls=true
-    start="<!-- begin software prerequisites -->"
-    end="Additional Software For Running"
-    heading-offset=1
-%}
----
-{%
-    include-markdown "pre-reqs.md"
-    rewrite-relative-urls=true
-    heading-offset=2
-    start="<!-- start tag for check script  include -->"
-    end="<!-- end tag for check-prereq script -->"
-%}
+  1. Install software prerequisites
+  1. Cleanup from previous runs
+  1. Create the KubeFlex hosting cluster and Kubestellar core components
+  1. Create and register two WECs.
+  1. Use KubeStellar to distribute a Deployment object to the two WECs.
 
 ---
+## Install software prerequisites
 
-### Delete debris from preivous trials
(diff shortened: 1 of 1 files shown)


docs/content/direct/get-started.md
```diff
@@ -1,32 +1,23 @@
...
-%}
+```shell
+bash <(curl https://raw.githubusercontent.com/kubestellar/kubestellar/v{{ config.ks_latest_regular_release }}/hack/check_pre_req.sh) kflex ocm helm kubectl docker kind
+```
 
----
+This quickstart uses [kind](https://kind.sigs.k8s.io/) to create three Kubernetes cluster on your machine.
+Note that `kind` does not support three or more concurrent clusters unless you raise some limits as described in this `kind` "known issue": [Pod errors due to “too many open files”](https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files).
 
 ### Delete debris from preivous trials
```
ezrasilvera, line 22: Not part of this PR - but I would move that to a special section at the end "Cleanup"

docs/content/direct/get-started.md
```diff
@@ -1,32 +1,23 @@
...
-This Quick Start outlines step 1 and shows a concrete example of steps 2--7 in the [Installation and Usage outline](usage-basics.md). In this example you will create three new `kind` clusters to serve as your KubeFlex hosting cluster and two WECs.
+This Quick Start outlines step 1, shows a concrete example of steps 2--7 in the [Installation and Usage outline](usage-basics.md), and forwards you to one example of the remaining steps. In this example you will create three new `kind` clusters to serve as your KubeFlex hosting cluster and two WECs.
 
   1. Before you begin, prepare your system (get the software prerequisites)
   2. Create the KubeFlex hosting cluster and Kubestellar core components
   3. Create and register two WECs.
+  4. Use KubeStellar to distribute a Deployment object to the two WECs.
 
 ---
 ## Before You Begin
```
ezrasilvera, line 11: Can't we just call it "Install pre-reqs" and move the other paragraph to a separate "Cleanup" section ?

docs/content/direct/get-started.md
MikeSpreitzer: I think that putting the cleanup at the end is bad for two reasons.

Tearing down the state as soon as the user finishes going through the written recipe is bad because it deprives the user of the ability to examine the state that the recipe established and do further experiments with it.

Granted, the user might figure that out and put off the cleanup. In this case, it is not unlikely that the user will forget about the cleanup by the time they go to start over again. I have done this, and I know that others have done this.

docs/content/direct/get-started.md
MikeSpreitzer: The paragraph about configuring `kind` to work with more than two clusters is not a matter of cleanup.

ezrasilvera: Sorry, My comment was on the wrong line :(     
I meant change the "Before you begin" to:  Cleanup and install pre-reqs   and also change the list item from "Before you begin, prepare your system (get the software prerequisites)"   to something short (maybe also Cleanup and install pre-reqs)

docs/content/direct/get-started.md
ezrasilvera: Sounds reasonable.  My comment was mainly about the sections and titles .  See the second comment which basically covers the same thing.
I think that "Delete debris from previous trials"  and "Before you begin, prepare your system (get the software prerequisites)"  can be simplified ...

MikeSpreitzer: Revised.

docs/content/direct/get-started.md
MikeSpreitzer: Revised.

PR #2270, opened 2024-06-27
📖 Apply meat cleaver to documentation

MikeSpreitzer:
## Summary
This PR deletes all the documents that we have been retaining because they have information that we do not want to lose but have not been properly integrated into the new outline. This is good because it cause searches to stop returning pages that have not been properly edited. This is bad because it will lose information, but not too bad because Git remembers.

This PR also updates a couple of docs with better context setting words.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-chop/

## Related issue(s)

Fixes #


MikeSpreitzer:
To be clear, this PR represents a shift in attitude about documentation: rather than it being a work in progress, we are shifting it to be a safe space where it is hard to find anything misleading or wrong.

MikeSpreitzer:
See #2271 for an alternative that stashes the unready docs for future scavenging.

PR #2276, opened 2024-06-28
📖 Fix broken links

MikeSpreitzer:
## Summary
This PR fixes some more broken links in the documentation, found now that the broken links crawler is working well enough.

Two of the problems are due to the fact that GOVERNANCE.md is symlinked, which means that relative links do not work from both locations. The solution was to use an absolute link to the website version named "main", which we have retracted. I have changed the solution to using relative links and settling for the fact that they only work when viewed from inside the docs dir.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-unbreak-628b/readme/

## Related issue(s)

Fixes #


KPRoche (commented):
/lgtm

MikeSpreitzer:
/hold
while testing improvements to the broken links crawler --- so that it has something to find.

MikeSpreitzer:
The current broken links crawler work is done.
/remove-hold

PR #2285, opened 2024-07-01
✨ Status controller with redesigns

waltforme:
## Summary
This PR changes the overall design of the status controller's handling of singleton status, from 'edge-driven' to 'edge-triggered, level-driven', thus fixes an eventual consistency bug and kills other potential bugs before they really appear.

Some initial thoughts and discussions are documented [here](https://docs.google.com/document/d/1KD0S79cxqDp2mD55FQ4BGeWtxEN2TM6g9uiiSa-LO80).


waltforme:
Force-push to https://github.com/kubestellar/kubestellar/commit/b7a8cf903ed1640406f9acdb0c03d7edb6b46a5f was
1. cherry-picking of code on top of latest (after merging of work on combined status) `main`;
2. changes to make code pass all e2e test cases.

waltforme:
With #2356 merged, we can now check the eventual consistency of the redesigned status controller.

First, I changed the expectation of the test in https://github.com/kubestellar/kubestellar/pull/2285/commits/425af149f6cd1e1fb878e1de93447ba1635d75bf. Then, the test [succeeded](https://github.com/kubestellar/kubestellar/actions/runs/10168273756/job/28122505399?pr=2285#step:7:210).

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,167 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	"github.com/kubestellar/kubestellar/pkg/util"
+)
+
+// Should be wr-locked
```
MikeSpreitzer, line 33: Is that "should" as in
- "this is something that would be true in an ideal world but is not true in this world now", or
- "this is something that I think is true but I have not fully checked", or
- "this is something that somebody else is responsible for making true"?

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (ss *singletonState) Get(id util.ObjectIdentifier) bool {
+	return ss.wObjSync[id]
+}
+
+func (ss *singletonState) Set(id util.ObjectIdentifier) {
+	ss.wObjSync[id] = true
+}
+
+func (ss *singletonState) Unset(id util.ObjectIdentifier) {
```
MikeSpreitzer, line 46: What deletes entries from this map?
If an entry is never deleted then this is a memory leak.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,167 @@
...
+	ss.wObjSync[id] = false
```
MikeSpreitzer, line 47: What deletes entries from this map?
If an entry is never deleted then this is a memory leak.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// Should it be a field of Contoller?
```
MikeSpreitzer, line 50: I think so.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,167 @@
...
+			labels := nsWObj.(metav1.Object).GetLabels()
+			if v, ok := labels[util.BindingPolicyLabelSingletonStatusKey]; ok {
+				logger.V(2).Info("Found namespace scoped singleton workload object", "name", nsWObjRef.Name)
+				wObjIdentifier := util.ObjectIdentifier{
+					GVK:        schema.GroupVersionKind{Group: nsWObjRef.Group, Version: nsWObjRef.Version, Kind: nsWObj.GetObjectKind().GroupVersionKind().Kind},
+					Resource:   nsWObjRef.Resource,
+					ObjectName: cache.NewObjectName(nsWObjRef.Namespace, nsWObjRef.Name),
+				}
+				if v == util.BindingPolicyLabelSingletonStatusValueSet && len(bdg.Spec.Destinations) > 0 {
+					logger.V(2).Info("Singleton workload object should has status synced", "name", nsWObjRef.Name)
```
MikeSpreitzer, line 86: s/has/have/

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+			labels := nsWObj.(metav1.Object).GetLabels()
+			if v, ok := labels[util.BindingPolicyLabelSingletonStatusKey]; ok {
+				logger.V(2).Info("Found namespace scoped singleton workload object", "name", nsWObjRef.Name)
+				wObjIdentifier := util.ObjectIdentifier{
+					GVK:        schema.GroupVersionKind{Group: nsWObjRef.Group, Version: nsWObjRef.Version, Kind: nsWObj.GetObjectKind().GroupVersionKind().Kind},
+					Resource:   nsWObjRef.Resource,
+					ObjectName: cache.NewObjectName(nsWObjRef.Namespace, nsWObjRef.Name),
+				}
+				if v == util.BindingPolicyLabelSingletonStatusValueSet && len(bdg.Spec.Destinations) > 0 {
```
MikeSpreitzer, line 85: Shouldn't this be `== 1` rather than `> 0` ?

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,167 @@
...
+					logger.V(2).Info("Singleton workload object should has status synced", "name", nsWObjRef.Name)
+					theSingletonState.Set(wObjIdentifier)
+				} else {
+					logger.V(2).Info("Singleton workload object should not have status synced", "name", nsWObjRef.Name)
+					theSingletonState.Unset(wObjIdentifier)
+				}
+				// reconcile should be out of the for loop over all bindings
```
MikeSpreitzer, line 92: I think this "should" means "would be in an ideal world but is not true here (now)". If I have this right then this deserves a "TODO" marking.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	return
+}
+
+func (c *Controller) reconcileSingletonByWObj(ctx context.Context, wObjID util.ObjectIdentifier) error {
+	logger := klog.FromContext(ctx)
+
+	if !theSingletonState.Get(wObjID) {
+		emptyStatus := make(map[string]interface{})
+		return updateObjectStatus(ctx, &wObjID, emptyStatus, c.listers, c.wdsDynClient)
```
MikeSpreitzer, line 118: I think this is too pushy, it will overwrite contents of the Status section that were written by other controllers. Instead this should consult the `ObjectMeta.ManagedFields` to find the `Status` contents that were written by this controller and delete only those.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	wsNameSuffix := strings.Join([]string{
+		wObjID.ObjectName.Namespace,
+		wObjID.ObjectName.Name,
+	}, "-")
+	// A (probably critical) question: How do I know which WEC is selected?
+	list, _ := c.workStatusLister.ByNamespace("cluster1").List(labels.Everything())
```
MikeSpreitzer, line 126: `"cluster1"` ?

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,167 @@
...
+	for _, obj := range list {
+		wsName := obj.(metav1.Object).GetName()
+		if !strings.HasSuffix(wsName, wsNameSuffix) {
```
MikeSpreitzer, line 129: Where is the construction of WorkStatus object names documented?
It looks to me like the current practice has the problems of possible name collisions and length overflows.
It would be better if this client code got its job done without depending on how names are constructed.
Why not look at the source object reference in the spec of the WorkStatus?

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,167 @@
...
+	logger.V(2).Info("Reconciling singleton status by binding", "name", string(ref))
+	return c.buildSingletonStateAndOptionallyReconcile(ctx, true)
+}
+
+func (c *Controller) reconcileSingletonByWs(ctx context.Context, ref singletonWorkStatusRef) error {
+	logger := klog.FromContext(ctx)
+	obj, _ := c.workStatusLister.ByNamespace(ref.wecName).Get(ref.name)
+	status, _ := util.GetWorkStatusStatus(obj)
+	if sync, ok := theSingletonState.wObjSync[ref.sourceObjectIdentifier]; !ok {
+		logger.V(2).Info("Not a singleton workload object, aborting", "sync", sync)
```
MikeSpreitzer, line 158: Someone reading the debug log is going to wonder _what_ is being aborted. I think this message would be clearer if the ", aborting" were removed.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,226 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	"github.com/kubestellar/kubestellar/pkg/util"
+)
+
+type singletonState struct {
+	sync.RWMutex
+	wObjSync map[util.ObjectIdentifier]bool
```
MikeSpreitzer, line 37: Why is this not just a `sets.Set[util.ObjectIdentifier]` ?

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,226 @@
...
+				return err
+			}
+			labels := nsWObj.(metav1.Object).GetLabels()
+			if v, ok := labels[util.BindingPolicyLabelSingletonStatusKey]; ok {
+				wObjIdentifier := util.ObjectIdentifier{
+					GVK:        schema.GroupVersionKind{Group: nsWObjRef.Group, Version: nsWObjRef.Version, Kind: nsWObj.GetObjectKind().GroupVersionKind().Kind},
+					Resource:   nsWObjRef.Resource,
+					ObjectName: cache.NewObjectName(nsWObjRef.Namespace, nsWObjRef.Name),
+				}
+				logger.V(2).Info("Found namespace scoped singleton workload object", "resource", wObjIdentifier.Resource, "objectName", wObjIdentifier.ObjectName)
```
MikeSpreitzer, line 120: I do not favor having two log messages per object. The `if` statement right below emits one log statement, let it carry all the interesting information.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,226 @@
...
+	delete(ss.wObjSync, id)
+}
+
+func (c *Controller) buildSingletonStateAndOptionallyReconcile(ctx context.Context, reconcile bool) error {
+	logger := klog.FromContext(ctx)
+	logger.V(2).Info("Building the desired state for singleton statuses")
+
+	for wObjIdentifier := range c.singletonState.wObjSync {
+		gvr := schema.GroupVersionResource{Group: wObjIdentifier.GVK.Group, Version: wObjIdentifier.GVK.Version, Resource: wObjIdentifier.Resource}
+		lister, found := c.listers.Get(gvr)
```
MikeSpreitzer, line 83:
```suggestion
		lister, found := c.listers.Get(wObjIdentifier.GVR())
```

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,226 @@
...
+		if !found {
+			return fmt.Errorf("could not get lister for gvr: %s", gvr)
+		}
+		_, err := lister.ByNamespace(wObjIdentifier.ObjectName.Namespace).Get(wObjIdentifier.ObjectName.Name)
```
MikeSpreitzer, line 87: What if this GVR is not namespaced?

pkg/status/singletonstatus.go
MikeSpreitzer, line 120: What do you suppose "singleton" means?
We do not currently have a glossary. I have been using it to mean "is distributed to exactly 1 WEC".

MikeSpreitzer:
Don't forget about https://docs.kubestellar.io/unreleased-development/direct/architecture/#status-controller .

MikeSpreitzer:
I do not understand why a bit of status controller internal state is externalized in the workload object's `managed-by.kubestellar.io/singletonstatus` label. It sure makes the design complicated. And this complicated design is undocumented.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,226 @@
...
+		logger.V(2).Info("Updating singleton status", "workstatus", wsRef.Name, "wecName", wsRef.WECName)
+		return updateObjectStatus(ctx, &wsRef.SourceObjectIdentifier, status, c.listers, c.wdsDynClient)
+	}
+	return nil
+}
+
+func (c *Controller) reconcileSingletonByBdg(ctx context.Context, ref bindingRef) error {
+	logger := klog.FromContext(ctx)
+	logger.V(2).Info("Reconciling singleton status by binding", "name", string(ref))
+	return c.buildSingletonStateAndOptionallyReconcile(ctx, true) // TODO: differentiate full build and incremental change
```
MikeSpreitzer, line 207: Yeah. It is not only a gross matter of efficiency, it is also a matter of correctness: this line can be run concurrently for multiple different values of `ref bindingRef` (which it discards because it does a full build). And further, each full build finishes by reconciling for a bunch of workload objects.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	list, _ := c.workStatusLister.ByNamespace("").List(labels.Everything())
+	for _, obj := range list {
+		sourceRef, err := util.GetSourceRefFromWorkStatus(obj)
+		if err != nil {
+			return err
+		}
+		sourceObjID := util.ObjectIdentifierFromSourceRef(sourceRef)
+		if !reflect.DeepEqual(wObjID, sourceObjID) {
```
MikeSpreitzer, line 187: What is wrong with using `!=` here?

pkg/util/unstructured.go
```diff
@@ -170,8 +171,42 @@ func SourceRefFromObjectIdentifier(objIdentifier ObjectIdentifier) *SourceRef {
...
+			Group:   sourceRef.Group,
+			Version: sourceRef.Version,
+			Kind:    sourceRef.Kind,
+		},
+		Resource:   sourceRef.Resource,
+		ObjectName: cache.NewObjectName(sourceRef.Namespace, sourceRef.Name),
+	}
+}
+
+func GetSourceRefFromWorkStatus(workStatus runtime.Object) (*SourceRef, error) {
```
MikeSpreitzer, line 186: Why do we need this in addition to `util.GetWorkStatusSourceRef`?

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,226 @@
...
+	logger.V(2).Info("Reconciling singleton workload object", "resource", wObjID.Resource, "objectName", wObjID.ObjectName)
+
+	if !c.singletonState.get(wObjID) {
+		emptyStatus := make(map[string]interface{})
+		return updateObjectStatus(ctx, &wObjID, emptyStatus, c.listers, c.wdsDynClient)
+	}
+
+	list, _ := c.workStatusLister.ByNamespace("").List(labels.Everything())
+	for _, obj := range list {
+		sourceRef, err := util.GetSourceRefFromWorkStatus(obj)
```
MikeSpreitzer, line 182: Why do we need this in addition to the `wsRef, _ := runtimeObjectToWorkStatusRef(obj)` (below)?

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+	}
+
+	allBdgs, err := c.wdsKsClient.ControlV1alpha1().Bindings().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, bdg := range allBdgs.Items {
+		logger.V(2).Info("Got Binding while building singleton state", "name", bdg.Name)
```
MikeSpreitzer, line 101: I think this should be at a numerically higher `V` level, it is a pretty low-level debugging detail.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,223 @@
...
+			}
+			labels := nsWObj.(metav1.Object).GetLabels()
+			if v, ok := labels[util.BindingPolicyLabelSingletonStatusKey]; ok {
+				wObjIdentifier := util.ObjectIdentifier{
+					GVK:        schema.GroupVersionKind{Group: nsWObjRef.Group, Version: nsWObjRef.Version, Kind: nsWObj.GetObjectKind().GroupVersionKind().Kind},
+					Resource:   nsWObjRef.Resource,
+					ObjectName: cache.NewObjectName(nsWObjRef.Namespace, nsWObjRef.Name),
+				}
+				if v == util.BindingPolicyLabelSingletonStatusValueSet && len(bdg.Spec.Destinations) > 0 {
+					logger.V(2).Info("Singleton workload object should have status synced", "resource", wObjIdentifier.Resource, "objectName", wObjIdentifier.ObjectName)
```
MikeSpreitzer, line 120: This statement is going to be executed many times for the same object, right? I am not a fan of spamming the debug log at `-v=2` with such low level boring stuff.

MikeSpreitzer, line 120: Did you want to include the binding name in this log message?
In many log collection/search frameworks, there is no such thing as looking at other earlier messages to provide context.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,223 @@
...
+			}
+		}
+	}
+
+	if !reconcile {
+		return nil
+	}
+
+	for wObjIdentifier := range c.singletonState.wObjSync {
+		if err := c.reconcileSingletonByWObj(ctx, wObjIdentifier); err != nil {
```
MikeSpreitzer, line 161: Why call `c.reconcileSingletonByWObj` when `c.singletonState.wObjSync[wObjIdentifier] == false` ?

pkg/status/controller.go
```diff
@@ -203,6 +206,8 @@ func (c *Controller) run(ctx context.Context, workers int, cListers chan interfa
 	}) // this will have the broker call the callback for all existing resolutions
 	c.logger.Info("Registered binding resolution broker callback")
 
+	c.buildSingletonStateAndOptionallyReconcile(ctx, false)
```
MikeSpreitzer, line 209: Why does this need to be done now? If I understand correctly, all this does is populate an internal data structure that is later consulted on a workload object-by-object basis. Why not just determine the needed bit when syncing that workload object?

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,223 @@
...
+		logger.V(2).Info("Updating singleton status", "workstatus", wsRef.Name, "wecName", wsRef.WECName)
+		return updateObjectStatus(ctx, &wsRef.SourceObjectIdentifier, status, c.listers, c.wdsDynClient)
+	}
+	return nil
+}
+
+func (c *Controller) reconcileSingletonByBdg(ctx context.Context, ref bindingRef) error {
+	logger := klog.FromContext(ctx)
+	logger.V(2).Info("Reconciling singleton status due to binding changes", "name", string(ref))
+	return c.buildSingletonStateAndOptionallyReconcile(ctx, true)
```
MikeSpreitzer, line 203: What happened to the "TODO" here? It still seems to me to be problematic to do this global operation every sync of a Binding.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,223 @@
...
+	defer ss.Unlock()
+
+	delete(ss.wObjSync, id)
+}
+
+func (c *Controller) buildSingletonStateAndOptionallyReconcile(ctx context.Context, reconcile bool) error {
+	logger := klog.FromContext(ctx)
+	logger.V(2).Info("Building the desired state for singleton statuses")
+
+	for wObjIdentifier := range c.singletonState.wObjSync {
```
MikeSpreitzer, line 80: Data race here.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,223 @@
...
+				}
+			}
+		}
+	}
+
+	if !reconcile {
+		return nil
+	}
+
+	for wObjIdentifier := range c.singletonState.wObjSync {
```
MikeSpreitzer, line 160: Data race here.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,223 @@
...
+	logger.V(2).Info("Reconciling singleton status due to binding changes", "name", string(ref))
+	return c.buildSingletonStateAndOptionallyReconcile(ctx, true)
+}
+
+func (c *Controller) reconcileSingletonByWs(ctx context.Context, ref singletonWorkStatusRef) error {
+	logger := klog.FromContext(ctx)
+	logger.V(2).Info("Reconciling singleton status due to workstatus changes", "name", string(ref.Name))
+	obj, _ := c.workStatusLister.ByNamespace(ref.WECName).Get(ref.Name)
+	status, _ := util.GetWorkStatusStatus(obj)
+	if sync, ok := c.singletonState.wObjSync[ref.SourceObjectIdentifier]; !ok {
```
MikeSpreitzer, line 211: Data race here.

pkg/status/singletonstatus.go
waltforme: I totally agree. I temporarily (in this _WIP_ PR) put a 2 here just because I wanted to see it when I was debugging with `-v 2`. Say I put 4s and do `-v 4` then the logging will be too verbose and the messages will be hard to find.

pkg/status/controller.go
MikeSpreitzer: This data structure is keeping track of one bit per workload object, indicating whether singleton status return should be done for that object. That is the "needed bit" that I meant.

waltforme: I see what you meant. To get the value of the that bit for _one_ workload object, the reconciliation needs to go over _all_ Binding objects. So I thought it is more efficient to do a full build (for _all_ workload object together) here, then do incremental updates in reconciliations. That's why I had this line here.

But right now I don't yet have an idea on how to do the incremental updates. Because an event of _one_ Binding object is not always sufficient to compute the incremental. For example, if a Binding object says "I don't select this workload object anymore", can I mark the corresponding bit to `false`? No, because other Binding object(s) may still cover the workload object, I still have to go over all the Binding objects to determine that bit. To me it seems that a full build every time is necessary so I removed the `TODO`.

pkg/status/singletonstatus.go
waltforme: Explained [here](https://github.com/kubestellar/kubestellar/pull/2285#discussion_r1709660194).

pkg/status/singletonstatus.go
waltforme: Yes, explained [here](https://github.com/kubestellar/kubestellar/pull/2285#discussion_r1709546859).

pkg/status/singletonstatus.go
waltforme: Added.

pkg/status/singletonstatus.go
waltforme: I am using it as has the label.

waltforme:
Latest force-push was a reorg and squash of commits by rebasing.

pkg/status/singletonstatus.go
waltforme: I think it should be `>0`, which says this Binding object asks some WEC to be candidates so that the overall set of candidates (after checking all Binding objects) is for sure not empty, thus the status should be synced.

pkg/status/singletonstatus.go
waltforme: Adopted the suggestion everywhere. Thanks!

pkg/status/singletonstatus.go
waltforme: Fixed.

waltforme:
> Don't forget about https://docs.kubestellar.io/unreleased-development/direct/architecture/#status-controller .

Updated the doc and the image.

pkg/status/singletonstatus.go
waltforme: Now all the log levels are regularized by following the content of #2390

pkg/status/singletonstatus.go
waltforme: Now the internal map was dropped, we don't have this problem. But still I fixed this before dropping.

pkg/status/singletonstatus.go
waltforme: Now the internal map was dropped, we don't have this problem. But still I fixed this before dropping.

pkg/status/singletonstatus.go
waltforme: I agree. Can we address the status ownership issue in another PR? I documented this as https://github.com/kubestellar/kubestellar/issues/2393

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,233 @@
...
+		if err = c.checkSingletonWorkloadInBinding(ctx, bdg, sync, false); err != nil {
+			return err
+		}
+	}
+
+	for wObjIdentifier, v := range sync {
+		if err := c.reconcileSingletonWObj(ctx, wObjIdentifier, v); err != nil {
+			return err
+		}
+	}
```
MikeSpreitzer, line 68: What this logic does is: first determine whether any Binding that is not being deleted associates an object in `sync` with at least one WEC, and then for each such object copy the status from *every* related WorkStatus to the status section of that object in the WDS. **Singleton** status return is supposed to be about returning status for an object that is associated with _exactly one_ WEC.

MikeSpreitzer, line 68: What this logic does is: first, for every object labeled to return singleton status, determine whether any Binding that is not being deleted associates that object with at least one WEC, and second, for each such object copy the status from _every_ related WorkStatus to the status section of that object in the WDS. **Singleton** status return is supposed to be about returning status for an object that is associated (by the effects of all the Bindings together) with _exactly one_ WEC.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,233 @@
...
+			return err
+		}
+		status, err := util.GetWorkStatusStatus(wsObj)
+		if err != nil {
+			return err
+		}
+		if status == nil {
+			return nil
+		}
+		return updateObjectStatus(ctx, &wObjID, status, c.listers, c.wdsDynClient)
```
MikeSpreitzer, line 165: There are two copies of this code. The other copy is in reconcileSingletonWObj. The other copy includes debug logging of this action, which is a good thing.

pkg/status/singletonstatus.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+		logger.V(4).Info("Updating singleton status", "workstatus", wsRef.Name, "wecName", wsRef.WECName)
+		return updateObjectStatus(ctx, &wsRef.SourceObjectIdentifier, status, c.listers, c.wdsDynClient)
+	}
+	return nil
+}
+
+// reconcileSingletonByWs decides sync or not for the corresponding workload object,
+// then maintains its status based on the decision.
+func (c *Controller) reconcileSingletonByWs(ctx context.Context, ref singletonWorkStatusRef) error {
```
MikeSpreitzer, line 107: I suppose that the "Ws" at the end of this method name is supposed to stand for "WorkStatus". If I have that right then I think that "WS" would do the job better.

MikeSpreitzer (approved):
More work is needed here. It can be done in follow-on PRs.

/lgtm
/approve
Thanks!

pkg/status/singletonstatus.go
waltforme: Addressed in #2409.

pkg/status/singletonstatus.go
waltforme: Addressed in #2409.

pkg/status/singletonstatus.go
waltforme: Improved in #2409 by checking whether there are exactly one WEC.

PR #2288, opened 2024-07-02
🌱 Add E2E test scenarios manipulating object label

MikeSpreitzer:
## Summary
This PR extends the ginkgo E2E test with a scenario that flaps the workload object label matched by the BindingPolicy, checking that the object is removed from and then restored to the WECs.

## Related issue(s)

Fixes #


MikeSpreitzer:
/cc @effi-ofer 
/cc @pdettori 
/cc @ezrasilvera 
@cwiklik

pdettori (approved):
/lgtm

PR #2174, opened 2024-05-28
:sparkles: CombinedStatus Calculation (part 2)

vMaroon:
## Summary
This PR outlines the design for the implementation of the calculation of CombinedStatus objects.
The PR introduces new components that maintain the resolutions of combined-statuses, by maintaining a resolution entry per each of the latter (in the new CombinedStatusResolver), in which information from StatusCollectors and the evaluation of WorkStatuses are kept up to date.

The code is incomplete, and has TODOs that should cover finishing the maintenance of CombinedStatus objects without error propagation.

Note that the outlined design attempts to find balance between over-optimizing (in my opinion there is none), and the efficiency required due to the highly distributed nature of the feature while respecting eventual consistency and multiple information update sources. 

## Related issue(s)
- #2141

PR #2293, opened 2024-07-05
:sparkles: CombinedStatus querying enhancements

vMaroon:
## Summary
This PR implements [Availability of source object's `.spec` and `.metadata` for CELs](https://github.com/kubestellar/kubestellar/pull/2174#discussion_r1625431727) as outlined in #2141 

It also involves general code improvements.

## Use Example
For the following StatusCollectors:
```
kind: StatusCollector
metadata:
  name: deployments-selector
spec:
  filter: obj.kind == "Deployment"
  limit: 10
  select:
    - name: wec
      def: inventory.name
    - name: full-status
      def: returned.status
    - name: replicas
      def: returned.status.replicas
---

kind: StatusCollector
metadata:
  name: deployments-aggregator
spec:
  limit: 10
  filter: obj.kind == "Deployment"
  groupBy:
    - name: apiVersion
      def: returned.status.conditions[0].lastUpdateTime
  combinedFields:
    - name: count
      type: COUNT
    - name: returned.status.replicasSum
      type: SUM
      subject: returned.status.availableReplicas
```

The given CombinedStatus is generated (collectors applied on our 2 multi-cluster test deployments)
```
- apiVersion: control.kubestellar.io/v1alpha1
  kind: CombinedStatus
  metadata:
    creationTimestamp: "2024-07-14T17:08:54Z"
    generation: 2
    labels:
      status.kubestellar.io/api-group: apps
      status.kubestellar.io/binding-policy: nginx-bindingpolicy
      status.kubestellar.io/name: nginx-deployment
      status.kubestellar.io/namespace: nginx
      status.kubestellar.io/resource: deployments
    name: 6e4ec320-146b-428d-8a87-577f89310a57.d491a841-4909-4e6a-b7ed-b942c9dd4894
    namespace: nginx
    resourceVersion: "7071"
    uid: 70bc24ed-b116-4679-aec4-9661fb1fb591
  results:
  - columnNames:
    - lastUpdateTime
    - count
    - returned.status.replicasSum
    name: deployments-aggregator
    rows:
    - columns:
      - string: "2024-07-14T17:08:51Z"
        type: String
      - float: "1"
        type: Number
      - float: "1"
        type: Number
    - columns:
      - string: "2024-07-14T17:08:53Z"
        type: String
      - float: "1"
        type: Number
      - float: "1"
        type: Number
  - columnNames:
    - wec
    - full-status
    - replicas
    name: deployments-selector
    rows:
    - columns:
      - string: cluster1
        type: String
      - object:
          availableReplicas: 1
          conditions:
          - lastTransitionTime: "2024-07-14T17:08:51Z"
            lastUpdateTime: "2024-07-14T17:08:51Z"
            message: Deployment has minimum availability.
            reason: MinimumReplicasAvailable
            status: "True"
            type: Available
          - lastTransitionTime: "2024-07-14T17:08:49Z"
            lastUpdateTime: "2024-07-14T17:08:51Z"
            message: ReplicaSet "nginx-deployment-7dd74d45b5" has successfully progressed.
            reason: NewReplicaSetAvailable
            status: "True"
            type: Progressing
          observedGeneration: 1
          readyReplicas: 1
          replicas: 1
          updatedReplicas: 1
        type: Object
      - float: "1"
        type: Number
    - columns:
      - string: cluster2
        type: String
      - object:
          availableReplicas: 1
          conditions:
          - lastTransitionTime: "2024-07-14T17:08:53Z"
            lastUpdateTime: "2024-07-14T17:08:53Z"
            message: Deployment has minimum availability.
            reason: MinimumReplicasAvailable
            status: "True"
            type: Available
          - lastTransitionTime: "2024-07-14T17:08:51Z"
            lastUpdateTime: "2024-07-14T17:08:53Z"
            message: ReplicaSet "nginx-deployment-7dd74d45b5" has successfully progressed.
            reason: NewReplicaSetAvailable
            status: "True"
            type: Progressing
          observedGeneration: 1
          readyReplicas: 1
          replicas: 1
          updatedReplicas: 1
        type: Object
      - float: "1"
        type: Number
```
## Related issue(s)
- #2141


pkg/status/combinedstatus-resolution.go
```diff
@@ -539,14 +581,20 @@ func handleSelect(scName string, scData *statusCollectorData) *v1alpha1.NamedSta
...
-			case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64:
-				numStr := fmt.Sprintf("%d", v)
+			case int, int8, int16, int32, int64:
+				numStr := strconv.Itoa(int(v.(int64)))
+				col = v1alpha1.Value{
+					Type:   v1alpha1.TypeNumber,
+					Number: &numStr,
+				}
+			case uint, uint8, uint16, uint32, uint64:
+				numStr := strconv.FormatInt(int64(v.(uint64)), 10)
```
MikeSpreitzer, line 591: I do not understand why casting to, and formatting as, a signed integer is the right thing to do here. `strconv.FormatUint` is a thing.

pkg/status/combinedstatus-resolution.go
```diff
@@ -657,7 +705,7 @@ func calculateCombinedFieldAggregation(combinedFieldNamedAgg v1alpha1.NamedAggre
 
 			sum += *subject
 		}
-		numStr = fmt.Sprintf("%f", sum)
+		numStr = strconv.FormatFloat(sum, 'f', -1, 64)
```
MikeSpreitzer, line 708: Wouldn't `'g'` be the more natural value for `fmt` here?

pkg/status/combinedstatus-resolution.go
```diff
@@ -827,3 +886,15 @@ func valueEqual(a, b *v1alpha1.Value) bool {
 		return false
 	}
 }
+
+func parseNumber(s string) (float64, error) {
+	if intValue, err := strconv.Atoi(s); err == nil {
+		return float64(intValue), nil
```
MikeSpreitzer, line 892: `float64(intValue)` can lose precision. How about returning an `any` that holds either an `int64`, a `uint64` that is not `==` any `int64`, or a `float64`?

pkg/status/combinedstatus-resolver.go
```diff
@@ -456,9 +478,26 @@ func (c *combinedStatusResolver) evaluateWorkStatusesPerBindingReadLocked(bindin
...
+
+				if objMap != nil {
+					for k, v := range objMap {
+						content[k] = v
+					}
+				}
+			}
+
+			// evaluate workstatus
+			if csResolution.evaluateWorkStatus(c.celEvaluator, workStatus.wecName, content) {
```
MikeSpreitzer, line 500: There is enough duplicated code here now to be worth creating a procedural abstraction for.

pkg/status/combinedstatus-resolver.go
```diff
@@ -468,6 +507,41 @@ func (c *combinedStatusResolver) evaluateWorkStatusesPerBindingReadLocked(bindin
...
+// The function is guaranteed not to return a key of `status` in the map.
+// If the resource contains any other subresources, they are fetched as well.
+func (c *combinedStatusResolver) getObjectMetaAndSpec(objectIdentifier util.ObjectIdentifier) (map[string]interface{}, error) {
+	// fetch object
+	lister, exists := c.wdsListers.Get(objectIdentifier.GVR())
+	if !exists {
+		return nil, fmt.Errorf("lister not found for gvr %s", objectIdentifier.GVR())
+	}
+
+	obj, err := lister.ByNamespace(objectIdentifier.ObjectName.Namespace).Get(objectIdentifier.ObjectName.Name)
```
MikeSpreitzer, line 521: What if the resource is not namespaced?

vMaroon: Cluster-wide resources do not make it this far, but yes this function should probably be agnostic of that.

On the first note: since an object cannot be both namespace scoped and cluster scoped, CombinedStatus currently supports namespace-scoped objects only. We can duplicate for cluster-scoped resources and fit them through the same pipeline later.

pkg/status/combinedstatus-resolver.go
```diff
@@ -468,6 +507,41 @@ func (c *combinedStatusResolver) evaluateWorkStatusesPerBindingReadLocked(bindin
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get object (%v) with gvr (%v): %w", objectIdentifier.ObjectName,
```
MikeSpreitzer, line 523: If the workload object does not exist in the WDS then (a) it must have been recently deleted and (b) the user no longer wants CombinedStatus objects for it and (c) doing nothing but debug logging here will be adequate because deletion events are inbound, right?

pkg/status/combinedstatus-resolution.go
```diff
@@ -334,21 +338,54 @@ func (c *combinedStatusResolution) evaluateWorkStatus(celEvaluator *celEvaluator
...
 }
 
+func (c *combinedStatusResolution) requiresSourceObjectMetaOrSpec() bool {
+	c.RLock()
+	defer c.RUnlock()
+
+	pred := func(s string) bool {
+		return strings.Contains(s, fmt.Sprintf("%s.metadata", expressionRootName)) ||
+			strings.Contains(s, fmt.Sprintf("%s.spec", expressionRootName)) ||
+			strings.Contains(s, fmt.Sprintf("%s.kind", expressionRootName)) ||
```
MikeSpreitzer, line 358: What if the expression contains `otherobj.kindly`?
What if the expression contains `someField.obj.kind`?
What if the expression contains `"some string about obj.kind"`?

MikeSpreitzer:
I am coming to the view that we should define four top-level identifiers for use in the CEL expressions here. I am not sure what are the best names; I have some suggestions below.

- `obj` refers to the workload object in the WDS.
- `returned.status` refers to the returned status in the WorkStatus object in the ITS.
- `inventory` refers to the InventoryObjectReference augmentation of #2055.
- `propagation` refers to the PropagationMeta augmentation of #2055.

This removes `inventory` and `propagation` from the space of names of fields of the object.

This approach plays nice with the possibility that not every field of `.status` originates in the WEC and propagates to the ITS.

This approach works even in a future where some reported state is elsewhere than in `.status` (`returned` can have more than `.status`).

This approach removes the need for the code to smush objects together.

pkg/status/combinedstatus-resolution.go
vMaroon: I saw that it made no difference on the examples I tried, done anyway.

pkg/status/combinedstatus-resolution.go
vMaroon: We can attach the source object's subresources in all cases. This here attempts to do so when we know that we **may** need them, and be sure not to fetch the object when we know it wont be needed.

pkg/status/combinedstatus-resolution.go
```diff
@@ -827,3 +870,34 @@ func valueEqual(a, b *v1alpha1.Value) bool {
...
+
+	if floatValue, err := strconv.ParseFloat(s, 64); err == nil {
+		return floatValue, nil
+	}
+
+	return nil, fmt.Errorf("failed to parse number")
+}
+
+// numericEqual compares two numeric values. It is assumed that the types of a
+// and b are the same.
```
MikeSpreitzer, line 891: I do not see why this assumption is justified. When the assumption is not true, this func will panic, right?

pkg/status/combinedstatus-resolver.go
MikeSpreitzer: What do you mean by "CombinedStatus currently supports namespace-scoped objects only"?

See https://github.com/kubestellar/kubestellar/blob/70ded98849b7236c5c160667c55c8b3f943b39d8/api/control/v1alpha1/types.go#L447

pkg/status/combinedstatus-resolution.go
```diff
@@ -357,32 +377,34 @@ func (c *combinedStatusResolution) evaluateWorkStatus(celEvaluator *celEvaluator
...
-		return true, nil
+		return true
 	}
 
 	// evaluate filter to determine if the workstatus is relevant
 	if scData.Filter != nil {
 		eval, err := celEvaluator.Evaluate(*scData.Filter, workStatusContent)
 		if err != nil {
-			return false, fmt.Errorf("failed to evaluate filter expression: %w", err)
+			runtime2.HandleError(fmt.Errorf("failed to evaluate filter expression: %w", err))
```
MikeSpreitzer, line 392: FYI, in Kubernetes release 1.31 we get https://github.com/kubernetes/apimachinery/blob/v0.31.0-alpha.1/pkg/util/runtime/runtime.go#L159-L180

MikeSpreitzer:
Also FYI, Kubernetes release 1.31 will introduce https://github.com/kubernetes/client-go/blob/v0.31.0-alpha.3/util/workqueue/rate_limiting_queue.go#L27

pkg/status/combinedstatus-resolver.go
vMaroon: I was wondering how this didnt bother me until now, thanks, forgot that. Will fix.

pkg/status/combinedstatus-resolution.go
vMaroon: Right, tried to save some type checks falsely by this line:
```
func valueEqual(a, b *v1alpha1.Value) bool {
	if a.Type != b.Type {
		return false
	}
```

Which is incorrect, thanks.

pkg/status/combinedstatus-resolver.go
vMaroon: Fixed.

pkg/status/combinedstatus-resolution.go
vMaroon: This will be addressed in PR for error handling.

MikeSpreitzer: I do not think we want to hold up that PR until after IBM Cloud offers a version of OpenShift that uses Kubernetes release 1.31 (which has not come out yet).

MikeSpreitzer:
Regarding top-level names:

When we get to augmentations along the lines shown in #2055 --- which I think we are agreed should happen --- I think it would be least disruptive to users if we do not have to forbid them to use certain field names in their object schemas.

We already have users asking to KubeStellar to support scenarios where some of the content in `.status` does not propagate between WDS and WEC. I think that we should plan on supporting this. When the content of `.status` in the two places is persistently different, I think that it is more natural to address it by two different names in the expressions in this feature.

Evaluating the expressions involved in this feature is done inside many layers of nested loops. The performance of these evaluations will be noticeable. This includes both direct CPU costs as well as the costs indirectly associated with dynamically allocated and garbage-collected memory. While the subject of this comment is only one component of those costs, I would still recommend making easy cost savings now.

pkg/status/combinedstatus-resolver.go
vMaroon: Fixed.

MikeSpreitzer: Huh? I still see a copy of this code at new lines 314--335.

pkg/status/combinedstatus.go
```diff
@@ -83,6 +87,13 @@ func (c *Controller) updateOrCreateCombinedStatus(ctx context.Context, combinedS
 	// set results
 	combinedStatus.Results = generatedCombinedStatus.Results
```
MikeSpreitzer, line 88: `combinedStatus` points to storage fetched from the informer's local cache, which is supposed to be immutable, right?

MikeSpreitzer:
`func (c *combinedStatusResolution) compareCombinedStatus` could be more efficient, avoiding constructing new objects in the happy case of equality. This can be addressed in follow-on work.

MikeSpreitzer:
`statusCombinationEqual` could do fewer comparisons by recognizing that in the case of a StatusCombiner that does aggregation, the set of rows is logically a map whose key is the columns holding the `GroupBy` values.

MikeSpreitzer:
`statusCombinationEqual` is comparing two relations. This can be sped up by using hashing, at the cost of memory. Can be considered in follow-on work.

pkg/status/cel-evaluator.go
```diff
@@ -37,7 +46,9 @@ type celEvaluator struct {
 func newCELEvaluator() (*celEvaluator, error) {
 	env, err := cel.NewEnv(
 		cel.Declarations(
-			decls.NewVar("obj", decls.NewMapType(decls.String, decls.Dyn)),
+			decls.NewVar(sourceObjectKey, decls.NewMapType(decls.String, decls.Dyn)),
+			decls.NewVar(statusKey, decls.NewMapType(decls.String, decls.Dyn)),
```
MikeSpreitzer, line 50: I do not think that "returned.status" is a legitimate variable name.

vMaroon: While I understand that the intention is to assign `returned` to whatever is returned from workstatuses, at the moment that is only `.status`. I thought of making this explicit. If you insist, I can make it implicit.

vMaroon:
@MikeSpreitzer thank you for pointing out possible efficiency improvements. I am aware of those, and the intention is to begin with a simpler working baseline and work from there (as you suggested).

pkg/status/cel-evaluator.go
```diff
@@ -83,7 +94,9 @@ func (e *celEvaluator) Evaluate(expression v1alpha1.Expression, objMap map[strin
 
 	// evaluate the expression with the unstructured object
 	result, _, err := prog.Eval(map[string]interface{}{
-		"obj": objMap,
+		sourceObjectKey: objMap[sourceObjectKey],
+		statusKey:       objMap[statusKey],
+		inventoryKey:    objMap[inventoryKey],
```
MikeSpreitzer, line 99: This seems almost exactly redundant with `func (c *combinedStatusResolver) getCombinedContentMap`

pkg/status/cel-evaluator.go
```diff
@@ -37,7 +46,9 @@ type celEvaluator struct {
 func newCELEvaluator() (*celEvaluator, error) {
 	env, err := cel.NewEnv(
 		cel.Declarations(
-			decls.NewVar("obj", decls.NewMapType(decls.String, decls.Dyn)),
+			decls.NewVar(sourceObjectKey, decls.NewMapType(decls.String, decls.Dyn)),
+			decls.NewVar(statusKey, decls.NewMapType(decls.String, decls.Dyn)),
+			decls.NewVar(inventoryKey, decls.NewMapType(decls.String, decls.Dyn)),
```
MikeSpreitzer, line 51: The schema for this object is fixed, right?

pkg/status/cel-evaluator.go
MikeSpreitzer: My concern here is with CEL formalities. I do not think that "returned.status" is a legitimate variable name in CEL.

pkg/status/combinedstatus.go
```diff
@@ -83,6 +87,13 @@ func (c *Controller) updateOrCreateCombinedStatus(ctx context.Context, combinedS
...
 
+	if combinedStatus.Namespace == metav1.NamespaceNone {
+		combinedStatus.Namespace = clusterScopedObjectsNamespace
+		if err := c.ensureNamespaceExists(ctx, combinedStatus.Namespace); err != nil {
+			return fmt.Errorf("failed to ensure namespace exists: %w", err)
+		}
+	}
```
MikeSpreitzer, line 95: I suspect that the code will be simpler to understand if `combinedStatus.Namespace` never holds the wrong value --- i.e., the construction in `func (c *combinedStatusResolution) generateCombinedStatus` always puts the right value there.

pkg/status/combinedstatus-resolver.go
```diff
@@ -468,6 +470,69 @@ func (c *combinedStatusResolver) evaluateWorkStatusesPerBindingReadLocked(bindin
...
+	unstrObj, ok := obj.(*unstructured.Unstructured)
+	if !ok {
+		return nil, fmt.Errorf("object cannot be cast to *unstructured.Unstructured: object: %s",
+			util.RefToRuntimeObj(obj))
+	}
+
+	objMap := map[string]interface{}{}
+	// get all but status from the READONLY object
+	// if the object has custom subresources, this will preserve them
+	for k, v := range unstrObj.Object { // rather than DeepCopy()
```
vMaroon, line 520: This is no longer required.

pkg/status/cel-evaluator.go
vMaroon: I tested that it works, at a simple level at least. In all cases, changing in next push.

pkg/status/combinedstatus-resolver.go
vMaroon: Were you referring to abstracting the creation of the map only or also the call to evaluateWorkStatus?

MikeSpreitzer: I used GitHub's feature where I can select a range of lines and put a comment on that range. That is the duplicated code I was referring to.

.github/workflows/pr-test-e2e.yml
```diff
@@ -123,6 +123,10 @@ jobs:
         if: always()
         run: kubectl --context its1 get workstatuses -A -o yaml
 
+      - name: show combinedstatuses
+        if: always()
+        run: kubectl --context wds1 get combinedstatuses -A -o yaml
+
```
MikeSpreitzer, line 129: +1 !

pkg/status/combinedstatus-resolver.go
MikeSpreitzer: In the current version of this PR, commit adcde7fd0d8bbce6d914570d33df6d9a41bc8495, much of the duplicated code has been swept into `c.getCombinedContentMap`. There remains some duplication outside that method: lines 314--320 and lines 460--466 are duplicates. Not a blocker, that is a small amount of duplicated material.

pkg/status/combinedstatus.go
```diff
@@ -68,46 +68,42 @@ func (c *Controller) syncCombinedStatus(ctx context.Context, ref string) error {
...
-	combinedStatus.Results = generatedCombinedStatus.Results
-
 	logger := klog.FromContext(ctx)
-	csEcho, err := c.wdsKsClient.ControlV1alpha1().CombinedStatuses(combinedStatus.Namespace).Update(ctx,
-		combinedStatus, metav1.UpdateOptions{FieldManager: controllerName})
+	csEcho, err := c.wdsKsClient.ControlV1alpha1().CombinedStatuses(generatedCombinedStatus.Namespace).Update(ctx,
+		generatedCombinedStatus, metav1.UpdateOptions{FieldManager: controllerName})
 
 	if err != nil {
 		if errors.IsNotFound(err) {
```
MikeSpreitzer, line 87: BTW, you expect `Update` to succeed if and only if the object's `ResourceVersion` is set to something, and you expect `Create` to succeed if and only if the object's `ResourceVersion` is the empty string, right? This suggests that it is certain which API call to make.

pkg/status/cel-evaluator.go
```diff
@@ -37,7 +46,9 @@ type celEvaluator struct {
 func newCELEvaluator() (*celEvaluator, error) {
 	env, err := cel.NewEnv(
 		cel.Declarations(
-			decls.NewVar("obj", decls.NewMapType(decls.String, decls.Dyn)),
+			decls.NewVar(sourceObjectKey, decls.NewMapType(decls.String, decls.Dyn)),
+			decls.NewVar(returnedKey, decls.NewMapType(decls.String, decls.Dyn)),
+			decls.NewVar(inventoryKey, decls.NewMapType(decls.String, decls.String)),
```
MikeSpreitzer, line 51: I suspect it is possible to say here that this map has only one string in its domain, `"name"`. I think it would be good to say that, at least by the time this code is developed enough to shift the type checking left. Which is not yet.

pkg/status/combinedstatus-resolution.go
```diff
@@ -264,6 +251,10 @@ func (c *combinedStatusResolution) generateCombinedStatus(bindingName string,
 		Results: make([]v1alpha1.NamedStatusCombination, 0, len(c.statusCollectorNameToData)),
 	}
 
+	if combinedStatus.Namespace == v1.NamespaceNone {
```
MikeSpreitzer, line 254: I see that this PR did not introduce this choice, but: there are a lot of package paths that end with "v1". Using that alone as a name is very ambiguous. There is a very dominant tradition of using "metav1" as the name of this particular package. Not a blocker.

pkg/abstract/primitive-maps.go
```diff
@@ -59,3 +59,11 @@ func PrimitiveMapSafeValMap[Key comparable, Val, Mapped any](lock *sync.RWMutex,
 
 	return PrimitiveMapValMap(source, mapper)
 }
+
+func PrimitiveMapKeys[Key comparable, Val any](rep map[Key]Val) []Key {
```
MikeSpreitzer, line 63: Maybe `PrimitiveMapKeySlice`, so that later we can add `PrimitiveMapKeySet`?

pkg/status/combinedstatus.go
```diff
@@ -61,53 +61,51 @@ func (c *Controller) syncCombinedStatus(ctx context.Context, ref string) error {
...
-			logger.Info("Created CombinedStatus", "ns", combinedStatus.Namespace,
-				"name", combinedStatus.Name, "resourceVersion", csEcho.ResourceVersion)
-			return nil
-		} else {
 			return fmt.Errorf("failed to update CombinedStatus (ns, name = %v, %v): %w",
-				combinedStatus.Namespace, combinedStatus.Name, err)
+				generatedCombinedStatus.Namespace, generatedCombinedStatus.Name, err)
 		}
+
+		return nil
```
MikeSpreitzer, line 96: Don't you want a log message in this case (successful update)?

pkg/status/combinedstatus-resolver.go
```diff
@@ -468,6 +470,60 @@ func (c *combinedStatusResolver) evaluateWorkStatusesPerBindingReadLocked(bindin
 	return combinedStatusesToQueue
 }
 
+// getCombinedContentMap returns a map of content for the given workstatus.
+func (c *combinedStatusResolver) getCombinedContentMap(workStatus *workStatus,
+	resolution *combinedStatusResolution) map[string]interface{} {
+	content := map[string]interface{}{
+		returnedKey:  workStatus.Content(),
```
MikeSpreitzer, line 477: I would make `returned.status` the way to access the workload object's status from the WEC, so that in the future we can easily add access to returned state in other parts of the object.

```suggestion
		returnedKey:  map[string]any{"status": workStatus.Content()},
```

pkg/status/combinedstatus-resolution.go
```diff
@@ -586,22 +617,22 @@ func handleCombinedFields(scName string, scData *statusCollectorData) *v1alpha1.
 	if len(scData.GroupBy) > 0 {
```
MikeSpreitzer, line 617: FYI, I would find it easier to follow the code if the comment on the `statusCollectorData` said that each one is specific to a (Binding, StatusCollector, workload object) combination.

pkg/status/combinedstatus-resolution.go
```diff
@@ -586,22 +617,22 @@ func handleCombinedFields(scName string, scData *statusCollectorData) *v1alpha1.
...
 		for _, groupByNamedExp := range scData.GroupBy {
 			groupByNameToValueToWorkStatus[groupByNamedExp.Name] = make(map[any][]*workStatusData)
+
+			for _, wsData := range scData.wecToData {
```
MikeSpreitzer, line 621: A `workStatusData` is specific to a Binding, it does not apply across all Bindings that reference the same StatusCollector, right? If so then it would help if the comment on that type said so.

pkg/status/combinedstatus-resolution.go
MikeSpreitzer, line 617: FYI, I find the method name "handleCombinedFields" to be not very clear about what this method is supposed to do. If you do not want to revise the name then I would find a comment on the method to be helpful.

pkg/status/combinedstatus-resolution.go
```diff
@@ -586,22 +617,22 @@ func handleCombinedFields(scName string, scData *statusCollectorData) *v1alpha1.
 	if len(scData.GroupBy) > 0 {
 		for _, groupByNamedExp := range scData.GroupBy {
 			groupByNameToValueToWorkStatus[groupByNamedExp.Name] = make(map[any][]*workStatusData)
```
MikeSpreitzer, line 619: FYI: I find myself working harder than I would like to figure out what are the keys in this map. Maybe a comment on `groupByNameToValueToWorkStatus` would help?

pkg/status/combinedstatus-resolution.go
```diff
@@ -640,13 +671,22 @@ func handleCombinedFields(scName string, scData *statusCollectorData) *v1alpha1.
 	return &namedStatusCombination
```
MikeSpreitzer, line 671: (I finally figured out that the purpose of this method is to compute the intended value of one NamedStatusCombination.)

The logic in this method has some problems (not introduced in this PR). One problem is that a NamedStatusCombination should have three kinds of columns, for the GroupBy, CombinedFields, and Select items in the StatusCollectorSpec.

There are several additional problems in the aggregation logic in this method. Perhaps I will try to update the documentation of what this feature is intended to do.

MikeSpreitzer:
I noticed something that was introduced earlier. https://github.com/vMaroon/kubestellar/blob/daef7ec4bebcc86f7d9e9c484c9017e7bf05e448/test/util/util.go#L201-L204 looks redundant with https://github.com/vMaroon/kubestellar/blob/daef7ec4bebcc86f7d9e9c484c9017e7bf05e448/test/util/util.go#L206-L209 .

MikeSpreitzer:
Here is something else introduced earlier. Rather than fetching a CombinedStatus by first fetching objects to get their UIDs and then fetching the CombinedStatus by name as in https://github.com/vMaroon/kubestellar/blob/daef7ec4bebcc86f7d9e9c484c9017e7bf05e448/test/util/util.go#L193-L204 , the labels of CombinedState are designed so that the CombinedStatus can be fetched directly by a label query.

Does not have to be changed in this PR.

MikeSpreitzer:
Regarding https://github.com/vMaroon/kubestellar/blob/daef7ec4bebcc86f7d9e9c484c9017e7bf05e448/test/util/util.go#L212 : I think the right approach is what is done elsewhere, wait for the expected values to show up.

Does not have to be changed in this PR.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
```diff
@@ -469,30 +475,28 @@ var _ = ginkgo.Describe("end to end testing", func() {
...
 			util.CreateBindingPolicy(ctx, ksWds, bpName, clusterSelector, testAndStatusCollection)
 
 			cs := util.GetCombinedStatus(ctx, ksWds, wds, ns, workloadName, bpName)
 
 			// Validate CombinedStatus results
 			gomega.Expect(len(cs.Results)).To(gomega.Equal(1))
-			gomega.Expect(cs.Results[0].Name).To(gomega.Equal(countAvailableReplicasStatusCollectorName))
+			gomega.Expect(cs.Results[0].Name).To(gomega.Equal(sumAvailableReplicasStatusCollectorName))
 
 			gomega.ExpectWithOffset(1, len(cs.Results[0].ColumnNames)).To(gomega.Equal(1))
```
MikeSpreitzer, line 498: Not introduced in this PR, but: why is there an offset of 1 here? I see inconsistent offsets within `ginkgo.Context("combined status testing", func() {...})`. If I understand correctly, an offset of 1 should be used in a function _called from_ a subject node like those in this ginkgo.Context.

pkg/status/combinedstatus-resolution.go
MikeSpreitzer: Please see #2325 .

pkg/status/combinedstatus.go
vMaroon: Added log and removed "synced" message from caller.

pkg/status/combinedstatus-resolver.go
vMaroon: This already is what is implemented. `returnedKey` is `"returned"`, and `workStatus::Content` is the function that maps the information returned from WECs:

```
func (ws *workStatus) Content() map[string]interface{} {
	return map[string]interface{}{
		"status": ws.status,
	}
}
```

vMaroon:
Thanks @MikeSpreitzer. I'll address the clarification of the desired output ASAP. Regarding the tests, @andreyod wrote a basic extension of tests to kickoff coverage. If possible, I'd like not to handle tests at this stage.

I propose:
1. I update the presentation of status combinations
2. I update the current tests to cover the resulting form
3. @andreyod enhances the implementation and addresses the comments

What do you think?

MikeSpreitzer:
More smaller PRs are easier to review than one big PR.

vMaroon:
@MikeSpreitzer addressed the last batch of comments (excluding the ones involving tests). Since your PR #2325 pushes feature documentation, do you think we can merge this one if there are no more comments?

pkg/status/combinedstatus-resolution.go
```diff
@@ -500,28 +531,22 @@ func validateCombinedStatusLabels(combinedStatus *v1alpha1.CombinedStatus,
...
+		ColumnNames: make([]string, 0, len(scData.Select)), // for now, manually add wecName to all rows
 		Rows:        make([]v1alpha1.StatusCombinationRow, 0, len(scData.wecToData)),
 	}
 
 	// add column names
-	namedStatusCombination.ColumnNames = append(namedStatusCombination.ColumnNames, "wecName")
 	namedStatusCombination.ColumnNames = append(namedStatusCombination.ColumnNames,
 		abstract.SliceMap(scData.Select, func(selectNamedExp v1alpha1.NamedExpression) string {
 			return selectNamedExp.Name
 		})...)
```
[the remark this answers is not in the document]
vMaroon: I personally prefer this approach too, but there are several instances of accessing the name of Named APIs (directly or through abstract functions) - I think this better attempt to address as many of them as makes sense, in follow-up work.

pkg/status/cel-evaluator.go
vMaroon: This was handled.

pkg/status/combinedstatus.go
MikeSpreitzer: > removed "synced" message from caller.

In that case it is best to have some log message at the end of every possible way the execution could go.

pkg/status/combinedstatus.go
```diff
@@ -61,53 +61,52 @@ func (c *Controller) syncCombinedStatus(ctx context.Context, ref string) error {
...
+			"name", generatedCombinedStatus.Name, "resourceVersion", csEcho.ResourceVersion)
+		return nil
+	}
+
+	csEcho, err := c.wdsKsClient.ControlV1alpha1().CombinedStatuses(generatedCombinedStatus.Namespace).Create(ctx,
+		generatedCombinedStatus, metav1.CreateOptions{FieldManager: controllerName})
+	if err != nil {
+		runtime2.HandleError(fmt.Errorf("failed to create CombinedStatus (ns, name = %v, %v): %w",
+			generatedCombinedStatus.Namespace, generatedCombinedStatus.Name, err))
+		return nil
```
MikeSpreitzer, line 105: Why is this logged instead of returned as an `error`?

pkg/status/combinedstatus-resolver.go
MikeSpreitzer: Good. Thanks.

pkg/status/combinedstatus.go
vMaroon: Strange, I did log it and pushed, not sure what happened. I rebased on upstream and force pushed.

pkg/status/combinedstatus.go
vMaroon: Fixed.

pkg/status/combinedstatus.go
```diff
@@ -61,53 +60,51 @@ func (c *Controller) syncCombinedStatus(ctx context.Context, ref string) error {
...
-			if err != nil {
-				runtime2.HandleError(fmt.Errorf("failed to create CombinedStatus (ns, name = %v, %v): %w",
-					combinedStatus.Namespace, combinedStatus.Name, err))
-				return nil
+	if generatedCombinedStatus.ResourceVersion != "" {
+		csEcho, err := c.wdsKsClient.ControlV1alpha1().CombinedStatuses(generatedCombinedStatus.Namespace).Update(ctx,
+			generatedCombinedStatus, metav1.UpdateOptions{FieldManager: controllerName})
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return nil // the object was deleted during the syncing procedure, event will be handled
```
MikeSpreitzer, line 87: No log message or error on this way out.

pkg/status/combinedstatus-resolution.go
```diff
@@ -357,32 +387,34 @@ func (c *combinedStatusResolution) evaluateWorkStatus(celEvaluator *celEvaluator
...
-	workStatusContent map[string]interface{}, scData *statusCollectorData) (bool, error) {
+	content map[string]interface{}, scData *statusCollectorData) bool {
 	wsData, exists := scData.wecToData[workStatusWECName]
 
-	if workStatusContent == nil && exists { // workstatus is empty/deleted, remove the workstatus data if it exists
+	if content == nil && exists { // workstatus is empty/deleted, remove the workstatus data if it exists
 		delete(scData.wecToData, workStatusWECName)
-		return true, nil
+		return true
 	}
```
MikeSpreitzer, line 396: If `content == nil && !exists`, shouldn't there be a quick return here?

pkg/status/combinedstatus-resolution.go
```diff
@@ -397,9 +429,10 @@ func evaluateWorkStatusAgainstStatusCollectorWriteLocked(celEvaluator *celEvalua
...
 	for _, selectNamedExp := range scData.Select {
-		eval, err := celEvaluator.Evaluate(selectNamedExp.Def, workStatusContent)
+		eval, err := celEvaluator.Evaluate(selectNamedExp.Def, content)
 		if err != nil {
-			return false, fmt.Errorf("failed to evaluate select expression: %w", err)
+			runtime2.HandleError(fmt.Errorf("failed to evaluate select expression: %w", err))
+			continue
 		}
 
 		updated = updated || eval.Equal(wsData.selectEval[selectNamedExp.Name]).Value() != true
```
MikeSpreitzer, line 438: `wsData.selectEval` could be `nil`, right? And `wsData.selectEval[selectNamedExp.Name]` could be `nil`, right?

pkg/status/combinedstatus-resolution.go
```diff
@@ -409,9 +442,10 @@ func evaluateWorkStatusAgainstStatusCollectorWriteLocked(celEvaluator *celEvalua
...
 	for _, groupByNamedExp := range scData.GroupBy {
-		eval, err := celEvaluator.Evaluate(groupByNamedExp.Def, workStatusContent)
+		eval, err := celEvaluator.Evaluate(groupByNamedExp.Def, content)
 		if err != nil {
-			return false, fmt.Errorf("failed to evaluate groupBy expression: %w", err)
+			runtime2.HandleError(fmt.Errorf("failed to evaluate groupBy expression: %w", err))
+			continue
 		}
 
 		updated = updated || eval.Equal(wsData.groupByEval[groupByNamedExp.Name]).Value() != true
```
MikeSpreitzer, line 451: nil pointer faults possible here too, right?

pkg/status/combinedstatus-resolution.go
```diff
@@ -431,9 +465,10 @@ func evaluateWorkStatusAgainstStatusCollectorWriteLocked(celEvaluator *celEvalua
...
 		// evaluate subject which should not be nil since the statuscollector is valid
-		eval, err := celEvaluator.Evaluate(*combinedFieldNamedAgg.Subject, workStatusContent)
+		eval, err := celEvaluator.Evaluate(*combinedFieldNamedAgg.Subject, content)
 		if err != nil {
-			return false, fmt.Errorf("failed to evaluate combinedFields expression: %w", err)
+			runtime2.HandleError(fmt.Errorf("failed to evaluate combinedFields subject expression: %w", err))
+			continue
 		}
 
 		updated = updated || eval.Equal(wsData.combinedFieldsEval[combinedFieldNamedAgg.Name]).Value() != true
```
MikeSpreitzer, line 474: More nil pointer faults?

pkg/status/combinedstatus-resolution.go
```diff
@@ -497,81 +532,31 @@ func validateCombinedStatusLabels(combinedStatus *v1alpha1.CombinedStatus,
...
 
+// handleSelect handles the select expressions of a statuscollector data.
+// This means that the function evaluates the select expressions against the
+// possibly filtered workstatuses and returns the result in a
+// NamedStatusCombination.
 func handleSelect(scName string, scData *statusCollectorData) *v1alpha1.NamedStatusCombination {
 	namedStatusCombination := v1alpha1.NamedStatusCombination{
 		Name:        scName,
-		ColumnNames: make([]string, 0, len(scData.Select)+1), // for now, manually add wecName to all rows
+		ColumnNames: make([]string, 0, len(scData.Select)), // for now, manually add wecName to all rows
```
MikeSpreitzer, line 542: Is the comment here still accurate?

pkg/status/combinedstatus-resolution.go
```diff
@@ -580,73 +565,180 @@ func handleSelect(scName string, scData *statusCollectorData) *v1alpha1.NamedSta
...
+			Type:   v1alpha1.TypeNumber,
+			Number: &numStr,
+		}
+	case bool:
+		return v1alpha1.Value{
+			Type: v1alpha1.TypeBool,
+			Bool: &v,
+		}
+	default:
+		evalJSON, err := json.Marshal(valValue)
```
MikeSpreitzer, line 605: This never produces an array Value.

pkg/status/combinedstatus-resolution.go
```diff
@@ -580,73 +565,180 @@ func handleSelect(scName string, scData *statusCollectorData) *v1alpha1.NamedSta
...
+	ValueToNumber := map[any]int{}
+	orderedTupleToWorkStatusGroup := map[string][]*workStatusData{}
+
+	if len(scData.GroupBy) > 0 {
+		for _, wsData := range scData.wecToData {
+			valuesTuple := make([]string, 0, len(scData.GroupBy))
+			for _, groupByNamedExp := range scData.GroupBy {
+				groupByValue := wsData.groupByEval[groupByNamedExp.Name]
+				// ensure unique number mapping for the value
+				uid, exists := ValueToNumber[groupByValue.Value()]
```
MikeSpreitzer, line 645: I am not sure that Golang equality does what is needed here.

pkg/status/combinedstatus-resolution.go
```diff
@@ -580,73 +565,180 @@ func handleSelect(scName string, scData *statusCollectorData) *v1alpha1.NamedSta
...
-		for value, wsDataGroup := range valueToWorkStatus {
-			rowName := fmt.Sprintf("%s.%v", groupByExpName, value)
-			row := v1alpha1.StatusCombinationRow{
-				Name:    &rowName,
-				Columns: make([]v1alpha1.Value, 0, len(scData.CombinedFields)),
-			}
+	for _, wsDataGroup := range orderedTupleToWorkStatusGroup {
+		if len(wsDataGroup) == 0 {
+			continue
+		}
```
MikeSpreitzer, line 698: What good does this do?

pkg/status/combinedstatus-resolution.go
MikeSpreitzer: I think a return is warranted here if `content == nil`, rather than trying to evaluate a nil row in all the code below.

pkg/status/combinedstatus-resolution.go
```diff
@@ -357,32 +387,34 @@ func (c *combinedStatusResolution) evaluateWorkStatus(celEvaluator *celEvaluator
...
-	workStatusContent map[string]interface{}, scData *statusCollectorData) (bool, error) {
+	content map[string]interface{}, scData *statusCollectorData) bool {
 	wsData, exists := scData.wecToData[workStatusWECName]
 
-	if workStatusContent == nil && exists { // workstatus is empty/deleted, remove the workstatus data if it exists
+	if content == nil && exists { // workstatus is empty/deleted, remove the workstatus data if it exists
 		delete(scData.wecToData, workStatusWECName)
-		return true, nil
+		return true
 	}
```
MikeSpreitzer, line 397: I think a return is warranted here if content == nil, rather than trying to evaluate a nil row in all the code below.

pkg/status/combinedstatus-resolution.go
MikeSpreitzer: I looked at the definition of the `Equal` method of the `Val` interface, and did not notice a promise about how it behaves if `other Val` is `nil`. I think that it would be best if KubeStellar did not rely on undocumented behavior in this case.

pkg/status/combinedstatus-resolution.go
MikeSpreitzer: If Golang equality is the same as the equality needed here, then there is a simpler and more direct approach than using `ValueToNumber` and `orderedTupleToWorkStatusGroup`. Instead it could be something like the following, using a recursive indexed data structure.

```
	grouped := IndexedOrNot{}
	for _, wsData := range scData.wecToData {
		grouped.insert(scData.GroupBy, wsData)
	}
```

where

```
type IndexedOrNot struct {
	indexed map[any]*IndexedOrNot
	not     []*workStatusData
}

func (ion *IndexedOrNot) insert(groupBy []v1alpha1.NamedExpression, wsData *workStatusData) {
	this := ion
	for j := 0; j < len(groupBy); j++ {
		if this.indexed == nil {
			this.indexed = map[any]*IndexedOrNot{}
		}
		groupByVal := wsData.groupByEval[groupBy[j].Name]
		groupByAny := groupByVal.Value()
		next := this.indexed[groupByAny]
		if next == nil {
			next = &IndexedOrNot{}
			this.indexed[groupByAny] = next
		}
		this = next
	}
	this.not = append(this.not, wsData)
}
```

pkg/status/combinedstatus-resolution.go
vMaroon: Fixed.

MikeSpreitzer (approved):
This is progress.
I still wonder whether golang equality on `ref.Val` is the same thing as the semantic equality that we need.

/lgtm
/approve
Thanks!

PR #2297, opened 2024-07-06
🌱 Bump Kubernetes version to 1.28.11 and container-runtime to 0.16.6

MikeSpreitzer:
## Summary
This PR bumps the Kubernetes version to 1.28.11 (the latest in the 1.28 line) and the controller-runtime version to 0.16.6 (the latest that requires only kube 1.28).

## Related issue(s)

Fixes #


cmd/controller-manager/main.go
```diff
@@ -93,9 +95,9 @@ func main() {
 	// manager here is mainly used for leader election and health checks
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                 scheme,
-		MetricsBindAddress:     metricsAddr,
+		Metrics:                crmetrics.Options{BindAddress: metricsAddr},
```
ezrasilvera, line 98: Why is this part of this PR ?

MikeSpreitzer: Because the old way was deprecated and is no longer available.

cmd/controller-manager/main.go
```diff
@@ -31,6 +31,8 @@ import (
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
+	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics/server"
+	crwebhook "sigs.k8s.io/controller-runtime/pkg/webhook"
```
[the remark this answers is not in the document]
MikeSpreitzer: This is part of the change discussed above.

cmd/controller-manager/main.go
MikeSpreitzer: We went through the same thing in https://github.com/kubestellar/ocm-status-addon/pull/76

PR #2318, opened 2024-07-11
:book: Add additional emphasis to website left navigation to reduce confusion

KPRoche:
## Summary
The rendering of the left menu tree in our existing rendered website only uses a color change to indicate the currently active page in the tree. This makes it less obvious to the user how they are traveling through the website.

This PR is a simple change to the css styling to make it much more obvious where a visitor is in the tree.
It also by way of example restores the previous behavior of having the getting started link in the masthead jump you to that as a page in the user guide tree

[The site may be previewed at https://kproche.github.io/kubestellar/doc-nav-emphasis/](https://kproche.github.io/kubestellar/doc-nav-emphasis/)

## How it looks before this PR
image

## How it looks after this PR

image

### DIFF
--- a/docs/overrides/home.html
+++ b/docs/overrides/home.html
@@ -10,7 +10,7 @@
     <div class="px-1 ctx-header-text conteiner">
       
        <span class="shortcut-bar Space-Bd-BT"> 
-        <a class="shortcut-bar-text" href="direct/get-started/" >START</a>&nbsp;&nbsp;&nbsp;
+        <a class="shortcut-bar-text" href="direct/get-started/">START</a>&nbsp;&nbsp;&nbsp;
         <a class="shortcut-bar-text" href="readme/">ABOUT</a>&nbsp;&nbsp;&nbsp;
         <a class="shortcut-bar-text" href="direct/user-guide-intro/">USE</a>&nbsp;&nbsp;&nbsp;
         <a class="shortcut-bar-text" href="direct/contribute/">CONTRIBUTE</a>&nbsp;&nbsp;&nbsp;&nbsp;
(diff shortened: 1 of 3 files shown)


docs/overrides/home.html
```diff
@@ -10,7 +10,7 @@
     <div class="px-1 ctx-header-text conteiner">
       
        <span class="shortcut-bar Space-Bd-BT"> 
-        <a class="shortcut-bar-text" href="direct/get-started/">START</a>&nbsp;&nbsp;&nbsp;
+        <a class="shortcut-bar-text" href="direct/get-started/" >START</a>&nbsp;&nbsp;&nbsp;
```
MikeSpreitzer, line 13: Is adding this space intentional? What difference does it make?

MikeSpreitzer (commented):
/lgtm
(But go ahead and delete that unintentional space, if it is unintentional.)

clubanderson:
@KPRoche the margins flip flop based on the footer (or something) as you click around. I think this needs to be addressed before going forward, but I like the highlighting.

docs/overrides/home.html
KPRoche: Oops. That was an artifact of attempting to add a hover tooltip and then deleting the tag.

KPRoche:
> @KPRoche the margins flip flop based on the footer (or something) as you click around. I think this needs to be addressed before going forward, but I like the highlighting.

They actually do it in the existing docs, too, just not as dramatically. If you work your way through the user guide pages you can see them move. This is somewhere in the main template for the site.

MikeSpreitzer (commented):
/lgtm

ezrasilvera:
Very nice 
/approve

PR #2324, opened 2024-07-12
:book: Update website landing page (home.html) for better clarity and accessibility

### LINKED ISSUE #2319: bug: Rename "Documentation" to "User Guide" on Landing Page
### Describe the bug

Kubestellar users can be referred to the documentation or the user guides without knowing they are the same. Using a search engine with these different terms will lead to different results.

@KPRoche 

### Steps To Reproduce

1. Go to Landing preview page [https://kproche.github.io/kubestellar/doc-nav-emphasis/](https://kproche.github.io/kubestellar/doc-nav-emphasis/)
2. View and click "Documentation" button.
3. Wait for redirect to Kubestellar "User Guide".

### Expected Behavior

Documentation titles should match.

image
image

OR

image
image

### Want to contribute?

### Additional Context

The alt text should match the button text or provide a description for those users using screen readers.

KPRoche:
## The landing page for the site (docs/overrides/home.html) had some labeling and attributes (or missing attributes) that were either confusing or could cause accessibility problems.

This PR cleans up bad title attributes on all 3 of the buttons on the landing page, and makes sure that all graphical elements have alt, title, or aria-labelledby attributes defined for improved accessibility and clarity.

(It also incorporates the CSS menu highlight modification in PR #2318)

The site can be previewed at https://kproche.github.io/kubestellar/doc-alttext/
## Related issue(s)

Fixes #2319


MikeSpreitzer:
Here are two snapshots from my screen, looking at release-0.23.1 and this PR's rendering. There is some unintentional overlap in this PR's rendering, maybe?

Screenshot 2024-07-12 at 2 22 52 PM

Screenshot 2024-07-12 at 2 22 29 PM

KPRoche:
> Here are two snapshots from my screen, looking at release-0.23.1 and this PR's rendering. There is some unintentional overlap in this PR's rendering, maybe, at the top and at the bottom?

That happens on the landing page already whenever you are in the non-latest version with a reduced width screen. I have not found a way to fix the yellow warning banner/top menu shortcut overlap reliably in the css or in the html

MikeSpreitzer:
@KPRoche: look also at the bottom, where the picture is on top of some text.

KPRoche:
> look also at the bottom, where the picture is on top of some text.
These are not new problems in the responsive layout in home.html; I think you and more people are observing them now. It definitely has room for improvement

MikeSpreitzer:
Another comparison, even narrower.

release-0.23.1:
Screenshot 2024-07-12 at 2 51 29 PM

This PR:
Screenshot 2024-07-12 at 2 50 31 PM

MikeSpreitzer:
Now here is something _really_ weird: I took this screenshot on my Mac. It has an active area around "docs.kubestellar.io" below the left button!
And there is no corresponding active area below the right button.

Screenshot 2024-07-12 at 3 59 35 PM

MikeSpreitzer:
Back to ordinary weirdness. Here are four pictures I took with a cellphone, so that you can see the cursor. They are exploring where I can actually click on the text below the buttons. At the browser widths shown, all four results are disappointing --- three of them completely so.

This one shows the cursor turned to a pointing finger over the left part of "docs.kubestellar.io":

PXL_20240712_200159121

And this one shows the cursor in its normal appearance over the right part of the same string:

PXL_20240712_200206956

There is no active area on the right side text. Here is testing one part of it:

PXL_20240712_200216929

At this other width, even the left part of "docs.kubestellar.io" is not active:

PXL_20240712_200445651

clubanderson:
/lgtm

Thank you for fixing the version number in bash2curl.

MikeSpreitzer:
The latest revision is better.

I tested at a large variety of browser window widths, and there are still three ranges of widths that have a problem.

At the very narrowest allowed, the text/link under the first button is invisible.

Screenshot 2024-07-13 at 11 44 22 PM

Making the window a bit wider, we get to a range where the links under the buttons are in the shadows of the buttons.

Screenshot 2024-07-13 at 11 44 59 PM

Going wider, we get to a range where the links/text under the buttons are also under the core+WECs picture and are not active (do not bother looking for the cursor in the screenshot below, it is not visible in this capture).

Screenshot 2024-07-13 at 11 45 15 PM

KPRoche:
@MikeSpreitzer  thank you for mapping out problems in the responsive layout. 
I'm pretty sure some of these exist in the original layout of the landing page but we weren't exercising it as hard, or they just weren't as apparent because they become more obvious when we have normal text in the flexible elements. (for instance, the earlier problem with only the left side of the docs.kubestellar.io link working was from the right-floated shortcut menu bar at the top being spec'ed as having 90% height, so it masked the elements below)

Solving them may be as simple as tweaking some margin settings, or figuring out how to specify which elements should be on the "top" layer

KPRoche:
I have managed to solve everything but the button overlap on the narrowest screen setting.

KPRoche:
OK, I think I have fixed all the margins and z-index attributes so all the text links and buttons remain usable as the screen reflows.

KPRoche:
One other item... the warning banner overlapping the top menu. I've elevated the z-index of the top menu so it has priority, but I haven't yet found the correct element attribute to prevent the overlap.

ezrasilvera:
There are some conflicts in the PR 
/lgtm

KPRoche:
> There are some conflicts in the PR /lgtm
@clubanderson had some lingering reservations about the menu label “overview” for the intro to the user guide, so I updated that to “Guide Overview”

KPRoche:
I may need a little hand-holding to fix the conflict (simple as it is) in accordance with best practices.

PR #2325, opened 2024-07-14
📖 Clarify status combining feature

MikeSpreitzer:
## Summary
This PR updates the user documentation about the combined status feature, in a couple of ways. One is to track developments that have happen since this doc was written. Another is to hopefully make it more precise and clear.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-update-713/direct/combined-status/

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/direct/combined-status.md
+++ b/docs/content/direct/combined-status.md
@@ -4,15 +4,20 @@ Note on terminology: the general idea that we wish to address is returning repor
 
 There are two methods of returning reported state: a general one and a special case. The general method returns reported state from any number of WECs. The special case applies when the number of WECs is exactly 1, and returns the reported state into the original object in the WDS.
 It may make sense to expand the latter into a general feature in the future, accommodating features such as scheduling.
+
 ## Introduction to the General Technique
 
 The general technique involves the following ideas.
 
-1. The way that reported state is combined is specified by the user, in a simple but powerful way modeled on SQL. This is chosen because it is a well worked out set of ideas, is widely known, and is something that we may someday want to use in our implementation. We do not need to support anything like full SQL (for any version of SQL). This proposal only involves one particular pattern of relatively simple SELECT statement, and a small subset of the expression language.
+1. The way that reported state is combined is specified by the user, in a simple but powerful way modeled on SQL. This is chosen because it is a well worked out set of ideas, is widely known, and is something that we may someday want to use in our implementation. We do not need to support anything like full SQL (for any version of SQL). This proposal only involves one particular pattern of relatively simple SELECT statement, and a different expression language (CEL, which is the most prominent expression language in the Kubernetes milieu).
 
-1. The specification of how to combine reported state is defined in `StatusCollector` objects. These objects are referenced in the BindingPolicies right next to the criteria for selecting workload objects. This saves users the trouble of having to write selection criteria twice. With the specification being separate, it is possible to have a library of `StatusCollectors` that can be reused across different BindingPolicies. In the future KubeStellar would provide a library of `StatusCollector` objects that cover convenient use-cases for kubernetes built-in resources such as deployments.
+1. The expressions may reference the content of the workload object as it sits in the WDS as well as the state returned from the WECs.
 
-1. The combined reported state appears in a new kind of object, one per (workload object, BindingPolicy object) pair.
+1. The specification of how to combine reported state is defined in `StatusCollector` objects. These objects are referenced in the `BindingPolicy` objects right next to the criteria for selecting workload objects. This saves users the trouble of having to write selection criteria twice. With the specification being separate rather than embedded, it is possible to have a library of `StatusCollectors` that can be reused across different `BindingPolicy` objects. In the future KubeStellar would provide a library of `StatusCollector` objects that cover convenient use-cases for kubernetes built-in resources such as deployments.
+
+1. The `Binding` objects also hold references to `StatusCollector` objects. Each reference to a workload object is paired with references to all the `StatusCollectors` mentioned in all the `DownsyncObjectTestAndStatusCollection` structs that matched the workload object.
+
+1. The combined reported state appears in a new kind of object, one per (workload object, `Binding` object) pair.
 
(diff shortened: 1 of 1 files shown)


docs/content/direct/combined-status.md
```diff
@@ -8,38 +8,81 @@ It may make sense to expand the latter into a general feature in the future, acc
...
+FROM PerWEC WHERE <filter condition>
+LIMIT <limit>
+```
+
+##### Aggregation without GROUP BY
+
+When there is aggregation but no plain selection and _no_ "GROUP BY", this is equivalent to the following form of SELECT statement.
+
+```sql
+SELECT <aggregation columns>
```
vMaroon, line 68: I think it would be helpful to expand on what aggregation columns are. Lines 42-43 give the idea, but it would be helpful to show the full equivalent SQL example in which a function is applied, at least once.

MikeSpreitzer: I have updated #2325

docs/content/direct/combined-status.md
```diff
@@ -8,38 +8,85 @@ It may make sense to expand the latter into a general feature in the future, acc
 
 The general technique involves the following ideas.
 
-1. The way that reported state is combined is specified by the user, in a simple but powerful way modeled on SQL. This is chosen because it is a well worked out set of ideas, is widely known, and is something that we may someday want to use in our implementation. We do not need to support anything like full SQL (for any version of SQL). This proposal only involves one particular pattern of relatively simple SELECT statement, and a small subset of the expression language.
+1. The way that reported state is combined is specified by the user, in a simple but powerful way modeled on SQL. This is chosen because it is a well worked out set of ideas, is widely known, and is something that we may someday want to use in our implementation. We do not need to support anything like full SQL (for any version of SQL). This proposal only involves one particular pattern of relatively simple SELECT statement, and a different expression language (one of growing currency in the Kubernetes milieu).
```
ezrasilvera, line 11: Should we mention which "different expression language"  ?

docs/content/direct/combined-status.md
```diff
@@ -8,38 +8,85 @@ It may make sense to expand the latter into a general feature in the future, acc
...
 
-1. The specification of how to combine reported state is defined in `StatusCollector` objects. These objects are referenced in the BindingPolicies right next to the criteria for selecting workload objects. This saves users the trouble of having to write selection criteria twice. With the specification being separate, it is possible to have a library of `StatusCollectors` that can be reused across different BindingPolicies. In the future KubeStellar would provide a library of `StatusCollector` objects that cover convenient use-cases for kubernetes built-in resources such as deployments.
+1. The expressions may reference the content of the workload object as it sits in the WDS as well as the state returned from the WECs.
```
ezrasilvera, line 13: I guess the as well as refers to the "content ..."   ?

docs/content/direct/combined-status.md
```diff
@@ -8,38 +8,85 @@ It may make sense to expand the latter into a general feature in the future, acc
...
 
-1. The combined reported state appears in a new kind of object, one per (workload object, BindingPolicy object) pair.
+1. The specification of how to combine reported state is defined in `StatusCollector` objects. These objects are referenced in the `BindingPolicy` objects right next to the criteria for selecting workload objects. This saves users the trouble of having to write selection criteria twice. With the specification being separate, it is possible to have a library of `StatusCollectors` that can be reused across different `BindingPolicy` objects. In the future KubeStellar would provide a library of `StatusCollector` objects that cover convenient use-cases for kubernetes built-in resources such as deployments.
```
ezrasilvera, line 15: Maybe the emphasis should be "With the specification being referenced from a separate definition rather than embedded..."

docs/content/direct/combined-status.md
```diff
@@ -8,38 +8,85 @@ It may make sense to expand the latter into a general feature in the future, acc
...
+
+1. The `Binding` objects also hold references to `StatusCollector` objects. Each reference to a workload object is paired with references to all the `StatusCollectors` mentioned in all the `DownsyncObjectTestAndStatusCollection` structs that matched the workload object.
```
ezrasilvera, line 17: I think the main point of the doc is to explain the "why"  (e.g., usage possibilities) , I'm not sure there is a point in just listing/specifying what are the object's fields

docs/content/direct/combined-status.md
```diff
@@ -8,38 +8,85 @@ It may make sense to expand the latter into a general feature in the future, acc
...
+      expressions is implicitly included in the output columns.
 
-   - A collection of named expressions using aggregation functions to define
-     additional output columns.
+    - A collection of named expressions using aggregation functions to define
+      additional output columns.
 
 1. The SELECT statement has a LIMIT on the number of rows that it will yield.
 
+#### Relation with SQL, more precisely
```
ezrasilvera, line 47: This is already sub section of relation with SQL. So maybe "Detailed description" ?

docs/content/direct/combined-status.md
MikeSpreitzer: I am not sure I understand the question. "as well as" is not a reference, it is a joiner.

docs/content/direct/combined-status.md
MikeSpreitzer: This is the whole user-facing document about this feature, so it needs to explain not only the "why" but also lay out the relevant concepts and how they fit together. The paragraph that is comment is attached to is not a list of fields of an object but is talking about how the content of a Binding is related to the content of a BindingPolicy. In particular, this paragraph is pointing out that multiple DownsyncObjectTestAndStatusCollection structs in a BindingPolicy can match the same workload object and in this case their StatusCollector references are combined into one set of StatusCollector references in the corresponding position in the Binding object.

docs/content/direct/combined-status.md
MikeSpreitzer: Yes, updated.

docs/content/direct/combined-status.md
MikeSpreitzer: Updated.

docs/content/direct/combined-status.md
MikeSpreitzer: Revised.

ezrasilvera:
@MikeSpreitzer It's LGTM - will let @vMaroon lgtm/approve when he is done as well.

PR #2326, opened 2024-07-15
:sparkles: StatusCollector validation & error propagation to CR

vMaroon:
## Summary
This PR implements [StatusCollector validation](https://github.com/kubestellar/kubestellar/pull/2174#discussion_r1630076164) as outlined in #2141 

It also involves minor code refactoring.

## Related issue(s)
- #2141

### DIFF
--- a/pkg/status/statuscollector.go
+++ b/pkg/status/statuscollector.go
@@ -46,13 +46,13 @@ func (c *Controller) syncStatusCollector(ctx context.Context, ref string) error 
 
 	// Validate the StatusCollector
 	if errs := c.validateStatusCollector(statusCollector); len(errs) > 0 {
-		if err := c.updateStatusCollectorErrors(ctx, statusCollector, errs); err != nil {
+		if err := c.updateStatusCollectorErrors(ctx, statusCollector.DeepCopy(), errs); err != nil {
 			return err
 		}
 
-		isDeleted = true // invalid statuscollector, should be deleted if it exists
+		isDeleted = true // invalid statuscollector, treated as if it doesn't exist
 	} else if statusCollector.Status.Errors != nil {
-		if err := c.updateStatusCollectorErrors(ctx, statusCollector, nil); err != nil {
+		if err := c.updateStatusCollectorErrors(ctx, statusCollector.DeepCopy(), nil); err != nil {
 			return err
 		}
 	}
@@ -66,6 +66,9 @@ func (c *Controller) syncStatusCollector(ctx context.Context, ref string) error 
 	return nil
 }
 
+// validateStatusCollector validates the StatusCollector resource
+// and returns a list of errors if any.
+// The passed statuscollector is not mutated.
 func (c *Controller) validateStatusCollector(statusCollector *v1alpha1.StatusCollector) []error {
 	var errs []error
 	// groupBy & CombinedFields empty if select is not
@@ -129,6 +132,8 @@ func (c *Controller) validateStatusCollector(statusCollector *v1alpha1.StatusCol
 	return errs
 }
 
+// updateStatusCollectorErrors updates the StatusCollector status with the
+// given errors. The given statuscollector is mutated.
 func (c *Controller) updateStatusCollectorErrors(ctx context.Context, statusCollector *v1alpha1.StatusCollector,
 	errs []error) error {
 	logger := klog.FromContext(ctx)
@@ -142,7 +147,8 @@ func (c *Controller) updateStatusCollectorErrors(ctx context.Context, statusColl
 
 	if err != nil {
 		if errors.IsNotFound(err) {
-			// StatusCollector has been deleted
+			logger.V(4).Info("StatusCollector not found (status updating skipped)",
+				"ns", statusCollector.Namespace, "name", statusCollector.Name)
 			return nil
 		} else {
 			return fmt.Errorf("failed to update StatusCollector status (ns, name = %s, %s): %w",
--- a/pkg/status/statuscollector.go
+++ b/pkg/status/statuscollector.go
@@ -19,12 +19,13 @@ package status
 import (
 	"context"
 	"fmt"
-	"github.com/kubestellar/kubestellar/pkg/abstract"
+
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 
 	"github.com/kubestellar/kubestellar/api/control/v1alpha1"
+	"github.com/kubestellar/kubestellar/pkg/abstract"
 )
 
 func (c *Controller) syncStatusCollector(ctx context.Context, ref string) error {
--- a/pkg/status/statuscollector.go
+++ b/pkg/status/statuscollector.go
@@ -18,8 +18,10 @@ package status
 
 import (
 	"context"
-
+	"fmt"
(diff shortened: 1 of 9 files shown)


pkg/status/combinedstatus-resolver.go
```diff
@@ -242,9 +242,6 @@ func (c *combinedStatusResolver) NoteBindingResolution(bindingName string, bindi
 			c.resolutionNameToKey[csResolution.getName()] = resolutionKey{bindingName, objectIdentifier}
 		}
 
-		// fetch missing statuscollector specs
-		c.fetchMissingStatusCollectorSpecsLocked(statusCollectorLister, objectData.StatusCollectors)
```
vMaroon, line 246: Missing collectors should be pending/about to be pending in the workqueue. Removal is simpler than validating them here as well.

MikeSpreitzer:
Is this PR listed in #2141?

pkg/status/statuscollector.go
```diff
@@ -41,7 +44,19 @@ func (c *Controller) syncStatusCollector(ctx context.Context, ref string) error
...
 	}
 
-	// TODO: validate the statusCollector if not nil
+	// Validate the StatusCollector
+	if errs := c.validateStatusCollector(statusCollector); len(errs) > 0 {
+		if err := c.updateStatusCollectorErrors(ctx, statusCollector, errs); err != nil {
+			return err
+		}
+
+		isDeleted = true // invalid statuscollector, should be deleted if it exists
```
MikeSpreitzer, line 53: I presume you mean that this method should pretend that the StatusCollector does not exist, rather than actively delete it. If so then more clarity in the comment would be good.

pkg/status/statuscollector.go
```diff
@@ -50,3 +65,92 @@ func (c *Controller) syncStatusCollector(ctx context.Context, ref string) error
 	logger.Info("Synced StatusCollector", "ref", ref)
 	return nil
 }
+
+func (c *Controller) validateStatusCollector(statusCollector *v1alpha1.StatusCollector) []error {
+	var errs []error
+	// groupBy & CombinedFields empty if select is not
```
MikeSpreitzer, line 71: Wouldn't it be better to inject these empty/non-empty constraints into the OpenAPI schema?

pkg/status/statuscollector.go
```diff
@@ -50,3 +65,92 @@ func (c *Controller) syncStatusCollector(ctx context.Context, ref string) error
...
+
+	return errs
+}
+
+func (c *Controller) updateStatusCollectorErrors(ctx context.Context, statusCollector *v1alpha1.StatusCollector,
+	errs []error) error {
+	logger := klog.FromContext(ctx)
+
+	statusCollector.Status.Errors = abstract.SliceMap(errs, func(err error) string {
+		return err.Error()
```
MikeSpreitzer, line 137: Wouldn't it be simpler to have the origins above directly produce strings?

pkg/status/statuscollector.go
```diff
@@ -41,7 +44,19 @@ func (c *Controller) syncStatusCollector(ctx context.Context, ref string) error
 		statusCollector.Name = ref
 	}
 
-	// TODO: validate the statusCollector if not nil
+	// Validate the StatusCollector
+	if errs := c.validateStatusCollector(statusCollector); len(errs) > 0 {
+		if err := c.updateStatusCollectorErrors(ctx, statusCollector, errs); err != nil {
```
MikeSpreitzer, line 49: It might be worth putting a comment on the variable `statusCollector` about its immutability (see https://github.com/kubestellar/kubestellar/blob/08d290a0dbcfe91714d7e140cd80630c5ed843b9/pkg/generated/listers/control/v1alpha1/statuscollector.go#L36).

pkg/status/statuscollector.go
```diff
@@ -50,3 +65,92 @@ func (c *Controller) syncStatusCollector(ctx context.Context, ref string) error
...
+		if err := c.celEvaluator.CheckExpression(combinedField.Subject); err != nil {
+			errs = append(errs, fmt.Errorf("combinedField expression (%s) subject invalid: %w",
+				combinedField.Name, err))
+		}
+	}
+
+	return errs
+}
+
+func (c *Controller) updateStatusCollectorErrors(ctx context.Context, statusCollector *v1alpha1.StatusCollector,
```
MikeSpreitzer, line 132: Since the value of `statusCollector` here comes from https://github.com/kubestellar/kubestellar/blob/08d290a0dbcfe91714d7e140cd80630c5ed843b9/pkg/generated/listers/control/v1alpha1/statuscollector.go#L36 , it is worth noting its immutability.

pkg/status/statuscollector.go
```diff
@@ -50,3 +65,92 @@ func (c *Controller) syncStatusCollector(ctx context.Context, ref string) error
...
+	errs []error) error {
+	logger := klog.FromContext(ctx)
+
+	statusCollector.Status.Errors = abstract.SliceMap(errs, func(err error) string {
```
MikeSpreitzer, line 136: `statusCollector` is immutable.

pkg/status/statuscollector.go
```diff
@@ -50,3 +65,92 @@ func (c *Controller) syncStatusCollector(ctx context.Context, ref string) error
...
+		return err.Error()
+	})
+
+	scEcho, err := c.wdsKsClient.ControlV1alpha1().StatusCollectors().UpdateStatus(ctx,
+		statusCollector, metav1.UpdateOptions{FieldManager: controllerName})
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// StatusCollector has been deleted
+			return nil
```
MikeSpreitzer, line 146: I would log something in this case.

pkg/status/statuscollector.go
vMaroon: While that's true, the current implementation is clearer logically in my opinion. If you disagree and insist, I can change to strings.

MikeSpreitzer: No, I do not insist.

pkg/status/statuscollector.go
```diff
@@ -50,3 +65,98 @@ func (c *Controller) syncStatusCollector(ctx context.Context, ref string) error
...
+// given errors. The given statuscollector is mutated.
+func (c *Controller) updateStatusCollectorErrors(ctx context.Context, statusCollector *v1alpha1.StatusCollector,
+	errs []error) error {
+	logger := klog.FromContext(ctx)
+
+	statusCollector.Status.Errors = abstract.SliceMap(errs, func(err error) string {
+		return err.Error()
+	})
+
+	scEcho, err := c.wdsKsClient.ControlV1alpha1().StatusCollectors().UpdateStatus(ctx,
```
MikeSpreitzer, line 145: FYI, it is not obvious from its functionality, but it is a surprising fact about the client stubs that this can also mutate `statusCollector`.

pkg/status/statuscollector.go
vMaroon: I gave it a go and it did not work out as expected, I rather not handle this now. Can open issue on using openapi validations where makes sense across our API CRDs.

MikeSpreitzer: OK

MikeSpreitzer (approved):
This is good enough to merge.
/lgtm
/approve
Thanks!

PR #2330, opened 2024-07-17
📖 Start writing troubleshooting guide

MikeSpreitzer:
## Summary
This PR introduces the start on a troubleshooting guide.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-troubleshooting/direct/troubleshooting/

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/direct/troubleshooting.md
+++ b/docs/content/direct/troubleshooting.md
@@ -4,9 +4,9 @@ This guide is a work in progress.
 
 ## Things to look at
 
-- Remember that for each of your BindingPolicy objects, there is a corresponding Binding object that reports what matched the policy object.
+- Remember that for each of your BindingPolicy objects, there is a corresponding Binding object that reports what is matching the policy object.
 - Although not part of the interface, when debugging you can look at the ManifestWork and WorkStatus objects in the ITS.
-- Look at logs of controllers. If they have had container restarts that look relevant, look also at the previous logs.
+- Look at logs of controllers. If they have had container restarts that look relevant, look also at the previous logs. Do not forget OCM controllers.
 - If a controller's `-v` is not at least 4, increase it.
 - Remember that Kubernetes controllers tend to report transient problems as errors without making it clear that the problem is transient and tend to not make it clear if/when the problem has been resolved (sigh).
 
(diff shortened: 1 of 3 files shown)


docs/content/direct/troubleshooting.md
```diff
@@ -0,0 +1,27 @@
...
+- Although not part of the interface, when debugging you can look at the ManifestWork and WorkStatus objects in the ITS.
+- Look at logs of controllers. If they have had container restarts that look relevant, look also at the previous logs. Do not forget OCM controllers.
+- If a controller's `-v` is not at least 4, increase it.
+- Remember that Kubernetes controllers tend to report transient problems as errors without making it clear that the problem is transient and tend to not make it clear if/when the problem has been resolved (sigh).
+
+## Making a good trouble report
+
+Basic configuration information.
+
+- Include the versions of all the relevant software; do not forget the OCM pieces.
```
vMaroon, line 17: I think this, plus the user-friendly display of combinedstatuses, can kickoff the writing of KS cli tools.

vMaroon (commented):
Overall this is nice. I think that the parts that deal with reporting can be streamlined by templates and cli support (e.g., generate a report of setup + versions).

MikeSpreitzer:
Yes, I definitely think that tools to help users gather helpful information are also interesting. I would like to leave the tools to a separate PR.

clubanderson:
I like the concept, there needs to be lots of pre-formatted commands that guide a user into what kinds of information can be captured and how to, prescriptively, capture it. It is not enough to say "gather info on binding policy objects" - you must add a kubectl command (complete with context) that can be used to explicitly get info on bp objects... and on its goes.

KPRoche:
If we have any documented "gotchas" that crop up repeatedly, we should have a section for that, even if it repeats information in some other documents.
One immediate candidate that comes to mind is the Kind "Pod errors due to too many open files" problem.

MikeSpreitzer:
@KPRoche: Yes, I think that a "known issues" compendium would be good. It seems to me to be a little different content than how to deal with an unknown problem. I think that perhaps the known issues should go in a separate document.

@clubanderson: yes, I think that showing commands for doing the investigation is a good idea. I am not entirely sure what is the best approach. Should this document start with an outline like is currently present and then have a section with details, or should the outline be interrupted by all the details? Some things are repeated: displaying the contents of an object, discovering the containers in a pod, getting the latest log from each container in a pod, getting the previous log from each container in a pod, getting from Deployment to Pod. I really do not think we want to duplicate all the common material; we will need to find a way so that the common things are written only once. Can we merge the existing outline and work on adding the detailed commands in follow-on PRs?

clubanderson:
@KPRoche - indeed - incremental progress is best - follow-on!

Issue #2336, opened 2024-07-18
feature: user can mark a workload object as "create-only"


MikeSpreitzer:
### Feature Description

Motivation: This is a simple way to deal with the problem of multiple controllers writing into the `.spec` of a workload object, some controllers writing to the WDS and some writing to the WEC. The application arm of OCM has only one way to support this, and it is marking an object so that the klusterlet only creates the object if it does not already exist, does not update the object once it exists.

Feature: a user can indicate that a workload object should be propagated in create-only mode. This mode means that the object will be created on the WEC if the object does not already exist there, but will not be updated on the WEC once it exists there. This mode does not change deletion; an object marked for create-only mode will still be deleted when the user no longer wants it to be projected to the WEC.

### Proposed Solution

Modify `BindingPolicy` so that the downsync spec associates a create-only bit with a `DownsyncObjectTest`. That is, replace `DownsyncObjectTestAndStatusCollection` (which pairs a `DownsyncObjectTest` with a `StatusCollectors []string`) with a new type that associates a `DownsyncObjectTestAndStatusCollection` with both a `StatusCollectors []string` and a `CreateOnly bool`.

Define what happens when multiple of those in one `BindingPolicy` match the same workload object. The definition will be that the CreateOnly bits are ORed together.

Make the analogous generalization in `Binding`.

Define what happens when multiple `Binding` objects match the same workload object with different CreateOnly bits. The definition will be that the Bindings with CreateOnly==true will cause at most one write, while any Bindings with CreateOnly==false will each lead to attempted continual maintenance of the object's spec on the WEC.

Because this is an interface change and a new feature, it can not be introduced in a patch release. Because we are at major version 0, this can be introduced with a new minor version number. This breaks Kubernetes rules for evolving an API object type, but we are not following those yet. We continue to not yet support in-place upgrades.

### Want to contribute?

- [X] I would like to work on this issue.

### Additional Context

This feature will give users a way to avoid one of the problems reported in #2332, which is that the klusterlet becomes wedged wrt a ManifestWork if it gets into a fight with another controller on the WEC.

MikeSpreitzer:
The implementation is done. The only thing lacking is better documentation. It is currently mentioned in https://github.com/kubestellar/kubestellar/blob/3e9a3be05747a83f2a8324421216df2a1528f7a1/docs/content/direct/binding.md#bindingpolicy

PR #2338, opened 2024-07-18
✨ Introduce create-only bit in downsync API

MikeSpreitzer:
## Summary
This PR introduces the API changes for a feature whereby a user can mark a workload object for propagation in "create-only" mode (#2336).

I plan to update the binding controller and generic transport controller in later PRs.

This feature can not be completed until a release of this repo is made, then changes are made in https://github.com/kubestellar/ocm-transport-plugin (https://github.com/kubestellar/ocm-transport-plugin/issues/34) using this new API, then a release of that repo, then this repo is updated to use that release.

## Related issue(s)

This addresses part of #2336.


pkg/transport/generic_transport_controller.go
```diff
@@ -847,8 +852,19 @@ func (c *genericTransportController) computeDestToCustomizedObjects(objectsToPro
 	return destToCustomizedObjects, bindingErrors
 }
 
+func TODO_FIXME_ADD_CONSTANT_CREATEONLY(obj *unstructured.Unstructured) Wrapee {
```
vMaroon, line 855: FYI: The standard is to leave a comment that includes "TODO", which is followed by modern IDEs to list pending work.

pkg/transport/generic_transport_controller.go
```diff
@@ -847,8 +852,19 @@ func (c *genericTransportController) computeDestToCustomizedObjects(objectsToPro
...
+	return Wrapee{obj, false}
+}
+
 func (c *genericTransportController) wrapBatch(batchToPropagate []*unstructured.Unstructured, binding *v1alpha1.Binding, numShard int, isSharded bool) (*unstructured.Unstructured, error) {
-	wrappedObject, err := convertObjectToUnstructured(c.transport.WrapObjects(batchToPropagate))
+	var wrapped runtime.Object
+	if t2, is := c.transport.(TransportWithCreateOnly); is {
```
vMaroon, line 861: Does this not beat the point of "generic" in `genericTransportController`? Shouldn't the difference in functionality be abstracted in the interface and implemented twice?

Due to the urgency (if i'm not mistaken) - it is fine to keep it as-is but add a TODO or document this in any way - that is assuming you agree.

pkg/transport/transport.go
```diff
@@ -26,3 +26,27 @@ type Transport interface {
...
 }
+
+// TransportWithCreateOnly is a subtype of Transport that knows how to handle
+// the create-only option.
+type TransportWithCreateOnly interface {
+	Transport
+
+	// WrapObjectsHavingCreateOnly gets slice of Wrapee and wraps them into a single wrapped object.
+	// In case slice is empty, the function should return an empty wrapped object.
+	WrapObjectsHavingCreateOnly(objects []Wrapee) runtime.Object
```
vMaroon, line 37: Does this function filter out non create-only wrapees or it expects the list to be of create-only enabled wrapees? The behavior is not clear.

vMaroon (commented):
Does the introduction of new transport type to solve the problem imply that some transports may implement the `Transport` interface and not the new one?

MikeSpreitzer:
@vMaroon: I see that the transition plan was not obvious. I have pushed a revision that includes an explicit outline of the plan. That should address the points you raised above.

MikeSpreitzer:
A little more detail on the problem: this repo (ks/ks) is tested together with the latest release of ks/OTP --- which does not implement the new wrapping method. So we have to break it into multiple steps, where each step does not break the other repo and building creates a system that works end-to-end (at least for the older functionality).

pkg/transport/generic_transport_controller.go
vMaroon: Resolved

pkg/transport/generic_transport_controller.go
vMaroon: Resolved

vMaroon:
Thanks @MikeSpreitzer that makes more sense.
There is 1 concern in the "unresolved" comment then good to go from my side.

pkg/transport/transport.go
MikeSpreitzer: The intent is that this method honors the create-only bit, correctly handling either value.

MikeSpreitzer: I pushed a revision that spells this out.

vMaroon: Thanks. So this means that once the transition is complete, the name will revert back to `WrapObjects`?

MikeSpreitzer: I was not planning on that but we can make that transition too with additional releases.

PR #2343, opened 2024-07-22
:sparkles: More combined status testing

andreyod:
## Summary
1. Additional combined status e2e test for the `obj` identifier in CEL expression.
2. Small fix to the combined status doc.

## Related issue(s)

Fixes #

### DIFF
--- a/test/e2e/ginkgo/multiple_cluster_deployment_test.go
+++ b/test/e2e/ginkgo/multiple_cluster_deployment_test.go
@@ -563,9 +563,9 @@ var _ = ginkgo.Describe("end to end testing", func() {
 			gomega.Expect(*cs.Results[0].Rows[0].Columns[0].Number).To(gomega.Equal("2"))
 		})
 
-		ginkgo.It("list of WECs with available nginx deployment", func(ctx context.Context) {
+		ginkgo.It("can list all the WECs where the reported number of availableReplicas equals the desired number of replicas", func(ctx context.Context) {
 
-			availableNginxCEL := ksapi.Expression("obj.spec.replicas == returned.status.availableReplicas && obj.metadata.labels['app.kubernetes.io/name'] == 'nginx'")
+			availableNginxCEL := ksapi.Expression("obj.spec.replicas == returned.status.availableReplicas")
 			util.CreateStatusCollector(ctx, ksWds, listNginxWecsStatusCollectorName,
 				ksapi.StatusCollectorSpec{
 					Filter: &availableNginxCEL,
@@ -591,6 +591,16 @@ var _ = ginkgo.Describe("end to end testing", func() {
 			gomega.Expect(cs.Results[0].ColumnNames[0]).To(gomega.Equal("wecName"))
 
 			gomega.ExpectWithOffset(1, len(cs.Results[0].Rows)).To(gomega.Equal(2))
+
+			// we don't know which row will hold data for which WEC
+			row0expectedWec := "cluster1"
+			row1expectedWec := "cluster2"
+			if *cs.Results[0].Rows[0].Columns[0].String != row0expectedWec {
+				row0expectedWec = "cluster2"
+				row1expectedWec = "cluster1"
+			}
+			gomega.Expect(*cs.Results[0].Rows[0].Columns[0].String).To(gomega.Equal(row0expectedWec))
+			gomega.Expect(*cs.Results[0].Rows[1].Columns[0].String).To(gomega.Equal(row1expectedWec))
 		})
 
 		ginkgo.It("can support multiple StatusCollectors", func(ctx context.Context) {
--- a/test/e2e/ginkgo/multiple_cluster_deployment_test.go
+++ b/test/e2e/ginkgo/multiple_cluster_deployment_test.go
@@ -474,6 +474,7 @@ var _ = ginkgo.Describe("end to end testing", func() {
 		sumAvailableReplicasStatusCollectorName := "sum-available-replicas"
 		selectAvailableStatusCollectorName := "select-available-replicas"
 		selectReplicasStatusCollectorName := "replicas"
+		listNginxWecsStatusCollectorName := "nginx-wecs"
 
 		clusterSelector := []metav1.LabelSelector{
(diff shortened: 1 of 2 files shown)


test/e2e/ginkgo/multiple_cluster_deployment_test.go
```diff
@@ -514,6 +515,36 @@ var _ = ginkgo.Describe("end to end testing", func() {
 			gomega.Expect(*cs.Results[0].Rows[0].Columns[0].Number).To(gomega.Equal("2"))
 		})
 
+		ginkgo.It("list of WECs with available nginx deployment", func(ctx context.Context) {
+
+			availableNginxCEL := ksapi.Expression("obj.spec.replicas == returned.status.availableReplicas && obj.metadata.labels['app.kubernetes.io/name'] == 'nginx'")
```
MikeSpreitzer, line 520: Why is the `&& obj.metadata.labels['app.kubernetes.io/name'] == 'nginx'` in here? The `obj.metadata.labels` is from the WDS and does not vary from WEC to WEC.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
```diff
@@ -514,6 +515,36 @@ var _ = ginkgo.Describe("end to end testing", func() {
 			gomega.Expect(*cs.Results[0].Rows[0].Columns[0].Number).To(gomega.Equal("2"))
 		})
 
+		ginkgo.It("list of WECs with available nginx deployment", func(ctx context.Context) {
```
MikeSpreitzer, line 518: I think that it would be better to be a little more specific in the description of the test here, perhaps something like "lists all the WECs where the reported number of availableReplicas equals the desired number of replicas".

test/e2e/ginkgo/multiple_cluster_deployment_test.go
andreyod: The intention was to filter out all the objects in the WDS without this label(the test name says: `available nginx`) , and to exercise `obj.metadata` along with `obj.spec`

test/e2e/ginkgo/multiple_cluster_deployment_test.go
andreyod: If we keep the label check from the comment above than we need to say `nginx deployment` in the test name.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
```diff
@@ -514,6 +515,36 @@ var _ = ginkgo.Describe("end to end testing", func() {
...
+
+			// Validate CombinedStatus results
+			gomega.Expect(len(cs.Results)).To(gomega.Equal(1))
+			gomega.Expect(cs.Results[0].Name).To(gomega.Equal(listNginxWecsStatusCollectorName))
+
+			gomega.ExpectWithOffset(1, len(cs.Results[0].ColumnNames)).To(gomega.Equal(1))
+			gomega.Expect(cs.Results[0].ColumnNames[0]).To(gomega.Equal("wecName"))
+
+			gomega.ExpectWithOffset(1, len(cs.Results[0].Rows)).To(gomega.Equal(2))
+		})
```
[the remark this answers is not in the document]
andreyod: Added the values check

test/e2e/ginkgo/multiple_cluster_deployment_test.go
MikeSpreitzer: The filtering wrt workload object is done prior. Remember that a CombinedStatus is already specific to a pair of workload object and StatusCollector.

If you want to test access to `obj.metadata` then I recommend doing it in a way that makes a difference to the resulting CombinedStatus.

MikeSpreitzer: Put another way, it is the job of the BindingPolicy and the Binding to associate StatusCollectors with workload objects. The filter in a StatusCollectorSpec looking at data that is specific to one workload object but (potentially) varying from one WEC to another.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
MikeSpreitzer: Yes. But remember that `ginkgo.It` takes not a name but an English sentence (with the subject elided) asserting the behavior that is to be tested.

andreyod: Changed.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
andreyod: ok. I removed this. Can't see a way to test an access to `obj.metadata`  in a way that makes a difference to the resulting CombinedStatus.

MikeSpreitzer: We can pursue that additional testing in follow-on PRs.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
```diff
@@ -562,6 +563,46 @@ var _ = ginkgo.Describe("end to end testing", func() {
...
+			cs := util.GetCombinedStatus(ctx, ksWds, wds, ns, workloadName, bpName)
+
+			// Validate CombinedStatus results
+			gomega.Expect(len(cs.Results)).To(gomega.Equal(1))
+			gomega.Expect(cs.Results[0].Name).To(gomega.Equal(listNginxWecsStatusCollectorName))
+
+			gomega.ExpectWithOffset(1, len(cs.Results[0].ColumnNames)).To(gomega.Equal(1))
+			gomega.Expect(cs.Results[0].ColumnNames[0]).To(gomega.Equal("wecName"))
+
+			gomega.ExpectWithOffset(1, len(cs.Results[0].Rows)).To(gomega.Equal(2))
```
MikeSpreitzer, line 593: The offset should be zero here, as in the other Expect calls in this func.
Can be pursued in follow-on.

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #2344, opened 2024-07-22
✨ Support create-only in generic transport controller

MikeSpreitzer:
## Summary
This PR updates the generic transport controller to respect the create-only bit.

## Related issue(s)

This implements part of #2336

### DIFF
--- a/pkg/transport/generic_transport_controller.go
+++ b/pkg/transport/generic_transport_controller.go
@@ -718,9 +718,9 @@ func (c *genericTransportController) updateWrappedObjectsAndFinalizer(ctx contex
 	return nil
 }
 
-func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, binding *v1alpha1.Binding) ([]*unstructured.Unstructured, sets.Set[metav1.GroupResource], error) {
+func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, binding *v1alpha1.Binding) ([]Wrapee, sets.Set[metav1.GroupResource], error) {
 	groupResources := sets.New[metav1.GroupResource]()
-	objectsToPropagate := make([]*unstructured.Unstructured, 0)
+	objectsToPropagate := make([]Wrapee, 0)
 	// add cluster-scoped objects to the 'objectsToPropagate' slice
 	for _, clusterScopedObject := range binding.Spec.Workload.ClusterScope {
 		gvr := schema.GroupVersionResource(clusterScopedObject.GroupVersionResource)
@@ -730,7 +730,7 @@ func (c *genericTransportController) getObjectsFromWDS(ctx context.Context, bind
 		}
 		gr := metav1.GroupResource{Group: clusterScopedObject.GroupVersionResource.Group, Resource: clusterScopedObject.GroupVersionResource.Resource}
 		groupResources.Insert(gr)
-		objectsToPropagate = append(objectsToPropagate, TransformObject(ctx, c.customTransformCollection, gr, object, binding.Name))
+		objectsToPropagate = append(objectsToPropagate, Wrapee{TransformObject(ctx, c.customTransformCollection, gr, object, binding.Name), clusterScopedObject.CreateOnly})
 	}
 	// add namespace-scoped objects to the 'objectsToPropagate' slice
 	for _, namespaceScopedObject := range binding.Spec.Workload.NamespaceScope {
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
/cc @nirrozenbaum 
/cc @effi-ofer 
/cc @vMaroon 
/cc @pdettori 
@cwiklik

pkg/transport/generic_transport_controller.go
```diff
@@ -718,54 +718,56 @@ func (c *genericTransportController) updateWrappedObjectsAndFinalizer(ctx contex
...
+		wrapees = append(wrapees, Wrapee{TransformObject(ctx, c.customTransformCollection, gr, object, binding.Name), clause.CreateOnly})
 	}
 
-	return objectsToPropagate, groupResources, nil
+	return wrapees, groupResources, nil
 }
 
 // computeDestToWrappedObjects returns the following three things.
-//   - the destToWrappedObject function. This maps a destination to the customized wrapped object
+//   - the destToWrappedObject function. This maps a destination to the slice of wrapped customized objects
```
vMaroon, line 754: s/customized/possibly customized/.

MikeSpreitzer: I am thinking "certainly as customized as the user is calling for". But that's a little too long to spell out in the name.

nirrozenbaum (commented):
@MikeSpreitzer there is a general point I don't understand with this PR and previous ones (previous PRs were merged before I wrote my comments).

why do we have two transport interfaces? it's either we say that one of the transport requirements is to have a create-only option or not.
in addition, as a side comment, is there such a word "Wrapee" in English?
 
it looks as if instead of having a generalized transport interface we keep moving towards OCM over time. IMO, it's ok to say KS transport will always be OCM, but in that case there are a lot of abstractions that are not needed in our codebase.

MikeSpreitzer:
@nirrozenbaum : The reason for adding a second transport interface instead of just changing the one is because of the separation into two repos and the way that we test. I first tried just changing the one interface, but the testing failed. The existence of two interfaces is a temporary thing during a few transitional releases. You should find there is a comment now outlining a transition strategy.

We need the create-only option because of the way apiservers handle Job objects. It is not because we are using OCM.

ezrasilvera:
@MikeSpreitzer If I understand correctly the test in this PR only validate that the field was added to the object, right ?
If so - do we want to add an actual e2e test that will validate the functionality - e.g., try to change the object on  the WEC with kubectl and see that the  value stays there (we can add comparison to another object without createOnly and see that the value is overwritten by KS/OCM)

MikeSpreitzer:
The E2E test in _this_ PR only checks that the create-only bit is set correctly in the Binding object, according to what the BindingPolicy says. In _this_ PR the test can not check behavior because ks/OTP does not yet implement the behavior. See the transition plan outlined in the trasnport.go file.

pkg/transport/generic_transport_controller.go
```diff
@@ -827,43 +831,38 @@ func (c *genericTransportController) computeDestToCustomizedObjects(objectsToPro
...
+	return destToCustomizedWrapees, bindingErrors
 }
 
-func (c *genericTransportController) wrapBatch(batchToPropagate []*unstructured.Unstructured, binding *v1alpha1.Binding, numShard int, isSharded bool) (*unstructured.Unstructured, error) {
+func (c *genericTransportController) wrapBatch(batchToPropagate []Wrapee, binding *v1alpha1.Binding, numShard int, isSharded bool) (*unstructured.Unstructured, error) {
 	var wrapped runtime.Object
 	if t2, is := c.transport.(TransportWithCreateOnly); is {
-		wrapees := abstract.SliceMap(batchToPropagate, FIXME_ADD_CONSTANT_CREATEONLY)
-		wrapped = t2.WrapObjectsHavingCreateOnly(wrapees)
+		wrapped = t2.WrapObjectsHavingCreateOnly(batchToPropagate)
```
vMaroon, line 862: Something is not clear to me in the logic. A binding may now contain objects with `CreateOnly==true` and objects with false. Does this not mean that bindings may now lead to the generation of 2 (or more) wrapped-objects? That is putting sharding aside.

Apart from this point, everything looks good.

vMaroon: I could not see an explicit answer in the referenced doc or API documentation clause. I can only see the desired behavior from setting `CreateOnly` bit which I agree with, therefore I assume that the point will be clearer in upcoming PRs.

MikeSpreitzer: Then I am not sure I understand your question. Is it answered by saying that each BindingPolicy object has exactly one corresponding Binding object, in which everything said about a given workload object is combined into one ClusterScopeDownsyncClause or NamespaceScopeDownsyncClause? You already knew that and implemented the StatusCollector aspect of that.

vMaroon: I was under the impression that to handle the requirement of create-only ManifestWorks we'd need to split selected objects (of a binding) into two groups: one that are wrapped in a possibly sharded manifestwork as we've done to this point, and another that are wrapped in a possibly sharded create-only manifestwork. I do not see the implementation of this separation in this PR, or the groundwork for it.

If my assumption is wrong, please let me know.

MikeSpreitzer: In a ManifestWork each workload object can have its own UpdateStrategy (in the ManifestConfigs of the ManifestWorkSpec).

MikeSpreitzer: But, more to the point, this is a detail in the transport plugin. The interface to the transport plugin passes the needed information, a create-only bit per workload object.

vMaroon: I see, thanks.

PR #2346, opened 2024-07-22
✨ Make the binding controller respect the create-only bit

MikeSpreitzer:
## Summary
This PR updates the binding controller to respect the create-only bit.

## Related issue(s)

This is part of #2336

### DIFF
--- a/pkg/binding/bindingpolicy-resolution.go
+++ b/pkg/binding/bindingpolicy-resolution.go
@@ -49,7 +49,7 @@ type bindingPolicyResolution struct {
 	requiresSingletonReportedState bool
 }
 
-// objectData stores the UID (why?), resource version, create-only bit,
+// objectData stores the UID, resource version, create-only bit,
 // and statuscollectors for an object.
 type objectData struct {
 	UID              string
--- a/pkg/binding/bindingpolicy-resolution.go
+++ b/pkg/binding/bindingpolicy-resolution.go
@@ -31,7 +31,7 @@ import (
 )
 
 // bindingPolicyResolution stores the selected object identifiers and
-// destinations for a single bindingpolicy. The mutex should be read locked
+// destinations and modulations for a single bindingpolicy. The mutex should be read locked
 // before reading, and write locked before writing to any field.
 type bindingPolicyResolution struct {
 	sync.RWMutex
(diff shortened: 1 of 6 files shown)


pkg/binding/bindingpolicy-resolution.go
```diff
@@ -49,23 +49,25 @@ type bindingPolicyResolution struct {
 	requiresSingletonReportedState bool
 }
 
-// objectData stores the resource version and statuscollectors for an object.
+// objectData stores the UID (why?), resource version, create-only bit,
```
vMaroon, line 52: The UID is cached for the CombinedStatusResolver's use.

pkg/binding/bindingpolicy.go
```diff
@@ -279,10 +279,11 @@ type mrObject interface {
 // testObject tests if the object matches the given tests.
 // The returned tuple is:
 //   - bool: whether the object matches ANY of the tests
+//   - bool: whether any of the tests says CreateOnly==true
 //   - sets.Set[string]: the UNION of the statuscollector names that appear within
 //     EACH of the tests that the object matches
 func (c *Controller) testObject(ctx context.Context, objIdentifier util.ObjectIdentifier, objLabels map[string]string,
```
vMaroon, line 285: Perhaps now is a good time to move this function.

pkg/binding/bindingpolicy.go
```diff
@@ -279,10 +279,11 @@ type mrObject interface {
 // testObject tests if the object matches the given tests.
 // The returned tuple is:
 //   - bool: whether the object matches ANY of the tests
+//   - bool: whether any of the tests says CreateOnly==true
```
vMaroon, line 282: s/any/any of the matching tests/.

pkg/binding/bindingpolicy.go
```diff
@@ -279,10 +279,11 @@ type mrObject interface {
...
 //   - sets.Set[string]: the UNION of the statuscollector names that appear within
 //     EACH of the tests that the object matches
```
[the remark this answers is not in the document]
MikeSpreitzer: This one already has "that the object matches".

pkg/binding/bindingpolicy-resolution.go
MikeSpreitzer: Thanks

pkg/binding/bindingpolicy.go
MikeSpreitzer: Let's do that in a separate PR.

pkg/binding/bindingpolicy.go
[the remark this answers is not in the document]
vMaroon: Yes. I meant to unify the syntax by making it:

```
//   - sets.Set[string]: the UNION of the statuscollector names that appear within
//     EACH of the matching tests
```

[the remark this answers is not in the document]
MikeSpreitzer: On second thought, I think "matching test" is a bit ambiguous because the purpose of a test is matching, so the "matching" in that phrase _could_ be read as a description of simply the test.

I have pushed a revision that unifies the wording, going with wording that is explicit about the test matching the object.

PR #2347, opened 2024-07-22
✨ Improve testability by allowing controllers to run independently

waltforme:
## Summary
This PR introduces a `controllers` flag into the controller manager, so we can specify which controller to be started by the controller manager. This feature improves the testability because it allows us to simulate the scenario where some of the controller(s) crush and restart.

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -75,7 +75,7 @@ func main() {
 	pflag.StringVar(&itsName, "its-name", "", "name of the Inventory and Transport Space to connect to (empty string means to use the only one)")
 	pflag.StringVar(&wdsName, "wds-name", "", "name of the workload description space to connect to")
 	pflag.StringVar(&allowedGroupsString, "api-groups", "", "list of allowed api groups, comma separated. Empty string means all API groups are allowed")
-	pflag.StringSliceVar(&controllers, "controllers", []string{}, "list of controllers to be started by the controller manager, comma separated, e.g. 'binding,status'")
+	pflag.StringSliceVar(&controllers, "controllers", []string{}, "list of controllers to be started by the controller manager, lower case and comma separated, e.g. 'binding,status'. If not specified (or emtpy list specifed), all controllers are started. Currently available controllers are 'binding' and 'status'.")
 	pflag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
@@ -98,11 +98,17 @@ func main() {
 
 	// parse allowed resources string
 	allowedGroupsSet := util.ParseAPIGroupsString(allowedGroupsString)
-	// parse controllers
-	ctlrsToStart, err := parseControllers(controllers)
-	if err != nil {
-		setupLog.Error(err, "unable to parse controllers")
+
+	// check controllers flag
+	ctlrsToStart := sets.New(controllers...)
+	if !sets.New(
+		strings.ToLower(binding.ControllerName),
+		strings.ToLower(status.ControllerName),
+	).IsSuperset(ctlrsToStart) {
+		setupLog.Error(fmt.Errorf("unkown controller specified"), "'controllers' flag has incorrect value")
+		os.Exit(1)
 	}
+
 	// setup manager
 	// manager here is mainly used for leader election and health checks
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
@@ -177,7 +183,7 @@ func main() {
 
 	cListers := make(chan interface{}, 1)
 
-	if shouldStartController(strings.ToLower(binding.ControllerName), ctlrsToStart) {
+	if len(ctlrsToStart) == 0 || ctlrsToStart.Has(strings.ToLower(binding.ControllerName)) {
 		setupLog.Info("Starting controller", "name", binding.ControllerName)
 		if err := bindingController.Start(ctx, workers, cListers); err != nil {
 			setupLog.Error(err, "error starting the binding controller")
@@ -186,7 +192,8 @@ func main() {
 	}
 
 	// check if status add-on present before starting the status controller
-	if util.CheckWorkStatusPresence(itsRestConfig) && shouldStartController(strings.ToLower(status.ControllerName), ctlrsToStart) {
+	if util.CheckWorkStatusPresence(itsRestConfig) &&
+		(len(ctlrsToStart) == 0 || ctlrsToStart.Has(strings.ToLower(status.ControllerName))) {
 		setupLog.Info("Starting controller", "name", status.ControllerName)
 		statusController, err := status.NewController(wdsRestConfig, itsRestConfig, wdsName,
 			bindingController.GetBindingPolicyResolutionBroker())
@@ -207,27 +214,4 @@ func main() {
 		os.Exit(1)
 	}
 	select {}
-}
-
-func parseControllers(controllers []string) (sets.Set[string], error) {
-	if len(controllers) == 0 {
-		return nil, nil
-	}
-
-	for _, name := range controllers {
-		if strings.ToLower(name) != name {
-			return nil, fmt.Errorf("upper case letters not allowed for controller name %q", name)
-		}
-	}
-
-	return sets.New(controllers...), nil
-}
-
-// shouldStartController checks if the given controller should be started,
-// empty controllers is equivalent to start all.
-func shouldStartController(controller string, controllers sets.Set[string]) bool {
-	if len(controllers) == 0 {
-		return true
-	}
-	return controllers.Has(controller)
 }
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -63,12 +63,14 @@ func main() {
 	var itsName string
 	var wdsName string
 	var allowedGroupsString string
+	var controllers string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The [host]:port from which /metrics is served.")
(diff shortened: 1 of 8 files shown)


cmd/controller-manager/main.go
```diff
@@ -90,6 +93,8 @@ func main() {
 
 	// parse allowed resources string
 	allowedGroupsSet := util.ParseAPIGroupsString(allowedGroupsString)
+	// parse controllers string
+	ctlrsToStart := util.ParseControllersString(controllers)
```
MikeSpreitzer, line 97: I think that ParseControllersString belongs in this package rather than util, because ParseControllersString is not of general interest.

pkg/util/resources.go
```diff
@@ -40,6 +40,21 @@ func ParseAPIGroupsString(apiGroups string) sets.Set[string] {
 	return groupsSet
 }
 
+func ParseControllersString(controllers string) sets.Set[string] {
+	if controllers == "*" {
+		return nil
+	}
+
+	return sets.New(allToLowerCase(strings.Split(controllers, ","))...)
```
MikeSpreitzer, line 48: I think that allowing variation in case is more complexity than we need here.

pkg/util/resources.go
```diff
@@ -40,6 +40,21 @@ func ParseAPIGroupsString(apiGroups string) sets.Set[string] {
 	return groupsSet
 }
 
+func ParseControllersString(controllers string) sets.Set[string] {
+	if controllers == "*" {
```
MikeSpreitzer, line 44: `controllers == ""` will produce an equivalent result (i.e., treated the same by ShouldStartController), right? If so then I do not see the need to support two ways to say the same thing.

cmd/controller-manager/main.go
```diff
@@ -63,12 +64,14 @@ func main() {
...
 	var wdsName string
 	var allowedGroupsString string
+	var controllers string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The [host]:port from which /metrics is served.")
 	flag.StringVar(&pprofAddr, "pprof-bind-address", ":8082", "The [host]:port fron which /debug/pprof is served.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&itsName, "its-name", "", "name of the Inventory and Transport Space to connect to (empty string means to use the only one)")
 	flag.StringVar(&wdsName, "wds-name", "", "name of the workload description space to connect to")
 	flag.StringVar(&allowedGroupsString, "api-groups", "", "list of allowed api groups, comma separated. Empty string means all API groups are allowed")
+	flag.StringVar(&controllers, "controllers", "*", "list of controllers to be started by the controller manager, comma separated, e.g. 'binding,status'")
```
MikeSpreitzer, line 74: FYI, pflag has StringArray

cmd/controller-manager/main.go
```diff
@@ -165,13 +170,17 @@ func main() {
 
 	cListers := make(chan interface{}, 1)
 
-	if err := bindingController.Start(ctx, workers, cListers); err != nil {
-		setupLog.Error(err, "error starting the binding controller")
-		os.Exit(1)
+	if util.ShouldStartController(strings.ToLower(binding.ControllerName), ctlrsToStart) {
+		setupLog.Info("Starting controller", "name", binding.ControllerName)
```
MikeSpreitzer, line 174: Since this line is for exactly one, known, controller, it could be simpler: `Info("Starting binding controller")`.

pkg/util/resources.go
waltforme: Yes, removed `*` related code.

cmd/controller-manager/main.go
waltforme: Used `pflag` instead.

pkg/util/resources.go
waltforme: Just want to clarify: Is the suggestion to (a) remove the case check/conversion, or (b) reject inputs with upper cases?

cmd/controller-manager/main.go
waltforme: Just wandering, is it common seen that there are *_test.go files in package main? Asking because I wrote a unit test for this func.

MikeSpreitzer: I don't know about common, but it is certainly acceptable.

pkg/util/resources.go
MikeSpreitzer: The suggestion is the just use one case consistently. All the provision and consumption of these strings is in the same repo. We have no user community with an established practice of using both cases. Just pick one and go with it.

pkg/util/resources.go
MikeSpreitzer: Sorry, I should have also pointed out that I am not a fan of irregularity.  When absence of value X from set S means one thing, except it means the opposite when set S is empty, that is irregular. I am not going to insist on this, but it is FYI.

cmd/controller-manager/main.go
waltforme: Moved code from util to main.

pkg/util/resources.go
waltforme: OK, rejected inputs with upper case letters.

cmd/controller-manager/main.go
```diff
@@ -63,34 +68,41 @@ func main() {
...
-	flag.StringVar(&allowedGroupsString, "api-groups", "", "list of allowed api groups, comma separated. Empty string means all API groups are allowed")
-	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
+	var controllers []string
+	pflag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The [host]:port from which /metrics is served.")
+	pflag.StringVar(&pprofAddr, "pprof-bind-address", ":8082", "The [host]:port fron which /debug/pprof is served.")
+	pflag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	pflag.StringVar(&itsName, "its-name", "", "name of the Inventory and Transport Space to connect to (empty string means to use the only one)")
+	pflag.StringVar(&wdsName, "wds-name", "", "name of the workload description space to connect to")
+	pflag.StringVar(&allowedGroupsString, "api-groups", "", "list of allowed api groups, comma separated. Empty string means all API groups are allowed")
+	pflag.StringSliceVar(&controllers, "controllers", []string{}, "list of controllers to be started by the controller manager, comma separated, e.g. 'binding,status'")
```
MikeSpreitzer, line 78: Since the meaning of the empty array is special, I think it should be mentioned in the doc string.

cmd/controller-manager/main.go
```diff
@@ -192,3 +208,26 @@ func main() {
...
+}
+
+// shouldStartController checks if the given controller should be started,
+// empty controllers is equivalent to start all.
+func shouldStartController(controller string, controllers sets.Set[string]) bool {
+	if len(controllers) == 0 {
+		return true
+	}
+	return controllers.Has(controller)
+}
```
MikeSpreitzer, line 233: FYI, I think we do not need to write functions here. Just use `sets.New`, `sets.IsSubsset`, and `sets.Has` directly in the `main` func.

cmd/controller-manager/main.go
```diff
@@ -192,3 +208,26 @@ func main() {
...
+	}
+
+	for _, name := range controllers {
+		if strings.ToLower(name) != name {
+			return nil, fmt.Errorf("upper case letters not allowed for controller name %q", name)
+		}
+	}
+
+	return sets.New(controllers...), nil
+}
```
MikeSpreitzer, line 224: FYI, I do not think we need a whole func definition. Just use `sets.New` and `sets.IsSubset` directly in the `main` func.

pkg/util/resources.go
MikeSpreitzer: I am trying to suggest something even simpler. There is a valid set of controller names. Those are literal exact strings, end of story. Any invalid controller name is either simply ineffective or is reported as an error. The user is responsible for passing valid controller names.

pkg/util/resources.go
MikeSpreitzer: Right. So I am OK with having "*" mean "every". My real complaint is having empty set mean full set.

cmd/controller-manager/main.go
waltforme: Killed `parseControllers` (and `shouldStartController` as well).

cmd/controller-manager/main.go
waltforme: Improved the doc string as suggested.

cmd/controller-manager/main.go
```diff
@@ -63,34 +68,47 @@ func main() {
...
 
 	// parse allowed resources string
 	allowedGroupsSet := util.ParseAPIGroupsString(allowedGroupsString)
 
+	// check controllers flag
+	ctlrsToStart := sets.New(controllers...)
+	if !sets.New(
+		strings.ToLower(binding.ControllerName),
+		strings.ToLower(status.ControllerName),
+	).IsSuperset(ctlrsToStart) {
```
MikeSpreitzer, line 107: (1) You could use sets.Difference to discover the invalid values, if any, and then include them in the error message.

(2) Why bother changing the case? Why not let the string values in `binding.ControllerName` and `status.ControllerName` be the very same string values used in the `controllers` array?

Neither of these is critical.

pkg/util/resources.go
MikeSpreitzer: Not a blocker.

MikeSpreitzer (approved):
This is good enough to merge.
There remain some suggestions that could be acted on in follow-on work.
/lgtm
/approve
Thaniks!

PR #2348, opened 2024-07-23
:seedling: 🌱 other/misc Added create-kubestellar-demo-env script

benzha14:
## Related issue(s)

Fixes #


benzha14:
@clubanderson should I add the boiler plate text, delete this pull request and try again?

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,149 @@
+#!/usr/bin/env bash
+
+# The formal instructions accompaning this script can be found at:
+# https://docs.kubestellar.io/release-0.23.1/direct/get-started/
```
MikeSpreitzer, line 4: I think it is best to use a relative reference, rather than introduce another place that requires editing for every release.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,149 @@
+#!/usr/bin/env bash
+
+# The formal instructions accompaning this script can be found at:
```
MikeSpreitzer, line 3: I am not sure I understand the relationship this text is trying to describe. Those instructions do not reference this script.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,149 @@
...
+# https://docs.kubestellar.io/release-0.23.1/direct/get-started/
+
+set -e
+
+echo -e "Checking that pre-req softwares are installed..."
+curl -s -o check_pre_req.sh https://raw.githubusercontent.com/kubestellar/kubestellar/v0.23.1/hack/check_pre_req.sh
+chmod +x check_pre_req.sh
+./check_pre_req.sh -V kflex ocm helm kubectl docker kind
```
MikeSpreitzer, line 11: Can/should this curl-to-bash instead of create a local copy?

Creating a local copy creates dependency on the current working directory being a place where this script can and should be written.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,149 @@
...
+
+output=$(./check_pre_req.sh -V kflex ocm helm kubectl docker kind)
+
+if echo "$output" | grep -q "X"; then
```
MikeSpreitzer, line 15: Can the exit code of that command be used instead? If not, then perhaps this is an issue to raise against that script.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,149 @@
...
+
+cluster_clean_up "kind delete cluster --name kubeflex"
+cluster_clean_up "kind delete cluster --name cluster1"
+cluster_clean_up "kind delete cluster --name cluster2"
+echo -e "Cluster space clean up has been completed"
+
+echo -e "\nStarting context clean up..."
+
+context_clean_up() {
+    output=$(kubectx)
```
MikeSpreitzer, line 41: What is `kubectx` ?

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,149 @@
...
+done
+
+kubectl config delete-context its1 || true
+kflex ctx its1
+kubectl config delete-context wds1 || true
+kflex ctx wds1
+kflex ctx
+
+echo -e "\nCreating cluster and context for cluster 1 and 2..."
+: set flags to "" if you have installed KubeStellar on an OpenShift cluster
```
MikeSpreitzer, line 99: That's not actually something that an invoker of this script can do.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,149 @@
...
+   clusteradm --context its1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk '{print $0 " --context '${cluster}' --singleton '${flags}'"}' | sh
+done
+
+echo -e "Checking that the CSR for cluster 1 and 2 appears..."
+checking_cluster1_and_cluster2() {
+    output=$(kubectl --context its1 get csr)
+    c1=false
+    c2=false
+    
+    while [ "$c1" = false ] || [ "$c2" = false ]; do
```
MikeSpreitzer, line 114: No need to check twice. This condition is also checked below, where it could be followed by a `break`.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,149 @@
...
+flags="--force-internal-endpoint-lookup"
+clusters=(cluster1 cluster2);
+for cluster in "${clusters[@]}"; do
+   kind create cluster --name ${cluster}
+   kubectl config rename-context kind-${cluster} ${cluster}
+   clusteradm --context its1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk '{print $0 " --context '${cluster}' --singleton '${flags}'"}' | sh
+done
+
+echo -e "Checking that the CSR for cluster 1 and 2 appears..."
+checking_cluster1_and_cluster2() {
```
MikeSpreitzer, line 109: FYI, I think that this could be simpler if instead of coding the OR in here, it just looked for one and gets invoked twice.

scripts/create-kubestellar-demo-env.sh
benzha14: I going to remove it to reduce confusion.

scripts/create-kubestellar-demo-env.sh
benzha14: I going to remove it to reduce confusion.

scripts/create-kubestellar-demo-env.sh
benzha14: I will work on this.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,149 @@
...
+echo "Context space clean up completed"
+
+echo -e "\nStarting the process to install kubestellar core: kind-kubeflex..."
+export KUBESTELLAR_VERSION=0.23.1
+
+curl -s -o create-kind-cluster-with-SSL-passthrough.sh https://raw.githubusercontent.com/kubestellar/kubestellar/v${KUBESTELLAR_VERSION}/scripts/create-kind-cluster-with-SSL-passthrough.sh
+
+chmod +x create-kind-cluster-with-SSL-passthrough.sh
+
+./create-kind-cluster-with-SSL-passthrough.sh --name kubeflex --port 9443
```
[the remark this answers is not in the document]
benzha14: I will work on this.

scripts/create-kubestellar-demo-env.sh
benzha14: kubectx is a tool to switch between contexts (clusters) on kubectl faster.

MikeSpreitzer:
```shell
mspreitz@mjs13 kubestellar % which kubectx
kubectx not found
```

`kubectx` is not among the listed prerequites for using KubeStellar.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: Perhaps this script should have a command-line flag that controls this behavior. Or auto-detect whether the cluster is OpenShift.

scripts/create-kubestellar-demo-env.sh
benzha14: fixed

scripts/create-kubestellar-demo-env.sh
benzha14: i changed it to use a kubectl command instead

scripts/create-kubestellar-demo-env.sh
benzha14: Im will comment/delete out this line at the moment because I am not sure how openshift clusters work.

MikeSpreitzer: I think that leaving a comment will be more helpful than just deleting the line.

scripts/create-kubestellar-demo-env.sh
benzha14: fixed

scripts/create-kubestellar-demo-env.sh
benzha14: fixed

scripts/create-kubestellar-demo-env.sh
benzha14: added

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: I see no change. I still see the output being examined with `grep`.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,161 @@
...
+# Deploys Kubestellar Environment
+
+set -e
+
+echo -e "Checking that pre-req softwares are installed..."
+curl -s -o check_pre_req.sh https://raw.githubusercontent.com/kubestellar/kubestellar/v0.23.1/hack/check_pre_req.sh
+chmod +x check_pre_req.sh
+./check_pre_req.sh -V kflex ocm helm kubectl docker kind
+
+output=$(./check_pre_req.sh -V kflex ocm helm kubectl docker kind)
```
MikeSpreitzer, line 25: I now see two identical invocations of the check_pre_req.sh script, one with the output going to stdout and one with the output going to a shell variable. FYI, you could just do the latter and echo the contents of the shell variable.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: Did you push your change? I still see `kubectx` being used.

benzha14:
Git getting a messy. Im going to open a new PR

PR #2353, opened 2024-07-24
📖 Start documenting the create-only bit

MikeSpreitzer:
## Summary
This PR adds user-facing documentation of the create-only feature.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-create-only/direct/binding/

## Related issue(s)

This is part of #2336


MikeSpreitzer:
/cc @nirrozenbaum 
/cc @pdettori 
/cc @KPRoche 
@cwiklik

docs/content/direct/binding.md
```diff
@@ -1,5 +1,179 @@
 # Binding workload with WEC
 
-This document is about BindingPolicy and Binding objects and the concept of downsync.
+This document is about associating WECs with workload objects. The
+primary concept is sometimes called "downsync", which confusingly
```
ezrasilvera, line 4: So why not just avoiding downsyncing in our docs and avoid 1) the confusion 2) this explanation about the confusion ?

docs/content/direct/binding.md
```diff
@@ -1,5 +1,179 @@
...
+The user controls downsync primarily through API objects of kinds
+`BindingPolicy` and `Binding`. These go in a WDS and associate
+workload objects in that WDS with WECs, along with adding some
+modulations on how downsync is done.
+
+`BindingPolicy` is a higher level concept than `Binding`. KubeStellar
+has a controller that translates each `BindingPolicy` to a
+`Binding`. A user _could_ eschew the `BindingPolicy` and directly
+maintain a `Binding` object or let a different controller maintain the
+`Binding` object (TODO: check that this is true). The `Binding` object
```
ezrasilvera, line 20: Unless we recommend doing that, or see a value in it, I don't see why we should add the " A user _could_ eschew the `BindingPolicy` and directly maintain a `Binding` object or let a different controller maintain the `Binding` object (TODO: check that this is true)."

docs/content/direct/binding.md
```diff
@@ -1,5 +1,179 @@
...
+The workload object selection part of a `DownsyncPolicyClause`
+consists of the following fields, all of which are optional. A
+workload object matches a `DownsyncPolicyClause` if that object
+matches all of these selection fields; the default value for each
+field specifies a test that every object passes. Note, however, that
+not every API object in a WDS is considered to be a _workload_
+object. The KubeStellar control objects and `CombinedStatus` objects
+are not considered to be workload objects. (TODO: also talk about the
+other objects that never match).
+
```
ezrasilvera, line 55: Instead of all the info in the sections bellow this point - I don't see why we can't just point to the object itself, and add all the needed info as comments there?
And maybe maintain just the yaml example.   Seems like a duplication of the code

docs/content/direct/binding.md
```diff
@@ -1,5 +1,179 @@
...
+A `DownsyncPolicyClause` also contains the following two fields to
+modulate the downsync.
+
+- `statusCollectors`: an array of references to `StatusCollector`
+  objects that specify how status from the WECs is to be combined and
+  reported in `CombinedStatus` objects in the WDS. See [Combining
+  returned status](combined-status.md) for more details.
+
+- `createOnly`: a boolean that, when true, suppresses on-going
+  maintenance of the object in the WEC. When this boolean is `true`,
```
ezrasilvera, line 113: suppresses on-going  maintenance of the object in the WEC by KubeStellar   
(you can still maintain it through the WEC)

docs/content/direct/binding.md
MikeSpreitzer: The term "downsync" is already all over the project. If you want to try eradicating it then I think that should be a separate PR.

docs/content/direct/binding.md
MikeSpreitzer: This was among the motivations given in the discussion about introducing the Binding object. It also is a bit of alignment with other projects, which is also a bit of value. So I definitely think it is worth mentioning.

docs/content/direct/binding.md
```diff
@@ -1,5 +1,179 @@
...
+
+## BindingPolicy
+
+The `spec` of a `BindingPolicy` has two predicates that (1) identify a
+subset of the WECs in the inventory of the ITS associated with the WDS
+and (2) identify a subset of the workload objects in the WDS. The
+primary function of the `BindingPolicy` is to assert the desired
+association between (1) and (2). A `BindingPolicy` can also add some
+modulations on how those workload objects are downsynced to/from those
+WECs.
```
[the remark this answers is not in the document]
MikeSpreitzer: Because here we have to say what a BindingPolicy does.

docs/content/direct/binding.md
MikeSpreitzer: I have pushed such a revision.

docs/content/direct/binding.md
MikeSpreitzer: Moot now.

docs/content/direct/binding.md
ezrasilvera: If  you believe it's valuable I'm OK with it.

PR #2354, opened 2024-07-24
 :seedling: 🌱 other/misc Added create-kubestellar-demo-env script + Update script based on suggestions

benzha14:
## Related issue(s)

Fixes #

### DIFF
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -1,0 +1,151 @@
+#!/usr/bin/env bash
+# Copyright 2024 The KubeStellar Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Deploys Kubestellar environment for demo purposes.
+
+set -e
+
+echo -e "Checking that pre-req softwares are installed..."
+curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v0.23.1/hack/check_pre_req.sh | bash -s -- -V kflex ocm helm kubectl docker kind
+
+output=$(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v0.23.1/hack/check_pre_req.sh | bash -s -- -A -V kflex ocm helm kubectl docker kind
+)
+
+echo -e "\nStarting environment clean up..."
+echo -e "Starting cluster clean up..."
+
+cluster_clean_up() {
+    error_message=$(eval "$1" 2>&1)
+    if [ $? -ne 0 ]; then
+        echo "clean up failed. Error:"
+        echo "$error_message"
+    fi
+}
+
+cluster_clean_up "kind delete cluster --name kubeflex"
+cluster_clean_up "kind delete cluster --name cluster1"
+cluster_clean_up "kind delete cluster --name cluster2"
+echo -e "Cluster space clean up has been completed"
+
+echo -e "\nStarting context clean up..."
+
+context_clean_up() {
+    output=$(kubectl config get-contexts -o name)
+
+    while IFS= read -r line; do
+        if [ "$line" == "kind-kubeflex" ]; then 
+            echo "Deleting kind-kubeflex context..."
+            kubectl config delete-context kind-kubeflex
+
+        elif [ "$line" == "cluster1" ]; then
+            echo "Deleting cluster1 context..."
+            kubectl config delete-context cluster1
+
+        elif [ "$line" == "cluster2" ]; then
+            echo "Deleting cluster2 context..."
+            kubectl config delete-context cluster2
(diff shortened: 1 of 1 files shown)


benzha14:
@MikeSpreitzer I needed to make a new PR because the previous one became too messy. Sorry for the inconvenience cause by my end. This script is the updated version of the previous one.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,151 @@
...
+
+# Deploys Kubestellar environment for demo purposes.
+
+set -e
+
+echo -e "Checking that pre-req softwares are installed..."
+curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v0.23.1/hack/check_pre_req.sh | bash -s -- -V kflex ocm helm kubectl docker kind
+
+output=$(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v0.23.1/hack/check_pre_req.sh | bash -s -- -A -V kflex ocm helm kubectl docker kind
+)
```
MikeSpreitzer, line 24: TODO: consider curl-to-bash here.

Also: how is the user going to invoke _this_ script (create-kubestellar-demo-env.sh)? How will this script invoke the corresponding _version_ of check_pre_req.sh?

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,151 @@
...
+cluster_clean_up "kind delete cluster --name cluster1"
+cluster_clean_up "kind delete cluster --name cluster2"
+echo -e "Cluster space clean up has been completed"
+
+echo -e "\nStarting context clean up..."
+
+context_clean_up() {
+    output=$(kubectl config get-contexts -o name)
+
+    while IFS= read -r line; do
```
MikeSpreitzer, line 47: FYI: tather than indirect through a shell variable, this could be simpler: `kubectl config get-contexts -o name | while ... done`.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,151 @@
...
+        fi
+
+    done <<< "$output"
+}
+
+context_clean_up
+echo "Context space clean up completed"
+
+echo -e "\nStarting the process to install KubeStellar core: kind-kubeflex..."
+export KUBESTELLAR_VERSION=0.23.1
```
MikeSpreitzer, line 77: cf question about invoking this script and it invoking corresponding versions of other scripts.

scripts/create-kubestellar-demo-env.sh
benzha14: The script can be invoke through this command once its in the pr goes through: bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v0.23.1/scripts/create-kubebestellar-demo-env.sh).

Regarding the version of check_pre_req.sh, I don't see any indications of version in the check_pre_req.sh script.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,151 @@
...
+done
+
+echo -e "Checking that the CSR for cluster 1 and 2 appears..."
+
+checking_cluster() {
+    found=false
+    
+    while true; do
+
+        output=$(kubectl --context its1 get csr)
```
MikeSpreitzer, line 118: FYI: if the contents of `output` are used in only one place then you could cut out the indirection through the variable and instead pipe the output of `kubectl` to the input of the `while` statement.

scripts/create-kubestellar-demo-env.sh
benzha14: I'll try that out

scripts/create-kubestellar-demo-env.sh
```diff
@@ -0,0 +1,151 @@
...
+echo""
+echo "Approving CSR for cluster1 and cluster2..."
+checking_cluster cluster1
+checking_cluster cluster2
+
+echo""
+echo "Checking the new clusters are in the OCM inventory and label them"
+kubectl --context its1 get managedclusters
+kubectl --context its1 label managedcluster cluster1 location-group=edge name=cluster1
+kubectl --context its1 label managedcluster cluster2 location-group=edge name=cluster2
```
MikeSpreitzer, line 151: FYI: GitHub complains less if the last line ends with a line terminator.

scripts/create-kubestellar-demo-env.sh
benzha14: Good catch, Im thinking about making the version number an argument to the command use to invoke it

scripts/create-kubestellar-demo-env.sh
benzha14: I'll try that out

benzha14:
Let have the code go through the PR as it is. I want to properly test download/invoke issue once its in the code base. I can also improve the script in future PRs.

clubanderson:
I am ok with that

/lgtm

clubanderson:
lets get something out there and then we can iterate on it with subsequent testing.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: My point is that scripts should reference each other consistently. If a user invokes version X of this script, it should in turn invoke version X of check_pre_requ.sh.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: I think that this PR raises the degree to which we would benefit from introducing the concept of asking a user to fetch and install (or at least use) a particular release.

PR #2356, opened 2024-07-25
✨ Test for status controller eventual consistency

waltforme:
## Summary
This PR introduce test on whether status controller achieves eventual consistency w.r.t. the singleton statuses of workload objects in WDS.

The test simulates a scenario where the status controller is absent (e.g. because of a crush) for some time during which there are changes in bindingpolicies. Upon the reappearance of the status controller, it should drive the singleton statuses into their desired state.

## Related
#2285 depends on this PR.
This PR uses #2347

### DIFF
--- a/test/e2e/ginkgo/multiple_cluster_deployment_test.go
+++ b/test/e2e/ginkgo/multiple_cluster_deployment_test.go
@@ -376,17 +376,17 @@ var _ = ginkgo.Describe("end to end testing", func() {
 						ObjectSelectors: []metav1.LabelSelector{{MatchLabels: map[string]string{"app.kubernetes.io/name": "nginx-singleton"}}},
 					}},
 				},
+				func(bp *ksapi.BindingPolicy) error {
+					bp.Spec.WantSingletonReportedState = true
+					return nil
+				},
 			)
-			patch := []byte(`{"spec":{"wantSingletonReportedState": true}}`)
-			_, err := ksWds.ControlV1alpha1().BindingPolicies().Patch(
-				ctx, "nginx-singleton", types.MergePatchType, patch, metav1.PatchOptions{})
-			gomega.Expect(err).NotTo(gomega.HaveOccurred())
 			util.ValidateNumDeployments(ctx, wec1, ns, 1)
 			util.ValidateNumDeployments(ctx, wec2, ns, 0)
 			util.ValidateSingletonStatus(ctx, wds, ns, "nginx-singleton")
-			patchAgain := []byte(`{"spec":{"clusterSelectors":[{"matchLabels":{"name":"CelestialNexus"}}]}}`)
-			_, err = ksWds.ControlV1alpha1().BindingPolicies().Patch(
-				ctx, "nginx-singleton", types.MergePatchType, patchAgain, metav1.PatchOptions{})
+			patch := []byte(`{"spec":{"clusterSelectors":[{"matchLabels":{"name":"CelestialNexus"}}]}}`)
+			_, err := ksWds.ControlV1alpha1().BindingPolicies().Patch(
+				ctx, "nginx-singleton", types.MergePatchType, patch, metav1.PatchOptions{})
 			gomega.Expect(err).NotTo(gomega.HaveOccurred())
 			util.ValidateNumDeployments(ctx, wec1, ns, 0)
 			util.ValidateNumDeployments(ctx, wec2, ns, 0)
@@ -410,11 +410,11 @@ var _ = ginkgo.Describe("end to end testing", func() {
 						ObjectSelectors: []metav1.LabelSelector{{MatchLabels: map[string]string{"app.kubernetes.io/name": "nginx-singleton"}}},
 					}},
 				},
+				func(bp *ksapi.BindingPolicy) error {
+					bp.Spec.WantSingletonReportedState = true
+					return nil
+				},
 			)
-			BPPatch := []byte(`{"spec":{"wantSingletonReportedState": true}}`)
-			_, err := ksWds.ControlV1alpha1().BindingPolicies().Patch(
-				ctx, "nginx-singleton", types.MergePatchType, BPPatch, metav1.PatchOptions{})
-			gomega.Expect(err).NotTo(gomega.HaveOccurred())
 			util.ValidateNumDeployments(ctx, wec1, ns, 1)
 			util.ValidateNumDeployments(ctx, wec2, ns, 0)
 			util.ValidateSingletonStatus(ctx, wds, ns, "nginx-singleton")
@@ -436,7 +436,7 @@ var _ = ginkgo.Describe("end to end testing", func() {
 			util.ExpectDepolymentAvailability(ctx, coreCluster, "wds1-system", "kubestellar-controller-manager")
 
 			// At this time, the status controller should not be running, this is a simulation of a crush.
-			BPPatch = []byte(`{"spec":{"clusterSelectors":[{"matchLabels":{"name":"CelestialNexus"}}]}}`)
+			BPPatch := []byte(`{"spec":{"clusterSelectors":[{"matchLabels":{"name":"CelestialNexus"}}]}}`)
 			_, err = ksWds.ControlV1alpha1().BindingPolicies().Patch(
 				ctx, "nginx-singleton", types.MergePatchType, BPPatch, metav1.PatchOptions{})
 			gomega.Expect(err).NotTo(gomega.HaveOccurred())
--- a/test/e2e/ginkgo/multiple_cluster_deployment_test.go
+++ b/test/e2e/ginkgo/multiple_cluster_deployment_test.go
@@ -18,6 +18,8 @@ package e2e
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"time"
 
 	"github.com/onsi/ginkgo/v2"
@@ -358,7 +360,7 @@ var _ = ginkgo.Describe("end to end testing", func() {
 		})
 	})
 
-	ginkgo.Context("singleton status testing", func() {
+	ginkgo.Context("singleton status creation and deletion", func() {
 		ginkgo.It("sets (or deletes) singleton status when a singleton bindingpolicy/deployment is created (or deleted)", func(ctx context.Context) {
 			util.DeleteDeployment(ctx, wds, ns, "nginx") // we don't have to delete nginx
 			util.CreateDeployment(ctx, wds, ns, "nginx-singleton",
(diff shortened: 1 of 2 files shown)


test/e2e/ginkgo/multiple_cluster_deployment_test.go
```diff
@@ -382,16 +382,78 @@ var _ = ginkgo.Describe("end to end testing", func() {
...
+				[]ksapi.DownsyncPolicyClause{
+					{DownsyncObjectTest: ksapi.DownsyncObjectTest{
+						ObjectSelectors: []metav1.LabelSelector{{MatchLabels: map[string]string{"app.kubernetes.io/name": "nginx-singleton"}}},
+					}},
+				},
+			)
+			BPPatch := []byte(`{"spec":{"wantSingletonReportedState": true}}`)
+			_, err := ksWds.ControlV1alpha1().BindingPolicies().Patch(
+				ctx, "nginx-singleton", types.MergePatchType, BPPatch, metav1.PatchOptions{})
+			gomega.ExpectWithOffset(1, err).NotTo(gomega.HaveOccurred())
```
MikeSpreitzer, line 415: Calls to `Expect` in this position (the body of the function given to `It`) should have an offset of 0.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
```diff
@@ -382,16 +382,78 @@ var _ = ginkgo.Describe("end to end testing", func() {
...
+					{MatchLabels: map[string]string{"name": "cluster1"}},
+				},
+				[]ksapi.DownsyncPolicyClause{
+					{DownsyncObjectTest: ksapi.DownsyncObjectTest{
+						ObjectSelectors: []metav1.LabelSelector{{MatchLabels: map[string]string{"app.kubernetes.io/name": "nginx-singleton"}}},
+					}},
+				},
+			)
+			BPPatch := []byte(`{"spec":{"wantSingletonReportedState": true}}`)
+			_, err := ksWds.ControlV1alpha1().BindingPolicies().Patch(
```
MikeSpreitzer, line 413: FYI, It would be OK with me if the BindingPolicy were created with the Spec that this test needs.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
```diff
@@ -382,16 +382,78 @@ var _ = ginkgo.Describe("end to end testing", func() {
...
+				ctx, "nginx-singleton", types.MergePatchType, BPPatch, metav1.PatchOptions{})
+			gomega.ExpectWithOffset(1, err).NotTo(gomega.HaveOccurred())
+			util.ValidateNumDeployments(ctx, wec1, ns, 1)
+			util.ValidateNumDeployments(ctx, wec2, ns, 0)
+			util.ValidateSingletonStatus(ctx, wds, ns, "nginx-singleton")
+			ginkgo.By("Singleton status synced")
```
MikeSpreitzer, line 419: FYI, this looks to me more like a log statement than a "By" clause. `ginkgo.GinkgoWriter` and `ginkgo.GinkgoLogr` are available.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
```diff
@@ -382,16 +382,78 @@ var _ = ginkgo.Describe("end to end testing", func() {
...
+
+			// Restart the controller manager without starting the status controller.
+			util.ScaleDeployment(ctx, coreCluster, "wds1-system", "kubestellar-controller-manager", 0)
+			CMPatch := []byte(`{"spec":{"template":{"spec":{"containers":[{"args":["--health-probe-bind-address=:8081","--leader-elect","--wds-name=wds1","--its-name=","--api-groups=","-v=2","--controllers=binding"],"name":"manager"}]}}}}`)
```
MikeSpreitzer, line 423: It might be a tad safer to read the current args and compute the correct new args.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
```diff
@@ -382,16 +382,78 @@ var _ = ginkgo.Describe("end to end testing", func() {
...
+			_, err := ksWds.ControlV1alpha1().BindingPolicies().Patch(
+				ctx, "nginx-singleton", types.MergePatchType, BPPatch, metav1.PatchOptions{})
+			gomega.ExpectWithOffset(1, err).NotTo(gomega.HaveOccurred())
+			util.ValidateNumDeployments(ctx, wec1, ns, 1)
+			util.ValidateNumDeployments(ctx, wec2, ns, 0)
+			util.ValidateSingletonStatus(ctx, wds, ns, "nginx-singleton")
+			ginkgo.By("Singleton status synced")
+
+			// Restart the controller manager without starting the status controller.
+			util.ScaleDeployment(ctx, coreCluster, "wds1-system", "kubestellar-controller-manager", 0)
```
MikeSpreitzer, line 422: Why scale down and up? Isn't the patch operation enough?

test/util/util.go
```diff
@@ -598,3 +606,51 @@ func Expect1PodOfEach(ctx context.Context, client *kubernetes.Clientset, ns stri
...
+
+func ScaleDeployment(ctx context.Context, client *kubernetes.Clientset, ns string, name string, target int32) {
+	ginkgo.By(fmt.Sprintf("Scaling Deployment %q in namespace %q to %d", name, ns, target))
+	gomega.EventuallyWithOffset(1, func() error {
+		gotSc, err := client.AppsV1().Deployments(ns).GetScale(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		gotSc.Spec.Replicas = target
+		_, err = client.AppsV1().Deployments(ns).UpdateScale(ctx, name, gotSc, metav1.UpdateOptions{})
```
MikeSpreitzer, line 618: Why not do this just once?

test/util/util.go
```diff
@@ -598,3 +606,51 @@ func Expect1PodOfEach(ctx context.Context, client *kubernetes.Clientset, ns stri
...
+		labelSelector := metav1.FormatLabelSelector(gotDeploy.Spec.Selector)
+		podList, err := client.CoreV1().Pods(ns).List(ctx, metav1.ListOptions{
+			LabelSelector: labelSelector,
+		})
+		if err != nil {
+			return err
+		}
+		if len(podList.Items) != int(target) {
+			return fmt.Errorf("pod count not converged yet")
+		}
```
MikeSpreitzer, line 635: This test seems to be assuming that the Deployment controller has reacted, creating a new ReplicaSet, and the ReplicaSet controller has reacted, creating more Pods. These things are not guaranteed. I think a correct test would check that the Deployment's `Status.ObservedGeneration` equals its `ObjectMeta.Generation` and check that `Status.UpdatedReplicas` equals `Spec.Replicas`.

test/util/util.go
```diff
@@ -598,3 +606,51 @@ func Expect1PodOfEach(ctx context.Context, client *kubernetes.Clientset, ns stri
...
+		if gotDeploy.Status.ReadyReplicas != target {
```
MikeSpreitzer, line 636: Why `ReadyReplicas` instead of `AvailableReplicas` ?

MikeSpreitzer:
What is the point of the following remark in the opening comment?

> This PR uses https://github.com/kubestellar/kubestellar/pull/2347

That PR has merged.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
waltforme: I see that func `CreateBindingPolicy` currently doesn't have an argument to pass the needed value of `wantSingletonReportedState`, so `Patch`es are used whenever that value needs to be set.

Are you suggesting some change to `CreateBindingPolicy`?

MikeSpreitzer: There are a number of possible ways to accomplish that.

You could generalize CreateBindingPolicy by adding a parameter specifically for that. Or by adding a variadic parameter (so that the existing calls do not have to be edited) taking any sequence of functions that tweak the constructed BindingPolicy. Or recognize that the Get check in CreateBindingPolicy is pointless (the success of the Create implies that Get will succeed) and just construct the BindingPolicy you want and call Create directly and expect no error.

waltforme: Adopted the variadic parameter approach.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
waltforme: Used `ginkgo.GinkgoWriter` instead of `By`.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
waltforme: Implemented the suggestion.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
waltforme: If we use patch, then there will be an interval during which the terminating and old pods and the starting and new pods coexist. Sometimes the terminating pods last long enough, such that the old pods are still functioning during the simulated absence --- which makes the test flaky.

So I did scale down and up to ensure the status controller is indeed not functioning during the simulated absence.

test/util/util.go
waltforme: Fixed.

test/util/util.go
waltforme: I tried the suggestion here, but I see the coexistence of the terminating old pods and the starting new pods, similar to the situation described here: https://github.com/kubestellar/kubestellar/pull/2356#discussion_r1697294754 . So I didn't change this.

test/util/util.go
waltforme: Used `AvailableReplicas` instead.

waltforme:
> What is the point of the following remark in the opening comment?
> 
> > This PR uses #2347
> 
> That PR has merged.

In case someone revisits #2347, he/she can see a link to this PR for better understanding of how testability is improved.

test/util/util.go
MikeSpreitzer: For comparison, see the stalled transition state exhibited at https://github.com/kubestellar/kubestellar/issues/2365#issuecomment-2258982513 et seq.

I think the comparison of `gotDeploy.Status.AvailableReplicas` with `target` is a sufficient test.
The additional test here is not harmful, just unnecessary. Can be removed in follow-on.

MikeSpreitzer (approved):
This is good enough to merge.
/lgtm
/approve
Thanks!

test/e2e/ginkgo/multiple_cluster_deployment_test.go
MikeSpreitzer: I just realized that I previously read that wrong. I thought the concern was that startup of the partially-disabled controller manager while the fully enabled one was also running would make the test flaky. I see now that the concern was that the test would proceed to probe behavior while both controller managers were running. I think that proper waiting in the transition code will rule out that scenario.

PR #2357, opened 2024-07-26
✨  Add monitoring tools (prometheus, pyroscope & grafana) for KS

dumb0002:
## Summary

This PR install monitoring tools for KS, more precisely:
   1)   Remove the obsolete installation and configuration of prometheus and grafana via Ansible playbooks
   2)  Add new scripts to Install prometheus, pyroscope and grafana monitoring tools in the host cluster
   3)  Add scripts to configure prometheus and pyroscope to scrape the KS ctl manager, KS transport ctl and WDS kube apiserver in a WDS space
   4) Configuration of grafana with the prometheus and pyroscope datasource and configuration of prometheus to retain its data in a persistent volume. 

An updated README file will be added in a new PR

## Related issue(s)

Fixes #


monitoring/configuration/ks-transport-ctl-svc.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: ks-transport-controller
+  name: ks-transport-controller
+spec:
+  ports:
+  - name: http1
```
MikeSpreitzer, line 9: Do you anticipate having a port named "http2"? I do not think that "1" vs "2" conveys any useful information. I suggest more informative names.

monitoring/configuration/prometheus-rbac.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: prometheus
```
MikeSpreitzer, line 4: Do we need to define this? Isn't it part of the monitoring infrastructure that is already present?

monitoring/configuration/sa-rbac-proxy
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: kubestellar-controller-manager
```
MikeSpreitzer, line 4: Mismatch between this name and the file name.

monitoring/configuration/sm-ks-controller-manager.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: k8s-controller-monitor
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    interval: 5s
+    path: /metrics
+    port: https
+    scheme: https
+    tlsConfig:
+      insecureSkipVerify: true
```
MikeSpreitzer, line 13: Why not use the CA cert that is also available?

monitoring/configure-metrics-tools.sh
```diff
@@ -0,0 +1,94 @@
...
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
+
+# set core context
+kubectl config use-context $ctx
+
+: --------------------------------------------------------------------
+: Configure kubestellar controller for prometheus scraping
+: --------------------------------------------------------------------
+: 1. configure the rbac for the kube-rbac-proxy
```
MikeSpreitzer, line 36: I would add "in the KubeStellar controller-manager Pod".

monitoring/configure-metrics-tools.sh
```diff
@@ -0,0 +1,94 @@
...
+
+: 2. Create svc for KS ctl manager
+kubectl -n $wds-system apply -f ${SCRIPT_DIR}/ks-ctl-manager-svc.yaml
+
+: 3. Create service monitor for KS ctl manager
+sed s/%WDS_NS%/$wds-system/g sm-ks-controller-manager.yaml | kubectl -n ks-monitoring apply -f -
+
+
+: --------------------------------------------------------------------
+: Configure kubestellar controller for pyroscope scraping
```
MikeSpreitzer, line 47: There are multiple "controllers" in KubeStellar. I would be more specific here, it is about the KubeStellar _controller-manager_.

monitoring/configure-metrics-tools.sh
```diff
@@ -0,0 +1,94 @@
...
+
+: --------------------------------------------------------------------
+: Configure transport controller for prometheus scraping
+: --------------------------------------------------------------------
+: 1. Create transport controller service:
+kubectl -n $wds-system apply -f ${SCRIPT_DIR}/ks-transport-ctl-svc.yaml
+
+: 2. Open the metrics and pprof ports:
+kubectl -n $wds-system get deploy transport-controller -o yaml | yq '(del(.status) |.spec.template.spec.containers.[0].ports[0].name |= "http1")' | yq '.spec.template.spec.containers.[0].ports[0].protocol |= "TCP"' | yq '.spec.template.spec.containers.[0].ports[0].containerPort |= 8090' | yq '.spec.template.spec.containers.[0].ports[1].name |= "http2"' | yq '.spec.template.spec.containers.[0].ports[1].protocol |= "TCP"' | yq '.spec.template.spec.containers.[0].ports[1].containerPort |= 8092' | kubectl --context $ctx apply --namespace=$wds-system -f -
+
```
MikeSpreitzer, line 65: Why not configure for Pyroscope scraping?

monitoring/configure-metrics-tools.sh
```diff
@@ -0,0 +1,94 @@
...
+: Configure kubestellar controller for pyroscope scraping
+: --------------------------------------------------------------------
+: 1. Open the pprof port:
+kubectl -n $wds-system get deploy kubestellar-controller-manager -o yaml | yq '(del(.status) |.spec.template.spec.containers.[1].ports[0].name |= "http")' | yq '.spec.template.spec.containers.[1].ports[0].protocol |= "TCP"' | yq '.spec.template.spec.containers.[1].ports[0].containerPort |= 8082' | kubectl --context $ctx apply --namespace=$wds-system -f -
+
+: 2. Add labels for pyroscope scraping:
+kubectl -n $wds-system get deploy kubestellar-controller-manager -o yaml | yq '.spec.template.metadata.annotations."profiles.grafana.com/cpu.port" |= "8082" |  .spec.template.metadata.annotations."profiles.grafana.com/cpu.scrape"|= "true" | .spec.template.metadata.annotations."profiles.grafana.com/goroutine.port" |= "8082" | .spec.template.metadata.annotations."profiles.grafana.com/goroutine.scrape" |= "true" |
+.spec.template.metadata.annotations."profiles.grafana.com/memory.port" |= "8082" | .spec.template.metadata.annotations."profiles.grafana.com/memory.scrape" |= "true"' | kubectl -n $wds-system apply -f -
```
MikeSpreitzer, line 54: See #2362

monitoring/configure-metrics-tools.sh
```diff
@@ -0,0 +1,94 @@
...
+
+
+: --------------------------------------------------------------------
+: Configure transport controller for prometheus scraping
+: --------------------------------------------------------------------
+: 1. Create transport controller service:
+kubectl -n $wds-system apply -f ${SCRIPT_DIR}/ks-transport-ctl-svc.yaml
+
+: 2. Open the metrics and pprof ports:
+kubectl -n $wds-system get deploy transport-controller -o yaml | yq '(del(.status) |.spec.template.spec.containers.[0].ports[0].name |= "http1")' | yq '.spec.template.spec.containers.[0].ports[0].protocol |= "TCP"' | yq '.spec.template.spec.containers.[0].ports[0].containerPort |= 8090' | yq '.spec.template.spec.containers.[0].ports[1].name |= "http2"' | yq '.spec.template.spec.containers.[0].ports[1].protocol |= "TCP"' | yq '.spec.template.spec.containers.[0].ports[1].containerPort |= 8092' | kubectl --context $ctx apply --namespace=$wds-system -f -
```
MikeSpreitzer, line 64: See https://github.com/kubestellar/ocm-transport-plugin/issues/35

monitoring/configure-metrics-tools.sh
```diff
@@ -0,0 +1,94 @@
...
+.spec.template.metadata.annotations."profiles.grafana.com/memory.port" |= "8082" | .spec.template.metadata.annotations."profiles.grafana.com/memory.scrape" |= "true"' | kubectl -n $wds-system apply -f -
+
+
+: --------------------------------------------------------------------
+: Configure transport controller for prometheus scraping
+: --------------------------------------------------------------------
+: 1. Create transport controller service:
+kubectl -n $wds-system apply -f ${SCRIPT_DIR}/ks-transport-ctl-svc.yaml
+
+: 2. Open the metrics and pprof ports:
```
MikeSpreitzer, line 63: s/Open/Declare/

monitoring/configure-metrics-tools.sh
```diff
@@ -0,0 +1,94 @@
...
+
+
+: --------------------------------------------------------------------
+: Configure WDS API server for prometheus scraping
+: --------------------------------------------------------------------
+: 1. Create a SA and give the right RBAC to talk to the wds1 API server
+kubectl --context $wds -n ks-monitoring apply -f ${SCRIPT_DIR}/prometheus-rbac.yaml
+
+: 2. Create token secret for prometheus in the target wds space
+kubectl --context $wds get namespace ks-monitoring --ignore-not-found || kubectl --context $wds create ns ks-monitoring
```
MikeSpreitzer, line 77: FYI: This might read a little more naturally if written as `if ! kubectl get ... ; then kubectl create ...; fi`

monitoring/configure-metrics-tools.sh
MikeSpreitzer: Or, could this be indirected through a Service (as in Prometheus scraping)? That is, define a Service and put the needed labels/annotations on the Service?

MikeSpreitzer:
What's up with the big file `monitoring/prometheus/prometheus-operator.yaml`? Is this copied from somewhere? We shouldn't be maintaining our own copy of someone else's source. Since the installation is scripted, isn't there a way to get what we need without maintaining our own copy?

monitoring/configuration/ks-ctl-manager-svc.yaml
```diff
@@ -0,0 +1,12 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: kubestellar-controller
```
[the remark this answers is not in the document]
dumb0002: Fixed

monitoring/configuration/ks-transport-ctl-svc.yaml
dumb0002: Fixed. Add a better name: `http-metrics`

monitoring/configuration/prometheus-rbac.yaml
dumb0002: Fixed - moved this file to the prometheus directory instead.

monitoring/configure-metrics-tools.sh
dumb0002: Fixed!

monitoring/configure-metrics-tools.sh
```diff
@@ -0,0 +1,94 @@
...
+kubectl -n $wds-system apply -f ${SCRIPT_DIR}/ks-ctl-manager-svc.yaml
+
+: 3. Create service monitor for KS ctl manager
+sed s/%WDS_NS%/$wds-system/g sm-ks-controller-manager.yaml | kubectl -n ks-monitoring apply -f -
+
+
+: --------------------------------------------------------------------
+: Configure kubestellar controller for pyroscope scraping
+: --------------------------------------------------------------------
+: 1. Open the pprof port:
```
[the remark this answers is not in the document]
dumb0002: Fixed - improved the text description to clearly state that we are naming ports

monitoring/configure-metrics-tools.sh
dumb0002: Fixed!

monitoring/configure-metrics-tools.sh
dumb0002: We are already configuring the transport controller for  Pyroscope scraping

monitoring/configure-metrics-tools.sh
dumb0002: Pyroscope to scrape pods directly by adding annotations to the pods - not sure if it can scrape it through a service - I'll look into it.

monitoring/configure-metrics-tools.sh
dumb0002: Fixed!

monitoring/configuration/sm-ks-controller-manager.yaml
dumb0002: The available certs are not available in the prometheus namespace. Hence, for configuration simplicity, it's better not to use them.

monitoring/configuration/sa-rbac-proxy
dumb0002: Removed this file - as it's not needed.

monitoring/configuration/ks-transport-ctl-svc.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: ks-transport-controller
+  name: ks-transport-controller
+spec:
+  ports:
+  - name: http-metric
```
MikeSpreitzer, line 9: This name reads like it is saying that this port Is for HTTP with metric system units.
Or it is about measuring a single metric.
I would use "metrics" rather than "metric".
I would drop the "http" or replace it with "prometheus", because HTTP is the standard protocol to use when scraping Prometheus metrics.

monitoring/configuration/sm-ks-controller-manager.yaml
MikeSpreitzer: I am not sure I understand. There is a CA certificate in `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`, right? What goes wrong if that one is used?

dumb0002:
> What's up with the big file `monitoring/prometheus/prometheus-operator.yaml`? Is this copied from somewhere? We shouldn't be maintaining our own copy of someone else's source. Since the installation is scripted, isn't there a way to get what we need without maintaining our own copy?

You're right! We shouldn't be maintaining our own copy indeed. Removed it and install prometheus-operator via a helm-chart

monitoring/configuration/sm-ks-controller-manager.yaml
MikeSpreitzer: I found the reason. A little experimentation shows that the RBAC proxy server's certificate is issued not by the cluster CA but rather by a special CA hacked up just for that pod.

monitoring/configuration/ks-transport-ctl-svc.yaml
dumb0002: Fixed. I dropped the "http" and replaced metric with "metrics"

monitoring/configuration/ks-ctl-manager-svc.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: v1
+metadata:
+  name: ks-controller-manager
+  labels:
+    app: ks-controller-manager
+spec:
+  selector:
+    control-plane: controller-manager
+  ports:
+  - name: https
```
MikeSpreitzer, line 11: Wouldn't "metrics" be a better name for this port?

monitoring/configure-metrics-tools.sh
```diff
@@ -0,0 +1,113 @@
...
+
+set -x # echo so that users can understand what is happening
+set -e # exit on error
+
+ctx="kind-kubeflex"
+wds="wds1"
+
+while [ $# != 0 ]; do
+    case "$1" in
+        (-h|--help) echo "$0 usage: (--host-cluster-context | --space-name )*"
```
MikeSpreitzer, line 25: Those flags take parameter values, that should be mentioned in the usage.

monitoring/install-ks-monitoring.sh
```diff
@@ -0,0 +1,102 @@
...
+set -x # echo so that users can understand what is happening
+set -e # exit on error
+
+ctx="kind-cluster1"
+ns="ks-monitoring"
+
+
+while [ $# != 0 ]; do
+    case "$1" in
+        (-h|--help) echo "$0 usage: (--host-cluster-context | --monitoring-ns )*"
```
MikeSpreitzer, line 26: Those flags take parameter values, that should be shown in the usage.

monitoring/configuration/ks-ctl-manager-svc.yaml
dumb0002: Fixed!

monitoring/configure-metrics-tools.sh
dumb0002: Fixed!

monitoring/install-ks-monitoring.sh
dumb0002: Fixed!

monitoring/configure-metrics-tools.sh
MikeSpreitzer: As I mentioned elsewhere, I have lately remembered that pulling from Pods is better than pulling from Services.

MikeSpreitzer (approved):
This is good enough to merge.
/lgtm
/approve
Thanks!

PR #2368, opened 2024-07-31
🌱 Simplify test of eventual consistency of singleton status

MikeSpreitzer:
## Summary
This PR simplifies the "cleans up previously synced but currently invalid singleton status" test spec, removing unnecessary scale down and up of the controller around its reconfiguration. This should be valid now that the test uses a proper wait for the reported state of the controller-manager Deployment to match its desired state.

## Related issue(s)

Fixes #


MikeSpreitzer:
I do not understand why, but the simplification does not work.
See #2369 instead.

PR #2378, opened 2024-08-02
🌱 Begin drafting example controller outline

MikeSpreitzer:
## Summary
This PR contributes an example outline of a controller, to illustrate a pattern for writing one that reacts to all sorts of schedules of stimulii.

## Related issue(s)

Fixes #


nirrozenbaum:
@MikeSpreitzer not sure I got the idea of this PR. 
why do we need an example of how to write controller?
there are plenty of examples out there. 
for example, the k8s sample controller - https://github.com/kubernetes/sample-controller.

pkg/controller-outline/controller.go
```diff
@@ -0,0 +1,463 @@
...
+)
+
+// This is an outline of a Kubernetes controller.
+// In general, a controller takes its input from the state of objects of N various types I1, I2, ... IN
+// and the controller's main job is to maintain some output in objects of M various types O1, O2, ... OM.
+// The controller may also report into the status section of the I1, I2, ... IN objects.
+
+// This is informed by https://github.com/kubernetes/sample-controller/blob/master/controller.go
+// and https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/controllers.md
+// and goes further by explicitly considering multiple types of input and multiple tyeps of output
```
waltforme, line 43: s/tyeps/types/

waltforme: Also, maybe adding some concrete use cases to support the need of `M >= 2`?

MikeSpreitzer: An example of `M >= 2` is the status controller. It outputs to `CombinedStatus` objects and to the `.status` sections of workload objects (note: neither of these are inputs to the status controller).

MikeSpreitzer:
@nirrozenbaum: I updated the introductory commentary with more of an explanation of what is distinctive about this outline.

nirrozenbaum:
> @nirrozenbaum: I updated the introductory commentary with more of an explanation of what is distinctive about this outline.

I still can't understand why would we want to have in our repo these 600 lines of code we need to maintain.
first, I don't see a reason why ks repo should supply an example of how to write a controller.
second, I don't see anything special/unique in this code, whoever wants to write k8s controller can find a lot of material out there.

of course if others think otherwise you can go ahead and merge, but I do not see the point in this PR/code.

pkg/controller-outline/controller.go
waltforme: I also thought about the status controller. But the handling of `.status` sections of all kinds doesn't fit the pattern. The status controller doesn't watch the kinds, nor put them in a workqueue, nor sync them in `func (ctlr *controller) sync`. IMHO the status controller is a _counter_ example of the pattern proposed in this PR, showing that in practice controllers are often not that 'regular'.

MikeSpreitzer: Or maybe the status controller should be changed to conform to this pattern.

clubanderson:
This design PR has been open as WIP for 17 months. To keep the repository clean while preserving the discussion, this has been converted to a GitHub Discussion.

**Continue the conversation here:** https://github.com/kubestellar/kubestellar/discussions/3621

Thank you @MikeSpreitzer for the design work. The discussion can be revisited when there's renewed interest in this controller pattern documentation.

PR #2381, opened 2024-08-02
✨ Create named container port for `/debug/pprof`

### LINKED ISSUE #2360: feature: named ports on KubeStellar controller-manager Pod
### Feature Description

A user (such as @dumb0002) who wants access to the ports at which `/metrics` and `/debug/pprof` can be reached finds it easier if the Pod object declares named ports for these.

### Proposed Solution

Declare those in the sources of the PodSpec.

### Want to contribute?

### Additional Context

_No response_

francostellari:
## Summary

1. rename the named container port 8443 of the proxy from `https` to `metrics`
2. add `debug-pprof` port 8082 to manager container
3. add flag to set the binding address

## Related issue(s)

Fixes #2360

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -57,7 +57,3 @@ spec:
         - "--its-name={{.Values.ITSName}}"
         - "--api-groups={{.Values.APIGroups}}"
         - -v={{.Values.ControllerManager.Verbosity}}
-        ports:
-        - containerPort: 8082
-          protocol: TCP
-          name: debug-pprof
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -39,7 +39,7 @@ spec:
         ports:
         - containerPort: 8443
           protocol: TCP
-          name: https
+          name: metrics
         resources:
           limits:
             cpu: 500m
@@ -51,8 +51,13 @@ spec:
         args:
         - "--health-probe-bind-address=:8081"
         - "--metrics-bind-address=127.0.0.1:8080"
+        - "--pprof-bind-address=:8082"
         - "--leader-elect"
         - "--wds-name={{.Values.ControlPlaneName}}"
         - "--its-name={{.Values.ITSName}}"
         - "--api-groups={{.Values.APIGroups}}"
         - -v={{.Values.ControllerManager.Verbosity}}
+        ports:
+        - containerPort: 8082
+          protocol: TCP
+          name: debug-pprof
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -54,5 +54,9 @@ spec:
           requests:
             cpu: 10m
             memory: 64Mi
+        ports:
+        - containerPort: 8082
+          protocol: TCP
+          name: debug-pprof
       serviceAccountName: controller-manager
       terminationGracePeriodSeconds: 10
--- a/monitoring/configure-metrics-tools.sh
+++ b/monitoring/configure-metrics-tools.sh
@@ -65,10 +65,7 @@ sed s/%WDS_NS%/$wds-system/g ${SCRIPT_DIR}/configuration/ks-ctl-manager-sm.yaml 
 : --------------------------------------------------------------------
 : Configure kubestellar controller manager pod for pyroscope scraping
 : --------------------------------------------------------------------
-: 1. Adding declarations of the pprof port, so that kubestellar service definition can refer to it by name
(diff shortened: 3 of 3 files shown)


config/default/manager_auth_proxy_patch.yaml
```diff
@@ -51,8 +51,13 @@ spec:
...
+        - "--pprof-bind-address=:8082"
         - "--leader-elect"
         - "--wds-name={{.Values.ControlPlaneName}}"
         - "--its-name={{.Values.ITSName}}"
         - "--api-groups={{.Values.APIGroups}}"
         - -v={{.Values.ControllerManager.Verbosity}}
+        ports:
+        - containerPort: 8082
+          protocol: TCP
+          name: debug-pprof
```
MikeSpreitzer, line 63: I think that it is best if we leave most of the configuration in the base YAML file rather than in this delta. The base file is in `config/manager/manager.yaml`.

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -51,8 +51,13 @@ spec:
         args:
         - "--health-probe-bind-address=:8081"
         - "--metrics-bind-address=127.0.0.1:8080"
+        - "--pprof-bind-address=:8082"
```
MikeSpreitzer, line 54: I do not think that it is necessary to explicitly provide a setting that is the same as the default. Also, not particularly harmful.

MikeSpreitzer:
I think that it would be good for this PR to make the corresponding changes in `monitoring/`, so that this PR does not leave things in an inconsistent state. I am particularly concerned about https://github.com/kubestellar/kubestellar/blob/1adcf8c120ac23c54c9990af5d5a4ef27363d9b6/monitoring/configure-metrics-tools.sh#L69, which will not play nice together with this change. This PR makes that hack unnecessary.

Also: @dumb0002, I am confused by the comment at https://github.com/kubestellar/kubestellar/blob/1adcf8c120ac23c54c9990af5d5a4ef27363d9b6/monitoring/configure-metrics-tools.sh#L68 . If I understand correctly, by "kubestellar service definition" you mean https://github.com/kubestellar/kubestellar/blob/1adcf8c120ac23c54c9990af5d5a4ef27363d9b6/monitoring/configuration/ks-ctl-manager-svc.yaml . But that thing _does not_ refer to the pprof port and does not refer to the metrics port by name. You can see the definitions of the fields of a Port declaration in https://github.com/kubernetes/api/blob/v0.28.12/core/v1/types.go#L4982 et seq.

config/default/manager_auth_proxy_patch.yaml
francostellari: To be honest, I could not figure out why you kept talking about port 8082 when it was nowhere to be seen. Only after your run the help command showing the flag and port it was clear to me.
I think clearly stating the port bring more clarity.
Otherwise, one in the future may wander why open a container port 8082 that does not seem used.

config/default/manager_auth_proxy_patch.yaml
francostellari: The separation of stuff in the files is maddening to me.
I put it in this file because the other containerport was in this file and the flag that open the port is in this file.
I'm ok to move it if you prefer

francostellari:
> I think that it would be good for this PR to make the corresponding changes in `monitoring/`, so that this PR does not leave things in an inconsistent state. I am particularly concerned about
> 
> https://github.com/kubestellar/kubestellar/blob/1adcf8c120ac23c54c9990af5d5a4ef27363d9b6/monitoring/configure-metrics-tools.sh#L69
> 
> , which will not play nice together with this change. This PR makes that hack unnecessary.

I removed the command

config/default/manager_auth_proxy_patch.yaml
MikeSpreitzer: I agree. The use of kustomize here brings us only pain.

MikeSpreitzer: However, I think that the code generation tool that we use produces output that requires modification.

For now, let's just minimize what goes in the delta. I think that this port declaration would go better in the base YAML file.

francostellari: ok, done

MikeSpreitzer:
I tested the current version of this PR (commit 86748451db2b9b50832c7a9e8e5557b289c2ada5).

I used `test/e2e/run-test.sh` to create a running system. Then I used `kubectl --context kind-kubeflex get -n wds1-system deploy kubestellar-controller-manager -o yaml` to inspect the KubeStellar controller-manager, and found the expected port declarations. Then I tested them by running `kubectl --context kind-kubeflex port-forward -n wds1-system deploy/kubestellar-controller-manager 1701:metrics 1702:debug-pprof` in one shell and using another to `kubectl --context kind-kubeflex get --raw https://localhost:1701/metrics > /tmp/metrics.txt` (file contents looked good) and `curl http://localhost:1702/debug/pprof/` (output looked good).

MikeSpreitzer (commented):
This looks good to me, but I really want to get @dumb0002's review too.
/approve
Thanks!

dumb0002:
@MikeSpreitzer (cc: @francostellari ), it looks good to me. I configured pyroscope to scrape by referring to port number only. I will clean up `configure-metrics-tools.sh` script in another PR.

PR #2384, opened 2024-08-05
🌱 add ocm transport to main repo

nirrozenbaum:
## Summary

this PR adds ocm code into ks main repo and uses the local ocm code in the e2e tests. in follow up PRs helm charts will be updated.

## Related issue(s)

Fixes #


Makefile
```diff
@@ -15,9 +15,10 @@
 # Image repo/tag to use all building/pushing image targets
 DOCKER_REGISTRY ?= ghcr.io/kubestellar/kubestellar
 CMD_NAME ?= controller-manager
+TRANSPORT_CMD_NAME ?= ocm-transport-controller
 IMAGE_TAG ?= $(shell git rev-parse --short HEAD)
-IMAGE_REPO ?= ${DOCKER_REGISTRY}/${CMD_NAME}
-IMG ?= ${IMAGE_REPO}:${IMAGE_TAG}
+IMAGE ?= ${DOCKER_REGISTRY}/${CMD_NAME}:${IMAGE_TAG}
```
MikeSpreitzer, line 20: Now that this Makefile is building two images, let's use distinctive names for each. If there is TRANSPORT_IMAGE, maybe also CTLMGR_IMAGE? Same for makefile target names.

scripts/deploy-transport-controller.sh
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-DEFAULT_TRANSPORT_CONTROLLER_IMAGE=ghcr.io/kubestellar/ocm-transport-plugin/transport-controller:0.1.11
+DEFAULT_TRANSPORT_CONTROLLER_IMAGE=ghcr.io/kubestellar/kubestellar/ocm-transport-controller:0.1.11
```
MikeSpreitzer, line 16: I would expect images in kubestellar/kubestellar to be tagged with ks/ks release numbers.

What is the purpose of this script? The answer should be documented. I am wondering if this script should now be using a released image or building an image from the local source.

MikeSpreitzer: Currently, this variable's setting does not matter. This script is always invoked with a third positional argument, so the default image does not matter. We currently have no documentation telling users that that they can call this script with only positional two arguments.

Perhaps we used to have such user instructions. Perhaps we will again in the future.

In the meantime, I think the simplest change in the definition of this variable is no change.

The repo merge campaign will include merging the release procedures and documents. Not necessarily in this PR.

nirrozenbaum: @MikeSpreitzer you're right. atm this variable is never used.  
maybe the best thing to do would be just to remove it. 
in case it will be needed in the future we can always re-add it.

Makefile
nirrozenbaum: @MikeSpreitzer yes, I agree.
in this PR I tried to avoid it cause there are scripts that uses make commands with the existing names. 
let's do in a followup PR.

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #2391, opened 2024-08-12
🌱 addressed previous PR comments

nirrozenbaum:
## Summary

addressed minor comments written by @MikeSpreitzer in #2384.

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/direct/packaging.md
+++ b/docs/content/direct/packaging.md
@@ -227,7 +227,7 @@ By our development practices and not doing any manual hacking we maintain the as
 
 The [release process](release.md) builds and publishes that container image.
 
-`make ko-build-local` will make a local image for just the local
+`make ko-build-controller-manager-local` will make a local image for just the local
 platform. This is used in local testing.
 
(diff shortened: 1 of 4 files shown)


docs/content/direct/packaging.md
```diff
@@ -227,7 +227,7 @@ By our development practices and not doing any manual hacking we maintain the as
 
 The [release process](release.md) builds and publishes that container image.
```
MikeSpreitzer, line 228: This file should also be updated to talk about the newly introduced image. Does not have to be done in this PR.

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

docs/content/direct/packaging.md
nirrozenbaum: @MikeSpreitzer right. atm we have no new image. the newly introduced image will be done in a followup PR that adds helm chart. this file will be updated accordingly in the same PR.

PR #2394, opened 2024-08-14
✨ Make controller-manager use plain Kubernetes

MikeSpreitzer:
## Summary
This PR switches the KubeStellar controller-manager from using controller-runtime to using plain Kubernetes.

There remains some relatively isolated use of just part of controller-runtime, the non-typed client, due to https://github.com/kubestellar/kubeflex/issues/277 .

## Related issue(s)

Fixes #

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -20,7 +20,6 @@ package main
 // to ensure that exec-entrypoint and run can make use of them.
 
 import (
-	"context"
 	"flag"
 	"fmt"
 	"os"
@@ -33,15 +32,17 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/component-base/metrics/legacyregistry"
+	_ "k8s.io/component-base/metrics/prometheus/clientgo"
+	_ "k8s.io/component-base/metrics/prometheus/version"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/healthz"
-	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics/server"
-	crwebhook "sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	v1alpha1 "github.com/kubestellar/kubestellar/api/control/v1alpha1"
 	clientopts "github.com/kubestellar/kubestellar/options"
 	"github.com/kubestellar/kubestellar/pkg/binding"
+	ksctlr "github.com/kubestellar/kubestellar/pkg/controller"
+	ksmetrics "github.com/kubestellar/kubestellar/pkg/metrics"
 	"github.com/kubestellar/kubestellar/pkg/status"
 	"github.com/kubestellar/kubestellar/pkg/util"
 )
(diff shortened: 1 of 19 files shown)


cmd/controller-manager/main.go
```diff
@@ -82,15 +84,16 @@ func main() {
...
+	processOpts.AddToFlags(pflag.CommandLine)
 	itsClientLimits.AddFlags(pflag.CommandLine)
 	wdsClientLimits.AddFlags(pflag.CommandLine)
 	klog.InitFlags(nil)
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.Parse()
-	ctx := context.Background()
+	ctx, _ := ksctlr.InitialContext()
 	logger := klog.FromContext(ctx)
 	ctrl.SetLogger(logger)
```
waltforme, line 95: Are we going to get rid of this?

options/process-options.go
```diff
@@ -0,0 +1,35 @@
...
+limitations under the License.
+*/
+
+package clientsopts
+
+type ProcessOptions[FS FlagSet] struct {
+	MetricsBindAddr     string
+	PProfBindAddr       string
+	HealthProbeBindAddr string
+}
```
waltforme, line 23: I searched throughout the PR, looks like *github.com/spf13/pflag.FlagSet is the only type argument that intantiate `ProcessOptions[FS]`. So the generic type is for some future compatibility?

waltforme (commented):
Maybe the package `pkg/metrics` was born for metrics, but now I see many more general stuff here in the package. As a result, it is a little strange that when creating a controller, many clients are types that imported from a package named 'ks_metrics_'.

cmd/controller-manager/main.go
MikeSpreitzer: Yes, when I am sure it is completely irrelevant.
This PR does not remove _all_ use of controller-runtime, that is blocked on https://github.com/kubestellar/kubeflex/issues/277

MikeSpreitzer:
> Maybe the package pkg/metrics was born for metrics, but now I see many more general stuff here in the package.

What do you see in that package that is not for metrics?

options/process-options.go
MikeSpreitzer: No, it is from the past. In the latest rev of this PR, I have removed the paramaterization by FlagSet type, from all the places where it was.

pkg/status/controller.go
```diff
@@ -58,10 +58,14 @@ const (
...
-	wdsDynClient dynamic.Interface
-	wdsKsClient  ksclient.Interface
-	itsDynClient dynamic.Interface
+	wdsName               string
+	wdsDynClient          dynamic.Interface
+	wdsKsClient           ksclient.Interface
+	bindingPolicyClient   ksmetrics.ClientModNamespace[*v1alpha1.BindingPolicy, *v1alpha1.BindingPolicyList]
+	bindingClient         ksmetrics.ClientModNamespace[*v1alpha1.Binding, *v1alpha1.BindingList]
+	statusCollectorClient ksmetrics.ClientModNamespace[*v1alpha1.StatusCollector, *v1alpha1.StatusCollectorList]
+	combinedStatusClient  ksmetrics.BasicNamespacedClient[*v1alpha1.CombinedStatus, *v1alpha1.CombinedStatusList]
```
waltforme, line 67: Here is an example of what I wanted to discuss at https://github.com/kubestellar/kubestellar/pull/2394#pullrequestreview-2258007020. These four clients are all of some type imported from `ksmetrics` package. In contrast, the `wdsKsClient` before these four is imported from `ksclient` package, which sounds more natural/intuitive.

Not a blocker though.

waltforme:
> > Maybe the package pkg/metrics was born for metrics, but now I see many more general stuff here in the package.
> 
> What do you see in that package that is not for metrics?

I made some comments [here](https://github.com/kubestellar/kubestellar/pull/2394/files#r1729547566).

PR #2396, opened 2024-08-15
🌱  Add GetSingletonReportedStateRequestForObject method

MikeSpreitzer:
## Summary
This PR extends the BindingPolicyResolver interface with a method to inquire about what all the resolutions together say about singleton reported state return. This is good because it will make it easier to implement that feature.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/binding/bindingpolicy-resolver.go
+++ b/pkg/binding/bindingpolicy-resolver.go
@@ -320,6 +320,8 @@ func (resolver *bindingPolicyResolver) GetSingletonReportedStateRequestForObject
 	resolver.RWMutex.RLock()
 	defer resolver.RWMutex.RUnlock()
 	var requested bool
+	// First, just compute whether singleton reported state return is requested for this object.
+	// Avoid thrashing the heap with that set collection unless it is really necessary.
 	for _, resolution := range resolver.bindingPolicyToResolution {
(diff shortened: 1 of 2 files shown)


pkg/binding/bindingpolicy-resolver.go
```diff
@@ -303,6 +316,33 @@ func (resolver *bindingPolicyResolver) ResolutionRequiresSingletonReportedState(
...
+			requested = true
+			break
+		}
+	}
+	if !requested {
+		return false, 0
+	}
+	requested = false
+	matchingWECs := sets.New[string]()
+	for _, resolution := range resolver.bindingPolicyToResolution {
```
vMaroon, line 335: Why not unify the loops?

MikeSpreitzer: The first loop only computes, does not allocate anything on the heap. For an object that does not request singleton reported state return, the early out saves a lot of heap thrashing. Heap thrashing is expensive, both in terms of "inline" costs and the deferred costs of garbage collection.

vMaroon: Fair, perhaps it can be commented in order to quickly be noticed if read in the future?

MikeSpreitzer: Done

PR #2397, opened 2024-08-15
:sparkles: CombinedStatus PropagationMeta Augmentation

vMaroon:
## Summary
Added propagation-meta to combined-status API along with all required BindingPolicy/Binding API changes, and implemented its support.

#### Specifications

BindingPolicy:
```yaml
kind: BindingPolicy
metadata:
  name: nginx-bindingpolicy
spec:
  clusterSelectors:
  - matchLabels: {"location-group":"edge"}
  downsync:
  - statusCollection:
      stalenessThresholdSecs: 100 # default is 120
      statusCollectors:
        - deployments-aggregator
        - deployments-stale
    objectSelectors:
    - matchLabels: {"app.kubernetes.io/name":"nginx"}
```

StatusCollectors:
```yaml
kind: StatusCollector
metadata:
  name: deployments-aggregator
spec:
  limit: 10
  filter: obj.kind == "Deployment"
  combinedFields:
    - name: deployed-replicas-sum
      type: SUM
      subject: returned.status.availableReplicas
--

kind: StatusCollector
metadata:
  name: deployments-stale
spec:
  limit: 10
  filter: obj.kind == "Deployment" && propagation.stale == true
  combinedFields:
    - name: count
      type: COUNT
```

#### Generated Objects

Binding:
```yaml
- apiVersion: control.kubestellar.io/v1alpha1
  kind: Binding
  metadata:
    creationTimestamp: "2024-08-15T15:43:17Z"
    finalizers:
    - transport.kubestellar.io/object-cleanup
    generation: 13
    name: nginx-bindingpolicy
    ownerReferences:
    - apiVersion: control.kubestellar.io/v1alpha1
      blockOwnerDeletion: false
      controller: true
      kind: BindingPolicy
      name: nginx-bindingpolicy
      uid: 5a9ffb8c-244f-459c-a7f4-9359d8be4f33
    resourceVersion: "2709"
    uid: 14eb026b-45cb-4437-a73e-071599e807c7
  spec:
    destinations:
    - clusterId: cluster1
    - clusterId: cluster2
    workload:
      clusterScope:
      - group: ""
        name: nginx
        resource: namespaces
        resourceVersion: "309"
        stalenessThresholdSecs: 100
        statusCollectors:
        - deployments-aggregator
        - deployments-stale
        version: v1
      namespaceScope:
      - group: apps
        name: nginx
        namespace: nginx
        resource: deployments
        resourceVersion: "342"
        stalenessThresholdSecs: 100
        statusCollectors:
        - deployments-aggregator
        - deployments-stale
        version: v1
```

CombinedStatus
```yaml
- apiVersion: control.kubestellar.io/v1alpha1
  kind: CombinedStatus
  metadata:
    creationTimestamp: "2024-08-15T15:43:22Z"
    generation: 21
    labels:
      status.kubestellar.io/api-group: apps
      status.kubestellar.io/binding-policy: nginx-bindingpolicy
      status.kubestellar.io/name: nginx
      status.kubestellar.io/namespace: nginx
      status.kubestellar.io/resource: deployments
    name: f4b1abaf-6149-4bdb-b162-5ad0cebdada3.5a9ffb8c-244f-459c-a7f4-9359d8be4f33
    namespace: nginx
    resourceVersion: "2712"
    uid: 878e457b-e7b9-47f5-968c-1b18b0a4b6e9
  results:
  - columnNames:
    - deployed-replicas-sum
    name: deployments-aggregator
    rows:
    - columns:
      - float: "2"
        type: Number
  - columnNames:
    - count
    name: deployments-stale
    rows:
    - columns:
      - float: "2"
        type: Number
```

When the `BindingPolicy` is changed with a high staleness-threshold, the resulting `CombinedStatus` would be:
```yaml
- apiVersion: control.kubestellar.io/v1alpha1
  kind: CombinedStatus
  metadata:
    creationTimestamp: "2024-08-15T15:43:22Z"
    generation: 24
    labels:
      status.kubestellar.io/api-group: apps
      status.kubestellar.io/binding-policy: nginx-bindingpolicy
      status.kubestellar.io/name: nginx
      status.kubestellar.io/namespace: nginx
      status.kubestellar.io/resource: deployments
    name: f4b1abaf-6149-4bdb-b162-5ad0cebdada3.5a9ffb8c-244f-459c-a7f4-9359d8be4f33
    namespace: nginx
    resourceVersion: "3085"
    uid: 878e457b-e7b9-47f5-968c-1b18b0a4b6e9
  results:
  - columnNames:
    - deployed-replicas-sum
    name: deployments-aggregator
    rows:
    - columns:
      - float: "2"
        type: Number
  - columnNames:
    - count
    name: deployments-stale
```
## Related issue(s)
- #2141


api/control/v1alpha1/types.go
```diff
@@ -153,7 +153,19 @@ type DownsyncPolicyClause struct {
 	// +optional
 	CreateOnly bool `json:"createOnly,omitempty"`
 
-	// statusCollectors is a list of StatusCollectors name references that are applied to the selected objects.
+	// `statusCollection` holds the rules for collecting status from the WECs.
```
MikeSpreitzer, line 156: Why introduce this level of structure for BindingPolicy objects and not for Binding objects?

api/control/v1alpha1/types.go
```diff
@@ -290,6 +302,11 @@ type NamespaceScopeDownsyncClause struct {
 
 	// `statusCollectors` is a list of StatusCollectors name references that are applied to the object.
 	StatusCollectors []string `json:"statusCollectors,omitempty"`
+
+	// `stalenessThresholdSecs` is the threshold used to derive
+	// the `stale` bit in PropagationMeta values.
+	// The value is the minimum of the `stalenessThresholdSecs` in the tests that match the object.
```
MikeSpreitzer, line 308: "The value"? _which_ value? This value here? I suspect not. I suspect you mean to say something like:

> The staleness threshold used in calculating a. workload object's `stale` bit is the lowest of the thresholds given in the matching Binding objects.

api/control/v1alpha1/types.go
```diff
@@ -443,6 +465,27 @@ const (
 // Type checking errors are posted to the status.Errors of the Binding and BindingPolicy.
 type Expression string
 
+// Augmentation defines what is implicitly added to every workload object from a WEC,
+// for purposes of status collection.
+type Augmentation struct {
+	Inventory   InventoryObjectReference `json:"inventory"`
+	Propagation PropagationMeta          `json:"propagation"`
+}
```
MikeSpreitzer, line 473: This is not a good explanation of the design actually being pursued. The design being pursued does not add fields to "the" object structure. Instead, the top-level names available to expressions are "obj", "returned", "inventory", and "propagation".

api/control/v1alpha1/types.go
```diff
@@ -443,6 +465,27 @@ const (
...
+// Inventory is queryable in StatusCollector expressions using the `inventory` object.
+type InventoryObjectReference struct {
+	// `name` is the inventory object's name
+	Name string `json:"name"`
+}
+
+// PropagationMeta is the last reported metadata about one workload object at one WEC.
+// PropagationMeta is queryable in StatusCollector expressions using the `propagation` object.
+type PropagationMeta struct {
+	// `stale` indicates whether `infoAgeSeconds > stalenessThresholdSecs`.
```
MikeSpreitzer, line 485: "infoAgeSeconds" is not defined.

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -49,13 +49,25 @@ type bindingPolicyResolution struct {
...
+	// by at least one downsync clause with the createOnly bit set.
+	// If enabled, it means that the object should only be created and not
+	// maintained
+	CreateOnly bool
+	// StatusCollectors is a set of status collector names that are associated
 	StatusCollectors sets.Set[string]
+	// stalenessThresholdSecs dictates the staleness threshold for determining
+	// when the object's collected status is considered stale.
+	// If more than one BindingPolicy downsync clause selected the object,
+	// the minimum staleness threshold is used.
```
MikeSpreitzer, line 69: s/is used/appears here/

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -49,13 +49,25 @@ type bindingPolicyResolution struct {
...
+	// CreateOnly is a boolean that indicates whether the object was selected
+	// by at least one downsync clause with the createOnly bit set.
+	// If enabled, it means that the object should only be created and not
+	// maintained
+	CreateOnly bool
+	// StatusCollectors is a set of status collector names that are associated
 	StatusCollectors sets.Set[string]
+	// stalenessThresholdSecs dictates the staleness threshold for determining
+	// when the object's collected status is considered stale.
+	// If more than one BindingPolicy downsync clause selected the object,
```
MikeSpreitzer, line 68: s/more than one BindingPolicy downsync clause selected the object/more than one resolution references the object/

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -49,13 +49,25 @@ type bindingPolicyResolution struct {
...
+	ResourceVersion string
+	// CreateOnly is a boolean that indicates whether the object was selected
+	// by at least one downsync clause with the createOnly bit set.
+	// If enabled, it means that the object should only be created and not
+	// maintained
+	CreateOnly bool
+	// StatusCollectors is a set of status collector names that are associated
 	StatusCollectors sets.Set[string]
+	// stalenessThresholdSecs dictates the staleness threshold for determining
+	// when the object's collected status is considered stale.
```
MikeSpreitzer, line 67: _collected_ status considered stale? As far as I understand, the propagation augmentation's `stale` bit is about whether what the core knows about status from the WEC is "stale", and this is _input_ to the aggregation function. BTW: since this input can change with the passage of time, what gets the aggregation re-evaluated when this input changes?

vMaroon: I'm not sure I understand the difference in the first part.

Regarding the 2nd, currently what causes for reevaluation is any time a workstatus is evaluated, which can occur on the three event streams the resolver is affected by (changes in bindings, changes in statuscollectors and changes in workstatuses).

I'm thinking about using delayed queuing to trigger re-evaluation, where once one of the above triggers an evaluation, the delay is calculated to be the time left for status to become stale, if it isn't.

api/control/v1alpha1/types.go
vMaroon: Added support for infoAgeSeconds.

MikeSpreitzer:
The documentation belongs in https://docs.kubestellar.io/unreleased-development/direct/combined-status/ . There is substantial content currently in the opening comment of this PR, which I would prefer to comment on using the usual GitHub features for commenting on file revisions.

api/control/v1alpha1/types.go
```diff
@@ -153,7 +153,19 @@ type DownsyncPolicyClause struct {
 	// +optional
 	CreateOnly bool `json:"createOnly,omitempty"`
 
-	// statusCollectors is a list of StatusCollectors name references that are applied to the selected objects.
+	// `statusCollection` holds the rules for collecting status from the WECs.
+	// +optional
+	StatusCollection *StatusCollection `json:"statusCollection,omitempty"`
```
MikeSpreitzer, line 158: Does a non-zero default value for StalenessThresholdSecs play nice with "omitempty" and "+optional"?

api/control/v1alpha1/types.go
```diff
@@ -443,6 +457,42 @@ const (
 // Type checking errors are posted to the status.Errors of the Binding and BindingPolicy.
 type Expression string
 
+// Augmentation defines what is implicitly queryable in the CEL expressions.
+// This means that while the listed fields are not part of the object's schema,
+// they can be referred to in the CEL expressions.
+//
+// Besides for Augmentations, the other objects queryable in CEL expressions are:
```
MikeSpreitzer, line 464: Rather than define Augmentation and also document two things "besides Augmentation", I think it would be simpler to just have one level of documentation that defines all four top-level variables usable in the expressions.

api/control/v1alpha1/types.go
```diff
@@ -443,6 +457,42 @@ const (
...
+// Inventory is queryable in StatusCollector expressions using the `inventory` object.
+type InventoryObjectReference struct {
+	// `name` is the inventory object's name
+	Name string `json:"name"`
+}
+
+// PropagationMeta is the last reported metadata about one workload object at one WEC.
+// PropagationMeta is queryable in StatusCollector expressions using the `propagation` object.
+type PropagationMeta struct {
+	// `infoAgeSeconds` is the age of the object in seconds.
```
MikeSpreitzer, line 490: With respect to a given WEC, an object has two ages: its age in the WDS and its age in the WEC.

My original intent for _info_ age is to characterize how old is the _information_ (about the object) that is in the core but originated in the WEC. How old is the core's copy of the object status and klusterlet success/fail/whatever from the WEC?

api/control/v1alpha1/types.go
vMaroon: Yes.

api/control/v1alpha1/types.go
MikeSpreitzer, line 490: Putting an _age_ field in derived data implies the need to continually update the derived data. One reason for quantization in my original design is to limit the frequency of those recalculations. Now that I think of it, it would be better to put a timestamp than an age field in the augmentation. There could still be quantization that explicitly gives permission for a limited amount of inaccuracy in the timestamp.

pkg/binding/bindingpolicy-resolution.go
MikeSpreitzer: Regarding the first part, another way to put it is to ask exactly what is meant by _collected status_. That is not exactly a term that is used elsewhere in this API. If "collected status" means the result of applying StatusCollectors, that is one thing. If "collected status" means the core's copy of (the object's reported state and the klusterlet's status wrt the object) from the WEC, that is a different thing. So the comment should be updated to be clear. I hope that the clarification says that the age in question is the age of the core's copy of (the object's reported state and the klusterlet's status wrt the object) from the WEC.

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -230,34 +253,52 @@ type objectRef struct {
...
+			ResourceVersion:        clusterScopeDownsyncClause.ResourceVersion,
+			CreateOnly:             clusterScopeDownsyncClause.CreateOnly,
+			StatusCollectors:       statusCollectors,
+			StalenessThresholdSecs: stalenessThresholdSecs,
 		}
 	}
 
 	for _, namespacedScopeDownsyncClause := range bindingSpecWorkload.NamespaceScope {
+		var statusCollectors sets.Set[string]
+		var stalenessThresholdSecs int32
```
MikeSpreitzer, line 284: Default is 120?

api/control/v1alpha1/types.go
MikeSpreitzer: As far as I am concerned, it would be OK to say that the default is zero and zero means "unspecified" and when no threshold is specified then a threshold of 120 seconds is used. Another way to handle "unspecified" is to make StalenessThresholdSecs a _pointer to_ an integer (so that `nil` can represent "unspecified").

pkg/binding/bindingpolicy-resolution.go
vMaroon: Default is 120 when `StatusCollection` is none nil, which is true only when the object matched at least one StatusCollector.

If the latter is true, the threshold will be correctly set. Otherwise its value is useless.

api/control/v1alpha1/types.go
vMaroon: I think it's fine and simple as it is currently: if StatusCollection is set, staleness defaults to 120. If it is not set, staleness is not included. Do you agree with keeping it as it is?

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -49,13 +49,25 @@ type bindingPolicyResolution struct {
...
-	ResourceVersion  string
-	CreateOnly       bool
+type ObjectData struct {
+	// UID is the UID of the object.
+	UID string
+	// ResourceVersion is the resource version of the object.
+	ResourceVersion string
+	// CreateOnly is a boolean that indicates whether the object was selected
+	// by at least one downsync clause with the createOnly bit set.
+	// If enabled, it means that the object should only be created and not
```
MikeSpreitzer, line 61: s/enabled/`true`/

pkg/binding/selectors.go
```diff
@@ -202,3 +207,80 @@ func (c *Controller) removeObjectFromBindingPolicies(ctx context.Context, objIde
...
+			SliceContains(test.ObjectNames, objIdentifier.ObjectName.Name)) {
+			continue
+		}
+		if len(test.ObjectSelectors) > 0 && !labelsMatchAny(c.logger, objLabels, test.ObjectSelectors) {
+			continue
+		}
+		if len(test.NamespaceSelectors) > 0 && !ALabelSelectorIsEmpty(test.NamespaceSelectors...) {
+			if objNS == nil {
+				var err error
+				objNS, err = c.kubernetesClient.CoreV1().Namespaces().Get(context.TODO(),
```
MikeSpreitzer, line 250: Why not use `ctx`?

api/control/v1alpha1/types.go
vMaroon: Agreed. Done.

api/control/v1alpha1/types.go
MikeSpreitzer: infoAgeSeconds is no longer defined.
LastUpdateTimestamp is not suitable for judging whether the core has stale information. The core might have very fresh information that the object was last updated a long time ago.

pkg/binding/selectors.go
```diff
@@ -202,3 +207,80 @@ func (c *Controller) removeObjectFromBindingPolicies(ctx context.Context, objIde
...
+		if test.StatusCollection != nil {
+			matchedStatusCollectors.Insert(test.StatusCollection.StatusCollectors...)
+			minStalenessThresholdSecs = minInt(minStalenessThresholdSecs,
+				test.StatusCollection.StalenessThresholdSecs)
+		}
+		matched = true
+		createOnly = createOnly || test.CreateOnly
+	}
+
+	if len(matchedStatusCollectors) > 0 && minStalenessThresholdSecs == math.MaxInt32 {
```
MikeSpreitzer, line 274: When would this condition be true?

pkg/status/cel-evaluator.go
```diff
@@ -48,7 +54,8 @@ func newCELEvaluator() (*celEvaluator, error) {
 		cel.Declarations(
 			decls.NewVar(sourceObjectKey, decls.NewMapType(decls.String, decls.Dyn)),
 			decls.NewVar(returnedKey, decls.NewMapType(decls.String, decls.Dyn)),
-			decls.NewVar(inventoryKey, decls.NewMapType(decls.String, decls.String)),
+			decls.NewVar(inventoryKey, decls.NewMapType(decls.String, decls.String)), // contains name:string only
```
MikeSpreitzer, line 57: Not new in this PR, but FYI: it is possible to also declare that this map has only one entry (with key = "name"), and that should be done (not necessarily in this PR).

pkg/status/cel-evaluator.go
```diff
@@ -48,7 +54,8 @@ func newCELEvaluator() (*celEvaluator, error) {
...
+			decls.NewVar(propagationMetaKey, decls.NewMapType(decls.String, decls.Dyn)),
```
MikeSpreitzer, line 58: It is possible, and preferable, to declare the specific keys in this map and the type of each one's associated value.

pkg/status/combinedstatus-resolution.go
```diff
@@ -346,10 +362,10 @@ func (c *combinedStatusResolution) requiresSourceObjectMetaOrSpec() bool {
...
 
 	pred := func(s string) bool {
-		return strings.Contains(s, fmt.Sprintf("%s.metadata", expressionRootName)) ||
-			strings.Contains(s, fmt.Sprintf("%s.spec", expressionRootName)) ||
-			strings.Contains(s, fmt.Sprintf("%s.kind", expressionRootName)) ||
-			strings.Contains(s, fmt.Sprintf("%s.apiVersion", expressionRootName))
+		return strings.Contains(s, fmt.Sprintf("%s.metadata", sourceObjectKey)) ||
+			strings.Contains(s, fmt.Sprintf("%s.spec", sourceObjectKey)) ||
+			strings.Contains(s, fmt.Sprintf("%s.kind", sourceObjectKey)) ||
+			strings.Contains(s, fmt.Sprintf("%s.apiVersion", sourceObjectKey))
```
MikeSpreitzer, line 368: Writing `sourceObjectKey + ".metadata"` makes a constant expression, which does not require CPU and heap activity every time requiresSourceObjectMetaOrSpec is invoked.

pkg/status/combinedstatus-resolution.go
```diff
@@ -346,10 +362,10 @@ func (c *combinedStatusResolution) requiresSourceObjectMetaOrSpec() bool {
 	defer c.RUnlock()
 
 	pred := func(s string) bool {
-		return strings.Contains(s, fmt.Sprintf("%s.metadata", expressionRootName)) ||
-			strings.Contains(s, fmt.Sprintf("%s.spec", expressionRootName)) ||
-			strings.Contains(s, fmt.Sprintf("%s.kind", expressionRootName)) ||
-			strings.Contains(s, fmt.Sprintf("%s.apiVersion", expressionRootName))
+		return strings.Contains(s, fmt.Sprintf("%s.metadata", sourceObjectKey)) ||
```
MikeSpreitzer, line 365: Writing `sourceObjectKey + ".metadata"` makes a constant expression, which does not require CPU and heap activity every time requiresSourceObjectMetaOrSpec is invoked.

pkg/binding/selectors.go
vMaroon: Right, removed. Thought this was causing the bug fixed in a push somewhere.

pkg/status/combinedstatus-resolution.go
vMaroon: This function specifically checks for spec and metadata. It should later be rewritten to check if any augmentation is used and make that process more efficient.

api/control/v1alpha1/types.go
vMaroon: Right. This can only be propagated by adding a new field to the WorkStatus API object and updating the ocm status agent.

pkg/status/combinedstatus-resolution.go
```diff
@@ -663,6 +679,10 @@ func handleAggregation(scName string, scData *statusCollectorData) *v1alpha1.Nam
 			valuesTuple := make([]string, 0, len(scData.GroupBy))
 			for _, groupByNamedExp := range scData.GroupBy {
 				groupByValue := wsData.groupByEval[groupByNamedExp.Name]
+				if groupByValue == nil {
```
MikeSpreitzer, line 682: Under what circumstances would `groupByValue == nil`?

BTW, this func's name should end with "ReadLocked", right?

pkg/status/workstatus.go
```diff
@@ -167,3 +171,26 @@ func runtimeObjectToWorkStatusRef(obj runtime.Object) (*workStatusRef, error) {
 		SourceObjectIdentifier: objIdentifier,
 	}, nil
 }
+
+func getObjectStatusLastUpdateTime(metaObj metav1.Object) *metav1.Time {
+	if managedFields := metaObj.GetManagedFields(); managedFields != nil && len(managedFields) > 0 {
```
MikeSpreitzer, line 176: `range` has defined behavior, which is the convenient behavior in this case, when the slice is `nil` or empty.

pkg/status/workstatus.go
```diff
@@ -167,3 +171,26 @@ func runtimeObjectToWorkStatusRef(obj runtime.Object) (*workStatusRef, error) {
...
+		// iterate through all managedFields entries to find the one that updated the status
+		for _, field := range managedFields {
+			if field.Operation == metav1.ManagedFieldsOperationUpdate &&
```
MikeSpreitzer, line 179: Why exclude the other operations?

pkg/status/workstatus.go
```diff
@@ -167,3 +171,26 @@ func runtimeObjectToWorkStatusRef(obj runtime.Object) (*workStatusRef, error) {
...
+				field.FieldsType == "FieldsV1" &&
+				field.FieldsV1 != nil {
+				// parse the FieldsV1.Raw to a map to check if "f:status" is present
+				var fieldsMap map[string]interface{}
+				if err := json.Unmarshal(field.FieldsV1.Raw, &fieldsMap); err != nil {
+					continue
+				}
+
+				if _, ok := fieldsMap["f:status"]; ok {
+					return field.Time
```
MikeSpreitzer, line 189: What if different FieldManagers wrote to different parts of the status?

api/control/v1alpha1/types.go
vMaroon: @MikeSpreitzer on 2nd thought, I think both may be interesting to the user.

It depends on how staleness is defined in this case. Staleness under the context of propagation does make sense when set by how recent the sync was. Staleness of objects on the WEC may be better suited to be a part of the `returned` object.

MikeSpreitzer:
The example shown in the opening comment suffers from both problems reported in #2400 .

pkg/status/combinedstatus-resolution.go
vMaroon: @MikeSpreitzer  Fixed.

pkg/status/combinedstatus-resolution.go
vMaroon: Was the wrong fix for a different no longer present bug, fixed.

vMaroon:
The latest force-push rebases on upstream main. Comments marked or replied to today were also addressed.

api/control/v1alpha1/types.go
MikeSpreitzer: Oh crud, I see that OCM's ManifestWork does not report on how fresh is the information in the core. It only reports on the time of the last change, not how recently the core was informed that there has been no further change. The point that I was trying to get at with "infoAge" is how old or fresh is the core's information from the WEC.

For this PR we have to content ourselves with something that is supported by ManifestWork. For a given workload object, I think the best that we can do is expose the binary condition of whether the version of that object that is currently in the WDS has been successfully propagated into the WEC. I think that would be derived from the conjunction of the following tests.

- Is the ResourceVersion read from the WDS equal to the ResourceVersion in the ManifestWork?
- Does the ObservedGeneration in the ManifestWork.Status.Conditions equal the Generation of the ManifestWork?
- Does the ManifestWork.Status.ResourceStatus.Manifests[the index for the object].Conditions have an entry for "Applied" with a status of "True"?
- Maybe also the same for "Available"? I am not sure exactly what this one means in the Conditions for an individual object (the comment in https://github.com/open-cluster-management-io/api/blob/main/work/v1/types.go looks like it was written only with ManifestWork.Status.Conditions in mind).

pkg/binding/bindingpolicy-resolution.go
MikeSpreitzer: This is moot now.

api/control/v1alpha1/types.go
MikeSpreitzer: This is moot now.

api/control/v1alpha1/types.go
MikeSpreitzer: I am not convinced, but I think it is moot since ManifestWork does not support the useful notion of "stale".

pkg/binding/bindingpolicy-resolution.go
MikeSpreitzer: I think this is moot now, having realized that ManifestWork does not support the useful kind of "stale".

pkg/status/combinedstatus-resolution.go
```diff
@@ -1008,3 +1038,46 @@ func sortedStringSlice(s []string) []string {
...
+		}
+	}
+}
+
+// isWholeWord checks whether the word at `idx` in `s` is a whole word.
+// s is assumed to be non-nil.
+func isWholeWord(s *string, idx, length int) bool {
+	// check preceding rune
+	if idx > 0 {
+		r := rune((*s)[idx])
```
MikeSpreitzer, line 1066: This does not fetch the _preceding_ rune.

pkg/status/combinedstatus-resolution.go
```diff
@@ -1008,3 +1038,46 @@ func sortedStringSlice(s []string) []string {
...
+	for {
+		idx = strings.Index((*query)[idx:], obj) // slices share the same storage, therefore efficient
+		if idx == -1 {
+			return false
+		}
+
+		if isWholeWord(query, idx, len(obj)) {
+			return true
+		}
+	}
```
MikeSpreitzer, line 1058: Overlooked increase in `idx` here?

pkg/status/combinedstatus-resolution.go
```diff
@@ -1008,3 +1038,46 @@ func sortedStringSlice(s []string) []string {
 	sort.Sort(sort.StringSlice(s))
 	return s
 }
+
+// objectIsQueried checks whether `obj` appears in `query` as a standalone
+// occurrence.
+// That means, it must not be surrounded by alphanumeric characters.
+// query is assumed to be non-nil.
+func objectIsQueried(query *string, obj string) bool {
```
MikeSpreitzer, line 1046: Why does this take a _pointer to_ a string instead of simply a string?

pkg/status/workstatus.go
```diff
@@ -168,3 +172,27 @@ func runtimeObjectToWorkStatusRef(obj runtime.Object) (*workStatusRef, error) {
...
+		if field.FieldsType == "FieldsV1" &&
+			field.FieldsV1 != nil {
+			// parse the FieldsV1.Raw to a map to check if "f:status" is present
+			var fieldsMap map[string]interface{}
+			if err := json.Unmarshal(field.FieldsV1.Raw, &fieldsMap); err != nil {
+				continue
+			}
+
+			if _, ok := fieldsMap["f:status"]; ok {
+				if field.Time.After(latestTime.Time) {
```
MikeSpreitzer, line 190: What if `latestTime == nil`?

pkg/status/workstatus.go
```diff
@@ -168,3 +172,27 @@ func runtimeObjectToWorkStatusRef(obj runtime.Object) (*workStatusRef, error) {
...
+					latestTime = field.Time
+				}
+			}
+		}
+	}
+
+	return nil
```
MikeSpreitzer, line 197: This looks wrong.

pkg/status/combinedstatus-resolution.go
vMaroon: Oops, fixed.

pkg/status/combinedstatus-resolution.go
vMaroon: To make it slightly more efficient under the non-nil guarantee., assuming this will become a hot-function.

Passing a string by value copies 16 bytes, passing by pointer copies 8 bytes. I haven't ran benchmarks on whether this brings more value given the dereferences.

pkg/status/workstatus.go
vMaroon: It cannot be nil. 

`latestTime := &metav1.Time{}` initializes it to zero time (and grabs the pointer to it).
Afterwards, it is only updated through this line.

pkg/status/workstatus.go
vMaroon: Fixed.

pkg/status/combinedstatus-resolution.go
vMaroon: Fixed.

MikeSpreitzer:
When a PR gets to this many commits, particularly when many of them going in different directions from earlier commits in the same PR, I think that it is beneficial to squash the commits before merging.

MikeSpreitzer:
The example in the opening comment of this PR no longer represents what this PR proposes to do.

The user guide should have an example.

api/control/v1alpha1/types.go
MikeSpreitzer: This is moot at the moment, since the full description has been removed from type.go. However, I think that types.go should have a full description.

pkg/status/workstatus.go
MikeSpreitzer: Indeed, that is true of the current code. I was wondering if you had intended to return `nil` in the case of zero matching fields. If the intent is never to return a nil then why is `workStatus` declared to hold a _pointer to_ a time rather than simply a time?

api/control/v1alpha1/types.go
```diff
@@ -329,8 +342,9 @@ type ClusterScopeDownsyncClause struct {
 	// +optional
 	CreateOnly bool `json:"createOnly,omitempty"`
 
-	// `statusCollectors` is a list of StatusCollectors name references that are applied to the object.
-	StatusCollectors []string `json:"statusCollectors,omitempty"`
+	// `statusCollection` holds the rules of status collection for the object.
+	// +optional
+	StatusCollection *StatusCollection `json:"statusCollection,omitempty"`
 }
```
MikeSpreitzer, line 349: I think that the types.go file should have a declaration of the data structure that the expressions are evaluated against. I see that this PR currently documents that in the user guide, but: (A) the description there is not as precise as a data structure, and programmers need precision, and (B) types.go is where programmers go for the real API definition.

It could look something like the following.

```go
// ExpressionContext defines what an Expression can reference regarding the workload object.
type ExpressionContext struct {
	// `inventory` holds the inventory record for the workload object
	Inventory InventoryRecord `json:"inventory"`

	// `obj` holds a copy of the workload object as read from the WDS
	Obj map[string]any `json:"obj"`

	// `returned` holds the fragment of the workload object that was returned to the core from the WEC
	Returned ReturnedState `json:"returned"`

	// `propagation` holds data about the current state of the work on propagating
	// the object's state from WDS to WEC and from WEC to WDS.
	Propagation PropagatationData `json:"propagatation"`
}

// InventoryRecord is what appears in the inventory for a given WEC.
type InventoryRecord struct {
	// the name of the WEC
	Name string `json:"name"`
}

type ReturnedState struct {
	Status map[string]any `json:"status"`
}

type PropagatationData struct {
	// `lastReturnedUpdateTimestamp` is the time of the last update to any
	// of the returned object state in the core.
	// Before the first such update, this is the zero value of `time.Time`.
	LastReturnedUpdateTimestamp time.Time `json:"lastReturnedUpdateTimestamp"`
}
```

docs/content/direct/combined-status.md
```diff
@@ -94,6 +96,22 @@ When there are N "GROUP BY" columns, the result has a row for each tuple of valu
...
+   - `inventory.name`: The name of the inventory object.
+
+1. `obj`: The workload object from the WDS:
+   - All fields of the workload object except the status subresource.
+
+1. `returned`: The reported state from the WEC:
+   - `returned.status`: The status section of the object returned from the WEC.
+
+1. `propagation`: Metadata about the end-to-end propagation process:
+   - `propagation.lastReturnedUpdateTimestamp`: Time of last update to any returned state.
```
MikeSpreitzer, line 113: Someone writing an expression using this will need to know exactly what sort of "time" this is.

docs/content/direct/combined-status.md
```diff
@@ -23,6 +23,8 @@ The general technique involves the following ideas.
 
 1. The reported state for a given workload object from a given WEC is implicitly augmented with metadata about the WEC and about the end-to-end propagation from WDS to that WEC. This extra information is available just like the regular contents of the object, for use in combining reported state.
 
+1. The reported state from each WEC is augmented with metadata about the WEC and the end-to-end propagation process. This augmented information is available alongside other object contents for use in combining reported state, as specified [Queryable Objects](#queryable-objects).
```
MikeSpreitzer, line 26: This item overlaps with the previous one.

pkg/binding/selectors.go
```diff
@@ -202,3 +207,76 @@ func (c *Controller) removeObjectFromBindingPolicies(ctx context.Context, objIde
...
+			if !labelsMatchAny(logger, objNS.Labels, test.NamespaceSelectors) {
+				continue
+			}
+		}
+
+		klog.FromContext(ctx).V(5).Info("Workload object matched test", "objIdentifier", objIdentifier, "objLabels", objLabels, "test", test)
+		// test is a match
+		if test.StatusCollection != nil {
+			matchedStatusCollectors.Insert(test.StatusCollection.StatusCollectors...)
+			minStalenessThresholdSecs = minInt(minStalenessThresholdSecs,
```
MikeSpreitzer, line 267: Why is the code still messing with a staleness threshold? There is no "stale" bit in this PR now.

api/control/v1alpha1/types.go
```diff
@@ -153,7 +153,19 @@ type DownsyncPolicyClause struct {
...
+	StatusCollection *StatusCollection `json:"statusCollection,omitempty"`
+}
+
+// StatusCollection holds the rules for collecting status from the WECs.
+type StatusCollection struct {
+	// `stalenessThresholdSecs` is the threshold used to derive
+	// the `stale` bit in PropagationMeta values.
+	// Default value is 120.
+	// +kubebuilder:default=120
+	StalenessThresholdSecs int32 `json:"stalenessThresholdSecs,omitempty"`
```
MikeSpreitzer, line 167: Since there is no "stale" bit anymore, we do not need a threshold here.

pkg/binding/selectors.go
vMaroon: Fixed all instances.

vMaroon:
The last force push is due to squashing and rebasing after addressing comments.

api/control/v1alpha1/types.go
```diff
@@ -457,6 +467,39 @@ const (
...
+
+type ReturnedState struct {
+	Status runtime.RawExtension `json:"status"`
+}
+
+type PropagatationData struct {
+	// `lastReturnedUpdateTimestamp` is the time of the last update to any
+	// of the returned object state in the core.
+	// Before the first such update, this is the zero value of `time.Time`.
+	LastReturnedUpdateTimestamp metav1.Time `json:"lastReturnedUpdateTimestamp"`
```
MikeSpreitzer, line 500: Is `metav1.Time` the best choice here? A previous version of this PR used a type from a protobuf library. I think that we want to optimize for convenience in use in the user's CEL expressions.

MikeSpreitzer (approved):
This is good enough to merge. Needs follow-on work on doc and perhaps the data typing for the timestamp.

/lgtm
/approve
Thanks!

api/control/v1alpha1/types.go
MikeSpreitzer: Looking at https://github.com/google/cel-spec/blob/master/doc/langdef.md#functions, I wonder if `google.protobuf.Timestamp` is a better choice.

MikeSpreitzer: Also, since this has seconds precision, I suspect that a viable choice would be simply an int32 or uint32 holding seconds since Unix epoch (1970.0).

PR #2405, opened 2024-08-16
🌱  Update IDL to forbid duplications in BindingSpec

MikeSpreitzer:
## Summary
This PR updates the definition of Binding objects to explicitly say (in a way that is checked by the apiserver) that duplicates are not allowed in the lists of destinations and workload objects. This is good because it is something that we intend and the consuming code has been silently assuming.

## Related issue(s)

Fixes #


MikeSpreitzer:
FYI, I tried to do this more directly but ran into what I think is a bug in Kubernetes (https://github.com/kubernetes/kubernetes/issues/126747).

vMaroon (commented):
/lgtm

PR #2406, opened 2024-08-19
⚠️  Focus scope of singleton status return request to one clause

MikeSpreitzer:
## Summary
This PR changes where the bit that requests singleton reported state return appears, from a whole BindingPolicy object to an individual clause in that object. This PR also updates the description, attempting to be more clear and precise about this request bit.

This PR also makes that bit appear in BindingObjects (it should have, but wasn't).

## Related issue(s)

Fixes #


MikeSpreitzer:
The force-push to 3dbebcd80 is a rebase onto `main`.

MikeSpreitzer:
I think that we are pursuing the API in #2463 instead.

PR #2407, opened 2024-08-19
⚠️  Generalize singleton return and merge into combined

MikeSpreitzer:
## Summary
This PR changes the API for singleton reported state return, primarily in two ways. One is to generalize so that instead of always returning exactly the whole contents of `.status`, the user programs what gets returned. The other is to move the request bit into the StatusCollector object.

I chose to not ignore the `filter`. It seems strange to ignore it, it is right there and has natural semantics. Except, after thinking it through for a while, it gets rather intricate to say exactly what the semantics are. I made one choice that is coarser than necessary: the counting of the relevant WECs does not take into account which parts of the object's state are selected. I made this choice for the sake of simplicity, and I think it fits with the use cases that I expect.

I really struggled with writing the English words. I assume that English is preferable to first order predicate logic, but the ideas being communicated are that complicated. In English "any" can mean either "every" or "some". And "some" implies an existential quantifier without being precise about where it appears.

## Related issue(s)

Fixes #


api/control/v1alpha1/types.go
```diff
@@ -366,9 +356,13 @@ type BindingList struct {
...
 
-// StatusCollector defines one way to collect status about a given workload object from
+// StatusCollector defines a way to collect and report status about a given workload object from
 // the set of WECs that it propagates to.
-// This is modeled after an SQL SELECT statement that does aggregation.
+// There are two styles of StatusCollectors, one general and one specialized.
+// The general style is modeled after a simple SQL SELECT statement and may or may not do aggregation;
+// the corresponding results go in CombinedStatus objects.
+// The specialized style is meant for the case where propagation is to exactly 1 WEC and picks out
+// parts of the object state to propagate back from that WEC to the WDS.
```
vMaroon, line 365: Could the status subresource contain fields that must be set once status is not empty?

MikeSpreitzer:
The force-push to 298bef35e is a rebase onto `main`.

MikeSpreitzer:
I think that we are pursuing #2463 instead.

PR #2408, opened 2024-08-20
⚠️  Narrow scope of singleton status return request

MikeSpreitzer:
## Summary
This PR is a variant on #2406 . The difference is that the design in this PR says that Bindings and BindingPolicies that do NOT request singleton reported state return do not affect the semantics or behavior regarding Bindings and BindingPolicies that _do_ request it.

The English description is more than a little different. I tried taking a different approach to the writing, defining a couple of concepts before using them in prescribing the API.

## Related issue(s)

Fixes #

### DIFF
--- a/api/control/v1alpha1/types.go
+++ b/api/control/v1alpha1/types.go
@@ -104,15 +104,6 @@ type BindingPolicySpec struct {
 	// the `createOnly` bits are ORed together, and the StatusCollector reference
 	// sets are combined by union.
 	Downsync []DownsyncPolicyClause `json:"downsync,omitempty"`
-
-	// WantSingletonReportedState means that for workload objects that are distributed --- taking
-	// all BindingPolicies into account --- to exactly one WEC, the object's reported state
-	// from the WEC should be written to the object in its WDS.
-	// If any of the workload objects are distributed to more or less than 1 WEC then
-	// the `.status.errors` of this policy will report that discrepancy for
-	// some of them.
-	// +optional
-	WantSingletonReportedState bool `json:"wantSingletonReportedState,omitempty"`
 }
 
 const (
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
The force-push to 29d155862 is a rebase onto `main`.

api/control/v1alpha1/types.go
```diff
@@ -144,6 +135,37 @@ type DownsyncPolicyClause struct {
...
+	// While singleton status return is _not_ requested, KubeStellar suppresses
+	// the existence of a label with that name (key).
+	// While singleton status return is requested _and_ the size of the qualified
+	// WEC set is 1, KubeStellar propagates the object's `.status` from that WEC
+	// to the `.status` section of the object in the WDS.
+	// While either singleton status return is NOT requested or the size of the
+	// qualified WEC set is NOT 1, the `.status` of the object in the WDS does
+	// not have contents propagated there from a WEC by KubeStellar.
+	// +optional
+	WantSingletonReportedState bool `json:"wantSingletonReportedState,omitempty"`
```
vMaroon, line 168: Why not move this inside `StatusCollection`?

MikeSpreitzer: Because `StatusCollection` is just a set of references to `StatusCollector` objects. The singleton state return does not modulate the meaning of those or otherwise affect the `CombinedStatus` objects.

Something that I _could_ support would be to refactor `DownsyncPolicyClause` so that it has just two fields: `DownsyncObjectTest` and `DownsyncModulation` (or some other appropriate new name), where the latter puts together all the things that modulate the downsync behavior --- i.e., `CreateOnly bool`, `WantSingletonReportedState bool`, and `StatusCollectors []string`.

vMaroon: That sounds good. My main concern is that such a flag should be at the test-level, since it may make no sense to apply it to some resources in a workload. Therefore the user may duplicate/narrow-down object selectors and enable status collection in only some of them.

MikeSpreitzer: OK. I made a new PR in that style. Please see #2463 .

MikeSpreitzer:
We are pursuing #2463 instead.

PR #2409, opened 2024-08-20
✨ Improve status controller using existing assets

waltforme:
## Summary
This PR uses some existing data structures and logic in (also adds some to) the binding controller, to help the status controller handle singleton statuses.

Now status controller
    - doesn’t have functionally duplicated code;
    - doesn't have to read the `managed-by.kubestellar.io/singletonstatus` label when reconciling singleton statuses.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/binding/bindingpolicy-resolver.go
+++ b/pkg/binding/bindingpolicy-resolver.go
@@ -115,6 +115,11 @@ type BindingPolicyResolver interface {
 	// otherwise the returned `int` is unspecified.
 	GetSingletonReportedStateRequestForObject(util.ObjectIdentifier) (bool, int)
 
+	// GetSingletonReportedStateRequestsForBinding calls GetSingletonReportedStateRequestForObject
+	// for each of workload objects in the resolution if the resolution exists.
+	// If the resolution doesn't exist, empty slices are returned.
+	GetSingletonReportedStateRequestsForBinding(bindingPolicyKey string) ([]util.ObjectIdentifier, []bool, []int)
+
 	// DeleteResolution deletes the resolution associated with the given key,
 	// if it exists.
 	DeleteResolution(bindingPolicyKey string)
@@ -343,6 +348,21 @@ func (resolver *bindingPolicyResolver) GetSingletonReportedStateRequestForObject
 		matchingWECs = matchingWECs.Union(thisDests)
 	}
 	return requested, matchingWECs.Len()
+}
+
+func (resolver *bindingPolicyResolver) GetSingletonReportedStateRequestsForBinding(bindingPolicyKey string) ([]util.ObjectIdentifier, []bool, []int) {
+	resolver.RWMutex.RLock()
+	defer resolver.RWMutex.RUnlock()
+
+	objIds, requests, nWECs := []util.ObjectIdentifier{}, []bool{}, []int{}
+	if !resolver.ResolutionExists(bindingPolicyKey) {
+		return objIds, requests, nWECs
+	}
+	for objId := range resolver.getResolution(bindingPolicyKey).objectIdentifierToData {
+		r, n := resolver.GetSingletonReportedStateRequestForObject(objId)
+		objIds, requests, nWECs = append(objIds, objId), append(requests, r), append(nWECs, n)
+	}
+	return objIds, requests, nWECs
 }
 
 // DeleteResolution deletes the resolution associated with the given key,
--- a/pkg/status/singletonstatus.go
+++ b/pkg/status/singletonstatus.go
@@ -171,6 +171,35 @@ func (c *Controller) reconcileSingletonByWs(ctx context.Context, ref singletonWo
 	}
 }
 
+func (c *Controller) alsoReconcileSingletonByWS(ctx context.Context, ref singletonWorkStatusRef) error {
+	logger := klog.FromContext(ctx)
+	logger.V(4).Info("Reconciling singleton status due to workstatus changes", "name", string(ref.Name))
+	wObjID := ref.SourceObjectIdentifier
+
+	requested, nWECs := c.bindingPolicyResolver.GetSingletonReportedStateRequestForObject(wObjID)
+	if !requested || nWECs != 1 {
+		logger.V(4).Info("Singleton workload object should not have status synced, cleaning",
+			"resource", ref.SourceObjectIdentifier.Resource, "objectName", ref.SourceObjectIdentifier.ObjectName,
+			"requested", requested, "nWECs", nWECs)
+		return c.reconcileSingletonWObj(ctx, wObjID, false)
+	} else {
+		logger.V(4).Info("Singleton workload object should have status synced, updating",
+			"resource", ref.SourceObjectIdentifier.Resource, "objectName", ref.SourceObjectIdentifier.ObjectName)
+		wsObj, err := c.workStatusLister.ByNamespace(ref.WECName).Get(ref.Name)
+		if err != nil {
+			return err
+		}
+		status, err := util.GetWorkStatusStatus(wsObj)
+		if err != nil {
+			return err
+		}
+		if status == nil {
+			return nil
+		}
+		return updateObjectStatus(ctx, &wObjID, status, c.listers, c.wdsDynClient)
+	}
+}
+
 func (c *Controller) checkSingletonWorkloadInBinding(ctx context.Context, binding v1alpha1.Binding, sync map[util.ObjectIdentifier]bool, init bool) error {
 	logger := klog.FromContext(ctx)
 	logger.V(5).Info("Checking singleton workload object", "binding", binding.Name)
--- a/pkg/status/singletonstatus.go
+++ b/pkg/status/singletonstatus.go
@@ -33,18 +33,18 @@ func (c *Controller) reconcileSingletonByBdg(ctx context.Context, bdgName string
 	for i := range wObjIDs {
 		wObjID, r, n := wObjIDs[i], requested[i], nWECs[i]
 		if !r || n != 1 {
-			logger.V(4).Info("Workload object should not have singleton status synced, cleaning",
-				"resource", wObjID.Resource, "objectName", wObjID.ObjectName,
-				"requested", r, "nWECs", n)
 			if err := c.reconcileSingletonWObj(ctx, wObjID, false); err != nil {
 				return err
 			}
+			logger.V(4).Info("Cleaned singleton status for workload object",
+				"gvk", wObjID.GVK, "objectName", wObjID.ObjectName,
+				"requested", r, "nWECs", n)
 		} else {
-			logger.V(4).Info("Workload object should have singleton status synced, updating",
-				"resource", wObjID.Resource, "objectName", wObjID.ObjectName)
 			if err := c.reconcileSingletonWObj(ctx, wObjID, true); err != nil {
 				return err
 			}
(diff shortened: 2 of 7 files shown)


pkg/status/singletonstatus.go
```diff
@@ -18,65 +18,71 @@ package status
...
-	if err != nil {
-		if errors.IsNotFound(err) {
-			// A deleted Binding is equivalent to one that references zero workload objects
-			return nil
+	wObjIDs, requested, nWECs := c.bindingPolicyResolver.GetSingletonReportedStateRequestsForBinding(bdgName)
+	for i := range wObjIDs {
+		wObjID, r, n := wObjIDs[i], requested[i], nWECs[i]
+		if !r || n != 1 {
+			logger.V(4).Info("Workload object should not have singleton status synced, cleaning",
+				"resource", wObjID.Resource, "objectName", wObjID.ObjectName,
```
MikeSpreitzer, line 37: Why `wObjID.Resource` instead of `wObjID.GVK`? We _do_ have cases of one Kind name appearing in multiple API groups. And one day --- if we are lucky enough to get this far --- users will report hard-to-diagnose problems caused by the fact that API group _version_ appears in the keys here.

pkg/status/singletonstatus.go
```diff
@@ -18,65 +18,71 @@ package status
...
-		if err := c.reconcileSingletonWObj(ctx, wObjIdentifier, v); err != nil {
+		status, err := util.GetWorkStatusStatus(wsObj)
+		if err != nil {
 			return err
 		}
+		if status == nil {
+			return nil
+		}
+		logger.V(4).Info("Updating singleton status", "workstatus", ref.Name, "wecName", ref.WECName)
+		return updateObjectStatus(ctx, &wObjID, status, c.listers, c.wdsDynClient)
```
MikeSpreitzer, line 79: Why does updateObjectStatus take a _pointer to_ the object ID rather than simply take the object ID?

MikeSpreitzer, line 79: Not introduced in this PR and so does not have to be addressed here, but why does https://github.com/kubestellar/kubestellar/blob/305e28105974b17fda2793eb4aed005ce164b553/pkg/status/workstatus.go#L97 unconditionally use a namespace to access what might be a non-namespaced object?

pkg/status/singletonstatus.go
```diff
@@ -18,65 +18,71 @@ package status
...
-	binding, err := c.wdsKsClient.ControlV1alpha1().Bindings().Get(ctx, bdgName, metav1.GetOptions{})
-	if err != nil {
-		if errors.IsNotFound(err) {
-			// A deleted Binding is equivalent to one that references zero workload objects
-			return nil
+	wObjIDs, requested, nWECs := c.bindingPolicyResolver.GetSingletonReportedStateRequestsForBinding(bdgName)
+	for i := range wObjIDs {
+		wObjID, r, n := wObjIDs[i], requested[i], nWECs[i]
+		if !r || n != 1 {
+			logger.V(4).Info("Workload object should not have singleton status synced, cleaning",
```
MikeSpreitzer, line 36: Look at how many `V(4)` log statements are on the code path here, all that get executed in this case. I think there are too many, and they all come _before_ the actual change. I think that there should be _one_ `V(4)` log statement in this case, after the change is successfully made, indicating that it was done and why. There may be additional `V(5)` log statements before the change explaining what is going on.

pkg/status/singletonstatus.go
```diff
@@ -18,65 +18,71 @@ package status
...
+		logger.V(4).Info("Updating singleton status", "workstatus", ref.Name, "wecName", ref.WECName)
+		return updateObjectStatus(ctx, &wObjID, status, c.listers, c.wdsDynClient)
 	}
-	return nil
 }
 
 func (c *Controller) reconcileSingletonWObj(ctx context.Context, wObjID util.ObjectIdentifier, sync bool) error {
 	logger := klog.FromContext(ctx)
-	logger.V(4).Info("Reconciling singleton workload object", "resource", wObjID.Resource, "objectName", wObjID.ObjectName)
+	logger.V(4).Info("Reconciling workload object for singleton status", "resource", wObjID.Resource, "objectName", wObjID.ObjectName)
```
MikeSpreitzer, line 85: If this log message included the value of the `sync` parameter then the other two in this func would not be needed.

MikeSpreitzer (commented):
Where is it documented how/why this removes the status section of a workload object in the WDS after a Binding is changed from referencing that object and requesting singleton reported state return to not referencing that object?

Other things noted in earlier individual comments.

pkg/status/singletonstatus.go
waltforme: Changed to `wObjID.GVK` everywhere.

pkg/status/singletonstatus.go
waltforme: Reorganized everywhere as suggested.

pkg/status/singletonstatus.go
waltforme: Included the value of `sync`.

BTW, appreciate for looking at all the logging related details so carefully.

waltforme:
> Where is it documented how/why this removes the status section of a workload object in the WDS after a Binding is changed from referencing that object and requesting singleton reported state return to not referencing that object?

Not sure I'm following the question. I think the code is quite self-explanatory on why the status section is removed, no? Are you suggesting that some more comments should be written?

pkg/status/singletonstatus.go
```diff
@@ -18,69 +18,78 @@ package status
...
-	if err != nil {
-		if errors.IsNotFound(err) {
-			// A deleted Binding is equivalent to one that references zero workload objects
-			return nil
+	wObjIDs, requested, nWECs := c.bindingPolicyResolver.GetSingletonReportedStateRequestsForBinding(bdgName)
+	for i := range wObjIDs {
+		wObjID, r, n := wObjIDs[i], requested[i], nWECs[i]
+		if !r || n != 1 {
+			if err := c.reconcileSingletonWObj(ctx, wObjID, false); err != nil {
+				return err
```
MikeSpreitzer, line 37: I think it would be better if this error got wrapped with some explanation, mentioning the binding name, the fact that it arose in the source of syncing the WDS copies of status for that binding, and in particular trying to delete that status. You could do one wrapping here, or do part of the wrapping job in `c.syncBinding` (the stuff that is known at that point) and the remainder here (giving the info not known at the higher wrap site).

This is a generic code pattern issue.

pkg/status/singletonstatus.go
```diff
@@ -18,69 +18,78 @@ package status
...
 	}
+	logger.V(4).Info("Updated singleton status for workload object",
+		"gvk", ref.SourceObjectIdentifier.GVK, "objectName", ref.SourceObjectIdentifier.ObjectName)
 	return nil
 }
 
 func (c *Controller) reconcileSingletonWObj(ctx context.Context, wObjID util.ObjectIdentifier, sync bool) error {
 	logger := klog.FromContext(ctx)
-	logger.V(4).Info("Reconciling singleton workload object", "resource", wObjID.Resource, "objectName", wObjID.ObjectName)
+	logger.V(4).Info("Reconciling workload object for singleton status", "gvk", wObjID.GVK, "objectName", wObjID.ObjectName, "sync", sync)
```
MikeSpreitzer, line 89: I would make this `V(5)`, since all the call sites also have a `V(4)` reporting the same information.

MikeSpreitzer:
The reason that I asked about documentation of how/why this controller removes the status section of a workload object in the WDS after a Binding is changed from referencing that object and requesting singleton reported state return to not referencing that object is that I do not remember the answer or where it is documented --- and I think that it should not be hard to find documented. For the binding controller there is a fairly detailed explanation of how it works. There should be such an explanation for the status controller too.

What I see in the status controller is a `func (c *Controller) reconcileSingletonByBdg` that reacts to things that are said in the given Binding but not to things that are absent from (e.g., recently removed from) that Binding. So I wondered what removes the status from the workload object in the WDS in this case.

The answer may be that the WorkStatus object will be removed and that will cause `func (c *Controller) reconcileSingletonByWS` to be called to handle it. But I think that is not necessarily true. What if there are two Bindings that bind the same workload object to the same WEC, one Binding requesting singleton reported state return and one not requesting it. If the first of those two is modified to no longer reference the workload object, the second Binding will cause the WorkStatus to continue to exist. This gets into the issue of behavior in a situation that we currently tell users to avoid, because we do not have a documented agreement on the proper behavior in this case. So I am not going to insist on a change in this PR.

MikeSpreitzer (approved):
This could use more work, but it can be done in follow-on PRs.

/lgtm
/approve
Thanks!

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -345,6 +350,21 @@ func (resolver *bindingPolicyResolver) GetSingletonReportedStateRequestForObject
 	return requested, matchingWECs.Len()
 }
 
+func (resolver *bindingPolicyResolver) GetSingletonReportedStateRequestsForBinding(bindingPolicyKey string) ([]util.ObjectIdentifier, []bool, []int) {
```
vMaroon, line 353: @MikeSpreitzer @waltforme why is this not a part of a new subscriber to the broker?
I don't think we want to bloat this interface.

This way the status can return to be a user of the broker, and not the resolver - which is a weaker dependency.

Apologies for the late review.

MikeSpreitzer: Can you pleae elaborate on the alternative that you are suggesting? I am not sure I understand exactly what you mean by "this" being "the responsibility of a new subscriber to the broker".

vMaroon: This PR switched the status controller from being a user of the Broker to a user of the Resolver, which is a stronger and unnecessary dependency imo.

All the new Resolver function does is iterate over objects of a resolution given a bindingKey, which can easily be implemented in a new status controller sub-structure that uses the Broker. This way, the status remains a user of the Broker only (which is sort of an informer), and the responsibility of calculating this information is better placed - instead of bloating the binding Resolver.

MikeSpreitzer: Let me see if I have got it right. `Broker` is essentially an informer of resolutions and the status controller should remain a user of only that (possibly indirectly through the new thing mentioned in the following sentence). The function of iterating over the objects of a resolution given a bindingKey should be a method of a new thing that only uses the Broker. Have I got that right?

Would the new thing that only uses the broker go in its own new file, or be added to `bindingpolicy-resolution-broker.go`?

MikeSpreitzer: @vMaroon: are you suggesting that `GetSingletonReportedStateRequestsForBinding` could depend only on one resolution rather than everything that a resolver holds? I do not follow that. `GetSingletonReportedStateRequestsForBinding` needs access to all the resolutions in order to do its job properly, which includes finding whether any resolution requests singleton state return and counting the number of all the associated WECs.

vMaroon: @MikeSpreitzer on the first message, yes. A new structure should reside in the status package and use the broker. On the latter, no, I'm not saying it can depend on a single resolution. I'm saying the broker as-is can provide everything needed in that function.

Right now, the consumer would need to keep track of existing bindings since the broker lacks a method that provides the list or a lazy iterator.

The end point is to keep the binding resolver compact, and have derived logic be built on using the broker.

Issue #2413, opened 2024-08-28
bug: status of cluster-scoped workload object is not propagated back


waltforme:
### Describe the bug

As shown in the title.

### Steps To Reproduce

Looking at the log of the controller manager, we can see something similar to
```
I0828 03:15:24.209486  202005 workstatus.go:101] "Could not find object" logger="Status.worker-0" objectIdentifier={"GVK":{"Group":"","Version":"v1","Kind":"Namespace"},"Resource":"namespaces","ObjectName":{"Namespace":"","Name":"nginx"}}
```

But namespace objects may _not_ be good representatives of cluster-scoped workload objects, because we may _not_ want the status of a namespace object to be propagated back from WEC to WDS.

### Expected Behavior

Status of cluster-scoped workload object should be able to propagated back from WEC to WDS, if such propagation is said to be expected (or, not prohibited) for that workload object.

### Want to contribute?

- [ ] I would like to work on this issue.

### Additional Context

The cause of the bug is documented [here](https://github.com/kubestellar/kubestellar/pull/2409#discussion_r1723979753). But merely adding a lister for cluster object is not sufficient. In addition to adding the missing lister, Kubestellar should be able to be told that, for each workload object, the status should be propagated or not. I think this is related to #1851 .

vMaroon:
@waltforme namespaces are not propagated on the ocm-status-addon level.

https://github.com/kubestellar/ocm-status-addon/blob/v0.2.0-rc10/pkg/agent/agent.go#L39-L41

The code does support it, but I never really tested it. You may want to test with a different cluster-scoped object.

waltforme:
@vMaroon Exactly, so in the opening comment I wrote that namespaces are not good representatives of cluster-scoped workload objects. I still used namespaces only because everyone can easily reproduce the "Could not find object" message.

On the topic of propagation of Namespace status, I saw @MikeSpreitzer has some different opinion. If I understand correctly, looks like @MikeSpreitzer thinks
- the status should be propagated from WEC to ITS;
- but overwrites should not happen in WDS.

MikeSpreitzer:
Right. I do not understand why Namespace, Service, and ServiceAccount are excepted from propagating status from WEC to WorkStatus. Note also that the code currently does NOT except CustomResourceDefinition.

vMaroon:
@MikeSpreitzer based on the comment above the linked code, I guess whoever wrote it wanted to save the creation of "empty" WorkStatuses.

I see two options here:
[...]

PR #2414, opened 2024-08-28
🐛 Add missing lister for cluster-scoped workload object when updating status

waltforme:
## Summary
As shown in the title.

Also handles https://github.com/kubestellar/kubestellar/pull/2409#discussion_r1723976138.
## Related issue(s)

Part of #2413

### DIFF
--- a/pkg/status/workstatus.go
+++ b/pkg/status/workstatus.go
@@ -90,17 +90,22 @@ func updateObjectStatus(ctx context.Context, objectIdentifier *util.ObjectIdenti
 	gvr := objectIdentifier.GVR()
 	lister, found := listers.Get(gvr)
 	if !found {
-		logger.V(2).Info("Could not find lister for gvr", "gvr", gvr)
+		logger.V(4).Info("Could not find lister for gvr", "gvr", gvr)
 		return nil
 	}
 
-	obj, err := lister.ByNamespace(objectIdentifier.ObjectName.Namespace).Get(objectIdentifier.ObjectName.Name)
+	var obj runtime.Object
+	var err error
+	if objectIdentifier.ObjectName.Namespace == "" {
+		obj, err = lister.Get(objectIdentifier.ObjectName.Name)
+	} else {
+		obj, err = lister.ByNamespace(objectIdentifier.ObjectName.Namespace).Get(objectIdentifier.ObjectName.Name)
+	}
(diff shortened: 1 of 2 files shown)


pkg/status/workstatus.go
```diff
@@ -83,25 +83,37 @@ func (c *Controller) syncSingletonWorkStatus(ctx context.Context, ref singletonW
...
+		}
+	} else {
+		obj, err = lister.ByNamespace(objectIdentifier.ObjectName.Namespace).Get(objectIdentifier.ObjectName.Name)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				logger.V(4).Info("Could not find object", "objectIdentifier", objectIdentifier)
+				return nil
+			}
+			return fmt.Errorf("failed to get object (%v): %w", objectIdentifier, err)
 		}
```
MikeSpreitzer, line 116: I would DRY this out

pkg/status/workstatus.go
```diff
@@ -83,25 +83,37 @@ func (c *Controller) syncSingletonWorkStatus(ctx context.Context, ref singletonW
...
-		if errors.IsNotFound(err) {
-			logger.V(2).Info("Could not find object", "objectIdentifier", objectIdentifier)
-			return nil
+	var obj runtime.Object
+	var err error
+	if objectIdentifier.ObjectName.Namespace == "" {
+		obj, err = lister.Get(objectIdentifier.ObjectName.Name)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				logger.V(4).Info("Could not find object", "objectIdentifier", objectIdentifier)
```
MikeSpreitzer, line 103: I would make this log message a little more explanatory, something like "Abandoning update for non-existent workload object" or "Did not update workload object because it is not in local cache, presumably because it was recently deleted".

pkg/status/workstatus.go
waltforme: Good point, done.

waltforme:
Force-push to https://github.com/kubestellar/kubestellar/commit/e44eef371217d23fc9e46bffce92547e4c020c61 includes changes that address the comments and squashes the changes to where they should go.

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

Issue #1851, opened 2024-03-04
feature: flexible trichotomy of downsynced object contents


MikeSpreitzer:
### Feature Description

A downsynced object has three kinds of contents:
1. contents that propagate from WDS to WEC;
2. contents that propagate from WEC to WDS; and
3. contents that do not propagate in either direction.

For contents that propagate from WDS to WEC, category 1 divides into two interesting subcategories:
1.a. The value in the WEC should continually track the value in the WDS; or
1.b. The value in the WDS is only to be used as the initial value in the WEC.

Currently:
- The division of an object's contents is not configurable, it is hard-coded in KubeStellar code.
- KubeStellar equates an object's `status` section with category 2.
- KubeStellar considers category 3 to consist only of fields managed by the kube-apiserver.
- KubeStellar does not consider anything to be in category 1.b. (legacy had a way to say "treat everything that would normally be 1.a as 1.b" --- but that is too coarse in some cases).
- The full truth of the matter is that users need to be able to configure what parts of an object fall in what category. The current behavior covers the common cases, but not all cases. In some cases, the current treatment of status, labels, annotations, finalizers, and other content is not what users want.

I think we should consider this an extension of the schema for a kind of object. I do not think this is something that we should support configuring on a different basis.

### Proposed Solution

Define a kind of API object that can be put in a WDS to express this added aspect of schema. It would be optional, only needed when the default categorization is not desired. The cardinality would be: 0 or 1 of these per resource (roughly, kind of object).

### Want to contribute?

- [ ] I would like to work on this issue.

ezrasilvera:
@MikeSpreitzer I agree with the topic. I would like first to validate that we want/plan to support all the modes you described. For example, I'm not sure we should support 1.b  (the one time deploy) . I think that some of those cases were an outcome of work arounds we did in the past.

MikeSpreitzer:
Issue #2336 takes a bite at a piece of this.

MikeSpreitzer:
Also: the only use cases that I know about for 1.b. are as a work-around to a system not offering any control over the categorization of object contents.

MikeSpreitzer:
I have realized that the characterization in the opening comment of this issue is not quite right. The question is not exactly what propagates in a given direction but rather what is the authority over a given part of an object's state. So the three categories are really the following.

1. The parts of the object for which the state from the WDS is the authority.
2. The parts of the object for which the state from the WEC is the authority.
3. The parts of the object for which neither side has authority over the other.

[...]

PR #2417, opened 2024-08-30
✨ Do not update statuses for namespaces in WDS

waltforme:
## Summary
Getting hands on with the authority issue of downsynced object content #1851.

### DIFF
--- a/pkg/status/workstatus.go
+++ b/pkg/status/workstatus.go
@@ -90,6 +90,11 @@ func updateObjectStatus(ctx context.Context, objectIdentifier util.ObjectIdentif
 	listers util.ConcurrentMap[schema.GroupVersionResource, cache.GenericLister], wdsDynClient dynamic.Interface) error {
 	logger := klog.FromContext(ctx)
 
+	if util.WEC2WDSExceptions.Has(objectIdentifier.GVK) {
+		logger.V(4).Info("Status from WEC shouldn't have authority to overwrite status in WDS", "object", objectIdentifier)
+		return nil
+	}
+
 	gvr := objectIdentifier.GVR()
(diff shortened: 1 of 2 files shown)


pkg/util/authority.go
```diff
@@ -0,0 +1,26 @@
...
+*/
+
+package util
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+var WEC2WDSExceptions = sets.New(
```
MikeSpreitzer, line 24: (A) I would leave version out of it, this does not vary from version to version.

(B) I would include CustomResourceDefinition.

pkg/status/workstatus.go
```diff
@@ -90,6 +90,11 @@ func updateObjectStatus(ctx context.Context, objectIdentifier util.ObjectIdentif
 	listers util.ConcurrentMap[schema.GroupVersionResource, cache.GenericLister], wdsDynClient dynamic.Interface) error {
 	logger := klog.FromContext(ctx)
 
+	if util.WEC2WDSExceptions.Has(objectIdentifier.GVK) {
+		logger.V(4).Info("Status from WEC shouldn't have authority to overwrite status in WDS", "object", objectIdentifier)
```
MikeSpreitzer, line 94: Since this error comes from the user, it should be reported to the user. OK if that is done in follow-on work.

MikeSpreitzer, line 94: This is a user error and should be reported to the user. OK if that is done in a later PR.

MikeSpreitzer (approved):
This is a start. We can do further work in follow-on PRs.
/lgtm
/approve
Thanks!

PR #2419, opened 2024-08-31
🐛 Fix concurrency bugs, stop copying binding resolutions

### LINKED ISSUE #2418: bug: Concurrency bugs in binding controller
### Describe the bug

https://github.com/kubestellar/kubestellar/blob/51e221c3f8ae167c1d7ae35c91947346bd9559c5/pkg/binding/bindingpolicy-resolution.go#L34-L35 says that the mutex must be held while accessing fields. But the mutex is not held at the following accesses.

- https://github.com/kubestellar/kubestellar/blob/51e221c3f8ae167c1d7ae35c91947346bd9559c5/pkg/binding/bindingpolicy-resolver.go#L212
- https://github.com/kubestellar/kubestellar/blob/51e221c3f8ae167c1d7ae35c91947346bd9559c5/pkg/binding/bindingpolicy-resolver.go#L322
- https://github.com/kubestellar/kubestellar/blob/51e221c3f8ae167c1d7ae35c91947346bd9559c5/pkg/binding/bindingpolicy-resolution-broker.go#L124
- https://github.com/kubestellar/kubestellar/blob/51e221c3f8ae167c1d7ae35c91947346bd9559c5/pkg/binding/bindingpolicy-resolver.go#L362

### Steps To Reproduce

This is from code examination.

### Expected Behavior

Locking discipline is followed.

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR changes, primarily the binding controller, to remove the need for `ResolutionBroker.GetResolution` to copy data. This PR also removes some concurrency bugs in the handling of binding resolutions.

## Related issue(s)

Fixes #2418


pkg/abstract/map.go
```diff
@@ -0,0 +1,48 @@
...
+*/
+
+package abstract
+
+// Map abstracts reading from a map from Key to Val.
+// The Key values must be comparable in some way
+// (which is not exposed in this interface).
+// The map may or may not vary over time.
+// Access through this interface may or may not be thread-safe.
+type Map[Key, Val any] interface {
```
vMaroon, line 24: This makes util.ConcurrentMap redundant.

vMaroon: Do we have an independent need (aside for replacing ConcurrentMap) for this interface? Or is its sole purpose to serve MapLocker?

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -36,36 +37,62 @@ import (
...
 	// If true, it means that the object should only be created and not
 	// maintained
 	CreateOnly bool
-	// StatusCollectors is a set of status collector names that are associated
+	// StatusCollectors is a set of status collector names that are associated.
+	// Every set ever stored here is immutable.
 	StatusCollectors sets.Set[string]
 }
 
+var _ Resolution = &bindingPolicyResolution{}
```
vMaroon, line 75: What is this for?

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -36,36 +37,62 @@ import (
...
+func (resolution *bindingPolicyResolution) GetDestinations() sets.Set[string] {
+	resolution.RLock()
+	defer resolution.RUnlock()
+	return resolution.destinations
+}
+
+func (resolution *bindingPolicyResolution) GetWorkload() abstract.Map[util.ObjectIdentifier, ObjectData] {
+	m1 := abstract.AsPrimitiveMap(resolution.objectIdentifierToData)
+	m2 := abstract.NewMapLocker(&resolution.RWMutex, m1)
+	m3 := abstract.MapMapValues(m2, NonNilPointerDeference)
```
vMaroon, line 92: These three lines seem very similar to functionalities in pkg/primitive-maps.go. It would be awkward to keep both around.

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -227,12 +252,26 @@ func (resolution *bindingPolicyResolution) getOwnerReference() metav1.OwnerRefer
 	return *resolution.ownerReference
 }
 
-// getSingletonReportedStateReturnRequest returns what this resolution requests regarding singleton reported state return.
+func (resolution *bindingPolicyResolution) setRequestsSingletonReportedStateReturn(request bool) {
+	resolution.RLock()
```
vMaroon, line 256: this should be a write lock.

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -95,6 +98,7 @@ type BindingPolicyResolver interface {
 	// The given destinations set is expected not to be mutated during and
 	// after this call by the caller.
 	// If no resolution is associated with the given key, an error is returned.
+	// Must not be called concurrently with any method that adds resolutions.
```
vMaroon, line 101: A. Why do we need this constraint
B. Can it really be enforced? What if two different BindingPolicies are processed concurrently?

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -235,6 +234,11 @@ func (resolver *bindingPolicyResolver) EnsureObjectData(bindingPolicyKey string,
 			bindingPolicyKey)
 	}
 
+	// Now the resolver's mutex is not held, so the resolution just fetched could be concurrently
```
vMaroon, line 237: `Now` makes sense to us code maintainers since things are changing, but to external readers it wont.

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -252,6 +256,11 @@ func (resolver *bindingPolicyResolver) RemoveObjectIdentifier(bindingPolicyKey s
 	if bindingPolicyResolution == nil {
 		return false
 	}
+	// The resolver's mutex is no longer held by this goroutine, so the resolution
```
vMaroon, line 259: Same as the previous comment.

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -275,15 +284,12 @@ func (resolver *bindingPolicyResolver) GetObjectIdentifiers(bindingPolicyKey str
...
-// SetDestinations updates the maintained bindingpolicy's
-// destinations resolution for the given bindingpolicy key.
-// The given destinations set is expected not to be mutated during and
-// after this call by the caller.
-// If no resolution is associated with the given key, an error is returned.
 func (resolver *bindingPolicyResolver) SetDestinations(bindingPolicyKey string,
 	destinations sets.Set[string]) error {
 	bindingPolicyResolution := resolver.getResolution(bindingPolicyKey) // thread-safe
-
+	// Now the resolver's mutex is not held, so the resolution just fetched could be removed.
```
vMaroon, line 290: Same as the previous comments.

pkg/abstract/map-lock.go
```diff
@@ -0,0 +1,59 @@
...
+
+// MapLocker wraps locking around an inner Map.
+// The outer Map does not allow a consumer of Iterate2 to
+// access the outer map.
+type MapLocker[Key, Val any] struct {
+	mutex *sync.RWMutex // never nil
+	inner Map[Key, Val]
+}
+
+var _ Map[func() int, func() bool] = &MapLocker[func() int, func() bool]{}
```
vMaroon, line 31: A comment on compile-time interface compliance check may add clarity.

pkg/binding/bindingpolicy-resolution.go
vMaroon, line 75: A comment on compile-time interface compliance check may add clarity.

pkg/abstract/map.go
MikeSpreitzer: This interface only provides read access, not write access. I have found it useful in the past (e.g., in the legacy system - https://github.com/kubestellar/kubestellar/tree/v0.14.2/pkg/placement) to separate interfaces for reading and writing. Further, I have also found it useful to define interfaces for fragments of reading functionality.

The current (in this PR) three-layered implementation of `(*bindingPolicyResolution).GetWorkload` uses three layers that I have found to be useful independently and in different combinations in the past. `util.ConcurrentMap` is partly redundant with the combination of two of them. Defining the `Map` interface makes it possible to have various layers and combine them as needed for a given use.

pkg/binding/bindingpolicy-resolution.go
MikeSpreitzer: I assume you mean the file in this package that this PR renames to `primitive-map-ops.go`. Yes, there is partial overlap. Yes, I think that overlap should be addressed in follow-on PRs.

pkg/abstract/map-lock.go
MikeSpreitzer: Done.

pkg/binding/bindingpolicy-resolution.go
MikeSpreitzer: Done

pkg/binding/bindingpolicy-resolver.go
MikeSpreitzer: I expect this method's implementation will only be read by code maintainers.

pkg/binding/bindingpolicy-resolution.go
MikeSpreitzer: Oops, right. Thanks! Fixed.

pkg/binding/bindingpolicy-resolver.go
MikeSpreitzer: Sorry, I had not written the comment precisely enough. I have now revised it, to clarify that the concern is only with concurrent activity wrt the same BindingPolicy name.

The value it brings is making the implementation correct. Without that constraint or something else about usage, the implementation has a concurrency bug: the implementation fetches a resolution out of the resolver's map and then releases the lock on that map before making any change in the resolution; this opens up the possibility that concurrent activity migh remove that resolution and introduce a replacement (a resolution under the same key).

vMaroon: I originally thought that the concurrency problem is not conforming to the constraint of not directly accessing fields within the returned resolutions.

If a mutating call such as `SetDestinations` gets to fetching the resolution, then before executing the mutating method a reconciliation of the same key swapped the resolution in memory, then the mutating call would be irrelevant to the new resolution. It would be executed, it may alert subscribers of change, but once the change is reconciled, the state remains correct (redundant consumption at worst).

A simple solution to this problem may be adding a "valid" field and performing the double-checked locking pattern.

MikeSpreitzer: I am looking at multiple concurrency issues.

I am not sure that I would use the word "enforce". The current code _obeys_ this constraint.

vMaroon: How does it obey it? Also, do you agree with what I said in the 2nd paragraph?

MikeSpreitzer: Both of your questions have the same answer. There never are two concurrent invocations of `BindingPolicyResolver.SetDestinations` for the same BindingPolicy name because it is only called from `Controller.syncBindingPolicy` and, because of the way the controller pattern works, there are never two concurrent calls on that with the same BindingPolicy name.

vMaroon: I don't mean two concurrent calls to the same method specifically. What I meant is the only concurrency bug in the code is instances of accessing a field directly within a returned resolution. Instead, all should be accessed by getters and setters.

The issue of working with outdated binding-resolutions in case one is fetched and then replaced in memory when the lock is released, is not a bug, but a choice. Its price is redundant actions on the orphaned resolution and possibly a redundant sync check. I took this option because I thought it is statistically worth it.

MikeSpreitzer: In this PR, I am addressing more than the concurrency bugs called out in issue #2418 . Looking at the current definition of BindingPolicyResolver (https://github.com/kubestellar/kubestellar/blob/f30a85da33d6b9df8f9629ff27ffce74326b4e56/pkg/binding/bindingpolicy-resolver.go#L33-L130), I noticed that this interface does not forbid concurrent updates to the same BindingPolicy. In particular, it does not forbid concurrent updates that conflict. Sure, fixing the bugs called out in #2418 will prevent the code from panicing or failing a mechanical data race checker, but there is still the semantic question of what happens if two concurrent goroutines make conflicting calls on a BindingPolicyResolver. So I tried adding comments explaining why the code (all of the code in the repo taken together) is correct. I am not happy with what is in this PR, because some of the comments are inside method implementations yet take into account what the only client of this abstraction does with it --- thus breaking the abstraction. I hope to do better in follow-on PRs.

PR #2422, opened 2024-09-01
🌱 Stop messing with singletonstatus label

### LINKED ISSUE #2421: bug: Unnecessary and incoherent handling of singletonstatus label
### Describe the bug

The label is maintained on workload objects (`func (c *Controller) handleSingletonLabel` is called on workload objects) but the code that reads the label (https://github.com/kubestellar/kubestellar/blob/51e221c3f8ae167c1d7ae35c91947346bd9559c5/pkg/status/controller.go#L336-L362) is looking at WorkStatus objects.

The fact that this code works anyway demonstrates that the label is not needed.

### Steps To Reproduce

This is from code inspection.

### Expected Behavior

No unnecessary and broken writes to workload objects.

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR removes the code that does stuff with the label whose name (key) is defined at https://github.com/kubestellar/kubestellar/blob/51e221c3f8ae167c1d7ae35c91947346bd9559c5/pkg/util/labels.go#L31-L32 . 

## Related issue(s)

Fixes #2421

### DIFF
--- a/pkg/binding/selectors.go
+++ b/pkg/binding/selectors.go
@@ -23,9 +23,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


MikeSpreitzer:
This was mis-informed; see discovery at https://github.com/kubestellar/kubestellar/issues/2421#issuecomment-2323684202

MikeSpreitzer:
See #2423 instead.

PR #2423, opened 2024-09-02
🌱 Stop using singletonstatus label

### LINKED ISSUE #2420: bug: event labels handling bugs in status controller
### Describe the bug

(A) https://github.com/kubestellar/kubestellar/blob/51e221c3f8ae167c1d7ae35c91947346bd9559c5/pkg/status/controller.go#L353 may modify an API object passed to an informer event handler, but these should not modify the given objects.

(B) https://github.com/kubestellar/kubestellar/blob/51e221c3f8ae167c1d7ae35c91947346bd9559c5/pkg/status/controller.go#L350-L351 introduces a map that is not used for anything. Sure, it gets written into at line 353, but then it is dropped.

### Steps To Reproduce

This is from code inspection.

### Expected Behavior

Informer event handlers should not modify the API objects.

### Want to contribute?

### Additional Context

_No response_

### LINKED ISSUE #2421: bug: Unnecessary and incoherent handling of singletonstatus label

### LINKED ISSUE #2505: bug: singleton reported state return does not play nice with create-only
### Describe the bug

For a workload object marked create-only, singleton reported state return does not work. That is because the singleton reported state return mechanism involves a label being applied to the workload object in the WDS and this label propagating to the workload object in the WEC (and later, from there to the WorkStatus in the ITS). With create-only engaged, the label does not propagate to the workload object in the WEC if, for any reason, the object was already present, without the label, in the WEC.

### Steps To Reproduce

Outlined above.

### Expected Behavior

Singleton reported state return should work even for an object marked create-only.

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR updates the KubeStellar controller-manager to do nothing with labels named `managed-by.kubestellar.io/singletonstatus`. This is good because it is not necessary to run controller internal state through API object labels.

This PR also enhances some debug logging and adds a missing FieldManager attribution.

## Related issue(s)

Fixes #2420 
Fixes #2421 
Fixes #2505

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -24,10 +24,12 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/spf13/pflag"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -141,8 +143,10 @@ func main() {
 	setupLog.Info("Got config for ITS", "name", itsName)
 	itsRestConfig = itsClientLimits.LimitConfig(itsRestConfig)
 
-	// start the binding controller
-	bindingController, err := binding.NewController(logger, wdsClientMetrics, itsClientMetrics, wdsRestConfig, itsRestConfig, wdsName, allowedGroupsSet)
+	workloadEventRelay := &workloadEventRelay{}
+
+	// create the binding controller
+	bindingController, err := binding.NewController(logger, wdsClientMetrics, itsClientMetrics, wdsRestConfig, itsRestConfig, wdsName, allowedGroupsSet, workloadEventRelay)
 	if err != nil {
 		setupLog.Error(err, "unable to create binding controller")
 		os.Exit(1)
@@ -158,9 +162,41 @@ func main() {
 		os.Exit(1)
 	}
 
+	startBindingController := len(ctlrsToStart) == 0 || ctlrsToStart.Has(strings.ToLower(binding.ControllerName))
+	startStatusCtlr := len(ctlrsToStart) == 0 || ctlrsToStart.Has(strings.ToLower(status.ControllerName))
+	var statusController *status.Controller
+
+	if startStatusCtlr {
+		if !startBindingController {
+			setupLog.Error(nil, "Status controller does not work without binding controller")
+			os.Exit(1)
+		}
+		// check if status add-on present before starting the status controller
+		for i := 1; true; i++ {
+			if util.CheckWorkStatusPresence(itsRestConfig) {
+				break
+			}
+			if (i & (i - 1)) == 0 {
+				setupLog.Info("Not statrting status controller yet because WorkStatus is not defined in the ITS")
+			}
+			i++
+			time.Sleep(15 * time.Second)
+		}
+		setupLog.Info("Creating controller", "name", status.ControllerName)
+		statusController, err = status.NewController(logger, wdsClientMetrics, itsClientMetrics, wdsRestConfig, itsRestConfig, wdsName,
+			bindingController.GetBindingPolicyResolver())
+		if err != nil {
+			setupLog.Error(err, "unable to create status controller")
+			os.Exit(1)
+		}
+		workloadEventRelay.statusController = statusController
+	} else {
+		setupLog.Info("Not creating status controller")
+	}
+
 	cListers := make(chan interface{}, 1)
 
-	if len(ctlrsToStart) == 0 || ctlrsToStart.Has(strings.ToLower(binding.ControllerName)) {
+	if startBindingController {
 		setupLog.Info("Starting controller", "name", binding.ControllerName)
 		if err := bindingController.Start(ctx, workers, cListers); err != nil {
 			setupLog.Error(err, "error starting the binding controller")
@@ -168,17 +204,8 @@ func main() {
 		}
 	}
 
-	// check if status add-on present before starting the status controller
-	if util.CheckWorkStatusPresence(itsRestConfig) &&
-		(len(ctlrsToStart) == 0 || ctlrsToStart.Has(strings.ToLower(status.ControllerName))) {
+	if startStatusCtlr {
 		setupLog.Info("Starting controller", "name", status.ControllerName)
-		statusController, err := status.NewController(logger, wdsClientMetrics, itsClientMetrics, wdsRestConfig, itsRestConfig, wdsName,
-			bindingController.GetBindingPolicyResolver())
-		if err != nil {
-			setupLog.Error(err, "unable to create status controller")
-			os.Exit(1)
-		}
-
 		if err := statusController.Start(ctx, workers, cListers); err != nil {
 			setupLog.Error(err, "error starting the status controller")
 			os.Exit(1)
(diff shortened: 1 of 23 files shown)


pkg/status/workstatus.go
```diff
@@ -50,6 +50,10 @@ func (ws *workStatus) Content() map[string]interface{} {
 func (c *Controller) syncWorkStatus(ctx context.Context, ref workStatusRef) error {
 	logger := klog.FromContext(ctx)
 
+	if err := c.reconcileSingletonByWS(ctx, ref); err != nil {
```
vMaroon, line 53: For every workstatus whos object does not require singleton status, this will run
```
if err := c.reconcileSingletonWObj(ctx, wObjID, false); err != nil {
	return err
}
logger.V(4).Info("Cleaned singleton status for workload object",
			"gvk", ref.SourceObjectIdentifier.GVK, "objectName", ref.SourceObjectIdentifier.ObjectName,
			"requested", requested, "nWECs", nWECs)
return nil
```

MikeSpreitzer: Yes. I also anticipate doing point 7 in https://github.com/kubestellar/kubestellar/issues/1851#issuecomment-2320123956 .

MikeSpreitzer: In fact, perhaps like you, I am nervous about merging this change without that additional change.

MikeSpreitzer:
/remove hold
I added a commit that suppresses the unnecessary writes to workload object `.status`.

MikeSpreitzer:
Actually, this is still not good enough. It will insist on completely empty `.status` for every potential workload object in the WDS.
/hold

MikeSpreitzer:
Latest revision is more restrained. It uses a label to remember objects that status was written into, but does not run the label through the WEC; the label is only on the workload object.

The latest revision also introduces workqueue items in the status controller for working on workload objects, to fix long-standing bugs with concurrent processing of the same object.

/remove-hold

MikeSpreitzer:
The current revision has some `klog.Info...` for debugging, will be removed.

cmd/controller-manager/main.go
```diff
@@ -158,27 +162,50 @@ func main() {
...
+			setupLog.Error(nil, "Status controller does not work without binding controller")
+			os.Exit(1)
+		}
+		// check if status add-on present before starting the status controller
+		for i := 1; true; i++ {
+			if util.CheckWorkStatusPresence(itsRestConfig) {
+				break
+			}
+			if (i & (i - 1)) == 0 {
+				setupLog.Info("Not statrting status controller yet because WorkStatus is not defined in the ITS")
```
vMaroon, line 180: s/statrting/starting/

cmd/controller-manager/main.go
```diff
@@ -187,3 +214,13 @@ func main() {
 
 	select {}
 }
+
+type workloadEventRelay struct {
+	statusController *status.Controller
+}
+
+func (wer *workloadEventRelay) HandleWorkloadObjectEvent(gvr schema.GroupVersionResource, oldObj, obj util.MRObject, eventType binding.WorkloadEventType, wasDeletedFinalStateUnknown bool) {
```
vMaroon, line 222: Missing documentation on exported function.

pkg/abstract/map-map.go
```diff
@@ -46,6 +46,13 @@ func (wrapper MapValueMapper[Key, Val, Mapped]) Get(key Key) (Mapped, bool) {
 	return wrapper.mapVal(val), true
 }
 
+func (wrapper MapValueMapper[Key, Val, Mapped]) ContGet(key Key, cont func(Mapped)) {
```
vMaroon, line 49: The name is not clear (to me). Missing documentation on exported (non-obvious, to me) function.

pkg/abstract/map-to-comparable.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+
+var _ MutableMapToComparable[string, int] = &IndexedMapToComparable[string, int]{}
+
+func (imc *IndexedMapToComparable[Key, Val]) Length() int {
+	return imc.forward.Length()
+}
+
+func (imc *IndexedMapToComparable[Key, Val]) ContGet(key Key, cont func(Val)) {
```
vMaroon, line 51: Same as above.

pkg/abstract/primitive-map-as-map.go
```diff
@@ -34,6 +34,13 @@ func (pm PrimitiveMap[Key, Val]) Get(key Key) (Val, bool) {
 	return val, have
 }
 
+func (pm PrimitiveMap[Key, Val]) ContGet(key Key, cont func(Val)) {
```
vMaroon, line 37: Same as above.

pkg/binding/binding.go
```diff
@@ -38,7 +38,7 @@ func (c *Controller) syncBinding(ctx context.Context, bindingName string) error
 		// then the bindingpolicy has been deleted, and the binding
 		// will eventually be garbage collected. We can safely ignore this.
 
-		c.bindingPolicyResolver.Broker().NotifyCallbacks(bindingName)
+		c.bindingPolicyResolver.Broker().NotifyBindingPolicyCallbacks(bindingName)
```
vMaroon, line 41: Who notifies singleton-status subscribers if this possibly newly recorded absence affects any?

vMaroon:
Looks good overall, binding and status controller docs should be updated to reflect the new logic and responsibilities. Can be done in a followup PR.

cmd/controller-manager/main.go
MikeSpreitzer: Fixed.

cmd/controller-manager/main.go
MikeSpreitzer: I updated the source around here to make it clear that this is implementing `binding.WorkloadEventHandler`.

I do not favor just copying comments from interface to implementation, that does not add information. I have added some comments on the interface and the implementation type, hopefully it is clearer now.

pkg/abstract/map-map.go
MikeSpreitzer: This is an implementation of a method of the `MapToLocked` interface, which has an extensive comment explaining this method and its name. This method is a continuation-passing-style "get".

pkg/binding/binding-controller.go
```diff
@@ -64,6 +65,18 @@ const (
 	defaultResyncPeriod = time.Duration(0)
 )
 
+type WorkloadEventHandler interface {
```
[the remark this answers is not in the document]
MikeSpreitzer: I added comments on this type and its method.

The idea here is not that the binding controller is given the status controller. The idea is that, like a broker, `WorkloadEventHandler` is an abstraction whose user does not care about its implementation. Passing this thing at binding controller construction time avoids the question of what would happen if such a handler were added after the binding controller had already started its informers and gotten some notifications. If we did things the latter way, the status controller would need the answer to be that the handler is called with all the notifications that the binding controller has already received --- which would be expensive to implement.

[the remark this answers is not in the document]
vMaroon: I understand but it's sort of "hacky" given that what's given at construction time is empty, and what really happens is that it's set before starting. Can implement the exact same behavior in a more formal broker-like approach. A matter of preference, not a blocker.

vMaroon:
@MikeSpreitzer LGTM'd, binding/status controller docs should be updated in a follow up, thank you for the cleanup & improvements.

MikeSpreitzer:
Yes, I plan follow-up work on doc and maybe the workload observer as suggested.

Issue #1954, opened 2024-03-22
bug: KS v0.21 latency for remote pod execution degraded relative to v0.20 for Braulio stress test


eaepstein:
### Describe the bug

Braulio's stress test creates pods in a WDS at a specified rate. The pods run "bash -c sleep 20" on the ubuntu:latest image and then terminate. In my test KS was deployed on a VM (32GB RAM, 8vCPU) using K3D clusters for hub and spoke. 

The two plots below show individual pod end-to-end latencies from creation time in WDS to status=complete in WDS. Pods were created 1 every 3 seconds in the same namespace. A single binding policy was used. There were no pod deletions during each test.

![k3d-ks020-end-to-end-onethird](https://github.com/kubestellar/kubestellar/assets/56738503/ad34f71f-8df9-4b14-9c53-82889e414a73)

![k3d-ks021-end-to-end-onethird](https://github.com/kubestellar/kubestellar/assets/56738503/5435c686-d8f4-4f0c-8ba7-8e202a6de607)

### Steps To Reproduce

A PR adding a document describing how to deploy KS on K3D should be available shortly.
Braulio will document how to run his stress test. Yaml files for the pod and and the binding policy are included here.

[yamls.zip](https://github.com/kubestellar/kubestellar/files/14727649/yamls.zip)

### Expected Behavior

KS overhead latency should not be so badly degraded for this test.

nirrozenbaum:
@eaepstein @dumb0002 thanks for opening this issue, indeed an important one that worth deeper investigation.
would it be possible to add more information about where do we spend the time?
it would be very helpful to understand what exactly takes time to identify where is the problem.
some examples - 
- does the distribution from WDS to ITS take more time?
- does OCM distribution from ITS to WECs take more time? in this case we might hit OCM bug (see issue #1868)
- it could be that distribution works perfectly fine but the issue is in the status?
- etc...

ezrasilvera:
@nirrozenbaum I don't think it's an issue of "where we spend the time" , the main problem I see is a (linear) constant increase in time as the number of resources  increases.  Which means this is a major scalability issue.

ezrasilvera:
@eaepstein @nirrozenbaum One test that can be helpful is:
1. Measure latency of creation
2. Populate the system with large number of objects
3. Wait for few minutes 
4. Repeat (1) - Measure latency of creation 
[...]

PR #2431, opened 2024-09-04
🌱 Improve default max-num-wrapped of transport controller

MikeSpreitzer:
## Summary
The PR changes the default value for the `--max-num-wrapped` command line flag of the generic transport controller. The old value was 9223372036854775807, a huge value that stuns the eyes does not correspond to anything reasonable. The new value is the same as the default for `--max-size-wrapped`, which is a much more sensible value. Each wrapped object will certainly require multiple bytes.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/transport/generic/cmd/options.go
+++ b/pkg/transport/generic/cmd/options.go
@@ -17,8 +17,6 @@ limitations under the License.
 package cmd
 
 import (
-	"math"
-
 	"github.com/spf13/pflag"
 
 	ksopts "github.com/kubestellar/kubestellar/options"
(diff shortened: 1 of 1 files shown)


eaepstein (commented):
Given the serious performance degradation reported in https://github.com/kubestellar/kubestellar/issues/1954, setting MaxNumWrapped to 500k is no less ridiculous than math.MaxInt,

MikeSpreitzer:
@eaepstein: can we please fight the changing default battle separately? I just want to replace a crazy large number with one that is not overwhelming.

francostellari:
@MikeSpreitzer what is the numeric value of `maxSizeWrapped`? I could not easily find it in the code change

MikeSpreitzer:
@francostellari: it is unchanged, at 51200. It is expressed in the code as `500 * 1024`. I lifted that expression to an earlier place so that the value could be used for both flags.

vMaroon (approved):
/lgtm

PR #2441, opened 2024-09-06
🐛 Update goreleaser to track Makefile change

### LINKED ISSUE #2440: bug: wrong image tag in controller-manager chart
### Describe the bug

When using release `0.24.0-experiment.5`, the command `test/e2e/run-test.sh --released --kubestellar-controller-manager-verbosity 2` produces a broken system. Also, the "Getting Started" recipe is also broken. Both use the same broken Helm chart (https://github.com/kubestellar/kubestellar/pkgs/container/kubestellar%2Fcontroller-manager-chart/269324623?tag=0.24.0-experiment.5).

`kubectl --context kind-kubeflex get deploy -n wds1-system kubestellar-controller-manager -o yaml` reveals the following in the controller-manager container config:

```yaml
        image: kubestellar/kubestellar/controller-manager:latest
```

### Steps To Reproduce

Shown above.

### Expected Behavior

The end-to-end test of the latest release succeeds.

### Want to contribute?

### Additional Context

The problem is that the Makefile was changed regarding tagging of that image, but the goreleaser workflow was not updated to track that change.

This is particularly tricky to debug because the goreleaser workflow sets relevant environment variables that affect the Makefile behavior when it is invoked in the incorrect way that the workflow does, but doing the same `make` invocation from a plain shell gets different behavior.

MikeSpreitzer:
## Summary
This PR updates the goreleaser workflow to track a recent change to the Makefile regarding the variable that defines the kubestellar-controller-manager image reference.

## Related issue(s)

Fixes #2440

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -62,7 +62,7 @@ jobs:
     - name: Package and push charts
       run: |
         chartVersion=$(echo ${{ github.ref_name }} | cut -c 2-)
-        make chart IMG=${{ env.REGISTRY }}/${{ env.CONTROLLER_MANAGER_IMAGE }}:${{github.ref_name}}
+        make chart CONTROLLER_MANAGER_IMAGE=${{ env.REGISTRY }}/${{ env.CONTROLLER_MANAGER_IMAGE }}:${{github.ref_name}}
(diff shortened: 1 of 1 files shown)


nirrozenbaum (approved):
/lgtm
/approve

PR #2446, opened 2024-09-06
🐛 Update the kubeflex version in the GitHub workflows

MikeSpreitzer:
## Summary
This PR updates the GitHub Actions workflows to use release 0.6.1 of kubeflex rather than 0.4.2, because 0.6.1 is now the minimum required version.

## Related issue(s)

Fixes #


MikeSpreitzer:
Successfully tested the case that shows the problem, see https://github.com/MikeSpreitzer/kcp-edge-mc/actions/runs/10742260653 .

PR #2450, opened 2024-09-11
🐛 Correct aggregation of zero rows

MikeSpreitzer:
## Summary
This PR fixes the implementation of status combination to correctly handle the case of aggregation over zero rows.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/status/combinedstatus-resolver.go
+++ b/pkg/status/combinedstatus-resolver.go
@@ -87,7 +87,7 @@ type CombinedStatusResolver interface {
 	//
 	// The returned set contains the identifiers of combinedstatus objects
 	// that should be queued for syncing.
-	NoteWorkStatus(workStatus *workStatus) sets.Set[util.ObjectIdentifier]
+	NoteWorkStatus(ctx context.Context, workStatus *workStatus) sets.Set[util.ObjectIdentifier]
 
 	// ResolutionExists returns true if a combinedstatus resolution is
 	// associated with the given name. The name is expected to follow the
@@ -244,8 +244,8 @@ func (c *combinedStatusResolver) NoteBindingResolution(ctx context.Context, bind
 		if !exists {
 			logger.V(3).Info("Introducing CombinedStatus resolution", "binding", bindingName, "objectId", objectIdentifier)
 			csResolution = &combinedStatusResolution{
-				name:                      getCombinedStatusName(policyUID, objectData.UID),
-				statusCollectorNameToData: make(map[string]*statusCollectorData),
+				Name:                      getCombinedStatusName(policyUID, objectData.UID),
+				StatusCollectorNameToData: make(map[string]*statusCollectorData),
 			}
 			objectIdentifierToResolution[objectIdentifier] = csResolution
 			c.resolutionNameToKey[csResolution.getName()] = resolutionKey{bindingName, objectIdentifier}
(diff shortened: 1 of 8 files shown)


pkg/status/combinedstatus-resolver.go
```diff
@@ -261,11 +261,12 @@ func (c *combinedStatusResolver) NoteBindingResolution(ctx context.Context, bind
 		removedDestinations, newDestinationsSet := csResolution.setCollectionDestinations(destinationsSet)
 
 		logger.V(5).Info("Updating CombinedStatus resolution", "binding", bindingName, "objectId", objectIdentifier,
+			"introduced", !exists,
 			"removedCollectors", removedCollectors, "addedCollectors", addedCollectors,
 			"removedDestinations", removedDestinations, "newDestinationsSet", newDestinationsSet)
 
 		// should queue the combinedstatus object for syncing if lost collectors / destinations
-		if removedCollectors || removedDestinations {
+		if removedCollectors || removedDestinations || !exists {
```
vMaroon, line 269: If I'm not mistaken, this change may be misleading since a CombinedStatus is synced/created only if its state differs from that in memory. If a key is synced only from this flag, it may not necessarily have state that is different (not empty) - therefore only content update events queue up combined-statuses.

MikeSpreitzer: I think that there is a user-visible change in this case. The internal state change is from no resolution present to some resolution is present. There is a corresponding external state change, from the CombinedStatus object not existing to existing.

My point here is that a CombinedStatus object should exist for every (Binding, bound workload object) pair. Period. End of story. If there is no returned state from any WECs for that object at the moment, or there are some WorkStatus objects but each one of them is rejected by the the filter of every relevant StatusCollector, then the CombinedStatus should hold the results of aggregating over zero rows.

MikeSpreitzer:
The force-push to 380ca42bf is a rebase onto `main`.

pkg/status/combinedstatus-resolver.go
vMaroon: That's a reasonable decision. My point with "misleading" is that the change in this line does not implement your intent.

A non-existent combinedstatus in the cluster and an empty combinedstatus in memory are equivalent currently - unless I missed a change in a merged PR.

MikeSpreitzer: Following is how it goes down.

1. When the internal combined status resolution is introduced, this PR also causes the reference to the CombinedStatus to be put in the status controller's workqueue.
2. When the status controller pulls this reference to work on it, it finds that the CombinedStatus does not exist (now, yet) and creates, for usage later in the method, a CombinedStatus data structure with results for zero StatusCollectors.
3. The controller then calls `c.combinedStatusResolver.CompareCombinedStatus`, which fetches the now-extant resolution data structure and then calls `resolution.compareCombinedStatus`.
4. That method first calls `c.generateCombinedStatus` to create a copy of what should exist externally. In the case probed in the new ginkgo test case, that will have one NamedStatusCombination (for the one StatusCollector involved in the test case). Yet the CombinedStatus ginned up in step 2, there are zero NamedStatusCombinations. So the subsequent comparison of len(status.Results) will indeed find a difference.

BTW, if the comparison had proceeded to `statusCombinationEqual`, that would have also returned `false` in the new test case. Assuming that the NamedStatusCombination ginned up in step 2 has zero rows, the correct number of rows is 1 (the result of aggregating over the empty all-group).

vMaroon: Ah right, the PR does indeed introduce a change to the state. Thanks.

MikeSpreitzer: Ah, my statement of intent omitted a qualification --- that some StatusCollectors are referenced. So let me try again.

There should be a CombinedStatus object for every (Binding B, workload object O) pair where B includes O with a request to apply at least one StatusCollector.

MikeSpreitzer: (A) I think that the API (types.go) should be clear on this. I will leave that to a different PR.

(B) I think that the current code gets it wrong in one case: the CombinedStatus object exists and has zero NamedStatusCombination, and indeed there are no applicable StatusCollectors. I will also leave fixing this to a follow-on.

vMaroon: Sounds good

PR #2451, opened 2024-09-12
✨ Stop scheduling

MikeSpreitzer:
## Summary
This PR changes KubeStellar to not pick a single WEC when the user requests singleton reported state return.
This PR also changes the contract for how a wrong number of WECs is reported, and implements the new contract.

## Related issue(s)

Fixes #


pkg/binding/bindingpolicy-resolver.go
```diff
@@ -124,7 +124,7 @@ type BindingPolicyResolver interface {
 	// GetSingletonReportedStateRequestsForBinding calls GetSingletonReportedStateRequestForObject
 	// for each of workload objects in the resolution if the resolution exists.
 	// If the resolution doesn't exist, empty slices are returned.
```
waltforme, line 126: We may also change the comment accordingly with the code change, regarding the returned result (a single slice and could be nil).

pkg/binding/binding.go
```diff
@@ -51,35 +53,81 @@ func (c *Controller) syncBinding(ctx context.Context, bindingName string) error
...
+		if err != nil {
+			policyErrors = append(policyErrors, fmt.Sprintf("Failed to json.Marshal some example blighted objects (%#v): %s", badSR, err))
+		} else {
+			policyErrors = append(policyErrors, fmt.Sprintf("Singleton reported status return is requested but some objects have the wrong number of associated WECs, for example: %s", string(badSRBytes)))
+		}
+	}
+	policyWithStatus := policy.DeepCopy()
+	policyWithStatus.Status = v1alpha1.BindingPolicyStatus{
+		ObservedGeneration: policy.Generation,
+		Errors:             policyErrors,
```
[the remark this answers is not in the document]
MikeSpreitzer: This controller is the authority for the BindingPolicyStatus and computes all of the right errors every time this code runs. This is a state-based approach. The right list of errors comes from the current state.

pkg/binding/bindingpolicy-resolver.go
MikeSpreitzer: Thanks, updated.

MikeSpreitzer:
The force-push to 947b1442a updates the comment as noted above.

api/control/v1alpha1/types.go
```diff
@@ -105,30 +105,17 @@ type BindingPolicySpec struct {
...
 	Downsync []DownsyncPolicyClause `json:"downsync,omitempty"`
 
-	// WantSingletonReportedState means that for objects that are distributed --- taking
+	// WantSingletonReportedState means that for workload objects that are distributed --- taking
 	// all BindingPolicies into account --- to exactly one WEC, the object's reported state
 	// from the WEC should be written to the object in its WDS.
-	// WantSingletonReportedState connotes an expectation that indeed the object will
-	// propagate to exactly one WEC, but there is no guaranteed reaction when this
-	// expectation is not met.
+	// If any of the workload objects are distributed to more or less than 1 WEC then
```
ezrasilvera, line 111: Is the rest of the comment makes sense for 0 WECs as indicated in the "less" here ?  How ?

pkg/binding/binding.go
```diff
@@ -51,35 +53,81 @@ func (c *Controller) syncBinding(ctx context.Context, bindingName string) error
...
 
 		// notify the bindingpolicy resolution broker that the binding has been updated
 		c.bindingPolicyResolver.Broker().NotifyCallbacks(bindingPolicyIdentifier)
-		return nil
+	}
+	srPerObj := c.bindingPolicyResolver.GetSingletonReportedStateRequestsForBinding(bindingPolicyIdentifier)
+	policyErrors := append([]string{}, binding.Status.Errors...)
+	badSR := []objectWithNumWECs{}
+	for _, srStatus := range srPerObj {
+		if srStatus.WantSingletonReportedState && srStatus.NumWECs != 1 {
```
ezrasilvera, line 94: Similar to the pervious comment on the code comments - what do we think the user expects  to see in the case of 0 wecs ?

waltforme:
My comments were addressed so
/lgtm

We may finish the discussions on Ezra's comments then do the merge.

api/control/v1alpha1/types.go
MikeSpreitzer: If the user expects the object to be distributed to 1 WEC and the actual number is 0 then this is a discrepancy to report.

pkg/binding/binding.go
MikeSpreitzer: The singleton reported state return feature only makes sense to use when the number of associated WECs is exactly 1. When the number is 0, there will be no return of reported state into the workload object; but since the user requested such return, there is a surprise here that should be reported.

PR #2457, opened 2024-09-16
✨ Extend PropagationData API - implementation not included

MikeSpreitzer:
## Summary
This PR proposes an extension to the API for "combined status", making more `PropagationData` available.

This PR is only the API change, the implementation change is not included here. This PR is, at first, for discussion/consensus on the API change. Once that is agreed, we can proceed to the implementation. That will be based on information in the `.status` of the `ManifestWork` objects. Following is an example of one (for the nginx Deployment object produced by `test/e2e/run-test.sh`).

```yaml
status:
  conditions:
  - lastTransitionTime: "2024-09-12T16:37:37Z"
    message: Apply manifest work complete
    observedGeneration: 2
    reason: AppliedManifestWorkComplete
    status: "True"
    type: Applied
  - lastTransitionTime: "2024-09-12T16:37:37Z"
    message: All resources are available
    observedGeneration: 2
    reason: ResourcesAvailable
    status: "True"
    type: Available
  resourceStatus:
    manifests:
    - conditions:
      - lastTransitionTime: "2024-09-12T16:37:37Z"
        message: Apply manifest complete
        reason: AppliedManifestComplete
        status: "True"
        type: Applied
      - lastTransitionTime: "2024-09-12T16:37:37Z"
        message: Resource is available
        reason: ResourceAvailable
        status: "True"
        type: Available
      - lastTransitionTime: "2024-09-12T16:37:37Z"
        message: ""
        reason: NoStatusFeedbackSynced
        status: "True"
        type: StatusFeedbackSynced
      resourceMeta:
        group: ""
        kind: Namespace
        name: nginx
        namespace: ""
        ordinal: 0
        resource: namespaces
        version: v1
      statusFeedback: {}
    - conditions:
      - lastTransitionTime: "2024-09-12T16:37:37Z"
        message: Apply manifest complete
        reason: AppliedManifestComplete
        status: "True"
        type: Applied
      - lastTransitionTime: "2024-09-12T16:37:37Z"
        message: Resource is available
        reason: ResourceAvailable
        status: "True"
        type: Available
      - lastTransitionTime: "2024-09-12T16:37:37Z"
        message: ""
        reason: NoStatusFeedbackSynced
        status: "True"
        type: StatusFeedbackSynced
      resourceMeta:
        group: apps
        kind: Deployment
        name: nginx-deployment
        namespace: nginx
        ordinal: 1
        resource: deployments
        version: v1
      statusFeedback: {}
```

## Related issue(s)

Fixes #


api/control/v1alpha1/types.go
```diff
@@ -490,8 +490,24 @@ type ReturnedState struct {
...
+
+	// `lastGeneration` is that last `ObjectMeta.Generation` from the WDS that
+	// propagated to the WEC. This is not to imply that it was successfully applied there;
+	// for that, see `lastGenerationIsApplied`.
+	// Zero means that none has yet propagated there.
+	LastGeneration int64 `json:"lastGeneration"`
+
+	// `lastGenerationIsApplied` indicates whether `lastGeneration` has been successfully
+	// applied in the WEC. This is `false` while no attempt has yet been made.
+	LastGenerationIsApplied bool `json:"lastGenerationIsApplied"`
```
nirrozenbaum, line 504: what about the situation where an attempt has been made but was not successful?
does it considered as applied or not? 
this is not clear from the comment.

api/control/v1alpha1/types.go
```diff
@@ -490,8 +490,24 @@ type ReturnedState struct {
...
+
+	// `lastCurrencyUpdateTime` is the time of the latest update to either
+	// `lastGeneration` or `lastGenerationIsApplied`. More precisely, it is
+	// the time when the core became informed of the update.
+	// Before the first such update, this holds the zero value of `time.Time`.
+	LastCurrencyUpdateTime metav1.Time `json:"lastCurrencyUpdateTime"`
```
vMaroon, line 510: Perhaps due to not being a native speaker, this name does not tell me much. With this timestamp and the `LastGenerationIsApplied` bit one can infer whether this is a `LastGenerationUpdateTimestamp` or a `LastGenerationApplicationTimestamp`. The current name/field-documentation do not explicitly imply this usefulness.

api/control/v1alpha1/types.go
```diff
@@ -490,8 +490,24 @@ type ReturnedState struct {
...
 	LastReturnedUpdateTimestamp metav1.Time `json:"lastReturnedUpdateTimestamp"`
+
+	// `lastGeneration` is that last `ObjectMeta.Generation` from the WDS that
+	// propagated to the WEC. This is not to imply that it was successfully applied there;
+	// for that, see `lastGenerationIsApplied`.
+	// Zero means that none has yet propagated there.
+	LastGeneration int64 `json:"lastGeneration"`
+
+	// `lastGenerationIsApplied` indicates whether `lastGeneration` has been successfully
+	// applied in the WEC. This is `false` while no attempt has yet been made.
```
vMaroon, line 503: I think the 2nd part of the sentence is redundant and may introduce confusion. This bit should be `true` only if `LastGeneration` is applied. That's it. Its previous state or the reason it is/was switched to false is irrelevant, in my opinion.

MikeSpreitzer: To mind mind also, the second sentence is redundant. I included it for clarity --- I think that _some_ other people will think that the first sentence does not cover this case.

api/control/v1alpha1/types.go
MikeSpreitzer: Do you see the qualifier "successfully" in the first sentence of the comment?

api/control/v1alpha1/types.go
MikeSpreitzer: @vMaroon: I am not sure I understand your point. Did you omit a word, meaning to say "can not infer" rather than "can infer"? The second sentence of the comment is intended to clarify that the time value in this field is read from a clock in the core in every case. Do you find that unclear?

Indeed I see no reason to expect `LastGenerationIsApplied` to change from `true` to `false` without _also_ there being a change to `LastGeneration`. I have not attempted to optimize the API in light of that exclusion, nor am I sure that we _should_ build that exclusion into the API.

api/control/v1alpha1/types.go
nirrozenbaum: I did, the last sentence got me confused.

MikeSpreitzer:
The force-push to 95a2d11d2 deletes a sentence that reviewers found confusing.

api/control/v1alpha1/types.go
MikeSpreitzer: Since it confused you and @vMaroon, I have removed it.

api/control/v1alpha1/types.go
MikeSpreitzer: Since it confused @nirrozenbaum, I have removed it.

api/control/v1alpha1/types.go
vMaroon: The point is I have no idea what this name is attempting to say. If there are alternatives, I think they are preferable.

`LastGenerationIsApplied` bit sets the meaning of this timestamp. Perhaps this can be reflected through the name/documentation.

MikeSpreitzer: In every case this is a timestamp read from a clock in the core. More particularly, this is `time.Now()` from the status controller, executed near the time it updates either `lastGeneration` or `lastGenerationIsApplied`.

MikeSpreitzer:
Some implementation notes.

`PropagationData.LastGeneration` can be set to `ManifestWorkStatus.Conditions[ Type==Applied ].ObservedGeneration` if there exists an `i` such that `ManifestWorkStatus.ResourceStatus.Manifests[i].Conditions[ Type==Applied ].Status == "True"`.

`lastGenerationIsApplied` should be set to the value of `ManifestWorkStatus.ResourceStatus.Manifests[i].Conditions[ Type==Applied ].Status` for the `i` for the object in question.

`lastCurrencyUpdateTime` should be set to `time.Now()` whenever either of the above makes a change.

MikeSpreitzer:
There is an important wrinkle that I should have included in my implementation note above. The Status controller consumes WorkStatus objects but not ManifestWork objects. The latter are a private detail of the OCM-based transport controller. To achieve the outlined implementation requires extending the information in the WorkStatus objects or creating additional API objects.

wangd14:
> There is an important wrinkle that I should have included in my implementation note above. The Status controller consumes WorkStatus objects but not ManifestWork objects. The latter are a private detail of the OCM-based transport controller. To achieve the outlined implementation requires extending the information in the WorkStatus objects or creating additional API objects.

Extending the WorkStatus object seems simpler to me since its already in use. If we go with this method, is the update to the workStatus struct below suitable? If there is a way you think it might better be stored or this isn't what you were thinking about lmk.

```
 type workStatus struct {
	 	workStatusRef
	 	status         map[string]interface{}
	 	lastUpdateTime *metav1.Time
	 	lastGeneration int64
                LastGenerationIsApplied  bool
                lastCurrencyUpdateTime  *metav1.Time
 }
```

MikeSpreitzer:
By "WorkStatus" I meant the API object. Look at https://github.com/kubestellar/ocm-status-addon/blob/v0.2.0-rc13/api/v1alpha1/workstatus_types.go#L32

MikeSpreitzer:
The force-push to bcb081aab is a rebase onto `main` to resolve a merge conflict.

clubanderson:
This design PR has been open as WIP for 16 months. To keep the repository clean while preserving the discussion, this has been converted to a GitHub Discussion.

**Continue the conversation here:** https://github.com/kubestellar/kubestellar/discussions/3622

Thank you @MikeSpreitzer for the design work. The discussion can be revisited when there's renewed interest in extending the PropagationData API.

PR #2459, opened 2024-09-18
 ✨ Adding Poisson tunning set for workload generation

dumb0002:
## Summary

This PR adds the Poisson tuningset in the workload configuration for clusterloader2 - we will use the tuning set to generate the KS benchmark workloads in our experiments. 

N.B: this PR also improves the README to allow users to easily copy & past the list commands. 

## Related issue(s)

Fixes #

### DIFF
--- a/test/performance/workloads/performance-test-config.yaml
+++ b/test/performance/workloads/performance-test-config.yaml
@@ -5,7 +5,7 @@
 {{$K8S_CLUSTER := DefaultParam .K8S_CLUSTER true}}
 {{$OPENSHIFT_CLUSTER := DefaultParam .OPENSHIFT_CLUSTER false}}
 {{$CREATE_NS := DefaultParam .CREATE_NS true}}
-{{$tuningSet := DefaultParam .tuningSet "RandomizedLoad"}}
+{{$tuningSet := DefaultParam .tuningSet "PoissonLoad"}}
 
 
 name: load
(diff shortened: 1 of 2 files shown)


test/performance/workloads/performance-test-config.yaml
```diff
@@ -41,6 +41,12 @@ tuningSets:
   parallelismLimitedLoad:
     parallelismLimit: 100
 
+# PoissonLoad defines a load with with Poisson arrival time
```
ezrasilvera, line 44: Typo:  Remove one of the "with"

ezrasilvera, line 44: Arrival time of what ?  Maybe add ?

test/performance/workloads/performance-test-config.yaml
```diff
@@ -41,6 +41,12 @@ tuningSets:
...
+# for a given rate parameter λ (ExpectedActionsPerSecond)
```
ezrasilvera, line 45: I would change the order 
for a given rate parameter ExpectedActionsPerSecond (λ in the Poisson model)

test/performance/workloads/performance-test-config.yaml
dumb0002: Fixed! thanks

dumb0002: The Poisson arrival time is for the load itself.

test/performance/workloads/performance-test-config.yaml
dumb0002: Fixed! Thanks

PR #2460, opened 2024-09-18
🐛 Make WorkStatus event handlers check WDS name

### LINKED ISSUE #2455: bug: The status controller ignores the origin WDS
### Describe the bug

When the status controller runs, it never checks the origin WDS of workstatuses. A label key is defined at
https://github.com/kubestellar/kubestellar/blob/4f114959d8923fbcbb41e274bacdda4a59e13732/pkg/status/controller.go#L54 But the definition is not referenced anywhere in the repository.

Note that the same label key is defined at
https://github.com/kubestellar/kubestellar/blob/4f114959d8923fbcbb41e274bacdda4a59e13732/pkg/transport/generic/generic_transport_controller.go#L70 And the definition is only referenced in pkg/transport/generic/generic_transport_controller.go.

When I was doing experiments with kueue + KS, I had two WDSes --- `wds1` of type 'k8s' and `wds2` of type 'host'. In my setup, I saw the status controller for `wds2` tries to sync workstatuses for `wds1`.

This issue does not cause any problem when there is exactly one WDS using one ITS. I think it issue is hiding for so long because most of the time we had only one WDS. But as our architecture doc suggests, there could be multiple WDSes sharing one ITS.

### Steps To Reproduce

- In addition to `wds1` in our example setup, create a second WDS, say `wds2`.
- Start KS for `wds2` with verbosity level of at least 4 for the controller manager.
- Make sure there are
  - at least one Workstatus object(s) in `its1` that carry the reported state of workload object(s) in `wds1`.
  - _zero_ Workstatus objects in `its1` that carry the reported state of workload objects in `wds2`.
- Check the log of the controller manager of `wds2`, and observe something like
```
I0916 17:53:15.790008 3989659 singletonstatus.go:55] "Reconciling singleton status due to workstatus changes" logger="Status.worker-3" name="ce862a8f-9cc4-4d53-84c4-219228319d1b-appsv1-deployment-nginx-nginx-singleton"
I0916 17:53:15.790061 3989659 singletonstatus.go:89] "Reconciling workload object for singleton status" logger="Status.worker-3" gvk="apps/v1, Kind=Deployment" objectName="nginx/nginx-singleton" sync=false
I0916 17:53:15.790101 3989659 workstatus.go:114] "Did not update workload object because it is not in local cache, presumably because it was recently deleted" logger="Status.worker-3" objectIdentifier={"GVK":{"Group":"apps","Version":"v1","Kind":"Deployment"},"Resource":"deployments","ObjectName":{"Namespace":"nginx","Name":"nginx-singleton"}}
```

### Expected Behavior

A KS controller manager should never sync Workstatus objects for other WDSes.

### Want to contribute?

### Additional Thoughts

Does combined status have the same issue?

waltforme:
## Summary
Fixes #2455

### DIFF
--- a/pkg/status/status-controller.go
+++ b/pkg/status/status-controller.go
@@ -333,13 +333,13 @@ func (c *Controller) runWorkStatusInformer(ctx context.Context) {
 	// add the event handler functions
 	c.workStatusInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			if shouldSkipAdd(c.wdsName, obj) {
+			if objNotInThisWDS(obj, c.wdsName) {
 				return
 			}
 			c.handleWorkStatus(ctx, obj)
 		},
 		UpdateFunc: func(old, new interface{}) {
-			if shouldSkipUpdate(c.wdsName, old, new) {
+			if objNotInThisWDS(new, c.wdsName) || shouldSkipUpdate(old, new) {
 				return
 			}
 
@@ -359,7 +359,7 @@ func (c *Controller) runWorkStatusInformer(ctx context.Context) {
 			c.handleWorkStatus(ctx, new)
 		},
 		DeleteFunc: func(obj interface{}) {
-			if shouldSkipDelete(c.wdsName, obj) {
+			if objNotInThisWDS(obj, c.wdsName) {
 				return
 			}
 			c.handleWorkStatus(ctx, obj)
@@ -377,28 +377,11 @@ func (c *Controller) runWorkStatusInformer(ctx context.Context) {
 	<-ctx.Done()
 }
 
-func shouldSkipAdd(wdsName string, obj interface{}) bool {
-	if objNotInThisWDS(obj, wdsName) {
-		return true
-	}
-	return false
-}
-
-func shouldSkipUpdate(wdsName string, old, new interface{}) bool {
-	if objNotInThisWDS(new, wdsName) {
-		return true
-	}
+func shouldSkipUpdate(old, new interface{}) bool {
 	oldMObj := old.(metav1.Object)
 	newMObj := new.(metav1.Object)
 	// do not enqueue update events for objects that have not changed
 	return newMObj.GetResourceVersion() == oldMObj.GetResourceVersion()
-}
-
-func shouldSkipDelete(wdsName string, obj interface{}) bool {
-	if objNotInThisWDS(obj, wdsName) {
-		return true
-	}
-	return false
 }
 
 func objNotInThisWDS(obj interface{}, thisWDS string) bool {
--- a/pkg/status/status-controller.go
+++ b/pkg/status/status-controller.go
@@ -333,10 +333,13 @@ func (c *Controller) runWorkStatusInformer(ctx context.Context) {
 	// add the event handler functions
 	c.workStatusInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
+			if shouldSkipAdd(c.wdsName, obj) {
+				return
+			}
 			c.handleWorkStatus(ctx, obj)
 		},
 		UpdateFunc: func(old, new interface{}) {
-			if shouldSkipUpdate(old, new) {
+			if shouldSkipUpdate(c.wdsName, old, new) {
 				return
 			}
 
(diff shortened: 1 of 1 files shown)


pkg/status/status-controller.go
```diff
@@ -333,10 +333,13 @@ func (c *Controller) runWorkStatusInformer(ctx context.Context) {
 	// add the event handler functions
 	c.workStatusInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
+			if shouldSkipAdd(c.wdsName, obj) {
```
ezrasilvera, line 336: Why not a shared ShouldSkip() function that all call and just decide if to skip or not (without performing the action)

pkg/status/status-controller.go
```diff
@@ -374,14 +377,36 @@ func (c *Controller) runWorkStatusInformer(ctx context.Context) {
...
 
-func shouldSkipUpdate(old, new interface{}) bool {
+func shouldSkipAdd(wdsName string, obj interface{}) bool {
+	if objNotInThisWDS(obj, wdsName) {
+		return true
+	}
+	return false
+}
+
+func shouldSkipUpdate(wdsName string, old, new interface{}) bool {
```
ezrasilvera, line 387: This is not really  shouldSkipUpdate()  as it performs the actual update in case of false. Maybe separate and have a general skip as I suggested and add a function for update() that is conditioned by the shouldSKip ?

pkg/status/status-controller.go
waltforme: I'm trying to follow here: What action is performed here?

pkg/status/status-controller.go
waltforme: Also, I'm trying to follow: I don't see actions performed in `shouldSkipUpdate()`, so what actions are we discussing about?

ezrasilvera: @waltforme You are correct. My bad.

pkg/status/status-controller.go
ezrasilvera: @waltforme Regarding the action - you are correct. 
Regarding the sharing?  ShouldAdd and ShouldDelete seems identical.  And also the first section of shouldUpdate. 
Maybe just add IsInThisWDS()   and have the delete and Add use this directly, and the update can use 
`If !IsInThisWDS() || SkipUpdate() { return}`
Not a big issue - if you decide you don't agree - I'm fine.  I just don't like identical implementation in different functions.

waltforme: I got the point. Thanks! And revised as suggested.

pkg/status/status-controller.go
```diff
@@ -381,7 +384,12 @@ func shouldSkipUpdate(old, new interface{}) bool {
 	return newMObj.GetResourceVersion() == oldMObj.GetResourceVersion()
 }
 
-func shouldSkipDelete(_ interface{}) bool {
+func objNotInThisWDS(obj interface{}, thisWDS string) bool {
```
MikeSpreitzer, line 387: This object is not going to be **in** the WDS, it might be **for** or **from** the WDS.

pkg/status/status-controller.go
```diff
@@ -381,7 +384,12 @@ func shouldSkipUpdate(old, new interface{}) bool {
...
+	if objWDS, ok := obj.(metav1.Object).GetLabels()[originWdsLabelKey]; ok {
```
MikeSpreitzer, line 388: What applies this label?

MikeSpreitzer: Oh, I see now that there are _two_ separate symbols for this label's name --- in different packages. Not the best arrangement.

waltforme: Yes I also noted that in the opening of #2455. If we want only one symbol, should the symbol be in pkg/transport/generic (generic_transport_controller.go), or in pkg/util (labels.go)?

MikeSpreitzer: @waltforme, @pdettori, @nirrozenbaum: This label is part of the interface with the transport plugin. Since systems typically distinguish between their main interface with clients and their one or more interfaces with plugin, we could follow that pattern and establish a package for the parts of the interface with the transport plugin that have no better home. Or we could simply not maintain that level of organization (yet), and declare this label in the same `types.go` where the main interface is declared.

PR #2462, opened 2024-09-19
🐛 fix load-image in kind/k3d cluster inside Makefile

francostellari:
## Summary

1. added `kind-load-image` and `k3d-load-image` dependency on `ko-build-controller-manager-local` and `ko-build-transport-local`
2.  added loading the transport image in the local cluster

## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,13 +188,15 @@ ko-build-transport-local: ## Build local transport container image with `ko`.
 
 # this is used for local testing
 .PHONY: kind-load-image
-kind-load-image:
+kind-load-image: ko-build-controller-manager-local ko-build-transport-local
 	kind load --name ${KIND_HOSTING_CLUSTER} docker-image ${CONTROLLER_MANAGER_IMAGE}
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
This is not a _bug fix_, right? What we have now is working, right? As far as I can see, the stuff modified in this PR is used only at https://github.com/kubestellar/kubestellar/blob/46dfeff809375e1ed39693c613f11e06de66d96e/test/e2e/common/setup-kubestellar.sh#L135 . And that line is immediately preceded by `make ko-build-controller-manager-local`.

The fact that the chart in question is simply called "chart" rather that something more descriptive such as controller-manager-chart is a wart of history. At first, that was the only Helm chart in this repo. Now there are also "core-chart" and "ocm-transport-controller-chart".

The controller-manager chart does not actually depend on building the transport controller.

francostellari:
@MikeSpreitzer the core chart will depend on the transport image when I'm done with the merge.

Currently running `make chart` fails without the fix unless in some previous step by the user a build was made.

francostellari:
Included in #2465

PR #2463, opened 2024-09-23
⚠️ Narrow scope of singleton status return request, and refactor

MikeSpreitzer:
## Summary
This PR is a variant on #2408. This moves the `wantSingletonReportedState` bit from the whole Binding[Policy] to its individual workload clauses. This PR also refactors the API a bit, gathering together in one struct all the things that modulate downsync behavior: `createOnly`, `statusCollectors`, `wantSingletonReportedState`.

This PR, so far, only has the API change --- not the implementation change.

## Related issue(s)

Fixes #


api/control/v1alpha1/types.go
```diff
@@ -128,28 +119,52 @@ const (
...
+	// While singleton status return is _not_ requested, KubeStellar suppresses
+	// the existence of a label with that name (key).
+	// While singleton status return is requested _and_ the size of the qualified
+	// WEC set is 1, KubeStellar propagates the object's `.status` from that WEC
+	// to the `.status` section of the object in the WDS.
+	// While either singleton status return is NOT requested or the size of the
+	// qualified WEC set is NOT 1, the `.status` of the object in the WDS does
+	// not have contents propagated there from a WEC by KubeStellar.
+	// +optional
+	WantSingletonReportedState bool `json:"wantSingletonReportedState,omitempty"`
```
vMaroon, line 167: This should not be the only place this API is explained thoroughly, the docs should reflect it as well.

MikeSpreitzer: I have revised the PR, clarifying the last sentence and adding doc updates.

docs/content/direct/combined-status.md
```diff
@@ -194,4 +193,52 @@ select:
...
+the workload object.
+
+For a given workload object in a given WDS, while singleton status
+return is requested, KubeStellar maintains a label on the object whose
+name (key) is `kubestellar.io/executing-count` and whose value is a
+string representation of the size of the qualified WEC set of that
+object.  While singleton status return is _not_ requested, KubeStellar
+suppresses the existence of a label with that name (key).  While
+singleton status return is requested _and_ the size of the qualified
+WEC set is 1, KubeStellar propagates the object's `.status` from that
```
vMaroon, line 240: The case of single status return is requested and the size of the qualified WEC set is bigger than 1 is not addressed. What happens in this case?

MikeSpreitzer: Huh? The last sentence of this paragraph covers that case.

vMaroon: Sorry I see it now, was not visible earlier for some reason.

You mean that if the system is operating well in a case of:
- A user's BP requests singleton status and selects a single WEC
- A different BP is deployed and the total of selected WECs per relevant objects from the above step increases
Then status would first be propagated and finally get dropped when the system reconciles the 2nd point?

MikeSpreitzer: Yes

vMaroon: Sorry we're having tough times here. Why is this preferable over allowing the first to "own" the status delivery and rejecting the rest that follow?

MikeSpreitzer: That would raise additional issues. Perhaps there should be a documented definition of which WEC's status is being returned to the WDS (I am not sure what you mean by "first" in our concurrent and dynamic world). For the sake of not thrashing, the implementation should fix on one WEC and stay latched onto that WEC as long as it is in the qualified set, but switch to another when that WEC drops out of the qualified set. The user will want a way of knowing _which_ WEC's status is being returned to the WDS. I think that it is simpler to just return the status to the WDS only when the qualified WEC set size is 1.

vMaroon: It would probably make a lot more sense with "WEC selection RBAC". 
I think for starters we shouldn't manage racing BPs, and that was not the point in the previous message. My concern is unnecessarily de-propagating status due to same-user errors (having another BP propagate a resource to another WEC) or a different user in another BP.

The challenge in the behavior + API you're proposing is that if one user marks a resource for singleton status, then no one can ship the same resource to another WEC, for any purpose.

Though it is fair to completely shift the responsibility onto the users. I just wonder if we can achieve a more robust behavior without heavy changes to the proposal.

MikeSpreitzer: > Another concern in the behavior + API you're proposing is that if one user marks a resource for singleton status, then no one can ship the same resource to another WEC, for any purpose.

No. Note that this PR and #2408 differ from #2406 in this matter. This is why the "qualified WEC set" consists only of WECs that the object is bound to with `wantSingletonReportedState=true`.

Because there is only one status section in a workload object in the WEC, I do not think we have any real chance of supporting independent uncoordinated users of the same workload object in the same WDS.

MikeSpreitzer: We can not merge just the API change, the PR that gets merged needs to change the implementation too. It should be a relatively small change.

vMaroon: Point is lets move forwards with this direction. I agree it is the simplest, I still think that at the very least some form of feedback should help users understand their incoordination/bad use.

MikeSpreitzer: Agreed on moving forward.
Agree on the desirability of feedback. We had a design, up hrough https://github.com/kubestellar/kubestellar/blob/release-0.24.0/api/control/v1alpha1/types.go#L119-L130 , but that declaration has been deleted since the implementation was deleted when we switched to KSLight. There is a similar design in https://github.com/MikeSpreitzer/kcp-edge-mc/blob/select-to-wds/api/control/v1alpha1/types.go#L427-L428 . There is also a design somewhere for giving feedback in the `.status` of each Binding[Policy].

MikeSpreitzer:
The force-push to 9b6e354b2 is a rebase onto master.

MikeSpreitzer:
The force-push to 75d1534ae is a rebase onto `main`.

api/control/v1alpha1/types.go
```diff
@@ -128,28 +119,52 @@ const (
...
 	StatusCollectors []string `json:"statusCollectors,omitempty"`
+
+	// WantSingletonReportedState, in short, indicates an expectation
+	// that the matching workload objects are distributed to exactly one WEC
+	// and requests that the `.status` of such objects propagate from the WEC
+	// to the WDS.
+	//
+	// For a precise description, start with a couple of definitions.
+	// For a given workload object, _singleton status return is requested_
+	// if and only if there exists at least one BindingPolicy or Binding
```
vMaroon, line 145: Does "at least one BindingPolicy or Binding" imply that Bindings could be made separately?

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -135,6 +125,43 @@ type BindingPolicyResolver interface {
...
+		StatusCollectors:           sets.List(dm.StatusCollectors),
+		WantSingletonReportedState: dm.WantSingletonReportedState,
+	}
+}
+
+func (left *DownsyncModulation) Equal(right DownsyncModulation) bool {
+	return left.CreateOnly == right.CreateOnly && left.WantSingletonReportedState == right.WantSingletonReportedState && left.StatusCollectors.Equal(right.StatusCollectors)
+}
+
+func (dm *DownsyncModulation) AddExternal(external v1alpha1.DownsyncModulation) {
```
vMaroon, line 159: I can't find any use of this function in this repo. I also wonder when would this be useful? A modulation is a member of a resolution, and a resolution is resolved from a single binding-policy.

api/control/v1alpha1/types.go
MikeSpreitzer: It does not imply that. It is worded that way in order to be a useful definition regardless of whether or not there can be a Binding that is not derived from a BindingPolicy.

MikeSpreitzer:
I think this is too big a change to slip in between RCs and the regular release; let's target this to the 0.26.0 release.
/hold

MikeSpreitzer:
Code freeze over.
/remove-hold

PR #2465, opened 2024-09-24
✨ Merge kubestellar and transport charts into the `core-chart`

### LINKED ISSUE #2389: Raise required version of KubeFlex to avoid accidental confusion
https://docs.kubestellar.io/unreleased-development/direct/pre-reqs/#software-prerequisites-for-using-kubestellar says the minimum version of KubeFlex to use is 0.6.1.

We have had confusion caused by the hidden state that `kflex` maintains in the user's kubeconfig file, the preference.extension that holds the name of the context to use when accessing the hosting cluster. Some users somehow got the wrong value in there, or never got any value in there, and it caused them no end of trouble. See #2257 for an example result of the confusion.

The behavior of `kfex` is more assertive in release 0.6.2. I think that the user confusion that we had would not have happened if the users had been using release 0.6.2.

### LINKED ISSUE #2445: bug: CI uses outdated version of kubeflex
### Describe the bug

O crud. All the github Actions workflows install kubeflex release 0.4.2, even though https://github.com/kubestellar/kubestellar/blob/main/docs/content/direct/pre-reqs.md#software-prerequisites-for-using-kubestellar says that the minimum required version is 0.6.1.

### Steps To Reproduce

Look at CI runs from 0.24.0-rc.1

### Expected Behavior

Using kubeflex features that are in 0.6.1 work even if those features are missing from earlier releases.

### Want to contribute?

### Additional Context

_No response_

francostellari:
## Summary

This PR merges the kubestellar and transport charts into the `core-chart` , thus enabling the full testability of local container images and Helm chart.

Specifically:
1. remove the `chart/` and `ocm-transport-controller-chart/` folders
2. remove `chart` and `install-local-chart` targets from the `Makefile`
3. remove the generation of the `controller-manager-chart` and `ocm-transport-controller-chart` chart generation by goreleaser
4. updated the `config` kustomization files to replace `controller-manager-chart` Helm parameters with PCH parameters
5. merged the `ocm-transport-controller-chart`  YAML into the `wds.yaml` PCH in the `core-chart`
6. added a `core-chart` target to the `Makefile` to create the kubestellar plugin yaml and merge it into the `core-chart`
7. added a `install-local-core-chart` Makefile target to install the local image and charts to a cluster
8. update kubeflex to 0.6.3 (required by this release)
9. updated the goreleaser to generate the new `core-chart`
10. updated the `e2e\common\setup-kubestellar.sh` script to use the new `core-chart` for local testing

**NOTE:** from the end-user point of view the `core-chart` Helm chart should behave exactly as before.

Proposed future follow up work:
1. remove the kustomization of kubestellar controller YAML and merge it as  static code into the `wds.yaml` PCH
2. identify commonalities in the transport-controller and kubestellar-controller-manager RBAC (role, rolebindings, etc) and reduce/simplify/share the resources created by the wds PCH (this will be much easier to figure out after 1.)
3.  should we rename `core-chart` to just `chart` since it is the only one?

## Related issue(s)

Fixes #2389
Fixes #2445


MikeSpreitzer:
There is something strange going on in the commits in the branch (https://github.com/francostellari/kubestellar/tree/merge-charts) that this PR is requesting to be merged into `main` --- some of them are already in `main`.

MikeSpreitzer:
This may show what is giving the verify job indigestion.

```console
mspreitz@mjs13 kubestellar % git status
On branch merge-charts
Your branch is up to date with 'fs/merge-charts'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	test/performance/.venv/
	test/performance/__pycache__/

nothing added to commit but untracked files present (use "git add" to track)

mspreitz@mjs13 kubestellar % go mod tidy

mspreitz@mjs13 kubestellar % git status
On branch merge-charts
Your branch is up to date with 'fs/merge-charts'.

	modified:   go.sum

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	test/performance/.venv/
	test/performance/__pycache__/

no changes added to commit (use "git add" and/or "git commit -a")
```

core-chart/values.yaml
```diff
@@ -5,11 +5,12 @@
...
 # please do not change them unless you know what you are doing.
-HELM_VERSION: "3.15.0"
-KUBECTL_VERSION: "1.30.1"
+HELM_VERSION: "3.16.1"
+KUBECTL_VERSION: "1.31.0"
 KUBESTELLAR_VERSION: "0.24.0"
+TRANSPORT_VERSION: "0.24.0"
 CLUSTERADM_VERSION: "0.8.2"
-OCM_STATUS_ADDON_VERSION: "0.2.0-rc11"
+OCM_STATUS_ADDON_VERSION: "0.2.0-rc12"
```
MikeSpreitzer, line 13: This change has already been done by #2468

core-chart/values.yaml
```diff
@@ -5,11 +5,12 @@
 
 # Container/chart image versions. These values are set during the chart release process,
 # please do not change them unless you know what you are doing.
-HELM_VERSION: "3.15.0"
-KUBECTL_VERSION: "1.30.1"
+HELM_VERSION: "3.16.1"
```
MikeSpreitzer, line 8: +1
(https://helm.sh/docs/topics/version_skew/)

Makefile
```diff
@@ -19,6 +19,7 @@ TRANSPORT_CMD_NAME ?= ocm-transport-controller
 IMAGE_TAG ?= $(shell git rev-parse --short HEAD)
 CONTROLLER_MANAGER_IMAGE ?= ${DOCKER_REGISTRY}/${CONTROLLER_MANAGER_CMD_NAME}:${IMAGE_TAG}
 TRANSPORT_IMAGE ?= ${DOCKER_REGISTRY}/${TRANSPORT_CMD_NAME}:${IMAGE_TAG}
+INSTALL_KUBEFLEX ?= true
```
MikeSpreitzer, line 22: Why is this an option? Our pre-reqs.md doc says that KubeFlex is a pre-req for building and using KubeStellar.

MikeSpreitzer:
This is a big PR. Let's get the easy preliminaries out of the way first. So far I have identified #2472, which should have been done earlier for its own reasons.

francostellari:
> There is something strange going on in the commits in the branch (https://github.com/francostellari/kubestellar/tree/merge-charts) that this PR is requesting to be merged into `main` --- some of them are already in `main`. I think that the the `merge-charts` branch should be rebased onto `main`. Following is what it looked like when I tried it.
> 
> ```
> mspreitz@mjs13 kubestellar % git checkout merge-charts
> branch 'merge-charts' set up to track 'fs/merge-charts'.
> Switched to a new branch 'merge-charts'
> 
> mspreitz@mjs13 kubestellar % git rebase main
> warning: skipped previously applied commit d61145583
> warning: skipped previously applied commit 1e18d2160
> warning: skipped previously applied commit 4bb04db75
> warning: skipped previously applied commit 544642709
> hint: use --reapply-cherry-picks to include skipped commits
> hint: Disable this message with "git config advice.skippedCherryPicks false"
> dropping 94d3477c44b18d6af7eb2bb751d4c3d9f7b3bbbd update status-adon version -- patch contents already upstream
> Successfully rebased and updated refs/heads/merge-charts.
> ```

```
 git rebase origin/main
warning: skipped previously applied commit d61145583
warning: skipped previously applied commit 1e18d2160
warning: skipped previously applied commit 4bb04db75
warning: skipped previously applied commit 544642709
hint: use --reapply-cherry-picks to include skipped commits
hint: Disable this message with "git config advice.skippedCherryPicks false"
dropping 94d3477c44b18d6af7eb2bb751d4c3d9f7b3bbbd update status-adon version -- patch contents already upstream
Successfully rebased and updated refs/heads/merge-charts.

 git status
On branch merge-charts
Your branch and 'origin/merge-charts' have diverged,
and have 21 and 17 different commits each, respectively.
  (use "git pull" if you want to integrate the remote branch with yours)

nothing to commit, working tree clean`
``

File: Makefile (hunk 3 above)
Comment (francostellari) [reply Makefile:22]: Since the test script create the kind cluster using `kflex init -c`, Kubeflex is already deployed in the cluster thus making necessary a flag to avoid trying to redeploy it with the `core-chart`.

**Comment by MikeSpreitzer** [comment]
@francostellari : you showed zero merge conflicts. It is normal and expected for `git status` to talk about divergence.

**Comment by francostellari** [comment]
> @francostellari : you showed zero merge conflicts. It is normal and expected for `git status` to talk about divergence.

So should I `git pull` as suggested? fastforward?

**Comment by MikeSpreitzer** [comment]
The thing to do after a successful rebase is to force-push back to your fork.

**Comment by MikeSpreitzer** [comment]
@francostellari: you need to do `go mod tidy`. It would be best if that were in the same commit as the change to `go.mod`, but adequate to add a commit making this change.

```console
mspreitz@mjs13 kubestellar % git status
On branch merge-charts
Your branch is up to date with 'fs/merge-charts'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	test/performance/.venv/
	test/performance/__pycache__/

nothing added to commit but untracked files present (use "git add" to track)

mspreitz@mjs13 kubestellar % go mod tidy

mspreitz@mjs13 kubestellar % git diff
diff --git a/go.sum b/go.sum
index f33b7fb0d..01427074f 100644
--- a/go.sum
+++ b/go.sum
@@ -332,8 +332,6 @@ github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
 github.com/kr/text v0.2.0 h1:5Nx0Ya0ZqY2ygV366QzturHI13Jq95ApcVaJBhpS+AY=
 github.com/kr/text v0.2.0/go.mod h1:eLer722TekiGuMkidMxC/pM04lWEeraHUUmBw8l2grE=
-github.com/kubestellar/kubeflex v0.6.1 h1:NsWbqU7KQbSayOny7/++3KP9/0GMhWFudwxKRRkWe/c=
-github.com/kubestellar/kubeflex v0.6.1/go.mod h1:kFHuAQkw1z/6Ul7vuX2xnvZAE1C5w8WxYweWMjH4UM4=
 github.com/kubestellar/kubeflex v0.6.3 h1:hajFsW+MSj1SskV8GgleBtZlzFkmEIi2BOZq4ymoXnA=
 github.com/kubestellar/kubeflex v0.6.3/go.mod h1:kFHuAQkw1z/6Ul7vuX2xnvZAE1C5w8WxYweWMjH4UM4=
 github.com/mailru/easyjson v0.7.7 h1:UGYAvKxe3sBsEDzO8ZeWOSlIQfWFlxbzLZe7hwFURr0=
```

MikeSpreitzer:
Regarding the proposed future work:

2. I am OK with having distinct YAMLs for the OCM transport controller, as an illustration of how a different transport controller could be made and used.
3. I am OK with keeping "core" in the name of the core-chart, because there is more to KubeStellar than the core.

Makefile
```diff
@@ -188,22 +189,37 @@ ko-build-transport-local: ## Build local transport container image with `ko`.
...
-		| yq '. | select(.kind == "ClusterRole*").metadata.name |= "{{.Values.ControlPlaneName}}-" + .' \
-		| yq '. | select(.kind == "ClusterRoleBinding").roleRef.name |= "{{.Values.ControlPlaneName}}-" + .' \
-		> chart/templates/controller-manager.yaml
+		| yq '. | select(.kind == "ClusterRole*").metadata.name |= "{{\"{{.ControlPlaneName}}\"}}-" + .' \
+		| yq '. | select(.kind == "ClusterRoleBinding").roleRef.name |= "{{\"{{.ControlPlaneName}}\"}}-" + .' \
+		| yq '[.]' \
+		| sed -e 's/^/  /' \
+		> /tmp/kubestellar-controller-manager.yaml
 	git checkout -- config/manager/kustomization.yaml
+	sed -i -e '/KUBESTELLAR_CONTROLLER_PLACEHOLDER/{r /tmp/kubestellar-controller-manager.yaml' -e 'd}' core-chart/templates/postcreatehooks/wds.yaml
```
MikeSpreitzer, line 211: This is not working for me.

```console
mspreitz@mjs13 kubestellar % make core-chart
test -s /Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin/controller-gen && /Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin/controller-gen --version | grep -q v0.13.0 || \
	GOTOOLCHAIN=go1.21.8 GOBIN=/Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.13.0
test -s /Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin/kustomize || GOBIN=/Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin go install sigs.k8s.io/kustomize/kustomize/v5@v5.3.0
/Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./api/..." output:crd:artifacts:config=config/crd/bases
/Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin/kustomize build config/crd/ > pkg/crd/files/crds.yaml
/Users/mspreitz/go/src/github.com/kubestellar/kubestellar/bin/kustomize build config/default \
		| yq '. | select(.kind == "ClusterRole*").metadata.name |= "{{\"{{.ControlPlaneName}}\"}}-" + .' \
		| yq '. | select(.kind == "ClusterRoleBinding").roleRef.name |= "{{\"{{.ControlPlaneName}}\"}}-" + .' \
		| yq '[.]' \
		| sed -e 's/^/  /' \
		> /tmp/kubestellar-controller-manager.yaml
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
git checkout -- config/manager/kustomization.yaml
sed -i -e '/KUBESTELLAR_CONTROLLER_PLACEHOLDER/{r /tmp/kubestellar-controller-manager.yaml' -e 'd}' core-chart/templates/postcreatehooks/wds.yaml
sed: 1: "/KUBESTELLAR_CONTROLLER ...": unexpected EOF (pending }'s)
make: *** [core-chart] Error 1
```

Makefile
```diff
@@ -188,22 +189,37 @@ ko-build-transport-local: ## Build local transport container image with `ko`.
...
-chart: manifests kustomize
-	cd config/manager && $(KUSTOMIZE) edit set image controller=$(shell echo ${CONTROLLER_MANAGER_IMAGE} | sed 's/\(:.*\)v/\1/')
+.PHONY: core-chart
+core-chart: manifests kustomize
 	$(KUSTOMIZE) build config/default \
-		| yq '. | select(.kind == "ClusterRole*").metadata.name |= "{{.Values.ControlPlaneName}}-" + .' \
-		| yq '. | select(.kind == "ClusterRoleBinding").roleRef.name |= "{{.Values.ControlPlaneName}}-" + .' \
-		> chart/templates/controller-manager.yaml
+		| yq '. | select(.kind == "ClusterRole*").metadata.name |= "{{\"{{.ControlPlaneName}}\"}}-" + .' \
+		| yq '. | select(.kind == "ClusterRoleBinding").roleRef.name |= "{{\"{{.ControlPlaneName}}\"}}-" + .' \
```
MikeSpreitzer, line 206: Why make these changes here with `yq` instead of just changing the source permanently or using `kustomize`?

core-chart/templates/postcreatehooks/wds.yaml
```diff
@@ -43,185 +38,132 @@ spec:
...
+          - name: shared-volume
+            emptyDir: {}
+          - name: config-volume
+            configMap:
+              name: transport-controller-config
+              defaultMode: 0744
+  # ^^^^ TRANSPORT CONTROLLER ^^^^
+
+  # vvvv KUBESTELLAR CONTROLLER vvvv
+  # KUBESTELLAR_CONTROLLER_PLACEHOLDER
```
MikeSpreitzer, line 166: I do not see the `# ` in the stuff being replaced in the `sed` command in the Makefile.

Makefile
```diff
@@ -188,22 +189,37 @@ ko-build-transport-local: ## Build local transport container image with `ko`.
...
+		| yq '[.]' \
+		| sed -e 's/^/  /' \
+		> /tmp/kubestellar-controller-manager.yaml
 	git checkout -- config/manager/kustomization.yaml
```
MikeSpreitzer, line 210: Why is this line still here?

core-chart/values.yaml
```diff
@@ -5,11 +5,12 @@
...
+KUBECTL_VERSION: "1.31.0"
```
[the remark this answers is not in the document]
francostellari: Perhaps, I but this has a limited usage and as long as it works, it should be ok to upgrade.

core-chart/values.yaml
francostellari: Perhaps, I but this has a limited usage and as long as it works, it should be ok to upgrade.

francostellari:
> @francostellari: you need to do `go mod tidy`. It would be best if that were in the same commit as the change to `go.mod`, but adequate to add a commit making this change. (BTW, #2471 has merged, so another rebase onto `main` would eliminate the noise about those untracked files.)
> 
> ```
> mspreitz@mjs13 kubestellar % git status
> On branch merge-charts
> Your branch is up to date with 'fs/merge-charts'.
> 
> Untracked files:
>   (use "git add <file>..." to include in what will be committed)
> 	test/performance/.venv/
> 	test/performance/__pycache__/
> 
> nothing added to commit but untracked files present (use "git add" to track)
> 
> mspreitz@mjs13 kubestellar % go mod tidy
> 
> mspreitz@mjs13 kubestellar % git diff
> diff --git a/go.sum b/go.sum
> index f33b7fb0d..01427074f 100644
> --- a/go.sum
> +++ b/go.sum
> @@ -332,8 +332,6 @@ github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
>  github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
>  github.com/kr/text v0.2.0 h1:5Nx0Ya0ZqY2ygV366QzturHI13Jq95ApcVaJBhpS+AY=
>  github.com/kr/text v0.2.0/go.mod h1:eLer722TekiGuMkidMxC/pM04lWEeraHUUmBw8l2grE=
> -github.com/kubestellar/kubeflex v0.6.1 h1:NsWbqU7KQbSayOny7/++3KP9/0GMhWFudwxKRRkWe/c=
> -github.com/kubestellar/kubeflex v0.6.1/go.mod h1:kFHuAQkw1z/6Ul7vuX2xnvZAE1C5w8WxYweWMjH4UM4=
>  github.com/kubestellar/kubeflex v0.6.3 h1:hajFsW+MSj1SskV8GgleBtZlzFkmEIi2BOZq4ymoXnA=
>  github.com/kubestellar/kubeflex v0.6.3/go.mod h1:kFHuAQkw1z/6Ul7vuX2xnvZAE1C5w8WxYweWMjH4UM4=
>  github.com/mailru/easyjson v0.7.7 h1:UGYAvKxe3sBsEDzO8ZeWOSlIQfWFlxbzLZe7hwFURr0=
> ```

ok done

Makefile
MikeSpreitzer: Also, stuffing a whole file unescaped into the `replacement` part of a `s/regexp/replacement/flags` command is problematic, because the `replacement` is interpreted.

francostellari: I'm not sure what the problem is. It's working for me.

```shell
$ sed --version
sed (GNU sed) 4.8
```

```shell
$ make core-chart
test -s /vagrant/kubestellar/bin/controller-gen && /vagrant/kubestellar/bin/controller-gen --version | grep -q v0.13.0 || \
GOTOOLCHAIN=go1.21.8 GOBIN=/vagrant/kubestellar/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.13.0   
test -s /vagrant/kubestellar/bin/kustomize || GOBIN=/vagrant/kubestellar/bin go install sigs.k8s.io/kustomize/kustomize/v5@v5.3.0
/vagrant/kubestellar/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./api/..." output:crd:artifacts:config=config/crd/bases
/vagrant/kubestellar/bin/kustomize build config/crd/ > pkg/crd/files/crds.yaml
/vagrant/kubestellar/bin/kustomize build config/default \
        | yq '. | select(.kind == "ClusterRole*").metadata.name |= "{{\"{{.ControlPlaneName}}\"}}-" + .' \
        | yq '. | select(.kind == "ClusterRoleBinding").roleRef.name |= "{{\"{{.ControlPlaneName}}\"}}-" + .' \
        | yq '[.]' \
        | sed -e 's/^/  /' \
        > /tmp/kubestellar-controller-manager.yaml
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
git checkout -- config/manager/kustomization.yaml
sed -i -e '/KUBESTELLAR_CONTROLLER_PLACEHOLDER/{r /tmp/kubestellar-controller-manager.yaml' -e 'd}' core-chart/templates/postcreatehooks/wds.yaml
```

Makefile
francostellari: The lines were already there in the Makefile.
I was trying to make the minimal changes to the code.
I think all this will be cleanup once we remove the kustomization of the KS template

core-chart/templates/postcreatehooks/wds.yaml
francostellari: The `#` does not matter because the entire line is replaced

Makefile
francostellari: The `git checkou`?
`make core-chart` is still blaming that file for the moment.
We will remove it when the kustomization is removed.

Makefile
francostellari: > Also, stuffing a whole file unescaped into the `replacement` part of a `s/regexp/replacement/flags` command is problematic, because the `replacement` is interpreted.

Perhaps but it's working. All will be removed once we remove the kustomization,
I'm really trying to do this in 2 steps instead one one.

MikeSpreitzer: FYI:

```console
mspreitz@mjs13 kubestellar % fgrep -n '# KUBESTELLAR_CONTROLLER_PLACEHOLDER' core-chart/templates/postcreatehooks/wds.yaml              
166:  # KUBESTELLAR_CONTROLLER_PLACEHOLDER
mspreitz@mjs13 kubestellar % 
mspreitz@mjs13 kubestellar % fgrep -n '# KUBESTELLAR_CONTROLLER_PLACEHOLDER' core-chart/templates/postcreatehooks/wds.yaml | cut -d: -f1
166
```

francostellari:
```shell
$ fgrep -n '# KUBESTELLAR_CONTROLLER_PLACEHOLDER' core-chart/templates/postcreatehooks/wds.yaml
166:  # KUBESTELLAR_CONTROLLER_PLACEHOLDER

$ make core-chart
test -s /vagrant/kubestellar/bin/controller-gen && /vagrant/kubestellar/bin/controller-gen --version | grep -q v0.13.0 || \
GOTOOLCHAIN=go1.21.8 GOBIN=/vagrant/kubestellar/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.13.0
test -s /vagrant/kubestellar/bin/kustomize || GOBIN=/vagrant/kubestellar/bin go 
install sigs.k8s.io/kustomize/kustomize/v5@v5.3.0
/vagrant/kubestellar/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./api/..." output:crd:artifacts:config=config/crd/bases
/vagrant/kubestellar/bin/kustomize build config/crd/ > pkg/crd/files/crds.yaml
/vagrant/kubestellar/bin/kustomize build config/default \
        | yq '. | select(.kind == "ClusterRole*").metadata.name |= "{{\"{{.ControlPlaneName}}\"}}-" + .' \
        | yq '. | select(.kind == "ClusterRoleBinding").roleRef.name |= "{{\"{{.ControlPlaneName}}\"}}-" + .' \
        | yq '[.]' \
        | sed -e 's/^/  /' \
        > /tmp/kubestellar-controller-manager.yaml
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. 
Run 'kustomize edit fix' to update your Kustomization automatically.
git checkout -- config/manager/kustomization.yaml
sed -i -e '/KUBESTELLAR_CONTROLLER_PLACEHOLDER/{r /tmp/kubestellar-controller-manager.yaml' -e 'd}' core-chart/templates/postcreatehooks/wds.yaml

$ fgrep -n '# KUBESTELLAR_CONTROLLER_PLACEHOLDER' core-chart/templates/postcreatehooks/wds.yaml
$

```

MikeSpreitzer: FYI, I find that the following works. Except for the fact that it injects LF-terminated lines into a file that is CRLF-terminated.

```
	(where=$$(fgrep -n '# KUBESTELLAR_CONTROLLER_PLACEHOLDER' core-chart/templates/postcreatehooks/wds.yaml | cut -d: -f1) &&\
	echo where=[$$where] &&\
	head -n $$(( where-1 )) core-chart/templates/postcreatehooks/wds.yaml > replaced.yaml &&\
	cat /tmp/kubestellar-controller-manager.yaml >> replaced.yaml &&\
	tail -n +$$(( where+1 )) core-chart/templates/postcreatehooks/wds.yaml >> replaced.yaml )
	mv replaced.yaml core-chart/templates/postcreatehooks/wds.yaml
```

MikeSpreitzer: See https://github.com/MikeSpreitzer/kcp-edge-mc/commits/merge-charts/

core-charttemplatespostcreatehookswds.yaml
```diff
@@ -0,0 +1,695 @@
...
+          - name: config-volume
+            configMap:
+              name: transport-controller-config
+              defaultMode: 0744
+
+  # ^^^^ TRANSPORT PLUGIN ^^^^
+
+  # vvvv KUBESTELLAR PLUGIN vvvv
+
+  - apiVersion: rbac.authorization.k8s.io/v1
```
MikeSpreitzer, line 331: GitHub does not make this clear, but: the lines above this one are terminated with CRLF, while the lines in this section are terminated by LF. This is a bit of a pain; it would be better if all the lines were terminated by LF.

core-charttemplatespostcreatehookswds.yaml
MikeSpreitzer, line 1: What is this file here for?

```console
mspreitz@mjs13 kubestellar % find * .github/workflows \( -name "*.svg" -prune \) -or \( -path "*venv" -prune \) -or \( -path hack/tools -prune \) -or \( -type f -exec grep charttemplate \{\} \; -print -exec echo \; \)
mspreitz@mjs13 kubestellar % 
```

Makefile
MikeSpreitzer: It needs to work for developers using a Mac.
I am using MacOS Sonoma, make version 3.81, my bash version is 3.2.57, and my default shell is zsh version 5.9.

core-chart/values.yaml
MikeSpreitzer: "+1" means approval. I like that you updated the Helm version to better match our usage.

core-chart/values.yaml
MikeSpreitzer: The rebase onto `main` picked that up. This is resolved now.

core-chart/values.yaml
```diff
@@ -5,9 +5,10 @@
...
 KUBESTELLAR_VERSION: "0.24.0"
+TRANSPORT_VERSION: "0.24.0"
```
MikeSpreitzer, line 11: I am not a fan of adding self-references. OTOH, I _am_ a fan of making it clear how an out-of-tree transport-controller could be created and used. If this were a Makefile or script then we could say _here_ that `TRANSPORT_VERSION` defaults to whatever `KUBESTELLAR_VERSION` is --- but, as far as I know, you can not do such a thing in a Helm values file.

core-chart/values.yaml
```diff
@@ -58,8 +59,20 @@ InstallPCHs: true
 
 # Transport controller bundling parameters
```
MikeSpreitzer, line 60: It's not just _bundling_ parameters any more.

Makefile
MikeSpreitzer: Oh, I misread the failing command. It is not a replace command, it is using a pattern to identify a line and then applied a pair of `r` and `d` commands to that line. I suspect the problem is just in the way that the bracketed pair of `r` and `d` commands is formed is the problem.

MikeSpreitzer: I am getting closer with the following.

```
	sed -e $$'/KUBESTELLAR_CONTROLLER_PLACEHOLDER/{=; r /tmp/kubestellar-controller-manager.yaml ; d\n}' -i.bak core-chart/templates/postcreatehooks/wds.yaml
```

`man sed` says that the closing curly bracket has to be preceded by a newline. But the replacement is not happening.

MikeSpreitzer: OK, I am succeeding with the following.

```
	sed -e $$'/KUBESTELLAR_CONTROLLER_PLACEHOLDER/{r /tmp/kubestellar-controller-manager.yaml\n d\n}' -i.bak core-chart/templates/postcreatehooks/wds.yaml
```

Except that:
1. This still mixes LF-terminated content into a CRLF-terminated file.
2. This creates a `.bak` file. MacOS and Linux are ferociously incompatible in how to tell `sed` to process a file in place without creating a backup.

Makefile
francostellari: It's used in the context of the helm chart and it's trying to mimic the option in the `value.yaml`

Makefile
francostellari: If you can provide code that works on both Linux and Mac, I personally don't care.
This will all be removed soon after the PR is merged

core-charttemplatespostcreatehookswds.yaml
```diff
@@ -0,0 +1,695 @@
...
+          - name: shared-volume
+            emptyDir: {}
+          - name: config-volume
+            configMap:
+              name: transport-controller-config
+              defaultMode: 0744
+
+  # ^^^^ TRANSPORT PLUGIN ^^^^
+
+  # vvvv KUBESTELLAR PLUGIN vvvv
```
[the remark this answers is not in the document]
francostellari: I renamed to controller for both

core-charttemplatespostcreatehookswds.yaml
francostellari: I pushed an update forcing the file to LF but I'm confused because my git config should be setup to convert all CRLF to LF

core-charttemplatespostcreatehookswds.yaml
francostellari: Sorry, I don't understand

core-chart/values.yaml
francostellari: No values in in `value.yaml` cannot reference other values in the same file.
We could compromise in the way that if `TRANSPORT_VERSION` is not set, the it defaults to `KUBESTELLAR_VERSION` by changing the PCH value substitution.

core-chart/values.yaml
francostellari: I changed to just "parameters"

Makefile
MikeSpreitzer: OK, I think I have something that works on both Mac and Linux. I tested it on MacOS Sonoma and Ubuntu 22.04.

Rather than go to all the lengths necessary to cope with the `-i` divergence between `sed` on Linux and MacOS, I took a different approach --- distinct input and output files. This has the added advantage that `make core-chart` no longer is modifying a file that is under source code control.

See https://github.com/MikeSpreitzer/kcp-edge-mc/tree/merge-charts

Makefile
MikeSpreitzer: Yes, that `git checkout ...`.
I do not understand what you mean by "`make core-chart` is still _blaming_ that file". As far as I can tell, deleting this `git checkout` command will make no difference (besides simplifying the recipe). I tested with that deletion, and found that `make core-chart` does not modify that file.

core-charttemplatespostcreatehookswds.yaml
MikeSpreitzer: I am asking why this PR is adding this file. As far as I can tell, this file is not used for anything.

core-chart/values.yaml
MikeSpreitzer: Yes, but that requires duplicating the defaulting logic at every use rather than declaring it once at the definition. I am not asking for a change here.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -106,55 +106,28 @@ esac
...
+make install-local-core-chart \
+  INSTALL_KUBEFLEX=false \
+  ITS_NAME=its1 \
+  DEFAULT_WDS_NAME=wds1 \
+  KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY=$KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY
+popd
 
 wait-for-cmd "(kubectl --context '$HOSTING_CONTEXT' -n wds1-system wait --for=condition=Ready pod/\$(kubectl --context '$HOSTING_CONTEXT' -n wds1-system get pods -l name=transport-controller -o jsonpath='{.items[0].metadata.name}'))"
 
 echo "transport controller is running."
```
MikeSpreitzer, line 123: Don't we also want to wait for the kubestellar-controller-manager to be running? Perhaps even wait for it to get the KubeStellar CRDs "established"?

MikeSpreitzer (commented):
With the added two commits in https://github.com/MikeSpreitzer/kcp-edge-mc/tree/merge-charts, I ran successfully on my Mac both the bash-based and ginkgo-based end-to-end tests.

core-charttemplatespostcreatehookswds.yaml
francostellari: Sorry, I still don't understand which file

core-charttemplatespostcreatehookswds.yaml
MikeSpreitzer: Thanks for the file fix.

core-chart/values.yaml
francostellari: it's not complicated, I made the change in the new commit
the PCH will use the TRANSPORT_VERSION if defined and default to KUBESTELLAR_VERSION otherwise

test/e2e/common/setup-kubestellar.sh
francostellari: I did not change any of the waiting compared to what was there before.
it seems to work as it is.

Makefile
MikeSpreitzer: @francostellari has taken a similar but different approach, using `sed -i.bak ..; rm ....bak`. I would prefer to not be in the business of modifying a file under Git control. This can be pursued in follow-on work.

Makefile
MikeSpreitzer: This line has been removed.

core-charttemplatespostcreatehookswds.yaml
MikeSpreitzer: I mean the individual file that this comment is attached to. The only individual file that shows up in the following listing. This file is not used by anything, so I wonder why this PR introduces it.

```console
mspreitz@mjs13 kubestellar % ls -ld core-chart*
drwxr-xr-x  10 mspreitz  staff    320 Oct  1 17:37 core-chart
-rw-r--r--   1 mspreitz  staff  21635 Oct  1 17:37 core-charttemplatespostcreatehookswds.yaml
```

core-chart/values.yaml
MikeSpreitzer: I see that there is only one use, besides the one in the unused file `core-charttemplatespostcreatehookswds.yaml`, so the complexity as measured in source tokens is about the same (currently).

test/e2e/common/setup-kubestellar.sh
MikeSpreitzer: Indeed, but you changed the earlier contents of the script, so maybe more waiting is in order now? Can be pursued in follow-on.

core-charttemplatespostcreatehookswds.yaml
francostellari: ok sorry for being slow, I just remove it

MikeSpreitzer (approved):
I tested this on my Mac:
- `test/e2e/run-test.sh` succeeded
- `test/e2e/run-test.sh --released` failed the first time I tried it, with `kubectl --context kind-kubeflex get -n its1-system secret vc-vcluster -o yaml` showing that the Secret did NOT have a data member named "config-incluster" (but it had the other expected data members). Then I tested three more times, and got success each time.

This is good enough to merge and continue working on in later PRs.

/lgtm
/approve
Thanks!

.github/workflows/goreleaser.yml
```diff
@@ -62,14 +58,11 @@ jobs:
     - name: Package and push charts
       run: |
         chartVersion=$(echo ${{ github.ref_name }} | cut -c 2-)
-        make chart CONTROLLER_MANAGER_IMAGE=${{ env.REGISTRY }}/${{ env.CONTROLLER_MANAGER_IMAGE }}:${{github.ref_name}}
-        helm package ${{ env.CM_CHART_PATH }} --destination . --version ${chartVersion} --app-version ${chartVersion}
-        helm push ./controller-manager-chart*.tgz oci://${{ env.REGISTRY }}/kubestellar/kubestellar
-        
+
+        make core-chart IMAGE_TAG=${{chartVersion}}
```
MikeSpreitzer, line 62: This caused the failure at https://github.com/kubestellar/kubestellar/actions/runs/11170183214 . The problem here is that the syntax `${{chartVersion}}` is asking for GitHub to expand `chartVersion` but that is not an environment version at the GitHub workflow level. It is a shell variable inside the shell script here. You want to use the same syntax seen elsewhere in the same script for expansion of that shell variable.

PR #2473, opened 2024-10-02
✨Remove Kustomizer usage to build the core-chart

francostellari:
## Summary

Remove the use of Kustomizer to dynamically generate the YAML for the KubeStellar controller deployment.

Key changes:
- update the WDS PCH to have static YAML for deploying KubeStellar controller
- remove several unused targets from Makefile, including `core-chart`
- remove `config/` folder

## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,28 +199,6 @@ k3d-load-image: ko-build-controller-manager-local ko-build-transport-local
 	k3d image import ${CONTROLLER_MANAGER_IMAGE} -c ${K3D_HOSTING_CLUSTER}
 	k3d image import ${CONTROLLER_MANAGER_IMAGE} -c ${TRANSPORT_IMAGE}
 
-.PHONY: core-chart
-core-chart: manifests kustomize
-	$(KUSTOMIZE) build config/default \
-		| yq '. | select(.kind == "ClusterRole*").metadata.name |= "{{\"{{.ControlPlaneName}}\"}}-" + .' \
-		| yq '. | select(.kind == "ClusterRoleBinding").roleRef.name |= "{{\"{{.ControlPlaneName}}\"}}-" + .' \
-		| yq '[.]' \
-		| sed -e 's/^/  /' \
-		> /tmp/kubestellar-controller-manager.yaml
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -140,13 +140,6 @@ $(CODE_GEN_DIR):
 
 ##@ Development
 
-## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects,
-## then Kustomize the CustomResourceDefinition objects for the 'crd' package to use.
-.PHONY: manifests
-manifests: controller-gen kustomize
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./api/..." output:crd:artifacts:config=config/crd/bases
```
MikeSpreitzer, line 147: We still need a `make` target that generates the CRDs, right?

config/crd/patches/more_validations_in_bindingpolicies.yaml
MikeSpreitzer, line 1: This file was doing something we still want, adding more validations to some CRDs.

MikeSpreitzer (commented):
This PR makes it clear that we need to document the uses of the Makefile. There is some in https://docs.kubestellar.io/release-0.24.0/direct/packaging/ , but it may not be comprehensive and gathered in one place and nobody seems to know about it.

It would be OK with me to remove the use of `kustomize` after #2474 .

Makefile
francostellari: Not for the core-helm chart.
What do you think is using that YAML?

config/crd/patches/more_validations_in_bindingpolicies.yaml
francostellari: I need help identifying where this would be used.
I was able to build the controllers and install the local Helm chart without it.

Makefile
MikeSpreitzer: KubeStellar certainly needs to maintain its CRDs! If I recall correctly, the KubeStellar controller-manager establishes some in each WDS and the OCM Status Add-On Controller establishes some in the ITS.

config/crd/patches/more_validations_in_bindingpolicies.yaml
MikeSpreitzer: These validations are being added into the Custom Resource _Definitions_. These validations are checking the user-supplied content for the custom resources. Our tests supply only valid input, so this additional validation is not tested by our tests. That should probably be fixed.

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #2474, opened 2024-10-02
🐛 Fix `run_test --released`

francostellari:
## Summary

Streamline the `run_test --released` to actually test the latest released core-chart and controller images.

## Related issue(s)

Fixes #

### DIFF
--- a/test/e2e/common/setup-kubestellar.sh
+++ b/test/e2e/common/setup-kubestellar.sh
@@ -73,10 +73,6 @@ then echo "$0: \$TRANSPORT_CONTROLLER_VERBOSITY must be an integer" >&2
      exit 1
 fi
 
-if [ "$use_release" = true ]
-then wds_extra="-p kubestellar --set ControllerManagerVerbosity=$KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY --set TransportControllerVerbosity=$TRANSPORT_CONTROLLER_VERBOSITY"
-else wds_extra=""
-fi
 
 if [[ "$KFLEX_DISABLE_CHATTY" = true ]] ; then
    disable_chatty_status="--chatty-status=false"
@@ -110,16 +106,27 @@ esac
 :
 
 pushd "${SRC_DIR}/../../.."
-make kind-load-image
-helm upgrade --install ks-core core-chart/ \
-  --dependency-update \
-  --set KUBESTELLAR_VERSION=$(git rev-parse --short HEAD) \
-  --kube-context $HOSTING_CONTEXT \
-  --set kubeflex-operator.install=false \
-  --set-json='ITSes=[{"name":"its1"}]' \
-  --set-json='WDSes=[{"name":"wds1"}]' \
-  --set verbosity.kubestellar=${KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY} \
-  --set verbosity.transport=${TRANSPORT_CONTROLLER_VERBOSITY}
+if [ "$use_release" = true ] ; then
+  helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart \
+    --version $(yq .KUBESTELLAR_VERSION core-chart/values.yaml) \
+    --kube-context $HOSTING_CONTEXT \
+    --set kubeflex-operator.install=false \
+    --set-json='ITSes=[{"name":"its1"}]' \
+    --set-json='WDSes=[{"name":"wds1"}]' \
+    --set verbosity.kubestellar=${KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY} \
+    --set verbosity.transport=${TRANSPORT_CONTROLLER_VERBOSITY}
+else
+  make kind-load-image
+  helm upgrade --install ks-core core-chart/ \
+    --dependency-update \
+    --set KUBESTELLAR_VERSION=$(git rev-parse --short HEAD) \
+    --kube-context $HOSTING_CONTEXT \
+    --set kubeflex-operator.install=false \
+    --set-json='ITSes=[{"name":"its1"}]' \
+    --set-json='WDSes=[{"name":"wds1"}]' \
+    --set verbosity.kubestellar=${KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY} \
+    --set verbosity.transport=${TRANSPORT_CONTROLLER_VERBOSITY}
+  fi
 popd
 
 wait-for-cmd "(kubectl --context '$HOSTING_CONTEXT' -n wds1-system wait --for=condition=Ready pod/\$(kubectl --context '$HOSTING_CONTEXT' -n wds1-system get pods -l name=transport-controller -o jsonpath='{.items[0].metadata.name}'))"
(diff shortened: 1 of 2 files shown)


test/e2e/common/setup-kubestellar.sh
```diff
@@ -109,14 +104,25 @@ esac
...
-pushd "${SRC_DIR}/../../.."
-make kind-load-image
-make install-local-core-chart \
-  INSTALL_KUBEFLEX=false \
-  ITS_NAME=its1 \
-  DEFAULT_WDS_NAME=wds1 \
-  KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY=$KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY
-popd
+if [ "$use_release" = true ] ; then
+  kubestellar_latest_release=$(curl -sL https://github.com/kubestellar/kubestellar/releases/latest | grep "</h1>" | tail -n 1 | sed -e 's/<[^>]*>//g' | xargs)
```
MikeSpreitzer, line 108: What we want is not the latest release on GitHub but rather the last release up through the commit that the user is using. See https://github.com/kubestellar/kubestellar/blob/5c18a50ccd0bbe7f2124b522a235215b7289bb66/test/e2e/README.md?plain=1#L7 . We do this by putting self-references in files. The list of self-references is tracked in https://docs.kubestellar.io/release-0.24.0/direct/release/#making-a-new-kubestellar-release . We could put a self-reference in this file (e.g., `KS_RELEASE=0.24.0`).

test/e2e/common/setup-kubestellar.sh
```diff
@@ -109,14 +104,25 @@ esac
...
+  helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart \
+    --version ${kubestellar_latest_release//v} \
```
MikeSpreitzer, line 110: That's a fragile way to strip off the leading "v" --- it actually deletes every "v". There is a better way.

```console
bash-3.2$ release=v0.24.1-preview.1

bash-3.2$ echo ${release//v}
0.24.1-preiew.1

bash-3.2$ echo ${release#v}
0.24.1-preview.1
```

test/e2e/common/setup-kubestellar.sh
```diff
@@ -109,14 +104,25 @@ esac
...
+    --set-json='WDSes=[{"name":"wds1"}]' \
+    --set verbosity.kubestellar=${KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY} \
+    --set verbosity.transport=${TRANSPORT_CONTROLLER_VERBOSITY}
+else
+  pushd "${SRC_DIR}/../../.."
+  make kind-load-image
+  make install-local-core-chart \
+    INSTALL_KUBEFLEX=false \
+    ITS_NAME=its1 \
+    DEFAULT_WDS_NAME=wds1 \
```
MikeSpreitzer, line 122: Did not originate here, should be fixed in a separate PR: this variable is named wrong. It is simply the WDS name, it does not hold a default that might get overridden somewhere.

test/e2e/common/setup-kubestellar.sh
francostellari: ok

test/e2e/common/setup-kubestellar.sh
francostellari: Yes it's an awkward name. I can fix in a separate PR.
I'm also wondering if there is a need to use the `make install-local-core-chart` instead of a direct helm command pointing to the local `core-chart/` folder.
The second would be more clear, I think... Then do we need the `install-local-core-chart` target, with the repeated arguments?

test/e2e/common/setup-kubestellar.sh
francostellari: While I understand your point, and the complications in digging the version out, I'm wondering if it is in reality moot in a real case scenario.

I was thinking that one would `run_test --released` basically only once after a new release (being alpha, preview, stable, etc) ... at that would be `latest` independently.

Do you envision a situation when one would like to `run_test --released` against v0.24.0 when you release `v0.24.1.alpha1`? In such caases I think we go back to my initial proposal where `--released [<ver>]` could take an optional version.

Finally, I think we should remove `/VERSION` since it is not maintained anymore.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -109,14 +104,25 @@ esac
...
+    KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY=$KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY
```
MikeSpreitzer, line 123: The `install-local-core-chart` target uses `KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY` to set _both_ the kubestellar controller-manager verbosity _and_ the OCM transport controller verbosity. This does not line up with the separate treatment in this script. IMHO the best resolution would be to update the `install-local-core-chart` target to be more articulate about verbosity controls.

francostellari: If you see my comment above, I was thinking that it is probably better to completely remove the target and just directly use helm with the correctly named args

francostellari: I implemented the change in #2473.
We will need to rebase.

test/e2e/common/setup-kubestellar.sh
francostellari: I implemented the change in https://github.com/kubestellar/kubestellar/pull/2473.
We will need to rebase.

test/e2e/common/setup-kubestellar.sh
francostellari: I removed the target in #2473

test/e2e/common/setup-kubestellar.sh
```diff
@@ -110,12 +106,24 @@ esac
...
 
 pushd "${SRC_DIR}/../../.."
-make kind-load-image
-make install-local-core-chart \
-  INSTALL_KUBEFLEX=false \
-  ITS_NAME=its1 \
-  DEFAULT_WDS_NAME=wds1 \
-  KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY=$KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY
+if [ "$use_release" = true ] ; then
+  kubestellar_release=$(sed -n -e 's/^.*KUBESTELLAR_VERSION: //p' "core-chart/values.yaml")
```
MikeSpreitzer, line 110: There is a much easier and more robust way:

```
mspreitz@mjs13 kubestellar % yq .KUBESTELLAR_VERSION core-chart/values.yaml 
0.24.0
```

Makefile
```diff
@@ -364,4 +364,4 @@ bin-dir:
 	mkdir -p bin
 
 
-include Makefile.venv
+include Makefile.venv
```
MikeSpreitzer, line 367: I do not think we need to remove the terminal linebreak here.

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #2483, opened 2024-10-07
✨ Finish implementing expanded transport ifc, simplify

### LINKED ISSUE #2332: bug: Kubestellar is not handling Jobs correctly
### Describe the bug

Environment:
k3d version v5.7.2
k3s version v1.29.6-k3s2 (default)

KUBESTELLAR_VERSION=0.23.0
OCM_STATUS_ADDON_VERSION=0.2.0-rc10
OCM_TRANSPORT_PLUGIN=0.1.11

Running with a hosting cluster (wds type=host) and two wecs. I submit a Job on the hosting cluster. It gets downsynced to the wec where it runs to completion. The status of the job on the hosting cluster is updated correctly. A job label matching binding policy is then deleted to force deletion of the job on the wec. 

The job on the wec is not deleted as expected.

Any subsequent jobs created on the hosting cluster never show on the wec. The manifestwork only contains the first/completed job. The bindings object only contains the new job.

When I delete klusterlet-agent on the wec, the old job gets deleted and the new one immediately starts. 
Attached is klusterlet-agent log and yamls for bindings and manifestworks
The completed job's name is  sample-job-scphw. The second job (which is stuck) is sample-job-t8tgh

[klusterlet-agent.log](https://github.com/user-attachments/files/16274665/klusterlet-agent.log)
[manifestworks.txt](https://github.com/user-attachments/files/16274667/manifestworks.txt)
[bindings.txt](https://github.com/user-attachments/files/16274669/bindings.txt)

Both the klusterlet-agent and manifestworks show errors. 

Paolo and Eddie debugged this problem with me and have an idea what the problem is. I will let them chime in to provide more details about why Kubestellar is failing to handle jobs. 

### Steps To Reproduce

See above

### Expected Behavior

Jobs should be deleted when a label matching bindingpolicy is deleted. Should be able run subsequent jobs.

### Want to contribute?

### Additional Context

_No response_

### LINKED ISSUE #2469: bug: Workload not deleted from a WEC after its label (matching BP) is deleted in WDS
### Describe the bug

My environment:
K3D 
Client Version: v1.31.0
Kustomize Version: v5.4.2
Server Version: v1.30.4+k3s1
Kubestallar version: 0.24.0 

I've configured  a job with a label matching BP of specific WEC. I've created it in WDS (type=host) and the job was pulled to the WEC and ran. On WDS, I've edited the job and removed the label matching BP. My expectation was that the job would be removed from the WEC. Instead, the job lingers and eventually after several minutes is cleaned up. The amount of time before the cleanup is done varies.

Another side effect of the problem is that any subsequent jobs created in WDS are not pulled to the WEC. 

The workaround is to use BP per workload which is not optimal. The BP deletion does cleanup the WEC.

### Steps To Reproduce

Create BP for a WEC (add   wantSingletonReportedState: true )
Add a label to a job which matches the BP
Create a job in WDS and observe its running and completing on the WEC
Edit the job in WDS (kubectl edit job ..) and delete the label that matches BP
Observe the job lingering on the WEC

### Expected Behavior

After deletion of the label, I expect the workload to be deleted in a timely manner (not several minutes) and subsequent jobs should be pulled and run on the WEC.

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR updates the OCM transport plugin to implement the expanded wrapping interface. And then cleans up the interface by removing the un-expanded wrapping method.

## Related issue(s)

Fixes #2469 
Fixes #2332 
This is a hack that alleviates the most pressing need to properly address #1851 .

### DIFF
--- a/test/e2e/ginkgo/multiple_cluster_deployment_test.go
+++ b/test/e2e/ginkgo/multiple_cluster_deployment_test.go
@@ -117,17 +117,15 @@ var _ = ginkgo.Describe("end to end testing", func() {
 			util.ValidateNumDeploymentReplicas(ctx, wec1, ns, 1)
 			util.ValidateNumDeploymentReplicas(ctx, wec2, ns, 1)
 
-			if false { // TODO: enable this after the transport plugin implements create-only
-				ginkgo.By("modifying the Deployment in the WDS and expecting no change in the WECs")
-				objPatch := []byte(`{"spec":{"replicas": 2}}`)
-				_, err := wds.AppsV1().Deployments(ns).Patch(
-					ctx, "nginx", types.MergePatchType, objPatch, metav1.PatchOptions{})
-				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			ginkgo.By("modifying the Deployment in the WDS and expecting no change in the WECs")
+			objPatch := []byte(`{"spec":{"replicas": 2}}`)
+			_, err := wds.AppsV1().Deployments(ns).Patch(
+				ctx, "nginx", types.MergePatchType, objPatch, metav1.PatchOptions{})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
-				time.Sleep(30 * time.Second)
-				util.ValidateNumDeploymentReplicas(ctx, wec1, ns, 1)
-				util.ValidateNumDeploymentReplicas(ctx, wec2, ns, 1)
-			}
+			time.Sleep(30 * time.Second)
+			gomega.Expect(util.GetNumDeploymentReplicas(ctx, wec1, ns)).To(gomega.Equal(1))
+			gomega.Expect(util.GetNumDeploymentReplicas(ctx, wec2, ns)).To(gomega.Equal(1))
 		})
 
 		ginkgo.It("handles changes in bindingpolicy ObjectSelector", func(ctx context.Context) {
--- a/test/util/util.go
+++ b/test/util/util.go
@@ -503,18 +503,29 @@ func ValidateNumManifestworks(ctx context.Context, ocmWorkIts *ocmWorkClient.Cli
 	}, timeout).Should(gomega.Equal(num))
 }
 
+const (
+	// GotNot means that the attempt to list Deployments failed
+	GotNoList int = -1
+	// GotNotExactly1 means that the number of Deployments is more or less than 1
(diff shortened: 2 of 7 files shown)


MikeSpreitzer:
In local testing, including the fix from #2489 seems to fix the test failure here too.

MikeSpreitzer:
The force-push to 05064454f is a rebase onto `main`.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
```diff
@@ -117,16 +117,16 @@ var _ = ginkgo.Describe("end to end testing", func() {
 			util.ValidateNumDeploymentReplicas(ctx, wec1, ns, 1)
 			util.ValidateNumDeploymentReplicas(ctx, wec2, ns, 1)
 
-			if false { // TODO: enable this after the transport plugin implements create-only
+			if true { // TODO: enable this after the transport plugin implements create-only
```
vMaroon, line 120: Perhaps should just completely remove.

MikeSpreitzer: Done. This the only thing changed in the force-push to 08960c7be .

test/util/util.go
```diff
@@ -503,6 +503,26 @@ func ValidateNumManifestworks(ctx context.Context, ocmWorkIts *ocmWorkClient.Cli
...
+func GetNumDeploymentReplicas(ctx context.Context, wec *kubernetes.Clientset, ns string) int {
+	ginkgo.GinkgoHelper()
+	var replicas int = -1
+	gomega.Eventually(func() error {
+		deployments, err := wec.AppsV1().Deployments(ns).List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+		if len(deployments.Items) != 1 {
+			replicas = -2
```
pdettori, line 515: What is the meaning of replica = -2  or -3 ? It would be helpful to use named constants for this values

MikeSpreitzer: There is no real meaning. These are just values that can not possibly come from `len(anything)`, so that someone debugging a failure gets a clue.

MikeSpreitzer: But I have pushed a revision that documents these values.

pdettori (approved):
/lgtm

PR #2485, opened 2024-10-07
📖 Note incompleteness of create-only

MikeSpreitzer:
## Summary
This PR updates the release notes to say that release 0.24.0 has an incomplete introduction of create-only and that every release (thus far) does not support Job objects properly.

## Related issue(s)

Proper release notes would probably have prevented @cwiklik from wasting his time testing release 0.24.0.


pdettori (approved):
/lgtm

PR #2489, opened 2024-10-09
🐛 Stop transport-controller thrashing finalizers

MikeSpreitzer:
## Summary
This PR changes the generic transport controller code so that it does not thrash the finalizers of the wrapped object.

## Related issue(s)

Fixes #


dumb0002:
@MikeSpreitzer, I tested this PR and it works as expected. Thanks!

PR #2493, opened 2024-10-10
✨ Change E2E setup to let core chart install kubeflex in hosting cluster

### LINKED ISSUE #2491: bug: Core Helm chart >0.24.0 does not deploy WDS controllers in a Kind cluster with an older version (<0.6.3) of KubeFlex controller
### Describe the bug

If a kind cluster was created using the command `kflex init` with a kflex version <0.6.3, then the helm chart will not be able to deploy the WDS controllers due to https://github.com/kubestellar/kubeflex/issues/278

This is a problematic situation especially for people that have tried kubeflex/kubestellar in the past and have an older version of kflex on their system and want to try the new Helm chart.  It's a very hard to debug problem for the user.

While the core Helm chart can install the correct version of KubeFlex automatically, it cannot override an existing -older- version of the controller in the cluster. 

### Steps To Reproduce

Install `kflex` CLI < 0.6.3
Run `test/e2e/run_test.sh` and observe that it fails.
Note that the `wds1` does not show controllers nor deployments in it

### Expected Behavior

Additional checks are needed.
Some way for the user to figure out this issue.

The issue arise from the combined use of `kflex init [-C]` and the `helm...` command where the versions don't match.
Perhaps we should discourage the the use of `kflex init` and let the core helm chart install the KubeFlex controller

### Want to contribute?

### Additional Context

/cc @MikeSpreitzer 
/cc @pdettori 
/cc @dumb0002

MikeSpreitzer:
## Summary
This PR changes the script that does setup for E2E testing to align it with what the Getting Started instructions tell users, which is to use `scripts/create-kind-cluster-with-SSL-passthrough.sh` to create the hosting cluster and let the core Helm chart install KubeFlex in the hosting cluster; formerly the E2E test setup used `kflex init` and had the core Helm chart _not_ install KubeFlex in the hosting cluster. This change is good because we want to regularly test the recipe that we tell users to do.

## Related issue(s)

This is also good because it moots #2491 . I am not saying this fully resolves #2491 , because we do not forbid users to use `kflex init` and we also have a dependency-checking script that does not check versions.

### DIFF
--- a/test/e2e/common/setup-kubestellar.sh
+++ b/test/e2e/common/setup-kubestellar.sh
@@ -85,17 +85,16 @@ source "$COMMON_SRCS/setup-shell.sh"
 
 :
 : -------------------------------------------------------------------------
-: Initialize KubeFlex, creating or using an existing hosting cluster.
+: Create the KubeFlex hosting cluster, if necessary.
 :
 case "$CLUSTER_SOURCE" in
     (kind)
-        kflex init --create-kind $disable_chatty_status
+        ${SRC_DIR}/../../../scripts/create-kind-cluster-with-SSL-passthrough.sh --name kubeflex
         : Kubeflex kind cluster created.
         ;;
     (existing)
         kubectl config use-context "$HOSTING_CONTEXT"
-        kflex init $disable_chatty_status
-        : KubeFlex initialized to use existing cluster in "$HOSTING_CONTEXT" context
+        : kubectl configured to use existing cluster in "$HOSTING_CONTEXT" context
         ;;
 esac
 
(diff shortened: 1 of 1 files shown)


test/e2e/common/setup-kubestellar.sh
```diff
@@ -110,18 +109,16 @@ if [ "$use_release" = true ] ; then
...
     --version $(yq .KUBESTELLAR_VERSION core-chart/values.yaml) \
     --kube-context $HOSTING_CONTEXT \
-    --set kubeflex-operator.install=false \
     --set-json='ITSes=[{"name":"its1"}]' \
     --set-json='WDSes=[{"name":"wds1"}]' \
     --set verbosity.kubestellar=${KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY} \
     --set verbosity.transport=${TRANSPORT_CONTROLLER_VERBOSITY}
 else
   make kind-load-image
+  helm dependency update core-chart/
```
MikeSpreitzer, line 118: I first tested without this line, expecting the `--dependency-update` flag to get the job done. But it did not. I had the 0.6.2 kubeflex chart in core-chart/charts, and it did not get replaced with 0.6.3 as it should. Note that the doc at https://helm.sh/docs/helm/helm_upgrade/ for that flag says that it updates dependencies **IF THEY ARE MISSING** (which I find a bit ambiguous, but experimentation resolved it). I tested with Helm version 3.16.1.

francostellari: @MikeSpreitzer I had the same experience myself. The flag does not seem to check the chart version. I'm not sure if it is a bug or the desired behavior. You can add a command, at least in the test script, to remove charts from `charts/`.
Overall I think this is the least evil.

dumb0002:
a) I tested this PR in a env with an old Kubeflex version installed: `v0.6.1`
b) the e2e test in this PR deployed the correct version for Kubeflex: 

```
k -n kubeflex-system logs kubeflex-controller-manager-749f8c7c6b-twshk
2024-10-10T14:51:54Z	INFO	controller-runtime.metrics	Metrics server is starting to listen	{"addr": "127.0.0.1:8080"}
2024-10-10T14:51:54Z	INFO	setup	starting manager	{"version": "v0.6.3.672cc8a"}
2024-10-10T14:51:54Z	INFO	Starting server	{"kind": "health probe", "addr": "[::]:8081"}
2024-10-10T14:51:54Z	INFO	starting server	{"path": "/metrics", "kind": "metrics", "addr": "127.0.0.1:8080"}
I1010 14:51:54.339982       1 leaderelection.go:250] attempting to acquire leader lease kubeflex-system/c6f71c85.kflex.kubestellar.org...
I1010 14:51:54.349438       1 leaderelection.go:260] successfully acquired lease kubeflex-system/c6f71c85.kflex.kubestellar.org
```

dumb0002:
@MikeSpreitzer, everything works as expected. Thanks!

/lgtm

PR #2495, opened 2024-10-10
📖 Improve doc linkage

MikeSpreitzer:
## Summary
This PR makes some organizational changes to the documentation: adding a few links, removing the "Contribution Notes" vertex in the hierarchy (promoting its children), recognizing that the "For Contributors" doc is really just about testing.

Preview of website part at https://mikespreitzer.github.io/kcp-edge-mc/doc-fixlinx-1010/readme/

## Related issue(s)

Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ KubeStellar simplifies this process by allowing developers to define a binding p
 
 ## Website
 
-See [kubestellar.io](https://kubestellar.io).
+For usage, architecture, and other documentation, see [the website](https://kubestellar.io).
 
 ## Contributing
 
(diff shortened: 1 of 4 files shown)


KPRoche:
Should that "getting started" section maybe have a second sentence and link to the "full" user guide?

And I realize I omitted adding a link in the text when I wrote this originally because I presumed this readme would be read as part of the website and the get started link exists in the masthead menu. I agree it should be in the text also.

MikeSpreitzer:
> Should that "getting started" section maybe have a second sentence and link to the "full" user guide?

Remember that 'the "full" user guide' is a branch in the website's hierarchy of pages, not a single page. There _is_ an overview page, and this section _could_ link to that.

Why only link to the user guide? Why not also link to the real architecture doc? It seems presumptuous to only link to one thing.

I first wrote a remark making a broad reference, but it looked stupid. I said basically "look around the website you are in", and that seems like telling people to have common sense.

pdettori (approved):
/lgtm

PR #2501, opened 2024-10-12
🐛  Check for empty DownsyncObjectTest

### LINKED ISSUE #2500: feature: `DownsyncObjectTest{}` should be rejected
### Feature Description

I got a problem report from a confused user who accidentally constructed a `DownsyncObjectTest` that has zero tests in it. In other words, the source was empty. The consequence is that it matched every matchable object in the WDS. I expect that this is never, or very nearly never, the desired functionality. We should make an empty `DownsyncObjectTest` invalid. If someone really does want to match every matchable object then they can explicitly write a `"*"` in there somewhere.

This is probably a duplicate of a very old Issue, but I am having trouble finding it. Perhaps it was one of many closed en masse when we moved away from the legacy system.

### Proposed Solution

Add a validation criterion to the CRD for `BindingPolicy` that forbids an empty `DownsyncObjectTest`.

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR adds validation checking to BindingPolicy regarding the prohibition on an empty DownsyncObjectTest, and updates the statement in the API about how validation errors are reported.

This PR adds this validation in two ways, because the first way does not work and I do not understand why. Hopefully it is some small thing that can be rectified, which is why I have not abandoned it. The first technique adds some validation to the JSON schema for BindingPolicy, but the kube-apiserver accepts this extra validation without complaint but also trashes it so that it has no effect (it empties out the schemas in the array value of the `anyOf`, so they all trivially validate; the integration test displays the returned CRD, where you can see the trashed schemas). This PR has a second commit which adds checking in the binding controller.

## Related issue(s)

Fixes #2500

### DIFF
--- a/test/integration/controller-manager/cr-validation_test.go
+++ b/test/integration/controller-manager/cr-validation_test.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2024 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmtest
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+
+	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/klog/v2/ktesting"
+	kastesting "k8s.io/kubernetes/cmd/kube-apiserver/app/testing"
+	"k8s.io/kubernetes/test/integration/framework"
+	"sigs.k8s.io/yaml"
+
+	ksapi "github.com/kubestellar/kubestellar/api/control/v1alpha1"
+	kscrd "github.com/kubestellar/kubestellar/pkg/crd"
+)
+
+func TestBindingPolicyValidation(t *testing.T) {
+	testWriter := framework.NewTBWriter(t)
+	logger, ctx := ktesting.NewTestContext(t)
+	logger.Info("Starting etcd server")
+	framework.StartEtcd(t, testWriter)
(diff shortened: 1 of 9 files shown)


pkg/binding/binding.go
```diff
@@ -69,6 +69,8 @@ func (c *Controller) syncBinding(ctx context.Context, bindingName string) error
 		return fmt.Errorf("failed to get BindingPolicy from informer cache (name=%v): %w", bindingName, policyErr)
 	}
 
+	policyErrors := extraValidateBindingPolicy(policy)
```
vMaroon, line 72: I missed the changes that follow this line (in previously merged PRs). Managing a BP in this call violates the design of objects being updated only if their keys are queued and picked by a worker.

I think all the parts that perform checks and updates on BPs should be moved to `Controller::SyncBindingPolicy` and instead this code should queue the key when necessary. Otherwise, looks good.

MikeSpreitzer: Yes, I definitely agree that some refactoring is in order. I am not so sure that it will be easy until we fully address a larger issue: making BindingPolicy just one way to manage a Binding. In any case, yes, for later PRs.

MikeSpreitzer:
The force-push to adabae73b fixes the validation so that it works, undoes the addition of checking to the binding controller, and adds a note about this prohibition to the user-facing doc (which was not wrong, because it refers the reader to the API, which already states the prohibition).

pkg/binding/binding.go
vMaroon: Would be good to track in an issue.

MikeSpreitzer: #2502

test/integration/controller-manager/cr-validation_test.go
```diff
@@ -0,0 +1,126 @@
...
+	dynIfc := dynClient.Resource(ksapi.SchemeGroupVersion.WithResource("bindingpolicies"))
+	converter := runtime.DefaultUnstructuredConverter
+	for _, testCase := range []struct {
+		name     string
+		specJSON string
+		expectOK bool
+	}{
+		{name: "junk-field-in-spec", specJSON: `{"junk": 1}`, expectOK: false},
+		{name: "junk-field-in-policy", specJSON: `{"downsync": [{"junq": true}]}`},
+		{name: "no-test-in-policy", specJSON: `{"downsync": [{"createOnly": true, "statusCollection": {"statusCollectors": ["phred"]}}]}`},
```
waltforme, line 94: Need some help to understand: Why no `expectOK` for these two test cases?

waltforme:
We may need to update docs/content/direct/contributor.md to mention the new integration tests.

test/integration/controller-manager/cr-validation_test.go
MikeSpreitzer: It is not necessary to specify every field in a struct constructor. You can omit a field when its zero value is the value you want.

waltforme: I see. It might be more consistent with line 92 (the first test case) --- not a blocker.

PR #2515, opened 2024-10-21
📖  Add doc for getting started from OCM

MikeSpreitzer:
## Summary
This PR adds a document describing how to install and use KubeStellar after having done the OCM Quick Start.

This PR also has some doc fixes and enhancements noticed along the way.

Note: when testing this, you will have to get the modified OCM Quick Start script from https://github.com/MikeSpreitzer/kcp-edge-mc/blob/doc-start-from-ocm/scripts/ocm-local-up-for-ingress.sh , because it has not merged into `main` yet.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-start-from-ocm/direct/start-from-ocm/ .

I plan to include a little more doc updating here, to factor the general outline so that it has _create_ and _register_ steps for both ITS and WDS, to recognize the possiblity of using the host cluster as either of those.

## Related issue(s)

Fixes #


docs/content/direct/start-from-ocm.md
```diff
@@ -0,0 +1,143 @@
...
+```shell
+export KUBESTELLAR_VERSION={{ config.ks_latest_release }}
+```
+
+### OCM Quick Start with Ingress
+
+This recipe uses a modified version of [the OCM Quick Start](https://raw.githubusercontent.com/open-cluster-management-io/OCM/main/solutions/setup-dev-environment/local-up.sh). The modification is because KubeStellar needs the hosting cluster to have an Ingress controller with SSL passthrough enabled. The modified Quick Start script has the following modifications compared to the baseline.
+
+1. The `kind` cluster created for the hub has an additional port mapping, where the Ingress controller listens.
+1. The script installs an Ingress controller (based on nginx) into the hub cluster, then patches it to enable SSL passthrough, and later waits for it to be in service.
```
pdettori, line 59: best just say "NGINX Ingress Controller" which is well known (see https://docs.nginx.com/nginx-ingress-controller/ ) rather than "Ingress controller (based on nginx) "

docs/content/direct/start-from-ocm.md
```diff
@@ -0,0 +1,143 @@
...
+kubectl --context kind-hub label managedcluster cluster1 location-group=edge name=cluster1
+kubectl --context kind-hub label managedcluster cluster2 location-group=edge name=cluster2
+```
+
+### Use Core Helm chart to initialize KubeFlex, recognize ITS, and create WDS
+
+This chart instance will do the following.
+
+- Install KubeFlex in the hosting cluster.
+- Tell that the hosting cluster will play the role of an ITS named "its1".
```
pdettori, line 89: s/Tell that the hosting cluster will play the role of an ITS named "its1"./Assign the hosting cluster the role of an ITS, named 'its1'."/

docs/content/direct/start-from-ocm.md
```diff
@@ -0,0 +1,143 @@
...
+
+```shell
+while [ -z "$(kubectl --context wds1 get crd bindingpolicies.control.kubestellar.io --no-headers -o name 2> /dev/null)" ] ;  do
+    sleep 5
+done
+kubectl --context wds1 wait --for condition=Established crd bindingpolicies.control.kubestellar.io
+```
+
+Proceed to [Scenario 1 (multi-cluster workload deployment with kubectl) in the example scenarios](example-scenarios.md#scenario-1-multi-cluster-workload-deployment-with-kubectl) _after_ defining the shell variables that characterize the setup done above. Following are setting for those variables, whose meanings are defined [at the start of the example scenarios document](example-scenarios.md#assumptions-and-variables).
+
```
pdettori, line 127: s/setting/settings/

docs/content/direct/start-from-ocm.md
MikeSpreitzer: Done, thanks.

docs/content/direct/start-from-ocm.md
MikeSpreitzer: Done.

docs/content/direct/start-from-ocm.md
MikeSpreitzer: Fixed.

MikeSpreitzer:
The force-push to 7d2c6ed69 makes the changes indicated in the review above.

eaepstein:
Using the documentation in Main, it all worked for me. However, there are two places in the documentation that could be better.

Place 1:
`That command will print some notes about how to get a kubeconfig "context" named "wds1" defined. Do that, because this context is used in the steps that follow. The notes assume that your current kubeconfig context is the one where the Helm chart was installed, which is not necessarily true --- so take care for that too.
`

Confuses me, particularly the "Do that" and "so take care for that too". 

Much better for me would be if this paragraph just said "Ignore the notes printed from the helm command and continue following directions below."

Place 2: Another confusion is nearer the end of the page where it says:

Proceed to [Scenario 1 (multi-cluster workload deployment with kubectl) in the example scenarios](https://mikespreitzer.github.io/kcp-edge-mc/doc-start-from-ocm/direct/example-scenarios/#scenario-1-multi-cluster-workload-deployment-with-kubectl) after defining the shell variables ...

What would work better for me would be:

In order to run the example scenarios, first define the following shell variables:
```
host_context=kind-hub
its_cp=its1
...
```
Then proceed to Scenario 1 (multi-cluster workload deployment with kubectl) ...

clubanderson:
I agree, there is no longer a need to add instructions like these. It has become confusing because most places where the package is used have a wrapper text around it from the script that calls it.

PR #2518, opened 2024-10-24
updated demo env script to a point where it works consistently for 0.25

clubanderson:
## Related issue(s)

Fixes #

### DIFF
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -17,16 +17,27 @@
 
 set -e
 
-kubestellar_version=0.25.0-rc.1
+export kubestellar_version=0.25.0-rc.1
+echo -e "KubeStellar Version: ${kubestellar_version}"
 
 echo -e "Checking that pre-req softwares are installed..."
 curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/hack/check_pre_req.sh | bash -s -- -V kflex ocm helm kubectl docker kind
 
-output=$(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/hack/check_pre_req.sh | bash -s -- -A -V kflex ocm helm kubectl docker kind
-)
+output=$(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/hack/check_pre_req.sh | bash -s -- -A -V kflex ocm helm kubectl docker kind)
 
-echo -e "\nStarting environment clean up..."
-echo -e "Starting cluster clean up..."
+##########################################
+function wait-for-cmd() (
+    cmd="$@"
+    wait_counter=0
+    while ! (eval "$cmd") ; do
+        if (($wait_counter > 100)); then
+            echo "Failed to ${cmd}."
+            exit 1
+        fi
+        ((wait_counter += 1))
+        sleep 5
+    done
+)
 
 cluster_clean_up() {
     error_message=$(eval "$1" 2>&1)
@@ -35,13 +46,6 @@ cluster_clean_up() {
         echo "$error_message"
     fi
 }
-
-cluster_clean_up "kind delete cluster --name kubeflex"
-cluster_clean_up "kind delete cluster --name cluster1"
-cluster_clean_up "kind delete cluster --name cluster2"
-echo -e "Cluster space clean up has been completed"
-
-echo -e "\nStarting context clean up..."
 
 context_clean_up() {
     output=$(kubectl config get-contexts -o name)
@@ -72,45 +76,6 @@ context_clean_up() {
     done <<< "$output"
 }
 
-context_clean_up
-echo "Context space clean up completed"
-
-echo -e "\nStarting the process to install KubeStellar core: kind-kubeflex..."
-
-curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/scripts/create-kind-cluster-with-SSL-passthrough.sh | bash -s -- --name kubeflex --port 9443
-
-helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart --version $kubestellar_version --set-json='ITSes=[{"name":"its1"}]' --set-json='WDSes=[{"name":"wds1"}]'
-
-echo -e "\nWaiting for new non-host KubeFlex Control Planes to be Ready:"
-for cpname in its1 wds1; do
-  while [[ `kubectl get cp $cpname -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}'` != "True" ]]; do
-    echo "Waiting for \"$cpname\"..."
-    sleep 5
-  done
-  echo "\"$cpname\" is ready."
-done
-
-kubectl config delete-context its1 || true
-kflex ctx its1
-kubectl config delete-context wds1 || true
-kflex ctx wds1
-kflex ctx
-
-echo -e "\nCreating cluster and context for cluster 1 and 2..."
-
-# this demo environment will make sure of kind cluster and not Openshift cluster
-# : set flags to "" if you have installed KubeStellar on an OpenShift cluster
-
-flags="--force-internal-endpoint-lookup"
-clusters=(cluster1 cluster2);
-for cluster in "${clusters[@]}"; do
-   kind create cluster --name ${cluster}
-   kubectl config rename-context kind-${cluster} ${cluster}
-   clusteradm --context its1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk '{print $0 " --context '${cluster}' --singleton '${flags}'"}' | sh
-done
-
-echo -e "Checking that the CSR for cluster 1 and 2 appears..."
-
 checking_cluster() {
     found=false
     
@@ -139,9 +104,59 @@ checking_cluster() {
 
     done
 }
+##########################################
+
+echo -e "\nStarting environment clean up..."
+echo -e "Starting cluster clean up..."
+
+cluster_clean_up "kind delete cluster --name kubeflex"
+cluster_clean_up "kind delete cluster --name cluster1"
+cluster_clean_up "kind delete cluster --name cluster2"
+echo -e "Cluster space clean up has been completed"
+
+echo -e "\nStarting context clean up..."
+
+context_clean_up
+echo "Context space clean up completed"
+
+echo -e "\nStarting the process to install KubeStellar core: kind-kubeflex..."
+
+curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${KUBESTELLAR_VERSION}/scripts/create-kind-cluster-with-SSL-passthrough.sh | bash -s -- --name kubeflex --port 9443
+
+helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart \
+    --version $KUBESTELLAR_VERSION \
+    --set-json='ITSes=[{"name":"its1"}]' \
+    --set-json='WDSes=[{"name":"wds1"}]' \
+    --set-json='verbosity.default=5'
+
+kubectl config delete-context its1 || true
+kflex ctx its1
+kubectl config delete-context wds1 || true
+kflex ctx wds1
+
+# switch back to its1 context to crete 2 remote clusters, add OCM agents to them, and register them back to the KubeStellar core
+kflex ctx its1
+
+echo -e "\nWaiting for OCM cluster manager to be ready..."
+
+wait-for-cmd "[[ \$(kubectl --context its1 get deployments.apps -n open-cluster-management -o jsonpath='{.status.readyReplicas}' cluster-manager 2>/dev/null) -ge 1 ]]"
+echo -e "OCM cluster manager is ready"
+
+echo -e "\nCreating cluster and context for cluster 1 and 2..."
+
+: set flags to "" if you have installed KubeStellar on an OpenShift cluster
+flags="--force-internal-endpoint-lookup"
+clusters=(cluster1 cluster2);
+for cluster in "${clusters[@]}"; do
+   kind create cluster --name ${cluster}
+   kubectl config rename-context kind-${cluster} ${cluster}
+   clusteradm --context its1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk '{print $0 " --context '${cluster}' --singleton '${flags}'"}' | sh
+done
+
+echo -e "Checking that the CSR for cluster 1 and 2 appears..."
 
 echo""
-echo "Approving CSR for cluster1 and cluster2..."
+echo "Waiting for cluster1 and cluster2 to be ready and then approve their CSRs"
 checking_cluster cluster1
 checking_cluster cluster2
 
@@ -150,3 +165,4 @@ echo "Checking the new clusters are in the OCM inventory and label them"
 kubectl --context its1 get managedclusters
 kubectl --context its1 label managedcluster cluster1 location-group=edge name=cluster1
 kubectl --context its1 label managedcluster cluster2 location-group=edge name=cluster2
+


scripts/create-kubestellar-demo-env.sh
```diff
@@ -17,16 +17,27 @@
 
 set -e
 
-kubestellar_version=0.25.0-rc.1
+export kubestellar_version=0.25.0-rc.1
```
MikeSpreitzer, line 20: Why the "export"? If it is not needed then I think that it is positively harmful to include it because that makes the reader unsure of whether any invoked executable depends on that variable.

clubanderson: can be removed I guess

MikeSpreitzer: Removing it would be my recommendation.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -139,9 +104,83 @@ checking_cluster() {
 
     done
 }
+##########################################
+
+echo -e "\nStarting environment clean up..."
+echo -e "Starting cluster clean up..."
+
```
MikeSpreitzer, line 111: I would include https://github.com/kubestellar/kubestellar/blob/b68baf3a590729e52ab3f1c4fc56efd9bf381068/docs/content/direct/get-started.md?plain=1#L35 here, for the same reason that it is there.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -139,9 +104,83 @@ checking_cluster() {
...
+     kind create cluster --name "${cluster}" >/dev/null 2>&1 &&
+     kubectl config rename-context "kind-${cluster}" "${cluster}" >/dev/null 2>&1
+     echo -e "${cluster} creation and context setup complete"
+   ) &
+done
+
+wait 
+
+echo -e "Creating KubeFlex cluster with SSL Passthrough"
+curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${KUBESTELLAR_VERSION}/scripts/create-kind-cluster-with-SSL-passthrough.sh | bash -s -- --name kubeflex --port 9443
```
MikeSpreitzer, line 138: The capitalization of the environment variable name needs to be fixed here and in the later places where it is still uppercase.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -17,16 +17,27 @@
 
 set -e
```
MikeSpreitzer, line 19: I strongly recommend including `set -x`, either unconditionally or controlled by a command-line option. This script is likely to be used by newbies, and we will get bug reports, and those will be much more comprehensible with the commands echoed.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -17,16 +17,27 @@
...
-kubestellar_version=0.25.0-rc.1
+export kubestellar_version=0.25.0-rc.1
+echo -e "KubeStellar Version: ${kubestellar_version}"
 
 echo -e "Checking that pre-req softwares are installed..."
 curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/hack/check_pre_req.sh | bash -s -- -V kflex ocm helm kubectl docker kind
 
-output=$(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/hack/check_pre_req.sh | bash -s -- -A -V kflex ocm helm kubectl docker kind
-)
+output=$(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/hack/check_pre_req.sh | bash -s -- -A -V kflex ocm helm kubectl docker kind)
```
MikeSpreitzer, line 26: Why is this curl-to-bash being done a second time? Nothing is done with the content of `output`, right?

scripts/create-kubestellar-demo-env.sh
```diff
@@ -139,9 +104,83 @@ checking_cluster() {
...
+
+wait 
+
+echo -e "Creating KubeFlex cluster with SSL Passthrough"
+curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${KUBESTELLAR_VERSION}/scripts/create-kind-cluster-with-SSL-passthrough.sh | bash -s -- --name kubeflex --port 9443 
+# >/dev/null 2>&1
+
+echo -e "Completed KubeFlex cluster with SSL Passthrough"
+
+(docker pull ghcr.io/loft-sh/vcluster:0.16.4 && kind load docker-image ghcr.io/loft-sh/vcluster:0.16.4 --name kubeflex) &
```
MikeSpreitzer, line 143: Here is a more direct way to get that done:

```console
mspreitz@mjs13 kubestellar % docker exec kubeflex-control-plane crictl pull ghcr.io/loft-sh/vcluster:0.16.4
Image is up to date for sha256:00428133d55e8b3f1b699f390bc4f0dd79e3a2635aad1d0604f6c8df09803166
```

scripts/create-kubestellar-demo-env.sh
```diff
@@ -139,9 +104,83 @@ checking_cluster() {
...
+curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${KUBESTELLAR_VERSION}/scripts/create-kind-cluster-with-SSL-passthrough.sh | bash -s -- --name kubeflex --port 9443 
+# >/dev/null 2>&1
+
+echo -e "Completed KubeFlex cluster with SSL Passthrough"
+
+(docker pull ghcr.io/loft-sh/vcluster:0.16.4 && kind load docker-image ghcr.io/loft-sh/vcluster:0.16.4 --name kubeflex) & 
+(docker pull rancher/k3s:v1.27.2-k3s1 && kind load docker-image rancher/k3s:v1.27.2-k3s1 --name kubeflex) &
+(docker pull quay.io/open-cluster-management/registration-operator:v0.13.2 && kind load docker-image quay.io/open-cluster-management/registration-operator:v0.13.2 --name kubeflex) & 
+
+# curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${KUBESTELLAR_VERSION}/scripts/create-kind-cluster-with-SSL-passthrough.sh | bash -s -- --name kubeflex --port 9443
```
MikeSpreitzer, line 147: This comment can be omitted, right?

scripts/create-kubestellar-demo-env.sh
```diff
@@ -139,9 +104,83 @@ checking_cluster() {
...
+
+
+helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart \
+    --version $KUBESTELLAR_VERSION \
+    --set-json='ITSes=[{"name":"its1"}]' \
+    --set-json='WDSes=[{"name":"wds1"}]' \
+    --set-json='verbosity.default=5'
+
+kubectl config delete-context its1 || true
```
MikeSpreitzer, line 156: This was done earlier in context_clean_up, right?

scripts/create-kubestellar-demo-env.sh
```diff
@@ -139,9 +104,83 @@ checking_cluster() {
...
+kflex ctx its1
+kubectl config delete-context wds1 || true
+kflex ctx wds1
+
+# switch back to its1 context to crete 2 remote clusters, add OCM agents to them, and register them back to the KubeStellar core
```
MikeSpreitzer, line 161: s/crete/create/ but, more importantly, that was already done.
Also, the commands that follow do not care what the kubeconfig current context is, they all explicitly specify a context (which is a good thing, I think we have agreed).

scripts/create-kubestellar-demo-env.sh
```diff
@@ -150,3 +189,4 @@ echo "Checking the new clusters are in the OCM inventory and label them"
 kubectl --context its1 get managedclusters
 kubectl --context its1 label managedcluster cluster1 location-group=edge name=cluster1
 kubectl --context its1 label managedcluster cluster2 location-group=edge name=cluster2
+
```
MikeSpreitzer, line 192: Maybe a "done" message is appropriate here?

scripts/create-kubestellar-demo-env.sh
```diff
@@ -139,9 +104,83 @@ checking_cluster() {
...
+    --set-json='WDSes=[{"name":"wds1"}]' \
+    --set-json='verbosity.default=5'
+
+kubectl config delete-context its1 || true
+kflex ctx its1
+kubectl config delete-context wds1 || true
+kflex ctx wds1
+
+# switch back to its1 context to crete 2 remote clusters, add OCM agents to them, and register them back to the KubeStellar core
+kflex ctx its1
```
MikeSpreitzer, line 162: The commands that follow do not care what the kubeconfig current context is, they all explicitly specify a context (which is a good thing, I think we have agreed). So I think this command can simply be removed. I have tested a modified version of this script that omits this command, and it worked.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: Parallelization demo:

```console
mspreitz@mjs13 test4 % docker exec kubeflex-control-plane bash -c "echo a & echo b"
b
a
```

MikeSpreitzer:
I rebased onto #2521, removed the commands discussed in my review comments above, and tested. My modified version succeeded, and took 336 seconds on my laptop.

scripts/create-kubestellar-demo-env.sh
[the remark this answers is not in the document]
clubanderson: Roger that - included now

clubanderson: Roger that - included now

scripts/create-kubestellar-demo-env.sh
clubanderson: cleaned up

scripts/create-kubestellar-demo-env.sh
clubanderson: deleted

scripts/create-kubestellar-demo-env.sh
clubanderson: this way is much slower than what I have in the script now. keeping the way I have it

scripts/create-kubestellar-demo-env.sh
clubanderson: fixed

scripts/create-kubestellar-demo-env.sh
clubanderson: included

scripts/create-kubestellar-demo-env.sh
clubanderson: removed

scripts/create-kubestellar-demo-env.sh
clubanderson: removed

scripts/create-kubestellar-demo-env.sh
clubanderson: too noisy - can be done by hand if needed

scripts/create-kubestellar-demo-env.sh
clubanderson: ive tested without this and it does not allow the OCM bits to work properly - leaving these commands in

MikeSpreitzer:
And you no longer need to worry about the `ko` goofup. https://github.com/kubestellar/kubestellar/pull/2521#issuecomment-2436832307

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: That is labor intensive. That is why our other scripts have a `-X` option on the command line.

Not a blocker.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: Ah, the existing way uses a cache on the host machine.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: You have re-ordered the first two context establishment commands and eliminated the third (as it would clearly be redundant).

scripts/create-kubestellar-demo-env.sh
```diff
@@ -134,14 +97,85 @@ checking_cluster() {
...
+echo -e "\nPulling container images local..."
+images=("ghcr.io/loft-sh/vcluster:0.16.4"
+        "rancher/k3s:v1.27.2-k3s1"
+        "quay.io/open-cluster-management/registration-operator:v0.13.2"
+        "docker.io/bitnami/postgresql:16.0.0-debian-11-r13")
+
+for image in "${images[@]}"; do
+    (
+        docker pull "$image" && kind load docker-image "$image" --name kubeflex
+    ) &
```
MikeSpreitzer, line 147: FYI: Because _all_ of these are forked, there is no wait for any of them to complete before starting the `helm upgrade`. Is that what you intend?

Here is what I see in my typescript from around that time:

```
...
Completed KubeFlex cluster with SSL Passthrough

Pulling container images local...
Release "ks-core" does not exist. Installing it now.
Pulled: ghcr.io/kubestellar/kubestellar/core-chart:0.25.0-rc.1
Digest: sha256:7b50ef982b212a94a77019fc6e72ee56bd55c960490b27a813ab00c60dcb1f67
16.0.0-debian-11-r13: Pulling from bitnami/postgresql
v1.27.2-k3s1: Pulling from rancher/k3s
Digest: sha256:3331ad89ba2d1af68e36521724440638be3834978ac8288c49e54929357143e6
Status: Image is up to date for bitnami/postgresql:16.0.0-debian-11-r13
...
```

scripts/create-kubestellar-demo-env.sh
```diff
@@ -134,14 +97,85 @@ checking_cluster() {
...
+done
+
```
MikeSpreitzer, line 149: FYI, I made a modified version of this script with an added `wait` here, to get all the prefetching done before starting the chart instantiation. After warming up the caches, in two consecutive trials I got the following timings.

```
scripts/create-kubestellar-demo-env.sh  6.40s user 6.90s system 4% cpu 5:22.06 total
scripts/create-kubestellar-demo-env.sh  6.38s user 6.74s system 3% cpu 5:45.40 total
```

Without the `wait`, the timings are a bit quicker:

```
scripts/create-kubestellar-demo-env.sh  6.18s user 6.71s system 4% cpu 5:10.15 total
scripts/create-kubestellar-demo-env.sh  6.33s user 6.98s system 4% cpu 5:25.90 total
```

scripts/create-kubestellar-demo-env.sh
```diff
@@ -134,14 +97,85 @@ checking_cluster() {
...
+helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart \
+    --version $kubestellar_version \
+    --set-json='ITSes=[{"name":"its1"}]' \
+    --set-json='WDSes=[{"name":"wds1"}]' \
+    --set-json='verbosity.default=5'
+
+kubectl config delete-context wds1 || true
```
MikeSpreitzer, line 156: I made a modified version of this script with the `kubectl config delete-context wds1 || true` and `kubectl config delete-context its1 || true` commands deleted, and tested. Works fine. These commands are not needed because they are redundant with the context deletions done above in the `context_clean_up` function.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -134,14 +97,85 @@ checking_cluster() {
...
+context_clean_up
+echo "Context space clean up completed"
+
+echo -e "\nStarting the process to install KubeStellar core: kind-kubeflex..."
+clusters=(cluster1 cluster2)
+for cluster in "${clusters[@]}"; do
+   (
+     echo -e "Creating cluster ${cluster}..."
+     kind create cluster --name "${cluster}" >/dev/null 2>&1 &&
+     kubectl config rename-context "kind-${cluster}" "${cluster}" >/dev/null 2>&1
```
MikeSpreitzer, line 128: I would not run two file-modifying commands concurrently.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -134,14 +97,85 @@ checking_cluster() {
...
+echo -e "\nStarting context clean up..."
+context_clean_up
+echo "Context space clean up completed"
+
+echo -e "\nStarting the process to install KubeStellar core: kind-kubeflex..."
+clusters=(cluster1 cluster2)
+for cluster in "${clusters[@]}"; do
+   (
+     echo -e "Creating cluster ${cluster}..."
+     kind create cluster --name "${cluster}" >/dev/null 2>&1 &&
```
MikeSpreitzer, line 127: The `>/devnull 2>&1`means that it is going to be hard for anyone to debug if anything goes wrong.

MikeSpreitzer, line 127: I would not assume that the concurrent updates to my kubeconfig implied here are safe. So I asked. Sean McGinnis told me that they are. See https://kubernetes.slack.com/archives/CEKK1KTN2/p1729873143785879

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: Benjamin Elder replied in https://kubernetes.slack.com/archives/C2GL57FJ4/p1729873069405909, saying that `kubectl config` _does_ serialize its modifications to your kubeconfig file.

scripts/create-kubestellar-demo-env.sh
clubanderson: yes, I initially went with this with normal output however, because the commands run in parallel, to save time, the output did not make sense anyway - it was just repeating each line (1 for each cluster). I think this is just a preference in the end, they are vanilla kind clusters. if this fails, we would be able to reproduce via cmd line

scripts/create-kubestellar-demo-env.sh
clubanderson: it actually does not seem to matter, the creation of the core chart starts and takes much longer to get to the containers that are being fetched. On top of this, the pre-fetch that takes the longest is OCM. I put the commands here because I knew it was the first opportunity to fetch and then upload them to kind-kubeflex, and by the time they were done uploading to kind from docker, they would be plenty of time before they were needed. There are more that could be pre-fetched, but those are cared for by the helm chart which is why it takes a long time to install.

scripts/create-kubestellar-demo-env.sh
clubanderson: I wouldn't wait, the ks-core is loading many other container images that takes about 3 to 4 minutes. This allows plenty of time for the 4 pre-fetches to complete. If you wait for the pre-fetch to complete, then your just delaying ks-core which does not depend immediately on any of the pre-fetches

scripts/create-kubestellar-demo-env.sh
clubanderson: I will test again with them removed, but I have had mixed results in the past. I know that it should not affect the run, but I suspect it has.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: Indeed. I thought it worth documenting the actual results.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: The full complexity of the behavior of this script would be practically impossible to reproduce via command line.

MikeSpreitzer (approved):
I think that follow-on work is warranted, but I can approve this as progress (and want to reduce its cross-section wrt other PRs in flight).

/lgtm
/approve
Thanks!

scripts/create-kubestellar-demo-env.sh
clubanderson: looks like removing the commands worked on my latest run - will check in this update in a bit

clubanderson: check in removal in latest commit

PR #2521, opened 2024-10-24
🌱 Pin ko to release that has binary assets

MikeSpreitzer:
## Summary
This PR updates the GitHub Actions workflows that use `ko` to specify release 0.16 of `ko`. This is good because they made a newer release today but it does not have the binary assets that we need and I am not confident that this will be remedied soon.

## Related issue(s)

Fixes #


MikeSpreitzer:
The `ko` team fixed their problem, now the latest `ko` release has binary artifacts and we do not need to pin the `ko` version.

PR #2522, opened 2024-10-25
📖  Use explicit releases instead of main

MikeSpreitzer:
## Summary
This PR is a bit of cleanup after #2515 . This PR does two things.

1. Use an explicit version in the reference to the OCM Quick Start.
2. Note that the upcoming release should replace the reference to KubeStellar `main` with that release.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-release-for-ocm/direct/release/

## Related issue(s)

Fixes #


pdettori (commented):
/lgtm

PR #2523, opened 2024-10-25
📖  Use shell variable rather than environment variable

MikeSpreitzer:
## Summary
This PR updates the two setup recipes to use a variable that is local to the shell rather than exported to processes invoked in the recipes, to hold the KubeStellar release string. This is good because it does not leave the reader wondering whether those processes are sensitive to such an environment variable.

## Related issue(s)

Fixes #


MikeSpreitzer:
@katamyra

pdettori (commented):
/lgtm

PR #2527, opened 2024-10-26
📖 Adding documentation for quick automated script

katamyra:
## Summary 

Modified get-started.md documentation to include  possibility of using quick automated script.

You should be able to see modifications [here directly](https://katamyra.github.io/kubestellar/doc-krish2/direct/get-started/)

### DIFF
--- a/docs/content/direct/get-started.md
+++ b/docs/content/direct/get-started.md
@@ -8,9 +8,9 @@ There are multiple ways to get started with Kubestellar, reflecting its flexible
 
 Both approaches create a basic environment suitable for learning, but remember that KubeStellar supports more sophisticated configurations including:
 
-- Multiple Independent Target Spaces (ITS)
-- Multiple Workload Distribution Spaces (WDS)
-- Dynamic addition and removal of Work Execution Clusters (WECs)
+- Multiple Inventory and Transport Spaces (ITS)
+- Multiple Workload Definition Spaces (WDS)
+- Dynamic addition and removal of Workload Execution Clusters (WECs)
 - Various deployment patterns to suit different organizational needs
 
 ## Quick Start Using the Automated Script
@@ -18,7 +18,7 @@ Both approaches create a basic environment suitable for learning, but remember t
 If you want to quickly setup a basic environment, you can use our automated installation script:
 
 ```shell
-bash <(curl https://raw.githubusercontent.com/kubestellar/kubestellar/v{{ config.ks_latest_regular_release }}/hack/check_pre_req.sh) kflex ocm helm kubectl docker kind
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/heads/main/scripts/create-kubestellar-demo-env.sh)
 ```
 
 This script sets up a basic environment with one ITS, one WDS, and two WECs. While this is great for getting started, you may want to follow the manual setup below to better understand the components and prepare for more advanced configurations.
--- a/docs/content/direct/get-started.md
+++ b/docs/content/direct/get-started.md
@@ -24,7 +24,7 @@ bash <(curl https://raw.githubusercontent.com/kubestellar/kubestellar/v{{ config
 This script sets up a basic environment with one ITS, one WDS, and two WECs. While this is great for getting started, you may want to follow the manual setup below to better understand the components and prepare for more advanced configurations.
 
 ## Detailed Step walkthrough 
-This following shows one concrete example of steps 2--7 from the [full Installation and Usage outline](user-guide-intro.md#the-full-story). This example produces a simple single-host system suitable for kicking the tires, using [kind](https://kind.sigs.k8s.io/) to create three new clusters to serve as your KubeFlex hosting cluster and two WECs. This page concludes with forwarding you to one example of the remaining steps. For general setup information, see [the full story](user-guide-intro.md#the-full-story).
+The following steps show one concrete example of steps 2--7 from the [full Installation and Usage outline](user-guide-intro.md#the-full-story). This example produces a simple single-host system suitable for kicking the tires, using [kind](https://kind.sigs.k8s.io/) to create three new clusters to serve as your KubeFlex hosting cluster and two WECs. This page concludes with forwarding you to one example of the remaining steps. For general setup information, see [the full story](user-guide-intro.md#the-full-story).
 
   1. [Setup](#setup)
     1. Install software prerequisites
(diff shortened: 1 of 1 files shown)


katamyra:
This is weird, I signed off using "-s" in my commits which worked for my earlier pull requests. Can I add signed-off by in these commits retroactively?

katamyra:
Okay it was because I was using -S rather than -s, oops. I added signoff to all commits, sorry for the mess.

clubanderson:
@katamyra you need a rebase here. Or you can close this PR, pull the latest in your fork and then re-add your change and re-PR

docs/content/direct/get-started.md
```diff
@@ -1,6 +1,30 @@
...
-This page shows one concrete example of steps 2--7 from the [full Installation and Usage outline](user-guide-intro.md#the-full-story). This example produces a simple single-host system suitable for kicking the tires, using [kind](https://kind.sigs.k8s.io/) to create three new clusters to serve as your KubeFlex hosting cluster and two WECs. This page concludes with forwarding you to one example of the remaining steps.
+There are multiple ways to get started with Kubestellar, reflecting its flexible architecture that supports various deployment patterns. This page presents two approaches:
+
+1. A **Step by step walkthrough** that demonstrates the core concepts and components, showing how to manually set up a simple single-host system.
+
+2. A **quick automated setup** using our installation script, which creates a basic working environment for those who want to start experimenting right away. 
+
+Both approaches create a basic environment suitable for learning, but remember that KubeStellar supports more sophisticated configurations including:
+
+- Multiple Independent Target Spaces (ITS)
```
clubanderson, line 11: Inventory and Transport Space

docs/content/direct/get-started.md
```diff
@@ -1,6 +1,30 @@
...
+- Multiple Workload Distribution Spaces (WDS)
```
clubanderson, line 12: Workload Definitions Space

docs/content/direct/get-started.md
```diff
@@ -1,6 +1,30 @@
...
+- Dynamic addition and removal of Work Execution Clusters (WECs)
```
clubanderson, line 13: Workload Execution Cluster

docs/content/direct/get-started.md
```diff
@@ -1,6 +1,30 @@
...
+- Various deployment patterns to suit different organizational needs
+
+## Quick Start Using the Automated Script
+
+If you want to quickly setup a basic environment, you can use our automated installation script:
+
+```shell
+bash <(curl https://raw.githubusercontent.com/kubestellar/kubestellar/v{{ config.ks_latest_regular_release }}/hack/check_pre_req.sh) kflex ocm helm kubectl docker kind
```
clubanderson, line 21: should be: https://raw.githubusercontent.com/kubestellar/kubestellar/refs/heads/main/scripts/create-kubestellar-demo-env.sh)

katamyra:
@clubanderson how does this look?

clubanderson:
@KPRoche please review and approve if so inclined

KPRoche (changes requested):
Items 1 and 2 (step by step vs quick automated setup) should be in the other order, so they match the organization of the rest of the page.

KPRoche:
Other than the order of items in the numbered list at the very top of the page, this looks good now

docs/content/direct/get-started.md
```diff
@@ -1,6 +1,30 @@
 # Getting Started with KubeStellar
 
-This page shows one concrete example of steps 2--7 from the [full Installation and Usage outline](user-guide-intro.md#the-full-story). This example produces a simple single-host system suitable for kicking the tires, using [kind](https://kind.sigs.k8s.io/) to create three new clusters to serve as your KubeFlex hosting cluster and two WECs. This page concludes with forwarding you to one example of the remaining steps.
+There are multiple ways to get started with Kubestellar, reflecting its flexible architecture that supports various deployment patterns. This page presents two approaches:
```
MikeSpreitzer, line 3: I would keep the remark explaining that both of these approaches only aim to produce a simple demo system and are not intended for production use.

docs/content/direct/get-started.md
```diff
@@ -1,6 +1,30 @@
...
+
+1. A **Step by step walkthrough** that demonstrates the core concepts and components, showing how to manually set up a simple single-host system.
+
+2. A **quick automated setup** using our installation script, which creates a basic working environment for those who want to start experimenting right away. 
+
+Both approaches create a basic environment suitable for learning, but remember that KubeStellar supports more sophisticated configurations including:
```
MikeSpreitzer, line 9: I would strengthen this remark to more clearly NOT suggest that this could be used for production use.

MikeSpreitzer, line 9: I would add an explicit and unqualified remark that both of these approaches are UNsuitable for production usage.

docs/content/direct/get-started.md
```diff
@@ -1,6 +1,30 @@
...
+
+## Quick Start Using the Automated Script
+
+If you want to quickly setup a basic environment, you can use our automated installation script:
+
+```shell
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/heads/main/scripts/create-kubestellar-demo-env.sh)
+```
+
+This script sets up a basic environment with one ITS, one WDS, and two WECs. While this is great for getting started, you may want to follow the manual setup below to better understand the components and prepare for more advanced configurations.
```
MikeSpreitzer, line 24: I think that it is also worth saying that this script creates three `kind` clusters, one to serve as the KubeFlex hosting cluster and two to serve as WECs.

docs/content/direct/get-started.md
```diff
@@ -1,6 +1,30 @@
...
+
```
MikeSpreitzer, line 25: It may also be worth putting an explicit forward pointer here to the "next step", which is the "Exercise KubeStellar" section.

Do both setup techniques call for exactly the same shell variable values?

docs/content/direct/get-started.md
```diff
@@ -1,6 +1,30 @@
...
+- Multiple Inventory and Transport Spaces (ITS)
+- Multiple Workload Definition Spaces (WDS)
+- Dynamic addition and removal of Workload Execution Clusters (WECs)
+- Various deployment patterns to suit different organizational needs
+
+## Quick Start Using the Automated Script
+
+If you want to quickly setup a basic environment, you can use our automated installation script:
+
```
MikeSpreitzer, line 19: I think it would be worthwhile to point here to the pre-requisites section, which is a little more expansive and helpful than the output of the checking script.

If we believe that the note for Windows users should stay, it would apply to both techniques, right?

docs/content/direct/get-started.md
MikeSpreitzer, line 3: I think it would be good to be a little more specific about the scope of the script and its relationship with the step-by-step. I might say things like:
- The script does the same things as the parts of the step-by-step from the checking of prerequisites through creating and registering two WECs.
- The other remarks below apply equally, including the establishing of the prerequisites (including making `kind` able to create three clusters) and the remarks about variables to set and what to do afterwards.
- The script does things in a different order, using as much concurrency as possible to get as much speed as possible. This makes it more complicated than the step-by-step but only for technical reasons related to the concurrency.

MikeSpreitzer:
To be clear, none of my review comments above are blockers.

PR #2539, opened 2024-10-30
✨ minor updates to script for formatting and verbiage

clubanderson:
## Related issue(s)

Fixes #


scripts/create-kubestellar-demo-env.sh
```diff
@@ -173,7 +173,8 @@ echo "Checking the new clusters are in the OCM inventory and label them"
 kubectl --context its1 get managedclusters
 kubectl --context its1 label managedcluster cluster1 location-group=edge name=cluster1
 kubectl --context its1 label managedcluster cluster2 location-group=edge name=cluster2
-echo "\n\nCongratulations! Your KubeStellar demo environment is now ready to use."
+echo""
```
MikeSpreitzer, line 176: This line issues a command named `echo` (composed of literal `echo` followed by empty string) with no arguments. It could be done more simply without the empty string being appended to the command name. Not a blocker.

MikeSpreitzer (approved):
Good enough to merge.
One comment about something to improve.
/lgtm
/approve
Thanks!

PR #2540, opened 2024-10-31
🌱  Give some love to ginkgo --fail-fast

MikeSpreitzer:
## Summary
This PR does two things. One is to change the GitHub Actions workflow that runs the ginkgo E2E test, to use `--fail-fast`. This is good because it means that the state dumping after the test shows the problem state.

The other thing is to generalize the `test/e2e/run-test.sh` script so that it can be given `--fail-fast`. This is good because rushing past a failure destroys state that a contributor would want to examine.

## Related issue(s)

Fixes #


MikeSpreitzer:
The force-push to ad4c9dc99 is a rebase onto `main`.

pdettori (commented):
/lgtm

PR #2543, opened 2024-10-31
✨ fix: sometimes cluster1 or cluster2 kubeconfig context alias does not create on the first try

clubanderson:
I have discovered that every so often that 1 of the cluster1 and cluster2 aliases in kubeconfig do not create propertly - I assume there is a blocking happening since both clusters are create simultaneously - to avoid this I added a retry for both clusters - not pretty, but necessary to avoid missed config

## Related issue(s)

Fixes #

### DIFF
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -125,7 +125,7 @@ for cluster in "${clusters[@]}"; do
 done
 
 for cluster in "${clusters[@]}"; do
-  if kubectl config get-contexts | grep -w "\b${cluster}\b" >/dev/null 2>&1; then
+  if kubectl config get-contexts | grep -w " ${cluster} " >/dev/null 2>&1; then
     echo -e "\033[33m✔\033[0m $cluster context exists."
   else
     if kubectl config rename-context "kind-${cluster}" "${cluster}" >/dev/null 2>&1; then
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -116,6 +116,7 @@ for cluster in "${clusters[@]}"; do
      echo -e "Creating cluster ${cluster}..."
      kind create cluster --name "${cluster}" >/dev/null 2>&1 &&
      kubectl config rename-context "kind-${cluster}" "${cluster}" >/dev/null 2>&1
+     kubectl config rename-context "kind-${cluster}" "${cluster}" >/dev/null 2>&1
      echo -e "${cluster} creation and context setup complete"
    ) &
 done
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -103,11 +103,11 @@ cluster_clean_up "kind delete cluster --name kubeflex" &
 cluster_clean_up "kind delete cluster --name cluster1" &
 cluster_clean_up "kind delete cluster --name cluster2" &
 wait
-echo -e "Cluster space clean up has been completed"
+echo -e "\033[33m✔\033[0m Cluster space clean up has been completed"
 
 echo -e "\nStarting context clean up..."
 context_clean_up
-echo "Context space clean up completed"
+echo -e "\033[33m✔\033[0m Context space clean up completed"
 
 echo -e "\nStarting the process to install KubeStellar core: kind-kubeflex..."
 clusters=(cluster1 cluster2)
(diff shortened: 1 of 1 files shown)


scripts/create-kubestellar-demo-env.sh
```diff
@@ -116,6 +116,7 @@ for cluster in "${clusters[@]}"; do
      echo -e "Creating cluster ${cluster}..."
      kind create cluster --name "${cluster}" >/dev/null 2>&1 &&
      kubectl config rename-context "kind-${cluster}" "${cluster}" >/dev/null 2>&1
+     kubectl config rename-context "kind-${cluster}" "${cluster}" >/dev/null 2>&1
```
MikeSpreitzer, line 119: It is not safe to just repeat this command! For example:

```console
mspreitz@mjs13 kubestellar % kubectl config rename-context kind-cluster1 cluster1
error: cannot rename the context "kind-cluster1", it's not in /Users/mspreitz/.kube/config
```

FYI, I asked (in the `#sig-cli` channel of the Kubernetes Slack organization) about concurrent `kubectl config rename-context` commands, and was told:

> so it's safe in the sense that it won't create a broken state from reading at different versions before mutating, but one of the commands might fail

> so it's not safe to assume they succeeded. but it is safe to assume that your configs won't be mangled.

See https://kubernetes.slack.com/archives/C2GL57FJ4/p1729873069405909 , while you still can, for the whole conversation.

I tried testing, and there is something strange going on. I got no output from the `kind create` commands in this loop.

Given that concurrent context renaming is problematic, and there is not much time to be saved there, and it is being lost anyway by repeating commands, I think the better approach is to do the context renamings serially, after waiting for the cluster creation commands to complete.

clubanderson:
issue still occurring - looking at another possible way to correct behavior

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: Here is a revised version of this section that I have successfully tested.

```shell
   (
     echo -e "Creating cluster ${cluster}..."
     kind create cluster --name "${cluster}" >/dev/null 2>&1
     echo -e "${cluster} creation and context setup complete"
   ) &
done
wait
kubectl config rename-context kind-cluster1 cluster1
kubectl config rename-context kind-cluster2 cluster2
```

MikeSpreitzer:
> issue still occurring - looking at another possible way to correct behavior

I put one in a thread above.

All that redirection to `/dev/null` makes it harder to even get a clue about what is going wrong.

MikeSpreitzer:
Also, regarding even noticing problems, note that `&` defeats `set -e`.

```console
bash-3.2$ set -e
bash-3.2$ false & true
[1] 14497
bash-3.2$ 
[1]+  Exit 1                  false
bash-3.2$ jobs
bash-3.2$ 
```

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: See #2546

clubanderson:
ready for review

scripts/create-kubestellar-demo-env.sh
```diff
@@ -103,27 +103,42 @@ cluster_clean_up "kind delete cluster --name kubeflex" &
...
    ) &
 done
 wait 
 
+for cluster in "${clusters[@]}"; do
+   kubectl config rename-context "kind-${cluster}" "${cluster}" >/dev/null 2>&1
+done
+
+for cluster in "${clusters[@]}"; do
+  if kubectl config get-contexts | grep -w "\b${cluster}\b" >/dev/null 2>&1; then
```
MikeSpreitzer, line 128: I suspect that this test is not doing what you expect:

```console
mspreitz@mjs13 kubestellar % echo kind-cluster1 | grep -w "\bcluster1\b"
kind-cluster1
```

BTW, when I do `man grep` on my Mac, it refers me to `man re_format` for the details on the r.e. language. The latter man page lists `\b` in the "ENHANCED FEATURES" section. There is a plainer way to get the job done, in this particular case.

```console
mspreitz@mjs13 kubestellar % echo " * cluster1 rest" | grep "[ ]cluster1[ ]"
 * cluster1 rest
mspreitz@mjs13 kubestellar % echo " kind-cluster1 rest" | grep "[ ]cluster1[ ]" 
mspreitz@mjs13 kubestellar % 
```

MikeSpreitzer, line 128: Doesn't the change, above, to doing the context renamings serially remove the need to check and retry?

clubanderson:
@MikeSpreitzer good find on the simpler grep. I updated. I left in the 'just in case context did not get renamed' bit because it is still a possibility. It can be removed, but then you're wasting your time waiting for the script to complete

MikeSpreitzer:
@clubanderson, what do you mean about waiting?

> It can be removed, but then you're wasting your time waiting for the script to complete

If you are referring to the `wait` command that is now on line 121 (after the "done" of the bash loop), that is needed even if there is no checking and retry of the renaming. The `wait` is needed regardless of retry, to make the first tries to rename follow the completion of the `kind create` commands.

MikeSpreitzer (approved):
This has an essential fix: doing the context renamings serially and after completion of cluster creation. We can haggle over other details in follow-on.

/lgtm
/approve
Thanks!

MikeSpreitzer:
For later work:

> I left in the 'just in case context did not get renamed' bit because it is still a possibility.

Is that something you observed with this script? I understand the need when the context renamings were done concurrently with other cluster creations and context renamings, but I would be surprised if the context renamings could fail now that they are totally ordered wrt each other and other kubeconfig manipulations.

PR #2546, opened 2024-10-31
🐛  Various fixes and improvements to create-kubestellar-demo-env.sh

MikeSpreitzer:
## Summary
This PR improves the demo environment creation script in the following ways.

- The script retains, and displays if there is a failure, the logging from `kind`.
- The whole script fails if any WEC creation fails.
- This PR adds a wait for the update-cluster-info Job to complete before running the `clusteradm join` commands, because the latter depends on the former.
- Change `scripts/create-kind-cluster-with-SSL-passthrough.sh` to explicitly specify the kubeconfig context in each command, so that this script can be used in parallel with other activity that sets/requires the current context to be something else.
- Change `create-kubestellar-demo-env.sh` to create the hub cluster concurrently with creating the WECs, for improved total execution time.
- Use the `check_pre_req.sh` script from the `main` branch, to pick up #2557 .

## Related issue(s)

Fixes #2558

PR #2565, opened 2024-11-07
🌱  Changed create-kubestellar-demo-env.sh ref from me to main

MikeSpreitzer:
## Summary
This PR does some clean-up of #2546 , which --- as a hack to enable testing --- referenced its own copy of `scripts/create-kind-cluster-with-SSL-passthrough.sh`. This PR changes that to reference the copy in the `main` branch. When making the next release, the reference will be changed again, to reference the released version.

## Related issue(s)

Fixes #


pdettori (commented):
/lgtm

PR #2569, opened 2024-11-08
✨ Autostarting docker in create demo env script

katamyra:
## Summary

Autostart docker engine when running the create demo script.

### DIFF
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -17,37 +17,13 @@
 
 set -e
 
-echo -e "Starting Docker engine..."
-# Check if Docker is running using OS-agnostic commands
-if ! docker info >/dev/null 2>&1; then
-    case "$(uname -s)" in
-        Linux*)
-            sudo systemctl start docker
-            ;;
-        Darwin*)
-            open -a Docker
-            # Wait for Docker to start
-            echo "Waiting for Docker to start..."
-            while ! docker info >/dev/null 2>&1; do
-                sleep 2
-            done
-            ;;
-        MINGW*|CYGWIN*|MSYS*)
-            echo "Please start Docker Desktop manually if not running"
-            echo "Waiting for Docker to be available..."
-            while ! docker info >/dev/null 2>&1; do
-                sleep 2
-            done
-            ;;
-        *)
-            echo "Unknown operating system. Please ensure Docker is running manually."
-            exit 1
-            ;;
-    esac
-    echo "Docker engine started."
-else
-    echo "Docker engine is already running."
+
+echo -e "Checking container runtime..."
+if ! docker ps -q; then
+    echo "Error: The script cannot continue because Docker or Podman is not running. Please start your container runtime before running the script again."
+    exit 1
 fi
+echo "Container runtime is running."
 
 kubestellar_version=0.25.0-rc.1
 echo -e "KubeStellar Version: ${kubestellar_version}"
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -19,7 +19,7 @@ set -e
 
 
 echo -e "Checking container runtime..."
-if ! docker ps -q; then
+if ! dunsel=$(docker ps 2>&1); then
     echo "Error: The script cannot continue because Docker or Podman is not running. Please start your container runtime before running the script again."
     exit 1
 fi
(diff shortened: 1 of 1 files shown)


katamyra:
@clubanderson is this a good idea? I remember people kept running the script without docker engine enabled so I thought we could auto start it in the script.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -17,6 +17,38 @@
...
 
+echo -e "Starting Docker engine..."
+# Check if Docker is running using OS-agnostic commands
+if ! docker info >/dev/null 2>&1; then
+    case "$(uname -s)" in
+        Linux*)
+            sudo systemctl start docker
+            ;;
+        Darwin*)
+            open -a Docker
```
MikeSpreitzer, line 28: On my Mac, where I am using Rancher Desktop and the docker container runtime:

```console
mspreitz@mjs13 kubestellar % open -a Docker
Unable to find application named 'Docker'
```

In general, remember that there is commonly a "desktop" application that provides a UI and the thing that actually provides the container runtime is something else.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -17,6 +17,38 @@
 
 set -e
 
+echo -e "Starting Docker engine..."
+# Check if Docker is running using OS-agnostic commands
+if ! docker info >/dev/null 2>&1; then
+    case "$(uname -s)" in
+        Linux*)
+            sudo systemctl start docker
```
MikeSpreitzer, line 25: Some Linux distros do not use systemd.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -17,6 +17,14 @@
 
 set -e
 
+
+echo -e "Checking container runtime..."
+if ! docker ps -q; then
```
MikeSpreitzer, line 22: FYI: `-q` does not suppress all output.

katamyra: @MikeSpreitzer Should I do something like this instead? This was my original idea. To my knowledge null device changes between linux/macos and windows. 

```bash
# Need to check if windows / linux&macos. 
if [[ "$OSTYPE" == "msys" || "$OSTYPE" == "cygwin" || "$OSTYPE" == "win32" ]]; then
    # windows (Git bash, Cygwin, or native Windows)
    NULL_DEVICE="NUL"
else
    NULL_DEVICE="/dev/null"
fi

if ! docker ps -q > "$NULL_DEVICE" 2>&1; then
    echo "Error: The script cannot continue because Docker or Podman is not running. Please start your container runtime before running the script again."
    # You can add any additional actions here if Docker is not running
fi

```

MikeSpreitzer: (BTW, if you are discarding the output then `-q` doesn't matter, right?)

I am not familiar with whether/what cygwin and msys have instead of /dev/null. Assuming that what you have written is accurate, that would work.

It _is_ a dismaying lot of code. Could we instead unify on directing output to a shell variable? Something like the following?

```shell
if ! dunsel=$(docker ps); then
    ...
fi
```

katamyra: You're right, this is a much simpler way to do it, and its easier to make sure it works across all devices.

MikeSpreitzer (commented):
/lgtm
Thanks!

MikeSpreitzer:
Oh, wait. We try to keep a simple Git history. Instead of merging `main` into your PR's branch, you should rebase.

/remove-lgtm

katamyra:
> Oh, wait. We try to keep a simple Git history. Instead of merging `main` into your PR's branch, you should rebase that branch onto `main`.
> 
> /remove-lgtm

@MikeSpreitzer Sorry I am pretty new to Git and I think I messed it up by following the Github websites rebase/merge conflicts instead of the command line and it also didn't signoff the commit on the website. I tried to interactive rebase to add the signoff but now I have a lot of messy commits. Should I start a clean PR for simplicity?

MikeSpreitzer:
Yes, since the change is so small and compact, starting over might be best.

It should also be possible to recover, if you can pair with a Git expert for a little while. But it would be fine to just start over.

MikeSpreitzer:
Also, the title of this PR needs to be updated.

clubanderson:
@katamyra would you like me to recreate this and put up for review?

clubanderson:
I superseded with #2618 @katamyra

Issue #2508, opened 2024-10-17
[IFOS TERM - 3 2025] Provide a CLI for KubeStellar (unpaid)


francostellari:
**READ THIS FIRST: We use PRs in our repos as our main evaluation criteria - if you are interested in this project - please begin by looking for issues marked "help-wanted" and start to PR them. PRs are our main source information to evaluate potential IFOS interns. If you have no PR history with KubeStellar or any of its repositories then you should not apply for consideration for our IFOS projects**

### Background

While the deployment of KubeStellar is now pretty straight forward by using the Helm core-chart, the subsequent inspection, usage, and debugging does not have a consistent CLI, requiring a combination of `kflex` and `kubectl` commands, sometimes not obvious to the non expert user.

### Project proposal

It would be beneficial to introduce a KubeStellar CLI to 
- deploy KS core via its Helm chart
- add/delete/update CPs
- access/get the kubeconfig of the CPs
- retrieve/depict the current status of the KubeStelalr deployment

Ideally the CLI should try to guess the correct context without necessarily being told (see what [import-cp-contexts.sh](https://github.com/kubestellar/kubestellar/blob/main/scripts/import-cp-contexts.sh) does).

### Mentees
Paid intern positions: 1
Unpaid intern positions: 1

clubanderson:
@rxinui it would be great for you to have a closer look at this. Thank you.

@francostellari any more context or comments you would like to add to help @rxinui get started?

rxinui:
Sure @clubanderson, I would need to know which direction the maintainers want to go.

If it were up to me, I'd go on a `kubestellar/cli` written in Go, such as `kflex` for the sake on being consistent with the current stack. That will also be handy when calling other golib written by `kflex` or other kubernetes resources.

It would be a great learning opportunity as well.

rxinui:
@francostellari any updates or details on that idea?

I'll start when I'll know more about it

francostellari:
Your proposal envisions a reciprocal dependency on the projects.
Right now kflex does not depend or know anything about kubesterllar specific things.
- kflex does know what a its or wds is
- kflex does not know about binding policies etc

I personally do not think it's a good idea to go down this road.

rxinui:
Got it! 

Thank you for the clarification.
Regarding the reciprocal dependencies you mentioned, it won't necessarily be true.

We could leverage kubernetes client, and instead of being static, we can be dynamic and make used of unstructured data.

[...]

PR #2570, opened 2024-11-08
✨Provide a script to take a snapshot of KubeStellar deployment

francostellari:
## Summary

Provide a new script to take a snapshot of KubeStellar deployment: `kubestellar-snapshot.sh`

```shell
Usage: ./kubestellar-snapshot.sh [--kubeconfig|-K <filename>] [--context|-C <name>] [--verbose|-V] [-X]
--kubeconfig|-K <filename> use the specified kubeconfig
--context|-C <name>        use the specified context
--logs|-L                  save the logs of the pods
--yaml|-Y                  save the YAML of the resources
--verbose|-V               output extra information
--version|-v               print out the script version
-X                         enable verbose execution of the script for debugging
```

Example of usage after running `test/e2c/run_test.sh`:

```shell
$ ./kubestellar-snapshot.sh -V -Y -L
KubeStellar Snapshot v0.10

Script run on 2024-11-07_23:21:53
Checking script dependencies:
✔ kubectl version v1.31.1 at /usr/local/bin/kubectl
✔ helm version v3.16.1 at /usr/sbin/helm
✔ jq version jq-1.6 at /usr/bin/jq
Using kubeconfig(s): /home/vagrant/.kube/config
Using contexts(s): cluster1, cluster2, its1, kind-kubeflex, wds1  

KubeStellar:
- Helm chart ks-core (v0.1.0) in context kind-kubeflex

KubeFlex:
- kubeflex-system namespace in context kind-kubeflex
- controller-manager: version=0.7.1, pod=kubeflex-controller-manager-54c7fcb78c-gmgr4, status=Running
- postgres-postgresql-0: pod=postgres-postgresql-0, status=Running

Control Planes:
- its1: type=vcluster, pch=its, context=kind-kubeflex, namespace=its1-system
  - Post Create Hook: pod=its-vsmt2, status=Succeeded
  - Status addon: pod=addon-status-controller-d8864f4fd-m57xq-x-open-clust-fbedb9af6a, status=Running
- wds1: type=k8s, pch=wds, context=kind-kubeflex, namespace=wds1-system
  - KubeStellar controller: version=bcda2ef6a, pod=kubestellar-controller-manager-85c5c954db-lv5lm namespace=wds1-system, status=Running
  - Transport controller: version=bcda2ef6a, pod=transport-controller-697d88bf55-b4ffg namespace=wds1-system, status=Running

Managed Clusters:
- cluster1 in its1: accepted=True, joined=True, available=True, synced=True
  {
    "location-group": "edge",
    "name": "cluster1",
    "region": "east"
  }
- cluster2 in its1: accepted=True, joined=True, available=True, synced=True
  {
    "location-group": "edge",
    "name": "cluster2",
    "region": "west"
  }

Binding Policies:
- nginx-bindingpolicy in control plane wds1
  {
    "clusterSelectors": [
      {
        "matchLabels": {
          "location-group": "edge"
        }
      }
    ],
    "downsync": [
      {
        "objectSelectors": [
          {
            "matchLabels": {
              "app.kubernetes.io/name": "nginx"
            }
          }
        ]
      }
    ]
  }

Manifest Works:
- nginx-bindingpolicy-wds1-0 in cp=its1, namespace=cluster1: wds1 --> nginx-bindingpolicy --> cluster1
  {
    "apiVersion": "v1",
    "kind": "Namespace",
    "name": "nginx"
  }
- nginx-bindingpolicy-wds1-1 in cp=its1, namespace=cluster1: wds1 --> nginx-bindingpolicy --> cluster1
  {
    "apiVersion": "apps/v1",
    "kind": "Deployment",
    "name": "nginx-deployment"
  }
- nginx-bindingpolicy-wds1-0 in cp=its1, namespace=cluster2: wds1 --> nginx-bindingpolicy --> cluster2
  {
    "apiVersion": "v1",
    "kind": "Namespace",
    "name": "nginx"
  }
- nginx-bindingpolicy-wds1-1 in cp=its1, namespace=cluster2: wds1 --> nginx-bindingpolicy --> cluster2
  {
    "apiVersion": "apps/v1",
    "kind": "Deployment",
    "name": "nginx-deployment"
  }

Saving logs and/or YAML to ./kubestellar-snapshot.tar.gz
```

Here are the file saved:

```shell
$ tree /tmp/kubestellar-snapshot/
/tmp/kubestellar-snapshot/
├── its1
│   ├── cp.yaml
│   ├── its-job-clusteradm.log
│   ├── its-job-status-addon.log
│   ├── its-job.yaml
│   ├── managed-clusters
│   │   ├── cluster1.yaml
│   │   └── cluster2.yaml
│   ├── manifest-works
│   │   ├── cluster1
│   │   │   ├── nginx-bindingpolicy-wds1-0.yaml
│   │   │   └── nginx-bindingpolicy-wds1-1.yaml
│   │   └── cluster2
│   │       ├── nginx-bindingpolicy-wds1-0.yaml
│   │       └── nginx-bindingpolicy-wds1-1.yaml
│   ├── status-addon.log
│   └── status-addon.yaml
├── kubeflex
│   ├── kubeflex-controller.log
│   └── postgresql.log
├── kubestellar
│   ├── its-pch.yaml
│   └── wds-pch.yaml
└── wds1
    ├── binding-politcies
    │   └── nginx-bindingpolicy.yaml
    ├── cp.yaml
    ├── kubestellar-controller.log
    ├── kubestellar-controller.yaml
    ├── transport-controller.log
    └── transport-controller.yam
```

## Related issue(s)
#2508 

Fixes #

### DIFF
--- a/scripts/kubestellar-snapshot.sh
+++ b/scripts/kubestellar-snapshot.sh
@@ -20,7 +20,7 @@ set -e
 
 # Global variables
 TIMESTAMP="$(date +%F_%T)"
-TMPFOLDER="$(mktemp -d --tmpdir "kubestellar-snapshot-XXXX")"
+TMPFOLDER="$(mktemp -d -p . "kubestellar-snapshot-XXXX")"
 trap 'rm -rf -- "$TMPFOLDER"' EXIT
 OUTPUT_FOLDER="$TMPFOLDER/kubestellar-snapshot"
 
@@ -136,7 +136,7 @@ is_installed() {
 
 # Get the kubeconfig of a particular Control Plane
 get_kubeconfig() {
-    context="$1" # context in the host kubeconfig or the config passed to the script using --kubeconfig flag
+    context="$1" # context in the relevant kubeconfig
     cp_name="$2" # name of the Control Plane
     cp_type="$3" # type of the Control Plane
 
@@ -217,7 +217,7 @@ fi
 ###############################################################################
 # Script info
 ###############################################################################
-echov -e "${COLOR_INFO}$(${BASH_SOURCE[0]} --version)${COLOR_NONE}\n"
+echov -e "${COLOR_INFO}${SCRIPT_NAME} v${SCRIPT_VERSION}{COLOR_NONE}\n"
 echov -e "Script run on ${COLOR_INFO}$TIMESTAMP${COLOR_NONE}"
 
 
@@ -303,7 +303,7 @@ for context in "${valid_contexts[@]}" ; do
         version="$(echo "$line" | awk '{print $NF}')"
         secret="$(kubectl --context $context get secret -n "$namespace" -l "owner=helm,name=$name" --no-headers -o name 2> /dev/null || true)"
         secret="${secret##*/}"
-        if [[ ! -z "$(kubectl --context $context get secret $secret -n "$namespace" -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.chart' | grep -i "KubeStellar" 2> /dev/null || true)" ]] ; then
+        if [[ ! -z "$(kubectl --context $context get secret $secret -n "$namespace" -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.chart.metadata.description' | grep -i "KubeStellar" 2> /dev/null || true)" ]] ; then
             helm_context="$context"
             echo -e "- Helm chart ${COLOR_INFO}${name}${COLOR_NONE} (${COLOR_INFO}v${version}${COLOR_NONE}) in namespace ${COLOR_INFO}${namespace}${COLOR_NONE} in context ${COLOR_INFO}${context}${COLOR_NONE}"
             echo -e "  - Secret=${COLOR_INFO}${secret}${COLOR_NONE} in namespace ${COLOR_INFO}${namespace}${COLOR_NONE}"
@@ -374,7 +374,7 @@ for i in "${!cps[@]}" ; do # for all control planes in context ${context}
     echo -e "- ${COLOR_INFO}${cp_name[cp_n]}${COLOR_NONE}: type=${COLOR_INFO}${cp_type[cp_n]}${COLOR_NONE}, pch=${COLOR_INFO}${cp_pch[cp_n]}${COLOR_NONE}, context=${COLOR_INFO}${cp_context[cp_n]}${COLOR_NONE}, namespace=${COLOR_INFO}${cp_name[cp_n]}-system${COLOR_NONE}"
     if [[ -z "$cp_kubeconfig_content" ]] ; then
         cp_kubeconfig[cp_n]=""
-        echo -e "  ${COLOR_WARNING}WARNING: the Control plane is no ready, the kubeconfig is not available!${COLOR_NONE}"
+        echo -e "  ${COLOR_WARNING}WARNING: the Control Plane is not ready, the kubeconfig is not available!${COLOR_NONE}"
     else
         cp_kubeconfig[cp_n]="$TMPFOLDER/$name-kubeconfig"
         echo "$cp_kubeconfig_content" > "${cp_kubeconfig[cp_n]}"
--- a/scripts/kubestellar-snapshot.sh
+++ b/scripts/kubestellar-snapshot.sh
@@ -284,11 +284,15 @@ for context in "${contexts[@]}" ; do # for all contexts
         helm_context=$context
         name="$(echo "$chart" | awk '{print $1}')" || true
         version="$(echo "$chart" | awk '{print $NF}')" || true
+        secret="$(k --context $helm_context get secret --all-namespaces -l "owner=helm"  --no-headers -o name 2> /dev/null | grep "$name" || true)"
+        secret="${secret##*/}"
         echo -e "- Helm chart ${COLOR_INFO}${name}${COLOR_NONE} (${COLOR_INFO}v${version}${COLOR_NONE}) in context ${COLOR_INFO}${context}${COLOR_NONE}"
+        echo -e "  - Secret=${COLOR_INFO}${secret}${COLOR_NONE}"
         if [[ "$arg_yaml" == "true" ]] ; then
-            mkdir -p "$output_folder/kubestellar"
-            k --context $helm_context get postcreatehook its -o yaml > "$output_folder/kubestellar/its-pch.yaml"
-            k --context $helm_context get postcreatehook wds -o yaml > "$output_folder/kubestellar/wds-pch.yaml"
+            mkdir -p "$output_folder/kubestellar-core-chart"
+            echo -e "$(k --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.info'     | sed -e 's/\"/"/g')" > "$output_folder/kubestellar-core-chart/info.json"
+            echo -e "$(k --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.chart'    | sed -e 's/\"/"/g')" > "$output_folder/kubestellar-core-chart/chart.json"
+            echo -e "$(k --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.manifest' | sed -e 's/\"/"/g')" > "$output_folder/kubestellar-core-chart/manifest.yaml"
         fi
         break
     fi
--- a/scripts/kubestellar-snapshot.sh
+++ b/scripts/kubestellar-snapshot.sh
@@ -19,13 +19,15 @@ set -e
 
 
 # Global variables
-timestamp="$(date +%F_%T)"
-output_folder="/tmp/kubestellar-snapshot"
+TIMESTAMP="$(date +%F_%T)"
+TMPFOLDER="$(mktemp -d --tmpdir "kubestellar-snapshot-XXXX")"
+trap 'rm -rf -- "$TMPFOLDER"' EXIT
+OUTPUT_FOLDER="$TMPFOLDER/kubestellar-snapshot"
 
 
 # Script info
 SCRIPT_NAME="KubeStellar Snapshot"
-SCRIPT_VERSION="0.10"
+SCRIPT_VERSION="0.2.0"
 
 
 # Colors
@@ -35,6 +37,7 @@ COLOR_GREEN="\033[1;32m"
 COLOR_BLUE="\033[94m"
 COLOR_YELLOW="\033[1;33m"
 COLOR_PURPLE="\033[1;35m"
+COLOR_WARNING="${COLOR_RED}"
 COLOR_ERROR="${COLOR_RED}"
 COLOR_STATUS_TRUE="${COLOR_GREEN}"
 COLOR_STATUS_FALSE="${COLOR_RED}"
@@ -54,8 +57,9 @@ arg_verbose=false
 # Display the command line help
 display_help() {
   cat << EOF
-Usage: $0 [--kubeconfig|-K <filename>] [--context|-C <name>] [--logs|-L] [--yaml|-Y] [--verbose|-V] [--version|-v] [--help|-h] [-X]
+Usage: $0 [options]
 
+Options:
 --kubeconfig|-K <filename> use the specified kubeconfig to find KubeStellar Helm chart
 --context|-C <name>        use the specified context to find KubeStellar Helm chart
 --logs|-L                  save the logs of the pods
@@ -65,10 +69,13 @@ Usage: $0 [--kubeconfig|-K <filename>] [--context|-C <name>] [--logs|-L] [--yaml
 --help|-h                  show this information
 -X                         enable verbose execution for debugging
 
-Note: After piping the output of the script to file use "more"
-      to inspect the file content in color. ALternatively install
-      "ansi2txt" and pipe the script through it to remove escape
-      characters and obtain a plan text report.
+Note 1: This script expects that KubeStellar was installed with its Helm chart and
+        to find the latter in the host or provided kubeconfig(s) and context(s).
+
+Note 2: After piping the output of the script to file use "more"
+        to inspect the file content in color. ALternatively install
+        "ansi2txt" and pipe the script through it to remove escape
+        characters and obtain a plan text report.
 EOF
 }
 
@@ -129,25 +136,26 @@ is_installed() {
 
 # Get the kubeconfig of a particular Control Plane
 get_kubeconfig() {
-    context="$1" # context in the kubeconfig
+    context="$1" # context in the host kubeconfig or the config passed to the script using --kubeconfig flag
     cp_name="$2" # name of the Control Plane
     cp_type="$3" # type of the Control Plane
 
-    # wait for CP ready
-    while [[ $(k --context $context get controlplane $cp_name -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do
-        sleep 5
-    done
+    # check if the CP is ready
+    if [[ $(kubectl --context $context get controlplane $cp_name -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; then
+        echo ""
+        return
+    fi
 
     # put into the shell variable "kubeconfig" the kubeconfig contents for use from outside of the hosting cluster
     if [[ "$cp_type" == "host" ]] ; then
-        kubeconfig=$(k --context $context config view --flatten --minify)
+        kubeconfig=$(kubectl --context $context config view --flatten --minify)
     else
         # determine the secret name and namespace
-                     key=$(k --context $context get controlplane $cp_name -o=jsonpath='{.status.secretRef.key}')
-             secret_name=$(k --context $context get controlplane $cp_name -o=jsonpath='{.status.secretRef.name}')
-        secret_namespace=$(k --context $context get controlplane $cp_name -o=jsonpath='{.status.secretRef.namespace}')
+                     key=$(kubectl --context $context get controlplane $cp_name -o=jsonpath='{.status.secretRef.key}')
+             secret_name=$(kubectl --context $context get controlplane $cp_name -o=jsonpath='{.status.secretRef.name}')
+        secret_namespace=$(kubectl --context $context get controlplane $cp_name -o=jsonpath='{.status.secretRef.namespace}')
         # get the kubeconfig
-        kubeconfig=$(k --context $context get secret $secret_name -n $secret_namespace -o=jsonpath="{.data.$key}" | base64 -d)
+        kubeconfig=$(kubectl --context $context get secret $secret_name -n $secret_namespace -o=jsonpath="{.data.$key}" | base64 -d)
     fi
 
     # return the kubeconfig file contents in YAML format
@@ -196,7 +204,7 @@ done
 
 
 ###############################################################################
-# Define alias definition
+# Alias definitions
 ###############################################################################
 # Define the echov function based on verbosity
 if [ "$arg_verbose" == "true" ]; then
@@ -205,17 +213,12 @@ else
     echov() { :; }
 fi
 
-# Define the special kubectl
-k() {
-    KUBECONFIG="$kubeconfig" kubectl $*
-}
-
 
 ###############################################################################
 # Script info
 ###############################################################################
-echov -e "${COLOR_INFO}$($0 --version)${COLOR_NONE}\n"
-echov -e "Script run on ${COLOR_INFO}$timestamp${COLOR_NONE}"
+echov -e "${COLOR_INFO}$(${BASH_SOURCE[0]} --version)${COLOR_NONE}\n"
+echov -e "Script run on ${COLOR_INFO}$TIMESTAMP${COLOR_NONE}"
 
 
 ###############################################################################
@@ -240,8 +243,7 @@ is_installed 'jq' \
 # Ensure output folder
 ###############################################################################
 if [[ "$arg_logs" == "true" || "$arg_yaml" == "true" ]] ; then
-    yes "yes" | rm -vRI "/tmp/kubestellar-snapshot" > /dev/null 2>&1 || true
-    mkdir -p "$output_folder"
+    mkdir -p "$OUTPUT_FOLDER"
 fi
 
 
@@ -249,42 +251,39 @@ fi
 # Determine the list of kubeconfigs to search
 # based on https://kubernetes.io/docs/reference/k/generated/k_config/
 ###############################################################################
-if [[ "$arg_kubeconfig" != "" ]] ; then
-    kubeconfig="$arg_kubeconfig"
-elif [[ "$KUBECONFIG" != "" ]] ; then
-    kubeconfig="$KUBECONFIG"
-else
-    kubeconfig="$HOME/.kube/config"
-fi
-echov -e "Using kubeconfig(s): ${COLOR_INFO}$kubeconfig${COLOR_NONE}"
+[[ -n "$arg_kubeconfig" ]] && export KUBECONFIG="$arg_kubeconfig"
+[[ -z "$KUBECONFIG" ]] && export KUBECONFIG="$HOME/.kube/config"
+echov -e "Using kubeconfig(s): ${COLOR_INFO}$KUBECONFIG${COLOR_NONE}"
 
 
 ###############################################################################
 # Determine the list of contexts to search
 ###############################################################################
 if [[ "$arg_context" == "" ]] ; then
-    contexts=($(k config get-contexts --no-headers -o name))
+    contexts=($(kubectl config get-contexts --no-headers -o name))
 else
     contexts=("$arg_context")
 fi
 
+if [[ -z "${contexts[@]}" ]] ; then
+    echoerr "No context(s) found in the kubeconfig file $KUBECONFIG!"
+    exit 1
+fi
+
 echov "Validating contexts(s): "
-vc_n=0
-valid_context=()
+valid_contexts=()
 for context in "${contexts[@]}" ; do # for all contexts
     [[ -z "$context" ]] && continue
-    if k --context $context get secrets -A > /dev/null 2>&1 ; then
+    if kubectl --context $context get secrets -A > /dev/null 2>&1 ; then
         echov -e "${COLOR_GREEN}\xE2\x9C\x94${COLOR_NONE} ${COLOR_INFO}${context}${COLOR_NONE}"
-        valid_context[vc_n]="$context"
-        vc_n=$((vc_n+1))
+        valid_contexts+=($context)
     else
         echov -e "${COLOR_RED}X${COLOR_NONE} ${COLOR_INFO}${context}${COLOR_NONE}"
     fi
 done
-contexts=("${valid_context[@]}")
 
-if [[ -z "${contexts[@]}" ]] ; then
-    echoerr "no context(s) found in the kubeconfig file $kubeconfig!"
+if [[ -z "${valid_contexts[@]}" ]] ; then
+    echoerr "No valid context(s) found in the kubeconfig file $KUBECONFIG!"
     exit 1
 fi
 
@@ -294,24 +293,30 @@ fi
 ###############################################################################
 echotitle "KubeStellar:"
 helm_context=""
-for context in "${contexts[@]}" ; do # for all contexts
-    chart="$(helm --kubeconfig $kubeconfig --kube-context $context list 2> /dev/null | grep 'core-chart' || true)"
-    if [[ "$chart" != "" ]] ; then
-        helm_context=$context
-        name="$(echo "$chart" | awk '{print $1}')" || true
-        version="$(echo "$chart" | awk '{print $NF}')" || true
-        secret="$(k --context $helm_context get secret --all-namespaces -l "owner=helm"  --no-headers -o name 2> /dev/null | grep "$name" || true)"
+for context in "${valid_contexts[@]}" ; do
+    HELM_LIST="$TMPFOLDER/helm-list.txt"
+    helm --kube-context $context list --no-headers -A 2> /dev/null > "$HELM_LIST" || true
+    while IFS= read -r line; do
+        [[ -z "$line" ]] && continue
+        name="$(echo "$line" | awk '{print $1}')"
+        namespace="$(echo "$line" | awk '{print $2}')"
+        version="$(echo "$line" | awk '{print $NF}')"
+        secret="$(kubectl --context $context get secret -n "$namespace" -l "owner=helm,name=$name" --no-headers -o name 2> /dev/null || true)"
         secret="${secret##*/}"
-        echo -e "- Helm chart ${COLOR_INFO}${name}${COLOR_NONE} (${COLOR_INFO}v${version}${COLOR_NONE}) in context ${COLOR_INFO}${context}${COLOR_NONE}"
-        echo -e "  - Secret=${COLOR_INFO}${secret}${COLOR_NONE}"
-        if [[ "$arg_yaml" == "true" ]] ; then
-            mkdir -p "$output_folder/kubestellar-core-chart"
-            echo -e "$(k --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.info'     | sed -e 's/\"/"/g')" > "$output_folder/kubestellar-core-chart/info.json"
-            echo -e "$(k --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.chart'    | sed -e 's/\"/"/g')" > "$output_folder/kubestellar-core-chart/chart.json"
-            echo -e "$(k --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.manifest' | sed -e 's/\"/"/g')" > "$output_folder/kubestellar-core-chart/manifest.yaml"
+        if [[ ! -z "$(kubectl --context $context get secret $secret -n "$namespace" -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.chart' | grep -i "KubeStellar" 2> /dev/null || true)" ]] ; then
+            helm_context="$context"
+            echo -e "- Helm chart ${COLOR_INFO}${name}${COLOR_NONE} (${COLOR_INFO}v${version}${COLOR_NONE}) in namespace ${COLOR_INFO}${namespace}${COLOR_NONE} in context ${COLOR_INFO}${context}${COLOR_NONE}"
+            echo -e "  - Secret=${COLOR_INFO}${secret}${COLOR_NONE} in namespace ${COLOR_INFO}${namespace}${COLOR_NONE}"
+            if [[ "$arg_yaml" == "true" ]] ; then
+                mkdir -p "$OUTPUT_FOLDER/kubestellar-core-chart"
+                kubectl --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.info'      > "$OUTPUT_FOLDER/kubestellar-core-chart/info.json"
+                kubectl --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.chart'     > "$OUTPUT_FOLDER/kubestellar-core-chart/chart.json"
+                kubectl --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.manifest'  > "$OUTPUT_FOLDER/kubestellar-core-chart/manifest.yaml"
+            fi
+            break
         fi
-        break
-    fi
+    done < "$HELM_LIST"
+    [[ ! -z "$helm_context" ]] && break
 done
 if [[ "$helm_context" == "" ]] ; then
     echoerr "Helm chart was not found in any of the context(s)!"
@@ -323,30 +328,30 @@ fi
 # Look for Kubeflex deployment
 ###############################################################################
 echotitle "KubeFlex:"
-if [ -z "$(k --context $helm_context get ns kubeflex-system --no-headers -o name 2> /dev/null || true)" ] ; then
+if [[ -z "$(kubectl --context $helm_context get ns kubeflex-system --no-headers 2> /dev/null || true)" ]] ; then
     echoerr "KubeFlex namespace not found!"
 else
     echo -e "- ${COLOR_INFO}kubeflex-system${COLOR_NONE} namespace in context ${COLOR_INFO}$helm_context${COLOR_NONE}"
-    kubeflex_pod=$(k --context $helm_context -n kubeflex-system get pod -l "control-plane=controller-manager" -o name 2> /dev/null | cut -d'/' -f2 || true)
-    if [ -z "$kubeflex_pod" ]; then
+    kubeflex_pod=$(kubectl --context $helm_context -n kubeflex-system get pod -l "control-plane=controller-manager" -o name 2> /dev/null | cut -d'/' -f2 || true)
+    if [[ -z "$kubeflex_pod" ]]; then
         echoerr "KubeFlex pod not found!"
     else
-        kubeflex_version=$(k --context $helm_context -n kubeflex-system get pod $kubeflex_pod -o json 2> /dev/null | jq -r '.spec.containers[] | select(.image | contains("kubestellar/kubeflex/manager")) | .image' | cut -d':' -f2 || true)
-        kubeflex_status=$(k --context $helm_context -n kubeflex-system get pod $kubeflex_pod -o jsonpath='{.status.phase}' 2> /dev/null || true)
+        kubeflex_version=$(kubectl --context $helm_context -n kubeflex-system get pod $kubeflex_pod -o json 2> /dev/null | jq -r '.spec.containers[] | select(.image | contains("kubestellar/kubeflex/manager")) | .image' | cut -d':' -f2 || true)
+        kubeflex_status=$(kubectl --context $helm_context -n kubeflex-system get pod $kubeflex_pod -o jsonpath='{.status.phase}' 2> /dev/null || true)
         echo -n -e "- ${COLOR_INFO}controller-manager${COLOR_NONE}: version=${COLOR_INFO}$kubeflex_version${COLOR_NONE}, pod=${COLOR_INFO}$kubeflex_pod${COLOR_NONE}, status="
         echostatus "$kubeflex_status"
-        postgresql_pod=$(k --context $helm_context -n kubeflex-system get pod postgres-postgresql-0 2> /dev/null || true)
+        postgresql_pod=$(kubectl --context $helm_context -n kubeflex-system get pod postgres-postgresql-0 2> /dev/null || true)
         if [ -z "$postgresql_pod" ]; then
             echoerr "postgres-postgresql-0 pod not found!"
         else
-            postgresql_status=$(k --context $helm_context -n kubeflex-system get pod postgres-postgresql-0 -o jsonpath='{.status.phase}' 2> /dev/null || true)
+            postgresql_status=$(kubectl --context $helm_context -n kubeflex-system get pod postgres-postgresql-0 -o jsonpath='{.status.phase}' 2> /dev/null || true)
             echo -n -e "- ${COLOR_INFO}postgres-postgresql-0${COLOR_NONE}: pod=${COLOR_INFO}postgres-postgresql-0${COLOR_NONE}, status="
             echostatus "$postgresql_status"
         fi
         if [[ "$arg_logs" == "true" ]] ; then
-            mkdir -p "$output_folder/kubeflex"
-            k --context $helm_context -n kubeflex-system logs $kubeflex_pod > "$output_folder/kubeflex/kubeflex-controller.log"
-            k --context $helm_context -n kubeflex-system logs postgres-postgresql-0 > "$output_folder/kubeflex/postgresql.log"
+            mkdir -p "$OUTPUT_FOLDER/kubeflex"
+            kubectl --context $helm_context -n kubeflex-system logs $kubeflex_pod > "$OUTPUT_FOLDER/kubeflex/kubeflex-controller.log"
+            kubectl --context $helm_context -n kubeflex-system logs postgres-postgresql-0 > "$OUTPUT_FOLDER/kubeflex/postgresql.log"
         fi
     fi
 fi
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
Some comments on what is exhibited in the opening comment here.

1. Some of the format is almost YAML. Should it be actual YAML? Remember that YAML is flexible. For example, an array of strings can be represented by `[ "string1", "string2", "and", "so", "on" ]`.
2. Other parts of the output are almost tabular. I am also a big fan of tabular output. For example, the dump of ManifestWork contents could, I think, be very nicely represented by a table with columns for: ITS name, ManifestWork name, WEC name (equals namespace of the ManifestWork), WDS name, Binding name (equals BindingPolicy name but I think we should take Bindings as central), Kind, apiVersion, namespace of the workload object, name of the workload object.
3. I really think that Binding objects deserve at least as much attention as ManifestWork objects. The Binding objects are closer to the user and provide an important bit of feedback (without entanglement with other concerns): how the predicates in the BindingPolicy matched up against the available WECs and workload objects.
4. OTOH, for "must gather" functionality, I think we want comprehensive JSON or YAML (and I would prefer JSON) rather than stuff that is pretty and shortened for easy consumption. `kubectl get -o json --show-managed-fields`.

MikeSpreitzer:
```console
mspreitz@mjs13 kubestellar % echo $PWD
/Users/mspreitz/go/src/github.com/kubestellar/kubestellar

mspreitz@mjs13 kubestellar % tr <<<"$PWD" ABCDEFGHIJKLMNOPQRSTUVWXYZ abcdefghijklmnopqrstuvwxyz
/users/mspreitz/go/src/github.com/kubestellar/kubestellar
```

MikeSpreitzer:
Adding `set -x` to my copy of the script reveals where the time is going:

```
+ contexts=($(k config get-contexts --no-headers -o name))
++ k config get-contexts --no-headers -o name
++ KUBECONFIG=/Users/mspreitz/.kube/config
++ kubectl config get-contexts --no-headers -o name
+ [[ -z /api-prod-dipc-res-ibm-com:6443/4A5349897@W3IDRealm cluster1 cluster2 cp1 default/api-epoc2e-ete14-res-ibm-com:6443/kube:admin default/c100-e-us-east-containers-cloud-ibm-com:30969/IAM#mspreitz@us.ibm.com default/c100-e-us-south-containers-cloud-ibm-com:31320/IAM#mspreitz@us.ibm.com default/c100-e-us-south-containers-cloud-ibm-com:32111/IAM#mspreitz@us.ibm.com default/c106-e-us-south-containers-cloud-ibm-com:30263/IAM#mspreitz@us.ibm.com default/c106-e-us-south-containers-cloud-ibm-com:32494/IAM#mspreitz@us.ibm.com default/c107-e-us-south-containers-cloud-ibm-com:31803/IAM#mspreitz@us.ibm.com default/c109-e-us-east-containers-cloud-ibm-com:32541/IAM#mspreitz@us.ibm.com default/c114-e-us-south-containers-cloud-ibm-com:30286/IAM#mspreitz@us.ibm.com default/c114-e-us-south-containers-cloud-ibm-com:30301/IAM#mspreitz@us.ibm.com default/c114-e-us-south-containers-cloud-ibm-com:31314/IAM#mspreitz@us.ibm.com default/c115-e-us-south-containers-cloud-ibm-com:30049/IAM#mspreitz@us.ibm.com default/c115-e-us-south-containers-cloud-ibm-com:30486/IAM#mspreitz@us.ibm.com default/c115-e-us-south-containers-cloud-ibm-com:30829/IAM#mspreitz@us.ibm.com default/c115-e-us-south-containers-cloud-ibm-com:30847/IAM#mspreitz@us.ibm.com default/c115-e-us-south-containers-cloud-ibm-com:31498/IAM#mspreitz@us.ibm.com default/c115-e-us-south-containers-cloud-ibm-com:32320/IAM#mspreitz@us.ibm.com default/c116-e-us-south-containers-cloud-ibm-com:30708/IAM#mspreitz@us.ibm.com default/c117-e-us-south-containers-cloud-ibm-com:30490/IAM#mspreitz@us.ibm.com default/c118-e-us-south-containers-cloud-ibm-com:31192/IAM#mspreitz@us.ibm.com default/c118-e-us-south-containers-cloud-ibm-com:32614/IAM#mspreitz@us.ibm.com default/c130-e-us-south-containers-cloud-ibm-com:31342/IAM#mspreitz@us.ibm.com default/c130-e-us-south-containers-cloud-ibm-com:32596/IAM#mspreitz@us.ibm.com ep1 imbs1 imbs2 its1 its2 kind-kubeflex ks-edge-cluster1 ks-edge-cluster2 kscore kscore-stage postgres-test/c115-e-us-south-containers-cloud-ibm-com:32320/IAM#mspreitz@us.ibm.com rancher-desktop tsi-kube01/c20o1lbd0mhbu12dbd0g wds1 wds2 ]]
+ echov 'Using contexts(s): '
+ :
+ for context in '"${contexts[@]}"'
+ k --context /api-prod-dipc-res-ibm-com:6443/4A5349897@W3IDRealm get pods -A
```

There is a long pause at this point. I have a lot of obsolete contexts in my kubeconfig.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,554 @@
...
+    contexts=("$arg_context")
+fi
+
+if [[ -z "${contexts[@]}" ]] ; then
+    echoerr "no context(s) found in the kubeconfig file $kubeconfig!"
+    exit 1
+else
+    echov "Using contexts(s): "
+    for context in "${contexts[@]}" ; do # for all contexts
+        if k --context $context get pods -A  > /dev/null 2>&1 ; then
```
MikeSpreitzer, line 270: All of our current spaces have a resource named "pods" but that I something that we hope to change. The "namespace" resource, OTOH, I would expect in every space.

MikeSpreitzer:
After fixing the parameter expansion problem, I tested and got surprisingly empty results.

```console
mspreitz@mjs13 kubestellar % time scripts/kubestellar-snapshot.sh 

KubeStellar:
- Helm chart ks-core (v0.1.0) in context kind-kubeflex
  - Secret=sh.helm.release.v1.ks-core.v1

KubeFlex:
- kubeflex-system namespace in context kind-kubeflex
- controller-manager: version=0.7.1, pod=kubeflex-controller-manager-76f8db9767-m5dds, status=running
- postgres-postgresql-0: pod=postgres-postgresql-0, status=running

Control Planes:
- its1: type=vcluster, pch=its, context=kind-kubeflex, namespace=its1-system
  - Post Create Hook: pod=its-j5bjm, ns=its1-system, status=succeeded
  - Status addon: pod=addon-status-controller-d8864f4fd-wl9l5-x-open-clust-a298b3c05d, ns=its1-system, status=running
- wds1: type=k8s, pch=wds, context=kind-kubeflex, namespace=wds1-system
  - KubeStellar controller: version=a6f73c3db, pod=kubestellar-controller-manager-dc95df885-6vccs namespace=wds1-system, status=running
  - Transport controller: version=a6f73c3db, pod=transport-controller-d6994657b-nznhw namespace=wds1-system, status=running

Managed Clusters:

Binding Policies:

Manifest Works:

Work Statuses:
scripts/kubestellar-snapshot.sh  4.78s user 2.57s system 1% cpu 6:40.00 total
```

But just looking with kubectl finds stuff:

```console
mspreitz@mjs13 kubestellar % kubectl --context wds1 get bindings.control
NAME                  AGE
nginx-bindingpolicy   4h59m

mspreitz@mjs13 kubestellar % kubectl --context its1 get managedclusters
NAME       HUB ACCEPTED   MANAGED CLUSTER URLS                  JOINED   AVAILABLE   AGE
cluster2   true           https://cluster2-control-plane:6443   True     True        4h59m
cluster1   true           https://cluster1-control-plane:6443   True     True        4h59m

mspreitz@mjs13 kubestellar % kubectl --context its1 get manifestworks -A
NAMESPACE   NAME                          AGE
cluster1    nginx-bindingpolicy-wds1-0    4h59m
cluster1    nginx-bindingpolicy-wds1-1    4h59m
cluster2    nginx-bindingpolicy-wds1-0    4h59m
cluster2    nginx-bindingpolicy-wds1-1    4h59m
cluster1    addon-addon-status-deploy-0   4h59m
cluster2    addon-addon-status-deploy-0   4h59m

mspreitz@mjs13 kubestellar % kubectl --context its1 get WorkStatus -A   
NAMESPACE   NAME                                                                            AGE
cluster1    edd79b63-cc53-4fbe-9891-b1efe299314c-v1-namespace--nginx                        4h59m
cluster2    c670f1a6-1c8d-4aaa-bf96-eca98bdfa1a7-v1-namespace--nginx                        4h59m
cluster1    0742a339-c2d0-4e92-aee6-23a813b713cc-appsv1-deployment-nginx-nginx-deployment   4h59m
cluster2    cde59121-962e-459f-a8c4-948e510bcc48-appsv1-deployment-nginx-nginx-deployment   4h59m
```

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,554 @@
...
+arg_context=""
+arg_logs=false
+arg_yaml=false
+arg_verbose=false
+
+
+# Display the command line help
+display_help() {
+  cat << EOF
+Usage: $0 [--kubeconfig|-K <filename>] [--context|-C <name>] [--verbose|-V] [-X]
```
MikeSpreitzer, line 57: Shouldn't `-L`, `-v`, and `-Y` be here too?

scripts/kubestellar-snapshot.sh
MikeSpreitzer: Another thought: every Space will have the "secrets" resource. Ones that fail to list secrets will also fail (and taking another long time to fail, if the "get secrets" does) to do the `helm list` command later. So perhaps the better choice here is to try `get secrets` --- and also compile a list of contexts where that succeeded, and do the `helm list` only in those contexts.

MikeSpreitzer:
I did a little more investigating of why the script does not find any managed clusters and so on. Allowing the error messages to show up, I see error messages indicative of completely missing/wrong kubeconfig. I am reminded of a previous PR where you also tried the `kubectl --kubeconfig <(command)` thing and it completely failed for me. I think that eventually the PR was changed to work a different way.

francostellari:
> I tested on my laptop, MacOS 15.1, bash 3.2.57. It took a few minutes before outputting anything at all.
> 
> ```
> mspreitz@mjs13 kubestellar % scripts/kubestellar-snapshot.sh 
> 
> KubeStellar:
> - Helm chart ks-core (v0.1.0) in context kind-kubeflex
>   - Secret=sh.helm.release.v1.ks-core.v1
> 
> KubeFlex:
> - kubeflex-system namespace in context kind-kubeflex
> - controller-manager: version=0.7.1, pod=kubeflex-controller-manager-76f8db9767-m5dds, status=scripts/kubestellar-snapshot.sh: line 100: ${status,,}: bad substitution
> 
> mspreitz@mjs13 kubestellar % git rev-parse HEAD
> 4abed364465eb50bb8e37a9b48fb71a3e149a948
> ```
> 
> It looks like the script relies on a feature than is not in my `bash`:
> 
> ```
> bash-3.2$ echo "${PWD,,}"
> bash: ${PWD,,}: bad substitution
> ```

I moved to use `tr` to handle old bash versions

scripts/kubestellar-snapshot.sh
francostellari: I changed to use `secrets`

scripts/kubestellar-snapshot.sh
francostellari: I have reworked the help text

francostellari:
> I did a little more investigating of why the script does not find any managed clusters and so on. After editing the script to allow the error messages to show up, I see error messages indicative of completely missing/wrong kubeconfig. I vaguely recall that there was a previous PR where you also tried the `kubectl --kubeconfig <(command)` thing and it completely failed for me. I think that eventually the PR was changed to work a different way.

The only alternative would be to dump the in-memory kubeconfig contet to a tmp file and use KUBECONFIG env variable to point to it... It's more work, less secure

Can you investigate your source of this specific issue?

MikeSpreitzer:
FYI, Apple ships an old version of `bash` because of a copyright(left) dispute with the authors of `bash`; I do not expect a modern `bash` from Apple any time soon. While Mac users can install variant software, that is inherently dodgy; I would not recommend that we require this.

(I share your dismay. It means that we can not have nice things like: associative arrays, `@Q` in parameter expansion.)

MikeSpreitzer:
Here is a simple demonstration of the problem (producing the same error messages that I got from testing this PR's script after making it not discard the error messages).

```console
bash-3.2$ kubectl --kubeconfig ~/.kube/config get ns
NAME                                  STATUS   AGE
default                               Active   27h
kube-node-lease                       Active   27h
kube-public                           Active   27h
kube-system                           Active   27h
local-path-storage                    Active   27h
nginx                                 Active   27h
open-cluster-management               Active   27h
open-cluster-management-agent         Active   27h
open-cluster-management-agent-addon   Active   27h

bash-3.2$ kubectl --kubeconfig <(cat ~/.kube/config) get ns
E1115 15:37:02.330638    9646 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp 127.0.0.1:8080: connect: connection refused
E1115 15:37:02.331242    9646 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp 127.0.0.1:8080: connect: connection refused
E1115 15:37:02.332225    9646 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp 127.0.0.1:8080: connect: connection refused
E1115 15:37:02.332702    9646 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp 127.0.0.1:8080: connect: connection refused
E1115 15:37:02.333825    9646 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp 127.0.0.1:8080: connect: connection refused
The connection to the server localhost:8080 was refused - did you specify the right host or port?
```

MikeSpreitzer:
Not much help here:

```console
bash-3.2$ set -x
bash-3.2$ kubectl -v=10 --kubeconfig <(cat ~/.kube/config) get ns
+ kubectl -v=10 --kubeconfig /dev/fd/63 get ns
++ cat /Users/mspreitz/.kube/config
I1115 15:40:48.578262   10161 loader.go:395] Config loaded from file:  /dev/fd/63
I1115 15:40:48.578488   10161 round_trippers.go:466] curl -v -XGET  -H "Accept: application/json;g=apidiscovery.k8s.io;v=v2beta1;as=APIGroupDiscoveryList,application/json" -H "User-Agent: kubectl1.29.2/v1.29.2 (darwin/arm64) kubernetes/4b8e819" 'http://localhost:8080/api?timeout=32s'
...
```

MikeSpreitzer:
This script shows only expected behavior. I infer that the _way_ that the file is read makes the difference.

```shell
#!/usr/bin/env bash

set -x

echo Args are "<$@>"
echo arg 1 is "<$1>"
cat $1 > copy-of-1
```

Here is using it:

```console
bash-3.2$ ./exam.sh ~/.kube/config
+ echo Args are '</Users/mspreitz/.kube/config>'
Args are </Users/mspreitz/.kube/config>
+ echo arg 1 is '</Users/mspreitz/.kube/config>'
arg 1 is </Users/mspreitz/.kube/config>
+ cat /Users/mspreitz/.kube/config

bash-3.2$ ls -l copy-of-1 
-rw-r--r--@ 1 mspreitz  staff  77170 Nov 15 15:49 copy-of-1

bash-3.2$ ls -l ~/.kube/config
-rw-------@ 1 mspreitz  staff  77170 Nov 14 11:51 /Users/mspreitz/.kube/config

bash-3.2$ diff copy-of-1 ~/.kube/config
bash-3.2$ 
```

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+arg_context=""
+arg_logs=false
+arg_yaml=false
+arg_verbose=false
+
+
+# Display the command line help
+display_help() {
+  cat << EOF
+Usage: $0 [--kubeconfig|-K <filename>] [--context|-C <name>] [--logs|-L] [--yaml|-Y] [--verbose|-V] [--version|-v] [--help|-H] [-X]
```
MikeSpreitzer, line 57: Is it `-H` or `-h`? The latter is a fairly well established convention.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+
+--kubeconfig|-K <filename> use the specified kubeconfig
+--context|-C <name>        use the specified context
```
MikeSpreitzer, line 60: use it for what?

scripts/kubestellar-snapshot.sh
francostellari: a typo

scripts/kubestellar-snapshot.sh
francostellari: to search for KS Helm chart

francostellari:
@MikeSpreitzer what is your kubectl version?
```shell
 kubectl version
Client Version: v1.31.1
Kustomize Version: v5.4.2
Server Version: v1.31.0
```

francostellari:
> I will investigate the `--kubeconfig <(command)` issue. I also wonder (without having tried to study the script in detail) why it is necessary, when the user already has a kubeconfig file.

The script does not require `kflxe` or contexts for any CP.
Once it locates KubeStellar Helm chart, it will track down the CPs and other resouces, hence the need to deal with in memory kubeconfigs

francostellari:
> Here is a simple demonstration of the problem (producing the same error messages that I got from testing this PR's script after making it not discard the error messages).
> 
> ```
> bash-3.2$ kubectl --kubeconfig ~/.kube/config get ns
> NAME                                  STATUS   AGE
> default                               Active   27h
> kube-node-lease                       Active   27h
> kube-public                           Active   27h
> kube-system                           Active   27h
> local-path-storage                    Active   27h
> nginx                                 Active   27h
> open-cluster-management               Active   27h
> open-cluster-management-agent         Active   27h
> open-cluster-management-agent-addon   Active   27h
> 
> bash-3.2$ kubectl --kubeconfig <(cat ~/.kube/config) get ns
> E1115 15:37:02.330638    9646 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp 127.0.0.1:8080: connect: connection refused
> E1115 15:37:02.331242    9646 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp 127.0.0.1:8080: connect: connection refused
> E1115 15:37:02.332225    9646 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp 127.0.0.1:8080: connect: connection refused
> E1115 15:37:02.332702    9646 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp 127.0.0.1:8080: connect: connection refused
> E1115 15:37:02.333825    9646 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp 127.0.0.1:8080: connect: connection refused
> The connection to the server localhost:8080 was refused - did you specify the right host or port?
> ```

What about this:

```shell
kubectl --kubeconfig <(cat ~/.kube/config) config view --flatten > /tmp/dump.txt

diff /tmp/dump.txt ~/.kube/config
```

for me there is no difference observed

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+arg_context=""
+arg_logs=false
+arg_yaml=false
+arg_verbose=false
+
+
+# Display the command line help
+display_help() {
+  cat << EOF
+Usage: $0 [--kubeconfig|-K <filename>] [--context|-C <name>] [--logs|-L] [--yaml|-Y] [--verbose|-V] [--version|-v] [--help|-h] [-X]
```
MikeSpreitzer, line 57: As long as you are showing a listing of the options flags with descriptions, you do not need to repeat the flags in the synopsis. You could follow the pattern seen in the example below.

```console
mspreitz@mjs13 kubestellar % make -h
Usage: make [options] [target] ...
Options:
  -b, -m                      Ignored for compatibility.
  -B, --always-make           Unconditionally make all targets.
  -C DIRECTORY, --directory=DIRECTORY
                              Change to DIRECTORY before doing anything.
  -d                          Print lots of debugging information.
...
```

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+}
+
+
+# Get the kubeconfig of a particular Control Plane
+get_kubeconfig() {
+    context="$1" # context in the kubeconfig
+    cp_name="$2" # name of the Control Plane
+    cp_type="$3" # type of the Control Plane
+
+    # wait for CP ready
```
MikeSpreitzer, line 136: Should _this_ script wait? Wouldn't it be more suited to the purpose of this script to report that the CP is _not_ ready?

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+        echoerr "missing dependency:"
+        echo -e "${COLOR_RED}X${COLOR_NONE} $1"
+        exit 1
+    fi
+}
+
+
+# Get the kubeconfig of a particular Control Plane
+get_kubeconfig() {
+    context="$1" # context in the kubeconfig
```
MikeSpreitzer, line 132: Context in the _host_ kubeconfig. Worth clarifying _which_ kubeconfig because there are multiple in sight.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+    (*)
+        echo "$0: unknown positional argument" >&2
+        exit 1;;
+    esac
+    shift
+done
+
+
+###############################################################################
+# Define alias definition
```
MikeSpreitzer, line 199: plural

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+# Define the special kubectl
+k() {
+    KUBECONFIG="$kubeconfig" kubectl $*
+}
+
+
+###############################################################################
+# Script info
+###############################################################################
+echov -e "${COLOR_INFO}$($0 --version)${COLOR_NONE}\n"
```
MikeSpreitzer, line 217: Can this script be used by piping it into `bash`? If so then `$0` will refer to `bash` rather than the script.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+is_installed 'jq' \
+    'jq' \
+    'jq --version'
+
+
+###############################################################################
+# Ensure output folder
+###############################################################################
+if [[ "$arg_logs" == "true" || "$arg_yaml" == "true" ]] ; then
+    yes "yes" | rm -vRI "/tmp/kubestellar-snapshot" > /dev/null 2>&1 || true
```
MikeSpreitzer, line 243: Why not replace this whole line with simply `rm -rf /tmp/kubestellar-snapshot` ?

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+fi
+
+echov "Validating contexts(s): "
+vc_n=0
+valid_context=()
+for context in "${contexts[@]}" ; do # for all contexts
+    [[ -z "$context" ]] && continue
+    if k --context $context get secrets -A > /dev/null 2>&1 ; then
+        echov -e "${COLOR_GREEN}\xE2\x9C\x94${COLOR_NONE} ${COLOR_INFO}${context}${COLOR_NONE}"
+        valid_context[vc_n]="$context"
```
MikeSpreitzer, line 278: Why not simply assign to `valid_context[${#valid_context[*]}]` instead of explicitly maintaining a count?

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+        vc_n=$((vc_n+1))
+    else
+        echov -e "${COLOR_RED}X${COLOR_NONE} ${COLOR_INFO}${context}${COLOR_NONE}"
+    fi
+done
+contexts=("${valid_context[@]}")
+
+if [[ -z "${contexts[@]}" ]] ; then
+    echoerr "no context(s) found in the kubeconfig file $kubeconfig!"
```
MikeSpreitzer, line 287: No _valid_ contexts

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+    exit 1
+fi
+
+
+###############################################################################
+# Look for KubeStellar Helm chart
```
MikeSpreitzer, line 293: If this script had documentation then it would mention that the script only works if KubeStellar was installed via the core Helm chart.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+###############################################################################
+echotitle "KubeStellar:"
+helm_context=""
+for context in "${contexts[@]}" ; do # for all contexts
+    chart="$(helm --kubeconfig $kubeconfig --kube-context $context list 2> /dev/null | grep 'core-chart' || true)"
```
MikeSpreitzer, line 298: Is it necessarily true that the Helm "release" appear in the context's default namespace?

MikeSpreitzer, line 298: Ugh, "core-chart" is a little too short and generic to give me confidence that nobody else has that substring in a name.

We could at least make a tiny improvement by grepping for `'\<core-chart-'`.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+    if [[ "$chart" != "" ]] ; then
+        helm_context=$context
+        name="$(echo "$chart" | awk '{print $1}')" || true
+        version="$(echo "$chart" | awk '{print $NF}')" || true
+        secret="$(k --context $helm_context get secret --all-namespaces -l "owner=helm"  --no-headers -o name 2> /dev/null | grep "$name" || true)"
```
MikeSpreitzer, line 303: Maybe `-l owner=helm,name=$name` instead of searching with grep?

Why `--all-namespaces` here but not in the `helm list` command above?

scripts/kubestellar-snapshot.sh
MikeSpreitzer: The reader should be told a bit about the purpose.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+        secret="${secret##*/}"
+        echo -e "- Helm chart ${COLOR_INFO}${name}${COLOR_NONE} (${COLOR_INFO}v${version}${COLOR_NONE}) in context ${COLOR_INFO}${context}${COLOR_NONE}"
+        echo -e "  - Secret=${COLOR_INFO}${secret}${COLOR_NONE}"
+        if [[ "$arg_yaml" == "true" ]] ; then
+            mkdir -p "$output_folder/kubestellar-core-chart"
+            echo -e "$(k --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.info'     | sed -e 's/\"/"/g')" > "$output_folder/kubestellar-core-chart/info.json"
```
MikeSpreitzer, line 309: Why hack the JSON? Why use `echo -e`? Why not just do `k --context ... | jq '.info' > "$output_folder/kubestellar-core-chart/info.json"` ?

MikeSpreitzer:
```console
mspreitz@mjs13 kubestellar % kubectl version --client
Client Version: v1.29.2
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
```

MikeSpreitzer:
```console
mspreitz@mjs13 kubestellar % kubectl --kubeconfig <(cat ~/.kube/config) config view --flatten > /tmp/dump.txt

mspreitz@mjs13 kubestellar % wc /tmp/dump.txt 
       7      14     106 /tmp/dump.txt

mspreitz@mjs13 kubestellar % wc ~/.kube/config
     539    1067   77170 /Users/mspreitz/.kube/config
```

MikeSpreitzer:
Here is a demonstration of keeping temp files private to the user. (Remember that the user's kubeconfig, which is (at best) private to the user, gives access to anything this script finds.)

```console
bash-3.2$ umask 077

bash-3.2$ mktemp -p . -d
./tmp.c3ifC8QXym

bash-3.2$ date > tmp.c3ifC8QXym/date

bash-3.2$ ls -l
total 0
drwx------  3 mspreitz  staff  96 Nov 16 14:04 tmp.c3ifC8QXym

bash-3.2$ ls -l tmp.c3ifC8QXym/
total 8
-rw-------  1 mspreitz  staff  29 Nov 16 14:04 date
```

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+    echoerr "Helm chart was not found in any of the context(s)!"
+    exit 1
+fi
+
+
+###############################################################################
+# Look for Kubeflex deployment
+###############################################################################
+echotitle "KubeFlex:"
+if [ -z "$(k --context $helm_context get ns kubeflex-system --no-headers -o name 2> /dev/null || true)" ] ; then
```
MikeSpreitzer, line 326: Why `-o name` here?

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+        if [[ "${cp_type[cp_n]}" == "host" ]] ; then
+            status_ns="open-cluster-management"
+        else
+            status_ns="${cp_ns[cp_n]}"
+        fi
+        status_pod=$(k --context $helm_context -n "$status_ns" get pod -o name 2> /dev/null | grep addon-status-controller | cut -d'/' -f2 || true)
+        status_status=$(k --context $helm_context -n "$status_ns" get pod $status_pod -o jsonpath='{.status.phase}' 2> /dev/null || true)
+        echo -n -e "  - Post Create Hook: pod=${COLOR_INFO}$its_pod${COLOR_NONE}, ns=${COLOR_INFO}${cp_ns[cp_n]}${COLOR_NONE}, status="
+        echostatus "$its_status"
+        echo -n -e "  - Status addon: pod=${COLOR_INFO}$status_pod${COLOR_NONE}, ns=${COLOR_INFO}$status_ns${COLOR_NONE}, status="
```
MikeSpreitzer, line 382: Status addon _controller_:

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+        echostatus "$status_status"
+    else
+        kubestellar_pod=$(k --context $helm_context -n "${cp_ns[cp_n]}" get pod -l "control-plane=controller-manager" -o name 2> /dev/null | cut -d'/' -f2 || true)
+        kubestellar_version=$(k --context $helm_context -n "${cp_ns[cp_n]}" get pod $kubestellar_pod -o json 2> /dev/null | jq -r '.spec.containers[] | select(.image | contains("kubestellar/controller-manager")) | .image' | cut -d':' -f2 || true)
+        kubestellar_status=$(k --context $helm_context -n "${cp_ns[cp_n]}" get pod $kubestellar_pod -o jsonpath='{.status.phase}' 2> /dev/null || true)
+        echo -e -n "  - KubeStellar controller: version=${COLOR_INFO}$kubestellar_version${COLOR_NONE}, pod=${COLOR_INFO}$kubestellar_pod${COLOR_NONE} namespace=${COLOR_INFO}"${cp_ns[cp_n]}"${COLOR_NONE}, status="
```
MikeSpreitzer, line 388: Kubestellar controller _manager_:

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+
+
+###############################################################################
+# Listing binding policies
+###############################################################################
+echotitle "Binding Policies:"
+bp_n=0
+for j in "${!cp_pch[@]}" ; do
+    if [[ "${cp_pch[$j]}" == "wds" ]] ; then
+        bps=($(kubectl --kubeconfig <(echo "${cp_kubeconfig[$j]}") get bindingpolicy -no-headers -o name 2> /dev/null || true))
```
MikeSpreitzer, line 460: Why is an array being built? I only see use of the current (i.e., currently last) element.

scripts/kubestellar-snapshot.sh
francostellari: ok, I have changed it to the proposed way to simplify maintenance

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+    fi
+}
+
+
+# Check if a pre-requisite is installed
+is_installed() {
+    # $1 == name
+    # $2 == command name to search
+    # $3 == command to get the version, unstructured
+    if which $2 > /dev/null ; then
```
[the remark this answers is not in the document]
francostellari: I don't think this code requires any specific version for the moment, so I don't think checking for a version make sense yet.

scripts/kubestellar-snapshot.sh
francostellari: I have removed the waiting. Now the script prints a warning and will not return impacted resources.

scripts/kubestellar-snapshot.sh
francostellari: I changed to `context="$1" # context in the host kubeconfig or the config passed to the script using --kubeconfig flag`

scripts/kubestellar-snapshot.sh
francostellari: I changed to `# Alias definitions`

scripts/kubestellar-snapshot.sh
francostellari: I changed to ``${BASH_SOURCE[0]}`

scripts/kubestellar-snapshot.sh
francostellari: I remove the line all together since I need a temp folder for KUBECONFIG

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+# Determine the list of kubeconfigs to search
+# based on https://kubernetes.io/docs/reference/k/generated/k_config/
+###############################################################################
+if [[ "$arg_kubeconfig" != "" ]] ; then
+    kubeconfig="$arg_kubeconfig"
+elif [[ "$KUBECONFIG" != "" ]] ; then
+    kubeconfig="$KUBECONFIG"
+else
+    kubeconfig="$HOME/.kube/config"
+fi
```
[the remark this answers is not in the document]
francostellari: Thank you for the good suggstion.
I have removed the alias `k`  and changed this code to:
```shell
[[ -n "$arg_kubeconfig" ]] && export KUBECONFIG="$arg_kubeconfig"
[[ -z "$KUBECONFIG" ]] && export KUBECONFIG="$HOME/.kube/config"
echov -e "Using kubeconfig(s): ${COLOR_INFO}$KUBECONFIG${COLOR_NONE}"
```

scripts/kubestellar-snapshot.sh
francostellari: I reworked the code to remove the index

scripts/kubestellar-snapshot.sh
francostellari: changed to `    echoerr "No valid context(s) found in the kubeconfig file $KUBECONFIG!"
`

scripts/kubestellar-snapshot.sh
francostellari: I'll add a mention but there is no other way to install Kubestellar, right?

francostellari:
```shell
Note 1: This script expects that KubeStellar was installed with its Helm chart and
        to find the latter in the host or provided kubeconfig(s) and context(s).
```

scripts/kubestellar-snapshot.sh
francostellari: I have changed it to search all namespaces

francostellari: At a cost of more complexity, I have  changed the code to analyze all the Helm charts found to look for the work "Kubestellar" inside the chart template.
Yes, this is also not a fool proof guarantee,
At the moment I cannot figure out a unique identifier, perhaps we could look for the controller image repo full path.

scripts/kubestellar-snapshot.sh
francostellari: ok, thank you

scripts/kubestellar-snapshot.sh
francostellari: Thank you for catching this.
It must be a leftover of when I was trying to display something for debug.

scripts/kubestellar-snapshot.sh
francostellari: ok

scripts/kubestellar-snapshot.sh
francostellari: It was done this way to match the other sections.
It's true that it's not currently being used but I would rather keep it for possible future if it's not a big deal

MikeSpreitzer: No, it is not a big deal.

scripts/kubestellar-snapshot.sh
MikeSpreitzer: We do not yet support any other way but I suspect that we might in the future. For example, Issue #2577 is someone asking for general setup instructions and we have not yet proven that the core Helm chart will work for them. For another example, recall the ambition to be able to add KubeStellar to a pre-existing OCM system without any KubeFlex already being there nor being added; that would differ more than a little from the current pattern (as well as require enabling changes in the controller codes).

scripts/kubestellar-snapshot.sh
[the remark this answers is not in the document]
MikeSpreitzer: I am sure that there are versions of some of the dependencies that would be problematic. But I am not sure that these are likely to actually be in use.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,579 @@
...
+    cp_context[cp_n]=$helm_context
+    cp_name[cp_n]=$name
+    cp_ns[cp_n]="${cp_name[cp_n]}-system"
+    cp_type[cp_n]=$(kubectl --context $helm_context get controlplane ${cp_name[cp_n]} -o jsonpath='{.spec.type}')
+    cp_pch[cp_n]=$(kubectl --context $helm_context get controlplane ${cp_name[cp_n]} -o jsonpath='{.spec.postCreateHook}')
+    cp_kubeconfig_content=$(get_kubeconfig "${helm_context}" "${cp_name[cp_n]}" "${cp_type[cp_n]}")
+    echo -e "- ${COLOR_INFO}${cp_name[cp_n]}${COLOR_NONE}: type=${COLOR_INFO}${cp_type[cp_n]}${COLOR_NONE}, pch=${COLOR_INFO}${cp_pch[cp_n]}${COLOR_NONE}, context=${COLOR_INFO}${cp_context[cp_n]}${COLOR_NONE}, namespace=${COLOR_INFO}${cp_name[cp_n]}-system${COLOR_NONE}"
+    if [[ -z "$cp_kubeconfig_content" ]] ; then
+        cp_kubeconfig[cp_n]=""
+        echo -e "  ${COLOR_WARNING}WARNING: the Control plane is no ready, the kubeconfig is not available!${COLOR_NONE}"
```
MikeSpreitzer, line 377: s/no/not/

scripts/kubestellar-snapshot.sh
MikeSpreitzer: Not a big deal, but: I would not write "host kubeconfig" --- because, as an adjective, "host" (or, as I prefer, "hosting" or "KubeFlex hosting") applies to a context rather than a kubeconfig. For your purpose here I would simply write "relevant kubeconfig" or "applicable kubeconfig" or "chosen kubeconfig".

scripts/kubestellar-snapshot.sh
MikeSpreitzer: I don't think that is much of an improvement. When a user does `curl $something | bash`, that invocation of `bash` knows nothing about the URL that stdin was fetched from.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,579 @@
...
+# limitations under the License.
+
+
+# Exit on error
+set -e
+
+
+# Global variables
+TIMESTAMP="$(date +%F_%T)"
+TMPFOLDER="$(mktemp -d --tmpdir "kubestellar-snapshot-XXXX")"
```
MikeSpreitzer, line 23: Was the omission of `-p .` here deliberate? Without `-p .`, the temp dir gets made in `/tmp`, which is probably less secure than the user's current working directory.

scripts/kubestellar-snapshot.sh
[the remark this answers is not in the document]
MikeSpreitzer: FYI, you could omit the second command in that trio and do the defaulting directly in the third (which is the only place where you need to explicitly duplicate that logic). The third command becomes the following.

```shell
echov -e "Using kubeconfig(s): ${COLOR_INFO}${KUBECONFIG:-$HOME/.kube/confg}${COLOR_NONE}"
```

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,579 @@
...
+            if [[ "$arg_yaml" == "true" ]] ; then
+                mkdir -p "$OUTPUT_FOLDER/kubestellar-core-chart"
+                kubectl --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.info'      > "$OUTPUT_FOLDER/kubestellar-core-chart/info.json"
+                kubectl --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.chart'     > "$OUTPUT_FOLDER/kubestellar-core-chart/chart.json"
+                kubectl --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.manifest'  > "$OUTPUT_FOLDER/kubestellar-core-chart/manifest.yaml"
+            fi
+            break
+        fi
+    done < "$HELM_LIST"
+    [[ ! -z "$helm_context" ]] && break
```
MikeSpreitzer, line 319: This seems to be assuming that the user can not have two KubeStellar cores.

scripts/kubestellar-snapshot.sh
MikeSpreitzer: What I see being searched is a JSON representation of the entire chart, in which the templates appear encoded in base64 --- so the search is not going to find "kubestellar" in the templates. OTOH, I _do_ see "kubestellar" in the `.metadata.description`, `.metadata.icon`, `.metadata.dependencies`, and `.metadata.lock` inside that JSON. So, for example, you could narrow that `jq` to `jq .chart.metadata.description`.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,567 @@
...
+            available="$(kubectl --kubeconfig <(echo "${cp_kubeconfig[$j]}") get managedcluster $name -o jsonpath='{.status.conditions}' | jq '.[] | select(.type == "ManagedClusterConditionAvailable") | .status' | tr -d '"')"
+            synced="$(kubectl --kubeconfig <(echo "${cp_kubeconfig[$j]}") get managedcluster $name -o jsonpath='{.status.conditions}' | jq '.[] | select(.type == "ManagedClusterConditionClockSynced") | .status' | tr -d '"')"
+            echo -e "- ${COLOR_INFO}${mc_name[mc_n]}${COLOR_NONE} in ${COLOR_INFO}${cp_name[j]}${COLOR_NONE}: accepted=$(echostatus $accepted), joined=$(echostatus $joined), available=$(echostatus $available), synced=$(echostatus $synced)"
+            if [[ "$arg_verbose" == "true" ]] ; then
+                echo -n -e "${COLOR_YAML}"
+                kubectl --kubeconfig <(echo "${cp_kubeconfig[$j]}") get managedclusters $name -o jsonpath='{.metadata.labels}' | jq '. |= with_entries(select(.key|(contains("open-cluster-management")|not)))' | indent
```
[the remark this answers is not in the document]
MikeSpreitzer: Oh, sorry, I overlooked the fact that you defined your own `indent`. I was thinking of the standard utility `indent`, which is designed for the C language.

scripts/kubestellar-snapshot.sh
francostellari: KS require several cluster level non-namespaced resources such as KubeFlex, PCHs, CP CRDs, etc. So installing a second KS chart may fail or would create conflicts and unpredictable behaviors.. The only supported way to installa a "second" chart is to add ITSes and WDSes by disabling the installation of cluster resources such as KubeFlex, PCHs. In this scenario, the script would still work by enumerating the total resources because both chart would be in the same context. Alas the second chart would not be clearly listed in the KS section.
Based on the above considerations, I suggest to leave as is.

scripts/kubestellar-snapshot.sh
francostellari: I made a choice to use tmp folder but if you rather use the cwd, I can change to it.

scripts/kubestellar-snapshot.sh
francostellari: I'm not sure that even in that case will will actually not adapt to use the helm chart.
Furthermore, I'm not sure how the resources structure will look like so I suggest that we revise the script when we know more.

scripts/kubestellar-snapshot.sh
francostellari: ok, I changed to
```
echov -e "${COLOR_INFO}${SCRIPT_NAME} v${SCRIPT_VERSION}{COLOR_NONE}\n"
```

scripts/kubestellar-snapshot.sh
eaepstein: What is the scenario calling for multiple KS cores? Is the idea that multiple KS cores could be sharing OCM hubs? Given a KS scheduler that must keep track of used and available WEC resources, two KS cores-schedulers would have to synchronize.

MikeSpreitzer: @francostellari , @eaepstein: remember kubeconfig "contexts". Maybe you do not have as many as I do. I have a lot. Roughly, I have one for every cluster or space that I have ever worked with. Different contexts can correspond to completely unrelated and different things. While a beginning user might only kick the tires once, someone developing KubeStellar or a system incorporating KubeStellar is not unlikely to work with multiple systems. For example, I have worked no only with the kind/k3d demo systems that we have automated but also multiple systems involving our ETE clusters. I have indeed simultaenously had a KubeStellar system in those two domains.

francostellari: @MikeSpreitzer I understand your concern but the idea here is to take a snapshot of "a" KS deployment among the many that you may have across different systems. It's up to the user to provide a meaningful kubeconfig/context to the script in that case.  All data reported is only meaningful for a specific helm deployment. The script would have to add another loop to run itself on every KS helm deployments that it finds.

An alternative would be to list all of the Helm charts that it finds everywhere and then as the user to choose if more than one. To me it makes more sense that the user chooses by passing the --kubeconfig/--context

francostellari: Perhaps one thing where the script fall short in the effort to help the user when running the script in cluster1/cluster2 context by mistake is that it does not respect the current user context. So it should check the current context first before looking into other contexts.

scripts/kubestellar-snapshot.sh
MikeSpreitzer: Indeed, it would be premature to make the script do anything else now. My point here is only that this specificity should be documented.

scripts/kubestellar-snapshot.sh
MikeSpreitzer: OK, we can start with a script that aims to work on only one KubeStellar system. We may want future work that deals with all the ones found in a given kubeconfig. A start would be to just get a list of all the contexts that have a KubeStellar core installed in them.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,580 @@
...
+
+
+# Exit on error
+set -e
+
+
+# Global variables
+TIMESTAMP="$(date +%F_%T)"
+TMPFOLDER="$(mktemp -d -p . "kubestellar-snapshot-XXXX")"
+trap 'rm -rf -- "$TMPFOLDER"' EXIT
```
MikeSpreitzer, line 24: FYI, the `--` is unnecessary because we know that the temp folder name does not start with a dash.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,580 @@
...
+--context|-C <name>        use the specified context to find KubeStellar Helm chart
+--logs|-L                  save the logs of the pods
+--yaml|-Y                  save the YAML of the resources
+--verbose|-V               output extra information
+--version|-v               print out the script version
+--help|-h                  show this information
+-X                         enable verbose execution for debugging
+
+Note 1: This script expects that KubeStellar was installed with its Helm chart and
+        to find the latter in the host or provided kubeconfig(s) and context(s).
```
MikeSpreitzer, line 73: s/to/tries to/
s/in the host or provided kubeconfig(s) and context(s)/in a context of the relevant kubeconfig/

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,580 @@
...
+
+Note 2: After piping the output of the script to file use "more"
+        to inspect the file content in color. ALternatively install
+        "ansi2txt" and pipe the script through it to remove escape
+        characters and obtain a plan text report.
+EOF
+}
+
+
+# Indent JSON
```
MikeSpreitzer, line 83: FYI, it looks to me like this does a constant indent of anything.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,580 @@
...
+if [[ -z "${contexts[@]}" ]] ; then
+    echoerr "No context(s) found in the kubeconfig file $KUBECONFIG!"
+    exit 1
+fi
+
+echov "Validating contexts(s): "
+valid_contexts=()
+for context in "${contexts[@]}" ; do # for all contexts
+    [[ -z "$context" ]] && continue
+    [[ "$context" == "$(kubectl config current-context || true)" ]] && cc="*" || cc=""
```
MikeSpreitzer, line 277: FYI: `$(kubectl config current-context || true)` does not need to be evaluated on every iteration, it could be done once before the loop.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,580 @@
...
+    if kubectl --context $context get secrets -A > /dev/null 2>&1 ; then
+        echov -e "${COLOR_GREEN}\xE2\x9C\x94${COLOR_NONE} ${COLOR_INFO}${context}${COLOR_NONE} ${COLOR_GREEN}$cc${COLOR_NONE}"
+        [[ -z "$cc" ]] && valid_contexts+=("$context") || valid_contexts=("$context" "${valid_contexts[@]}")
+    else
+        echov -e "${COLOR_RED}X${COLOR_NONE} ${COLOR_INFO}${context}${COLOR_NONE} ${COLOR_GREEN}$cc${COLOR_NONE}"
+    fi
+done
+
+if [[ -z "${valid_contexts[@]}" ]] ; then
+    echoerr "No valid context(s) found in the kubeconfig file $KUBECONFIG!"
```
MikeSpreitzer, line 287: FYI, I find the wording "valid context(s)" to be a bit misleading. "Validity" of contexts is something that I expect the unwashed user to suppose is defined by Kubernetes. I think it would be better to say something like "No KubeStellar core installation found in any context in the kubeconfig file $KUBECONFIG".

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,580 @@
...
+fi
+
+
+###############################################################################
+# Look for KubeStellar Helm chart
+###############################################################################
+echotitle "KubeStellar:"
+helm_context=""
+for context in "${valid_contexts[@]}" ; do
+    HELM_LIST="$TMPFOLDER/helm-list.txt"
```
MikeSpreitzer, line 298: This is constant, it could be defined once before the loop.

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,580 @@
...
+            echo -e "  - Secret=${COLOR_INFO}${secret}${COLOR_NONE} in namespace ${COLOR_INFO}${namespace}${COLOR_NONE}"
+            if [[ "$arg_yaml" == "true" ]] ; then
+                mkdir -p "$OUTPUT_FOLDER/kubestellar-core-chart"
+                kubectl --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.info'      > "$OUTPUT_FOLDER/kubestellar-core-chart/info.json"
+                kubectl --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.chart'     > "$OUTPUT_FOLDER/kubestellar-core-chart/chart.json"
+                kubectl --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.manifest'  > "$OUTPUT_FOLDER/kubestellar-core-chart/manifest.yaml"
+            fi
+            break
+        fi
+    done < "$HELM_LIST"
```
MikeSpreitzer, line 319: This could be `<<< $helm_list` (take stdin from the content of a shell variable) rather than `< $HELM_LIST`, saving the trouble to define a special filename (in addition to the shell variable).

francostellari: ok, I started to be worried about what works in old bashes

scripts/kubestellar-snapshot.sh
francostellari: removed

scripts/kubestellar-snapshot.sh
francostellari: This is not related to KubeStellar, it's trying to remove configs from the list that are not working.
If you hit this erro, there is nowhere to search

scripts/kubestellar-snapshot.sh
francostellari: yes

scripts/kubestellar-snapshot.sh
francostellari: I have changed the script, hopefully to show all the charts, but it will only use the first one

scripts/kubestellar-snapshot.sh
francostellari: I think it is

scripts/kubestellar-snapshot.sh
```diff
@@ -0,0 +1,585 @@
...
+                    mkdir -p "$OUTPUT_FOLDER/kubestellar-core-chart"
+                    kubectl --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.info'      > "$OUTPUT_FOLDER/kubestellar-core-chart/info.json"
+                    kubectl --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.chart'     > "$OUTPUT_FOLDER/kubestellar-core-chart/chart.json"
+                    kubectl --context $helm_context get secret $secret -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq -r '.manifest'  > "$OUTPUT_FOLDER/kubestellar-core-chart/manifest.yaml"
+                fi
+            else
+                more_than_one="true"
+            fi
+        fi
+    done <<< "$(helm --kube-context $context list --no-headers -A 2> /dev/null || true)"
```
MikeSpreitzer, line 322: Ah, right, no shell variable needed.
FYI, this could be even simpler: `(helm ...true) | while ... done`.

MikeSpreitzer:
It tested the current rev of this PR, commit bb2b5843a50b9891047f83debb645ac6f108ccc3.

I first built a system with `test/e2e/run-test.sh --released`. Then I ran the snapshot script. I have attached a typscript of running the script. Unless I missed something, it worked as intended.

[snap-trial-4.log](https://github.com/user-attachments/files/17909147/snap-trial-4.log)

scripts/kubestellar-snapshot.sh
MikeSpreitzer: My point here is that the comment is overly specific. Not a big deal.

scripts/kubestellar-snapshot.sh
MikeSpreitzer: So something like "broken" or "malfunctioning" would be more on-point than "[not] valid". "valid" often means something more shallow about syntax and self-consistency. Not a big deal.

MikeSpreitzer (approved):
This is good enough to merge. More work can be done in follow-on PRs. Documentation is needed.

/lgtm
/approve
Thanks!

MikeSpreitzer:
This script prints the `.status.phase` of every WorkStatus object, regardless of whether there _is_ a `phase` defined in the schema of the relevant `.status`.

PR #2584, opened 2024-11-12
📖  Various enhancements to the demo env creation script doc

MikeSpreitzer:
## Summary
This PR attempts to make a few improvements to the documentation of the demo environment creation script.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-update-1112b/direct/get-started/

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/direct/get-started.md
+++ b/docs/content/direct/get-started.md
@@ -1,38 +1,39 @@
 # Getting Started with KubeStellar
 
-There are multiple ways to get started with Kubestellar, reflecting its flexible architecture that supports various deployment patterns. This page presents two approaches:
+This page shows one concrete example of steps 2--7 from the [full Installation and Usage outline](user-guide-intro.md#the-full-story). This example produces a simple single-host system that is suitable for kicking the tires, using [kind](https://kind.sigs.k8s.io/) to create three new clusters to serve as your KubeFlex hosting cluster and two WECs. This page concludes with forwarding you to one example of the remaining steps. For general setup information, see [the full story](user-guide-intro.md#the-full-story).
 
-1. A **Step by step walkthrough** that demonstrates the core concepts and components, showing how to manually set up a simple single-host system.
-
-2. A **quick automated setup** using our installation script, which creates a basic working environment for those who want to start experimenting right away. 
-
-Both approaches create a basic environment suitable for learning, but remember that KubeStellar supports more sophisticated configurations including:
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
/cc @KPRoche 
/cc @francostellari 
/cc @clubanderson 
@katamyra

docs/content/direct/get-started.md
```diff
@@ -1,38 +1,39 @@
 # Getting Started with KubeStellar
 
-There are multiple ways to get started with Kubestellar, reflecting its flexible architecture that supports various deployment patterns. This page presents two approaches:
+This page shows one concrete example of steps 2--7 from the [full Installation and Usage outline](user-guide-intro.md#the-full-story). This example produces a simple single-host system that is suitable for kicking the tires, using [kind](https://kind.sigs.k8s.io/) to create three new clusters to serve as your KubeFlex hosting cluster and two WECs. This page concludes with forwarding you to one example of the remaining steps. For general setup information, see [the full story](user-guide-intro.md#the-full-story).
```
clubanderson, line 3: Why is it a KubeFlex hosting cluster?  That's confusing. This is a KubeStellar Core and that language is consistent with (or should be) with the rest of our documentation

clubanderson (commented):
There is way too much text to describe an automated script. It takes longer to read all this then to run it.

docs/content/direct/get-started.md
MikeSpreitzer: This is not new text, it has just been moved around a bit.

This text deliberately speaks the same language as the cited "full story". The full story gets into the internal structure of what we call "the core", because the full story is that the consumer of KubeStellar has choices for what they put in their core.

MikeSpreitzer:
> If this is getting started then it should be ...

The documentation about the new script has been injected into the pre-existing step-by-step "Getting Started" document.

KPRoche:
I thought I had sent this a while ago:

I would suggest adding a What's Next section before the Troubleshooting section and moving a chunk of this to that section,  structuring the page so the simplest content opens it and gory details are available further down.

Also the Detailed Step Walkthrough section repeats a bunch of text and seems out of sync headings-wise with the hierarchy in the rest of the page (as manifested in the TOC on the right).

MikeSpreitzer:
@KPRoche: I like that outline. I created another PR, #2590, to replace this one using that outline. I had a little different opinion about some of the words, but kept your outline.

PR #2588, opened 2024-11-13
🐛  Make setup-kubestellar overwrite kflex hidden state

### LINKED ISSUE #2587: bug: test/e2e/run-test.sh fails when kflex hidden state is wrong
### Describe the bug

Testing `main` (when that points to commit 9ce248e08d5980f33206db2c96381d32e4c1958b), I did the following.

- First I tried https://docs.kubestellar.io/release-0.25.0/direct/start-from-ocm/  and example Scenarios 1 and 3; got success.
- then I tried `test/e2e/run-test.sh` and it failed to create the hosting cluster because another cluster is already using host port 9443.
- Then I deleted that other cluster.
- Then I retried `test/e2e/run-test.sh`, and this time it failed because the hosting cluster extension in my kubeconfig file still existed and held a now-incorrect value.

I have attached a typescript.

[kflex-fail.txt](https://github.com/user-attachments/files/17739182/kflex-fail.txt)

### Steps To Reproduce

Described above.

### Expected Behavior

The hosting cluster extension should have been overwritten with the correct value.

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR fixes the setup-kubestellar.sh script to unconditionally overwrite any pre-existing setting in kflex's kubeconfig extension that holds the name of the hosting cluster context.

## Related issue(s)

Fixes #2587

### DIFF
--- a/test/e2e/common/setup-kubestellar.sh
+++ b/test/e2e/common/setup-kubestellar.sh
@@ -135,10 +135,11 @@ wait-for-cmd "(kubectl --context '$HOSTING_CONTEXT' -n wds1-system wait --for=co
 
 echo "transport controller is running."
 
-kubectl config delete-context its1 || true
-kflex ctx its1
-kubectl config delete-context wds1 || true
-kflex ctx wds1
+kubectl config use-context "$HOSTING_CONTEXT"
+kflex ctx --set-current-for-hosting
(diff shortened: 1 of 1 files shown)


clubanderson (commented):
/lgtm
/approve

PR #2589, opened 2024-11-14
🐛  Revise generic transport controller update logic

### LINKED ISSUE #2511: bug: Setting max_num_wrapped=2 in transport_controller does not work correctly in version 0.25.0-rc.1
### Describe the bug

Scenario:
- configure transport with max_num_wrapped=2
- create pods 1 .. 5
- delete pods 1 .. 5
- track pods and manifestwork "pod contents" after each create and delete

The observed behavior is shown in the attached console log:
[num-wrapped-2.txt](https://github.com/user-attachments/files/17436844/num-wrapped-2.txt)

In summary, on deletions pods are duplicated in multiple manifestworks with the result that pods deleted from WDS1 are left in one or more manifestworks and thus also in the managed cluster. In addition, the deletion of pod1 causes multiple manifestworks to be recreated, which seems costly and could lead to synchronization issues.

### Steps To Reproduce

1. create a k8s host cluster as per kubestellar examples
2. configure max_num_wrapped=2 by adding the following argument when starting the KS core using helm. For example:
```
export KUBESTELLAR_VERSION=0.25.0-rc.1
helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart   --version $KUBESTELLAR_VERSION  --set-json='ITSes=[{"name":"its1"}]'  --set-json='WDSes=[{"name":"wds1"}]' --set transport_controller.max_num_wrapped=2
```
3. create and join a managed cluster as per kubestellar examples
4. create a binding policy that will assign work to the managed cluster
5. create pod yaml files pod1 .. pod5 that will be covered by the binding policy
6. in WDS1, start pods named pod1 .. pod5 and then delete them in order pod1 .. pod5

### Expected Behavior

- Pods should be deleted from the managed cluster when they are deleted from WDS1. 
- reordering of many manifestworks on a single pod deletion could be a major performance problem

### Want to contribute?

### Additional Context

_No response_

### LINKED ISSUE #2586: bug: Objects deleted from the Hub are not removed from the WEC
### Describe the bug

I am running on k3d with the latest KS 0.25.0 release. 
On the HUB I've created BP and a ConfigMap with a label matching the BP.
The CM was pulled to the WEC. I then deleted CM from the HUB.
The CM was deleted from the HUB but lingers on the WEC

The Bindings object does not contain deleted CM. In its1 however there is a manifestworks object which does contain the CM

The CM I used:

apiVersion: v1
kind: ConfigMap
metadata:
  name: game-demo
  labels:
    kubestellar.io/cluster: cluster2
data:
  # property-like keys; each key maps to a simple value
  player_initial_lives: "3"
  ui_properties_file_name: "user-interface.properties"

  # file-like keys
  game.properties: |
    enemy.types=aliens,monsters
    player.maximum-lives=5
  user-interface.properties: |
    color.good=purple
    color.bad=yellow
    allow.textmode=true

This is a regression from v.0.25.0-rc1 which does not have this problem. 
Attached are files: BindingPolicy, Bindings,ManifestWorks
[binding.txt](https://github.com/user-attachments/files/17738740/binding.txt)
[mw2.txt](https://github.com/user-attachments/files/17738746/mw2.txt)
[bp2.txt](https://github.com/user-attachments/files/17738753/bp2.txt)

### Steps To Reproduce

See ^^^

### Expected Behavior

Deleted object should be cleaned up from the WEC

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR changes the way that the transport controller decides what changes it has to make in wrapped objects. The old code was not logical.

## Related issue(s)

Fixes #2586 
Fixes #2511

### DIFF
--- a/pkg/transport/generic/generic_transport_controller.go
+++ b/pkg/transport/generic/generic_transport_controller.go
@@ -94,7 +94,7 @@ func NewTransportController(ctx context.Context,
 	transportClientset kubernetes.Interface,
 	transportDynamicClient dynamic.Interface,
 	maxSizeWrapped int, maxNumWrapped int, wdsName string) (*genericTransportController, error) {
-	emptyWrappedObject := transportInstance.WrapObjects(make([]transport.Wrapee, 0)) // empty wrapped object to get GVR from it.
+	emptyWrappedObject := transportInstance.WrapObjects(make([]transport.Wrapee, 0), nil) // empty wrapped object to get GVR from it.
 	wrappedObjectGVR, err := getGvrFromWrappedObject(transportClientset, emptyWrappedObject)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get wrapped object GVR - %w", err)
@@ -653,8 +653,13 @@ func (c *genericTransportController) deleteWrappedObjectsAndFinalizer(ctx contex
 }
 
 func (c *genericTransportController) deleteWrappedObject(ctx context.Context, namespace string, objectName string) error {
+	logger := klog.FromContext(ctx)
 	err := c.transportClient.Resource(c.wrappedObjectGVR).Namespace(namespace).Delete(ctx, objectName, metav1.DeleteOptions{})
-	if err != nil && !errors.IsNotFound(err) { // if object is already not there, we do not report an error cause desired state was achieved.
+	if err == nil {
+		logger.V(2).Info("Successful deletion of wrapped object", "wrappedObjectName", objectName, "destination", namespace)
+	} else if errors.IsNotFound(err) {
+		logger.V(2).Info("Deletion of wrapped object was not needed because it is already gone", "wrappedObjectName", objectName, "destination", namespace)
+	} else {
 		return fmt.Errorf("failed to delete wrapped object '%s' from destination WEC mailbox namespace '%s' - %w", objectName, namespace, err)
 	}
 	return nil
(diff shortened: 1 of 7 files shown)


MikeSpreitzer:
/cc @vMaroon 
/cc @waltforme 
@cwiklik

pkg/transport/generic/generic_transport_controller.go
```diff
@@ -1022,49 +1039,51 @@ func (c *genericTransportController) customizeForDestination(object *unstructure
...
-			desiredWrappedObjects, _ := destToDesiredWrappedObject(destination)
-			if currentWrappedObject != nil {
-				for _, desiredWrappedObject := range desiredWrappedObjects {
-					// Can't use apiequality.Semantic.DeepEqual to compare the two objects
-					if currentWrappedObject.GetAnnotations() != nil &&
-						currentWrappedObject.GetAnnotations()[originOwnerGenerationAnnotation] == desiredWrappedObject.GetAnnotations()[originOwnerGenerationAnnotation] {
-						foundMatch = true
-						break
-					}
+		tts, _ := destToDesiredWrappedObjects(destination)
```
vMaroon, line 1057: tts?

MikeSpreitzer:
The force-push to c5f5e2ed0 is a squash and a couple of renamings in response to the review comment above.

MikeSpreitzer:
.. and the force-push to 25fb9204e fixes the comment on the type `Gloss`.

MikeSpreitzer:
And the force-push to 4644e56f5 fixes the comment on `computeDestToWrappedObjects`

MikeSpreitzer:
.. and the force-push to 2eb730f58 does a couple of local variable renamings to bring them up to date.

MikeSpreitzer:
PR #2598 demonstrates good behavior with the limit on number of workload objects in a bundle set to 2.

/approve

vMaroon:
@MikeSpreitzer I think it may make sense to document (on a fine-enough granularity) the logic of the transport component as it is getting complexly distributed.

PR #2590, opened 2024-11-15
📖  Reorganize Getting Started Page

MikeSpreitzer:
## Summary
This PR reorganizes the Getting Started page a bit, to follow @KPRoche's outline shown in https://github.com/kubestellar/kubestellar/pull/2584#issuecomment-2476820172 . This is good because it puts the fast path first without reducing what can be learned from the whole page.

This version also draws on the version in release 0.25.0 (https://docs.kubestellar.io/release-0.25.0/direct/get-started/).

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-demo-script/direct/get-started/

## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/direct/get-started.md
+++ b/docs/content/direct/get-started.md
@@ -1,46 +1,42 @@
 # Getting Started with KubeStellar
 
-There are multiple ways to get started with Kubestellar, reflecting its flexible architecture that supports various deployment patterns. This page presents two approaches:
+## Set Up A Demo System
 
-1. A **Step by step walkthrough** that demonstrates the core concepts and components, showing how to manually set up a simple single-host system.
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
/cc @KPRoche
/cc @francostellari
/cc @clubanderson
@katamyra

docs/content/direct/get-started.md
```diff
@@ -137,6 +133,30 @@ wec2_context=$wec2_name
 label_query_both=location-group=edge
 label_query_one=name=cluster1
 ```
+
+## Exercise KubeStellar
+
+Now that your system is running, you can try some example scenarios
+
+1. Define the needed shell variables, using either the settings output as the script completes or the settings shown just above from the step-by-step instructions). Their meanings are defined [at the start of the example scenarios document](example-scenarios.md#assumptions-and-variables).
```
KPRoche, line 141: Minor typo here: there is an orphan closing parenthesis after "instructions"

MikeSpreitzer:
The force-push to cb14584e0 fixes the extra paren noted above.

docs/content/direct/get-started.md
MikeSpreitzer: Thanks. Fixed.

PR #2597, opened 2024-11-18
✨  Define API for flexible trichotomy of workload objects

MikeSpreitzer:
## Summary
This PR defines a new API that gives the user control over the classification of workload object content in terms of which direction (if any) it will propagate.

This PR currently only has the API and doc updates. This is what should be reviewed first.

## Related issue(s)

Fixes #


api/control/v1alpha1/types.go
```diff
@@ -665,3 +665,56 @@ type CustomTransformList struct {
...
+// ContentClassifierSpec specifies the overrides.
+// Currently there are limitations:
+// 1. You can not add to the reported state;
+// 2. You can not declare any part of `.status` to be desired state.
+// Invalid overrides will be reported in `.status.errors` and otherwise ignored.
+// Currently the overrides are expressed by identifying object parts
+// by JSON Pointers (RFC 6901).
+type ContentClassifierSpec struct {
+	DesiredStatePointers  []ParsedJSONPointer `json:"desiredStatePointers,omitempty"`
+	DontPropagatePointers []ParsedJSONPointer `json:"dontPropagatePointers,omitempty"`
```
vMaroon, line 704: Two points:
1. Can both fields be set? If I understood what they specify, then the answer is no. This should be clarified in the struct comment/docs.
2. The name `dontPropagatePointers` is more instructional than declarative in nature. Perhaps it can be changed to something declarative like `restrictedPaths` (or both fields can be declarative counterparts).

MikeSpreitzer: 1. These two fields hold _overrides_ to the default classification; an object field not covered by any override is classified in the default way. Yes, one kind of object can have both categories of overrides.
2. While we are on the subject of naming, I should point out that "desired state" and "reported state" do not get to the relevant point, which is about propagation.

clubanderson:
This design PR has been open as WIP for 14 months. To keep the repository clean while preserving the discussion, this has been converted to a GitHub Discussion.

**Continue the conversation here:** https://github.com/kubestellar/kubestellar/discussions/3623

Thank you @MikeSpreitzer for the design work. The discussion can be revisited when there's renewed interest in the flexible trichotomy API.

PR #2621, opened 2024-11-21
✨ add option to create kubestellar on k3d

clubanderson:
## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/direct/get-started.md
+++ b/docs/content/direct/get-started.md
@@ -25,9 +25,16 @@ Be sure to [install the software prerequisites](pre-reqs.md) _before_ running th
 The script will check for the pre-reqs and exit if they are not present.
 
 ### Run the script!
+The script can install KubeStellar's demonstration environment on top of Kind or k3d
 
+For Kind
 ```shell
-bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/tags/v{{ config.ks_latest_release }}/scripts/create-kubestellar-demo-env.sh)
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/tags/v{{ config.ks_latest_release }}/scripts/create-kubestellar-demo-env.sh) --platform kind
+```
+
+For k3d
+```shell
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/tags/v{{ config.ks_latest_release }}/scripts/create-kubestellar-demo-env.sh) --platform k3d
 ```
 
 If successful, the script will output the variable definitions that you would use when proceeding to the example scenarios. After successfully running the script, proceed to the [Exercise KubeStellar](#exercise-kubestellar) section below.
--- a/docs/content/direct/get-started.md
+++ b/docs/content/direct/get-started.md
@@ -25,14 +25,14 @@ Be sure to [install the software prerequisites](pre-reqs.md) _before_ running th
 The script will check for the pre-reqs and exit if they are not present.
 
 ### Run the script!
-The script can install KubeStellar's demonstration environment on top of Kind or k3d
+The script can install KubeStellar's demonstration environment on top of kind or k3d
 
-For Kind
+For use with kind
 ```shell
 bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/tags/v{{ config.ks_latest_release }}/scripts/create-kubestellar-demo-env.sh) --platform kind
 ```
 
-For k3d
+For use with k3d
 ```shell
 bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/tags/v{{ config.ks_latest_release }}/scripts/create-kubestellar-demo-env.sh) --platform k3d
 ```
--- a/docs/content/direct/get-started.md
+++ b/docs/content/direct/get-started.md
@@ -34,7 +34,7 @@ bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/ta
 
 For use with k3d
 ```shell
-bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/tags/v{{ config.ks_latest_release }}/scripts/create-kubestellar-demo-env.sh) --platform k3d
+bash <(curl -s https://raw.githubusercontent.com/clubanderson/kubestellar/refs/heads/add-k3d-to-create-demo-env/scripts/create-kubestellar-demo-env.sh) --platform k3d
 ```
 
 If successful, the script will output the variable definitions that you would use when proceeding to the example scenarios. After successfully running the script, proceed to the [Exercise KubeStellar](#exercise-kubestellar) section below.
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -117,20 +117,45 @@ checking_cluster() {
 
     done
 }
+
+# Function to check if a command exists
+command_exists() {
+    command -v "$1" >/dev/null 2>&1
+}
 ##########################################
 
 echo -e "\nStarting environment clean up..."
 echo -e "Starting cluster clean up..."
 
+# Check for k8s_platform and perform cleanup in the correct order
 if [ "$k8s_platform" == "kind" ]; then
+    # Check if k3d exists and delete clusters in k3d first
+    if command_exists "k3d"; then
+        cluster_clean_up "k3d cluster delete kubeflex" &
+        cluster_clean_up "k3d cluster delete cluster1" &
+        cluster_clean_up "k3d cluster delete cluster2" &
+        wait
+    fi
+    # Then delete clusters in kind
     cluster_clean_up "kind delete cluster --name kubeflex" &
     cluster_clean_up "kind delete cluster --name cluster1" &
     cluster_clean_up "kind delete cluster --name cluster2" &
+    wait
 else
+    # Check if kind exists and delete clusters in kind first
+    if command_exists "kind"; then
+        cluster_clean_up "kind delete cluster --name kubeflex" &
+        cluster_clean_up "kind delete cluster --name cluster1" &
+        cluster_clean_up "kind delete cluster --name cluster2" &
+        wait
+    fi
+    # Then delete clusters in k3d
     cluster_clean_up "k3d cluster delete kubeflex" &
     cluster_clean_up "k3d cluster delete cluster1" &
     cluster_clean_up "k3d cluster delete cluster2" &
+    wait
 fi
+
 wait
 echo -e "\033[33m✔\033[0m Cluster space clean up has been completed"
 
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -24,6 +24,7 @@ k8s_platform="kind"
 while [[ "$#" -gt 0 ]]; do
     case $1 in
         --platform) k8s_platform="$2"; shift ;;
+        -X) set -x ;;
         -h|--help)
             echo "Usage: $0 [--platform <kind|k3d>]"
             exit 0
@@ -59,8 +60,7 @@ echo -e "Checking that pre-req softwares are installed..."
 if [ "$k8s_platform" == "kind" ]; then
     curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/hack/check_pre_req.sh | bash -s -- -V kflex ocm helm kubectl docker kind
 else
-    ../hack/check_pre_req.sh -V kflex ocm helm kubectl docker k3d
-    # curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/hack/check_pre_req.sh | bash -s -- -V kflex ocm helm kubectl docker k3d
+    curl -s https://raw.githubusercontent.com/clubanderson/kubestellar/refs/heads/add-k3d-to-create-demo-env/hack/check_pre_req.sh | bash -s -- -V kflex ocm helm kubectl docker k3d
 fi
 
 ##########################################
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -13,9 +13,37 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Deploys Kubestellar environment for demo purposes.
+# Deploys Kubestellar environment for demo purposes on kind or k3d.
 
 set -e
+
+# Default Kubernetes platform parameter
+k8s_platform="kind"
+
+# Parse command-line arguments
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --platform) k8s_platform="$2"; shift ;;
+        -h|--help)
+            echo "Usage: $0 [--platform <kind|k3d>]"
+            exit 0
+            ;;
+        *)
+            echo "Unknown parameter passed: $1"
+            echo "Usage: $0 [--platform <kind|k3d>]"
+            exit 1
+            ;;
+    esac
+    shift
+done
+
+if [[ "$k8s_platform" != "kind" && "$k8s_platform" != "k3d" ]]; then
+    echo "Invalid platform specified: $k8s_platform"
+    echo "Supported platforms are: kind, k3d"
+    exit 1
+fi
+
+echo "Selected Kubernetes platform: $k8s_platform"
 
 echo -e "Checking container runtime..."
 if ! dunsel=$(docker ps 2>&1); then
@@ -28,7 +56,12 @@ kubestellar_version=0.25.1
 echo -e "KubeStellar Version: ${kubestellar_version}"
 
 echo -e "Checking that pre-req softwares are installed..."
-curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/hack/check_pre_req.sh | bash -s -- -V kflex ocm helm kubectl docker kind
+if [ "$k8s_platform" == "kind" ]; then
+    curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/hack/check_pre_req.sh | bash -s -- -V kflex ocm helm kubectl docker kind
+else
+    # ../hack/check_pre_req.sh -V kflex ocm helm kubectl docker k3d
+    curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/hack/check_pre_req.sh | bash -s -- -V kflex ocm helm kubectl docker k3d
+fi
 
 ##########################################
 cluster_clean_up() {
@@ -89,9 +122,15 @@ checking_cluster() {
 echo -e "\nStarting environment clean up..."
 echo -e "Starting cluster clean up..."
 
-cluster_clean_up "kind delete cluster --name kubeflex" &
-cluster_clean_up "kind delete cluster --name cluster1" &
-cluster_clean_up "kind delete cluster --name cluster2" &
+if [ "$k8s_platform" == "kind" ]; then
+    cluster_clean_up "kind delete cluster --name kubeflex" &
+    cluster_clean_up "kind delete cluster --name cluster1" &
+    cluster_clean_up "kind delete cluster --name cluster2" &
+else
+    cluster_clean_up "k3d cluster delete kubeflex" &
+    cluster_clean_up "k3d cluster delete cluster1" &
+    cluster_clean_up "k3d cluster delete cluster2" &
+fi
 wait
 echo -e "\033[33m✔\033[0m Cluster space clean up has been completed"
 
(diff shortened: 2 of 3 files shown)


MikeSpreitzer:
There also needs to be a doc update.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -13,10 +13,38 @@
...
+# Deploys Kubestellar environment for demo purposes on kind or k3d.
 
 set -e
 
+# Default Kubernetes platform parameter
+k8s_platform="kind"
+
+# Parse command-line arguments
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
```
MikeSpreitzer, line 25: As long as we are parsing command line arguments, we should accept `-X` too.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -28,7 +56,12 @@ kubestellar_version=0.25.1
 echo -e "KubeStellar Version: ${kubestellar_version}"
 
 echo -e "Checking that pre-req softwares are installed..."
-curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/hack/check_pre_req.sh | bash -s -- -V kflex ocm helm kubectl docker kind
+if [ "$k8s_platform" == "kind" ]; then
+    curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/hack/check_pre_req.sh | bash -s -- -V kflex ocm helm kubectl docker kind
+else
+    ../hack/check_pre_req.sh -V kflex ocm helm kubectl docker k3d
```
MikeSpreitzer, line 62: (A) using a local file does not work when this script is used in curl-to-bash style.

(B) This particular way of navigating to a local file adds an undocumented assumption about current working directory.

MikeSpreitzer: You could make this testable before merging by curl-to-bash of https://raw.githubusercontent.com/clubanderson/kubestellar/refs/heads/add-k3d-to-create-demo-env/hack/check_pre_req.sh

clubanderson: that worked - will check in the update

scripts/create-kubestellar-demo-env.sh
clubanderson: added

clubanderson:
@MikeSpreitzer Thank you - I do not see the offending container in your docker ps

MikeSpreitzer:
I did another trial of commit 61f3d79ff2ab3b76125e7dd85c6ee7cd84aa0fdf, this time first making sure there was no debris laying around. Even so, there were error messages printed during cluster creation. However, the script still claimed success and I was able to complete example scenario 1.

[k3d-trial5.log](https://github.com/user-attachments/files/17875459/k3d-trial5.log)

clubanderson:
yes, I have seen errors as well. I think they are specific to ks core creation and kubeflex.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -84,35 +117,78 @@ checking_cluster() {
...
+}
 ##########################################
 
 echo -e "\nStarting environment clean up..."
 echo -e "Starting cluster clean up..."
 
-cluster_clean_up "kind delete cluster --name kubeflex" &
-cluster_clean_up "kind delete cluster --name cluster1" &
-cluster_clean_up "kind delete cluster --name cluster2" &
+# Check for k8s_platform and perform cleanup in the correct order
```
MikeSpreitzer, line 130: What difference does the order make?

clubanderson: if creating kind you need to delete k3s first. If creating k3s you need to create kind first

clubanderson:
preview doc here - https://clubanderson.github.io/kubestellar/add-k3d-to-create-demo-env/direct/get-started/#run-the-script

scripts/create-kubestellar-demo-env.sh
```diff
@@ -84,35 +117,78 @@ checking_cluster() {
...
 done
 
 echo -e "Creating KubeFlex cluster with SSL Passthrough"
-curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/scripts/create-kind-cluster-with-SSL-passthrough.sh | bash -s -- --name kubeflex --nosetcontext
+if [ "$k8s_platform" == "kind" ]; then
+    curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/scripts/create-kind-cluster-with-SSL-passthrough.sh | bash -s -- --name kubeflex --nosetcontext
+else
+    k3d cluster create -p "9443:443@loadbalancer" --k3s-arg "--disable=traefik@server:*" kubeflex
+    sleep 15
+    helm install ingress-nginx ingress-nginx --set "controller.extraArgs.enable-ssl-passthrough=" --repo https://kubernetes.github.io/ingress-nginx --version 4.6.1 --namespace ingress-nginx --create-namespace
```
MikeSpreitzer, line 185: According to https://github.com/kubernetes/ingress-nginx/blob/main/README.md#supported-versions-table , it looks to me like we should be using version 4.11.3 of this Helm chart.

(And scripts/create-kind-cluster-with-SSL-passthrough.sh should probably use a specific version rather than `main`, but that's a separate PR.)

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: What goes wrong if I use a different order?

clubanderson: On delete. Nothing. On create it causes problems. The comment will be removed.

clubanderson: And the logic updates. Don't need two loops

scripts/create-kubestellar-demo-env.sh
clubanderson: fixed

scripts/create-kubestellar-demo-env.sh
clubanderson: fixed

scripts/create-kubestellar-demo-env.sh
```diff
@@ -13,10 +13,39 @@
...
+# Default Kubernetes platform parameter
+k8s_platform="kind"
+
+# Parse command-line arguments
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --platform) k8s_platform="$2"; shift ;;
+        -X) set -x ;;
+        -h|--help)
+            echo "Usage: $0 [--platform <kind|k3d>]"
```
MikeSpreitzer, line 29: I would mention all the options here.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -146,26 +208,42 @@ images=("ghcr.io/loft-sh/vcluster:0.16.4"
...
+    helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart \
+        --version $kubestellar_version \
+        --set-json='ITSes=[{"name":"its1"}]' \
+        --set-json='WDSes=[{"name":"wds1"}]' \
+        --set-json='verbosity.default=5'
+else
+    helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart \
+        --version $kubestellar_version \
+        --set-json='ITSes=[{"name":"its1"}]' \
+        --set-json='WDSes=[{"name":"wds1"},{"name":"wds2", "type":"host"}]' \
```
MikeSpreitzer, line 231: Why make two WDSes in this case and not in the `kind` case? I think that it would be best if this script behaved analogously in the two cases.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -84,35 +117,64 @@ checking_cluster() {
...
+        cluster_clean_up "k3d cluster delete cluster2" &
+        wait
+    fi
+    if command_exists "kind"; then
+        cluster_clean_up "kind delete cluster --name kubeflex" &
+        cluster_clean_up "kind delete cluster --name cluster1" &
+        cluster_clean_up "kind delete cluster --name cluster2" &
+        wait
+    fi
+fi
```
MikeSpreitzer, line 143: Maybe the thrust of my earlier comments was not clear. I do not debate the need for cleanup of from both platforms. I was asking whether/why the _order of_ the cleanup matters.

Without the cleanup of `kind` clusters when creating `k3d` clusters, I get a failure due to port 9443 already being taken. Following is an excerpt from a trial.

```
Creating two k3d clusters to serve as example WECs
Creating KubeFlex cluster with SSL Passthrough
INFO[0000] portmapping '9443:443' targets the loadbalancer: defaulting to [servers:*:proxy agents:*:proxy] 
INFO[0000] Prep: Network                                
INFO[0000] Re-using existing network 'k3d-kubeflex' (2bce747eb25b10de393f0ce44d360ee9bf90b076d5d22de7166211e66f563170) 
INFO[0000] Created image volume k3d-kubeflex-images     
INFO[0000] Starting new tools node...                   
INFO[0000] Starting node 'k3d-kubeflex-tools'           
INFO[0001] Creating node 'k3d-kubeflex-server-0'        
INFO[0001] Creating LoadBalancer 'k3d-kubeflex-serverlb' 
INFO[0001] Using the k3d-tools node to gather environment information 
INFO[0001] HostIP: using network gateway 172.22.0.1 address 
INFO[0001] Starting cluster 'kubeflex'                  
INFO[0001] Starting servers...                          
INFO[0001] Starting node 'k3d-kubeflex-server-0'        
INFO[0004] All agents already running.                  
INFO[0004] Starting helpers...                          
INFO[0004] Starting node 'k3d-kubeflex-serverlb'        
ERRO[0004] Failed Cluster Start: Failed to add one or more helper nodes: runtime failed to start node 'k3d-kubeflex-serverlb': docker failed to start container for node 'k3d-kubeflex-serverlb': Error response from daemon: driver failed programming external connectivity on endpoint k3d-kubeflex-serverlb (dc3dd5564bd8cbae1d7c1b7193c1f860012e71e5f81501267af15467870e4333): Bind for 0.0.0.0:9443 failed: port is already allocated 
ERRO[0004] Failed to create cluster >>> Rolling Back    
INFO[0004] Deleting cluster 'kubeflex'                  
INFO[0004] Deleting 1 attached volumes...               
FATA[0004] Cluster creation FAILED, all changes have been rolled back! 
```

scripts/create-kubestellar-demo-env.sh
```diff
@@ -146,26 +208,42 @@ images=("ghcr.io/loft-sh/vcluster:0.16.4"
 
 for image in "${images[@]}"; do
     (
-        docker pull "$image" && kind load docker-image "$image" --name kubeflex
+        if [ "$k8s_platform" == "kind" ]; then
+            docker pull "$image" && kind load docker-image "$image" --name kubeflex
+        else
+            docker pull "$image" && k3d image import "$image" --cluster kubeflex
```
MikeSpreitzer, line 214: Looking at the error messages from my trial5, I suspect that the problem is the attempt to do two concurrent image loads into the same cluster. Here is an excerpt from that trial5 log:

```
INFO[0000] Importing image(s) into cluster 'kubeflex'   
INFO[0000] Importing image(s) into cluster 'kubeflex'   
INFO[0000] Starting new tools node...                   
INFO[0000] Starting new tools node...                   
ERRO[0000] Failed to run tools container for cluster 'kubeflex' 
ERRO[0000] Failed to import image(s) into cluster 'kubeflex': failed to ensure that tools node is running: failed to run k3d-tools node for cluster 'kubeflex': failed to create node 'k3d-kubeflex-tools': runtime failed to create node 'k3d-kubeflex-tools': failed to create container for node 'k3d-kubeflex-tools': docker failed to create container 'k3d-kubeflex-tools': Error response from daemon: Conflict. The container name "/k3d-kubeflex-tools" is already in use by container "01f565a9d5465e6f555ef464bc1dd341f030d346e38005a38d895bf5c2b78525". You have to remove (or rename) that container to be able to reuse that name. 
WARN[0000] At least one error occured while trying to import the image(s) into the selected cluster(s) 
```

My hypothesis is consistent with the fact that these errors do not cause the script to fail: a failure to pre-load an image is only a failed optimization, not a correctness fail.

I expect that it would be fine to do the `k3d image import` commands in series after `wait`ing for all the parallel `docker pull`s to finish.

scripts/create-kubestellar-demo-env.sh
clubanderson: done

scripts/create-kubestellar-demo-env.sh
clubanderson: thats correct, that is why both cleanups are handled here in this control block. Cleanup is done here to remove debris before trying to create new clusters of any sort. So there is no code change you are proposing, your just asking for info?

scripts/create-kubestellar-demo-env.sh
clubanderson: updated for both kind and k3s to first pull then load into platform

scripts/create-kubestellar-demo-env.sh
clubanderson: done

docs/content/direct/get-started.md
```diff
@@ -25,9 +25,16 @@ Be sure to [install the software prerequisites](pre-reqs.md) _before_ running th
...
 
+For use with kind
 ```shell
-bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/tags/v{{ config.ks_latest_release }}/scripts/create-kubestellar-demo-env.sh)
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/tags/v{{ config.ks_latest_release }}/scripts/create-kubestellar-demo-env.sh) --platform kind
+```
+
+For use with k3d
+```shell
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/tags/v{{ config.ks_latest_release }}/scripts/create-kubestellar-demo-env.sh) --platform k3d
```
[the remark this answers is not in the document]
clubanderson: when it is approved then we can take that temp link out

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: The latest revision addresses my concern.

scripts/create-kubestellar-demo-env.sh
clubanderson: the second WDS is needed for the Kueue integration. We can either go back to making 1 for kind and 2 for k3d or keep 2 wds for each. I see no harm in leaving 2 in each version

docs/content/direct/get-started.md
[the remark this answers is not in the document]
MikeSpreitzer: Approval and even merging will not fix the problem, which is instructions telling the reader to use a URL that is 404. The problem will go away after this PR is merged and then a release is made.

This is not a blocker, just an FYI. If you wanted to fix it so that someone could succeed in the meantime, I think that it could be done as follows.

1. Make this PR say to use a URL that points into the `main` branch rather than the latest release.
2. Update the steps in`docs/content/direct/release.md` for making the next release to include (a) updating the URL here to refer to the latest release and (b) removing this step from that list of steps.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: Is this script becoming the install script for all purposes?

MikeSpreitzer: I think we want the "getting started" configuration to not be more complicated than necessary. If the Kueue integration requires a more complicated setup then I think that should be documented but in a way that does not make the "getting started" configuration more complicated.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -84,35 +117,62 @@ checking_cluster() {
...
 done
 
 echo -e "Creating KubeFlex cluster with SSL Passthrough"
-curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/scripts/create-kind-cluster-with-SSL-passthrough.sh | bash -s -- --name kubeflex --nosetcontext
+if [ "$k8s_platform" == "kind" ]; then
+    curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/scripts/create-kind-cluster-with-SSL-passthrough.sh | bash -s -- --name kubeflex --nosetcontext
+else
+    k3d cluster create -p "9443:443@loadbalancer" --k3s-arg "--disable=traefik@server:*" kubeflex
+    sleep 15
+    helm install ingress-nginx ingress-nginx --set "controller.extraArgs.enable-ssl-passthrough=" --repo https://kubernetes.github.io/ingress-nginx --version 4.6.1 --namespace ingress-nginx --create-namespace
```
MikeSpreitzer, line 169: I think there was an unintentional slip back to the older version of the ingress chart here.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -146,26 +206,47 @@ images=("ghcr.io/loft-sh/vcluster:0.16.4"
...
-    --set-json='WDSes=[{"name":"wds1"}]' \
-    --set-json='verbosity.default=5'
+echo -e "\nStarting the process to install KubeStellar core: $k8s_platform-kubeflex..."
+if [ "$k8s_platform" == "k3d" ]; then
+    helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart \
+        --version $kubestellar_version \
+        --set-json='ITSes=[{"name":"its1"}]' \
+        --set-json='WDSes=[{"name":"wds1"},{"name":"wds2", "type":"host"}]' \
+        --set-json='verbosity.default=5' \
+        --set "kubeflex-operator.hostContainer=k3d-kubeflex-server-0"
```
MikeSpreitzer, line 229: FYI, you don't need the dquotes on this line. They are harmless here, but killed the previous commit.

scripts/create-kubestellar-demo-env.sh
clubanderson: yep - was trying to see if it helped with the CSR stuff - ill put it back now

MikeSpreitzer:
I tested commit 98d64ff1e05b2a52f21c05395b4394beadad4948 on my Mac and it worked.

[k3d-trial7.txt](https://github.com/user-attachments/files/18029750/k3d-trial7.txt)

scripts/create-kubestellar-demo-env.sh
clubanderson: the intent of this script is to be a baseline for the getting started and the integrations. We should not need an entire set of scripts to run each time we want to run our own getting started or anything in galaxy repo. This script creates the basic "platform" that is KubeStellar. You can experiment with more than just the getting started with this new baseline.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: I still do not see all the options listed here.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -13,10 +13,39 @@
...
+    case $1 in
+        --platform) k8s_platform="$2"; shift ;;
+        -X) set -x ;;
+        -h|--help)
+            echo "Usage: $0 [--platform <kind|k3d>]"
+            exit 0
+            ;;
+        *)
+            echo "Unknown parameter passed: $1"
+            echo "Usage: $0 [--platform <kind|k3d>] [-X] [-h|--help]"
```
MikeSpreitzer, line 34: FYI, you could use `>&2` to send these error messages to stderr.

scripts/]
MikeSpreitzer:1]: Is this intentional?

docs/content/direct/get-started.md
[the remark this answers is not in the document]
MikeSpreitzer: It looks like you are planning the three-step approach that I suggested.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: If we expand the purpose of this script beyond establishing the configuration used in "get started" then that breaks the equivalence between this script and the step-by-step instructions in "get started". That means that this script needs distinct documentation of its own and the "get startede" page has to explain the difference between what this script does and what the step-by-step instructions do.

The expansion suggested here is still only for kicking tires, not production usage. So I would not say that the configuration produced by this script is 'the basic "platform" that is KubeStellar'.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -146,26 +206,47 @@ images=("ghcr.io/loft-sh/vcluster:0.16.4"
...
+    helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart \
+        --version $kubestellar_version \
+        --set-json='ITSes=[{"name":"its1"}]' \
+        --set-json='WDSes=[{"name":"wds1"},{"name":"wds2", "type":"host"}]' \
+        --set-json='verbosity.default=5'
+fi
 
 kflex ctx --set-current-for-hosting # make sure the KubeFlex CLI's hidden state is right for what the Helm chart just did
 kflex ctx --overwrite-existing-context wds1
+if [ "$k8s_platform" == "k3d" ]; then
```
MikeSpreitzer, line 240: You no longer intend this to be conditional, right?

scripts/]
clubanderson:1]: no, removed

scripts/create-kubestellar-demo-env.sh
clubanderson: added

scripts/create-kubestellar-demo-env.sh
clubanderson: this doesn't break the original getting started - there is an extra WDS - that is all

scripts/create-kubestellar-demo-env.sh
clubanderson: your right - removed control block

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: I understand that. My point is that adding extra functionality to this script causes structural problems. The functionality that we expect users to use should be documented. Currently the only documentation of this script is the get-started page, which only documents one WDS. Also, that page asserts equivalence between the script and the step-by-step instructions. See https://github.com/kubestellar/kubestellar/blob/v0.26.0-alpha.1/docs/content/direct/get-started.md?plain=1#L5 ("two ways to create one particular simple configuration) and https://github.com/kubestellar/kubestellar/blob/v0.26.0-alpha.1/docs/content/direct/get-started.md?plain=1#L35 ("the script does the same things as described in the [Step by Step Setup]").

MikeSpreitzer: I am not necessarily opposed to having just one script that does a bit more, but:

(A) it needs to be documented, and
(B) it needs to stay limited to producing only a demo configuration, or else this becomes a bigger project (indeed one that we. should do, but not as part of this PR).

scripts/create-kubestellar-demo-env.sh
```diff
@@ -13,10 +13,39 @@
...
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --platform) k8s_platform="$2"; shift ;;
+        -X) set -x ;;
+        -h|--help)
+            echo "Usage: $0 [--platform <kind|k3d>] [-X] [-h|--help]" >&2
+            exit 0
+            ;;
+        *)
+            echo "Unknown parameter passed: $1"
```
MikeSpreitzer, line 33: FYI, this seems like an error message to me, something that I would write to stderr.

scripts/create-kubestellar-demo-env.sh
clubanderson: this will be the demo platform from which demos can be constructed upon. It need not go any further than how it is structured now. The steps after this script are to be handled but the specific integrations in places like galaxy. This is intended to be the baseline. And, yes, it can be documented once it is approved

docs/content/direct/get-started.md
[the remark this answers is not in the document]
MikeSpreitzer: Now it looks like we are back to the original problem. I see `{{ config.ks_latest_release }}` in `get-started.md`, which means that readers of `main` will get failures until the next release. Is this what you intend?

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: We have made exceptions before and I am willing to do so again, but I think that our normal procedure should be to update the documentation in the same PR as the code. It helps reviewers, and it avoids confusing readers of `main`.

MikeSpreitzer (commented):
This is good enough to merge, excepting our normal procedure of including doc updates.
As noted in the relevant discussion, the included doc update inserts an instruction that can not be used directly until the next release. I have adapted it in my tested, as shown in the attached log. With that adaptation, I was able to test successfully, including running example scenarios 1 and 4.

[k3d-trial8.log](https://github.com/user-attachments/files/18044446/k3d-trial8.log)

/lgtm

docs/content/direct/get-started.md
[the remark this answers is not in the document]
clubanderson: I updated the getting started for the interim step outlined here.

MikeSpreitzer (approved):
/lgtm
/approve

PR #2624, opened 2024-11-21
✨ Enable prefetch in create kind cluster script

francostellari:
## Summary

Introduce a `--prefetch` argument to prefetach nginx ingress images to the kind cluster.

NOTE:  the sha256 portion of the image tags needs to be removed from the nginx YAML before applying it because the images loaded into the kind cluster will have a different sha256. The YAML would keep pulling new images even though we loaded them.

## Related issue(s)

Fixes #

### DIFF
--- a/scripts/create-kind-cluster-with-SSL-passthrough.sh
+++ b/scripts/create-kind-cluster-with-SSL-passthrough.sh
@@ -81,7 +81,7 @@ if [[ "$prefetch" == "true" ]] ; then
   IFS=' ' read -r -a prefetch_images <<< "$(echo "$nginx_ingress" | grep 'image: ' | awk '{print $2}' | uniq | tr '\n' ' ')"
   for image in "${prefetch_images[@]}" ; do
     echo -n "Prefetching image: \"${image}\"... "
-    docker pull "${image}" > /dev/null &
+    docker pull "${image}" -q &
     prefetch_pids+=("$!")
     echo "pid=${prefetch_pids[-1]}"
   done
--- a/scripts/create-kind-cluster-with-SSL-passthrough.sh
+++ b/scripts/create-kind-cluster-with-SSL-passthrough.sh
@@ -17,18 +17,22 @@
 
 set -o errexit
 
+NGINX_INGRESS_URL="https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml"
+NGINX_INGRESS_FILE="/tmp/nginx-ingress.yaml"
 name=kubestellar
 port=9443
 wait=true
+prefetch=false
 setcontext=true
 
 display_help() {
   cat << EOF
-Usage: $0 [--name name] [--port port] [--nowait] [--nosetcontext] [-X]
+Usage: $0 [options]
 
 --name name     set a specific name of the kind cluster (default: kubestellar)
 --port port     map the specified host port to the kind cluster port 443 (default: 9443)
 --nowait        when set to true, the script proceeds without waiting for the nginx ingress patching to complete
+--prefetch      prefetch the nginx ingress images for the kind cluster while the kind cluster is being created
 --nosetcontext  when set to true, the script does not change the current kubectl context to the newly created cluster
 -X              enable verbose execution of the script for debugging
 EOF
@@ -48,6 +52,8 @@ while (( $# > 0 )); do
     fi;;
   (--nowait)
     wait=false;;
+  (--prefetch)
+    prefetch=true;;
   (--nosetcontext)
     setcontext=false;;
   (-X)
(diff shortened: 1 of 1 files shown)


scripts/create-kind-cluster-with-SSL-passthrough.sh
```diff
@@ -65,6 +71,27 @@ while (( $# > 0 )); do
...
 done
 
+
+###############################################################################
+# Prefetch the images required by nginx ingress
+###############################################################################
+prefetch_images=()
+prefetch_pids=()
+if [[ "$prefetch" == "true" ]] ; then
+  curl -s "$NGINX_INGRESS_URL" | sed 's/@sha256.*$//' > "$NGINX_INGRESS_FILE"
```
MikeSpreitzer, line 81: Why is sequentially fetching the URL now better than later?

MikeSpreitzer, line 81: Here is a way to suppress the progress meter without suppressing error messages:

```console
bash-3.2$ curl --no-progress-meter "$NGINX_INGRESS_URL" | sed 's/@sha256.*$//' > "$NGINX_INGRESS_FILE"
bash-3.2$ 
```

scripts/create-kind-cluster-with-SSL-passthrough.sh
```diff
@@ -65,6 +71,27 @@ while (( $# > 0 )); do
...
+  IFS=' ' read -r -a prefetch_images <<< "$(cat $NGINX_INGRESS_FILE | grep 'image: ' | awk '{print $2}' | uniq | tr '\n' ' ')"
```
MikeSpreitzer, line 82: FYI, you can do this with less and lower tech, because newlines are also separators in `for ...; do ...; done`. Following is a demonstration (also making the grep and uniq less fragile).

```console
bash-3.2$ prefetch_images=$(cat $NGINX_INGRESS_FILE | grep -w 'image:' | awk '{print $2}' | sort | uniq)
bash-3.2$ 
bash-3.2$ echo $prefetch_images
registry.k8s.io/ingress-nginx/controller:v1.12.0-beta.0 registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.4.4
bash-3.2$ echo "$prefetch_images"
registry.k8s.io/ingress-nginx/controller:v1.12.0-beta.0
registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.4.4
bash-3.2$ 
bash-3.2$ for img in $prefetch_images; do echo "<$img>"; done
<registry.k8s.io/ingress-nginx/controller:v1.12.0-beta.0>
<registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.4.4>
```

scripts/create-kind-cluster-with-SSL-passthrough.sh
```diff
@@ -65,6 +71,27 @@ while (( $# > 0 )); do
...
+  for image in "${prefetch_images[@]}" ; do
+    echo -n "Prefetching image: \"${image}\"... "
+    docker pull "${image}" > /dev/null &
```
MikeSpreitzer, line 85: Is this really an effective prefetch? The image gets used inside the kubeflex hosting cluster, not on the host machine.

MikeSpreitzer (commented):
This is good enough to merge. Suggestions for improvements above.

/lgtm
Thanks!

francostellari:
@MikeSpreitzer I can work on the suggested changes if you can give enough time. Up to you. Otherwise I can make another PR later.

MikeSpreitzer:
I am not in a hurry.

francostellari:
@MikeSpreitzer I pushed a new commit removing the use of the tmp file.

scripts/create-kind-cluster-with-SSL-passthrough.sh
```diff
@@ -65,8 +70,29 @@ while (( $# > 0 )); do
...
+prefetch_images=()
+prefetch_pids=()
+if [[ "$prefetch" == "true" ]] ; then
+  nginx_ingress="$(curl --no-progress-meter "$NGINX_INGRESS_URL" | sed 's/@sha256.*$//')"
+  IFS=' ' read -r -a prefetch_images <<< "$(echo "$nginx_ingress" | grep 'image: ' | awk '{print $2}' | uniq | tr '\n' ' ')"
+  for image in "${prefetch_images[@]}" ; do
+    echo -n "Prefetching image: \"${image}\"... "
+    docker pull "${image}" -q &
+    prefetch_pids+=("$!")
+    echo "pid=${prefetch_pids[-1]}"
```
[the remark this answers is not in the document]
francostellari: @MikeSpreitzer to clarify, is it the `+=(...)` syntax?
It's not totally clear to me from the log.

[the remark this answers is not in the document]
MikeSpreitzer: The error message is pointing to line 86, which is the following.

```shell
    echo "pid=${prefetch_pids[-1]}"
```

I suspect the problem is subscripting with `-1`. I think you could easily avoid that by first storing `$!` in a shell variable and then both appending it to `prefetch_pids` and echoing it.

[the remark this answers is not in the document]
MikeSpreitzer:
```bash
bash-3.2$ foo=(abc def)
bash-3.2$ echo ${foo[1]}
def
bash-3.2$ echo ${foo[-1]}
bash: foo: bad array subscript
```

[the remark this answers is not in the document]
francostellari: Fixed in the last commit

scripts/create-kind-cluster-with-SSL-passthrough.sh
```diff
@@ -65,8 +70,29 @@ while (( $# > 0 )); do
...
+prefetch_images=()
+prefetch_pids=()
+if [[ "$prefetch" == "true" ]] ; then
+  nginx_ingress="$(curl --no-progress-meter "$NGINX_INGRESS_URL" | sed 's/@sha256.*$//')"
+  IFS=' ' read -r -a prefetch_images <<< "$(echo "$nginx_ingress" | grep 'image: ' | awk '{print $2}' | uniq | tr '\n' ' ')"
+  for image in "${prefetch_images[@]}" ; do
+    echo -n "Prefetching image: \"${image}\"... "
+    docker pull "${image}" -q &
+    prefetch_pids+=("$!")
+    echo "pid=${prefetch_pids[${#prefetch_pids[@]}-1]}"
```
MikeSpreitzer, line 86: This should work.

```shell
bash-3.2$ foo=(abc def)
bash-3.2$ echo ${foo[1]}
def
bash-3.2$ echo ${foo[${#foo[@]}-1]}
def
```

MikeSpreitzer (approved):
Successfully tested, both `--prefetch` and not, on my Mac.

/lgtm
/approve
Thanks!

PR #2625, opened 2024-11-21
✨ Improve scripts for building some of the service images

francostellari:
## Summary

Improve scripts for building some of the service images
- remove `jq` and `yq` from the `kubectl` container image
- fix dir change in `build-clusteradm-image.sh`

## Related issue(s)

Fixes #

### DIFF
--- a/hack/build-clusteradm-image.sh
+++ b/hack/build-clusteradm-image.sh
@@ -22,7 +22,7 @@
 set -e
 
 clusteradm_version="" # ==> latest
-clusteradm_folder=clusteradm
+clusteradm_folder="/tmp/clusteradm"
 registry=quay.io/kubestellar
 platform=linux/amd64,linux/arm64,linux/ppc64le
 
@@ -73,11 +73,12 @@ echo "Using clusteradm v${clusteradm_version}."
 
 git clone -b "v$clusteradm_version" --depth 1 https://github.com/open-cluster-management-io/clusteradm.git "$clusteradm_folder"
 
-cd "$clusteradm_folder"
+pushd "$clusteradm_folder"
 
 export KO_DOCKER_REPO=$registry
 
 ko build -B ./cmd/clusteradm -t $clusteradm_version --sbom=none --platform $platform
 
-cd ~
+popd
+
 rm -rf $clusteradm_folder
--- a/hack/build-clusteradm-image.sh
+++ b/hack/build-clusteradm-image.sh
@@ -72,8 +72,8 @@ clusteradm_version=${clusteradm_version#v}
 
 echo "Using clusteradm v${clusteradm_version}."
 
-git clone -b "v$clusteradm_version" --depth 1 https://github.com/open-cluster-management-io/clusteradm.git "$clusteradm_folder"
 cd "$clusteradm_folder"
+git clone -b "v$clusteradm_version" --depth 1 https://github.com/open-cluster-management-io/clusteradm.git .
 
 export KO_DOCKER_REPO=$registry
 ko build -B ./cmd/clusteradm -t $clusteradm_version --sbom=none --platform $platform
--- a/hack/build-clusteradm-image.sh
+++ b/hack/build-clusteradm-image.sh
@@ -21,7 +21,7 @@
 
 set -e
 
-clusteradm_folder="$(mktemp -d -p . "clusteradm-XXXX")"
+clusteradm_folder="$(mktemp -d -p $PWD "clusteradm-XXXX")"
 trap 'rm -rf "$clusteradm_folder"' EXIT # delete the temporary folder on exit
 clusteradm_version="" # ==> latest
 registry=quay.io/kubestellar
@@ -73,9 +73,7 @@ clusteradm_version=${clusteradm_version#v}
 echo "Using clusteradm v${clusteradm_version}."
 
 git clone -b "v$clusteradm_version" --depth 1 https://github.com/open-cluster-management-io/clusteradm.git "$clusteradm_folder"
+cd "$clusteradm_folder"
 
-( # Ensure that the temprary folder is deketed even on error
-    cd "$clusteradm_folder"
-    export KO_DOCKER_REPO=$registry
-    ko build -B ./cmd/clusteradm -t $clusteradm_version --sbom=none --platform $platform
-)
+export KO_DOCKER_REPO=$registry
+ko build -B ./cmd/clusteradm -t $clusteradm_version --sbom=none --platform $platform
--- a/hack/build-clusteradm-image.sh
+++ b/hack/build-clusteradm-image.sh
@@ -21,8 +21,9 @@
 
 set -e
 
+clusteradm_folder="$(mktemp -d -p . "clusteradm-XXXX")"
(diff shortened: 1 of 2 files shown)


hack/build-clusteradm-image.sh
```diff
@@ -22,7 +22,7 @@
 set -e
 
 clusteradm_version="" # ==> latest
-clusteradm_folder=clusteradm
+clusteradm_folder="/tmp/clusteradm"
```
MikeSpreitzer, line 25: What makes `/tmp/clusteradm` better than `./clusteradm`? I can see one downside to using `/tmp`: it is less protected by the filesystem.

MikeSpreitzer: Are you saying that the binary should go to `/tmp` rather than `.` because it is annoying to have `git` complain about the binary not being staged for commit? That can easily be avoided by adding `clusteradm` to `.gitignore`.

francostellari: No, I'm concerned that somebody already has a local `./clusteradm` folder with stuff in it that he/she does not want to be overwritten and deleted at the end. I will change to use tempdir when I have a chance.

MikeSpreitzer: The scripts in `hack/` are intended for use by contributors. These are people who have checked out the kubestellar repository. The normal expectation is that the current working directory is `kubestellar` (or possibly a subdirectory), unless the contributor is explicitly told otherwise. This script is used in the process of building and publishing KubeStellar, so I do not think that we need to worry about it being used with an arbitrary current working directory.

francostellari:
@MikeSpreitzer I changed the script to use `mktemp` with a `trap` to delete the temporary folder on exist.
Note that the block at the end in `( ... )` is needed to allow the script to delete the temporary folder if `ko` fails, which would be prevented by the combination of using `set -e` and the script being in the `clusteradm_folder` folder.

hack/build-clusteradm-image.sh
```diff
@@ -73,11 +74,8 @@ echo "Using clusteradm v${clusteradm_version}."
...
 
-cd "$clusteradm_folder"
-
-export KO_DOCKER_REPO=$registry
-
-ko build -B ./cmd/clusteradm -t $clusteradm_version --sbom=none --platform $platform
-
-cd ~
-rm -rf $clusteradm_folder
+( # Ensure that the temprary folder is deketed even on error
```
MikeSpreitzer, line 77: (A) typo in "deketed".

(B) the `trap 'rm ...' EXIT` ensures that the `rm` command is executed no matter what.

(C) The parentheses here add nothing. Further, they are harmful because they cause a subshell to be used, so all the local shell variables are lost. It would be better to use curly braces rather than parentheses, if anything.

(D) I am not a fan of deleting the evidence when something goes wrong. In my opinion, it would be better to `{ ... commands to build ...} && rm -rf "$clusteradm_folder"` so that the evidence goes away only on success.

MikeSpreitzer:
On further thought: the script here works ONLY if the current working directory is `kubestellar`. So: (a) that should be documented and (b) there is no concern about a conflicting use of a subdirectory named "clusteradm".

hack/build-clusteradm-image.sh
francostellari: @MikeSpreitzer 

(B) if the script fails on the `ko` command, because if the `set -e` it will terminate and exit immediately, the `trap` will indeed trigger but since the script has `cd` to the temp folder that would like to remove, it will NOT remove it. I did observe this behavior and took actions

(C) indeed the parenthesis add the extra shell execution so that if it fails, the "main" script will NOT be holding the temp folder, it will still exit immediately due to `set -e` but the `trap` will work in this case... also tested it

(D) the temp folder does not contain anything useful as it is an exact copy of somebody else repo. 99% of ko fail are related to the user having forgotten to login into quay.io before running the script. I'm not a fan to keep garbage in `.`. I would be ok to let it be in `/tmp` as it would be eventually deleted by the system. A second successful run of the script after a previous failure will leave the previous temp folder in place

francostellari: @MikeSpreitzer I just poshed a new version with a different solution to the problem that does nor require the temporary shell

francostellari:
> On further thought: the script here works ONLY if the current working directory is `kubestellar`. So: (a) that should be documented and (b) there is no concern about a conflicting use of a subdirectory named "clusteradm".

The script works from any folder as long as you provide full path to the script.
It does not depend on anything in kubestellar repo
It can be run as curt|bash

hack/build-clusteradm-image.sh
```diff
@@ -72,12 +73,7 @@ clusteradm_version=${clusteradm_version#v}
 echo "Using clusteradm v${clusteradm_version}."
 
 git clone -b "v$clusteradm_version" --depth 1 https://github.com/open-cluster-management-io/clusteradm.git "$clusteradm_folder"
```
MikeSpreitzer, line 75: This command would get a bit simpler if it came after the `cd "$clusteradm_folder"`.

hack/build-clusteradm-image.sh
```diff
@@ -72,12 +73,7 @@ clusteradm_version=${clusteradm_version#v}
...
-
 cd "$clusteradm_folder"
 
 export KO_DOCKER_REPO=$registry
-
-ko build -B ./cmd/clusteradm -t $clusteradm_version --sbom=none --platform $platform
-
-cd ~
-rm -rf $clusteradm_folder
+ko build -B ./cmd/clusteradm -t $clusteradm_version --sbom=none --platform $platform
```
MikeSpreitzer, line 79: Reviewers would see less red if this file ended with a linebreak.

MikeSpreitzer:
> The script works from any folder as long as you provide full path to the script.
> It does not depend on anything in kubestellar repo

Oh, right.

hack/build-clusteradm-image.sh
MikeSpreitzer: Oh, right, the parentheses _were_ doing something.

Since there is so little distinct information in the build directory, I am OK with it being deleted even on failure.

The build directory does contain one bit of information not in Git: _which_ version was attempted to be built. Maybe if the script echoed that, there would be no doubt.

hack/build-clusteradm-image.sh
```diff
@@ -71,13 +72,8 @@ clusteradm_version=${clusteradm_version#v}
 
 echo "Using clusteradm v${clusteradm_version}."
 
-git clone -b "v$clusteradm_version" --depth 1 https://github.com/open-cluster-management-io/clusteradm.git "$clusteradm_folder"
-
 cd "$clusteradm_folder"
+git clone -b "v$clusteradm_version" --depth 1 https://github.com/open-cluster-management-io/clusteradm.git .
```
MikeSpreitzer, line 76: Ugh, sorry I suggested this. It was more informative the other way. Now the user of the script is not told the name of the build directory, instead is told "Cloning into '.'..." and that's confusing. Can be fixed now or in follow-on.

MikeSpreitzer (approved):
Successfully tested with `--registry $a_test`.

This is good enough to merge.

/lgtm
/approve

PR #2629, opened 2024-11-25
🌱  Always save spellcheck output to workflow artifact

### LINKED ISSUE #2628: Spellcheck workflow never saves its output
It always fails the checking step, so the following step --- the one that saves the output --- does not run. See https://stackoverflow.com/questions/58858429/how-to-run-a-github-actions-step-even-if-the-previous-step-fails-while-still-f for ways to make it run.

MikeSpreitzer:
## Summary
This PR changes the condition under which an attempt is made to save the spell checking output to a workflow artifact. This PR removes the restriction to the "kubestellar" organization, so that contributors can also get these artifacts for their own branches. This PR allows the saving to be attempted even if the spell checking finds errors.

## Related issue(s)

Fixes #2628

### DIFF
--- a/.github/workflows/spellcheck_action.yml
+++ b/.github/workflows/spellcheck_action.yml
@@ -40,7 +40,7 @@ jobs:
 #         source_files: "docs/content/*"
         
     - uses: actions/upload-artifact@v4
-      if: github.repository_owner == 'kubestellar'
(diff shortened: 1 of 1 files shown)


clubanderson (commented):
/lgtm
/approve

PR #2631, opened 2024-11-25
✨  Bump KubeFlex dependency to 0.7.2

MikeSpreitzer:
## Summary
This PR increases the imported and required KubeFlex release to 0.7.2. This is part of the upgrade campaign.

Note that the currently documented release procedure (in docs/content/direct/release.md) does not contemplate a required KF version that is different from the imported version. That is, we currently have no wiggle room. I think this is too strict in general. But it is not a battle that we need to fight right now, because the upgrade campaign is for (among other things) security improvements that should not be optional.

## Related issue(s)

Fixes #


go.mod
```diff
@@ -1,19 +1,21 @@
 module github.com/kubestellar/kubestellar
 
-go 1.22
+go 1.22.7
+
+toolchain go1.23.2
```
waltforme, line 5: Just want to understand this. Why do we need a toolchain version higher than the one in `go 1.22.7`?

pdettori: I have the same question

MikeSpreitzer: Oops, I thought that I had removed that. Fixed now.

Issue #2613, opened 2024-11-20
feature: SBOM needed to clomonitor tests


clubanderson:
### Feature Description

clomonitor requires a valid SBOM is present in a repo to improve the repo score

**SOFTWARE BILL OF MATERIALS (SBOM)**
ID: sbom

List of components in a piece of software, including licenses, versions, etc.

This check passes if:

The latest release on Github includes an asset which name contains sbom. Regexps used:
"(?i)sbom"
The repository’s README file contains a SBOM section that explains where they are published to, format used, etc. Regexps used to locate the title header:
"(?im)^#+.*sbom.*$"
"(?im)^#+.*software bill of materials.*$"
"(?im)^sbom$"
"(?im)^software bill of materials$"

### Proposed Solution

produce SBOM with our CICD

### Want to contribute?

- [ ] I would like to work on this issue.

jaydee029:
Hey @clubanderson, I'd like to work on this,  have done something similar for wasmedge before.

clubanderson:
@jaydee029 yes. Please take the issue. Thank you for you contribution in advance.

MikeSpreitzer:
If we remove the binary archives from the releases (see #2653), would that make this moot?

jaydee029:
> If we remove the binary archives from the releases (see #2653), would that make this moot?

So I'm not sure, but I think SBOM here is concerned with everything you used to build this particular software, If I go into the release section on github I find this

![image](https://github.com/user-attachments/assets/0a20a67e-1a18-47e5-ac7a-f78df20c0177)

Since we do have packages/binaries of some sort or the other in our release, SBOM would be required.
I might be wrong.

MikeSpreitzer:
If the binary archives go, would that make clomonitor stop caring about an SBOM, or would it still be looking for an SBOM covering the source archives?

jaydee029:
Please check the discussion [here](https://github.com/kubestellar/kubestellar/issues/2653#issuecomment-2576959955)

MikeSpreitzer:
I just created a test release, 0.26.0-alpha.5, to test the intended fix for this Issue, PR #2718 . The test release did _not_ get an SBOM in its assets.

MikeSpreitzer:
@jaydee029

jaydee029:
Sure ,I'll look into this soon, I've been busy for a last few days, sorry for the delay

jaydee029:
[...]

PR #2635, opened 2024-12-01
✨ CI: Added SBOM

jaydee029:
## Summary
Added SBOM action in the CI workflows.
Looking at the other workflows, I added it in a different workflow, which is triggered only when the test-release workflow is completed and successful.
please recommend changes.

## Related issue(s)

Tries fixing #2613


.github/workflows/releaseutils.yml
```diff
@@ -0,0 +1,25 @@
...
+              uses: actions/checkout@v4
+
+            - uses: advanced-security/sbom-generator-action@v0.0.1
+              id: sbom
+              env: 
+                    GITHUB_TOKEN: ${{ github.token }}
+            - uses: actions/upload-artifact@v3.1.0
+              with: 
+                path: ${{steps.sbom.outputs.fileName }}
+                name: "SBOM"
```
MikeSpreitzer, line 25: We will see less red if there is a terminal linebreak here.

.github/workflows/releaseutils.yml
```diff
@@ -0,0 +1,25 @@
...
+        runs-on: ubuntu-latest
+        if: ${{github.event.workflow_run.conclusion=='success'}}
+        steps:
+            - name: checkout code
+              uses: actions/checkout@v4
+
+            - uses: advanced-security/sbom-generator-action@v0.0.1
+              id: sbom
+              env: 
+                    GITHUB_TOKEN: ${{ github.token }}
```
MikeSpreitzer, line 21: Is there a way to give this Action only read access to the repo? That would be better.

jaydee029: done

.github/workflows/releaseutils.yml
jaydee029: sure, missed that

.github/workflows/releaseutils.yml
jaydee029: currently gave read all permissions, but I think more granular access can be provided in the future

MikeSpreitzer: I think read all is OK, since the repo already gives read all permission to everybody.

.github/workflows/releaseutils.yml
MikeSpreitzer: I still see the red "no eol here" icon.
image

jaydee029: I think i didn't leave a line at the end of the file , is that it?

jaydee029: I made another commit does it look good now?

MikeSpreitzer: yes

.github/workflows/releaseutils.yml
```diff
@@ -0,0 +1,30 @@
+name: Add sbom
+
+on:
+  workflow_dispatch:
+  workflow_run:
+        workflows: [test-latest-release]
+        types: [completed]
+
+permissions: read-all
```
MikeSpreitzer, line 9: Does this thwart the upload step?

.github/workflows/releaseutils.yml
```diff
@@ -0,0 +1,30 @@
...
+              uses: actions/checkout@v4
+
+            - name: Create SBOM
+              uses: advanced-security/sbom-generator-action@v0.0.1
+              id: sbom
+              env: 
+                  GITHUB_TOKEN: ${{ github.token }}
+              
+            - name: Upload SBOM artefact
+              uses: actions/upload-artifact@v3.1.0
```
MikeSpreitzer, line 27: Do you want an "artifact" or an "asset"? AFAIK, an "artifact" is associated with a workflow run, and "asset" is associated with a release.

.github/workflows/releaseutils.yml
MikeSpreitzer: BTW, if you want to grant only the needed write privileges, I think https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents identifies what is needed for writing "assets".

MikeSpreitzer: BTW, if the SBOM is less than 1MB then https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/passing-information-between-jobs shows how you can link-up differently-privileged jobs in a workflow.

.github/workflows/releaseutils.yml
jaydee029: right, that's why i think the read permission thing might work, for an artefact since its stored with the workflow, it might not required the write permissions, but an asset/package surely does

jaydee029: if we need as a release asset, i'll make the required changes

.github/workflows/releaseutils.yml
jaydee029: answered in the other review, and yes I'll use granular write for the specific job if required.

.github/workflows/releaseutils.yml
```diff
@@ -0,0 +1,30 @@
...
+        if: ${{github.event.workflow_run.conclusion=='success'}}
+        steps:
+            - name: checkout code
+              uses: actions/checkout@v4
+
+            - name: Create SBOM
+              uses: advanced-security/sbom-generator-action@v0.0.1
+              id: sbom
+              env: 
+                  GITHUB_TOKEN: ${{ github.token }}
```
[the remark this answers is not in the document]
jaydee029: i think with read-all permissions for the token are said to 'read', since we only need to read the token while doing this it won't thwart that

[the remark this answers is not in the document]
jaydee029: As for the github token, I provided since its mention that way in its documentation, I can look for a different github action for the same , if you want me to.

.github/workflows/releaseutils.yml
jaydee029: I looked at the release assets for Kubestellar, they are in the tar.gz format, what we can do is convert it into the tar.gz format and upload it along with the release as an asset. This would also be helpful later when we decide to sign the artefacts in as mentioned in one of the issues.

MikeSpreitzer: #2613 seems to be clear that what is needed is an "asset" getting attached to the release.

jaydee029: sure

.github/workflows/releaseutils.yml
MikeSpreitzer: You are replying to a confused comment of mine that I have since deleted after I found in https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#about-the-github_token-secret the statement that "The token is also available in the github.token context".

jaydee029: fine ,probably a delay. I'll redo this workflow and ping you once its done.

MikeSpreitzer:
Since our releases are built using `goreleaser`, and that can create SBOMs, is that the way to go? https://goreleaser.com/customization/sbom/

MikeSpreitzer:
Also: do we even need this? https://github.com/kubestellar/kubestellar/issues/2613#issuecomment-2523710038

jaydee029:
> Since our releases are built using `goreleaser`, and that can create SBOMs, is that the way to go? https://goreleaser.com/customization/sbom/

I looked at this, sure its possible, but at the end of the page in the limitations it says that goreleaser wont take in account the container images that are created, I think it kind of defeats the purpose.

image

MikeSpreitzer:
The complaint in the opening comment of #2613 seems to be about the assets of a release, not about container images.

For the container images, since they are built with `ko`, that would be what I expect to create the SBOM. However, I am not sure whether the sort of SBOM that `ko` builds is acceptable to clomonitor.

PR #2639, opened 2024-12-03
🐛  [release-0.25.1] Stop MW thrashing when max_num_wrapped is 1

MikeSpreitzer:
## Summary
This PR attempts to fix Issue #2634 in the release-0.25.1 branch for the case where max_num_wrapped==1, by making the wrapper's name a function of its contents (as well as source).

This PR includes the test expansion from #2640 , and the fix here makes the expanded test succeed.

This PR is built on #2638 .

DO NOT MERGE THIS PR. This PR is just for CI testing. We do not have a documented procedure for making new patch releases after `main` has diverged, so I am making this up as I go along. I think that what we should do is: after #2638 merges into the `release-0.25.1` branch, we should clone that branch into a new one named `release-0.25.2`, close this PR, and create a clone of this PR that updates the `release-0.25.2` branch. That branch should also get corresponding doc updates. From that branch create the 0.25.2 release.

## Related issue(s)

Fixes #

PR #2638, opened 2024-12-03
📖  [release-0.25.1] Document workload object thrashing

MikeSpreitzer:
## Summary
This PR updates the documentation for release 0.25.1 to document the pathology that I realized while thinking about Issue #2634 .

This PR also adds a note to the "known issues" index page to explicitly state its relation with bugs.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-mw-thrash-0251/direct/release-notes/

## Related issue(s)

Fixes #

Issue #2634, opened 2024-11-27
bug: Pods for a completed Job start again


cwiklik:
### Describe the bug

My environment:
k3d version v5.7.4
k3s version v1.30.4-k3s1 (default)
KubeStellar: 0.25.1 

I create a job in the mgmt cluster, it gets pulled to a WEC  and runs to completion. When the job completes I start another ( same job spec). I create multiple jobs this way. What I notice eventually on the WEC is that pods for a job, which previously completed, start again. This happens with KS v.0.25.0 and 0.25.1. In testing I've removed Kueue and my Kueue-KS integration controllers so its just KS in the picture. Below are two images, one showing all pods in completed state; and another showing the previously completed pods as running again.

![image](https://github.com/user-attachments/assets/3b476a23-b7e1-43aa-bd54-abb23f69c218)

I start a new job: sample-job-8qksl. Note that previously completed pods for a job sample-job-fmfzw are in Running state.

![image](https://github.com/user-attachments/assets/8a9858cf-3041-404b-8278-2cfe5e394328)

Below are manifestworks, bp, binding, and my job spec:
[mw1.txt](https://github.com/user-attachments/files/17937453/mw1.txt)
[mw3.txt](https://github.com/user-attachments/files/17937454/mw3.txt)
[mw4.txt](https://github.com/user-attachments/files/17937455/mw4.txt)
[mw5.txt](https://github.com/user-attachments/files/17937456/mw5.txt)
[wec-klusterlet.log](https://github.com/user-attachments/files/17937457/wec-klusterlet.log)

Job:
```
apiVersion: batch/v1
kind: Job
metadata:
  generateName: sample-job-
  namespace: default
  labels:
    kueue.x-k8s.io/queue-name: user-queue-ks
    kubestellar.io/cluster: cluster2
spec:
  parallelism: 3
  completions: 3
  suspend: true
  template:
    spec:
      containers:
      - name: dummy-job
        image: gcr.io/k8s-staging-perf-tests/sleep:v0.1.0
        args: ["50s"]
        resources:
          requests:
            cpu: 1
            memory: "200Mi"
          limits:
            memory: "200Mi"
            cpu: 1
      restartPolicy: Never
```
BP:
```
apiVersion: control.kubestellar.io/v1alpha1
kind: BindingPolicy
metadata:
  creationTimestamp: "2024-11-27T16:01:04Z"
  finalizers:
  - bindingpolicy.kubestellar.io/kscontroller
  generation: 1
[...]

PR #2640, opened 2024-12-03
✨  [release 0.25.1] Test for workload object thrashing

MikeSpreitzer:
## Summary
This PR extends the ginkgo-based end-to-end test to cover thrashing of workload objects that have create-only set. This PR fails CI because this test fails as it should.

This PR is built on #2638 

DO NOT MERGE THIS PR, because it adds testing that exposes a bug. Instead, look at #2639, which also includes the fix for the bug.

## Related issue(s)

This tests for Issue #2634 in release 0.25.1

### DIFF
--- a/.github/workflows/pr-test-e2e.yml
+++ b/.github/workflows/pr-test-e2e.yml
@@ -16,12 +16,14 @@ on:
   pull_request:
     branches:
       - main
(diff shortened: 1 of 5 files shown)


MikeSpreitzer:
This PR has served its purpose

PR #2643, opened 2024-12-03
✨ Report missing StatusCollector objects

### LINKED ISSUE #2574: feature: State based error reporting of missing StatusCollector objects
### Feature Description

Currently, if a DownsyncPolicyClause references a StatusCollector that does not exist, then there is no error message about this in any API object.

### Proposed Solution

Put in the BindingPolicy, Binding, and/or CombinedStatus object(s) status section an error message about this condition.

### Want to contribute?

### Additional Context

_No response_

### LINKED ISSUE #2722: bug: binding controller's binding informer not setup properly
### Describe the bug

`ResourceEventHandlerFuncs` of the informer are adding Binding**Policy**'s reference to the workqueue, but it is Binding's reference that should be added here.

I'm fixing this during my ongoing work on #2643.

https://github.com/kubestellar/kubestellar/blob/d7ecbc939fc26123345f17f6261a193d8ff312f2/pkg/binding/binding-controller.go#L572-L602

_No response_

### Steps To Reproduce

Review lines 578, 585, and 594 in the code snippet above.

### Expected Behavior

Binding's reference should be added instead.

### Want to contribute?

### Additional Context

_No response_

waltforme:
## Summary
This PR makes the controller manager to report missing StatusCollector objects (that are required by bindingpolicies) in two ways (instead of failing silently).

First, the controller manager logs the name of the missing StatusCollector. Second, a Condition is introduced to indicate the availability of the required StatusCollector objects.

## Related issue(s)

Fixes #2574
Fixes #2722


api/control/v1alpha1/conditions.go
```diff
@@ -76,6 +78,18 @@ func SetCondition(conditions []BindingPolicyCondition, newCondition BindingPolic
 	return conditions
 }
 
+// UnsetCondition unsets the BindingPolicyCondition of the supplied type in
+// the given slice, if the BindingPolicyCondition of that type exists.
+func UnsetCondition(conditions []BindingPolicyCondition, conditionType ConditionType) []BindingPolicyCondition {
```
MikeSpreitzer, line 83: I do not see a need for this func. Whenever the controller has an opinion about a given type of condition for a given object, it should ensure the presence of a Condition with that type and the proper values.

pkg/binding/binding.go
```diff
@@ -109,6 +109,7 @@ func (c *Controller) syncBinding(ctx context.Context, bindingName string) error
 	policyWithStatus := policy.DeepCopy()
 	policyWithStatus.Status = v1alpha1.BindingPolicyStatus{
 		ObservedGeneration: policy.Generation,
+		Conditions:         policy.Status.Conditions,
```
MikeSpreitzer, line 112: This is going backwards. Since Binding is derived from BindingPolicy, the flow should be: BindingPolicy desired state -> Binding desired state -> stuff happens -> Binding reported state -> BindingPolicy reported state.

pkg/binding/bindingpolicy.go
```diff
@@ -88,6 +88,47 @@ func (c *Controller) syncBindingPolicy(ctx context.Context, bindingPolicyName st
 		logger.V(5).Info("Enqueued Binding for syncing, while handling BindingPolicy", "name", bindingPolicy.Name)
 		c.enqueueBinding(bindingPolicy.GetName())
 
+		statusCollectorName, err := c.identifyMissingStatusCollector(bindingPolicy)
```
MikeSpreitzer, line 91: Why not identify all the missing StatusCollector objects? It would be more helpful to a user to reveal all the problems.

pkg/binding/bindingpolicy.go
```diff
@@ -88,6 +88,47 @@ func (c *Controller) syncBindingPolicy(ctx context.Context, bindingPolicyName st
...
+		if err != nil {
+			return fmt.Errorf("failed to list StatusCollector %s for BindingPolicy %s: %w", statusCollectorName, bindingPolicyName, err)
```
MikeSpreitzer, line 93: I see no list operation here. If you insist on wrapping the error, the wrap should talk about this call site is doing.

pkg/binding/bindingpolicy.go
```diff
@@ -88,6 +88,47 @@ func (c *Controller) syncBindingPolicy(ctx context.Context, bindingPolicyName st
...
+		}
+
+		// If at least one StatusCollector object is missing, ensure the Condition and requeue the BindingPolicy,
```
MikeSpreitzer, line 96: A missing StatusCollector should not cause requeuing. A missing StatusCollector should cause the proper error report in status, that is the successful reaction to the current state. Should a relevant StatusCollector object be created later, just as if one is deleted later, this is cause for reconsidering the current state.

pkg/binding/bindingpolicy.go
```diff
@@ -101,6 +142,23 @@ func (c *Controller) syncBindingPolicy(ctx context.Context, bindingPolicyName st
 	return c.deleteResolutionForBindingPolicy(ctx, bindingPolicyName)
 }
 
+func (c *Controller) identifyMissingStatusCollector(bp *v1alpha1.BindingPolicy) (string, error) {
+	for _, clause := range bp.Spec.Downsync {
+		if clause.DownsyncModulation.StatusCollectors == nil {
+			continue
+		}
```
MikeSpreitzer, line 149: https://go.dev/ref/spec#For_range, case 1, explicitly allows `range` over `nil` and it does what you would expect.

MikeSpreitzer (commented):
I have finished my first review. I left individual comments.

pkg/binding/binding.go
MikeSpreitzer: We already have reported state flowing from Binding to BindingPolicy in https://github.com/kubestellar/kubestellar/blob/v0.26.0-alpha.1/pkg/binding/binding.go#L91 through https://github.com/kubestellar/kubestellar/blob/v0.26.0-alpha.1/pkg/binding/binding.go#L112

pkg/binding/bindingpolicy.go
```diff
@@ -88,6 +88,55 @@ func (c *Controller) syncBindingPolicy(ctx context.Context, bindingPolicyName st
...
+			}
+
+			return fmt.Errorf("failed to sync BindingPolicy %s because StatusCollector(s) %s are missing", bindingPolicyName, statusCollectorNames)
+		}
+
+		scCondition := v1alpha1.BindingPolicyCondition{
+			Type:               v1alpha1.TypeStatusCollectorsAvailable,
+			Status:             corev1.ConditionTrue,
+			Reason:             v1alpha1.ReasonReconcileSuccess,
+			LastTransitionTime: metav1.Now(),
```
MikeSpreitzer, line 127: If this is not a change from the previous state of this Condition then LastTransitionTime should not be updated.

pkg/binding/bindingpolicy.go
```diff
@@ -101,6 +150,28 @@ func (c *Controller) syncBindingPolicy(ctx context.Context, bindingPolicyName st
...
+func (c *Controller) identifyMissingStatusCollectors(bp *v1alpha1.BindingPolicy) ([]string, error) {
+	statusCollectorNames := []string{}
+	for _, clause := range bp.Spec.Downsync {
+		if clause.DownsyncModulation.StatusCollectors == nil {
+			continue
+		}
+		for _, statusCollectorName := range clause.DownsyncModulation.StatusCollectors {
+			if _, err := c.statusCollectorLister.Get(statusCollectorName); err != nil {
+				if errors.IsNotFound(err) {
+					statusCollectorNames = append(statusCollectorNames, statusCollectorName)
```
MikeSpreitzer, line 165: If multiple clauses reference the same missing collector, do we really want it listed multiple times? I think that could be annoyingly verbose.

pkg/binding/bindingpolicy.go
MikeSpreitzer: Although GitHub marks this as "Outdated", it is still relevant.

pkg/binding/bindingpolicy.go
```diff
@@ -101,6 +151,25 @@ func (c *Controller) syncBindingPolicy(ctx context.Context, bindingPolicyName st
 	return c.deleteResolutionForBindingPolicy(ctx, bindingPolicyName)
 }
 
+// identifyMissingStatusCollectors finds all missing StatusCollector objects for bp.
+// If an error other than "not found" occurs, identifyMissingStatusCollectors returns a nil slice and the error;
+// otherwise, identifyMissingStatusCollectors returns a slice of StatusCollector object names and a nil error.
+func (c *Controller) identifyMissingStatusCollectors(bp *v1alpha1.BindingPolicy) ([]string, error) {
+	statusCollectorNames := sets.NewString([]string{}...)
```
MikeSpreitzer, line 158: https://github.com/kubernetes/apimachinery/blob/v0.29.12/pkg/util/sets/string.go#L21

It would also be less verbose to simply `sets.New[string]()`.

pkg/binding/bindingpolicy.go
```diff
@@ -101,6 +118,73 @@ func (c *Controller) syncBindingPolicy(ctx context.Context, bindingPolicyName st
...
+	for _, conditionActual := range bp.Status.Conditions {
+		if v1alpha1.AreConditionsEqual(conditionActual, conditionTentative) {
+			hasSame = true
+			break
+		}
+	}
+	// create or update if necessary
+	if hasSame {
+		return nil
+	}
```
MikeSpreitzer, line 150: Wouldn't it be easier to have v1alpha1.SetCondition do the equality test (and return the result) and update the time if it is a change?

pkg/binding/bindingpolicy.go
```diff
@@ -88,6 +90,21 @@ func (c *Controller) syncBindingPolicy(ctx context.Context, bindingPolicyName st
...
+		}
+
+		// If at least one StatusCollector object is missing, use a Condition to indicate the missing object and requeue the BindingPolicy,
+		// o.w. use the Condition to indicate all StatusCollector object(s) are available and proceed
+		if err = c.createOrUpdateStatusCollectorAvailableCondition(ctx, bindingPolicy, statusCollectorNames); err != nil {
+			return fmt.Errorf("failed to update status for BindingPolicy %s: %w", bindingPolicyName, err)
+		}
+		if len(statusCollectorNames) > 0 {
+			logger.V(4).Info("Missing StatusCollector(s)", "statusCollectorNames", statusCollectorNames, "bindingPolicyName", bindingPolicyName)
+			return fmt.Errorf("failed to sync BindingPolicy %s because StatusCollector(s) %s are missing", bindingPolicyName, statusCollectorNames)
```
MikeSpreitzer, line 105: This is still not really a sync failure, I see no need to requeue. Requeuing in this case causes continual controller work as long as any StatusCollector is missing, which is under user control and could be forever.

waltforme: Thanks! Removed the requeueing in latest push.

pkg/binding/bindingpolicy.go
waltforme: Improved.

pkg/binding/bindingpolicy.go
waltforme: Improved.

pkg/binding/bindingpolicy.go
waltforme: Improved.

pkg/binding/bindingpolicy.go
waltforme: Removed the requeuing.

MikeSpreitzer:
BTW, +1 on putting this in a "condition" rather than an "error". Because the usual Kubernetes way is to not require sequencing between independent objects, it is not even an _error_ to reference a StatusCollector before it exists.

pkg/binding/binding.go
MikeSpreitzer: > Are you suggesting that, that logic in SyncBindingPolicy should be moved here into SyncBinding to fit into the flow you described?

No. My main point was that there is already a flow updating the BindingPolicy's status and we should not introduce a second one. Here a a few principles to keep in mind: (a) when updating status, update `.status.reportedGeneration` so that the reader knows what the controller was looking at when doing that update; (b) for a given atom in a given API object, only one controller should author it; and (c) in a given controller, only one goroutine at a time should be writing to a given API object (with "subresources" considered distinct objects for this purpose).

The current code structure is dissatisfying to me. A Binding object is derived from a BindingPolicy object, is essentially the implementation of a BindingPolicy object. Thus, the logical flow is BindingPolicy desired state -> Binding desired state -> stuff happens --> Binding reported state -> BindingPolicy reported state. However, the history of the development of this code leaves us in a different place. Originally BindingPolicy was implemented directly; later Binding objects were introduced as a side-effect of BindingPolicy. I think this is an inferior view, I prefer the one I outlined at the start of this paragraph. But the code structure is not fully there yet. But I think that we should be moving in that direction.

The latest draft of this PR introduces logic in the binding controller that is redundant with logic in the status controller. The status controller already reacts to informer events about StatusCollector objects (in `func (c *combinedStatusResolver) NoteStatusCollector`) by going through all the Binding objects and reacting to the current state of that StatusCollector. I think that the simplest and most consistent thing would be for this processing to be extended so that the condition gets reported first in the Binding's Conditions and then this, along with the Errors, propagates to the BindingPolicy's Status.

It seems to me that the current split between "status controller" and "binding controller" is spurious. Both are concerned with implementing aspects of BindingPolicy and Binding objects. I think the split was originally motivated by the observation that the status functionality can just react to informer events about WorkStatus objects. I think that is not true; we see that fully implementing all the proper reactivity involves a tight integration between the two controllers, casting doubt on whether they should be two separate controllers.

waltforme: Thanks for sharing these insights and thoughts!

On the 'redundant' logic: I tried to squeeze the logic into `func (c *combinedStatusResolver) NoteStatusCollector`, as shown in experimental commit https://github.com/kubestellar/kubestellar/pull/2643/commits/52b2b0aea4da2450367c9a1f461ed7567e24229a. The resulting code looks a little awkward to me because of three reasons. (a) the signature of the method has to be changed a lot to make it work, making it sort of overstuffed. The method should be a highly cohesive and internal thing. (b) there is no logging and error handling inside the method. (c) Most importantly, the experimental commit doesn't work because of #2687 .

On the 'redundant' logic, too: I don't think the logic in binding controller is redundant, because the informer events from _both_ bindingpolicies and statuscollectors should trigger the updates of the Condition. For example, say BindingPolicy `bp` requires StatusCollector `sc`, and user creates `bp` but hasn't created `sc` yet. At this time the user should be able to see the Condition of `bp`, without any informer events from statuscollectors.

On the split of the binding controller and the status controller: I fully agree that the split is obsolete now. During my coding I felt more and more data and logic are shared tightly between the two. One example is the shared 'broker' from binding controller to status controller, second example is the shared listers for workload objects, third but not last example is the logic for updating the Condition in this PR.

waltforme: BTW, I have a question regarding (b) in the listed principles.
> (b) for a given atom in a given API object, only one controller should author it

I remember that k8s allows multiple field managers to co-own a field. Is this feature conflicting with principle (b)?

MikeSpreitzer: > k8s allows multiple field managers to co-own a field

Yes, but in my opinion only one should be the decider of what the value is. Another might be in the habit of writing the same value, but even this gets tricky to keep out of trouble. In my opinion each atom [_should_](https://datatracker.ietf.org/doc/html/rfc2119) have one author.

MikeSpreitzer: I looked at commit 52b2b0aea4da2450367c9a1f461ed7567e24229a, and it puts more in NoteStatusCollector than I was thinking. I think that an approach that is more consistent with the overall style of these controllers would be to expand what is in a `combinedStatusResolution` to include a set of names of missing StatusCollector objects. The NoteStatusCollector method would update these sets incrementally, and the existing path for writing BindingStatus would naturally be extended with this information.

MikeSpreitzer: > the experimental commit doesn't work because of #2687 .

So I opened #2690 to look into that. The first commit changed the testing, and that exposed a bug. Later commits investigate and fix the bug. The fix does the kernel of what I described above, making `combinedStatusResolution` hold information about all relevant StatusCollector objects --- including ones that do not exist at the moment.

MikeSpreitzer:
Now that #2690 has merged, I think that this PR can be greatly revised to build on the fact that `combinedStatusResolution` now has the needed information to derive the new Condtion.

waltforme:
The [last commit](https://github.com/kubestellar/kubestellar/pull/2643/commits/ce6cbbd66d8b4d05c4eb08327fbd5ef6de3d2135) tries to use combinedStatusResolver's internal data to derive the Condition. Built on the fact that #2690 made "combinedStatusResolution hold information about all relevant StatusCollector objects --- including ones that do not exist at the moment."

This PR is not finished yet. Next thing is to where the Condition should go.

waltforme:
The plan https://github.com/kubestellar/kubestellar/pull/2643#issuecomment-2588534816 looks great so I'm revising this PR to align with it.

I reverted previous changes (and squashed corresponding commits), then implemented the steps in the plan.

waltforme:
I also ran into #2722 and this PR is fixing it.

pkg/binding/binding-controller.go
```diff
@@ -575,14 +577,14 @@ func (c *Controller) setupBindingInformer(ctx context.Context) error {
...
 			bdg := obj.(*v1alpha1.Binding)
 			logger.V(5).Info("Enqueuing reference to Binding because of informer add event", "name", bdg.Name, "resourceVersion", bdg.ResourceVersion)
-			c.workqueue.Add(bindingPolicyRef(bdg.Name))
+			c.workqueue.Add(bindingRef(bdg.Name))
 		},
 		UpdateFunc: func(old, new interface{}) {
 			oldBdg := old.(*v1alpha1.Binding)
 			newBdg := new.(*v1alpha1.Binding)
-			if oldBdg.Generation != newBdg.Generation {
+			if oldBdg.ResourceVersion != newBdg.ResourceVersion {
```
waltforme, line 585: Some notes on this change: I'm not sure why we use Generation here. What I observed is that (status controller) updating status of bindings won't increase Generation, thus would inform the binding controller on the updates of the Condition.

Use ResourceVersion may not be a good idea as well. Maybe we just remove this if condition?

api/control/v1alpha1/types.go
```diff
@@ -349,6 +349,10 @@ type Destination struct {
 }
 
 type BindingStatus struct {
+	// Currently Conditions is by design to be copied to BindingPolicy's Conditions.
+	// +optional
+	Conditions []BindingPolicyCondition `json:"conditions"`
```
MikeSpreitzer, line 354: It is Kubernetes design style to put `,omitempty` in the json tag.

pkg/binding/binding-controller.go
MikeSpreitzer: In general the filtering here should be only sensitive to stuff that matters. For example, changes to labels and annotations do not matter. In this particular case, I think the right thing to do is compare Generation (as a summary of Spec) and also compare Status.Conditions and Status.Errors (it would suffice to do an element-by-element comparison, without trying to recognize that reordering is insignificant).

pkg/binding/binding.go
```diff
@@ -109,7 +109,8 @@ func (c *Controller) syncBinding(ctx context.Context, bindingName string) error
 	policyWithStatus := policy.DeepCopy()
 	policyWithStatus.Status = v1alpha1.BindingPolicyStatus{
 		ObservedGeneration: policy.Generation,
-		Errors:             policyErrors,
+		Conditions:         binding.Status.Conditions,
+		Errors:             append(policyErrors, binding.Status.Errors...),
```
MikeSpreitzer, line 113: `binding.Status.Errors` were already added at line 91.
I think that it would be fine to change line 91 to initialize `policyErrors` to an empty slice.

pkg/status/binding.go
```diff
@@ -42,6 +49,82 @@ func (c *Controller) syncBinding(ctx context.Context, key string) error {
...
+	nameList := sets.List(statusCollectorNames)
+	return nameList, nil
+}
+
+func (c *Controller) createOrUpdateStatusCollectorAvailableCondition(ctx context.Context, bdg *v1alpha1.Binding, missingSCs []string) error {
+	// compose tentative condition where LastTransitionTime is TBD
+	conditionTentative := v1alpha1.BindingPolicyCondition{}
+	// Sort the names so that the string represenation of the list is deterministic,
+	// which is a useful property when comparing a tentive Condition with an existing Condition before updating.
+	sort.Strings(missingSCs)
```
MikeSpreitzer, line 103: `sets.List` already did the sort.

pkg/status/combinedstatus-resolver.go
```diff
@@ -80,7 +80,8 @@ type CombinedStatusResolver interface {
 	// The returned set contains the identifiers of combinedstatus objects
 	// that should be queued for syncing.
```
MikeSpreitzer, line 81: This comment needs to be updated so that it describes both returned values.

pkg/status/binding.go
```diff
@@ -42,6 +49,82 @@ func (c *Controller) syncBinding(ctx context.Context, key string) error {
 		c.workqueue.AddAfter(combinedStatusRef(combinedStatus.ObjectName.AsNamespacedName().String()), queueingDelay)
 	}
 
+	bp, err := c.bindingPolicyLister.Get(key)
+	if err != nil {
+		return fmt.Errorf("failed to get BindingPolicy %s from cache: %w", key, err)
+	}
+	missingStatusCollectors, err := c.identifyMissingStatusCollectors(bp)
```
MikeSpreitzer, line 56: This is still working too hard. It is not necessary to lookup the StatusCollector objects. Each `combinedStatusResolution` already knows which StatusCollectors it is missing; it is those for which `combinedStatusResolution.StatusCollectorNameToData[collectorName]` is `nil`.

MikeSpreitzer (commented):
Why still have 10 commits?

pkg/binding/binding-controller.go
waltforme: Done.

api/control/v1alpha1/types.go
waltforme: Done. Also fixed same thing in `type BindingPolicyStatus`.

pkg/binding/binding.go
waltforme: Done.

pkg/status/binding.go
waltforme: Right, you and me know that but the function knows nothing about the order. And, other developers may not pass in a sorted slice. Furthermore, I have some concern that `sets.List` may change its API. So let's be extra safe here?

pkg/status/combinedstatus-resolver.go
waltforme: Updated.

pkg/status/binding.go
waltforme: Done.

waltforme:
I've responded to all the comments and consolidated the git history.

pkg/binding/binding-controller.go
```diff
@@ -601,6 +603,29 @@ func (c *Controller) setupBindingInformer(ctx context.Context) error {
...
+	for i, c := range old.Status.Conditions {
+		if !v1alpha1.AreConditionsEqual(c, new.Status.Conditions[i]) {
+			return true
+		}
+	}
+	for i, e := range old.Status.Errors {
+		if e != new.Status.Errors[i] {
+			return true
+		}
+	}
```
MikeSpreitzer, line 624: `abstract.SliceEqual` will handle cases where golang `==` is the right equality test.

pkg/binding/binding-controller.go
```diff
@@ -125,6 +125,7 @@ type Controller struct {
 	bindingLister               controllisters.BindingLister
 	bindingPolicyInformer       cache.SharedIndexInformer
 	bindingPolicyLister         controllisters.BindingPolicyLister
+	statusCollectorLister       controllisters.StatusCollectorLister
```
MikeSpreitzer, line 128: This is not needed now, right?

pkg/status/binding.go
MikeSpreitzer: I am not concerned about `sets.List` ceasing to sort. The sorting is the leading thing said about this function; see https://github.com/kubernetes/apimachinery/blob/v0.29.13/pkg/util/sets/set.go#L202 . It is the reason that this is a function rather than a method. We rely on this function doing sorting in other places.

pkg/status/combinedstatus-resolver.go
```diff
@@ -102,6 +103,8 @@ type CombinedStatusResolver interface {
 	//
 	// The returned pointers are expected to be read-only.
 	ResolutionExists(name string) (string, util.ObjectIdentifier, bool)
+
+	ResolutinosForBinding(bdgName string) map[util.ObjectIdentifier]*combinedStatusResolution
```
MikeSpreitzer, line 107: (A) Typo in this method's name: "Resolutinos" instead of "Resolutions".

(B) When a mutable data structure is returned, you need a comment about whether caller and/or anyone else can mutate it after the return.

pkg/status/combinedstatus-resolver.go
```diff
@@ -428,6 +436,13 @@ func (c *combinedStatusResolver) ResolutionExists(name string) (string, util.Obj
 	return key.bindingName, key.sourceObjectIdentifier, true
 }
 
+func (c *combinedStatusResolver) ResolutinosForBinding(bdgName string) map[util.ObjectIdentifier]*combinedStatusResolution {
+	c.RLock()
+	defer c.RUnlock()
+
+	return c.bindingNameToResolutions[bdgName]
```
MikeSpreitzer, line 443: `c.bindingNameToResolutions` should only be accessed while holding the lock in `c.RWMutex`.

I think a workable alternative is to redefine this method to return the set of missing StatusCollector names.

pkg/binding/binding-controller.go
MikeSpreitzer: Also, I think that it would be good to enhance `abstract` with an alternative form of `SliceEqual` that takes a comparison function. I would curry it too, so that we can have an algebra of comparison functions.

```golang
func SliceEqualFunction[Elt comparable](equal func (Elt, Elt) bool) func (slice1, slice2 []Elt) bool {...}
```

waltforme: Used `abstract.SliceEqual`, thanks!

pkg/status/combinedstatus-resolver.go
waltforme: Fixed.

pkg/status/combinedstatus-resolver.go
waltforme: Done.

pkg/status/binding.go
waltforme: Okay, I removed the sort here.

pkg/binding/binding-controller.go
waltforme: Just saw the 2nd comment. Would that be overkill here? Or are you saying we contribute to Go?

MikeSpreitzer: I am not proposing a contribution to golang here.

pkg/binding/binding-controller.go
```diff
@@ -601,6 +602,23 @@ func (c *Controller) setupBindingInformer(ctx context.Context) error {
 	return nil
 }
 
+func shouldQueueBindingOnUpdate(old, new *v1alpha1.Binding) bool {
+	if old.Generation != new.Generation {
+		return true
+	}
+	if len(old.Status.Conditions) != len(new.Status.Conditions) || len(old.Status.Errors) != len(new.Status.Errors) {
```
MikeSpreitzer, line 609: FYI, `abstract.SliceEqual` checks for equal lengths, among other things.

pkg/status/combinedstatus-resolver.go
MikeSpreitzer: Sorry, I was not clear enough. `c.bindingNameToResolutions` is a two-layer map. It maps a string to another _map_. `c.RWMutex` applies to both layers. It is not valid to return an inner map after releasing `c.RWMutex`.

The comment on `c.bindingNameToResolutions` does not promise that the inner maps are immutable. And indeed, they are mutated in some places.

waltforme: Indeed. I implemented the suggested 'workable alternative' above.

pkg/binding/binding-controller.go
waltforme: Improved.

BTW glad to see your recent contribution to `abstract.SliceEqual`!

pkg/status/combinedstatus-resolver.go
```diff
@@ -102,6 +103,10 @@ type CombinedStatusResolver interface {
 	//
 	// The returned pointers are expected to be read-only.
 	ResolutionExists(name string) (string, util.ObjectIdentifier, bool)
+
+	// MissingStatusCollectors returns the names of StatusCollector object(s)
+	// that are required by a binding but do not exist.
```
MikeSpreitzer, line 108: Since the caller is relying on it, I would document in the comment here that the returned list is sorted.

pkg/status/combinedstatus-resolver.go
```diff
@@ -102,6 +103,10 @@ type CombinedStatusResolver interface {
...
+	MissingStatusCollectors(bindingName string) []string
```
MikeSpreitzer, line 109: Since the caller is relying on it, I would document in the comment here that the returned list is sorted.

And again, here is a return of a mutable data structure. The comment should document that the caller may mutate the slice and the callee does not cause anything to mutate the slice after return.

MikeSpreitzer (approved):
I think that this is good enough to merge.
There also needs to be a corresponding update to the architecture document.

/lgtm
/approve

PR #2645, opened 2024-12-04
✨Introduce the capability of controlling if OCM should be installed in a Control Plane

francostellari:
## Summary

Introduce the capability of controlling if OCM should be installation in a Control Plane.

The `core-chart` YAML has been modified to accept a `install_clusteradm: true|false` value for each ITS CP to enable/disable the installation of OCM inside the CP.

Usage example: `--set-json='ITSes=[{"name":"its0","type":"host","install_clusteradm":false}]'`

NOTES:
1. this feature is mostly useful for an ITS of type `host` in a cluster with OCM already installed.
2. if OCM is not installed and it is not present, then the installation of the `status-addon` will fail

## Related issue(s)

Fixes #


core-chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,69 @@
...
+    value: "/tmp"
+  - name: HELM_CACHE_HOME
+    value: "/tmp"
+  - name: KUBECONFIG
+    value: "{{"/etc/kube/{{.ITSkubeconfig}}"}}"
+  volumeMounts:
+  - name: kubeconfig
+    mountPath: "/etc/kube"
+    readOnly: true
+{{- end }}
```
MikeSpreitzer, line 69: FYI, reviewers would see less red if there were a linebreak at the end of this file.

core-chart/templates/postcreatehooks/its-with-clusteradm.yaml
MikeSpreitzer, line 1: This file has a lot in common with `its-without-clusteradm.yaml`. There would be less duplicated material if there was one unified template that had some content conditional on the setting of a PostCreateHook "var".

core-chart/templates/controlplanes/its.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
 spec:
   backend: shared
   type: {{ $cp.type | default "vcluster" }}
-  postCreateHook: its
+  postCreateHook: {{ (eq $cp.install_clusteradm false) | ternary "its-without-clusteradm" "its-with-clusteradm" }}
```
MikeSpreitzer, line 10: What happens here if the user has not provided a value for `install_clusteradm` in the `cp` ?

docs/content/direct/core-chart.md
```diff
@@ -86,7 +86,7 @@ InstallPCHs: true
 
 # List the Inventory and Transport Spaces (ITSes) to be created by the chart
 # Each ITS consists of a mandatory unique name and an optional type, which could be either host or vcluster (default to vcluster, if not specified)
-ITSes: # ==> installs ocm + ocm-status-addon
+ITSes: # ==> installs ocm (optional) + ocm-status-addon
```
MikeSpreitzer, line 89: I think that your earlier suggestion of how to frame this is better: it is a question of whether OCM is already installed. Having OCM is not optional; the question is whether it is already there. In the case where the user says that OCM is already installed, the chart should check and somehow let the user know if OCM is not actually already there.

MikeSpreitzer: Or perhaps it is unnecessary to have an ordering requirement that OCM is installed in the ITS _before_ the core Helm chart is instantiated.

core-chart/templates/postcreatehooks/its-with-clusteradm.yaml
francostellari: Unfortunately as discussed elsewhere, helm code cannot be templated using PostCreateHook vars.

core-chart/templates/controlplanes/its.yaml
francostellari: absence of the value will evaluate as NOT == false and thus install OCM

docs/content/direct/core-chart.md
francostellari: AFAK, helm cannot verify the existence or wait for a specified condition in the cluster... being the presence of a CRD or the existence/reediness of a pod. As the user needs to go out of his/her way to not install OCM with the chart we can hope that he/she chose with some consideration. We can also add more text to clarify your point in words.

core-chart/templates/postcreatehooks/its-with-clusteradm.yaml
francostellari: We can abstract the RBAC stuff to the `_helper.tpl`

francostellari: I personally do not like very much having 2 separated but similar PCHs... more confusion.... this is what I could come up in that scenario.

While more of a tric that good coding, using the `init` | `--help` would have been the less code change and less  confusing

ITS install_clusteradm=T|F ==> PCH var clusteradm_action=`init` | `--help` ==> container command `clusteradm {{.clusteradm_action}}`

scripts/create-kubestellar-demo-env.sh
```diff
@@ -242,8 +242,8 @@ kflex ctx --overwrite-existing-context wds2
 kflex ctx --overwrite-existing-context its1
 
 echo -e "\nWaiting for OCM cluster manager to be ready..."
-kubectl --context $k8s_platform-kubeflex wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.its}=true' --timeout 24h
-kubectl --context $k8s_platform-kubeflex wait -n its1-system job.batch/its --for condition=Complete --timeout 24h
+kubectl --context kind-kubeflex wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.its-with-clusteradm}=true' --timeout 24h
+kubectl --context kind-kubeflex wait -n its1-system job.batch/its-with-clusteradm --for condition=Complete --timeout 24h
```
MikeSpreitzer, line 246: The changes here from `$k8s_platform` to `kind` look like a regression?

core-chart/templates/postcreatehooks/its-with-clusteradm.yaml
MikeSpreitzer: Oh, sorry, I had temporary oversight. You are right about the basic problem here.

core-chart/templates/controlplanes/its.yaml
MikeSpreitzer: Thank you.

scripts/create-kubestellar-demo-env.sh
francostellari: I must have messed up the rebase. I will fix it.

MikeSpreitzer (commented):
/lgtm

PR #2655, opened 2024-12-10
✨ add kubestellar-snapshot output placeholder to bug_report.…

clubanderson:
…yaml

adding kubestellar-snapshot output to bug report template

## Related issue(s)

Fixes #


francostellari:
@clubanderson I suggest to require the verbose flag `-V`|`--verbose` for extra info.

Optionally, we could also ask for the the deployment YAML with the flag `-Y`|`--yaml` and/or the logs of the pods with the flag `-L`|`--log`.
Either of the last options would generate a local `kubestellar-snaphot.tar.gz` that could be uploaded in the bug report.

For some non-developer users, who may not close the repo, it may also be useful to enable running the script without downloading it:

```shell
bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/heads/main/scripts/k
ubestellar-snapshot.sh) -V -Y -L
```

PR #2663, opened 2024-12-11
✨  Add Binding objects to snapshot output

MikeSpreitzer:
## Summary
This PR extends the snapshot script to dump the Binding objects too.

## Related issue(s)

Fixes #

### DIFF
--- a/scripts/kubestellar-snapshot.sh
+++ b/scripts/kubestellar-snapshot.sh
@@ -495,6 +495,7 @@ for j in "${!cp_pch[@]}" ; do
     fi
 done
 
+
 ###############################################################################
 # Listing Bindings
 ###############################################################################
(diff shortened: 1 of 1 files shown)


francostellari:
- stylistically rest of the script uses 2 line spaces to separate blocks
- I suggest replacing COLOR_RED with either COLOR_ERROR|COLOR_WARNING in case down the line somebody changes the color scheme

MikeSpreitzer:
I revised the way Binding objects are listed: now the normal listing enumerates the objects, and in the case of empty cross product it shows the reason(s) why, and there is no longer an empty namespace listed for cluster-scoped objects. Following is a picture of example output.

image

francostellari:
/lgtm
Thank you

PR #2664, opened 2024-12-11
🌱  Remove the extra ClusterRoleBinding from example 2

### LINKED ISSUE #2658: Confusing TODO notes in example scenario 2

### LINKED ISSUE #2659: docs: scenario 2 why the need for ClusterRoleBinding
In the docs in https://docs.kubestellar.io/release-0.25.1/direct/example-scenarios/#scenario-2-out-of-tree-workload

Can you provide more explanation why suddenly I need to make the service account kubeflex-controller-manager cluster-admin to be able to deploy custom CRs to WECs.

### LINKED ISSUE #2660: Example scenario 2 too mum on early deletion of ClusterRoleBinding
Currently example scenario 2 has the following remark, right after the loop installing the AppWrapper CRD in the WECs.

> If desired, you may remove the `kubeflex-manager-cluster-admin-rolebinding` after the kubestellar-controller-manager is started, with the command `kubectl --context "$host_context" delete clusterrolebinding kubeflex-manager-cluster-admin-rolebinding`.

A new user does not necessarily understand whether or why they might desire this. It needs a little more explanation.

What this remark is trying to say is that one particular step of the teardown can be moved earlier in the recipe, possibly as far as where that remark appears.

MikeSpreitzer:
## Summary
Example scenario 2 had a poorly motivated extra ClusterRoleBinding. I tested without it, and had no problem. I think it is not needed anymore. This PR removes it.

## Related issue(s)

Fixes #2658 
Fixes #2659 
Fixes #2660

### DIFF
--- a/docs/content/direct/example-scenarios.md
+++ b/docs/content/direct/example-scenarios.md
@@ -132,26 +132,6 @@ KubeStellar can handle the case where the CRD is part of the workload,
 this example concerns the case where the CRD is established in the
 WECs by some other means.
 
-In order to run this scenario using the post-create-hook method you need
-the raise the permissions for the kubeflex controller manager (TODO 1: move this material and its undo to [the doc on WDS](wds.md); TODO 2: why is this needed? Is it needed for the core chart too? can we remove the need for this?):
-
-```shell
-kubectl --context "$host_context" apply -f - <<EOF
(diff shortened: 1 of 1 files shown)


pdettori (commented):
/lgtm

PR #2680, opened 2024-12-13
✨  Update get-started.md to match demo-env

clubanderson:
## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/direct/get-started.md
+++ b/docs/content/direct/get-started.md
@@ -133,7 +133,6 @@ host_context=kind-kubeflex
 its_cp=its1
 its_context=its1
 wds_cp=wds1
-wds_host=wds2
 wds_context=wds1
 wec1_name=cluster1
 wec2_name=cluster2


MikeSpreitzer (commented):
This is functional, but I remain unconvinced that having `scripts/create-kubestellar-demo-env.sh` created wds2 is the best approach.

/lgtm

Having `scripts/create-kubestellar-demo-env.sh` create wds2 strikes me as putting unnecessary complexity in the "Get Started" configuration and recipe. I think that we could both have a simpler "Get Started" configuration&recipe and support Kueue integration, because adding wds2 is an addition not a change. I verified this as follows. I modified `scripts/create-kubestellar-demo-env.sh` to _not_ create wds2, and ran it. Then I issued the following command.

```shell
helm upgrade --install ks-core-ext oci://ghcr.io/kubestellar/kubestellar/core-chart \
    --version $kubestellar_version \
    --set='kubeflex-operator.install=false,InstallPCHs=false' \
    --set-json='WDSes=[{"name":"wds2", "type":"host"}]' \
    --set-json='verbosity.default=5'
```

That caused wds2 to be created. Then I defined `wds_cp=wds2` and `wds_context=$host_context`, and then proceeded to do Example Scenario 2. It succeeded.

PR #2689, opened 2024-12-18
✨  Add OSA and OCM versions to snapshot script

MikeSpreitzer:
## Summary
This PR extends the kubestellar-snapshot script to report the versions of the ocm-status-addon and of the OCM manager.

## Related issue(s)

Fixes #

### DIFF
--- a/scripts/kubestellar-snapshot.sh
+++ b/scripts/kubestellar-snapshot.sh
@@ -400,10 +400,13 @@ for i in "${!cps[@]}" ; do # for all control planes in context ${context}
         echostatus "$its_status"
         echo -n -e "  - Status addon controller: pod=${COLOR_INFO}$status_pod${COLOR_NONE}, ns=${COLOR_INFO}$status_ns${COLOR_NONE}, version=${COLOR_INFO}$status_version${COLOR_NONE}, status="
         echostatus "$status_status"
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
Following is an example of the revised Control Planes.ITS section of the output.

```
- its1: type=vcluster, pch=its-with-clusteradm, context=default/c106-e-us-south-containers-cloud-ibm-com:31891/IAM#mspreitz@us.ibm.com, namespace=its1-system
  - Post Create Hook: pod=its-with-clusteradm-pnjp7, ns=its1-system, status=succeeded
  - Status addon controller: pod=addon-status-controller-87c8ff67d-8xwjm-x-open-clust-47858caac8, ns=its1-system, version=0.2.0-rc14, status=running
  - Open-cluster-manager version=0.14.0
```

MikeSpreitzer:
The second commit changes the way that the version of the OCM manager is determined, from looking at the shadow Pod in the KubeFlex hosting cluster to looking at the original Deployment in the ITS. This is good because it removes a place where it is assumed that the ITS is either the hosting cluster or from a ControlPlane of type "vcluster". BTW, more of the script should be updated to work with general spaces.

francostellari:
I suggest adding `:` after `- Open-cluster-manager` for consistency

MikeSpreitzer:
And the force-push to e600831ea adds the colon in the other place.

Issue #2579, opened 2024-11-11
bug: Cleanup needs a wait for port 9443 to really be free


MikeSpreitzer:
### Describe the bug

I was running the following loop and in one iteration it failed.

```shell
while time bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/tags/v0.25.0-rc.3/scripts/create-kubestellar-demo-env.sh); do
    date
    sleep 1  # Optional: adds a delay of 1 second between runs
done; date
```

The end of the failing iteration looked like the following.

```
Creating cluster "kubeflex" ...
 ✓ Ensuring node image (kindest/node:v1.29.2) 🖼
 ✗ Preparing nodes 📦
Deleted nodes: ["kubeflex-control-plane"]
ERROR: failed to create cluster: command "docker run --name kubeflex-control-plane --hostname kubeflex-control-plane --label io.x-k8s.
kind.role=control-plane --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --
volume /var --volume /lib/modules:/lib/modules:ro -e KIND_EXPERIMENTAL_CONTAINERD_SNAPSHOTTER --detach --tty --label io.x-k8s.kind.clu
ster=kubeflex --net kind --restart=on-failure:1 --init=false --cgroupns=private --publish=0.0.0.0:9443:443/TCP --publish=127.0.0.1:407
09:6443/TCP -e KUBECONFIG=/etc/kubernetes/admin.conf kindest/node:v1.29.2@sha256:51a1434a5397193442f0be2a297b488b6c919ce8a3931be0ce822
606ea5ca245" failed with error: exit status 125
Command Output: fb6fc48106c423edac2c20678f63268bcd2e20e468b267a3d028a06d04122d34
docker: Error response from daemon: driver failed programming external connectivity on endpoint kubeflex-control-plane (993de472395027
a2963b5a0d4c5cb59c17dc0752617672dd957de358cc62adb2): Bind for 127.0.0.1:40709 failed: port is already allocated.

real    0m9.620s
user    0m2.141s
sys     0m1.478s
Mon Nov 11 07:03:45 UTC 2024
mspreitz@mjs-dev7a:~$ touch: cannot touch '/tmp/tmp.uKbG9H5hwf/cluster1.success': No such file or directory
touch: cannot touch '/tmp/tmp.uKbG9H5hwf/cluster2.success': No such file or directory
```

[...]

PR #2698, opened 2024-12-30
🌱  Start the move of check_pre_reqs.sh

MikeSpreitzer:
## Summary
This PR takes the first steps in moving the `check_pre_req.sh` script from `hack/` to `scripts/`. The remaining steps will have to wait until after releases, as noted in new material in `docs/content/direct/release.md`.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-start-moving-check-prereqs/direct/pre-reqs/

## Related issue(s)

This partially addresses #2697

PR #2699, opened 2025-01-03
🐛 bug fix:Improve k3s script to handle port allocation issues

pratik-mahalle:
This pull request improves the k3s script to handle port allocation issues more gracefully. The changes include:

  - Adding a wait mechanism to ensure that port 9443 is free before attempting to use it.
  - Enhancing the cleanup process to ensure that all resources are properly cleaned up, preventing port conflicts.

These improvements aim to address the issue where the k3s script fails due to port allocation conflicts, as described in issue [#2579](https://github.com/kubestellar/kubestellar/issues/2579).


scripts/create-kubestellar-demo-env.sh
```diff
@@ -123,6 +123,12 @@ checking_cluster() {
 command_exists() {
     command -v "$1" >/dev/null 2>&1
 }
+
+# Function to check if a port is free
+is_port_free() {
+    ! lsof -i:9443
```
MikeSpreitzer, line 129: Wouldn't `-itcp:9443` be better?

scripts/create-kubestellar-demo-env.sh
```diff
@@ -177,10 +189,10 @@ kubectl config use-context $k8s_platform-kubeflex
...
 for cluster in "${clusters[@]}"; do
     if ! [ -f "${cluster_log_dir}/${cluster}.success" ]; then
-	echo -e "\033[0;31mX\033[0m Creation of cluster $cluster failed!" >&2
-	cat "${cluster_log_dir}/${cluster}.log" >&2
-	some_failed=true
-	continue
+    echo -e "\033[0;31mX\033[0m Creation of cluster $cluster failed!" >&2
+    cat "${cluster_log_dir}/${cluster}.log" >&2
+    some_failed=true
+    continue
```
MikeSpreitzer, line 195: Replacing tabs is good, but I think that these tabs need to be replaced by eight spaces, so that the body of the "then" clause is nested more than the "if" statement.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -287,4 +299,4 @@ wec1_context=\$wec1_name
 wec2_context=\$wec2_name
 label_query_both=location-group=edge
 label_query_one=name=cluster1
-EOF
+EOF
```
MikeSpreitzer, line 302: Why delete the final linebreak? This deletion makes GitHub show a red circle at the end.

MikeSpreitzer:
Please do not merge the branch named "main" into a branch being used to create a PR. See https://docs.google.com/document/d/1iwkhpoV76-J-3G-Rif4B1cmL9SWBO_f_fE6WK8jvsDc for an explanation of why.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -157,10 +168,11 @@ for cluster in "${clusters[@]}"; do
         kind create cluster --name "${cluster}" >"${cluster_log_dir}/${cluster}.log" 2>&1 && touch "${cluster_log_dir}/${cluster}.success" &
     else
         k3d cluster create --network k3d-kubeflex "${cluster}" >"${cluster_log_dir}/${cluster}.log" 2>&1 && touch "${cluster_log_dir}/${cluster}.success" &
-        wait
```
MikeSpreitzer, line 160: I do not know why @clubanderson made the previous command a fork and immediately followed it with a `wait`. This combination looks unnecessarily complex. It is equivalent (since there are no other forked processes in sight) to making the previous command _not_ fork, right? @clubanderson, is there a specific reason for this structure? Is there a reason to not simply fork the previous command?

scripts/create-kubestellar-demo-env.sh
```diff
@@ -157,10 +168,11 @@ for cluster in "${clusters[@]}"; do
...
     fi
 done
 
+wait
```
MikeSpreitzer, line 174: Why add a `wait` here, when there is already one on new line 186?

pratik-mahalle: I see your point about the redundant wait. I'll remove the extra wait on line 186.

scripts/create-kubestellar-demo-env.sh
pratik-mahalle: You're right, the command could be simplified. @clubanderson, could you clarify if there's a specific reason for using fork followed by wait?

scripts/create-kubestellar-demo-env.sh
pratik-mahalle: I didn't realize the final linebreak deletion would cause an issue. I'll restore it to avoid the red circle.

scripts/create-kubestellar-demo-env.sh
pratik-mahalle: Good point about the tabs. I'll replace them with eight spaces to ensure proper nesting.

scripts/create-kubestellar-demo-env.sh
pratik-mahalle: Thank you for the suggestion! -itcp:9443 does sound better. I'll make the update.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: But the `wait` on line 186 waits for an additional forked process, and we still want that to happen, right?

MikeSpreitzer: Are you sure we want the wait earlier rather than later? The later wait allows more concurrency (which, to be fair, is a mixed blessing).

scripts/create-kubestellar-demo-env.sh
clubanderson: I don't recall why that wait is after the fork right now. It might just be a left over from the last time this went through review and edit. If it serves no purpose, just remove it and test. The goal is to have a working script that is efficient with respect to the time it takes to setup a demo. If there are places where async execution can be used and can help, then let's use it. If not, then lets remove it.

pratik-mahalle: Ok, Thanks for this!

scripts/create-kubestellar-demo-env.sh
pratik-mahalle: Not sure, I will remove it and will check if script run perfectly or not.

MikeSpreitzer: That sounds backwards to me; keeping the wait on line 186 and deleting the earlier wait is the approach with more concurrency.

In addition to #2579, I have found other failures: #2700, #2701, and #2702. The first of those three is clearly a concurrency problem. The others are not clear to me. And I have been continued testing --- and none of those failures has recurred.

MikeSpreitzer:
Don't forget https://github.com/kubestellar/kubestellar/pull/2699#issuecomment-2569726615

pratik-mahalle:
@MikeSpreitzer I am not sure why my test is failing as script is working perfect locally. Can you please explain it?

scripts/check_pre_req.sh
MikeSpreitzer, line 1: We can't just _move_ this file in one PR; see #2698 .

MikeSpreitzer:
https://github.com/kubestellar/kubestellar/actions/runs/12748399137/job/35528487572?pr=2699#step:7:23 shows the problem with the current version of this PR.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -167,7 +172,7 @@ if [ "$k8s_platform" == "kind" ]; then
 else
     k3d cluster create -p "9443:443@loadbalancer" --k3s-arg "--disable=traefik@server:*" kubeflex
     sleep 15
-    helm install ingress-nginx ingress-nginx --set "controller.extraArgs.enable-ssl-passthrough=" --repo https://kubernetes.github.io/ingress-nginx --version 4.11.3 --namespace ingress-nginx --create-namespace
+    helm install ingress-nginx ingress-nginx --set "controller.extraArgs.enable-ssl-passthrough=" --repo https://kubernetes.github.io/ingress-nginx --version 4.11.3 --namespace ingress-nginx --create-[...]
```
MikeSpreitzer, line 175: I don't know what went wrong here, but I suspect that you did not intend this change.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -13,8 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Deploys Kubestellar environment for demo purposes on kind or k3d.
```
MikeSpreitzer, line 16: Why delete this comment?

scripts/create-kubestellar-demo-env.sh
```diff
@@ -119,26 +133,18 @@ checking_cluster() {
...
-##########################################
-
 echo -e "\nStarting environment clean up..."
 echo -e "Starting cluster clean up..."
 
 if command_exists "k3d"; then
     cluster_clean_up "k3d cluster delete kubeflex" &
     cluster_clean_up "k3d cluster delete cluster1" &
     cluster_clean_up "k3d cluster delete cluster2" &
-    wait
```
MikeSpreitzer, line 135: We deliberately put a wait between the k3d cleanup and the kind cleanup because of observed problems with doing them concurrently.

MikeSpreitzer:
Merging `main` into this PR's branch is not going to fix the things that I pointed out about the changes in this PR's branch.

Merging `main` into a PR's branch is not recommended practice in this project. See https://docs.google.com/document/d/1iwkhpoV76-J-3G-Rif4B1cmL9SWBO_f_fE6WK8jvsDc for a description of the relevant discipline that we are trying to follow. The recommended way to pick up changes that have merged since starting work on a PR is to `git rebase` the PR's branch onto the latest edition of `main`.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -152,58 +173,80 @@ echo -e "\nCreating two $k8s_platform clusters to serve as example WECs"
...
-	some_failed=true
-	continue
+        echo -e "\033[0;31mX\033[0m Creation of cluster $cluster failed!" >&2
+        cat "${cluster_log_dir}/${cluster}.log" >&2
+        some_failed=true
+        continue
     fi
     echo -e "\033[33m✔\033[0m Cluster $cluster was successfully created"
-    kubectl config rename-context "${k8s_platform}-${cluster}" "${cluster}" >/dev/null 2>&1
+    kubectl config rename-context "${k8s_platform}-${cluster}" "${cluster}" >/dev/null 2>&1 || true
```
MikeSpreitzer, line 224: Why add `|| true` here? It masks failures.

hack/check_pre_req.sh
MikeSpreitzer, line 1: We can't just delete this file. I suspect you did not intend this deletion?

MikeSpreitzer:
I see that this PR now also makes pervasive changes to linebreaking. FYI, mixing a pervasive change to formatting in with substantive changes makes reviewing harder.

pratik-mahalle:
Thanks for the feedback! I’ll remove `|| true` so that the script fails when a command fails—that was unintentional. As for the line breaks, I’ll revert those changes. I understand that mixing formatting changes with substantive updates makes reviewing harder. In the future, I’ll keep formatting-only changes in a separate commit or PR.

hack/check_pre_req.sh
MikeSpreitzer, line 1: Do not delete this file. Doing so will create problems in scripts that fetch it.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer, line 135: Do not delete this wait. Past experience shows that we get failures if there is no synchronization barrier between the two groups of cluster delete commands.

pratik-mahalle:
Hey @MikeSpreitzer, My branch is messed up. I will make another pull request to address this issue.

pratik-mahalle:
I haved open another pull request for this issue

MikeSpreitzer:
OK

PR #2718, opened 2025-01-12
✨CI: Added sbom generation in goreleaser

### LINKED ISSUE #2613: feature: SBOM needed to clomonitor tests

jaydee029:
## Summary
Added sbom generation within go-releaser itself.
## Related issue(s)

Fixes #2613

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -53,4 +53,9 @@ release:
   prerelease: auto
   mode: keep-existing
   skip_upload: true
+sboms:
+  - id: source
+    artifacts: archive
+  - id: archive
+    artifacts: archive
 
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -55,7 +55,7 @@ release:
   skip_upload: true
 sboms:
   - id: source
-    artifacts: archive
+    artifacts: source
   - id: archive
     artifacts: archive
 
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -56,6 +56,4 @@ release:
 sboms:
   - id: source
     artifacts: source
-  - id: archive
-    artifacts: archive
 
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -25,7 +25,7 @@ jobs:
     - uses: actions/setup-go@v5
       with:
         go-version: v1.22
-
+         
     - name: Delete non-semver tags
       run: 'git tag -d $(git tag -l | grep -v "^v")'
 


.goreleaser.yaml
```diff
@@ -53,4 +53,9 @@ release:
   prerelease: auto
   mode: keep-existing
   skip_upload: true
+sboms:
```
MikeSpreitzer, line 56: (A) FYI, I found some documentation about this at https://goreleaser.com/customization/sbom/#usage

(B) The two objects in the list here have the same value for `artifacts`. That is, they make SBOMs about the same subject. I suspect this is not what you intended?

.goreleaser.yaml
```diff
@@ -53,4 +53,9 @@ release:
...
+  - id: source
+    artifacts: archive
+  - id: archive
+    artifacts: archive
```
MikeSpreitzer, line 60: Given that the archives are not being attached to the release (see line 55), why create SBOMs for those archives?

MikeSpreitzer:
Thank you for this suggestion. However, I do not understand how this creates any SBOMs that we need. The things that are published by doing a release of this repository are:
- The GitHub release, which has archives of the sources attached;
- Some container images, built with `ko` (which has its own ability to create SBOMs);
- A Helm chart (published as an OCI image).

.goreleaser.yaml
jaydee029: My apologies, this was a typo , I have corrected this.

.goreleaser.yaml
jaydee029: Right only the source archives exits currently in the Github release, but I added archive for forward compatibility, should I remove that ?

jaydee029:
> Thank you for this suggestion. However, I do not understand how this creates any SBOMs that we need. The things that are published by doing a release of this repository are:
> 
> * The GitHub release, which has archives of the sources attached;
> * Some container images, built with `ko` (which has its own ability to create SBOMs);
> * A Helm chart (published as an OCI image).

Right, but the thing is, this issue was specifically created for the Github release SBOMs, the clomonitor tests won't pass unless there is an sbom in the github release as long as there is even a single artifact even the source artifact.

For the other things that you mentioned, I can work on another PR or we an raise another issue.
Thanks

MikeSpreitzer:
Oh. I see that https://clomonitor.io/docs/topics/checks/#software-bill-of-materials-sbom says that an SBOM is required, without regard to whether there are binary archives. And indeed https://clomonitor.io/projects/cncf/kubestellar is still complaining about a lack of SBOM.

.goreleaser.yaml
MikeSpreitzer: Yes, I think it is better to not try to make an SBOM for something that does not exist.

jaydee029: made the changes.

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

MikeSpreitzer:
I just tested this by making test release 0.26.0-alpha.5, and the SBOM did _not_ get created.

PR #2765, opened 2025-01-29
 🐛 bug fix:Improve k3s script to handle port allocation issues and improved concurrrency

pratik-mahalle:
This pull request improves the k3s script to handle port allocation issues more gracefully. The changes include:

    Adding a wait mechanism to ensure that port 9443 is free before attempting to use it.
    Enhancing the cleanup process to ensure that all resources are properly cleaned up, preventing port conflicts.

These improvements aim to address the issue where the k3s script fails due to port allocation conflicts, as described in issue https://github.com/kubestellar/kubestellar/issues/2579.


pratik-mahalle:
Hey @MikeSpreitzer, Look at this pr as I think all looks good now to check!
Please tell me anything went wrong!

scripts/create-kubestellar-demo-env.sh
```diff
@@ -119,26 +133,18 @@ checking_cluster() {
...
-##########################################
-
 echo -e "\nStarting environment clean up..."
 echo -e "Starting cluster clean up..."
 
 if command_exists "k3d"; then
     cluster_clean_up "k3d cluster delete kubeflex" &
     cluster_clean_up "k3d cluster delete cluster1" &
     cluster_clean_up "k3d cluster delete cluster2" &
-    wait
```
MikeSpreitzer, line 135: This `wait` should not be deleted. Experience shows that for a user who has both `kind` and `k3d` clusters, sometimes the two sets of delete commands interfere with each other when run concurrently.

MikeSpreitzer (approved):
I will restore that needed `wait` in a follow-on PR.
/lgtm
/approve
Thanks!

scripts/create-kubestellar-demo-env.sh
```diff
@@ -46,6 +44,23 @@ fi
...
+wait_for_port_free() {
+    local port=$1
+    while lsof -i:$port >/dev/null 2>&1; do
+        echo "Waiting for port $port to be free..."
+        sleep 5
+    done
+}
+
+# Wait for port 9443 to be free before proceeding
+wait_for_port_free 9443
```
MikeSpreitzer, line 62: This is too soon. The wait for the port to free needs to come after the deletion of the old KubeFlex hosting cluster.

MikeSpreitzer:
BTW, this is an alternative version of #2699

scripts/create-kubestellar-demo-env.sh
pratik-mahalle: Move the port-free check after deleting the old KubeFlex cluster, rather than doing it at the start. This ensures you only wait for port 9443 **once the old cluster has actually** **been torn down**, thus freeing that port.

pratik-mahalle:
> BTW, this is an alternative version of #2699

Yes

PR #2768, opened 2025-01-29
✨Update the Helm chart to support external cluster as an ITS CP

### LINKED ISSUE #2684: Too-strict prereqs in core Helm chart doc?
https://docs.kubestellar.io/release-0.26.0-alpha.2/direct/core-chart/#pre-requisites says that this chart can only be used with a hosting cluster that is OpenShift or `kind` or `k3s`. But this chart can actually be used more broadly, right?

francostellari:
## Summary

Example usage:

```shell
kind create cluster --name ext1
clusteradm init

scripts/create-kind-cluster-with-SSL-passthrough.sh --name kubeflex --port 9443

scripts/create-external-bootstrap-secret.sh --controlplane its1 --source-context kind-ext1 --address https://ext1-control-plane:6443 --verbose

helm dependency update core-chart

  --set-json='ITSes=[{"name":"its1","type":"external","install_clusteradm":false}]' \
  --set-json='WDSes=[{"name":"wds1"}]'
```

Requirements: kubeflex >=0.8

## Related issue(s)

Fixes #2684


core-chart/values.yaml
```diff
@@ -75,14 +75,14 @@ transport_controller:
 
 
 # List the Inventory and Transport Spaces (ITSes) to be created by the chart
-# Each ITS consists of a mandatory unique name and an optional type, which could be either host or vcluster (default to vcluster, if not specified)
+# Each ITS consists of a mandatory unique name and an optional type, which could be either host, vcluster, or external (default to vcluster, if not specified)
 # The optional boolean flag install_clusteradm can be used to determine if OCM should be installed in the CP.
 ITSes: # all the CPs in this list will execute the its.yaml PCH
   # - name: its1 # name of the control plane
-  #   type: host # optional type of control plane host or vcluster (default to vcluster, if not specified)
+  #   type: host # optional type of control plane host, vcluster, or external (default to vcluster, if not specified)
```
pdettori, line 82: s/plane host/plane: host/

core-chart/values.yaml
```diff
@@ -75,14 +75,14 @@ transport_controller:
...
   #   install_clusteradm: true|false  # optional flag to enable/disable the installation of OCM in the control plane (default to true, if not specified)
   # - name: its2 # name of the control plane
-  #   type: host # optional type of control plane host or vcluster (default to vcluster, if not specified)
+  #   type: host # optional type of control plane host, vcluster, or external (default to vcluster, if not specified)
```
pdettori, line 85: s/plane host/plane: host/

docs/content/direct/core-chart.md
```diff
@@ -111,15 +111,27 @@ The third section of the `values.yaml` file allows one to create a list of Inven
 ```yaml
 ITSes: # all the CPs in this list will execute the its.yaml PCH
   - name: <its1>          # mandatory name of the control plane
-    type: <vcluster|host> # optional type of control plane host or vcluster (default to vcluster, if not specified)
+    type: <vcluster|host|external> # optional type of control plane host, vcluster, or external (default to vcluster, if not specified)
```
pdettori, line 114: s/plane host/plane: host/

docs/content/direct/core-chart.md
```diff
@@ -111,15 +111,27 @@ The third section of the `values.yaml` file allows one to create a list of Inven
...
-    type: <vcluster|host> # optional type of control plane host or vcluster (default to vcluster, if not specified)
+    type: <vcluster|host|external> # optional type of control plane host, vcluster, or external (default to vcluster, if not specified)
     install_clusteradm: true|false  # optional flag to enable/disable the installation of OCM in the control plane (default to true, if not specified)
+    bootstrapSecret: # this section is ignored unless type is "external"
+      name: <secret-name> # default: "<control-plane-name>-bootstrap"
+      namespace: <secret-namespace> # default: "default"
+      key: <key-name> # default: "kubeconfig-incluster"
   - name: <its2>          # mandatory name of the control plane
-    type: <vcluster|host> # optional type of control plane host or vcluster (default to vcluster, if not specified)
+    type: <vcluster|host|external> # optional type of control plane host, vcluster, or external (default to vcluster, if not specified)
```
pdettori, line 121: s/plane host/plane: host/

docs/content/direct/core-chart.md
```diff
@@ -111,15 +111,27 @@ The third section of the `values.yaml` file allows one to create a list of Inven
...
   ...
 ```
 
-where `name` must specify a name unique among all the control planes in that KubeFlex deployment, the optional `type` can be either vcluster (default) or host, see [here](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md) for more information, and the optional `install_clusteradm`can be either true (default) or false to enable or disable the installation of OCM in the control plane.
+where `name` must specify a name unique among all the control planes in that KubeFlex deployment, the optional `type` can be vcluster (default), host, or external, see [here](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md) for more information, and the optional `install_clusteradm`can be either true (default) or false to enable or disable the installation of OCM in the control plane.
+
+When the ITS `type` is `external`, the `bootstrapSecret` sub-section can be used to indicate the bootstrap secret used by KubeFlex to connect to the external cluster. Specifically, it can be sued to indicate any of the optional values for the name of the secret, the namespace containing the secret, and the name of the key containg the kubeconfig of the external cluster.
```
pdettori, line 132: s/sued/used/

pdettori:
This looks great @francostellari ! it may be helpful to have the "example usage" somewhere in the docs.

pdettori:
while testing, I have run into this issue:

```console
$ scripts/create-external-bootstrap-secret.sh --controlplane its1 --context kind-ext1 --address https://ext1-control-plane:6443 --verbose
bash: scripts/create-external-bootstrap-secret.sh: Permission denied
$ ls -al scripts/create-external-bootstrap-secret.sh
-rw-r--r--@ 1 paolo  staff  6206 Feb  3 10:48 scripts/create-external-bootstrap-secret.sh
```

Seems like you need to adjust permissions for the script.

pdettori:
I manually updated the script permissions to move forward, but then got into another hurdle:

```console
$ chmod u+x scripts/create-external-bootstrap-secret.sh
$ scripts/create-external-bootstrap-secret.sh --controlplane its1 --context kind-ext1 --address https://ext1-control-plane:6443 --verbose
Extracting context kind-ext1 from kubeconfig /Users/paolo/.kube/config...
Setting server internal address to https://ext1-control-plane:6443...
Creating namespace default...
Error from server (AlreadyExists): namespaces "default" already exists
```

scripts/create-external-bootstrap-secret.sh
```diff
@@ -0,0 +1,195 @@
...
+# Create secret
+###############################################################################
+create_ns=true
+for ns in $(kubectl get ns --no-headers -o name) ; do
+    if [[ "namespace/${arg_ns}" == "$ns" ]] ; then
+        create_ns=false
+        break
+    fi
+done
+if [[ $create_ns ]] ; then
```
pdettori, line 190: The condition inside [[ $create_ns ]] checks if the create_ns variable is non-empty. In Bash, any non-empty string is considered true, so this is always executed.

pdettori:
After manual fix for the bootstrap secret script, I have run into this issue:

```console
$ helm dependency build core-chart
Error: the lock file (Chart.lock) is out of sync with the dependencies file (Chart.yaml). Please update the dependencies
```

is there any guidance to resolve this?

scripts/create-external-bootstrap-secret.sh
francostellari: I fixed it

francostellari:
> ```
> helm upgrade --install ks-core core-chart \
>   --set-json='ITSes=[{"name":"its1","type":"external","install_clusteradm":false}]' \
>   --set-json='WDSes=[{"name":"wds1"}]'
> ```

I fixed the bug that I introduced when I made a change to the YAML.

francostellari:
> ```shell
> helm upgrade --install ks-core core-chart \
>   --set-json='ITSes=[{"name":"its1","type":"external","install_clusteradm":false}]' \
>   --set-json='WDSes=[{"name":"wds1"}]'
> ```

I have added a section to the doc to detail a usage example with kind and how to use the local chart.

docs/content/direct/core-chart.md
```diff
@@ -152,7 +152,14 @@ where `name` must specify a name unique among all the control planes in that Kub
 
 ## KubeStellar Core Chart usage
 
-A specific version of the KubeStellar core chart can be simply installed in an existing cluster using the following command:
+The local copy of the core chart can be installed in an existing cluster using the commands:
+
+```shell
+helm dependency update core-chart
+helm upgrade --install core-chart
```
pdettori, line 159: don't we need also `helm dependency build` ?

docs/content/direct/core-chart.md
```diff
@@ -193,6 +200,32 @@ After the initial installation is completed, there are two main ways to install
       --set-json='WDSes=[{name":"wds2"}]'
     ```
 
+The core chart also supports the use of external clusters as ITS Control Planes. A specific application is to connect to existing OCM clusters. As an exmaple, create a first local kind cluster with OCM installed in it:
+
```
pdettori, line 204: s/exmaple/example/

docs/content/direct/core-chart.md
```diff
@@ -193,6 +200,32 @@ After the initial installation is completed, there are two main ways to install
...
+Then, create a second kind cluster suitable for KubeStellar installation and create a bootstrap secret in the new cluster with the kubeconfig information of the `ext1` cluster:
+
+```shell
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v$KUBESTELLAR_VERSION/scripts/create-kind-cluster-with-SSL-passthrough.sh) --name kubeflex --port 9443
+
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v$KUBESTELLAR_VERSION/scripts/create-external-bootstrap-secret.sh) --controlplane its1 --context kind-ext1 --address https://ext1-control-plane:6443 --verbose
+```
+
+Note the above command creates a secret named `its1-bootstrap` will be created in the `default` namespace of the `kind-kubeflex` cluster. Finally, install the core chart using the `ext1` cluster as ITS:
```
pdettori, line 219: s/creates a secret named `its1-bootstrap` will be created in/creates a secret named `its1-bootstrap` in/

docs/content/direct/core-chart.md
```diff
@@ -24,9 +24,9 @@ to join Workload Execution Clusters (WECs) (_e.g._, clusteradm),
 and to interact with Control Planes (_e.g._, kubectl), _etc_.
 For such purpose, a full list of executable that may be required can be found [here](./pre-reqs.md).
 
-The setup of KubeStellar via the Core chart requires the existence of a KubeFlex hosting cluster.
+The setup of KubeStellar via the Core chart requires the existence of a Kubernetes hosting cluster.
```
MikeSpreitzer, line 27: I think the original is better. We have a role called "KubeFlex hosting cluster", and that is what is needed here.

docs/content/direct/core-chart.md
```diff
@@ -24,9 +24,9 @@ to join Workload Execution Clusters (WECs) (_e.g._, clusteradm),
 and to interact with Control Planes (_e.g._, kubectl), _etc_.
 For such purpose, a full list of executable that may be required can be found [here](./pre-reqs.md).
 
-The setup of KubeStellar via the Core chart requires the existence of a KubeFlex hosting cluster.
+The setup of KubeStellar via the Core chart requires the existence of a Kubernetes hosting cluster.
 
-This can be:
+While not a complete list of supported hosting clusters, we systematically test our releases using:
```
MikeSpreitzer, line 29: Being more explicit about the fact that this list is not exhaustive is good.
However, all that we test in CI is `kind`. I am not sure that we need this sentence to try to make assertions about testing.

docs/content/direct/core-chart.md
```diff
@@ -111,15 +111,27 @@ The third section of the `values.yaml` file allows one to create a list of Inven
...
   ...
 ```
 
-where `name` must specify a name unique among all the control planes in that KubeFlex deployment, the optional `type` can be either vcluster (default) or host, see [here](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md) for more information, and the optional `install_clusteradm`can be either true (default) or false to enable or disable the installation of OCM in the control plane.
+where `name` must specify a name unique among all the control planes in that KubeFlex deployment, the optional `type` can be vcluster (default), host, or external, see [here](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md) for more information, and the optional `install_clusteradm`can be either true (default) or false to enable or disable the installation of OCM in the control plane.
+
+When the ITS `type` is `external`, the `bootstrapSecret` sub-section can be used to indicate the bootstrap secret used by KubeFlex to connect to the external cluster. Specifically, it can be used to indicate any of the optional values for the name of the secret, the namespace containing the secret, and the name of the key containg the kubeconfig of the external cluster.
```
MikeSpreitzer, line 132: The parsing of the second sentence here is not clear. At first I read it like this: "... used to indicate (a) any of the optional values for the name of the secret, (b) the namespace containing the secret, and (c) the name of the key containg the kubeconfig of the external cluster". I would rewrite it a little. Such as:

> Specifically, it can be used to indicate any of the optional values for: the name of the secret, the namespace containing the secret, and the name of the key containg the kubeconfig of the external cluster.

docs/content/direct/core-chart.md
```diff
@@ -140,7 +152,14 @@ where `name` must specify a name unique among all the control planes in that Kub
...
 
-A specific version of the KubeStellar core chart can be simply installed in an existing cluster using the following command:
+The local copy of the core chart can be installed in an existing cluster using the commands:
+
+```shell
+helm dependency update core-chart
+helm upgrade --install core-chart
+```
+
+Alternatively, a specific version of the KubeStellar core chart can be simply installed in an existing cluster using the following command:
```
MikeSpreitzer, line 162: The introduction a second way to instantiate the chart breaks the subject of the sentence on new line 168:

> The above command will ...

MikeSpreitzer:
A preview of the updated web site would be nice.

core-chart/values.yaml
```diff
@@ -75,14 +75,14 @@ transport_controller:
 
 
 # List the Inventory and Transport Spaces (ITSes) to be created by the chart
-# Each ITS consists of a mandatory unique name and an optional type, which could be either host or vcluster (default to vcluster, if not specified)
+# Each ITS consists of a mandatory unique name and an optional type, which could be either host, vcluster, or external (default to vcluster, if not specified)
 # The optional boolean flag install_clusteradm can be used to determine if OCM should be installed in the CP.
 ITSes: # all the CPs in this list will execute the its.yaml PCH
   # - name: its1 # name of the control plane
-  #   type: host # optional type of control plane host or vcluster (default to vcluster, if not specified)
+  #   type: host # optional type of control plane: host, vcluster, or external (default to vcluster, if not specified)
```
MikeSpreitzer, line 82: Shouldn't `bootstrapSecret` be described here too?

docs/content/direct/core-chart.md
```diff
@@ -181,6 +200,32 @@ After the initial installation is completed, there are two main ways to install
       --set-json='WDSes=[{name":"wds2"}]'
     ```
 
+The core chart also supports the use of external clusters as ITS Control Planes. A specific application is to connect to existing OCM clusters. As an example, create a first local kind cluster with OCM installed in it:
```
MikeSpreitzer, line 203: To my taste, this material should come before the material about "After the initial installation is completed, there are two main ways to install additional control planes", because this material is germane to the initial installation.

MikeSpreitzer, line 203: At the KubeStellar layer of abstraction we speak of spaces rather than (KF) ControlPlanes. I would rewrite the initial the initial sentence to be something like the following (also switching from plurals to singular for clarity).

> The core chart also supports the use of a pre-existing cluster (or any space, really) as an ITS.

docs/content/direct/core-chart.md
```diff
@@ -181,6 +200,32 @@ After the initial installation is completed, there are two main ways to install
...
+
+Then, create a second kind cluster suitable for KubeStellar installation and create a bootstrap secret in the new cluster with the kubeconfig information of the `ext1` cluster:
+
+```shell
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v$KUBESTELLAR_VERSION/scripts/create-kind-cluster-with-SSL-passthrough.sh) --name kubeflex --port 9443
+
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v$KUBESTELLAR_VERSION/scripts/create-external-bootstrap-secret.sh) --controlplane its1 --context kind-ext1 --address https://ext1-control-plane:6443 --verbose
+```
+
+Note the above command creates a secret named `its1-bootstrap` in the `default` namespace of the `kind-kubeflex` cluster. Finally, install the core chart using the `ext1` cluster as ITS:
```
MikeSpreitzer, line 219: There are two commands, so I would say something like "Note that the last command above creates ...".

docs/content/direct/core-chart.md
```diff
@@ -181,6 +200,32 @@ After the initial installation is completed, there are two main ways to install
...
+
+clusteradm init
+```
+
+Then, create a second kind cluster suitable for KubeStellar installation and create a bootstrap secret in the new cluster with the kubeconfig information of the `ext1` cluster:
+
+```shell
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v$KUBESTELLAR_VERSION/scripts/create-kind-cluster-with-SSL-passthrough.sh) --name kubeflex --port 9443
+
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v$KUBESTELLAR_VERSION/scripts/create-external-bootstrap-secret.sh) --controlplane its1 --context kind-ext1 --address https://ext1-control-plane:6443 --verbose
```
MikeSpreitzer, line 216: That `--address https://ext1-control-plane:6443` part is black magic unless explained. I might add an explanation along the following lines.

The `--address` URL needs to be one that the KubeFlex controller can use to open a connection to the external cluster's Kubernetes apiserver(s). In this example, the external cluster is a `kind` cluster with one kube-apiserver and it listens on port 6443. We rely on the DNS resolver in Docker networking to map the domain name `ext1-control-plane` to the address of the container of that same name on the relevant Docker network.

scripts/create-external-bootstrap-secret.sh
```diff
@@ -0,0 +1,195 @@
...
+COLOR_STATUS_TRUE="${COLOR_GREEN}"
+COLOR_STATUS_FALSE="${COLOR_RED}"
+COLOR_INFO="${COLOR_BLUE}"
+COLOR_TITLE="${COLOR_YELLOW}"
+COLOR_YAML="${COLOR_PURPLE}"
+
+
+# Command line arguments
+arg_cp=""
+arg_kubeconfig="$HOME/.kube/config"
```
MikeSpreitzer, line 44: Shouldn't this be the following?
```suggestion
arg_kubeconfig="${KUBECONFIG:-$HOME/.kube/config}"
```

scripts/create-external-bootstrap-secret.sh
```diff
@@ -0,0 +1,195 @@
...
+    shift
+done
+
+
+###############################################################################
+# Check arguments
+###############################################################################
+if [[ -z "$arg_cp" ]] ; then
+    echoerr "a control plane name is required!"
+    $0 --help
```
MikeSpreitzer, line 136: In curl-to-bash style usage, `$0` will be `bash` rather than a pathname of this script, right?

scripts/create-external-bootstrap-secret.sh
```diff
@@ -0,0 +1,195 @@
...
+    case ${#contexts[@]} in
+    (0)
+        echoerr "there are no contexts in the kubeconfig file!"
+        $0 --help
+        exit 1;;
+    (1)
+        arg_context="${contexts[0]}";;
+    (*)
+        for context in "${contexts[@]}" ; do # for all contexts
+            if [[ "$context" =~ "$arg_cp" ]] ; then
```
MikeSpreitzer, line 150: This is substring matching. It is treacherous, in a variety of ways.

scripts/create-external-bootstrap-secret.sh
```diff
@@ -0,0 +1,195 @@
...
+###############################################################################
+# Check arguments
+###############################################################################
+if [[ -z "$arg_cp" ]] ; then
+    echoerr "a control plane name is required!"
+    $0 --help
+    exit 1
+fi
+if [[ -z "$arg_context" ]] ; then
+    contexts=($(kubectl --kubeconfig "$arg_kubeconfig" config get-contexts --no-headers -o name 2> /dev/null))
```
MikeSpreitzer, line 140: I appreciate the work to be convenient, but:
(1) it is undocumented, right? Nobody will rely on it if it is undocumented;
(2) it is unusual --- the usual default is the current context.

I think that it would be fine to default to the current context.

scripts/create-external-bootstrap-secret.sh
```diff
@@ -0,0 +1,195 @@
...
+        if (( $# > 1 ));
+        then { arg_cp="$2"; shift; }
+        else { echo "$0: missing control plane name" >&2; exit 1; }
+        fi;;
+    (--kubeconfig|-K)
+        if (( $# > 1 ));
+        then { arg_kubeconfig="$2"; shift; }
+        else { echo "$0: missing kubeconfig filename" >&2; exit 1; }
+        fi;;
+    (--context|-C)
```
MikeSpreitzer, line 108: This script is working on two contexts, right? The one that it reads a kubeconfig from, and one that it writes a Secret into. Should both be parameters?

scripts/create-external-bootstrap-secret.sh
```diff
@@ -0,0 +1,195 @@
...
+    kubectl --kubeconfig=$BOOTSTRAP_KUBECONFIG config set-cluster $(kubectl --kubeconfig=$BOOTSTRAP_KUBECONFIG config current-context) --server=$arg_addr > /dev/null
+fi
+
+
+###############################################################################
+# Create secret
+###############################################################################
+create_ns=true
+for ns in $(kubectl get ns --no-headers -o name) ; do
+    if [[ "namespace/${arg_ns}" == "$ns" ]] ; then
```
MikeSpreitzer, line 185: This seems like an overly complex way to test whether the namespace exists. Why not just `kubectl get ns "$arg_ns" &> /dev/null` ?

scripts/create-external-bootstrap-secret.sh
```diff
@@ -0,0 +1,195 @@
...
+        create_ns=false
+        break
+    fi
+done
+if [[ $create_ns == true ]] ; then
```
MikeSpreitzer, line 190: The `create_ns` variable is read just once. Why not simplify to the following?

```shell
if ! kubectl get ns "$arg_ns" &> /dev/null; then
```

MikeSpreitzer (commented):
I have finished a review. I left independent comments.

docs/content/direct/core-chart.md
francostellari: ok, changed it back

docs/content/direct/core-chart.md
francostellari: I changed the sentence trying to satisfy both your comment and Paolo's

docs/content/direct/core-chart.md
francostellari: I made it absolute to the release version

docs/content/direct/core-chart.md
```diff
@@ -111,15 +111,27 @@ The third section of the `values.yaml` file allows one to create a list of Inven
...
+
+When the ITS `type` is `external`, the `bootstrapSecret` sub-section can be used to indicate the bootstrap secret used by KubeFlex to connect to the external cluster. Specifically, it can be used to indicate any of the optional values for the name of the secret, the namespace containing the secret, and the name of the key containg the kubeconfig of the external cluster.
+
+If the secret was created using the [create-external-bootstrap-secret.sh](../../../scripts/create-external-bootstrap-secret.sh) script and the value passed to the argument `--controlplane` matches the name of the Control Plane specified by the Helm chart, then the sub-section `bootstrapSecret` is not required because all default values will identify the bootstrap secret created by the script. More specifically, if an external kind cluster was created with the command `kind create cluster --name its1` and the `create-external-bootstrap-secret.sh --controlplane its1 --verbose` command was used to create the bootstrap secret, then it would be enough to inform the Helm chart with `--set-json='ITSes=[{"name":"its1","type":"external"}]'`.
```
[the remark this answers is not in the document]
francostellari: made it absolute to the release

docs/content/direct/core-chart.md
francostellari: I changed it to `Either if of the previous way of installing KubeStellar chart will install`

core-chart/values.yaml
francostellari: I added some comments here

docs/content/direct/core-chart.md
francostellari: Good idea, I switched the chunks

francostellari: ok, changed

scripts/create-external-bootstrap-secret.sh
francostellari: yes, thank you

scripts/create-external-bootstrap-secret.sh
francostellari: Perhaps but we have used this way in all other scripts, I believe.
I'm not sure if there is a solution besides hardcoding it.
I think it's a minor gripe given the small amount of `cur|bash --help`.
People will care about the list of commands.
But I'm open to a better solution if available.

scripts/create-external-bootstrap-secret.sh
francostellari: @MikeSpreitzer you need to explain more, I don't understand the issue.

docs/content/direct/core-chart.md
francostellari: It seems to me that the first way your read it was the correct way... perhaps I can add `any combination of the optional values...`  (a), (b), (c)

scripts/create-external-bootstrap-secret.sh
francostellari: I actually designed this way to save of arguments. One can specify the CP alone. Defaulting to the current context is not a good idea in this case because it will be the context of KubeFlex core-hub where you are creating the secret.

What about renaming the argument `--context` to `--adopted-context` to match kflex cli?

To clarify the `--kubeconfig` is also referring to where the adopted context is located. Perhaps that also need to change to be renamed `--adopted-kubeconfig`?

Bottom time currently `--kubeconfig` and `--context` are used to extract/ the secret data.... but not to create the secret object which is using the host kubeconfig/context... currently.

scripts/create-external-bootstrap-secret.sh
francostellari: I guess so, see above comment.

scripts/create-external-bootstrap-secret.sh
francostellari: yes, sorry for the poor coding

scripts/create-external-bootstrap-secret.sh
francostellari: yes, see above

scripts/create-external-bootstrap-secret.sh
MikeSpreitzer: Maybe something like "source-context" and "source-kubeconfig"? Or "bootstrap-context" and "bootstrap-kubeconfig"? The latter is not great, because they are not where the "bootstrap Secret" go.

scripts/create-external-bootstrap-secret.sh
MikeSpreitzer: Because we document using this script in curl-to-bash style, I think it should function properly when used in that style.

Since this script's handling of `--help` is simply to invoke `display_help`, that could be in-lined here.

docs/content/direct/core-chart.md
MikeSpreitzer: The problem with my first reading is that it confines "any of the optional values" to (a) but I think you meant "any of the optional values" to apply to the list of (a), (b), and (c).

docs/content/direct/core-chart.md
```diff
@@ -111,15 +115,27 @@ The third section of the `values.yaml` file allows one to create a list of Inven
...
+
+If the secret was created using the [create-external-bootstrap-secret.sh](https://github.com/kubestellar/kubestellar/tree/v$KUBESTELLAR_VERSION/scripts/create-external-bootstrap-secret.sh) script and the value passed to the argument `--controlplane` matches the name of the Control Plane specified by the Helm chart, then the sub-section `bootstrapSecret` is not required because all default values will identify the bootstrap secret created by the script. More specifically, if an external kind cluster was created with the command `kind create cluster --name its1` and the `create-external-bootstrap-secret.sh --controlplane its1 --verbose` command was used to create the bootstrap secret, then it would be enough to inform the Helm chart with `--set-json='ITSes=[{"name":"its1","type":"external"}]'`.
```
MikeSpreitzer, line 138: When rendered on the website, that URL will still contain `$KUBESTELLAR_VERSION` (please try it!).
See https://docs.kubestellar.io/release-0.26.0/contribution-guidelines/operations/document-management/#global-variables and look for examples in other source documents.

core-chart/values.yaml
MikeSpreitzer: You do not seem to have pushed those changes yet.

docs/content/direct/core-chart.md
```diff
@@ -163,6 +186,36 @@ helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart
...
+
+```shell
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v$KUBESTELLAR_VERSION/scripts/create-kind-cluster-with-SSL-passthrough.sh) --name kubeflex --port 9443
+
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v$KUBESTELLAR_VERSION/scripts/create-external-bootstrap-secret.sh) --controlplane its1 --context kind-ext1 --address https://ext1-control-plane:6443 --verbose
+```
+
+Note that the last command above creates a secret named `its1-bootstrap` in the `default` namespace of the `kind-kubeflex` cluster.
+
+The `--address` URL needs to be one that the KubeFlex controller can use to open a connection to the external cluster's Kubernetes apiserver(s). In this example, the external cluster is a kind cluster with one kube-apiserver and it listens on port 6443. This example relies on the DNS resolver in Docker networking to map the domain name `ext1-control-plane` to the Docker network address of the container of that same name.
```
MikeSpreitzer, line 207: Oops, I forgot one critical detail. That kube-apiserver listens on port 6443 _in its node's network namespace_.

scripts/create-external-bootstrap-secret.sh
MikeSpreitzer: Suppose I have contexts named "home" and "homestuck". A substring search for "home" will also match "homestuck".

A little more specifically, it is not uncommon to name something "foo" and later realize that you want another similar thing and name it "foo1" or "foo2".

docs/content/direct/core-chart.md
francostellari: I tried to change it.

docs/content/direct/core-chart.md
francostellari: I changed the text

core-chart/values.yaml
francostellari: it should be in the latest version

scripts/create-external-bootstrap-secret.sh
francostellari: ok, I misunderstood the first comment, I changed it now in all places

scripts/create-external-bootstrap-secret.sh
francostellari: I renamed the current `--kubeconfig` ==> `--source-kubeconfig` and `--context` ==> `--source-context`

francostellari: I also reworked `--kuneconfig` and `--context` to be the cluster where the secret is creted

docs/content/direct/core-chart.md
```diff
@@ -111,15 +115,27 @@ The third section of the `values.yaml` file allows one to create a list of Inven
...
+where `name` must specify a name unique among all the control planes in that KubeFlex deployment, the optional `type` can be vcluster (default), host, or external, see [here](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md) for more information, and the optional `install_clusteradm`can be either true (default) or false to enable or disable the installation of OCM in the control plane.
+
+When the ITS `type` is `external`, the `bootstrapSecret` sub-section can be used to indicate the bootstrap secret used by KubeFlex to connect to the external cluster. Specifically, it can be used to change any combination of (a) the name of the secret, (b) the namespace containing the secret, and (c) the name of the key containg the kubeconfig of the external cluster from their default value.
```
MikeSpreitzer, line 136: I would say "specify" instead of "change", because change implies that something already exists but we are talking here about specifying the Secret and key (not changing the Secret or key).

core-chart/values.yaml
MikeSpreitzer: I still do not see it. Here is what I see (showing commit ID):
image

scripts/create-external-bootstrap-secret.sh
MikeSpreitzer: Yes, I see that you changed it in some places. However, there are still uses of `$0` to refer to the script name. Go ahead and do a text search for "$0". It appears in error messages, which should not pin the blame on bash.

scripts/create-external-bootstrap-secret.sh
```diff
@@ -0,0 +1,212 @@
...
+
+
+# Display help
+display_help() {
+  cat << EOF
+Usage: $0 [options]
+
+--controlplane|-c <name>       control plane name used to name the secret: <name>-bootstrap
+--namespace|-n <name>          namespace name where to create the secret, default is "default"
+--kubeconfig|-K <filename>     use the specified kubeconfig used to generate the bootstrat secret
```
MikeSpreitzer, line 60: "generate" is a bit odd here, it calls to mind computation/processing. I think "store" is more to the point.

scripts/create-external-bootstrap-secret.sh
MikeSpreitzer: "big" is a substring of "bigger"; "near" is a substring of "nearer"; and so on.
You have seen people use names like "big", "bigger", and "biggest".

docs/content/direct/core-chart.md
```diff
@@ -111,15 +115,27 @@ The third section of the `values.yaml` file allows one to create a list of Inven
 ```yaml
 ITSes: # all the CPs in this list will execute the its.yaml PCH
   - name: <its1>          # mandatory name of the control plane
-    type: <vcluster|host> # optional type of control plane host or vcluster (default to vcluster, if not specified)
+    type: <vcluster|host|external> # optional type of control plane: host, vcluster, or external (default to vcluster, if not specified)
     install_clusteradm: true|false  # optional flag to enable/disable the installation of OCM in the control plane (default to true, if not specified)
+    bootstrapSecret: # this section is ignored unless type is "external"
+      name: <secret-name> # default: "<control-plane-name>-bootstrap"
+      namespace: <secret-namespace> # default: "default"
```
MikeSpreitzer, line 122: Wouldn't it make more sense for the default to be the namespace that the chart is being instantiated in?

docs/content/direct/core-chart.md
```diff
@@ -140,13 +156,20 @@ where `name` must specify a name unique among all the control planes in that Kub
 
 ## KubeStellar Core Chart usage
 
-A specific version of the KubeStellar core chart can be simply installed in an existing cluster using the following command:
+The local copy of the core chart can be installed in an existing cluster using the commands:
+
+```shell
+helm dependency update core-chart
```
MikeSpreitzer, line 162: When testing, I got complaint, and it was not clear to me whether this is a fatal problem or not. I checked the exit code, and it claimed success.

```console
mspreitz@mjs13 kubestellar % helm dependency update core-chart
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "minio" chart repository
...Successfully got an update from the "spiffe" chart repository
...Successfully got an update from the "kubestellar" chart repository
...Successfully got an update from the "prometheus-community" chart repository
...Successfully got an update from the "grafana" chart repository
...Successfully got an update from the "bitnami" chart repository
Update Complete. ⎈Happy Helming!⎈
Saving 1 charts
Downloading kubeflex-operator from repo oci://ghcr.io/kubestellar/kubeflex/chart
Pulled: ghcr.io/kubestellar/kubeflex/chart/kubeflex-operator:v0.8.0
Digest: sha256:6e889586cbc08d580c100139fce1cf3800f8aed3ce44bec5bee652941d7f5598
Deleting outdated charts
Could not verify core-chart/charts/.DS_Store for deletion: file 'core-chart/charts/.DS_Store' does not appear to be a gzipped archive; got 'application/octet-stream' (Skipping)

mspreitz@mjs13 kubestellar % echo $?
0
```

docs/content/direct/core-chart.md
```diff
@@ -163,6 +186,36 @@ helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart
...
+
+clusteradm init
+```
+
+Then, create a second kind cluster suitable for KubeStellar installation and create a bootstrap secret in the new cluster with the kubeconfig information of the `ext1` cluster:
+
+```shell
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v$KUBESTELLAR_VERSION/scripts/create-kind-cluster-with-SSL-passthrough.sh) --name kubeflex --port 9443
+
+bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v$KUBESTELLAR_VERSION/scripts/create-external-bootstrap-secret.sh) --controlplane its1 --context kind-ext1 --address https://ext1-control-plane:6443 --verbose
```
MikeSpreitzer, line 202: The flag names here need to be updated, right?

docs/content/direct/core-chart.md
```diff
@@ -140,13 +156,20 @@ where `name` must specify a name unique among all the control planes in that Kub
 
 ## KubeStellar Core Chart usage
 
-A specific version of the KubeStellar core chart can be simply installed in an existing cluster using the following command:
+The local copy of the core chart can be installed in an existing cluster using the commands:
+
+```shell
+helm dependency update core-chart
+helm upgrade --install core-chart
```
MikeSpreitzer, line 163: Something is missing on this command line.

```console
mspreitz@mjs13 kubestellar % helm upgrade --install core-chart
Error: "helm upgrade" requires 2 arguments

Usage:  helm upgrade [RELEASE] [CHART] [flags]
```

MikeSpreitzer (commented):
I tested, got as far as instantiating the chart, with only reported issues.

core-chart/values.yaml
francostellari: Sorry, I misread your initial comment and kept thinking to the `core-chart.md`. I updated the `value.yaml` file this time

scripts/create-external-bootstrap-secret.sh
francostellari: The strict was changed to detect the situation when more than context was being matched, inform the use that the match is not unique, and instruct the suer to explicitly use the `--source-context` argument. Please see the new logic at ~L171:

```bash
                if [[ -n "$arg_source_context" ]] ; then
                    echoerr "there are multiple matching contexts in the source kubeconfig file, specify one with \"--source-context\"!"
                    display_help
                    exit 3
                fi
```

I think this is a good compromise between convenience and clarity

docs/content/direct/core-chart.md
francostellari: ok

docs/content/direct/core-chart.md
francostellari: No for 2 reasons:
1. we do not instruct to install the `core-helm` chart to anything but default
2. this "default" value needs to match the "default" value of the script more than where the chart is installed
I recommend leaving it as it is.

docs/content/direct/core-chart.md
francostellari: I cannot reproduce your error and I have never seen it before.
It seems to have pulled the latest kubeflex, so I assume it worked.
@pdettori also mentioned that he he needed to do a dependency build to get it to work.
It may be a Helm version issue or a Mac issue.

docs/content/direct/core-chart.md
francostellari: Yes, I forgot the name

scripts/create-external-bootstrap-secret.sh
MikeSpreitzer: I think that we agreed offline to handle this pervasive issue in a separate PR.

scripts/create-external-bootstrap-secret.sh
MikeSpreitzer: OK.

scripts/create-external-bootstrap-secret.sh
MikeSpreitzer: +1

docs/content/direct/core-chart.md
MikeSpreitzer: 1. Neither do we forbid instantiating the core Helm chart in another namespace.
2. In my opinion it is better to follow the usual Kubernetes conventions. Two come to mind. One is that for a command, the default namespace to use is the one set in the current context (remember that a context can have a namespace set), if any, otherwise "default". The other is that references from one object to another normally stay in the same namespace. These two play nice together. For example, when using `helm install` without specifying a namespace, the chart will be installed in the namespace determined according to the first rule, and a reference from that chart to a bootstrap secret will reference a Secret in that same namespace when the user does not explicitly specify a namespace.

docs/content/direct/core-chart.md
MikeSpreitzer: Looks like a Mac issue --- `.DS_Store` is a special metadata thing that Darwin adds to a directory when you view it in the GUI.

This complaint was not introduced by this PR.

MikeSpreitzer (commented):
I have completed another round of review.

docs/content/direct/core-chart.md
MikeSpreitzer: FYI:

```console
mspreitz@mjs13 ~ % kubectl config view -o json | jq '.["current-context"] as $cc | .contexts[] | select(.name == $cc) | .namespace // "default"'
"default"
```

scripts/create-external-bootstrap-secret.sh
francostellari: I added a note in the display_help() function to document the behavior.

scripts/kubestellar-snapshot.sh
MikeSpreitzer, line 1: I see a lot of duplication of code in this script. There ought to be a way to DRY it out. Not a blocker.

MikeSpreitzer:
Testing shows that the test at https://github.com/kubestellar/kubestellar/blob/1ca512d3d99b0fcb3d9eec89b336e5392f8b078d/core-chart/templates/NOTES.txt#L15 also needs to be updated to also exclude "external" type ControlPlanes.

MikeSpreitzer (approved):
This is good enough to commit as progress. There are a few things to follow up on in following PR(s).

/lgtm
/approve
Thanks!

Issue #2716, opened 2025-01-10
bug: create-kubestellar-demo-env.sh fails with kind "Klusterlet" not known


MikeSpreitzer:
### Describe the bug

While investigating #2579, #2700, #2701, #2702, and #2715, I did some more testing. In this case I was testing release 0.25.0-rc.3 on an x86 64 bit VM (IBM Cloud, Classic not VPC), 16 virtual CPUs, 32 GB of memory. Ubuntu 22.04.5; Linux 5.15.0-130-generic. Again, the loop used is the following.

```shell
while time bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/tags/v0.25.0-rc.3/scripts/create-kubestellar-demo-env.sh); do
    date
    sleep 1
done; date
```

Following is a tail of the log.

```console
Registering cluster 1 and 2 for remote access with KubeStellar Core...
Preflight check: HubKubeconfig check Passed with 0 warnings and 0 errors
Preflight check: DeployMode Check Passed with 0 warnings and 0 errors
Preflight check: ClusterName Check Passed with 0 warnings and 0 errors
Please log onto the hub cluster and run the following command:

    clusteradm accept --clusters cluster1

This is not needed when the ManagedClusterAutoApproval feature is enabled
Preflight check: HubKubeconfig check Passed with 0 warnings and 0 errors
Preflight check: DeployMode Check Passed with 0 warnings and 0 errors
Preflight check: ClusterName Check Passed with 0 warnings and 0 errors
Error: resource mapping not found for name: "klusterlet" namespace: "" from "local": no ma
tches for kind "Klusterlet" in version "operator.open-cluster-management.io/v1"
ensure CRDs are installed first

real    3m1.126s
user    0m14.357s
sys     0m11.174s
Fri Jan 10 15:52:10 UTC 2025
```

I have attached the whole log.

[fail7.log](https://github.com/user-attachments/files/18383398/fail7.log)

### Output from KubeStellar-Snapshot.sh

```console
mspreitz@mjs-dev7a:~$ bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/heads/main/scripts/kubestellar-snapshot.sh) -V -Y -L
KubeStellar Snapshot v0.2.0{COLOR_NONE}

Script run on 2025-01-10_18:43:11
Checking script dependencies:
✔ kubectl version v1.29.10 at /usr/local/bin/kubectl
[...]

PR #2769, opened 2025-01-30
🌱  Debuggable and terse clusteradm join

MikeSpreitzer:
Make the `clusteradm join` output terse when successful, otherwise include debug logging,
in the demo env creation script.

## Summary
This PR improves the demo env creation script regarding the `clusteradm get token` and `clusteradm join` commands, used to add a WEC to the system. Regarding the `get token`, this PR changes the script to do that once rather than once per WEC. Regarding `join`, this PR changes the script to have more terse output in the success case and show debug logging in the failure case (which I hope will shed some light on #2716).

## Related issue(s)

Fixes #


MikeSpreitzer:
FYI, here is the loop that I am using to test this now:

```shell
while time bash <(curl -s https://raw.githubusercontent.com/MikeSpreitzer/kcp-edge-mc/refs/heads/better-join/scripts/create-kubestellar-demo-env.sh); do
    date
    sleep 1
done; date; bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/heads/main/scripts/kubestellar-snapshot.sh) -V -Y -L
```

PR #2774, opened 2025-01-31
🐛  Pin the version of ingress-nginx

### LINKED ISSUE #2772: Dead reckoning patch against main
https://github.com/kubestellar/kubestellar/blob/v0.26.0-alpha.5/scripts/create-kind-cluster-with-SSL-passthrough.sh#L146 references the `main` branch and https://github.com/kubestellar/kubestellar/blob/v0.26.0-alpha.5/scripts/create-kind-cluster-with-SSL-passthrough.sh#L146 patches in a fixed set of command line arguments. Since the contents of that `main` branch can change at any time without warning, this is dangerous.

MikeSpreitzer:
## Summary
This PR pins the version of the nginx ingress controller used, so that the fixed patch to its arguments is safe.

## Related issue(s)

Fixes #2772

### DIFF
--- a/scripts/create-kind-cluster-with-SSL-passthrough.sh
+++ b/scripts/create-kind-cluster-with-SSL-passthrough.sh
@@ -17,7 +17,7 @@
 
 set -o errexit
 
-NGINX_INGRESS_URL="https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml"
+NGINX_INGRESS_URL="https://raw.githubusercontent.com/kubernetes/ingress-nginx/refs/tags/controller-v1.12.0/deploy/static/provider/kind/deploy.yaml"
 name=kubestellar
 port=9443
 wait=true


scripts/create-kind-cluster-with-SSL-passthrough.sh
```diff
@@ -17,7 +17,7 @@
 
 set -o errexit
 
-NGINX_INGRESS_URL="https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml"
+NGINX_INGRESS_URL="https://raw.githubusercontent.com/kubernetes/ingress-nginx/blob/controller-v1.12.0/deploy/static/provider/kind/deploy.yaml"
```
francostellari, line 20: https://raw.githubusercontent.com/kubernetes/ingress-nginx/release-1.12/deploy/static/provider/kind/deploy.yaml

MikeSpreitzer: Yes, I messed up the URL.
`release-1.12` is a branch. I chose to use `controller-v1.12.0` because that is a tag, which is supposed to have an immutable association with a particular commit.

MikeSpreitzer:
The force-push above fixes the URL.

PR #2784, opened 2025-02-06
✨  Add the kubectl-rbac-flatten utility

MikeSpreitzer:
## Summary
This PR adds a utility, kubectl-rbac-flatten, that reads all of a clusters RoleBinding and ClusterRoleBinding objects and the Roles and ClusterRoles that the reference and computes and outputs the outer product of subjects and rules implied by those objects.

## Related issue(s)

Fixes #


cmd/kubectl-rbac-flatten/README.md
```diff
@@ -0,0 +1,113 @@
...
+The tabular output does a maximal cross product, so that no tuple in
+the result has a list/slice/array-structured field. The tabular output
+has two tables, one for resource-based rules and one for non-resource
+URLs. The second one is omitted if `--api-goups` is given anything
+other than `*`.
+
+The fields in one row are separated by TABs. The empty string API
+group is output as an empty string (thus between two TABs).
+
+The resource-based table omits the column for the role, for brevity.
```
waltforme, line 87: Would this omitted column introduce irregularity and confusion to users? People may ask why one table has the column and the other table doesn't.

And the role names contain useful information to users, no?

cmd/kubectl-rbac-flatten/README.md
```diff
@@ -0,0 +1,113 @@
...
+Following are excerpts from example tabular output.
+
+```
+BINDING                                                         SUBJECT                                                        VERB               APIGROUP                        RESOURCE                                    OBJNAME
+/cluster-admin                                                  G:system:masters                                               *                  *                               *                                           *
+/ingress-nginx                                                  SA:ingress-nginx/ingress-nginx                                 list                                               configmaps                                  *
+/ingress-nginx                                                  SA:ingress-nginx/ingress-nginx                                 list                                               endpoints                                   *
+...
+
+BINDING                                       ROLE                                      SUBJECT                                          VERB   NRURL
```
waltforme, line 98: I think we can safely say ClusterRole for non-resource URLs.

cmd/kubectl-rbac-flatten/README.md
```diff
@@ -0,0 +1,113 @@
...
+[
+{"Binding":{"Namespace":"","Name":"cluster-admin"},"RoleName":"cluster-admin","Subject":{"kind":"Group","apiGroup":"rbac.authorization.k8s.io","name":"system:masters"},"Rule":{"verbs":["*"],"apiGroups":["*"],"resources":["*"]}}
+,
+{"Binding":{"Namespace":"","Name":"cluster-admin"},"RoleName":"cluster-admin","Subject":{"kind":"Group","apiGroup":"rbac.authorization.k8s.io","name":"system:masters"},"Rule":{"verbs":["*"],"nonResourceURLs":["*"]}}
+,
+...
+]
+```
+
+### Tabular Ouptput
```
waltforme, line 76: Typo here.

cmd/kubectl-rbac-flatten/README.md
MikeSpreitzer: The reason for the irregularity is documented right here.

The binding specifies the role, so omitting loses no information about the "why". Omitting the binding would leave the user guessing, but omitting the role only requires one easy `get` to resolve. The purpose of this table is to help users who already know something about the "what" to discover the "why".

cmd/kubectl-rbac-flatten/README.md
MikeSpreitzer: True, but I think that would be a confusing irregularity: to say "ClusterRole" in one table and not "Role or ClusterRole" in the other.

cmd/kubectl-rbac-flatten/README.md
MikeSpreitzer: I tried to strike a balance between brevity and convenience.

I used as a criterion: will it fit in one line in the widest window that I can fit on my screen, using the default font. That admits ROLE for the second table but not the first.

cmd/kubectl-rbac-flatten/README.md
MikeSpreitzer: Fixed.

MikeSpreitzer:
The force-push to b5674c62a fixes the typo.

cmd/kubectl-rbac-flatten/README.md
waltforme: I would put heavy weight on regularity for this multi objective optimization. But looks like the discussion is more about personal preferences, so I'll also respect your choice.

/lgtm

PR #2785, opened 2025-02-06
📖Resolved OAuth Token Authorization Failure

### LINKED ISSUE #2544: bug: Forbidden 403 on helm upgrade
### Describe the bug

While following the "Getting Started" instructions, @milanahrae got the failure shown in the following screen scrape.

image

### Steps To Reproduce

Try "Getting Started" on Windows using some (which?) Linux distro under WSL.

### Expected Behavior

Success.

### Want to contribute?

### Additional Context

_No response_

Shruti78:
Fixes #2544

### DIFF
--- a/docs/content/direct/knownissue-wsl-ghcr-helm.md
+++ b/docs/content/direct/knownissue-wsl-ghcr-helm.md
@@ -10,7 +10,7 @@ This error indicates that there was an issue fetching the OAuth token from GitHu
 
 ## Root Cause
 
-The issue is caused by missing or misconfigured credentials for `ghcr.io` in the Docker configuration. Docker fails to authenticate and retrieve the necessary OAuth token from GitHub's container registry.
+The issue is caused by missing or misconfigured credentials for `ghcr.io` in the Docker configuration. Docker fails to authenticate & retrieve the necessary OAuth token from GitHub's container registry.
 ## Workaround
 
 To bypass this issue, follow these steps:
--- a/docs/content/direct/knownissue-wsl-ghcr-helm.md
+++ b/docs/content/direct/knownissue-wsl-ghcr-helm.md
@@ -1,20 +1,62 @@
-# Authorization failure in WSL while fetching Helm chart from ghcr.io
+# Authorization failure in WSL/Ubuntu while fetching Helm chart from ghcr.io
 
 ## Description of the Issue
 
-When using WSL (Windows Subsystem for Linxu) and following the
-[Getting Started recipe](get-started.md) you might get a failure from
-the command to instantiate KubeStellar's core Helm chart. The error
-message is as follows.
+When working with WSL (Windows Subsystem for Linux)/Ubuntu and trying to instantiate KubeStellar's core Helm chart, you might encounter the following error:
 
-> Error: failed to authorize: failed to fetch oauth token: unexpected status from GET request to https://ghcr.io/token?scope=repository%3Akubestellar%2Fkubestellar%2Fcore-chart%3Apull&service=ghcr.io: 403 Fobidden
+> Error: failed to authorize: failed to fetch oauth token: unexpected status from GET request to https://ghcr.io/token?scope=repository%3Akubestellar%2Fkubestellar%2Fcore-chart%3Apull&service=ghcr.io: 403 Forbidden
 
-This is [Issue 2544](https://github.com/kubestellar/kubestellar/issues/2544).
+This error indicates that there was an issue fetching the OAuth token from GitHub's container registry (`ghcr.io`), which is preventing you from pulling the Helm chart.
 
 ## Root Cause
 
-Unknown.
-
+The issue is caused by missing or misconfigured credentials for `ghcr.io` in the Docker configuration. Docker fails to authenticate and retrieve the necessary OAuth token from GitHub's container registry.
 ## Workaround
 
-Run the Getting Started recipe as root user in Linux. For example, use `sudo su -` to get a shell as root.
+To bypass this issue, follow these steps:
+
+1. Run the Getting Started recipe as the root user in Linux. This can be done using the `sudo su -` command to get a shell as root.
+
+```bash
+sudo su -
+```
+
+This will allow you to execute the Helm chart instantiation command with the necessary privileges.
+
+## 2. Authenticate Docker with GitHub Container Registry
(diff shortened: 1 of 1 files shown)


docs/content/direct/knownissue-wsl-ghcr-helm.md
```diff
@@ -1,20 +1,62 @@
-# Authorization failure in WSL while fetching Helm chart from ghcr.io
+# Authorization failure in WSL/Ubuntu while fetching Helm chart from ghcr.io
```
MikeSpreitzer, line 1: The issue is not specific to WSL nor Ubuntu. I would remove "in WSL/Ubuntu".

docs/content/direct/knownissue-wsl-ghcr-helm.md
```diff
@@ -1,20 +1,62 @@
...
 
 ## Description of the Issue
 
-When using WSL (Windows Subsystem for Linxu) and following the
-[Getting Started recipe](get-started.md) you might get a failure from
-the command to instantiate KubeStellar's core Helm chart. The error
-message is as follows.
+When working with WSL (Windows Subsystem for Linux)/Ubuntu and trying to instantiate KubeStellar's core Helm chart, you might encounter the following error:
```
MikeSpreitzer, line 5: Because the issue is not specific to WSL nor Ubuntu, I would remove "working with WSL (Windows Subsystem for Linux)/Ubuntu and".

docs/content/direct/knownissue-wsl-ghcr-helm.md
```diff
@@ -1,20 +1,62 @@
...
 
-> Error: failed to authorize: failed to fetch oauth token: unexpected status from GET request to https://ghcr.io/token?scope=repository%3Akubestellar%2Fkubestellar%2Fcore-chart%3Apull&service=ghcr.io: 403 Fobidden
+> Error: failed to authorize: failed to fetch oauth token: unexpected status from GET request to https://ghcr.io/token?scope=repository%3Akubestellar%2Fkubestellar%2Fcore-chart%3Apull&service=ghcr.io: 403 Forbidden
 
-This is [Issue 2544](https://github.com/kubestellar/kubestellar/issues/2544).
+This error indicates that there was an issue fetching the OAuth token from GitHub's container registry (`ghcr.io`), which is preventing you from pulling the Helm chart.
```
MikeSpreitzer, line 9: I would add a remark about diagnosing this problem with the simpler command `helm show chart oci://ghcr.io/kubestellar/kubestellar/core-chart`.

docs/content/direct/knownissue-wsl-ghcr-helm.md
```diff
@@ -1,20 +1,62 @@
...
+
+```bash
+sudo su -
+```
+
+This will allow you to execute the Helm chart instantiation command with the necessary privileges.
+
+## 2. Authenticate Docker with GitHub Container Registry
+
+Docker requires authentication to access `ghcr.io` (GitHub's container registry). To authenticate, you need to log into the registry using your GitHub token.
```
MikeSpreitzer, line 28: Not exactly. The chart is public, no login is needed. Your problem was that your Docker Desktop installation was broken, due to not installing `pass`. The Docker Desktop installation instructions say that you should install `pass`. Without it, Docker Desktop just does not work.

docs/content/direct/knownissue-wsl-ghcr-helm.md
MikeSpreitzer, line 9: I would write "says" rather than "indicates", because the error message does not identify the cause of the problem, only a confusing symptom.

docs/content/direct/knownissue-wsl-ghcr-helm.md
```diff
@@ -1,20 +1,62 @@
...
+> Error: failed to authorize: failed to fetch oauth token: unexpected status from GET request to https://ghcr.io/token?scope=repository%3Akubestellar%2Fkubestellar%2Fcore-chart%3Apull&service=ghcr.io: 403 Forbidden
 
-This is [Issue 2544](https://github.com/kubestellar/kubestellar/issues/2544).
+This error indicates that there was an issue fetching the OAuth token from GitHub's container registry (`ghcr.io`), which is preventing you from pulling the Helm chart.
 
 ## Root Cause
 
-Unknown.
-
+The issue is caused by missing or misconfigured credentials for `ghcr.io` in the Docker configuration. Docker fails to authenticate & retrieve the necessary OAuth token from GitHub's container registry.
```
MikeSpreitzer, line 13: The cause is the following chain.

- The user has installed Docker Desktop on Linux but not installed and initialized the package named "pass". Installing and initializing that package is recommended in https://docs.docker.com/desktop/setup/install/linux/#general-system-requirements and explained in the linked document https://docs.docker.com/desktop/setup/sign-in/#credentials-management-for-linux-users .
- When fetching a public Helm chart from an OCI registry, it is not necessary to do a Docker login to that registry. However, sometimes the registry will require the client to first acquire a certain "token" to then use when fetching the chart.
- The `helm` commands will use the Docker config file `~/.docker/config.json` if it exists. If that file _does not exist_ then the `helm` commands can fetch public charts from OCI registries with no problem; `helm` acquires and uses the token if and when required with no problem.
- If the user's Docker config file `~/.docker/config.json` exists but Docker's credential management is broken due to `pass` not being installed and initialized then the `helm` commands will get an error when they try to acquire the needed token.

docs/content/direct/knownissue-wsl-ghcr-helm.md
MikeSpreitzer: Another diagnostic would be to test whether https://github.com/actions/actions-runner-controller/issues/3381#issuecomment-2619817888 solves the problem.

MikeSpreitzer:
FYI: You can see a barely adequate version of the logic involved, in a utility script at https://github.com/kubestellar/kubestellar/blob/v0.25.1/hack/inspect-image.py#L124-L125 . This never attempts to use Docker, it always does the bare minimum to fetch a public image. Try `hack/inspect-image.py ghcr.io/kubestellar/kubestellar/core-chart` to see it work.

MikeSpreitzer:
Thanks for making a start on this. I covered it in #2798 .

PR #2790, opened 2025-02-11
🐛  Wait for the right version of the ingress controller Pod

### LINKED ISSUE #2788: bug: demo env creation script assumes fast reaction to patch
### Describe the bug

Abhijeet Das reported a failure when testing `scripts/create-kubestellar-demo-env.sh` on Ubuntu under WSL. The following screenshot is his report.

Image

Note that `scripts/create-kind-cluster-with-SSL-passthrough.sh` patches the Deployment object for the controller and then waits for the Deployment's Pod object to exist and be "ready" --- without checking that the existing and ready Pod corresponds to the _post-patch_ version of the Deployment. In the reported log you can see that the pod that satisfies the wait in `scripts/create-kind-cluster-with-SSL-passthrough.sh` is from a different ReplicaSet than the pod that generates the later complaint.

not run

### Steps To Reproduce

Try the script. Get unlucky.

### Expected Behavior

No sign of failure

### Want to contribute?

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR updates `scripts/create-kind-cluster-with-SSL-passthrough.sh` so that it no longer assumes that the first controller Pod that it sees is the one corresponding to the patched version of the controller Deployment.

## Related issue(s)

Fixes #2788

### DIFF
--- a/scripts/create-kind-cluster-with-SSL-passthrough.sh
+++ b/scripts/create-kind-cluster-with-SSL-passthrough.sh
@@ -147,8 +147,19 @@ kubectl --context "kind-${name}" patch deployment ingress-nginx-controller -n in
 
 if [[ "$wait" == "true" ]] ; then
   echo "Waiting for nginx ingress with SSL passthrough to be ready..."
-  while [ -z "$(kubectl --context kind-${name} get pod --namespace ingress-nginx --selector=app.kubernetes.io/component=controller --no-headers -o name 2> /dev/null)" ] ;  do
+  while true; do
       sleep 5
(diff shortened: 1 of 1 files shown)


francostellari:
I tested it and it seems to work /lgtm
I'm not sure what this is trying to solve though.
Also the style of the `if..then` is inconsistent with the rest of the script.

MikeSpreitzer:
The problem that this is trying to solve is reported in #2788: before this PR, the create-kind-cluster-with-SSL-passthrough.sh would patch the Ingress controller Deployment and wait for _some_ Pod to exist and be ready but this could get confused by a Pod created from the pre-patch version of the Deployment.

PR #2795, opened 2025-02-12
✨  Check sysctl fs stuff in preflight

MikeSpreitzer:
## Summary
This PR updates the `{scripts,hack}/check_pre_req.sh` script to check that the container runtime's OS has been adjusted according to https://kind.sigs.k8s.io/docs/user/known-issues#pod-errors-due-to-too-many-open-files . This is good because we have been having trouble with that requirement not being met and people being confused by the resulting behavior.

This PR also updates the prereqs doc to state this requirement more prominently.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-check-kind-config/direct/pre-reqs/#additional-software-for-the-getting-started-setup and #additional-software-for-building-kubestellar-from-source-and-testing and #examples .

## Related issue(s)

Fixes #

### DIFF
--- a/hack/check_pre_req.sh
+++ b/hack/check_pre_req.sh
@@ -57,6 +57,24 @@ is_installed() {
     fi
 }
 
+check_sysctl() {
+    name="$1"
+    minval="$2"
+    # because the OS running this script might not be the Linux that runs the containers, use `docker run` to interrogate that Linux
+    if report=$(docker run --rm busybox sysctl "$name"); then
+        val=$(cut -d= -f2 <<<$report | awk '{ print $1 }')
+        if [[ $val -ge "$minval" ]]; then
+            echo -e "\033[0;32m\xE2\x9C\x94\033[0m $name is $val"
+        else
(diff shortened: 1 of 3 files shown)


hack/check_pre_req.sh
```diff
@@ -57,6 +57,24 @@ is_installed() {
     fi
 }
 
+check_sysctl() {
+    name="$1"
+    minval="$2"
+    # because the OS running this script might not be the Linux that runs the containers, use `docker run` to interrogate that Linux
```
waltforme, line 63: Trying to understand this: Is WSL one possible "Linux that runs the containers"? Any other possibilities?

MikeSpreitzer: I am not sure I understand the question. Can you restate or elaborate?

waltforme: I'm trying to understand the reason that we need docker instead of running sysctl directly without docker. Trying to find an example of "the OS running this script" and "Linux that runs the containers" are not the same.

waltforme: I see. That's also what I meant by mentioning Windows Subsystem for Linux in the opening comment.

/lgtm

MikeSpreitzer: There is more to WSL than the guest OS. When I hear "WSL", I think mainly of the Windows machinery that hosts the guest.

PR #2796, opened 2025-02-13
✨  Check helm health in preflight

MikeSpreitzer:
## Summary
This PR enhances the check_pre_req.sh script to check whether `helm show chart oci://ghcr.io/kubestellar/kubestellar/core-chart` succeeds. This is good because helm is unusable when that fails but nothing gives the user a clear clue about this.

## Related issue(s)

Fixes #

### DIFF
--- a/hack/check_pre_req.sh
+++ b/hack/check_pre_req.sh
@@ -67,11 +67,11 @@ check_sysctl() {
             echo -e "\033[0;32m\xE2\x9C\x94\033[0m $name is $val"
         else
             echo -e "\033[0;31mX\033[0m sysctl ${name} is only $val but must be at least $minval (see https://kind.sigs.k8s.io/docs/user/known-issues#pod-errors-due-to-too-many-open-files)" >&2
-            exit 4
+            [ "$assert" == "true" ] && exit 4
         fi
     else
         echo -e "\033[0;31mX\033[0m Failed to extract sysctl ${name}" >&2
-        exit 3
+        [ "$assert" == "true" ] &&  exit 3
     fi
 }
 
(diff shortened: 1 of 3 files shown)


hack/check_pre_req.sh
```diff
@@ -112,12 +112,22 @@ is_installed_go() {
...
 is_installed_helm() {
+    local log
     is_installed 'Helm' \
         'helm' \
         'helm version' \
         'helm version --template={{.Version}}' \
         'https://helm.sh/docs/intro/install/' \
         'v3'
+    if log=$(helm show chart oci://ghcr.io/kubestellar/kubestellar/core-chart 2>&1)
+    then echo -e "\033[0;32m\xE2\x9C\x94\033[0m helm can fetch public charts"
```
francostellari, line 123: I guess capitalizing Helm like in the line before would be stylistically better

hack/check_pre_req.sh
```diff
@@ -112,12 +112,22 @@ is_installed_go() {
...
+    else
+        echo "$log" >&2
+        echo >&2
+        echo -e "\033[0;31mX\033[0m helm can not fetch public charts!" >&2
```
francostellari, line 127: same "helm" -> "Helm"

hack/check_pre_req.sh
```diff
@@ -112,12 +112,22 @@ is_installed_go() {
...
+        exit 5
```
francostellari, line 128: For consistency, the script should exit only if the `--assert` flag is used... see line 54

```
 [ "$assert" == "true" ] && exit 5
```

francostellari:
While not part of this PR, the `exit 3` and  `exit 4` in `check_sysctl()` should also be prefixed by `[ "$assert" == "true" ] && exit ...`

hack/check_pre_req.sh
MikeSpreitzer: I chose "helm" because I am referring to the CLI command.

hack/check_pre_req.sh
MikeSpreitzer: I chose "helm" because I am referring to the CLI command.

hack/check_pre_req.sh
MikeSpreitzer: Oh, right. Fixed in latest revision.

PR #2815, opened 2025-02-25
🐛 Fix a problem with `echov` in `kubestellar-snapshot` script

francostellari:
## Summary

Fix a problem with `echov` being invoked by the `on_exit` function before being defined

Here is an example of the problem being solved:

```shell
$ kubestellar-snapshot.sh --foo
kubestellar-snapshot.sh: unknown flag
kubestellar-snapshot.sh: line 182: echov: command not found
```

## Related issue(s)

Fixes #

### DIFF
--- a/scripts/kubestellar-snapshot.sh
+++ b/scripts/kubestellar-snapshot.sh
@@ -79,10 +79,15 @@ EOF
 }
 
 
+# Define default behavior
+echov() { :; }
+
+
 # Indent JSON
 indent() {
     sed 's/^/  /'
 }
+
 
 # Echo in color
 echocolor() {
@@ -239,12 +244,8 @@ done
 ###############################################################################
 # Alias definitions
 ###############################################################################
-# Define the echov function based on verbosity
-if [ "$arg_verbose" == "true" ]; then
-    echov() { echo "$@" ; }
-else
-    echov() { :; }
-fi
+# Redefine the echov function based on verbosity
+[ "$arg_verbose" == "true" ] && echov() { echo "$@" ; }
 
 
 ###############################################################################
--- a/scripts/kubestellar-snapshot.sh
+++ b/scripts/kubestellar-snapshot.sh
@@ -232,10 +232,10 @@ while (( $# > 0 )); do
         display_help
         exit 0;;
     (-*)
-        echo "$0: unknown flag" >&2
+        echo "$0: unknown flag \"$1\"" >&2
         exit 1;;
     (*)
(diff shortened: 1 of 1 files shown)


scripts/kubestellar-snapshot.sh
```diff
@@ -79,11 +79,16 @@ EOF
 }
 
 
+# Define default behavior
+echov() { :; }
```
MikeSpreitzer, line 83: Is this the best choice for the initial behavior? Wouldn't it be better to show the error message?

francostellari: Yes, it's the best choice. I'm trying to fix the error reported on line 182. The error about the wrong flag will stay.

MikeSpreitzer: > I'm trying to fix the error reported on line 182.

I suppose you mean that you are trying to make the command `echov ""` on old line 182 not crash. I was more interested in old line 191, `echov -e "Removing temporary folder: ${COLOR_INFO}$TMPFOLDER${COLOR_NONE}"`. It is a little more substantial. My thinking was, if something really basic has gone wrong, let's err on the side of giving more information.

> The error about the wrong flag will stay.

I am not sure what you are referring to. What error, what flag, wrong how?

More importantly, isn't there an easier way to fix this? Couldn't the `mktemp` and `trap` statements come after the flag processing and helper function definitions? The temporary directory is not used until after that.

francostellari: @MikeSpreitzer `echov` is controlled by the `-V` flag. If you do not request verbosity none of the echov command will print anything. The above command does not request verbosity so the message will not be printed in any case.

There is a bug in the script that it exists/crashes before the echov function is defined it display an error about echov is not defined. This is what I'm trying to fix.

If you desire to see the text on line 192 then you have to precede the crashing flag with `-V`: `kubestellar-snalpshot.sh -V --crash-me`

This PR does not change the behavior or output of the script in any way expect to remove the the error `kubestellar-snapshot.sh: line 182: echov: command not found`

MikeSpreitzer: > both the creation and trapping would have to be moved after echov is defined and after argument parsing

Yes, that is what I am suggesting.

> which seems to me a more confusing change.

I think that this will make for a simpler script, no need to define the same function twice.

MikeSpreitzer (commented):
This is functional. We can consider simplifying in this PR or follow-on.
/lgtm

PR #2817, opened 2025-02-27
🌱 Update year for copyright and license

waltforme:
## Summary
As shown in the title.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2023-2024 The KubeStellar Authors
+   Copyright 2023-2025 The KubeStellar Authors
 
    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


MikeSpreitzer (commented):
/lgtm
.. but I think it is unnecessary. Copyright is good for MANY years after it is established. There are copyright statements in essentially all the files. I really do not want to sign up for updating all of them every year. I understand that this PR updates just a few of the more prominent ones. The fact that it is few makes it tolerable even though unnecessary.

waltforme:
/approve

I agree all the points here.
The reason I'm doing this is '2024' may give people the impression that the project is unattended or lack of maintenance.

> /lgtm .. but I think it is unnecessary. Copyright is good for MANY years after it is established. There are copyright statements in essentially all the files. I really do not want to sign up for updating all of them every year. I understand that this PR updates just a few of the more prominent ones. The fact that it is few makes it tolerable even though unnecessary.

Issue #2809, opened 2025-02-20
feature: Add logic to the KubeStellar controller to map back the status to the objects READY values


francostellari:
### Feature Description

Currently a user needs to look at the status objects to figure out if a certain objects was actually propagated to the WECs etc.

I would be nice to add logic to bring back the states to the original object in the WDS.

Specifically, the values READ x/y should depict the actual summarized state.
For example, with `n`  replicas going to `m` clusters, `y` should be equal to `n*m`

### Proposed Solution

 I'm not sure

### Want to contribute?

- [ ] I would like to work on this issue.

MikeSpreitzer:
I am trying to understand the general request here. It sounds like a new feature to me: filling in the reported state of objects with an approximation of the reported state from all the WECs even when the number of them is not 1. If we attempt this then I think that the answer should work even for kinds of objects that _we_ do not know about.

I think that this request is motivated by ArgoCD's display of `Deployment` objects. I am trying to understand exactly what is the approximation function being requested for these. You can see the definition of the status section of a `Deployment` at https://github.com/kubernetes/api/blob/v0.29.14/apps/v1/types.go#L480-L517 . You will see that it has several integer-valued counter fields. Filling one in with the sum over the same field in all the WECs is going to be confusing when compared with the `spec.replicas`. I think that average (rounded down) would be better than sum. The rounding down works well when scaling up, but is overly optimistic when scaling down. Also, there is a set of "conditions", which have boolean values. I suppose the most obvious way to approximate is to combine with AND, on the grounds that these are usually defined in a positive way. Following is an example of the conditions for a `Deployment`.

```yaml
  - lastTransitionTime: "2025-02-24T23:14:46Z"
    lastUpdateTime: "2025-02-24T23:14:46Z"
    message: Deployment has minimum availability.
    reason: MinimumReplicasAvailable
    status: "True"
    type: Available
  - lastTransitionTime: "2025-02-24T23:14:42Z"
    lastUpdateTime: "2025-02-24T23:14:46Z"
    message: ReplicaSet "coredns-76f75df574" has successfully progressed.
    reason: NewReplicaSetAvailable
    status: "True"
    type: Progressing
```

There are also other fields, such as `observedGeneration`.

Pardhasardhiraob29:
Hi! I'm new to open source and would love to make my first contribution. If this issue is still open would love work upon it.
Thanks a lot!
/assign

Rupam-It:
[...]

Issue #2808, opened 2025-02-20
feature: Add logic in KubeStellar controller to add that `ownerReference` in the status object


francostellari:
### Feature Description

Currently, when using ArgoCD to deploy/visualize the workloads the WDS, the status objects are not displayed because there is no `ownerReference` associating them to any of the objects

![Image](https://github.com/user-attachments/assets/9d501fbe-8c63-45c1-b33d-8a3c29cfd21d)

### Proposed Solution

It would be useful to update the KubeStelalr controller to add the `ownerReference` label to the status objects and, perhaps, associate them to the binding policy,

### Want to contribute?

- [ ] I would like to work on this issue.

PR #2818, opened 2025-02-27
✨ Add KubeStellar core chart integration with Argo CD

francostellari:
## Summary

This PR aims to extend KubeStellar Core chart to enable a user friendly integration with a locally deployed Argo CD instance.

In particular the modified Helm chart enables the following:
- installation of a local instance of Argo CD: `--set "argo-cd.install=true"`
- create an ingress to Argo CD: `--set "argo-cd.global.domain=<url>"`, for example https://argocd.localtest.me:9443
- automatically register all WDSes as target clusters in Argo CD
- allow the creating of Argo CD applications directly from values or cli

Here is an example bringing up scenario-1 in a Kind cluster:

```shell
helm dependency update core-chart

  --set-json='ITSes=[{"name":"its1"}]' \
  --set-json='WDSes=[{"name":"wds1"}]' \
  --set "argo-cd.install=true" \
  --set-json='argo-cd.applications=[{ "name": "scenario-1", "repoURL": "https://github.com/francostellari/kubestellar", "targetRevision": "argocd", "path": "config/argocd/scenario-1", "destinationWDS": "wds1" }]'
```

## Related issue(s)

This PR would benefit from a resolution of these issues: #2808 and #2809

Fixes #


francostellari:
image
image
image

MikeSpreitzer:
Could the example be revised to use an application from a project rather than an individual?

core-chart/values.yaml
```diff
@@ -35,11 +35,32 @@ kubeflex-operator:
   # hostContainer: kubeflex-control-plane # used to set the name of the container that runs the KubeFlex hosting cluster
   # verbosity: 0 $ used to set the verbosity of KubeFlex controller (this is not impacted by the global VERBOSITY value)
 
+
+# Argo CD values
+argo-cd:
+  fullnameOverride: argocd # do not change
```
MikeSpreitzer, line 41: Why is it here if "do not change"?

docs/content/direct/core-chart-argocd..md
MikeSpreitzer, line 1: Did you intend two dots in this filename?

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,151 @@
+{{- if index .Values "argo-cd" "install" }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: argocd-kubestellar-operator
+data:
+  # k logs $(k get pods -l job-name=argocd-kubestellar-operator -o 'jsonpath={.items[0].metadata.name}')
+  # k delete -f /vagrant/add-wds-clusters-job.yaml ; k apply -f /vagrant/add-wds-clusters-job.yaml && sleep $WAIT_BEFORE_RETRY ; k logs $(k get pods -l job-name=argocd-kubestellar-operator -o 'jsonpath={.items[0].metadata.name}')
```
MikeSpreitzer, line 8: What is this line here for?

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,151 @@
...
+        if [ -z "$ARGOCD_POD" ]; then
+            >&2 echo "==> ERROR: Argo CD pod not found! ...retrying in $WAIT_BEFORE_RETRY seconds..."
+            sleep $WAIT_BEFORE_RETRY
+        fi
+    done
+    echo "==> Argo CD pod found: $ARGOCD_POD"
+
+    echo "Waiting for Argo CD pod to be ready..."
+    kubectl -n $NS wait --for=condition=Ready pod/$ARGOCD_POD
+    echo "==> Argo CD show condition Ready."
```
MikeSpreitzer, line 31: s/show/shows/

Is this necessary, since `kubectl wait` will log about this?

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,151 @@
...
+            sleep $WAIT_BEFORE_RETRY
+        done
+        echo "==> Control Plane \"$cp_name\" is ready."
+
+        # determine the secret name and namespace
+        key=$(kubectl get controlplane $cp_name -o=jsonpath='{.status.secretRef.inClusterKey}')
+        secret_name=$(kubectl get controlplane $cp_name -o=jsonpath='{.status.secretRef.name}')
+        secret_namespace=$(kubectl get controlplane $cp_name -o=jsonpath='{.status.secretRef.namespace}')
+
+        kubeconfig="/tmp/${cp_name}.kubeconfig"
```
MikeSpreitzer, line 69: This leaves credentials in /tmp, which is not the best choice. You could instead, for example, hold the kubeconfig in a shell variable.

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,151 @@
+{{- if index .Values "argo-cd" "install" }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: argocd-kubestellar-operator
+data:
+  # k logs $(k get pods -l job-name=argocd-kubestellar-operator -o 'jsonpath={.items[0].metadata.name}')
+  # k delete -f /vagrant/add-wds-clusters-job.yaml ; k apply -f /vagrant/add-wds-clusters-job.yaml && sleep $WAIT_BEFORE_RETRY ; k logs $(k get pods -l job-name=argocd-kubestellar-operator -o 'jsonpath={.items[0].metadata.name}')
+  argocd-kubestellar-operator.sh: |
+    #!/bin/env bash
```
MikeSpreitzer, line 10: Maybe add a comment saying what this script does and who/what uses it?

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,151 @@
...
+  annotations:
+    # This is what defines this resource as a hook. Without this line, the
+    # job is considered part of the release.
+    "helm.sh/hook": post-install, post-upgrade
+    "helm.sh/hook-weight": "5" # IMPORTANT: set this weight higher than KubeFlex post-install hook, otherwise postgresql will not install
+    "helm.sh/hook-delete-policy": before-hook-creation, hook-succeeded
+spec:
+  template:
+    metadata:
+      name: argocd-kubestellar-operator
```
MikeSpreitzer, line 123: Is it necessary and good to specify this? Doesn't the Job controller take care of it?

docs/content/direct/core-chart-argocd..md
```diff
@@ -0,0 +1,57 @@
+# Using Argo CD with KubeStellar Core chart
+
+This documents explains how to use KubeStellar Core chart to:
+
+- Deploy an instance of Argo CD using KubeStellar Core chart that automatically lists the WDSes as desitnation clusters
+- Automatically crete Argo CD applications, including scenarios examples
```
MikeSpreitzer, line 6: The first bullet is awkward grammatically. Writing about "chart that automatically ..." suggests there is another chart that does NOT "automatically ...". And the second bullet is also part of the functionality of the one and only chart. And the sentence introducing the bullets already says that this is something that the one and only chart does. So I would rewrite the bullets to something like the following.

- deploy ArgoCD in the KubeFlex hosting cluster;
- register every WDS as a cluster in ArgoCD; and
- create ArgoCD Applications as specified in one of the chart's values.

docs/content/direct/core-chart-argocd..md
```diff
@@ -0,0 +1,57 @@
+# Using Argo CD with KubeStellar Core chart
+
+This documents explains how to use KubeStellar Core chart to:
```
MikeSpreitzer, line 3: "... use _the_ KubeStellar core Helm chart ..."

docs/content/direct/core-chart-argocd..md
```diff
@@ -0,0 +1,57 @@
+# Using Argo CD with KubeStellar Core chart
+
+This documents explains how to use KubeStellar Core chart to:
+
+- Deploy an instance of Argo CD using KubeStellar Core chart that automatically lists the WDSes as desitnation clusters
+- Automatically crete Argo CD applications, including scenarios examples
+
+## Pre-requisites
+
+The prerequisites are the same of [installing KubeStellar using the Core chart](./core-chart.md#pre-requisites).
```
MikeSpreitzer, line 10: "same _as_"

I am not sure if the "./" at the start of the relative URL works when rendering for the website. I am sure that it works without the "./".

It would help to see a rendering of the website.

docs/content/direct/core-chart-argocd..md
MikeSpreitzer, line 1: This file should be linked into the doc tree. This includes:

- reference from the main page for the core Helm chart.
- reference from the navigation tree in `mkdocs.yml`.
- Decide how the addition of this page changes the shape of the doc tree and do not make Issue #2825 worse.

docs/content/direct/core-chart-argocd..md
```diff
@@ -0,0 +1,57 @@
...
+
+## Installing Argo CD using KubeStellar Core chart
+
+To enable the installation of Argo CD by the KubeStellar Core chart, use the flag `--set "argo-cd.install=true"`. Besides deploying an instance of Argo CD, KubeStellar Core chart will take care of registering all the WDSes installed by the chart as Argo CD target clusters.
+
+When deploying in an **OpenShift** cluster, add the flag `--set "argo-cd.openshift.enabled=true"`.
+
+When deploying in a **Kubernetes** cluster, use the flag `--set "argo-cd.global.domain=<url>"` to provide the URL for the **nginx** ingress, default `argocd.localtest.me`.
+
+Note that when creating a local **Kubernetes** cluster using our scripts for **Kind** or **k3s**, the **nginx** ingress will be accessible on port `9443`; therefore the Argo CD UI can be accessed at the address `https://argocd.localtest.me:9443`.
```
MikeSpreitzer, line 22: Maybe make it "_host_ port", because there is a different port number inside the cluster.

docs/content/direct/core-chart-argocd..md
```diff
@@ -0,0 +1,57 @@
...
+
+The initial password for the `admin` user can be retrieved using the command:
+
+```shell
+kubectl get secret argocd-initial-admin-secret -o jsonpath="{.data.password}" | base64 -d
+```
+
+## Deploying Argo CD applications
+
+KubeStellar Core chart can also be used to automatically deploy Argo CD applications. The example below show an example deploying an application to corresponding to `scenario 1` in [KubeStellar docs](./example-scenarios.md#scenario-1---multi-cluster-workload-deployment-with-kubectl).
```
MikeSpreitzer, line 32: "_The_ KubeStellar core chart ..."

MikeSpreitzer, line 32: "automatically" is superfluous and potentially confusing here.

MikeSpreitzer, line 32: "... an example _of_ deploying ..."

MikeSpreitzer, line 32: s/to corresponding to/corresponding to/

core-chart/values.yaml
```diff
@@ -35,11 +35,32 @@ kubeflex-operator:
...
+  install: false # enable/disaple the installation of Argo CD
+  # openshift:
+  #   enabled: true
+  global:
+    domain: argocd.localtest.me # When using the "create-kind-cluster-with-SSL-passthrough.sh" script, Argo UI will be accessible from the host at https://argocd.localtest.me:9443
+  applications: # list of Argo CD applications to be create
+  # Uncomment the lines below to deploy an application for the scenario-1 example in KubeStellar docs
+  # - name: scenario-1 # required, must be unique
+  #   project: default # default: default
```
MikeSpreitzer, line 50: A reader who only uses plain Kubernetes is wondering "what is `project` for?"

docs/content/direct/core-chart-argocd..md
```diff
@@ -0,0 +1,57 @@
...
+
+```yaml
```
MikeSpreitzer, line 34: The words before this display of YAML should explain that it is a display of the relevant fragment of the chart's values.

MikeSpreitzer (commented):
I have finished a review, left independent comments.

core-chart/templates/NOTES.txt
```diff
@@ -19,3 +19,11 @@ kflex ctx --overwrite-existing-context {{ $cp.name }}
 
 Finally, you can use `kflex ctx` to switch back to the kubeconfig
 context for your KubeFlex hosting cluster.
+
+{{- if (index .Values "argo-cd" "install") }}
+
+Access Argo CD UI at https://{{ index .Values "argo-cd" "global" "domain" }} (append :9443 for Kind installations).
```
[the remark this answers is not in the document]
francostellari: I added `k3s`.

core-chart/values.yaml
francostellari: It's here because I want this override to prevent prefixing all argocd resources with `{{.Released.Name}}` (e.s `ks-core-`) but I don't want the final suer to mess with it.
I don't know of a way to hide the override from the user.

docs/content/direct/core-chart-argocd..md
francostellari: Sorry, I missed it

core-chart/templates/argocd/operator.yaml
francostellari: I added a comment before the script

core-chart/templates/argocd/operator.yaml
francostellari: @MikeSpreitzer I understand your concern.
I can add steps to delete the temp files after use in both pods.
Also remember that the operator pod is terminated upon completion
Argo pod will "know" the kubeconfig in order to access the WDS even if deleting the file.

The issue here is that I need to gather a kubeconfig in the operator pod, copy it to the argo pod and execute a command there using the kubeconfig.

I did try piping as follow:

```shell
echo "==> Getting \"$key\" from \"$secret_name\" secret in \"$secret_namespace\" for control plane \"$cp_name\"..."
KUBECONFIG_DATA=$(kubectl get secret $secret_name -n $secret_namespace -o=jsonpath="{.data.$key}")

echo "==> Registering Control Plane \"$cp_name\" as an Argo CD cluster..."
kubectl exec $ARGOCD_POD -- argocd cluster add ${cp_name} -y --kubeconfig <(echo $KUBECONFIG_DATA | base64 -d)
```
but I could not get it to work... `arggocd` cli does not seem to the the kubeconfig data.

core-chart/templates/argocd/operator.yaml
francostellari: I removed

docs/content/direct/core-chart-argocd..md
francostellari: removed

docs/content/direct/core-chart-argocd..md
francostellari: For the moment I added a section to the core-chart.yaml.

core-chart/values.yaml
francostellari: It's an argo concept to group applications:
https://argo-cd.readthedocs.io/en/stable/user-guide/projects/

francostellari: I added some text to the values

docs/content/direct/core-chart-argocd..md
francostellari: reworded

francostellari:
> I think that we envision distinct roles for (a) author/source/importer of workload and (b) author/source of KubeStellar control objects. So I think that a better example would be to use an existing workload, and this would not bundle the control objects with the workload objects. I think an obvious candidate would be the one mentioned in the ArgoCD "Getting Started": https://argo-cd.readthedocs.io/en/stable/getting_started/#6-create-an-application-from-a-git-repository

@MikeSpreitzer can you please elaborate a little more?

I interpret the first part of the comment as I should create 2 apps, one syncing a workload and one synching the binding policy.
But I don't understand the connection with link to Argo.

In my current understanding I would change the example repo to:

```text
scenario-1/
+workload/
 +workload.yaml
+policy/
 +policy.yaml
```

Then have 2 apps:

```bash
  --set-json='argo-cd.applications=[ \
      { \
        "name": "scenario-1-workload", \
        "repoURL": "https://github.com/francostellari/kubestellar", \
        "path": "config/argocd/scenario-1/workload", \
        "destinationWDS": wds1" \
      },\
      { \
        "name": "scenario-1-policy", \
        "repoURL": "https://github.com/francostellari/kubestellar", \
        "path": "config/argocd/scenario-1/policy", \
        "destinationWDS": wds1" \
      }\    ]'
```

Would this do what you discuss?

core-chart/templates/NOTES.txt
```diff
@@ -19,3 +19,11 @@ kflex ctx --overwrite-existing-context {{ $cp.name }}
 
 Finally, you can use `kflex ctx` to switch back to the kubeconfig
 context for your KubeFlex hosting cluster.
+
+{{- if (index .Values "argo-cd" "install") }}
+
+Access Argo CD UI at https://{{ index .Values "argo-cd" "global" "domain" }} (append :9443 for Kind or k3s installations).
```
MikeSpreitzer, line 25: What about other sorts of plain Kubernetes clusters? Are they guaranteed to have their Ingress controller listening at external port 443?

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,153 @@
+{{- if index .Values "argo-cd" "install" }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: "{{ .Release.Name }}-argocd-kubestellar-operator"
+data:
+  # This script by the "{{ .Release.Name }}-argocd-kubestellar-operator" job to
```
MikeSpreitzer, line 7: Did you forget to write the words "is used"?

core-chart/templates/argocd/operator.yaml
MikeSpreitzer: Yes, I think you did half of what I was trying to suggest. The `--kubeconfig` flag expects a filename, not file contents. I was thinking that you could create the file in the ARGOCD_POD without having a file on the user's machine. Perhaps by execing something like `echo "$(echo "$KUBECONFIG_DATA")" > /tmp/kubeconfig` on the ARGOCD_POD. Or maybe, if there is an issue with command line length, use something like `kubectl exec $'cat >/tmp/kubeconfig <<<EOF\n${KUBECONFIG_DATA}\nEOF\n'`.

docs/content/direct/core-chart.md
```diff
@@ -346,6 +346,10 @@ the `kflex` CLI and one not.
     - `--silent|-s` quiet mode, do not print information. This may be useful when using `-o -`.
     - `-X` enable verbose execution of the script for debugging
 
+## Argo CD integration
+
+KubeStellar Core Helm chart allows to deploy ArgoCD in the KubeFlex hosting cluster, register every WDS as a target cluster in Argo CD, and create Argo CD applications as specified by chart values, as asxplained [here](core-chart-argocd.md).
```
MikeSpreitzer, line 351: s/asxplained/explained/

docs/content/direct/core-chart-argocd..md
MikeSpreitzer: Do please render the website for this PR. See https://docs.kubestellar.io/unreleased-development/contribution-guidelines/operations/document-management/#rendering-and-previewing-modifications-to-the-website if you need clues.

When a website page is not listed in mkdocs.yml, the left-hand navigation panel is missing on that page. So the absolutely minimal PR would also add a line in mkdocs.yml. For clues, look at the line for the main core chart page.

docs/content/direct/core-chart-argocd.md
```diff
@@ -0,0 +1,58 @@
...
+
+The initial password for the `admin` user can be retrieved using the command:
+
+```shell
+kubectl get secret argocd-initial-admin-secret -o jsonpath="{.data.password}" | base64 -d
+```
+
+## Deploying Argo CD applications
+
+The KubeStellar Core chart can also be used to deploy Argo CD applications as specified by chart values. The example below show the relevant fragment of the chart values that could be used to deploying an application corresponding to `scenario 1` in [KubeStellar docs](example-scenarios.md#scenario-1---multi-cluster-workload-deployment-with-kubectl).
```
MikeSpreitzer, line 33: s/show/shows/

MikeSpreitzer:
Regarding the example, maybe it would be good to dovetail with the existing argocd example (scenario 6) rather than scenario 1? BTW, @pdettori , why does this not use the same GitHub source as the ArgoCD Getting Started (https://argo-cd.readthedocs.io/en/stable/getting_started/#6-create-an-application-from-a-git-repository) or another one from their repo?

I am also suggesting that we do not add git content for this example. The application source can be from nginx or argocd. For the BindingPolicy, just refer the reader to example scenario 6.

MikeSpreitzer (commented):

docs/content/direct/core-chart-argocd..md
francostellari: ok, I have added a line

core-chart/templates/NOTES.txt
francostellari: No, I don't think there is a general solution. I'm trying to handle the cases that we document for new users to help them getting going.
If you have a specific suggestion, I'm open to make a change.

francostellari:
> Regarding the example, maybe it would be good to dovetail with the existing argocd example (scenario 6) rather than scenario 1? BTW, @pdettori , why does this not use the same GitHub source as the ArgoCD Getting Started (https://argo-cd.readthedocs.io/en/stable/getting_started/#6-create-an-application-from-a-git-repository) or another one from their repo?
> 
> I am also suggesting that we do not add git content for this example. The application source can be from nginx or https://github.com/argoproj/argocd-example-apps . For the BindingPolicy, just refer the reader to example scenario 6.

I have changed the example in `core-chart-argocd.md` to match scenario-6. However, I think it's a miss not including a binding Policy because it will not be visible to a user just playing with Argo CD. A manual stepp will be required to apply the policy. Instead in the previous example the user would have edited the policy in argo.

core-chart/templates/argocd/operator.yaml
MikeSpreitzer: I think that we can do better, but that can be pursued in follow-on work.

core-chart/templates/NOTES.txt
MikeSpreitzer: If I understand correctly, there _is_ a general story. Access the ArgoCD UI at `https://{{ index .Values "argo-cd" "global" "domain" }}:portnum` where domain:portnum is where the Ingress or Route controller is listening. For OpenShift clusters, portnum is 443 (and thus can be omitted because it is the default port for https). For k3d and kind clusters made by our scripting, the portnum is 9443. For other clusters it depends on the construction of the cluster.

MikeSpreitzer:
> However, I think it's a miss not including a binding Policy because it will not be visible to a user just playing with Argo CD.

Good point! In fact, that is a good point for example 6 in general. I suggest that we improve example 6 in follow-on work, keeping the doc introduced here with example 6.

docs/content/direct/core-chart-argocd.md
```diff
@@ -0,0 +1,51 @@
+# Using Argo CD with KubeStellar Core chart
+
+This documents explains how to use the KubeStellar core Helm chart to:
+
+- deploy Argo CD in the KubeFlex hosting cluster;
+- register every WDS as a target cluster in Argo CD; and
+- create Argo CD applications as specified by the chart values.
```
MikeSpreitzer, line 7: s/the//

docs/content/direct/core-chart-argocd.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Pre-requisites
+
+The prerequisites are the same as [installing KubeStellar using the Core chart](core-chart.md#pre-requisites).
+
+The settings described in this document are an extesnion of the KubeStellar Core chart settings described [here](core-chart.md#kubestellar-core-chart-values).
+
+## Installing Argo CD using KubeStellar Core chart
+
+To enable the installation of Argo CD by the KubeStellar Core chart, use the flag `--set "argo-cd.install=true"`. Besides deploying an instance of Argo CD, KubeStellar Core chart will take care of registering all the WDSes installed by the chart as Argo CD target clusters.
```
MikeSpreitzer, line 17: dquotes not needed here, right?

docs/content/direct/core-chart-argocd.md
```diff
@@ -0,0 +1,51 @@
...
+
+When deploying in an **OpenShift** cluster, add the flag `--set "argo-cd.openshift.enabled=true"`.
+
+When deploying in a **Kubernetes** cluster, use the flag `--set "argo-cd.global.domain=<url>"` to provide the URL for the **nginx** ingress, default `argocd.localtest.me`.
```
MikeSpreitzer, line 21: s/default/which defaults to/

I would add a remark at the end that this default works for clusters created by `kind` and `k3d` but not in general.

docs/content/direct/core-chart-argocd.md
```diff
@@ -0,0 +1,51 @@
...
+
+The initial password for the `admin` user can be retrieved using the command:
+
+```shell
+kubectl get secret argocd-initial-admin-secret -o jsonpath="{.data.password}" | base64 -d
+```
+
+## Deploying Argo CD applications
+
+The KubeStellar Core chart can also be used to deploy Argo CD applications as specified by chart values. The example below shows the relevant fragment of the chart values that could be used to deploying an application corresponding to `scenario-6` in [KubeStellar docs](example-scenarios.md#scenario-6---multi-cluster-workload-deployment-of-app-with-serviceaccount-with-argocd).
```
MikeSpreitzer, line 33: s/used to deploying/used for deploying/

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,158 @@
...
+        ARGOCD_POD=$(kubectl -n $NS get pods -l app.kubernetes.io/name=argocd-server -o 'jsonpath={.items[0].metadata.name}' 2> /dev/null || true)
+        if [ -z "$ARGOCD_POD" ]; then
+            >&2 timestamped_echo "==> ERROR: Argo CD pod not found! ...retrying in $WAIT_BEFORE_RETRY seconds..."
+            sleep $WAIT_BEFORE_RETRY
+        fi
+    done
+    timestamped_echo "==> Argo CD pod found: $ARGOCD_POD"
+
+    timestamped_echo "Waiting for Argo CD pod to be ready..."
+    kubectl -n $NS wait --for=condition=Ready pod/$ARGOCD_POD
```
MikeSpreitzer, line 35: The default timeout is 30 seconds. Do we want to wait only that long?

docs/content/direct/core-chart-argocd.md
```diff
@@ -0,0 +1,51 @@
...
+    targetRevision: HEAD # default: HEAD
+    path: nginx
+    destinationWDS: wds1
+    destinationNamespace: nginx-sa # default: default
+```
+
+Alternatively, the same result can be achieved from Helm CLI by using the followig minimal argument (note that the default values are not explicitely set):
+
+```shell
+--set-json='argo-cd.applications=[ { "name": "scenario-6", "repoURL": "https://github.com/pdettori/sample-apps.git", "path": "nginx", "destinationWDS": wds1", "destinationNamespace": "nginx-sa" } ]'
```
MikeSpreitzer, line 50: Missing dquote before `wds1"`.

MikeSpreitzer:
I did some testing. After creating the KubeFlex hosting cluster, I did the following commands, based on extending https://github.com/kubestellar/kubestellar/blob/5b1197aeb9a96287bf5a18a992849b70ec1c7eb9/docs/content/direct/get-started.md#use-core-helm-chart-to-initialize-kubeflex-and-create-its-and-wds with the content added in this PR.

```shell
helm dependency update core-chart
    --version $kubestellar_version \
    --set-json='ITSes=[{"name":"its1"}]' \
    --set-json='WDSes=[{"name":"wds1"},{"name":"wds2", "type":"host"}]' \
    --set-json='verbosity.default=5' \
    --set argo-cd.install=true \
    --set-json='argo-cd.applications=[ { "name": "scenario-6", "repoURL": "https://github.com/pdettori/sample-apps.git", "path": "nginx", "destinationWDS": "wds1", "destinationNamespace": "nginx-sa" } ]'
```

That completed without complaint. Then I did argocd login on web and CLI.

wds2 looks troubled.

```console
bash-3.2$ argocd cluster list
SERVER                                      NAME  VERSION  STATUS      MESSAGE                                                  PROJECT
https://kubernetes.default.svc              wds2           Unknown     Cluster has no applications and is not being monitored.  
https://wds1.wds1-system.svc.cluster.local  wds1  1.29     Successful                                                           
```

I also looked at the following. It is a bit disappointing, but what I should expect at this point, right?

```console
bash-3.2$ argocd app list
NAME                CLUSTER  NAMESPACE  PROJECT  STATUS     HEALTH   SYNCPOLICY  CONDITIONS  REPO                                         PATH   TARGET
default/scenario-6  wds1     nginx-sa   default  OutOfSync  Missing  <none>      <none>      https://github.com/pdettori/sample-apps.git  nginx  HEAD
```

MikeSpreitzer:
I then proceeded with the rest of the step-by-step setup and thence to example scenario 6. The connection is not entirely there.

```console
bash-3.2$ argocd app sync nginx-sa
FATA[0000] rpc error: code = PermissionDenied desc = permission denied 
```

However, this was more rewarding.

```console
bash-3.2$ argocd app sync scenario-6
TIMESTAMP                  GROUP        KIND       NAMESPACE                  NAME    STATUS    HEALTH        HOOK  MESSAGE
2025-03-11T03:33:45-04:00          Namespace                              nginx-sa  OutOfSync  Missing              
2025-03-11T03:33:45-04:00         ServiceAccount    nginx-sa              nginx-sa  OutOfSync  Missing              
2025-03-11T03:33:45-04:00   apps  Deployment        nginx-sa   nginx-sa-deployment  OutOfSync  Missing              
2025-03-11T03:33:45-04:00          Namespace                          nginx-sa    Synced  Missing              
2025-03-11T03:33:45-04:00         ServiceAccount    nginx-sa              nginx-sa    Synced  Missing              
2025-03-11T03:33:45-04:00         ServiceAccount    nginx-sa              nginx-sa    Synced   Missing              serviceaccount/nginx-sa created
2025-03-11T03:33:45-04:00   apps  Deployment        nginx-sa   nginx-sa-deployment  OutOfSync  Missing              deployment.apps/nginx-sa-deployment created
2025-03-11T03:33:45-04:00          Namespace        nginx-sa              nginx-sa   Running    Synced              namespace/nginx-sa created
2025-03-11T03:33:45-04:00   apps  Deployment    nginx-sa   nginx-sa-deployment    Synced  Progressing              deployment.apps/nginx-sa-deployment created

Name:               default/scenario-6
Project:            default
Server:             wds1
Namespace:          nginx-sa
URL:                https://argocd.localtest.me/applications/scenario-6
Repo:               https://github.com/pdettori/sample-apps.git
Target:             HEAD
Path:               nginx
SyncWindow:         Sync Allowed
Sync Policy:        <none>
Sync Status:        Synced to HEAD (ff20d34)
Health Status:      Progressing

Operation:          Sync
Sync Revision:      ff20d344c858f5008af5049930d98c882f9afdb8
Phase:              Succeeded
Start:              2025-03-11 03:33:44 -0400 EDT
Finished:           2025-03-11 03:33:44 -0400 EDT
Duration:           0s
Message:            successfully synced (all tasks run)

GROUP  KIND            NAMESPACE  NAME                 STATUS   HEALTH       HOOK  MESSAGE
       Namespace       nginx-sa   nginx-sa             Running  Synced             namespace/nginx-sa created
       ServiceAccount  nginx-sa   nginx-sa             Synced                      serviceaccount/nginx-sa created
apps   Deployment      nginx-sa   nginx-sa-deployment  Synced   Progressing        deployment.apps/nginx-sa-deployment created
       Namespace                  nginx-sa             Synced                      
```

.. and then

```console
bash-3.2$ argocd app list
NAME                CLUSTER  NAMESPACE  PROJECT  STATUS  HEALTH       SYNCPOLICY  CONDITIONS  REPO                                         PATH   TARGET
default/scenario-6  wds1     nginx-sa   default  Synced  Progressing  <none>      <none>      https://github.com/pdettori/sample-apps.git  nginx  HEAD
```

core-chart/templates/argocd/operator.yaml
francostellari: instead of making the wait longer, I made a loop to print a periodic message consistent with the other checks. I also noticed that I was missing `watch` rbac.

francostellari:
> I did some testing. After creating the KubeFlex hosting cluster, I did the following commands, based on extending https://github.com/kubestellar/kubestellar/blob/5b1197aeb9a96287bf5a18a992849b70ec1c7eb9/docs/content/direct/get-started.md#use-core-helm-chart-to-initialize-kubeflex-and-create-its-and-wds with the content added in this PR.
> 
> ```shell
> helm dependency update core-chart
> helm upgrade --install ks-core core-chart \
>     --version $kubestellar_version \
>     --set-json='ITSes=[{"name":"its1"}]' \
>     --set-json='WDSes=[{"name":"wds1"},{"name":"wds2", "type":"host"}]' \
>     --set-json='verbosity.default=5' \
>     --set argo-cd.install=true \
>     --set-json='argo-cd.applications=[ { "name": "scenario-6", "repoURL": "https://github.com/pdettori/sample-apps.git", "path": "nginx", "destinationWDS": "wds1", "destinationNamespace": "nginx-sa" } ]'
> ```
> 
> That completed without complaint. Then I did argocd login on web and CLI.
> 
> wds2 looks troubled.
> 
> ```
> bash-3.2$ argocd cluster list
> SERVER                                      NAME  VERSION  STATUS      MESSAGE                                                  PROJECT
> https://kubernetes.default.svc              wds2           Unknown     Cluster has no applications and is not being monitored.  
> https://wds1.wds1-system.svc.cluster.local  wds1  1.29     Successful                                                           
> ```

I think that the message is correct and it means what it says. Since there is no app associated to wds2, Argo is ignoring it. I tried assigning the app to wds2 and it looks ok.

Here is an example where no app is deployed (wds0==wds2):

```shell
$ argocd cluster list
SERVER                                      NAME  VERSION  STATUS   MESSAGE                                                  PROJECT
https://kubernetes.default.svc              wds0           Unknown  Cluster has no applications and is not being monitored.
https://wds1.wds1-system.svc.cluster.local  wds1           Unknown  Cluster has no applications and is not being monitored.
```

Here is an example where the scenario-6 app is associated to wds0 instead of wds1 (wds0==wds2)::

```shell
$ argocd cluster list
SERVER                                      NAME  VERSION  STATUS      MESSAGE                                                  PROJECT
https://kubernetes.default.svc              wds0  1.32     Successful
https://wds1.wds1-system.svc.cluster.local  wds1           Unknown     Cluster has no applications and is not being monitored.
```

francostellari:
> I also looked at the following. It is a bit disappointing, but what I should expect at this point, right?
> 
> ```
> bash-3.2$ argocd app list
> NAME                CLUSTER  NAMESPACE  PROJECT  STATUS     HEALTH   SYNCPOLICY  CONDITIONS  REPO                                         PATH   TARGET
> default/scenario-6  wds1     nginx-sa   default  OutOfSync  Missing  <none>      <none>      https://github.com/pdettori/sample-apps.git  nginx  HEAD
> ```

I think this is expected since 1. the app was not syncked and 2, health will be missing

francostellari:
In the [last commit](https://github.com/kubestellar/kubestellar/pull/2818/commits/ee414740dd874092314950917f7831b75a32102c) I have enable app auto sync capability.

core-chart/templates/NOTES.txt
MikeSpreitzer: So there is editing to be done here.

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,163 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: "{{ .Release.Name }}-argocd-kubestellar-operator"
+data:
+  # This script is used by the "{{ .Release.Name }}-argocd-kubestellar-operator" job to
+  # 1. register KubeStellar WDSes as Argo CD target clusters
+  # 2. create any Argo CD application listed in KubeStelar values
+  argocd-operator.sh: |
+    #!/bin/env bash
```
MikeSpreitzer, line 11: `env` is at `/usr/bin/env`. That's the standard place. That is the only place that it is found on my Mac and in Ubuntu 22.04 and Ubuntu 20.04. That is the standard place used in the files in `scripts/`. That location is mentioned in https://man7.org/linux/man-pages/man1/env.1.html .

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,163 @@
...
+    done
+    timestamped_echo "==> Argo CD pod found: $ARGOCD_POD"
+
+    timestamped_echo "Waiting for Argo CD pod to be ready..."
+
+    until kubectl -n $NS wait --for=condition=Ready pod/$ARGOCD_POD --timeout=${WAIT_BEFORE_RETRY}s
+    do
+      >&2 timestamped_echo "==> ERROR: Argo CD pod is not ready yet! ...retrying..."
+    done
+    timestamped_echo "==> Argo CD shows condition Ready."
```
MikeSpreitzer, line 40: Did you want the word "pod" in this message (like its companion above)?

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,163 @@
...
+      timestamped_echo "==> Control Plane \"$cp_name\" is ready."
+
+      # determine the secret name and namespace
+      key=$(kubectl get controlplane $cp_name -o=jsonpath='{.status.secretRef.inClusterKey}')
+      secret_name=$(kubectl get controlplane $cp_name -o=jsonpath='{.status.secretRef.name}')
+      secret_namespace=$(kubectl get controlplane $cp_name -o=jsonpath='{.status.secretRef.namespace}')
+
+      kubeconfig="/tmp/${cp_name}.kubeconfig"
+
+      timestamped_echo "==> Getting \"$key\" from \"$secret_name\" secret in \"$secret_namespace\" for control plane \"$cp_name\"... \"$kubeconfig\""
```
MikeSpreitzer, line 80: The "..." is not very explanatory. How about something like the following?
```suggestion
      timestamped_echo "==> Populating \"$kubeconfig\" from key \"$key\" in secret \"$secret_name\" in namespace \"$secret_namespace\" for control plane \"$cp_name\""
```

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,163 @@
...
+      kubectl get secret $secret_name -n $secret_namespace -o=jsonpath="{.data.$key}" | base64 -d > "$kubeconfig"
+
+      timestamped_echo "==> Copying kubeconfig \"${cp_name}.kubeconfig\" to Argo CD pod \"$ARGOCD_POD\"..."
```
MikeSpreitzer, line 83: Why use the duplicative and directory-less expression `\"${cp_name}.kubeconfig\"` here instead of `"$kubeconfig"` ?

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,163 @@
...
+
+      timestamped_echo "==> Creating application \"{{ $app.name }}\":"
+      timestamped_echo "    - project=\"{{ $app.project | default "default" }}\""
+      timestamped_echo "    - repoURL=\"{{ $app.repoURL }}\""
+      timestamped_echo "    - targetRevision=\"{{ $app.targetRevision | default "HEAD" }}\""
+      timestamped_echo "    - path=\"{{ $app.path | default "/" }}\""
+      timestamped_echo "    - destinationWDS=\"{{ $app.destinationWDS }}\""
+      timestamped_echo "    - destinationNamespace=\"{{ $app.destinationNamespace | default "default" }}\""
+
+      kubectl exec $ARGOCD_POD -- argocd app create {{ $app.name }} \
```
MikeSpreitzer, line 104: All these user-provided string values should be quoted to prevent shell injection attacks (equivalently, be robust agains mistakes that inject characters interpreted by the shell). Unfortunately, I am not familiar with a function available to Helm templates that does exactly quoting/escaping as needed for bash. The closest that I can recommend is `js`, which formats the string as in JavaScript.

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,163 @@
...
+        --project {{ $app.project | default "default" }} \
+        --repo {{ $app.repoURL }} \
+        --revision {{ $app.targetRevision | default "HEAD" }} \
+        --path {{ $app.path | default "/" }} \
+        --dest-name {{ $app.destinationWDS }} \
+        --dest-namespace {{ $app.destinationNamespace | default "default" }} \
+        --sync-option CreateNamespace=true \
+        --sync-policy {{ $app.syncPolicy | default "manual" }} || true
```
MikeSpreitzer, line 112: The sync policy is not mentioned in the messaging above.

core-chart/templates/argocd/operator.yaml
```diff
@@ -0,0 +1,163 @@
...
+        app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+        app.kubernetes.io/instance: {{ .Release.Name | quote }}
+        helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    spec:
+      restartPolicy: Never # OnFailure
+      securityContext:
+        runAsNonRoot: true
+      serviceAccountName: "{{ .Release.Name }}-argocd-kubestellar-operator"
+      containers:
+      - name: add-wdses-clusters
```
MikeSpreitzer, line 146: FYI, I don't think you need two plurals in this name. I think "add-wds-clusters" would be fine.

core-chart/values.yaml
```diff
@@ -35,11 +35,32 @@ kubeflex-operator:
...
+
+# Argo CD values
+argo-cd:
+  install: false # enable/disaple the installation of Argo CD
+  # openshift:
+  #   enabled: true
+  global:
+    domain: argocd.localtest.me # When using the "create-kind-cluster-with-SSL-passthrough.sh" script, Argo UI will be accessible from the host at https://argocd.localtest.me:9443
+  applications: # list of Argo CD applications to be create
+  # Uncomment the lines below to deploy an application for the scenario-1 example in KubeStellar docs
```
MikeSpreitzer, line 47: We're suggesting example 6 now, right?

core-chart/templates/argocd/operator.yaml
MikeSpreitzer, line 1: In my opinion, calling this an "operator" is less helpful that using terminology like "initialization". The word "operator" conjures up much great expectations (e.g., see https://kubernetes.io/docs/concepts/extend-kubernetes/operator/ and https://www.redhat.com/en/topics/containers/what-is-a-kubernetes-operator).

MikeSpreitzer (commented):
This is good enough to merge. More work could be done in the PR and/or follow-ons.

/lgtm
Thanks!

core-chart/templates/argocd/operator.yaml
francostellari: I changed it to "setup"

core-chart/templates/argocd/operator.yaml
MikeSpreitzer: I see that you put literal dquotes in the source. That is a good start. We should do better in the future.

core-chart/values.yaml
```diff
@@ -35,11 +35,32 @@ kubeflex-operator:
...
+# Argo CD values
+argo-cd:
+  install: false # enable/disaple the installation of Argo CD
+  # openshift:
+  #   enabled: true
+  global:
+    domain: argocd.localtest.me # When using the "create-kind-cluster-with-SSL-passthrough.sh" script, Argo UI will be accessible from the host at https://argocd.localtest.me:9443
+  applications: # list of Argo CD applications to be create
+  # Uncomment the lines below to deploy an application for the scenario-6 example in KubeStellar docs
+  # - name: scenario-1 # Argo CD application name required, must be unique
```
MikeSpreitzer, line 48: Finish edits changing 1 to 6.

MikeSpreitzer (commented):
I left a few more independent comments. Can be addressed in this PR or later.

/lgtm

core-chart/templates/argocd/setup.yaml
```diff
@@ -0,0 +1,164 @@
+{{- if index .Values "argo-cd" "install" }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: "{{ .Release.Name }}-argocd-kubestellar-operator"
```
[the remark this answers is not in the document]
francostellari: I replace all the occurrences in the file.

core-chart/values.yaml
francostellari: I changed it but I think we will need to work on the scenarios in more detail

core-chart/templates/argocd/setup.yaml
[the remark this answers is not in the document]
MikeSpreitzer: I foud some in the RBAC too.

```
mspreitz@mjs13 kubestellar % find * .github/workflows \( -name "*.svg" -prune \) -or \( -path "*venv" -prune \) -or \( -path hack/tools -prune \) -or \( -type f -exec fgrep kubestellar-operator \{\} \; -print -exec echo \; \)
  name: "{{ .Release.Name }}-argocd-kubestellar-operator"
  name: "{{ .Release.Name }}-argocd-kubestellar-operator"
  name: "{{ .Release.Name }}-argocd-kubestellar-operator"
  name: "{{ .Release.Name }}-argocd-kubestellar-operator"
  name: "{{ .Release.Name }}-argocd-kubestellar-operator"
core-chart/templates/argocd/rbac.yaml
```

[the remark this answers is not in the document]
MikeSpreitzer: The good news is, these are not functional bugs.

MikeSpreitzer (approved):
Let's merge this and continue work in follow-on PRs.

/lgtm
/approve

Issue #2816, opened 2025-02-26
bug: curl-to-bash scripts use `$0` to get script name


MikeSpreitzer:
### Describe the bug

In the bash scripts that are documented to be fetched by curl (that is not a critical part) and piped to `bash`, `$0` is used to get the script name. But when the script is piped into `bash`, `$0` will expand to "bash" rather than the name of the script.

### Output from KubeStellar-Snapshot.sh

N/A

### Steps To Reproduce

Discussed above.

### Expected Behavior

The scripts do not attribute their problems to "bash".

### Want to contribute?

- [ ] I would like to work on this issue.

francostellari:
Possible approaches could be:

1. hard coding the name of the script by replacing all `$0` with the name of the script

2. add the script name as placeholder on shebang

```bash
#!/bin/bash -s script-name
echo "Script name: ${0##*-}"
```

3.  digging and parsing through history ... but I guess it is syntax depending `curl|bash` or `bash <(curl ...)`

```bash
# Inside the script
history | tail -n 1 | awk '{print $2}'
```

4. passing the script name on the `curl|bash` command

```shell
curl <url> | bash -s <script-name>
# Inside the script
script_name="$1"
```

In my opinion solution 2 is probably the most clear and easiest to implement.
Solution 1 requires maintaining the script name is multiple places
Solution 3 is risky to cover al possible syntaxes
Solution 4 relies on the user (especially when not cutting and pasting)

DeeBi9:
Hey @francostellari, I'd like to work on this issue

francostellari:
> Hey [@francostellari](https://github.com/francostellari), I'd like to work on this issue

Thank you for helping out. In my testing it looks like that the best solution is to replace the references `$0` with a variable like `$script_name` that is hard coded to the the script name.

DeeBi9:
Sure I'll go ahead with this approach and replace `$0` with a hardcoded variable like `$script_name`. I'll test it and open a PR soon.

PR #2819, opened 2025-03-01
🐛 Fix script name handling in Bash scripts when piped via curl

DeeBi9:
## Summary

Fix script name handling in Bash scripts when piped via curl. Previously, `$0` was used to get the script name, but when the script is piped into Bash, `$0` expands to "bash" instead of the actual script name. This update replaces `$0` with a hardcoded variable `$script_name` to ensure correct script name attribution.  

## Related issue(s)

Fixes #
#2816

### DIFF
--- a/scripts/create-k3s-cluster-with-SSL-passthrough.sh
+++ b/scripts/create-k3s-cluster-with-SSL-passthrough.sh
@@ -17,12 +17,15 @@
 
 set -o errexit
 
+# Script info
+SCRIPT_NAME="Create K3s Cluster with SSL Passthrough" 
+
 port=9443
 wait=true
 
 display_help() {
   cat << EOF
-Usage: $0 [--port port] [--nowait][-X]
+Usage: ${SCRIPT_NAME} [--port port] [--nowait][-X]
 
 --port port     map the specified host port to the kind cluster port 443 (default: 9443)
 --nowait        when set to true, the script proceeds without waiting for the nginx ingress patching to complete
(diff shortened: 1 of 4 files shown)


DeeBi9:
> @DeeBi9, the `SCRIPT_NAME` variable should be the actual script filename, e.g. `kubestellar-snapshot.sh`, not the brief description. For example, the output of `kubestellar-snapshot -h` should be roughly the same before and after your modification.

Thanks for the feedback. I will update the PR and push changes shortly.

francostellari:
thank you
/lgtm

francostellari:
@DeeBi9 if you change your PR description to have the fixed issues number on the same line like below, the issue will be automatically closed upon merging the PR.

```
Fixes #2816 
``

MikeSpreitzer (commented):
Looks good.
One minor style point to consider: capitalize the envar name or not? I prefer envars that are NOT exported to be lower-case, exported ones upper-case. I do not know a strict functional requirement for this, it just seems to be more in line with common style. Also, adjacently, there is a rule in `make` files that upper-case variables are automatically exported.

/ok-to-test
/lgtm
Thanks!

DeeBi9:
> Looks good. One minor style point to consider: capitalize the envar name or not? I prefer envars that are NOT exported to be lower-case, exported ones upper-case. I do not know a strict functional requirement for this, it just seems to be more in line with common style. Also, adjacently, there is a rule in `make` files that upper-case variables are automatically exported.
> 
> /ok-to-test /lgtm Thanks!

Thanks for the suggestion! Currently, `SCRIPT_NAME` is not exported, should I rename it to `$script_name` ?

MikeSpreitzer:
> should I rename it to $script_name ?

I think that would be a small improvement. Not strictly necessary, up to you.

DeeBi9:
> > should I rename it to $script_name ?
> 
> I think that would be a small improvement. Not strictly necessary, up to you.

Since it's not a strict requirement and everything is working fine, I'll keep it as `SCRIPT_NAME` for now. If any issues arise in the future, we can revisit this and make the change.

MikeSpreitzer:
/approve
Thanks!

PR #2704, opened 2025-01-07
✨ skip uploading binaries during release

jaydee029:
## Summary
Skips uploading binaries during release.

## Related issue(s)

Fixes #2653

Issue #2653, opened 2024-12-06
Our release binary archives are pointless and frowned upon


MikeSpreitzer:
Our current release-producing GitHub Actions workflow (https://github.com/kubestellar/kubestellar/blob/main/.github/workflows/goreleaser.yml) uses `goreleaser` to create the GitHub "release". This attaches two binary archives. Look at the "assets" of release https://github.com/kubestellar/kubestellar/releases/tag/v0.25.1 for an example; I am talking about `kubestellar_0.25.1_linux_amd64.tar.gz` and `kubestellar_0.25.1_linux_arm64.tar.gz`.

```console
mspreitz@mjs13 kubestellar % tar tf ~/Downloads/kubestellar_0.25.1_linux_amd64.tar.gz 
LICENSE
README.md
bin/ocm-transport-controller
bin/controller-manager
```

Those two binaries (the `bin/*controller*`) do not belong there. We do not intend users to use those binaries (users should instead be using the container images from ghcr), we do not document using those binaries, we have no scripts that use those binaries, and AFAIK no user **is** using those binaries.

A little more broadly, we have no scripts or instructions that tell users to fetch those archives and do anything with them. There is no usage of those archives.

Can we stop adding these archives to our releases, or at least get the pointless executables out of the archives?

pdettori:
@MikeSpreitzer I have noticed that as well. The problem seems to be that the `ko` extension in the goreleaser tool requires also to define the builds in the `builds` section and without that I could not find a way to convince the `kos` section to produce and publish the images.

MikeSpreitzer:
Can we define the builds but also say to not publish the built executables?

pdettori:
I found the option `skip_upload`  in https://goreleaser.com/customization/release/#github - not sure if it does what we want but we could try.

MikeSpreitzer:
Feel free to offer a PR making this change. I think that it would be OK to test the change by making another alpha release.

jaydee029:
Hey @MikeSpreitzer @pdettori I made the necessary changes and raised a PR [here](https://github.com/kubestellar/kubestellar/pull/2704) , but then I realized in the [test-latest-release](https://github.com/kubestellar/kubestellar/blob/main/.github/workflows/test-latest-release.yml) workflow, these binaries are being used for testing once published as a release.

![image](https://github.com/user-attachments/assets/3008048a-43c6-428f-a85b-4fe3e989c513)

[...]

PR #2821, opened 2025-03-02
✨: Added sbom generation

### LINKED ISSUE #2613: feature: SBOM needed to clomonitor tests

jaydee029:
## Summary

After multiple reviews, I found out, if we are to build the sbom using the go-releaser functionality, we have remove the `skip_upload: true ` option , this would obviously lead to creation of some unrequired binaries as seen before, but still serves our purpose. Another way to do it is to create a hook in go-releaser and use an external script to create an sbom.

I have tested it in a test repo [here](https://github.com/jaydee029/test)
It consists of a basic go-releaser config, and multiple releases (with variations)

As soon as this is merged I'll create a PR for the signed binary issue
## Related issue(s)

Fixes #2613

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -52,8 +52,14 @@ release:
   draft: false
   prerelease: auto
   mode: keep-existing
-  skip_upload: true
+
 sboms:
   - id: source
     artifacts: source
+    documents:
+      - "{{ .ProjectName }}_{{ .Version }}_source.sbom"
+  - id: binary
+    artifacts: binary
+    documents:
+      - "{{ .Binary }}_{{ .Os }}_{{ .Arch }}.sbom"
 


MikeSpreitzer:
To be clear, this undoes PR #2704 and brings back Issue #2643.

jaydee029:
> To be clear, this undoes PR #2704 and brings back Issue #2653.

Right, it does. Seems like the only option, creating a hook in go-releaser and writing a script to generate an sbom seems like a lot of technical debt

MikeSpreitzer (commented):
Yeah, I suppose that having pointless archives is less of a problem than lacking an SBOM. And yes, we are not in the business of developing/maintaining goreleaser.

/lgtm
Thanks

MikeSpreitzer:
@jaydee029: we made a new (test) release, 0.27.0-rc.1. And it has some SBOMs! Does it match your expectations?

clubanderson:
here is what clomonitor is looking for - https://clomonitor.io/docs/topics/checks/#software-bill-of-materials-sbom

please make sure it matches their search criteria, or let me know and I will create an exception. This should get us higher than a 92 score. Thank you @jaydee029 - great job! Please look around and see if there is anything else you can help us with - we would love to have you stick around.

MikeSpreitzer:
The criterion at https://clomonitor.io/docs/topics/checks/#software-bill-of-materials-sbom is very loose --- it only requires that there exists at least one attachment whose name looks like an SBOM. I am sure that criterion will be met when next evaluated. Oh, it also requires the README to reference one or more things that look like SBOMs; we still have to do that.

What I am asking @jaydee029 and the rest of us is whether we have any higher ambition and, if so, whether it is met. Kind of SBOM? Thoroughness? Accuracy? Granularity? Coverage?

jaydee029:
Thanks @clubanderson , I am looking at the signed releases issue now.
@MikeSpreitzer I think what we can do is just create an sbom , sign it outside of the go-releaser action and just upload it along with the release , since we have to implement signed releases as well. This will avoid those unnecessary binaries as well since then we can  just `skip upload` them. This would give us more control over the sbom.

This can be followed by the `workflow permission configuration` for every workflow, this would further increase the ossf score.

clubanderson:
Clomonitor did not register our SBOM in its latest release. Let's make the change to the Readme to reference it before moving on to signing of releases. Let's complete SBOM recognition first. Thank you

jaydee029:
Sure , we can do that, but are you certain its not because the release is 'pre release' cause I have done this for orgs before, it has always worked without mentioning it in the project readme, almost none of those projects have this mentioned in their readme.

clubanderson:
it could be, but having the readme updated will be a change to main and that is likely where they scan.

MikeSpreitzer:
The words in the doc are "latest release". Whether they consider a GitHub "pre-release" to be a "release" is, unfortunately, not totally obvious to me.

But in any case, we need to get the README fixed.

jaydee029:
since clomonitor has now picked up SBOM can we move forward with signed releases?
Also I am looking at ways to avoid uploading the binaries (not req ones) and still get the sbom.
one way being producing sbom outside the goreleaser, as we planned earlier.

Issue #2844, opened 2025-03-17
bug: GitHub actions not pinned to hash


MikeSpreitzer:
### Describe the bug

https://www.wiz.io/blog/github-action-tj-actions-changed-files-supply-chain-attack-cve-2025-30066 did not hit us, but another like it could. The recommended practice is to reference an Action by hash rather than tag. https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

### Output from KubeStellar-Snapshot.sh

N/A

### Steps To Reproduce

N/A

### Expected Behavior

Supply chain not so easily hacked.

### Want to contribute?

- [ ] I would like to work on this issue.

rxinui:
Hey, if there is a need for a contributor, I'm in @MikeSpreitzer

MikeSpreitzer:
The fix is going to keep getting undone until we document and enforce it.

PR #2845, opened 2025-03-17
🐛  Pin actions/checkout to SHA in pr-test-integration

MikeSpreitzer:
## Summary
The PR changes the way that the pr-test-integration workflow refers to actions/checkout, to improve supply chain security by pinning to a specific SHA256 instead of a tag.

## Related issue(s)

These address part of #2844

### DIFF
--- a/.github/workflows/pr-test-integration.yml
+++ b/.github/workflows/pr-test-integration.yml
@@ -18,7 +18,7 @@ jobs:
     name: Test
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
 
       - uses: actions/setup-go@v5
         with:


.github/workflows/pr-verifier.yaml
```diff
@@ -18,6 +18,6 @@ jobs:
     steps:
     - name: Verifier action
       id: verifier
-      uses: kubernetes-sigs/kubebuilder-release-tools@v0.4.3
+      uses: kubernetes-sigs/kubebuilder-release-tools@012269a88fa4c034a0acf1ba84c26b195c0dbab4
```
MikeSpreitzer, line 21: This is only half of what needs to be done for this particular action, because https://github.com/kubernetes-sigs/kubebuilder-release-tools/blob/012269a88fa4c034a0acf1ba84c26b195c0dbab4/action.yml#L12 . Perhaps we should take more drastic action, as suggested in https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#pr-verification-github-action-deprecated . However, we seem to be using a workaround that is not limited to March 18, so not sure what to do when.

MikeSpreitzer:
The force-push to f93ec800b is a rebase onto `main`.

pdettori (commented):
/lgtm

PR #2858, opened 2025-03-26
🐛  More excluding clusteradm release 0.11

MikeSpreitzer:
## Summary
This PR updates the pre-reqs doc and checking script to exclude release 0.11 and higher of clusteradm.

This PR also updates the core Helm chart's default version of clusteradm from 0.9.0 to 0.10.0.

## Related issue(s)

Fixes #


core-chart/values.yaml
```diff
@@ -9,7 +9,7 @@ HELM_VERSION: "3.16.1"
 KUBECTL_VERSION: "1.31.0"
 KUBESTELLAR_VERSION: "0.27.0"
 # TRANSPORT_VERSION: e.g., "0.24.0"; defaults to KUBESTELLAR_VERSION value, if not defined
-CLUSTERADM_VERSION: "0.9.0"
+CLUSTERADM_VERSION: "0.10.0"
```
francostellari, line 12: Why not use 0.10.1?

MikeSpreitzer: Because we have not built a container image for release 0.10.1

MikeSpreitzer: Updated to 0.10.1, now that it exists (thanks for putting it there).

PR #2873, opened 2025-04-03
:book: fixup: incorrect links to Getting Started page

andriisoldatenko:
## Summary

fixup wrong links in CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,7 +15,7 @@ Please read the following guidelines if you're interested in contributing to Kub
 
 ### Contributing Code -- Prerequisites
 
-Please make sure that your environment has all the necessary versions as spelled out in the prerequisites section of our [user guide](../direct/pre-reqs.md)
+Please make sure that your environment has all the necessary versions as spelled out in the prerequisites section of our [user guide](./docs/content/direct/pre-reqs.md)
 <!--end-first-include-->
 (If you are viewing this file in the repository, the [pre-req listing is in the docs subfolder](./docs/content/direct/pre-reqs.md))
 <!--start-second-include-->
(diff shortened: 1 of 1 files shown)


KPRoche:
Hi @andriisoldatenko  -- 
The links you are correcting were put in this document are there for a specific reason.
In order to avoid keeping two copies of this file in sync in the repository, the contents of this file are pulled into another page on the kubestellar.io website via the include-markdown plugin. 
This particular file, however, is required to exist in the repository root on github.

If you look you will see that the links you changed things to are *already* in this file very near to the "wrong" links,  with a notation "If you are viewing this file in the repository"
If we change the links as suggested in your PR they will work in the repository but break on the website. It's an issue with how mkdocs parses relative links (which we have to use because of the multiple releases of KubeStellar, each with their own corresponding version of the mkdocs-generated site) 

I _can_ see a way to fix this, but it will entail slicing the includes in more places, so the website links are all in the calling file (contributing-inc.md) and the repository links are all excluded from the include segments.

andriisoldatenko:
That's okey I can close PR.

KPRoche:
> That's okey I can close PR.

I'll see if I can do the re-slice I described later today and then I'll ask you to test it for me, to make sure the repository version and the website version both work?

Your PR made it clear to me that the reason for the double links was not obvious enough, so it's actually quite helpful!

KPRoche:
@andriisoldatenko Check out PR #2876 
You can help immensely by comparing the link behaviors on the webpage version and the repository version as noted in the PR

KPRoche:
Hi @andriisoldatenko we made the fix via PR #2876 so you can close this PR. Thank you for spotting the problem!

PR #2875, opened 2025-04-08
📖  Updated spellcheck wordlist and fixed typos across files.

trinaya-kantevari:


.github/spellcheck/.wordlist.txt
```diff
@@ -552,6 +597,7 @@ metav
 mishi
 mixup
 mkdocs
+mmmm
```
MikeSpreitzer, line 600: Why is this here?

.github/spellcheck/.wordlist.txt
```diff
@@ -301,17 +329,20 @@ ValidatingWebhookConfiguration
 VolumeAttachment
 WDS
 WDSes
+WDSs
```
MikeSpreitzer, line 332: This is for one item in olddocs/, right? Since those are not maintained, maybe we could skip them in the spell-checking?

docs/content/direct/get-started.md
```diff
@@ -8,7 +8,7 @@ The two ways to create this simple configuration are as follows.
 
 1. A [quick automated setup](#quick-start-using-the-automated-script) using our demo setup script, which creates a basic working environment for those who want to start experimenting right away.
 
-2. A [Step by step walkthrough](#step-by-step-setup) that demonstrates the core concepts and components, showing how to manually set up a simple single-host system.
+2. A [Step by step walk through](#step-by-step-setup) that demonstrates the core concepts and components, showing how to manually set up a simple single-host system.
```
MikeSpreitzer, line 11: IMO "walk-through" would be better.

oldocs/deploy-on-k3d.md
```diff
@@ -51,9 +51,9 @@ In addition to [pre-reqs](pre-reqs.md), install k3d v5.6.0 (only k3d version tes
     ```
 
 1. Install kubestellar controller and OCM space:
-   We are using nginx ingress with tls passthru.
+   We are using nginx ingress with tls pass through.
```
MikeSpreitzer, line 54: "TLS passthrough" actually seems to be common usage. IMO it would be fine to add "passthrough" to the dictionary. Using "TLS pass-through" would be better than "TLS pass through", IMO.

MikeSpreitzer (commented):
/lgtm
But @trinaya-kantevari , you need to do the DCO signoff. See https://docs.kubestellar.io/release-0.27.2/direct/pr-signoff/

PR #2883, opened 2025-04-27
🐛  ci now uses commitsha over tag

### LINKED ISSUE #2844: bug: GitHub actions not pinned to hash

rxinui:
## Summary

To avoid any [CI supply chain attack](https://www.wiz.io/blog/github-action-tj-actions-changed-files-supply-chain-attack-cve-2025-30066), all actions tag has been replaced by its commit sha.

To facilitate the process, as commit sha are humanly hard to read, a new file `.github/.reversemap.yaml` is added. Its purpose is informational and allows easy lookup on sha/tag locally to further develop the ci.

```yaml
actions/checkout:
  commit-sha: 11bd71901bbe5b1630ceea73d27597364c9af683
  tag: v4
actions/add-to-project:
  tag: v1.0.2
  commit-sha: 244f685bbc3b7adfa8466e08b698b5577571133e
```

For that, a `hack/ci-update-reversemap.sh` is added to automatically scan the `.github/workflows`. Whenever an action is used with a tag, launching the script will lookup for its commitsha and add both to the reversemap to keep track. 

ℹ️ _Note: this was useful to implement the fix, may no longer be from now on. Also, the script does not do in-replacement within workflows as of now._

## Related issue(s)

Fixes #2844

### DIFF
--- a/hack/ci-update-reversemap.sh
+++ b/hack/ci-update-reversemap.sh
@@ -1,0 +1,107 @@
+#!/bin/bash
+
+# author: @rxinui
+
+# Define the required version of yq
+GITHUB_PAT=
+GITHUB_WORKFLOWS_PATH="./.github/workflows/"
+REVERSEMAP_FILE="./.github/.reversemap.yml"
+YQ_REQUIRED_VERSION="v4.45.1"
+YQ_IS_INSTALLED=0
+GIT_COMMITSHA_LENGTH=40
+
+# Function to check if yq is installed and has the required version
+check_yq_version() {
+    if command -v yq >/dev/null 2>&1; then
+        INSTALLED_VERSION=$(yq --version 2>/dev/null)
+        if [[ "$INSTALLED_VERSION" =~ "$YQ_REQUIRED_VERSION" ]]; then
+            echo "yq $YQ_REQUIRED_VERSION is already installed."
+            YQ_IS_INSTALLED=1
+        else
+            echo "yq is installed but the version is $INSTALLED_VERSION. Required version is $YQ_REQUIRED_VERSION."
+        fi
+    else
+        echo "yq is not installed."
+    fi
+}
+
+# Function to download and install yq
+install_yq() {
+    OS=$(uname | tr '[:upper:]' '[:lower:]')
+    ARCH=$(uname -m)
+
+    if [[ "$ARCH" == "x86_64" ]]; then
+        ARCH="amd64"
+    elif [[ "$ARCH" == "aarch64" ]]; then
+        ARCH="arm64"
+    elif [[ "$ARCH" == "arm64" ]]; then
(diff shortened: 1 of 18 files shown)


hack/ci-update-reversemap.sh
```diff
@@ -0,0 +1,107 @@
...
+    curl -L -o /tmp/${YQ_BINARY}.tar.gz "$DOWNLOAD_URL"
+
+    if [[ $? -ne 0 ]]; then
+        echo "Failed to download yq."
+        exit 1
+    fi
+
+    tar xzf /tmp/${YQ_BINARY}.tar.gz -C /tmp
+    sudo mv /tmp/${YQ_BINARY} /usr/local/bin/yq
+    sudo chmod +x /usr/local/bin/yq
```
MikeSpreitzer, line 57: Maybe do this before the `mv`, thinking of possible concurrent usage?

rxinui: very low likelihood to me but i can switch it!

rxinui:
@MikeSpreitzer changes applied

MikeSpreitzer:
For supply chain security, the part we would want automating is propagating from `.github/.reversemap.yml` to the `.github/workflows/*.yml` files, not assuming that the latest version of everything is good.

rxinui:
> For supply chain security, the part we would want automating is propagating from `.github/.reversemap.yml` to the `.github/workflows/*.yml` files, not assuming that the latest version of everything is good.

It was not part of the issue requirements but I can implement it. The question would then be: "when/where does this hack should be triggered and by who"

MikeSpreitzer:
What I am thinking is to replace `hack/ci-update-reversemap.sh` with two scripts.

One script updates all of the `.github/workflows/*.yml` files based on the current contents of `.github/.reversemap.yml`. The idea is that `.github/.reversemap.yml` holds the dependency versions that we want to be using. The point of being concerned with supply chain security is that this catalog should be updated carefully and deliberately, not by just taking the latest version of everything. Instead, the version of a dependency should be updated only once some confidence has been gained that this is a good version. The second script would be one that we run in this case, updating the version of one dependency in `.github/.reversemap.yml`.

BTW, the `.github` directory should only contain stuff defined and understood by GitHub. Our catalog of good dependency versions should live elsewhere.

rxinui:
@MikeSpreitzer your new request has been issued on a seperate PR https://github.com/kubestellar/kubestellar/pull/2893

MikeSpreitzer:
This has been superseded by #2895, right?

MikeSpreitzer:
Closing this because it is superseded by #2895.

PR #2892, opened 2025-05-03
🌱  Move kubectl Dockerfile up to RHEL 9

MikeSpreitzer:
## Summary
This PR changes the base image from which the kubectl image is constructed from RHEL 8 to RHEL 9. This is good because it picks up many fixes.

This PR also changes where the kubectl binaries are fetched from, to the more comprehensive source documented in https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/#install-kubectl-binary-with-curl-on-linux

This PR also removes a duplicate tag flag from the docker build command.

## Related issue(s)

Fixes #


MikeSpreitzer:
I can not test exactly this because I am not authorized to push to quay.io/kubestellar. However, I was able to hack the registry to ghcr.io and test that, and succeeded.

PR #2893, opened 2025-05-03
🐛 Bug/2844-1 Supply chain security in CI

### LINKED ISSUE #2844: bug: GitHub actions not pinned to hash

rxinui:
## Summary

As per of [the following discussion](https://github.com/kubestellar/kubestellar/pull/2883#issuecomment-2848030049), a rework has been done to enhance the functionnality of the reverse map and its script.

Update includes:
- Usage of a the script as a CLI with `apply`, `update` and `sync` operations.
- Reverse map contains `sha`, `tag`, `url-sha` and `url-tag` as a KV for each actions

```yaml
actions/checkout:
  sha: 11bd71901bbe5b1630ceea73d27597364c9af683
  sha-url: |-
    https://github.com/actions/checkout/commit/11bd71901bbe5b1630ceea73d27597364c9af683
  tag: v4.2.2
  tag-url: https://github.com/actions/checkout/tree/v4.2.2
actions/add-to-project:
  sha: 244f685bbc3b7adfa8466e08b698b5577571133e
  sha-url: |-
    https://github.com/actions/add-to-project/commit/244f685bbc3b7adfa8466e08b698b5577571133e
  tag: v1.0.2
  tag-url: https://github.com/actions/add-to-project/tree/v1.0.2
# ...
```

### Enhancement proposal (new issue)

Make use of this hack script as part of CI to:
- test to ensure that no actions has been used with a tag or branch as version. 
- update on approval an action version in its reverse map and workflows right from GitHub.

If part of the CI, perhaps creating `.github/scripts/` would be idiomatic as GitHub recognises this directory as scripts for CI.

## Related issue(s)

Fixes #2844


.github/workflows/elapsed-time-collector.yml
```diff
@@ -84,4 +84,4 @@ jobs:
           echo "cat elapsed-time/${{ matrix.manifest }}.dmd"
           cat elapsed-time/${{ matrix.manifest }}.dmd
         
-      - run: git config --global user.email "kubestellar@kubestellar-dev@google.groups.com" && git config --global user.name "KubeStellar" && git add . && git commit -m "Add changes" && git push origin gh-pages || true
+      - run: git config --global user.email "kubestellar@null
```
MikeSpreitzer, line 87: Why this change?

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,260 @@
...
+ERR_YQ_NOT_INSTALLED=60
+ERR_GITHUB_TOKEN_INVALID=70
+ERR_ARCH_UNSUPPORTED=80
+
+help() {
+    cat <<EOF
+Harmonise '$GITHUB_WORKFLOWS_PATH' and the reversemap '$REVERSEMAP_FILE'
+
+Usage:
+    $0  <operation>
```
MikeSpreitzer, line 41: In the case of operation=update, more appears on the command line.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,260 @@
...
+        exit 2
+        ;;
+    esac
+
+}
+
+# Execute the main program
+main() {
+    if [[ -z $GITHUB_TOKEN ]]; then
+        _error_exit $ERR_GITHUB_TOKEN_INVALID "no github pat"
```
MikeSpreitzer, line 250: I think it would be better to give a more information in the error message, perhaps something like "environment variable GITHUB_TOKEN is not set" or "environment variable GITHUB_TOKEN needs to hold a token that is authorized to XXX" (and, by the way, what DOES this token need to be authorized to do?).

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,260 @@
...
+    operation_arg=$1
+    case $operation_arg in
+    "apply")
+        _loginfo "running $1"
+        run_apply
+        ;;
+    "update")
+        dependency_arg=$2
+        if [[ -z $dependency_arg ]]; then
+            _error_exit -1 "missing dependency name to update."
```
MikeSpreitzer, line 230: Should this be a negative number?

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,260 @@
...
+        fi
+        _loginfo "running $1"
```
MikeSpreitzer, line 232: I would log the dependency_arg too.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,260 @@
...
+}
+
+# Get commitsha from an action ref upstream
+_fetch_sha_from_upstream_ref() {
+    action_ref=$1
+    tag_or_branch=$2
+    action_ref_safe=$(echo "$action_ref" | cut -d '/' -f 1,2)
+    API_GITHUB_BRANCH="https://api.github.com/repos/${action_ref_safe}/git/refs/heads/${tag_or_branch}"
+    API_GITHUB_TAG="https://api.github.com/repos/${action_ref_safe}/git/refs/tags/${tag_or_branch}"
+    HTTP_STATUS=$(curl -o /dev/null -s -w "%{http_code}\n" -H "Authorization: Bearer $GITHUB_TOKEN" "$API_GITHUB_TAG")
```
MikeSpreitzer, line 124: FYI, I think that you do not need to newline in that output format.

MikeSpreitzer, line 124: FYI, you could reduce the number of `curl` invocations here from 3 to 2 by combining `-o /tmp/$$.json` and `-w %{http_code}` in one.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,260 @@
...
+    API_GITHUB_LATEST_RELEASE=https://api.github.com/repos/$action_ref/releases/latest
+    _return "$(curl -s -H "Authorization: Bearer $GITHUB_TOKEN" "$API_GITHUB_LATEST_RELEASE" | jq -r '.tag_name')"
+}
+
+# Update action version within a file
+_update_action_version_infile() {
+    file=$1
+    action_ref=$2
+    action_sha=$3
+    sed -E -i '' "s;$action_ref@.+;$action_ref@$action_sha;g" "$file"
```
MikeSpreitzer, line 175: You have only tested this on MacOS, right? On Ubuntu, it fails:

```console
mspreitz@mjs-dev7a:~/go/src/github.com/kubestellar/kubestellar$ sed -i '' s/the/das/ README.md 
sed: can't read s/the/das/: No such file or directory
```

BTW, I do not know a command that works on both Ubuntu and MacOS. I have solved this by testing whether `sed --help` output includes "-i[SUFFIX]" and then using one command line or the other.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,260 @@
...
+    action_tag=$2
+    action_sha=$3
+    action_ref_safe=$(echo "$action_ref" | cut -d '/' -f 1,2) # In case the action is a sub-action of a repo (ie. user/action/subaction@v1)
+    yq ".${action_ref}.sha = \"${action_sha}\"" -i $REVERSEMAP_FILE
+    yq ".${action_ref}.sha-url = \"https://github.com/${action_ref_safe}/commit/${action_sha}\"" -i $REVERSEMAP_FILE
+    yq ".${action_ref}.tag = \"${action_tag}\"" -i $REVERSEMAP_FILE
+    yq ".${action_ref}.tag-url = \"https://github.com/${action_ref_safe}/tree/${action_tag}\"" -i $REVERSEMAP_FILE
+}
+
+# Synchronise reverse map values with an action full ref (ie. actions/checkout@v4.2.0)
```
MikeSpreitzer, line 147: The current wording of this comment is not very illuminating for me. I recommend (a) explaining the arguments to the command and (b) explaining that the function is to iterate over all the "uses: actionref@actionver" lines and take the actionver, if it is not already a SHA, as the version whose SHA should be written into the reverse map file.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,260 @@
...
+# Run operation update
+run_update() {
+    dependency_arg=$1
+    _loginfo "updating dependency '$dependency_arg' tag to latest version available inside reverse map '$REVERSEMAP_FILE'"
+    latest_tag=$(_fetch_latest_tag "$dependency_arg")
+    action_sha=$(_fetch_sha_from_upstream_ref "$dependency_arg" "$latest_tag")
+    _yq_update_reversemap "$dependency_arg" "$latest_tag" "$action_sha"
+}
+
+# Run operation sync
```
MikeSpreitzer, line 211: I would use a little more expansive comment, perhaps something like "updates the reverse map file based on the tags/branches currently referenced in all the workflows".

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,260 @@
...
+            _loginfo "found $action_ref in reversemap with sha=$action_sha"
+            _update_action_version_infile "$file" "$action_ref" "$action_sha"
+        done
+    done
+}
+
+# Run operation update
+run_update() {
+    dependency_arg=$1
+    _loginfo "updating dependency '$dependency_arg' tag to latest version available inside reverse map '$REVERSEMAP_FILE'"
```
MikeSpreitzer, line 205: The description printed here does not match what the commands do. They fetch the latest regular (i.e., not "pre-release") release's tag and put it and its SHA in the reverse map file.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,260 @@
...
+}
+
+# Update reverse map KV using yq
+_yq_update_reversemap() {
+    action_ref=$1
+    action_tag=$2
+    action_sha=$3
+    action_ref_safe=$(echo "$action_ref" | cut -d '/' -f 1,2) # In case the action is a sub-action of a repo (ie. user/action/subaction@v1)
+    yq ".${action_ref}.sha = \"${action_sha}\"" -i $REVERSEMAP_FILE
+    yq ".${action_ref}.sha-url = \"https://github.com/${action_ref_safe}/commit/${action_sha}\"" -i $REVERSEMAP_FILE
```
MikeSpreitzer, line 142: Why write the sha-url into the file? I do not see it being used for anything. It bears a constant relationship with other information in the file.

Same question for the tag-url.

.gha-reversemap.yml
```diff
@@ -0,0 +1,89 @@
...
+  tag-url: https://github.com/actions/add-to-project/tree/v1.0.2
+titoportas/update-project-fields:
+  sha: 421a54430b3cdc9eefd8f14f9ce0142ab7678751
+  sha-url: |-
+    https://github.com/titoportas/update-project-fields/commit/421a54430b3cdc9eefd8f14f9ce0142ab7678751
+  tag: v0.1.0
+  tag-url: https://github.com/titoportas/update-project-fields/tree/v0.1.0
+ScholliYT/Broken-Links-Crawler-Action:
+  sha: a074e6a0c83f29560300752af773c7356041ceb4
+  sha-url: |-
```
MikeSpreitzer, line 21: Why is `|-` used for some actions and not for others?

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,260 @@
...
+Usage:
+    $0  <operation>
+
+Example:
+    # Update actions/checkout to its latest release
+    $0  update actions/checkout
+
+Operations:
+    apply                 -   Apply the commit sha of every actions listed in the reversemap as the version to use in all github workflows
+    update <action-ref>   -   Update the version of the given action reference to its latest release within the reverse map (sha, tag, urls)
```
MikeSpreitzer, line 49: The English grammar here is a bit awkward: the qualifier "within the reverse map" comes a long way after the thing it is modifying ("Update the version of the given action"); normally a qualifier immediately follows the thing that it is qualifying. Perhaps something more like the following?

> In the reverse map file update the tag and SHA of the given action reference to its latest release.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,260 @@
...
+        _error_exit "$query"
+    fi
+    _return "$query"
+}
+
+# Run operation apply
+run_apply() {
+    for file in "$GITHUB_WORKFLOWS_PATH"/*.yml; do
+        _loginfo "applying '$REVERSEMAP_FILE' commit sha to be used in '$file' ..."
+        action_refs=$(grep -E -o "$RE_ACTION_REF_AT" "$file")
```
MikeSpreitzer, line 192: The use of `grep` here is surprisingly coarse. Wouldn't it be better to use `yq` to look for `uses` fields in tasks?

.github/workflows/elapsed-time-collector.yml
rxinui: I made a huge mistake on the refactoring. I deleted a part of the regex. I'll fix it. Thank you for the review, that would be have  been terrible

hack/gha-reversemap.sh
rxinui: Unauthenticated requests are limited to 60 requests per hour per IP address. Therefore, even tho one execution might do the job, there will be cases where you want to run-it again (ie. quick fix on the reverse map before updating workflows). It is better to be authenticated IMO. 

On a CI side, if this script is executed part of the workflows as I mentioned, the usage of `secrets.GITHUB_TOKEN` can be done.

hack/gha-reversemap.sh
rxinui: I can put a define error instead of 255 (which is -1).

hack/gha-reversemap.sh
rxinui: On it

hack/gha-reversemap.sh
rxinui: Thank you. I was worried about the compatibility between these two OS.

After verification, there the difference is the usage of `-i ''` for MacOS/BSD and its equivalent on Linux is `-i`

hack/gha-reversemap.sh
rxinui: You are right but I did struggle to find better name. Would you recommend naming operations?

I'm willing to fully change them.

hack/gha-reversemap.sh
rxinui: So, this is purely informational and quite handy. It allows us from the reverse map to have a quick look onto the commit or the tag used. 

If somehow, you have a doubt on a commit sha used in a workflow, we can use the 'sha-url' in the reversemap to quickly access the action's github, read about the changes or just ensure the comformity of it.

I believe it is a great addition.

.gha-reversemap.yml
rxinui: I did not, this is purely `yq` doing

hack/gha-reversemap.sh
rxinui: Will do.

hack/gha-reversemap.sh
rxinui: `grep` just seemed handy over a query as it enable the usage of regexp over a file.

A `yq` could be used of course, but it would look as follow:

```bash
yq '.jobs[].steps[] | select(.uses) | .uses' $file
```

Both works. For the sake of comformity, i will replace `grep` to `yq` query.

.github/workflows/elapsed-time-collector.yml
rxinui: This is now fixed by replacing `grep` by `yq`

hack/gha-reversemap.sh
rxinui: I have refactored the usage to be more verbose and explicit.

I also change the operations names. Since it is not really a CLI but a Script with a seldom usage, i chose to use multiple words separated by a `-`.

```
Operations:
    apply-reversemap                     -   Apply the commit sha of every actions listed in the reversemap as the version to use in all github workflows
    update-action-version ACTION-REF     -   Update the version of the given action reference to its latest release within the reverse map (sha, tag, urls)
    update-reversemap                    -   Sync the reverse map values (sha, tag, urls) from the actions used in all github workflows

Arguments:
    update-action-version:
        ACTION-REF:   {gh_owner}/{gh_repo}    [Required]
```

hack/gha-reversemap.sh
rxinui: Fixed

hack/gha-reversemap.sh
rxinui: A good example is when the tag store is relative (ie. `v4`), its commit will be pointing to a specific semantic tag (for instance `v4.1.0`) and over time `v4` might points to `v4.2.4`. 

Knowing that, the usage of `tag-url` can be a handy shortcut to quickly compare the commit from the tag and the stored `sha`. It could help to decide whether we should update the value in the reversemap.

MikeSpreitzer:
I am confused. This PR asks to merge the same branch as PR #2883 .

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+Usage:
+    $0  <operation> [ARGUMENTS]
+
+Example:
+    # Update actions/checkout to its latest release
+    $0  update actions/checkout
+
+Operations:
+    apply-reversemap        [...WORKFLOW_FILE]          -   Apply the commit sha of every actions listed in the reversemap 
+                                                            as the version to use in all github workflows
```
MikeSpreitzer, line 47: "... or the given one(s)" ?

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+        exit 1
+        ;;
+    esac
+
+}
+
+# Execute the main program
+main() {
+    if [[ -z $GITHUB_TOKEN ]]; then
+        _exit_with_error $ERR_GITHUB_TOKEN_INVALID "environment variable GITHUB_TOKEN is not set."
```
MikeSpreitzer, line 281: This seems like an unnecessary restriction. One or more invocations could be done as an unauthenticated client, depending on the details of the invocations. Not a blocker.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+    fi
+    _check_yq_version
+    if [[ $? -eq $ERR_YQ_NOT_INSTALLED ]]; then
+        _install_yq
```
MikeSpreitzer, line 285: The automatic installation of `yq` slips in a requirement for password-less `sudo`. Not everybody does that. It should at least be explained to a potential user.

More importantly, overwriting the `yq` that the user already has installed is an undocumented surprise and exceptionally pushy. What if the user has a more recent version of `yq` installed, for their own reasons, and it is adequate for this script? Or even, what if the user has an older version of `yq` installed because they need it for some reason? I think that it would be better if this script fails if it finds that an unusable version of `yq` is already installed.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+
+Usage:
+    $0  <operation> [ARGUMENTS]
+
+Example:
+    # Update actions/checkout to its latest release
+    $0  update actions/checkout
+
+Operations:
+    apply-reversemap        [...WORKFLOW_FILE]          -   Apply the commit sha of every actions listed in the reversemap
```
MikeSpreitzer, line 46: (A) The usual way to write this would be `apply-reversemap [WORKFLOW_FILE...]` . I.e., put the ellipsis after the argument name.

(B) s/every actions/every action/

(C) the description says "all github workflows" but that's just one case. You could rewrite the description to something like the following.

> Update the given workflow files with the information in the reversemap file; if no workflow files are given then all are updated.

or

> Update the given workflow files, or all if none are given, with the information in the reversemap file.

Analogous comments apply to update-reversemap.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+
+Example:
+    # Update actions/checkout to its latest release
+    $0  update actions/checkout
+
+Operations:
+    apply-reversemap        [...WORKFLOW_FILE]          -   Apply the commit sha of every actions listed in the reversemap 
+                                                            as the version to use in all github workflows
+
+    update-action-version   ACTION_REF [...ACTION_REF]  -   Update the version of the given action reference within the reversemap
```
MikeSpreitzer, line 49: I think that `update-action-version ACTION_REF...` would get the message across.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+                                                            (sha, tag, urls) to its latest regular release tag
+
+    update-reversemap       [...WORKFLOW_FILE]          -   Update the reverse map values (sha, tag, urls) from the 
+                                                            actions' version used in the github workflows
+
+Arguments:
```
MikeSpreitzer, line 55: This section is partly redundant with the material above. What is new here is the explanation of the syntax of the two kinds of arguments. I would pare this section down to focus on just adding that information. Alternatively, I think that this section could be eliminated. I think that it will be obvious enough to a reader what a WORKFLOW_FILE is. BTW, I see no reason to restrict the filename to end in ".yml". GitHub accepts ".yaml" too. The syntax of ACTION_REF is germane to only one subcommand, and could be explained in the description of that subcommand.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+
+    echo "yq $YQ_REQUIRED_VERSION installed successfully."
+}
+
+# Get commitsha from an action ref upstream
+_fetch_sha_from_upstream_ref() {
+    action_ref=$1
+    tag_or_branch=$2
+    action_ref_safe=$(echo "$action_ref" | cut -d '/' -f 1,2)
+    TMP_OUTPUT=/tmp/output.json
```
MikeSpreitzer, line 131: I would pick a filename that (a) is not so generic that someone else might be using it and (b) differs in every invocation of this script (nothing stops a user from making two concurrent invocations of this script).

hack/gha-reversemap.sh
MikeSpreitzer: OK

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+    API_GITHUB_LATEST_RELEASE=https://api.github.com/repos/$action_ref/releases/latest
+    _return "$(curl -s -H "Authorization: Bearer $GITHUB_TOKEN" "$API_GITHUB_LATEST_RELEASE" | jq -r '.tag_name')"
+}
+
+# Update action version within a file
+_update_action_version_infile() {
+    file=$1
+    action_ref=$2
+    action_sha=$3
+    sed_replace_expr="s;(uses:) $action_ref@[a-zA-Z0-9 \._\-]+;\1 $action_ref@$action_sha ;g"
```
MikeSpreitzer, line 185: Why the backslashes in the bracket expression? Each one just stands for itself, why two?

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+        sed -E -i "$sed_replace_expr" "$file"
+    fi
+}
+
+# Get commit sha of an action ref stored in the reverse map
+_get_sha_from_reversemap() {
+    action_ref=$1
+    query=$(yq ".\"$action_ref\".sha" $REVERSEMAP_FILE)
+    if [[ $? -ne 0 ]]; then
+        _exit_with_error "$query"
```
MikeSpreitzer, line 199: A little more explanation in the error message would be good.

hack/gha-reversemap.sh
MikeSpreitzer: You asked a question then marked this as resolved, not sure what the two mean together.

The current names are OK with me.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+    fi
+    _return "$query"
+}
+
+# Apply the commit sha of every actions listed in the reversemap as the version to use in all github workflows
+run_apply_reversemap() {
+    files=$@
+    for file in $files; do
+        _loginfo "applying '$REVERSEMAP_FILE' commit sha to be used in '$file' ..."
+        action_fullrefs=$(yq '.jobs[].steps[] | select(.uses) | .uses' "$file")
```
MikeSpreitzer, line 209: FYI, this differs from the `select` on line 160. Not really a problem.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+_update_action_version_infile() {
+    file=$1
+    action_ref=$2
+    action_sha=$3
+    sed_replace_expr="s;(uses:) $action_ref@[a-zA-Z0-9 \._\-]+;\1 $action_ref@$action_sha ;g"
+    if [[ "$(uname -s)" = "Darwin" ]]; then
+        # MacOS sed usage of -i option differs from Linux version.
+        sed -E -i '' "$sed_replace_expr" "$file"
+    else
+        sed -E -i "$sed_replace_expr" "$file"
```
MikeSpreitzer, line 190: Using `yq` here rather than `sed` would be more robust. Not a blocker.

.gha-reversemap.yml
```diff
@@ -0,0 +1,85 @@
...
+  tag-url: https://github.com/actions/add-to-project/tree/v1.0.2
+titoportas/update-project-fields:
+  sha: 421a54430b3cdc9eefd8f14f9ce0142ab7678751
+  sha-url: https://github.com/titoportas/update-project-fields/commit/421a54430b3cdc9eefd8f14f9ce0142ab7678751
+  tag: v0.1.0
+  tag-url: https://github.com/titoportas/update-project-fields/tree/v0.1.0
+ScholliYT/Broken-Links-Crawler-Action:
+  sha: a074e6a0c83f29560300752af773c7356041ceb4
+  sha-url: https://github.com/ScholliYT/Broken-Links-Crawler-Action/commit/a074e6a0c83f29560300752af773c7356041ceb4
+  tag: fix-http-redirects
```
MikeSpreitzer, line 19: This is not the latest release. OK to bump in a different PR.

.gha-reversemap.yml
```diff
@@ -0,0 +1,85 @@
...
+  sha-url: https://github.com/anchore/sbom-action/commit/9f7302141466aa6482940f15371237e9d9f4c34a
+  tag: v0.19.0
+  tag-url: https://github.com/anchore/sbom-action/tree/v0.19.0
+goreleaser/goreleaser-action:
+  sha: 9c156ee8a17a598857849441385a2041ef570552
+  sha-url: https://github.com/goreleaser/goreleaser-action/commit/9c156ee8a17a598857849441385a2041ef570552
+  tag: v6
+  tag-url: https://github.com/goreleaser/goreleaser-action/tree/v6
+azure/setup-helm:
+  sha: 37dd2562cae2f186a3dc13f2bb5d4abe08dbfec4
```
MikeSpreitzer, line 47: Where did this hash come from? That is not what I see at GitHub now.

image

.gha-reversemap.yml
```diff
@@ -0,0 +1,85 @@
...
+  sha-url: https://github.com/azure/setup-helm/commit/37dd2562cae2f186a3dc13f2bb5d4abe08dbfec4
+  tag: v4
+  tag-url: https://github.com/azure/setup-helm/tree/v4
+docker/login-action:
+  sha: 74a5d142397b4f367a81961eba4e8cd7edddf772
+  sha-url: https://github.com/docker/login-action/commit/74a5d142397b4f367a81961eba4e8cd7edddf772
+  tag: v3
+  tag-url: https://github.com/docker/login-action/tree/v3
+actions/first-interaction:
+  sha: fb2402657b4a28582200150d0a145671d0e50597
```
MikeSpreitzer, line 57: I do not see this hash listed at https://github.com/actions/first-interaction/tags

hack/gha-reversemap.sh
rxinui: I disagree. In this very case it seems better to read the file from top to bottom.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+        help
+        exit 1
+        ;;
+    esac
+
+}
+
+# Execute the main program
+main() {
+    if [[ -z $GITHUB_TOKEN ]]; then
```
MikeSpreitzer, line 280: This should not prevent the `help` subcommand from working.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+        _exit_with_error $ERR_GITHUB_TOKEN_INVALID "environment variable GITHUB_TOKEN is not set."
+    fi
+    _check_yq_version
+    if [[ $? -eq $ERR_YQ_NOT_INSTALLED ]]; then
```
MikeSpreitzer, line 284: This is not relevant to the `help` subcommand (which could even provide a clue about the `yq` dependency).

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,290 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Purpose: Harmonise '$GITHUB_WORKFLOWS_PATH' and the reversemap '$REVERSEMAP_FILE'
+
+# Usage: see help function
+# Working directory must be the project root
+
+# Define the required version of yq
+GITHUB_TOKEN=
```
MikeSpreitzer, line 23: This does not play nice together with the check at the start of `main()`.

rxinui: So what do you want me to do?

hack/gha-reversemap.sh
MikeSpreitzer: I am not sure I understand. `yq` reads the whole file.

hack/gha-reversemap.sh
MikeSpreitzer: Remove one or the other.

MikeSpreitzer: Actually, I think this line should definitely go. The user should be able to supply a GITHUB_TOKEN.

hack/gha-reversemap.sh
rxinui: Regarding the space, im very aware but this is specific to the bracket within sed. Without the space, the expression is not interpreted not even [a-zA-Z0-9]. Regarding the dot, it was to be safe, i think it depending on the regexp library it behave differently.

hack/gha-reversemap.sh
rxinui: That is a hassle to correct and it is also not realistic to´that the script is invoked concurrently. If someone wants to break or find a way to break a program, he will find it. But that is not the responsibility of us to ensure that edge case, especially that none of the regulars contributors will use this hack. This hack, again interact with the CI, probably nobody aside the maintainers will know about it, hence only use by them.

Morever, this will be probably integrated to a CI to ensure the supply chain security.

hack/gha-reversemap.sh
```diff
@@ -0,0 +1,255 @@
...
+}
+
+# Update reverse map KV using yq
+_yq_update_reversemap() {
+    action_ref=$1
+    action_tag=$2
+    action_sha=$3
+    action_ref_safe=$(echo "$action_ref" | cut -d '/' -f 1,2) # In case the action is a sub-action of a repo (ie. user/action/subaction@v1)
+    yq ".${action_ref}.sha = \"${action_sha}\"" -i $REVERSEMAP_FILE
+    yq ".${action_ref}.sha-url = \"https://github.com/${action_ref_safe}/commit/${action_sha}\"" -i $REVERSEMAP_FILE
```
MikeSpreitzer, line 120: Wouldn't this field be more appropriately named "commit-url"?
Not a blocker.

rxinui: I prefer sha, personal preference. Gotta leave my touch to it.

.gha-reversemap.yml
rxinui: At first I thought it was a mistake from the action but now that the same issue occurs on Azure, this is definitely on my side.

One side effect I could see is when the same action is used in different workflows with different tag. In such cases, it should take the latest tag matched. However, the tag and its commit should be updated...

hack/gha-reversemap.sh
MikeSpreitzer: Not a blocker.

.gha-reversemap.yml
MikeSpreitzer: If I understand correctly, you are talking about the case where the update-reversemap operation when invoked on multiple workflow files that use different tags of the same action. I am willing to let that be addressed in a later PR, if at all.

hack/gha-reversemap.sh
MikeSpreitzer: OK

.gha-reversemap.yml
rxinui: Ok good! Let's do it like that. I'm on my phone at the moment. Will correct it in 1h

MikeSpreitzer:
Last commit was e36f1fc3925906dad3efa911a25b049a0c88f687

MikeSpreitzer:
PR #2895 was opened instead of this one, in hopes of avoiding the wedge wrt pull-kubestellar-kubestellar-validate-prow-yaml. But the new PR got wedged in the same way.

PR #2895, opened 2025-05-05
🐛 implementation of the script as CLI to solve Supply Chain security on CI

### LINKED ISSUE #2844: bug: GitHub actions not pinned to hash

rxinui:
## Summary

As per of [the following discussion](https://github.com/kubestellar/kubestellar/pull/2883#issuecomment-2848030049), a rework has been done to enhance the functionnality of the reverse map and its script.

**Goal**: move away of tag/branch as actions' version. Usage of commit sha instead to avoid supply chain attack.

Update includes:
- Usage of the script as a CLI with `update/reversemap`, `update-action-version` and `apply-reversemap` operations.
- Reverse map contains `sha`, `tag`, `url-sha` and `url-tag` as a KV for each actions

### Enhancement proposal (new issue)

Make use of this hack script as part of CI to:
- test to ensure that no actions has been used with a tag or branch as version. 
- update on approval an action version in its reverse map and workflows right from GitHub.

If part of the CI, perhaps creating `.github/scripts/` would be idiomatic as GitHub recognises this directory as scripts for CI.

## Related issue(s)

Fixes #2844

### DIFF
--- a/.gha-reversemap.yml
+++ b/.gha-reversemap.yml
@@ -1,0 +1,85 @@
+actions/checkout:
+  sha: 11bd71901bbe5b1630ceea73d27597364c9af683
+  sha-url: https://github.com/actions/checkout/commit/11bd71901bbe5b1630ceea73d27597364c9af683
+  tag: v4.2.2
+  tag-url: https://github.com/actions/checkout/tree/v4.2.2
+actions/add-to-project:
+  sha: 244f685bbc3b7adfa8466e08b698b5577571133e
+  sha-url: https://github.com/actions/add-to-project/commit/244f685bbc3b7adfa8466e08b698b5577571133e
+  tag: v1.0.2
+  tag-url: https://github.com/actions/add-to-project/tree/v1.0.2
+titoportas/update-project-fields:
+  sha: 421a54430b3cdc9eefd8f14f9ce0142ab7678751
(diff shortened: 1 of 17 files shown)


.gha-reversemap.yml
```diff
@@ -0,0 +1,85 @@
+  actions/checkout:
```
MikeSpreitzer, line 1: why the two leading spaces here?

hack/gha-reversemap.sh
MikeSpreitzer, line 1: This script should `set -e` so that failsures cause an immediate exit.

.gha-reversemap.yml
rxinui: fixed it

hack/gha-reversemap.sh
rxinui: done

rxinui:
@MikeSpreitzer ready to be merged :D

MikeSpreitzer:
close and re-open to see if it will help clear pull-kubestellar-kubestellar-validate-prow-yaml

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

MikeSpreitzer:
merging manually because tide is broken.

PR #2903, opened 2025-05-09
✨ Liberalize yq version check, bump setup-go to 5.5.0

MikeSpreitzer:
## Summary
This PR makes two changes. One is to liberalize the check that the `hack/gha-reversemap.sh` script makes on the version of yq, so that any v4.45.X is accepted. This is good because 4.45.2 was recently released.

The other change is to pick up the latest release of the setup-go action.

This will supersede #2902 

## Related issue(s)

Fixes #


MikeSpreitzer:
/cc @clubanderson 
/cc @francostellari 
/cc @pdettori 
@rxinui

PR #2916, opened 2025-05-14
✨   ci: automate usage gha-reversemap.sh

### LINKED ISSUE #2844: bug: GitHub actions not pinned to hash

rxinui:
## Summary

To accelerate the adoption of `hack/gha-reversemap.sh` to control and fix github actions version, I propose the following workflow which triggers the script through Github Actions UI.

When the command selected is executed, a PR will be raised which updated version. Reviewers (@clubanderson and @MikeSpreitzer) would either approve or deny the change.

## Related issue(s)

Fixes #2844

### DIFF
--- a/.github/workflows/sync-actions.yml
+++ b/.github/workflows/sync-actions.yml
@@ -1,0 +1,46 @@
+name: sync-actions
+on:
+  workflow_dispatch:
+    inputs:
+      cmd:
+        description: "specify the command of gha-reversemap.sh to execute"
+        required: true
+        type: choice
+        options:
+        - apply-reversemap
+        - update-action-version
+        - update-reversemap
+      args:
+        description: "arguments to selected command"
+        required: true
+        default: ""
+        type: string
+jobs:
+  trigger-gha-reversemap:
+    # only collaborators can trigger the workflow
+    if: contains(github.event.repository.collaborators, github.actor)
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    defaults:
+      run:
+        working-directory: ./
+    steps:
+    - name: Checkout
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
I agree that something needs to be done, thanks for starting on this.

Regardless of anything else, this new convention and mechanism needs to be documented in the contributor's section of the website.

I think that a simpler --- in the sense of requiring less privilege --- and more comprehensive solution is to add to CI a check for consistency. That would be a check that `hack/gha-reversemap.sh apply-reversemap` makes no change. Then it would be up to any PR that proposes a change to make a consistent change.

rxinui:
I think, because it is a CI and it is on workflow_dispatch event, it must be merged in the default branch (main). 

Therefore, let me do some testing on my forked repository and eventually shared the action result. I think I might missed something on it either way. 

@MikeSpreitzer

MikeSpreitzer:
I still think that the better approach, due to simplicity, reduced privileges, and better coverage, is to have CI check for consistency and require each PR to maintain consistency.

rxinui:
That can be done. Then should I close this PR? As to enable a CI check has nothing to do with `gha-reversemap.sh` instead rely on a linter.

MikeSpreitzer:
I think that the way to do a consistency check is to invoke `hack/gha-reversemap.sh apply-reversemap` and see if it makes a change (save a copy of the files before, compare with files after).

You could either change the branch in this PR or make a new PR. Up to you.

MikeSpreitzer:
Closing this PR, since the work has been done in other PRs. The latest increment is in #3016 .

PR #2918, opened 2025-05-16
🐛  Make clusteradm build get some later dependencies

MikeSpreitzer:
## Summary
This PR updates the script for building the clusteradm container image so that it gets later versions of some dependencies, picking up security fixes.

I have rebuilt quay.io/kubestellar/clusteradm:0.10.1 and quay.io/kubestellar/clusteradm:0.11.0 with the revised script.

## Related issue(s)

Fixes #


pdettori (commented):
/lgtm

PR #2920, opened 2025-05-16
✨  Bump kubeflex to 0.8.8

MikeSpreitzer:
## Summary
This PR bumps the KubeFlex dependency to 0.8.8, to pick up the actual improvement to the reference to the kube-rbac-proxy.

## Related issue(s)

Fixes #

PR #2921, opened 2025-05-16
✨  Prepare for test release 0.28.0-alpha.1

MikeSpreitzer:
## Summary
This PR updates the self-references in preparation for test release 0.28.0-alpha.1

This is an alpha rather than an RC because it does not have the intent to limit further development before the next regular release.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-prep4-0280a1/readme/

This is built on PR #2920, and will be rebased if that one merges first.

## Related issue(s)

Fixes #


pdettori (commented):
/lgtm

PR #2938, opened 2025-05-25
📖 More refined output added in the core-chart.md file

Rupam-It:
## Related issue(s)

Fixes #
As of now core-chart.md file only consist the command !
PR: 
1. added the output ( this shows what one can see after successful installation)
2. proper index at the top of the page

PR #2940, opened 2025-05-26
📖 Argocd docs update with Output

Rupam-It:
## Related issue(s)

Fixes #
PR update : 
1. add a proper index in this page 
2. add output of the argocd intregation!


Rupam-It:
@francostellari , @MikeSpreitzer 
please take a look!

clubanderson:
@francostellari can you comment/approve/merge?

docs/content/direct/core-chart-argocd.md
```diff
@@ -25,8 +34,139 @@ Note that when creating a local **Kubernetes** cluster using our scripts for **K
...
+
+kflex ctx --overwrite-existing-context its1
+kflex ctx --overwrite-existing-context wds1
+
+Finally, you can use `kflex ctx` to switch back to the kubeconfig
+context for your KubeFlex hosting cluster.
+
+Access Argo CD UI at https://argocd.localtest.me (append :9443 for Kind or k3s installations).
+Obtain Argo CD admin password using the command:
+kubectl -n kubeflex-system get secret argocd-initial-admin-secret -o jsonpath="{.data.password}" | base64 -d
```
francostellari, line 101: I guess the core-chart README file is also currently wrong. Can you please double check the location of the secret and fix this file as well?

docs/content/direct/core-chart-argocd.md
```diff
@@ -25,8 +34,139 @@ Note that when creating a local **Kubernetes** cluster using our scripts for **K
...
+**Output:**
+```
+Switched to context "kind-kubeflex".
+```
+
+### Step 1: Upgrade KubeStellar with Argo CD
+
+```bash
+helm upgrade --install ks-core ./core-chart \
+  --namespace kubeflex-system \
```
francostellari, line 61: In out instructions we do not specify the namespace to be `kubeflex-system`. I think this is causing some of the issues with namespaces

francostellari:
Can we add a couple of screen shots of the resulting ArgoCD UI with the app?

francostellari:
Additionally, currently kubestellar controller does not return the status of the resources correctly, so the deployed resources do not show as ready/green in Argo. It may be beneficial to discuss this for new users.

francostellari:
@Rupam-It thank you!
/lgtm

MikeSpreitzer:
For documentation PRs, please generate a website preview.

docs/content/direct/core-chart-argocd.md
```diff
@@ -1,11 +1,21 @@
...
+
+## Overview
+
 This documents explains how to use the KubeStellar core Helm chart to:
 
 - deploy Argo CD in KubeFlex hosting cluster;
 - register every WDS as a target cluster in Argo CD; and
 - create Argo CD applications as specified by the chart values.
 
+For a detailed step-by-step installation guide with expected outputs, see [Step-by-Step Installation Guide](core-chart.md).
```
MikeSpreitzer, line 17: I am going hold my comments on this for now, because #2938 is in concurrent progress.

docs/content/direct/core-chart-argocd.md
```diff
@@ -22,12 +32,97 @@ When deploying in a **Kubernetes** cluster, use the flag `--set argocd.global.do
...
+
+```bash
+kubectl get secret argocd-initial-admin-secret -o jsonpath="{.data.password}" | base64 -d
+```
+
+**Expected output (similar to):**
+```
+SYttSMZy6TQ2-alF
+```
+
```
MikeSpreitzer, line 94: There should be words or a section header here, saying something about what is coming next.

docs/content/direct/core-chart-argocd.md
```diff
@@ -22,12 +32,97 @@ When deploying in a **Kubernetes** cluster, use the flag `--set argocd.global.do
...
+**Access Argo CD UI:**
+
+Open your browser and navigate to: `https://argocd.localtest.me:9443/`
+
+- **Username**: `admin`
+- **Password**: Use the password obtained from the previous command (e.g., `SYttSMZy6TQ2-alF`)
+
+> **Note:** If you encounter SSL certificate warnings in your browser, proceed with "Advanced" → "Proceed to argocd.localtest.me (unsafe)" or similar option, as this is expected for local development setups.
```
MikeSpreitzer, line 122: I think that it would be good to be more specific about which browser warning this is trying to tell the reader to brush past --- it is possible that a user has something else go wrong! This is specifically about the warning about self-signed server certificate, right?

docs/content/direct/core-chart-argocd.md
```diff
@@ -22,12 +32,97 @@ When deploying in a **Kubernetes** cluster, use the flag `--set argocd.global.do
```
MikeSpreitzer, line 40: Now that we are slowing down and saying more in this section and the next, I think the words should be clear about the relationship between the two. The start of this section should clearly say something like the following.

> The KubeStellar core chart can optionally be used to install ArgoCD in the KubeFlex hosting cluster and register each KubeStellar WDS as an ArgoCD target cluster. The core chart also has the option to define some ArgoCD Applications. This section will cover installing ArgoCD and mapping WDSes to target clusters; the next section will show how to also define some Applications.

MikeSpreitzer:
I have created a website preview, at https://mikespreitzer.github.io/kcp-edge-mc/doc-pr2940-merge/direct/core-chart-argocd/

docs/content/direct/core-chart-argocd.md
```diff
@@ -22,12 +32,97 @@ When deploying in a **Kubernetes** cluster, use the flag `--set argocd.global.do
...
+**Expected output:**
+```
+Switched to context "kind-kubeflex".
+```
+
 The initial password for the `admin` user can be retrieved using the command:
 
 ```shell
 kubectl get secret argocd-initial-admin-secret -o jsonpath="{.data.password}" | base64 -d
 ```
```
MikeSpreitzer: Doesn't this come _after_ the installation?

MikeSpreitzer (commented):
I have finished a round of review.

Rupam-It:
@MikeSpreitzer , @francostellari please take a look!
https://rupam-it.github.io/kubestellar/argocd-docs/direct/core-chart-argocd/

docs/content/direct/core-chart-argocd.md
francostellari: @Rupam-It I think you should add @MikeSpreitzer statement

francostellari:
@Rupam-It I think that the new figures look good, thank you!

docs/content/direct/core-chart-argocd.md
Rupam-It: ok!

docs/content/direct/core-chart-argocd.md
Rupam-It: yes! so i change the unnecessary step and just after adding argocd , the very next step is to get the password and for then checking for the pods so that user can make sure that every pod is running (also I think if user got the password then every thing is fine , not mandatory need to check the pods)!

docs/content/direct/core-chart-argocd.md
Rupam-It: yes! I have already change the warning note, and now it will be more clear to user!

docs/content/direct/core-chart-argocd.md
MikeSpreitzer: Now that #2938 has merged there is something definite to compare with here.

The text "Step-by-Step Installation Guide" does not match the relative URL (core-chart.md). The latter refers to the whole document, not just the step-by-step part.

docs/content/direct/core-chart-argocd.md
MikeSpreitzer: There can be certificate problems other than self-signed certificates. If I understand correctly, this text is trying to tell the reader to push past the warning about a self-signed certificate. If I have this right then a more specific note is warranted.

docs/content/direct/core-chart-argocd.md
```diff
@@ -1,16 +1,34 @@
...
+- All prerequisites from [installing KubeStellar using the Core chart](core-chart.md#pre-requisites)
+- A properly configured KubeFlex hosting cluster
+- Helm installed and configured
+- kubectl access to your cluster
 
-The settings described in this document are an extesnion of the KubeStellar Core chart settings described [here](core-chart.md#kubestellar-core-chart-values).
+The settings described in this document are an extension of the KubeStellar Core chart settings described [here](core-chart.md#kubestellar-core-chart-values).
+
+> The KubeStellar core chart can optionally be used to install Argo CD in the KubeFlex hosting cluster and register each KubeStellar WDS as an Argo CD target cluster. The core chart also has the option to define some Argo CD Applications.  
+> This section will cover installing Argo CD and mapping WDSes to target clusters; the next section will show how to also define some Applications.
```
MikeSpreitzer, line 31: This text does not belong under the section heading "Pre-requisites", the subject is different.

MikeSpreitzer:
I updated the preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-pr2940-merge/direct/core-chart-argocd/

docs/content/direct/core-chart-argocd.md
```diff
@@ -22,12 +40,88 @@ When deploying in a **Kubernetes** cluster, use the flag `--set argocd.global.do
...
+
+### Retrieve Argo CD admin password
+
+The initial password for the `admin` user can be retrieved using the following command:
+
+```bash
+kubectl -n default get secret argocd-initial-admin-secret -o jsonpath="{.data.password}" | base64 -d
+```
+
+**Expected output(similar):**
```
MikeSpreitzer, line 90: The English wording here is a bit awkward (same for similar other lines). A few suggested alternatives:
- Expect output similar to the following
- Example output

docs/content/direct/core-chart-argocd.md
```diff
@@ -50,3 +144,6 @@ Alternatively, the same result can be achieved from Helm CLI by using the follow
 ```shell
 --set-json='argocd.applications=[ { "name": "scenario-6", "repoURL": "https://github.com/pdettori/sample-apps.git", "path": "nginx", "destinationWDS": "wds1", "destinationNamespace": "nginx-sa" } ]'
 ```
+
+![alt text](images/argocd-application.png)
```
MikeSpreitzer, line 148: You're supposed to put the alternate text for the image where I currently see the literal `alt text`.

Issue #2506, opened 2024-10-17
feature: expand regular testing


MikeSpreitzer:
### Feature Description

The current testing only exercises one deployment configuration: three new `kind` clusters, with fresh new KubeFlex ControlPlanes for WDS and ITS. We think that KubeStellar works in more configurations, and they should be tested.

- Things other than `kind` providing Kubernetes clusters.
- The WDS and/or ITS role played by the hosting cluster.
- The WDS and/or ITS role played by some other pre-existing cluster.
- One new ControlPlane playing both ITS and WDS roles.

### Proposed Solution

Do something. Without expanding the gating CI tests too much (note: I am not asserting exactly how much is too much).

### Want to contribute?

- [ ] I would like to work on this issue.

harsv688:
Hey can I pick this up ?

clubanderson:
Please do. Thank you

Pardhasardhiraob29:
/assign
Hi, maintainers!  
I’m new to open-source contributions and would love to work on this issue.  
Could you please assign it to me? Thanks!

clubanderson:
Welcome. Go right ahead and PR.  Thank you @Pardhasardhiraob29

Pardhasardhiraob29:
Hi @clubanderson, @MikeSpreitzer,

Apologies for the late response—I've been busy with exams. I'm excited to get started on this issue now and will be active in contributing from here on.

Could you please guide me on how to get started, or if there are any specific steps I should take to begin? I want to make sure I'm on the right track and aligned with the team's expectations.

Thanks for your patience, and I’m looking forward to collaborating!

Best,
Pardhasardhiraob29

blazethunderstorm:
@clubanderson @MikeSpreitzer @francostellari is it open to work

blazethunderstorm:
@MikeSpreitzer can i work on this pls

MikeSpreitzer:
PR #2945, which partially addressed this Issue, has been closed.

MikeSpreitzer:
@blazethunderstorm: You can work on this if you want to. You would need to re-assign it to yourself.

PR #2945, opened 2025-05-27
✨ add infrastructure for testing multiple configurations

AritraDey-Dev:
:sparkles: ✨ feature

This PR enhances the [create-kubestellar-demo-env.sh](https://probable-fortnight-pjjq5qr9qgr4h7pgx.github.dev/) script to support more flexible test environments. It adds options to:
-   Use a single ControlPlane for both ITS and WDS roles (--single-cp)
-   Use existing clusters/contexts for WDS and/or ITS (--wds-context, --its-context)
-   Output the correct environment variable hints based on the provided options

These improvements enable testing and demoing `KubeStellar` in a wider variety of deployment scenarios, including with pre-existing clusters and more compact control plane setups.
Related issue(s)

Partially addresses #2506

### DIFF
--- a/test/e2e/common/setup-kubestellar.sh
+++ b/test/e2e/common/setup-kubestellar.sh
@@ -93,20 +93,16 @@ source "$COMMON_SRCS/setup-shell.sh"
 source "$COMMON_SRCS/test-configs.sh"
 
 # Get scenario configuration
-scenario_config=( $(get_scenario_config "$TEST_SCENARIO") )
+set_scenario_config "$TEST_SCENARIO"
 if [ $? -ne 0 ]; then
     echo "Invalid scenario: $TEST_SCENARIO" >&2
     list_scenarios
     exit 1
 fi
 
-# Instead of associative array, parse scenario_config into variables
-config_name="${scenario_config[0]}"
-config_description="${scenario_config[1]}"
-config_wds_type="${scenario_config[2]}"
-config_its_type="${scenario_config[3]}"
-config_cluster_source="${scenario_config[4]}"
-config_combined_control_plane="${scenario_config[5]}"
+# Variables are now set directly by set_scenario_config:
+# config_name, config_description, config_wds_type, config_its_type, 
+# config_cluster_source, config_combined_control_plane
 
 :
 : -------------------------------------------------------------------------
--- a/test/e2e/common/test-configs.sh
+++ b/test/e2e/common/test-configs.sh
@@ -1,5 +1,3 @@
-# shellcheck shell=bash
-# test/e2e/common/test-configs.sh
 # Copyright 2024 The KubeStellar Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -13,26 +11,51 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-
-get_scenario_config() {
+set_scenario_config() {
   case "$1" in
     default)
-      # name description wds_type its_type cluster_source combined_control_plane
-      echo "default 'Three new kind clusters' new new kind false"
+      config_name="default"
+      config_description="Three new kind clusters"
+      config_wds_type="new"
+      config_its_type="new"
+      config_cluster_source="kind"
+      config_combined_control_plane="false"
       ;;
     wds-on-host)
-      echo "wds-on-host 'WDS on hosting cluster, new ITS' hosting new kind false"
+      config_name="wds-on-host"
+      config_description="WDS on hosting cluster, new ITS"
+      config_wds_type="hosting"
+      config_its_type="new"
+      config_cluster_source="kind"
+      config_combined_control_plane="false"
       ;;
     its-on-host)
-      echo "its-on-host 'ITS on hosting cluster, new WDS' new hosting kind false"
+      config_name="its-on-host"
+      config_description="ITS on hosting cluster, new WDS"
+      config_wds_type="new"
+      config_its_type="hosting"
+      config_cluster_source="kind"
+      config_combined_control_plane="false"
       ;;
     combined)
-      echo "combined 'Single control plane for both WDS and ITS' new new kind true"
+      config_name="combined"
+      config_description="Single control plane for both WDS and ITS"
+      config_wds_type="new"
+      config_its_type="new"
+      config_cluster_source="kind"
+      config_combined_control_plane="true"
       ;;
     *)
       echo "Unknown scenario: $1" >&2
       return 1
       ;;
   esac
+}
+
+list_scenarios() {
+  echo "Available test scenarios:"
+  echo "  default      - Three new kind clusters"
+  echo "  wds-on-host  - WDS on hosting cluster, new ITS"
+  echo "  its-on-host  - ITS on hosting cluster, new WDS"
+  echo "  combined     - Single control plane for both WDS and ITS"
 } 


AritraDey-Dev:
@MikeSpreitzer Thank you for your clarification. As I am new to this project, I’m currently working on familiarizing myself with the codebase. I have made the requested changes in this PR, and would greatly appreciate your feedback.

AritraDey-Dev:
@MikeSpreitzer any thoughts on this PR?

test/e2e/run-test.sh
```diff
@@ -21,11 +21,12 @@ set -x # so users can see what is going on
...
 test="bash"
 fail_flag=""
-
+setup_flags=""
+existing_wec_contexts=""
 
 while [ $# != 0 ]; do
     case "$1" in
-        (-h|--help) echo "$0 usage: (--released | --env | --test-type | --kubestellar-controller-manager-verbosity \$num | --transport-controller-verbosity \$num | --fail-fast)*"
+        (-h|--help) echo "$0 usage: (--released | --env $kind_or_ocp | --test-type $bash_or_ginkgo | --kubestellar-controller-manager-verbosity $num | --transport-controller-verbosity $num | --fail-fast | --existing-wec-contexts $context1,$context2,...)*"
```
MikeSpreitzer: Those backslashes are needed. Compare

```console
bash-3.2$ echo "$0 usage: (--released | --env $kind_or_ocp | --test-type $bash_or_ginkgo | --kubestellar-controller-manager-verbosity $num | --transport-controller-verbosity $num | --fail-fast | --existing-wec-contexts $context1,$context2,...)*"
bash usage: (--released | --env  | --test-type  | --kubestellar-controller-manager-verbosity  | --transport-controller-verbosity  | --fail-fast | --existing-wec-contexts ,,...)*
```

with

```console
bash-3.2$ echo "$0 usage: (--released | --env \$kind_or_ocp | --test-type \$bash_or_ginkgo | --kubestellar-controller-manager-verbosity \$num | --transport-controller-verbosity \$num | --fail-fast | --existing-wec-contexts \$context1,\$context2,...)*"
bash usage: (--released | --env $kind_or_ocp | --test-type $bash_or_ginkgo | --kubestellar-controller-manager-verbosity $num | --transport-controller-verbosity $num | --fail-fast | --existing-wec-contexts $context1,$context2,...)*
```

MikeSpreitzer: +1 on adding exposition of the possible values given to the flags.

test/e2e/run-test.sh
```diff
@@ -53,6 +54,14 @@ while [ $# != 0 ]; do
             exit 1;
           fi;;
         (--fail-fast) fail_flag="--fail-fast";;
+        (--existing-wec-contexts)
+          if (( $# > 1 )); then
+            existing_wec_contexts="$2"
```
MikeSpreitzer, line 73: Why this special handling instead of adding to `$setup_flags`?

test/e2e/run-test.sh
```diff
@@ -81,7 +90,7 @@ scripts_dir="${SRC_DIR}/../../scripts"
 
 "${COMMON_SRCS}/cleanup.sh" --env "$env"
 source "${COMMON_SRCS}/setup-shell.sh"
-"${COMMON_SRCS}/setup-kubestellar.sh" $setup_flags --env "$env"
+"${COMMON_SRCS}/setup-kubestellar.sh" $setup_flags --env "$env" $([ -n "$existing_wec_contexts" ] && echo "--existing-wec-contexts $existing_wec_contexts")
```
MikeSpreitzer: The `setup-kubestellar.sh` script does not accept an `--existing-wec-contexts` flag.

MikeSpreitzer:
Ping @AritraDey-Dev

test/e2e/run-test.sh
AritraDey-Dev: fixed in latest commit

test/e2e/common/setup-kubestellar.sh
```diff
@@ -83,6 +91,20 @@ SRC_DIR="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)"
...
+# Get scenario configuration
+scenario_config=($(get_scenario_config "$TEST_SCENARIO"))
+if [ $? -ne 0 ]; then
+    echo "Invalid scenario: $TEST_SCENARIO" >&2
+    list_scenarios
+    exit 1
+fi
+
+# Parse scenario configuration
+declare -A config
```
MikeSpreitzer: On my Mac:

```console
bash-3.2$ declare -A config
bash: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
```

MikeSpreitzer: Mac users are blighted with version 3 of bash, which does not support associative arrays. We should not be in the business of telling users to make a major version change in their system software.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -83,6 +91,20 @@ SRC_DIR="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)"
...
+for i in "${!scenario_config[@]}"; do
+    config[$i]="${scenario_config[$i]}"
+done
```
MikeSpreitzer: Why have two copies of the same associative array? Why not just have the code below consume the first copy?

test/e2e/common/setup-kubestellar.sh
```diff
@@ -83,6 +91,20 @@ SRC_DIR="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)"
 COMMON_SRCS="${SRC_DIR}/../common"
 source "$COMMON_SRCS/setup-shell.sh"
 
+# Get scenario configuration
+scenario_config=($(get_scenario_config "$TEST_SCENARIO"))
```
MikeSpreitzer: Where is `get_scenario_config` defined?

test/e2e/run-test.sh
```diff
@@ -21,13 +21,25 @@ set -x # so users can see what is going on
...
 test="bash"
 fail_flag=""
-
+setup_flags=""
+existing_wec_contexts=""
 
 while [ $# != 0 ]; do
     case "$1" in
-        (-h|--help) echo "$0 usage: (--released | --env | --test-type | --kubestellar-controller-manager-verbosity \$num | --transport-controller-verbosity \$num | --fail-fast)*"
+        (-h|--help) echo "$0 usage: (--released | --env \$kind_or_ocp | --test-type \$bash_or_ginkgo | --kubestellar-controller-manager-verbosity \$num | --transport-controller-verbosity \$num | --fail-fast | --existing-wec-contexts \$context1,\$context2,...)*"
```
MikeSpreitzer: Shouldn't the `--scenario` option be mentioned here?

MikeSpreitzer: BTW, since this is already using `(` and `|` to organize syntax options, I think that it would be simpler to write `(bash | ginkgo)` than both write `$bash_or_ginkgo` AND later explain that `$bash_or_ginkgo` means `bash` or `ginkgo`.

test/e2e/run-test.sh
```diff
@@ -21,13 +21,25 @@ set -x # so users can see what is going on
...
+                    echo "  \$num: integer verbosity level"
+                    echo "  \$context1,\$context2,...: comma-separated list of existing WEC contexts"
                     exit;;
-        (--released) setup_flags="$setup_flags $1";;
+        (--released|--scenario)
+          if [ "$1" = "--scenario" ] && (( $# > 1 )); then
+            setup_flags="$setup_flags $1 $2"
+            shift
+          else
+            setup_flags="$setup_flags $1"
```
MikeSpreitzer: `test/e2e/common/setup-kubestellar.sh` does not tolerate `--scenario` without an additional argument.

test/e2e/common/setup-kubestellar.sh
MikeSpreitzer: What are the possible scenarios? How/where are they defined?

test/e2e/common/setup-kubestellar.sh
AritraDey-Dev: i am in linux so didn't notice it.replaced with positional array.

test/e2e/common/setup-kubestellar.sh
AritraDey-Dev: defined in test.configs.sh

test/e2e/run-test.sh
AritraDey-Dev: yes thanks for noticing.updated it.

test/e2e/run-test.sh
```diff
@@ -21,13 +21,28 @@ set -x # so users can see what is going on
...
+setup_flags=""
+existing_wec_contexts=""
 
 while [ $# != 0 ]; do
     case "$1" in
-        (-h|--help) echo "$0 usage: (--released | --env | --test-type | --kubestellar-controller-manager-verbosity \$num | --transport-controller-verbosity \$num | --fail-fast)*"
+        (-h|--help) echo "$0 usage: (--released | --env (kind|ocp) | --test-type (bash|ginkgo) | --kubestellar-controller-manager-verbosity \$num | --transport-controller-verbosity \$num | --fail-fast | --existing-wec-contexts \$context1,\$context2,... | --scenario \$scenario_name)*"
+                    echo "where:"
+                    echo "  \$num: integer verbosity level"
+                    echo "  \$context1,\$context2,...: comma-separated list of existing WEC contexts"
```
MikeSpreitzer, line 32: IMO, these are obvious enough that you could save space by not explicitly saying them.

test/e2e/run-test.sh
```diff
@@ -21,13 +21,28 @@ set -x # so users can see what is going on
...
+                    echo "  \$scenario_name: scenario to use (default, wds-on-host, its-on-host, combined)"
```
MikeSpreitzer, line 33: Like the other enumerated types, this enumeration could be folded into the main syntax expression.

test/e2e/run-test.sh
```diff
@@ -53,6 +68,14 @@ while [ $# != 0 ]; do
             exit 1;
           fi;;
         (--fail-fast) fail_flag="--fail-fast";;
+        (--existing-wec-contexts)
+          if (( $# > 1 )); then
+            existing_wec_contexts="$2"
```
MikeSpreitzer, line 73: `existing_wec_contexts` is not used for anything. Why is it here?

test/e2e/common/setup-kubestellar.sh
```diff
@@ -82,6 +90,23 @@ fi
 SRC_DIR="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)"
 COMMON_SRCS="${SRC_DIR}/../common"
 source "$COMMON_SRCS/setup-shell.sh"
+source "$COMMON_SRCS/test-configs.sh"
+
+# Get scenario configuration
+scenario_config=( $(get_scenario_config "$TEST_SCENARIO") )
```
MikeSpreitzer: Indirecting through an array of strings is fragile. Why not just have the `case` over scenario directly assign to the variables (it could be `source`d)?

```console
bash-3.2$ a=b c=d
bash-3.2$ echo $a
b
bash-3.2$ echo $c
d
```

test/e2e/common/test-configs.sh
```diff
@@ -0,0 +1,38 @@
...
+      ;;
+    combined)
+      echo "combined 'Single control plane for both WDS and ITS' new new kind true"
+      ;;
+    *)
+      echo "Unknown scenario: $1" >&2
+      return 1
+      ;;
+  esac
+}
```
MikeSpreitzer, line 61: I see red here.
image

MikeSpreitzer:
This PR does not actually expand the testing done in CI, right? The existing title of this PR, "expand regular testing", suggests that CI testing is being expanded. Either fix the title or the functionality.

Available functionality should be regularly tested by CI.

The existing E2E test suite takes about half an hour to run. That is painfully long.

test/e2e/run-test.sh
AritraDey-Dev: yes will remove it

MikeSpreitzer:
https://prow2.kubestellar.io/view/s3/prow-logs/pr-logs/pull/kubestellar_kubestellar/2945/pull-kubestellar-verify/1941002794169274368 says

> 1 files have incorrect boilerplate headers:
> test/e2e/common/test-configs.sh

MikeSpreitzer:
Ping @AritraDey-Dev

test/e2e/common/test-configs.sh
AritraDey-Dev: fixed!

test/e2e/run-test.sh
MikeSpreitzer: But they're still here.

test/e2e/common/test-configs.sh
MikeSpreitzer: Did you forget to push? I still see red here. Have you ever seen the red circle-dash here?

image

test/e2e/common/test-configs.sh
```diff
@@ -0,0 +1,61 @@
...
+
+list_scenarios() {
+  echo "Available test scenarios:"
+  echo "  default      - Three new kind clusters"
+  echo "  wds-on-host  - WDS on hosting cluster, new ITS"
+  echo "  its-on-host  - ITS on hosting cluster, new WDS"
+  echo "  combined     - Single control plane for both WDS and ITS"
```
MikeSpreitzer, line 60: BTW, while not totally incorrect, calling these "scenarios" does not, in my opinion, do the best job of conveying what they are about. They are about configurations to test. I might call them "configurations" rather than "scenarios".

MikeSpreitzer, line 60: Did you test these?

test/e2e/common/setup-kubestellar.sh
```diff
@@ -98,19 +119,40 @@ case "$CLUSTER_SOURCE" in
...
+    WDS_CONFIG='[{"name":"combined-cp"}]'
+else
+    if [ "$config_wds_type" = "hosting" ]; then
+        WDS_CONFIG='[{"name":"wds1","hosting":true}]'
+    else
+        WDS_CONFIG='[{"name":"wds1"}]'
+    fi
+
+    if [ "$config_its_type" = "hosting" ]; then
+        ITS_CONFIG='[{"name":"its1","hosting":true}]'
```
MikeSpreitzer, line 144: If I recall correctly, `"hosting":true` is not the way to get this done. See https://github.com/kubestellar/kubestellar/blob/8b33007d864576b18379751433c4d8ce720ffa7d/core-chart/values.yaml#L108 .

MikeSpreitzer:
ping @AritraDey-Dev

MikeSpreitzer:
Ping @AritraDey-Dev

MikeSpreitzer:
Closed for lack of activity.

PR #2959, opened 2025-05-31
🐛  bug: fixed insufficient default rbac for syncing cluster scoped resources 

### LINKED ISSUE #2947: bug: Insufficient Default RBAC for Syncing Cluster-Scoped Resources like StorageClass
### Describe the bug

While working with KubeStellar to sync objects from the hub to edge clusters(WEC), I encountered permission issues when trying to deploy cluster-scoped resources, particularly StorageClass. It appears that the service account (klusterlet-work-sa) used by the OCM agent does not have the necessary RBAC permissions by default to handle certain resource types. The klusterlet-work-sa service account lacks the necessary RBAC permissions to get or apply StorageClass objects (and possibly other cluster-scoped resources) by default.

`'Failed to apply manifest: storageclasses.storage.k8s.io "local-storage" is forbidden: User "system:serviceaccount:open-cluster-management-agent:klusterlet-work-sa" cannot get resource "storageclasses" in API group "storage.k8s.io" at the cluster scope'
`

$ ./KubeStellar-Snapshot.sh
[kubestellar-snapshot.tar.gz](url)
# Collected state from the hub and WEC clusters

### Steps To Reproduce

1.Set up KubeStellar with the automated kind-based bootstrap script
2.Apply the following BindingPolicy to the hub cluster:
kind: BindingPolicy
metadata:
  name: hub-bpolicy
spec:
  clusterSelectors:
  - matchLabels: {"location-group": "edge"}
  downsync:
  - objectSelectors:
    - matchLabels: {"app":"ipp"}

3.Apply a StorageClass object to the hub:

`apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: local-storage
  labels: {app: "ipp"}
provisioner: kubernetes.io/no-provisioner
volumeBindingMode: WaitForFirstConsumer
`

4.Observe failure in manifestwork on WEC cluster:

`kubectl --context its1 -n cluster1 get manifestworks <manifest-name> -o yaml
`

### Expected Behavior

The StorageClass (or any cluster-scoped resource) should be successfully synced and applied on the WEC without manual RBAC modification, or at the very least, the need for RBAC extension should be clearly documented.

### Want to contribute?

### Additional Context

1.This is not a one-time bug, but rather a missing configuration in the default RBAC setup
2.A possible solution is either:
        -Update default RBAC in the setup script
        -Add documentation on required roles/permissions for syncing cluster-scoped resources

ritankarsaha:
## Summary

- Updated manager-role `ClusterRole` to grant full access to `storageclasses` in the `storage.k8s.io API group`.
- Updated the `ClusterRoleBinding` to bind manager-role to the controller-manager service account (not default).

The root cause for this issue according to me was missing RBAC rules for cluster-scoped resources. The solution was to update the ClusterRole to include those resources and ensure the correct service account is bound to that role, so the controller/agent can manage cluster-scoped resources like StorageClass without manual RBAC intervention.

## Related issue(s)

Fixes #2947

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -29,3 +29,15 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - "storage.k8s.io"
+  resources:
+  - storageclasses
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
--- a/config/rbac/role_binding.yaml
+++ b/config/rbac/role_binding.yaml
@@ -15,5 +15,5 @@ roleRef:
   name: manager-role
 subjects:
 - kind: ServiceAccount
-  name: default
+  name: controller-manager
   namespace: '{{"{{.Namespace}}"}}'


ritankarsaha:
@francostellari , @MikeSpreitzer
please take a look!

config/rbac/role.yaml
```diff
@@ -29,3 +29,15 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
```
francostellari, line 32: For these changes to be used by the Helm chart, they need to be implemented in https://github.com/kubestellar/kubestellar/blob/main/core-chart/templates/postcreatehooks/wds.yaml as well

config/rbac/role_binding.yaml
```diff
@@ -15,5 +15,5 @@ roleRef:
   name: manager-role
 subjects:
 - kind: ServiceAccount
-  name: default
+  name: controller-manager
```
francostellari, line 18: As you can see in https://github.com/kubestellar/kubestellar/blob/7a38c4896dd0b386e46fc47bcd41f1a4ce29e92f/config/rbac/kustomization.yaml#L7 and https://github.com/kubestellar/kubestellar/blob/main/config/manager/manager.yaml  the SA is not used. The controller-manager uses the default SA. The same is true in the core chart

config/rbac/role.yaml
ritankarsaha: @francostellari so i will be needing to add these changes to these files and make the updates in those as well right?

francostellari: yes, thank you

ritankarsaha:
@francostellari @clubanderson @MikeSpreitzer  will make the updates and push a new commit into the PR. do i need to rebase and squash the commits to one as well? ... was going through other PR works, found out a few rebases so?

MikeSpreitzer:
@ritankarsaha: GitHub does not think that you _need_ to rebase.

image

MikeSpreitzer:
This PR is small enough that keeping all of its changes in one commit looks like a good idea to me. You could either `git commit ... --amend` to replace the existing commit, or add a commit and then squash the two together (yes, I do that with `git rebase -i ...`).

ritankarsaha:
> This PR is small enough that keeping all of its changes in one commit looks like a good idea to me. You could either `git commit ... --amend` to replace the existing commit, or add a commit and then squash the two together (yes, I do that with `git rebase -i ...`).

Ahh yess @MikeSpreitzer , when i said `rebase and squash` I really just meant the good ol’ `git rebase -i trick` to squash commits like you do 😅, wasn’t planning on rebasing onto the base branch.

@francostellari suggested some updates, will be done with them by tomorrow, will squash both the commits and push !! Sorry for the delay, was away for a few days due to an emergency.

MikeSpreitzer:
I disagree that there is an implementation bug that needs to be fixed. See https://github.com/kubestellar/kubestellar/issues/2947#issuecomment-2979069743 .

ritankarsaha:
> I disagree that there is an implementation bug that needs to be fixed. See [#2947 (comment)](https://github.com/kubestellar/kubestellar/issues/2947#issuecomment-2979069743) .

Hey @MikeSpreitzer , took a look into the comment. 
I think you are right that granting elevated permissions by default may pose security concerns, especially in production setups. The lack of documentation around these permissions made it hard to diagnose why certain cluster-scoped resources (like StorageClass) were silently failing to sync to the edge clusters.
Then should I submit a documentation PR that clearly outlines the minimal RBAC permissions required for syncing cluster-scoped resources like StorageClass, with examples of Role/ClusterRole and binding strategies ? I think that should be the correct scope for this now ?

clubanderson:
At a minimum this must/should be documented. We can at least start with that.

ritankarsaha:
> At a minimum this must/should be documented. We can at least start with that.

okaay! , should I then work on this PR itself, or make this a draft one and open up a new PR for the documentation ? @clubanderson

clubanderson:
different PR is preferable - we try to keep PRs as small as possible to improve turn around time on review and merge. Thank you!

ritankarsaha:
> different PR is preferable - we try to keep PRs as small as possible to improve turn around time on review and merge. Thank you!

Suree !!

MikeSpreitzer:
Closing this one, in anticipation of work in other PR(s).
Don't forget that a PR that modifies the website should include a preview.

Issue #2932, opened 2025-05-22
Improve KubeStellar Installation Flow: Resource Ordering, Readiness, and Dependency Validation


Rupam-It:
## Current Pain Points
The current Helm-based installation flow for KubeStellar and its components (KubeFlex, ITS, WDS) has several issues:

1. No mechanism to wait for resource readiness (e.g., ControlPlane) before proceeding
2. Race conditions due to unordered or parallel resource creation
3. Silent failures when resources already exist (e.g., Namespace, CRDs)
4. No prerequisite validation for CRDs, CRs, or required dependencies
5. One-shot install – no retry/reconciliation for transient cluster delays
6. Poor feedback on dependency issues, making debugging hard
7. High manual setup effort (pre-create NS, CRDs, CRs)
8. Difficult to manage optional components (e.g., PostgreSQL, operators) based on cluster state

## Proposed Solution
Implement a lookup-based approach using Helm's built-in `lookup` function to:
- Validate resource existence and readiness
- Handle dependencies in the correct order
- Provide clear error messages
- Support optional components
- Implement retry logic for transient failures

Rupam-It:
proposed solution: https://github.com/Rupam-It/kubestellar/blob/custom-perator/custom-operator-proposal.md#2-core-reconciliation-algorithm
https://docs.google.com/document/d/1VVFrRg-np6S5EuLxX4tcEjFrPeoI1f3c0Zd2eSEgyYk/edit?usp=sharing

MikeSpreitzer:
No, the doc cited in the previous comment is not the proposal. The cited doc is the analysis of the problem.

The proposal is in https://github.com/Rupam-It/kubestellar/blob/custom-perator/custom-operator-proposal.md

MikeSpreitzer:
Regarding the analysis in https://docs.google.com/document/d/1VVFrRg-np6S5EuLxX4tcEjFrPeoI1f3c0Zd2eSEgyYk , I have the following comments/questions.

In "1. Reuse of Problematic InitContainers", the stated problem is "These containers require elevated cluster permissions and run custom bash logic during initialization, which introduces fragility and security risks". Does the proposed solution not require the same permissions and have the same security risks? Is the thinking that Go code is inherently less fragile than bash code?

In "2. Complex Bash Logic in ConfigMap", the displayed code is purely a convenience. It need not be used in production, or indeed ever. If we really do not like it at all then we can simply remove it.

[...]

PR #2966, opened 2025-06-10
DO NOT MERGE: review copy of proposal for Issue 2932

MikeSpreitzer:
## Summary
This PR exists to create a way to comment on the proposal in https://github.com/kubestellar/kubestellar/issues/2932#issuecomment-2957936769 . This PR only exists to create a way to attach comments to specific places in the proposal.

## Related issue(s)

Fixes #

### DIFF
--- a/custom-operator-proposal.md
+++ b/custom-operator-proposal.md
@@ -90,6 +90,12 @@ spec:
 ### 2. **Core Reconciliation Algorithm**
 ![alt text](image.png)
 
+**Types of Actions:**
+- **create** - Create new Kubernetes resources
+- **template** - Create resources using data from dependencies  
+- **patch** - Update existing resources
+- **delete** - Remove resources
+
 ### 3. **Dependency Checking Algorithm**
 
 ```go


custom-operator-proposal.md
```diff
@@ -0,0 +1,804 @@
...
+
+### 1. **Resource Definition & Priority Algorithm**
+
+```yaml
+apiVersion: tools.kubestellar.io/v1
+kind: ResourceDependency
+metadata:
+  name: kubestellar-bootstrap
+spec:
+  # Priority-based execution (lower number = higher priority)
```
MikeSpreitzer, line 39: What is "priority" used for? Don't all the waits have to complete?

I see "priority" in three places: there is `spec.priority`, `spec.dependencies[i].priority`, and `spec.actions[i].priority`. How do these three sorts of priority interact?

custom-operator-proposal.md
```diff
@@ -0,0 +1,804 @@
...
+  priority: 1
+  
+  # Dependencies to wait for
+  dependencies:
+    - resource: "namespace/kubestellar-system"
```
MikeSpreitzer, line 44: (A) It looks like you mean "object" rather than "resource". A "resource" is a kind of objects.

(B) I see some things not mentioned here: API group, namespace. What is the full syntax of these strings?

(C) Which space is this object expected to appear in?

(D) I am not sure I fully understand the intent here. After I ran `scripts/create-kubestellar-demo-env.sh`, I did not find a namespace named "kubestellar-system".

MikeSpreitzer:
/cc @francostellari 
@Rupam-It

custom-operator-proposal.md
```diff
@@ -0,0 +1,804 @@
...
+      timeout: "2m"
+      priority: 1
+    - resource: "deployment/kubeflex-controller-manager"
+      namespace: "kubeflex-system"
+      state: "Ready"
```
MikeSpreitzer, line 49: What does `state: "Ready"` mean? What is the full range of conditions that may be awaited upon?

```console
mspreitz@mjs13 ~ % kubectl --context kind-kubeflex get deploy -A
NAMESPACE            NAME                          READY   UP-TO-DATE   AVAILABLE   AGE
ingress-nginx        ingress-nginx-controller      1/1     1            1           2m30s
kube-system          coredns                       2/2     2            2           2m33s
kubeflex-system      kubeflex-controller-manager   0/1     1            0           43s
local-path-storage   local-path-provisioner        1/1     1            1           2m31s
```

custom-operator-proposal.md
```diff
@@ -0,0 +1,804 @@
...
+      minReplicas: 1
```
MikeSpreitzer, line 50: How does `minReplicas: 1` contribute?

custom-operator-proposal.md
```diff
@@ -0,0 +1,804 @@
...
+          timestamp: "{{ .Now }}"
+
+    - priority: 2
+      type: "template"
+      template: |
+        apiVersion: v1
+        kind: Secret
+        metadata:
+          name: kubestellar-bootstrap
+          namespace: {{ .Dependencies.namespace_kubestellar_system.Name }}
```
MikeSpreitzer, line 78: Isn't the value of this a constant, `kubestellar-system`?

custom-operator-proposal.md
```diff
@@ -0,0 +1,804 @@
...
+        patch: |
+          {"spec":{"template":{"metadata":{"annotations":{"kubectl.kubernetes.io/restartedAt":"{{ .Now }}"}}}}}}
+```
+
+### Example 2: ITS to WDS Dependency Chain
+
+#### **Current Pattern with InitContainers**
+
+```yaml
+# Current approach with initContainers
```
MikeSpreitzer, line 375: Where does this appear currently?

custom-operator-proposal.md
```diff
@@ -0,0 +1,804 @@
...
+        command: ['sh', '-c', 'until kubectl get controlplane its1 -o jsonpath="{.status.phase}" | grep Running; do echo "Waiting for ITS"; sleep 5; done']
+      - name: wait-for-ocm
+        image: bitnami/kubectl:1.25.8
+        command: ['sh', '-c', 'until kubectl get deployment -n its1 cluster-manager -o jsonpath="{.status.readyReplicas}" | grep 1; do echo "Waiting for OCM"; sleep 5; done']
+```
+
+#### **New Pattern with ResourceDependency**
+
+```yaml
+# New approach with ResourceDependency operator
```
MikeSpreitzer, line 396: There is one of these per what?

custom-operator-proposal.md
```diff
@@ -0,0 +1,804 @@
...
+    - priority: 2
+      type: "patch"
+      patch:
+        target:
+          kind: "Deployment" 
+          name: "wds-controller"
+          namespace: "kubestellar-system"
+        patchType: "strategic"
+        patch: |
+          {"spec":{"template":{"metadata":{"annotations":{"kubectl.kubernetes.io/restartedAt":"{{ .Now }}"}}}}}}
```
MikeSpreitzer, line 445: This annotation is not owned by KubeStellar and thus should not be written into by KubeStellar. Writing to this annotation does not cause a Deployment's Pod(s?) Container(s?) to be restarted. And if it did, I suppose that means that they were already running and logging complaints about stuff not being ready.

custom-operator-proposal.md
```diff
@@ -0,0 +1,804 @@
...
+### Example 6: Complete KubeStellar Deployment Chain
+
+```yaml
+apiVersion: tools.kubestellar.io/v1
+kind: DependencyGroup
+metadata:
+  name: kubestellar-full-deployment
+  namespace: kubestellar-system
+spec:
+  groups:
```
MikeSpreitzer, line 655: How does a singular group contain "groups"?

custom-operator-proposal.md
```diff
@@ -0,0 +1,804 @@
...
+    # Phase 1: Infrastructure
+    - name: "core-infrastructure"
+      priority: 1
+      dependencies:
+        - name: "kubestellar-namespaces"
+          resource: "namespace/kubestellar-system"
```
MikeSpreitzer, line 661: This is _not_ a reference to a ResourceDependency object, right? Doesn't that mean that there is no way to say a bunch of needed stuff here?

What does a DependencyGroup object do that a collection of ResourceDependency objects does not do?

image.png
MikeSpreitzer, line 1: In step 6, why requeue on success? If the awaited things have appeared, and the actions done, why do we want anything else to happen?

MikeSpreitzer:
This seems a lot more complicated than necessary. I am not clear on why we want anything more than a couple of synchronous executables to be used in initContainers:
- one that simply waits for some condition, not crashing if objects involved do not exist yet (and better yet, upstream this into `kubectl wait`);
- a `kflex` command that creates the desired kubeconfig file, waiting for inputs as necessary.

Or better yet:
- make all the controllers wait for whatever they need, without alarming log messages;
- define a plugin mechanism by which the transport controller can get its kubeconfigs, and implement plugins that wait as necessary and then produce the kubeconfig.

PR #2974, opened 2025-06-13
🌱 other/misc: add log for better understanding of docker context error!

Rupam-It:
## Summary
```
KO_DOCKER_REPO=ko.local ko build -B ./cmd/controller-manager -t c0f945577 --platform linux/amd64
2025/06/13 12:15:21 Using base cgr.dev/chainguard/static:latest@sha256:797e62f43d04d792e9f930913e7d9f5a63e92bd19ca5e7e5139a692decee2dbc for github.com/kubestellar/kubestellar/cmd/controller-manager
2025/06/13 12:15:22 Building github.com/kubestellar/kubestellar/cmd/controller-manager for linux/amd64
2025/06/13 12:15:45 Loading ko.local/controller-manager:35529aa9721bd8be3f09f3daaa2d6185ef93d83b71190e9432d5de1ef31d4ef3
2025/06/13 12:15:46 Loaded ko.local/controller-manager:35529aa9721bd8be3f09f3daaa2d6185ef93d83b71190e9432d5de1ef31d4ef3
2025/06/13 12:15:46 Adding tag c0f945577
2025/06/13 12:15:46 Added tag c0f945577
ko.local/controller-manager:35529aa9721bd8be3f09f3daaa2d6185ef93d83b71190e9432d5de1ef31d4ef3
docker tag ko.local/controller-manager:c0f945577 ghcr.io/kubestellar/kubestellar/controller-manager:c0f945577
Error response from daemon: No such image: ko.local/controller-manager:c0f945577
make: *** [Makefile:183: ko-build-controller-manager-local] Error 1

```

this error is not common and not every one face so, i have added a log if any one face that understand the error and solve that !
## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,17 +179,23 @@ run: manifests generate fmt vet ## Run a controller from your host.
 
 .PHONY: ko-build-controller-manager-local
 ko-build-controller-manager-local: ## Build local controller manager container image with ko
-	$(shell (docker version | { ! grep -qi podman; } ) || echo "DOCKER_HOST=unix://$$HOME/.local/share/containers/podman/machine/qemu/podman.sock ") KO_DOCKER_REPO=ko.local ko build -B ./cmd/${CONTROLLER_MANAGER_CMD_NAME} -t ${IMAGE_TAG} --platform linux/${ARCH}
-	@if docker image inspect ko.local/${CONTROLLER_MANAGER_CMD_NAME}:${IMAGE_TAG} > /dev/null 2>&1; then \
-		docker tag ko.local/${CONTROLLER_MANAGER_CMD_NAME}:${IMAGE_TAG} ${CONTROLLER_MANAGER_IMAGE}; \
-	else \
-		echo "[WARN] Failed to tag image. This might be due to a Docker context mismatch or missing tag '${IMAGE_TAG}'. Try setting: export DOCKER_HOST=unix:///home/$(shell whoami)/.docker/desktop/docker.sock"; \
+	@if ! docker images &> /dev/null; then \
+		echo "Error: docker command is not working. Please ensure you have a working docker setup."; \
+		echo "For podman users, you may need to set up docker compatibility or configure DOCKER_HOST."; \
+		exit 1; \
 	fi
+	KO_DOCKER_REPO=ko.local ko build -B ./cmd/${CONTROLLER_MANAGER_CMD_NAME} -t ${IMAGE_TAG} --platform linux/${ARCH}
+	docker tag ko.local/${CONTROLLER_MANAGER_CMD_NAME}:${IMAGE_TAG} ${CONTROLLER_MANAGER_IMAGE}
 
 
 .PHONY: ko-build-transport-local
 ko-build-transport-local: ## Build local transport container image with `ko`.
-	$(shell (docker version | { ! grep -qi podman; } ) || echo "DOCKER_HOST=unix://$$HOME/.local/share/containers/podman/machine/qemu/podman.sock ") KO_DOCKER_REPO=ko.local ko build -B ./pkg/transport/${TRANSPORT_CMD_NAME} -t ${IMAGE_TAG} --platform linux/${ARCH}
+	@if ! docker images &> /dev/null; then \
+		echo "Error: docker command is not working. Please ensure you have a working docker setup."; \
+		echo "For podman users, you may need to set up docker compatibility or configure DOCKER_HOST."; \
+		exit 1; \
+	fi
+	KO_DOCKER_REPO=ko.local ko build -B ./pkg/transport/${TRANSPORT_CMD_NAME} -t ${IMAGE_TAG} --platform linux/${ARCH}
 	docker tag ko.local/${TRANSPORT_CMD_NAME}:${IMAGE_TAG} ${TRANSPORT_IMAGE}
 
 # this is used for local testing
--- a/Makefile
+++ b/Makefile
@@ -179,34 +179,11 @@ run: manifests generate fmt vet ## Run a controller from your host.
 
 .PHONY: ko-build-controller-manager-local
 ko-build-controller-manager-local: ## Build local controller manager container image with ko
-	@echo "Building controller manager image..."
-	@# Try standard ko build first (case 1: simple Docker setup)
-	@if KO_DOCKER_REPO=ko.local ko build -B ./cmd/${CONTROLLER_MANAGER_CMD_NAME} -t ${IMAGE_TAG} --platform linux/${ARCH} > /dev/null 2>&1; then \
-		echo "Build successful with standard Docker setup"; \
-	elif docker version 2>/dev/null | grep -qi podman; then \
-		echo "Podman detected, retrying with Podman socket..."; \
-		if DOCKER_HOST=unix://$$HOME/.local/share/containers/podman/machine/qemu/podman.sock KO_DOCKER_REPO=ko.local ko build -B ./cmd/${CONTROLLER_MANAGER_CMD_NAME} -t ${IMAGE_TAG} --platform linux/${ARCH} > /dev/null 2>&1; then \
-			echo "Build successful with Podman setup"; \
-		else \
-			echo "[ERROR] Ko build failed even with Podman configuration"; \
-			exit 1; \
-		fi; \
-	else \
-		echo "[ERROR] Ko build failed with standard Docker setup"; \
-		echo "[WARN] This might be due to a Docker context mismatch. Try one of:"; \
-		echo "  - export DOCKER_HOST=unix:///home/$$(whoami)/.docker/desktop/docker.sock"; \
-		echo "  - docker context list (to see available contexts)"; \
-		echo "  - docker context use <context-name>"; \
-		exit 1; \
-	fi
-	@# Tag the image if it exists
+	$(shell (docker version | { ! grep -qi podman; } ) || echo "DOCKER_HOST=unix://$$HOME/.local/share/containers/podman/machine/qemu/podman.sock ") KO_DOCKER_REPO=ko.local ko build -B ./cmd/${CONTROLLER_MANAGER_CMD_NAME} -t ${IMAGE_TAG} --platform linux/${ARCH}
 	@if docker image inspect ko.local/${CONTROLLER_MANAGER_CMD_NAME}:${IMAGE_TAG} > /dev/null 2>&1; then \
 		docker tag ko.local/${CONTROLLER_MANAGER_CMD_NAME}:${IMAGE_TAG} ${CONTROLLER_MANAGER_IMAGE}; \
-		echo "Successfully tagged image as ${CONTROLLER_MANAGER_IMAGE}"; \
 	else \
-		echo "[WARN] Built image not found for tagging. Image may have been created with different tag."; \
-		echo "Available images:"; \
-		docker images | grep -E "(ko\.local|${CONTROLLER_MANAGER_CMD_NAME})" || echo "No matching images found"; \
+		echo "[WARN] Failed to tag image. This might be due to a Docker context mismatch or missing tag '${IMAGE_TAG}'. Try setting: export DOCKER_HOST=unix:///home/$(shell whoami)/.docker/desktop/docker.sock"; \
 	fi
 
 
--- a/Makefile
+++ b/Makefile
@@ -178,12 +178,35 @@ run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./cmd/controller-manager/main.go $(ARGS)
 
 .PHONY: ko-build-controller-manager-local
-ko-build-controller-manager-local:
-	@KO_DOCKER_REPO=ko.local ko build -B ./cmd/${CONTROLLER_MANAGER_CMD_NAME} -t ${IMAGE_TAG} --platform linux/${ARCH}
+ko-build-controller-manager-local: ## Build local controller manager container image with ko
+	@echo "Building controller manager image..."
+	@# Try standard ko build first (case 1: simple Docker setup)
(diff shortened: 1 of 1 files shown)


francostellari:
@Rupam-It can you please explain a  bit more your strategy here. I'm not very familiar with this particular problem.
The original fail was checking for docker vs podman and automatically setting the `DOCKER_HOST` variable.
It seems to me that your PR does more than add a useful user comment.
Also please show the new behavior in the same condition.

Rupam-It:
first thing is that there is podman in my system but it is not running !
so , when you check our code 
KO build an image !
by docker it try to tag it!

```
 docker context ls 
NAME              DESCRIPTION                               DOCKER ENDPOINT                                  ERROR
default           Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                      
desktop-linux *   Docker Desktop                            unix:///home/rupam/.docker/desktop/docker.sock  
```

now the thing is that 
-KO is doing things by default context 
-and docker command try to tag via desktop-linux!


that's why i have to export  the `docker_host = unix:///home/rupam/.docker/desktop/docker.sock `! then all things work . 
I think this is very rare issue, so i add a comment if there is issue 

```
ko.local/controller-manager:e666fd13cf736421be19e99efc6ef4fa5525cdf9f7adf782169de9c66bed64d9
[WARN] Failed to tag image. This might be due to a Docker context mismatch or missing tag '2a9993577'. Try setting: export DOCKER_HOST=unix:///home/rupam/.docker/desktop/docker.sock
KO_DOCKER_REPO=ko.local ko build -B ./pkg/transport/ocm-transport-controller -t 2a9993577 --platform linux/amd64
2025/06/13 20:48:46 Using base cgr.dev/chainguard/static:latest@sha256:797e62f43d04d792e9f930913e7d9f5a63e92bd19ca5e7e5139a692decee2dbc for github.com/kubestellar/kubestellar/pkg/transport/ocm-transport-controller
2025/06/13 20:48:47 Building github.com/kubestellar/kubestellar/pkg/transport/ocm-transport-controller for linux/amd64
2025/06/13 20:48:49 Loading ko.local/ocm-transport-controller:e90604218610bc846012895c86c50aa0d54251209757a90d0685d5341f35989b
2025/06/13 20:48:50 Loaded ko.local/ocm-transport-controller:e90604218610bc846012895c86c50aa0d54251209757a90d0685d5341f35989b
2025/06/13 20:48:50 Adding tag 2a9993577
2025/06/13 20:48:50 Added tag 2a9993577
ko.local/ocm-transport-controller:e90604218610bc846012895c86c50aa0d54251209757a90d0685d5341f35989b
docker tag ko.local/ocm-transport-controller:2a9993577 ghcr.io/kubestellar/kubestellar/ocm-transport-controller:2a9993577
Error response from daemon: No such image: ko.local/ocm-transport-controller:2a9993577
make: *** [Makefile:193: ko-build-transport-local] Error 1
```
so if some get error , will check the warn here else if every thing working good the warn won't come up!

francostellari:
@Rupam-It thank you for clarifying your problem and solution.

In my opinion I would change the workflow:
1. try KO build without checking `podman` or setting `DOCKER_HOST` ... should work in most simple cases
2. if fail: check if `podman` is installed/used and set the most likely `DOCKER_HOST` value and retry build ... should work for podman user
3. if none above work then print your WARN message and perhaps show extra wording if indeed more than one docker context is present

The original implementation worked for 1 & 2 but no working for 3
Your current implementation does not work for 2 but just create a working

What do you think?

francostellari:
@MikeSpreitzer do you have an opinion on the proposed workflow?

Rupam-It:
one thing you saw that 
[warn] consist my name but that is dynamic !
means will show the user name !

francostellari:
> one thing you saw that [warn] consist my name but that is dynamic ! means will show the user name !

Yes, understood

Rupam-It:
> The original implementation worked for 1 & 2 but no working for 3 Your current implementation does not work for 2 but just create a working
> 
> What do you think?

sorry sorry! missed it! adding that!

francostellari:
> first thing is that there is podman in my system but it is not running ! so , when you check our code KO build an image ! by docker it try to tag it!

To clarify, your `docker` command is an alias to `podman` but podman is not used?

I'm a bit puzzled by why the podman check is triggered in the Makefile in your scenario

Rupam-It:
> > first thing is that there is podman in my system but it is not running ! so , when you check our code KO build an image ! by docker it try to tag it!
> 
> To clarify, your `docker` command is an alias to `podman` but podman is not used?
> 
> I'm a bit puzzled by why the podman check is triggered in the Makefile in your scenario

I don't have a Podman issue. I have a Docker context mismatch issue. KO and my docker commands are using different Docker contexts, so the image gets built in one context but the tagging command looks for it in another context."

francostellari:
Additionally, that as I reported in slack channel, there is a potential bug where the `ko-local-build` is not listed as a dependency of `kind-load-image`... just wanted to make sure that your problem is not related to it.

https://kubernetes.slack.com/archives/C058SUSL5AA/p1749816831005229?thread_ts=1749752013.441509&cid=C058SUSL5AA

Rupam-It:
> Additionally, that as I reported in slack channel, there is a potential bug where the `ko-local-build` is not listed as a dependency of `kind-load-image`... just wanted to make sure that your problem is not related to it.
> 
> https://kubernetes.slack.com/archives/C058SUSL5AA/p1749816831005229?thread_ts=1749752013.441509&cid=C058SUSL5AA

ya , I saw that I think it is quite different issue! I'm looking onto that!



in simple word 
1. KO build a image as a root user, i can access the image by `sudo docker images`
2. but when it try to tag it use normal `docker tag` command obviously it does not have sudo/root access so it can't find the image and give an error!

Rupam-It:
> Additionally, that as I reported in slack channel, there is a potential bug where the `ko-local-build` is not listed as a dependency of `kind-load-image`... just wanted to make sure that your problem is not related to it.
> 
> https://kubernetes.slack.com/archives/C058SUSL5AA/p1749816831005229?thread_ts=1749752013.441509&cid=C058SUSL5AA

I have already seen this comments , going to open an issue for that!

but here the thing is that : 
1. KO is build a image as sudo user !
2. and try to tag is by a normal `docker tag` command ! that's why it is giving error to me !

MikeSpreitzer:
The significance of the displayed console output in the opening comment is not clear. Is it the result of issuing a `make` command? It would be so much clearer to include the command that produced the output.

Makefile
```diff
@@ -180,7 +180,12 @@ run: manifests generate fmt vet ## Run a controller from your host.
 .PHONY: ko-build-controller-manager-local
 ko-build-controller-manager-local: ## Build local controller manager container image with ko
 	$(shell (docker version | { ! grep -qi podman; } ) || echo "DOCKER_HOST=unix://$$HOME/.local/share/containers/podman/machine/qemu/podman.sock ") KO_DOCKER_REPO=ko.local ko build -B ./cmd/${CONTROLLER_MANAGER_CMD_NAME} -t ${IMAGE_TAG} --platform linux/${ARCH}
-	docker tag ko.local/${CONTROLLER_MANAGER_CMD_NAME}:${IMAGE_TAG} ${CONTROLLER_MANAGER_IMAGE}
+	@if docker image inspect ko.local/${CONTROLLER_MANAGER_CMD_NAME}:${IMAGE_TAG} > /dev/null 2>&1; then \
+		docker tag ko.local/${CONTROLLER_MANAGER_CMD_NAME}:${IMAGE_TAG} ${CONTROLLER_MANAGER_IMAGE}; \
+	else \
+		echo "[WARN] Failed to tag image. This might be due to a Docker context mismatch or missing tag '${IMAGE_TAG}'. Try setting: export DOCKER_HOST=unix:///home/$(shell whoami)/.docker/desktop/docker.sock"; \
```
MikeSpreitzer: I see no failure to tag image here. Instead, I see that tagging was not even attempted.

MikeSpreitzer:
Some of the discussion here seems to have slipped from talking about this repo to talking about the kubeflex repo or something older or hypothetical. There is no make target named `ko-build-local` in this repo's Makefile.

MikeSpreitzer:
In this repo's Makefile, the image load already depends on both image builds:

```
kind-load-image: ko-build-controller-manager-local ko-build-transport-local
```

(I am not actually a fan of requiring a fresh build every time an image is loaded, but that is a separate issue.)

MikeSpreitzer:
Please be careful about jumping to conclusions about the difference that "sudo" makes. We have seen other cases where what turned out to be important is simply the change to a different user, special privileges were not relevant.

MikeSpreitzer:
I am unable to reproduce this error. I have both podman and Rancher Desktop installed on my laptop. When I stop Rancher Desktop and `podman machine start`, and add a symlink that makes podman appear as `docker` in my `$PATH`, that is how is succeeded to build in years gone by. But today it fails like this:

```console
bash-3.2$ export PATH=$HOME/local/bin:$PATH

bash-3.2$ which docker
/Users/mspreitz/local/bin/docker

bash-3.2$ ls -l $(which docker)
lrwxr-xr-x  1 mspreitz  staff  24 Aug 24  2023 /Users/mspreitz/local/bin/docker -> /opt/homebrew/bin/podman

bash-3.2$ podman machine start
Starting machine "podman-machine-default"

This machine is currently configured in rootless mode. If your containers
require root permissions (e.g. ports < 1024), or if you run into compatibility
issues with non-podman clients, you can switch using the following command:

	podman machine set --rootful

API forwarding listening on: /var/folders/94/6n2mkxzn3px1dj6drjdtjvpc0000gn/T/podman/podman-machine-default-api.sock

The system helper service is not installed; the default Docker API socket
address can't be used by podman. If you would like to install it, run the following commands:

        sudo /opt/homebrew/Cellar/podman/5.5.1/bin/podman-mac-helper install
        podman machine stop; podman machine start

You can still connect Docker API clients by setting DOCKER_HOST using the
following command in your terminal session:

        export DOCKER_HOST='unix:///var/folders/94/6n2mkxzn3px1dj6drjdtjvpc0000gn/T/podman/podman-machine-default-api.sock'

Machine "podman-machine-default" started successfully

bash-3.2$ podman images
REPOSITORY              TAG         IMAGE ID      CREATED        SIZE
docker.io/kindest/node  <none>      1cf551538f7d  16 months ago  966 MB

bash-3.2$ docker images
REPOSITORY              TAG         IMAGE ID      CREATED        SIZE
docker.io/kindest/node  <none>      1cf551538f7d  16 months ago  966 MB

bash-3.2$ make ko-build-transport-local
DOCKER_HOST=unix:///Users/mspreitz/.local/share/containers/podman/machine/qemu/podman.sock  KO_DOCKER_REPO=ko.local ko build -B ./pkg/transport/ocm-transport-controller -t 352e8df04 --platform linux/arm64
2025/06/17 04:35:43 Using base cgr.dev/chainguard/static:latest@sha256:797e62f43d04d792e9f930913e7d9f5a63e92bd19ca5e7e5139a692decee2dbc for github.com/kubestellar/kubestellar/pkg/transport/ocm-transport-controller
2025/06/17 04:35:43 Building github.com/kubestellar/kubestellar/pkg/transport/ocm-transport-controller for linux/arm64
2025/06/17 04:35:48 Loading ko.local/ocm-transport-controller:a96a33dc4879fea17ae1309cdc3f88156452bc7a4c1c66d05e9c723559b42bbb
2025/06/17 04:35:48 daemon.Write response:  
Error: failed to publish images: error publishing ko://github.com/kubestellar/kubestellar/pkg/transport/ocm-transport-controller: error loading image: Cannot connect to the Docker daemon at unix:///Users/mspreitz/.local/share/containers/podman/machine/qemu/podman.sock. Is the docker daemon running?
make: *** [ko-build-transport-local] Error 1
```

However, if I just do as I am told, it succeeds:

```console
bash-3.2$ DOCKER_HOST='unix:///var/folders/94/6n2mkxzn3px1dj6drjdtjvpc0000gn/T/podman/podman-machine-default-api.sock' KO_DOCKER_REPO=ko.local ko build -B ./pkg/transport/ocm-transport-controller -t 352e8df04 --platform linux/arm64
2025/06/17 04:36:10 Using base cgr.dev/chainguard/static:latest@sha256:797e62f43d04d792e9f930913e7d9f5a63e92bd19ca5e7e5139a692decee2dbc for github.com/kubestellar/kubestellar/pkg/transport/ocm-transport-controller
2025/06/17 04:36:11 Building github.com/kubestellar/kubestellar/pkg/transport/ocm-transport-controller for linux/arm64
2025/06/17 04:36:12 Loading ko.local/ocm-transport-controller:a96a33dc4879fea17ae1309cdc3f88156452bc7a4c1c66d05e9c723559b42bbb
2025/06/17 04:36:14 Loaded ko.local/ocm-transport-controller:a96a33dc4879fea17ae1309cdc3f88156452bc7a4c1c66d05e9c723559b42bbb
2025/06/17 04:36:14 Adding tag 352e8df04
2025/06/17 04:36:14 Added tag 352e8df04
ko.local/ocm-transport-controller:a96a33dc4879fea17ae1309cdc3f88156452bc7a4c1c66d05e9c723559b42bbb

bash-3.2$ DOCKER_HOST='unix:///var/folders/94/6n2mkxzn3px1dj6drjdtjvpc0000gn/T/podman/podman-machine-default-api.sock' docker tag ko.local/ocm-transport-controller:352e8df04 ghcr.io/kubestellar/kubestellar/ocm-transport-controller:352e8df04

bash-3.2$ echo $?
0
```

MikeSpreitzer:
So I think that the simplest and best solution here may be to remove the DOCKER_HOST hackery from the Makefile and tell the user that they are responsible for having a working `docker` command. A user of podman will know how to do that (or at least, it should be their problem to know that). We could add a test of whether `docker images &> /dev/null` succeeds or fails, and in the failure case abort with an error message that the user does not have a working `docker` command.

Makefile
Rupam-It: Cause the image is created but docker can't access it ( as it does not have sudo privilege)!
That's why there is no it won't show any tag failure!

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

Issue #1844, opened 2024-03-02
bug: kubestellar controller-manager runs leader election and ignores it


MikeSpreitzer:
### Describe the bug

The kubestellar controller-manager engages leader election but does not wait to be elected before going into business. Tha is wrong.

### Steps To Reproduce

Look in the log from the KCM container and see something like the following excerpt near the start.

```
...
I0302 05:36:20.591816       1 controller.go:505] "Successfully synced" logger="Binding.worker-3" objectIdentifier={"GVK":{"Group":"apiextensions.k8s.io","Version":"v1","Kind":"CustomResourceDefinition"},"Resource":"customresourcedefinitions","ObjectName":{"Namespace":"","Name":"bindingpolicies.control.kubestellar.io"},"ResourceVersion":"236"}
I0302 05:36:20.591825       1 crd.go:142] "New API added. Starting informer for:" logger="Binding.worker-3" group="control.kubestellar.io" version="control.kubestellar.io/v1alpha1" kind="BindingPolicy"
I0302 05:36:23.377500       1 controller.go:147] "waiting for caches to sync" logger="Status"
I0302 05:36:23.377529       1 controller.go:161] "All caches synced" logger="Status"
I0302 05:36:23.377534       1 controller.go:163] "Starting workers" logger="Status" count=4
I0302 05:36:23.377538       1 controller.go:167] "Started workers" logger="Status"
I0302 05:36:23.377557       1 controller.go:187] "waiting for bindingpolicy cache to sync" logger="Status"
I0302 05:36:23.377559       1 controller.go:225] "waiting for workstatus cache to sync" logger="Status"
I0302 05:36:23.478378       1 controller.go:191] "bindingpolicy cache synced" logger="Status"
I0302 05:36:23.480571       1 controller.go:229] "workstatus cache synced" logger="Status"
I0302 05:36:26.380548       1 main.go:172] "starting manager" logger="setup"
I0302 05:36:26.381857       1 internal.go:360] "Starting server" kind="health probe" addr="[::]:8081"
I0302 05:36:26.381894       1 server.go:50] "starting server" path="/metrics" kind="metrics" addr="127.0.0.1:8080"
I0302 05:36:26.381915       1 leaderelection.go:250] attempting to acquire leader lease wds1-system/c6f71c85.kflex.kubestellar.org...
I0302 05:36:26.387006       1 leaderelection.go:260] successfully acquired lease wds1-system/c6f71c85.kflex.kubestellar.org
[...]

PR #2985, opened 2025-06-19
🐛bug: ensure controller-manager waits for leader election before starting controllers

antedotee:
## Summary
Fix bug where KubeStellar controller-manager would start controllers before becoming the leader, even when leader election was enabled.

## Problem
The controller-manager was engaging in leader election but ignoring it - controllers would start immediately after parsing flags, before the process had become the leader. This defeated the purpose of leader election and could lead to multiple active controllers.

## Solution
Implemented proper leader election logic using controller-runtime's built-in manager. Controllers now only start after the process becomes the leader when `--leader-elect=true`.

## Changes
- Added `startControllersWithLeaderElection()` function that uses controller-runtime manager
- Added `startControllersDirectly()` function for non-leader-election mode
- Modified main logic to choose between leader election and direct startup based on `--leader-elect` flag
- Made leader election namespace configurable based on WDS name (`wdsName + "-system"`)
- Preserved original behavior when `--leader-elect=false`

## Testing
- ✅ Verified leader election flag is correctly parsed
- ✅ Confirmed controllers don't start before leader election when `--leader-elect=true`
- ✅ Verified original behavior preserved when `--leader-elect=false`
- ✅ Built successfully with no compilation errors

## Impact
This fix ensures that when leader election is enabled, only one controller-manager instance will be active at a time, preventing potential conflicts and resource waste from multiple active controllers.

Fixes: #1844

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -20,12 +20,14 @@ package main
 // to ensure that exec-entrypoint and run can make use of them.
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"os"
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	"github.com/spf13/pflag"
 
 	"k8s.io/apimachinery/pkg/runtime"
@@ -34,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
 	"k8s.io/component-base/metrics/legacyregistry"
 	_ "k8s.io/component-base/metrics/prometheus/clientgo"
 	_ "k8s.io/component-base/metrics/prometheus/version"
@@ -122,8 +125,6 @@ func main() {
 	wdsClientMetrics := spacesClientMetrics.MetricsForSpace("wds")
 	itsClientMetrics := spacesClientMetrics.MetricsForSpace("its")
 
-	// TODO: engage leader election if requested
-
 	// get the config for WDS
 	setupLog.Info("Getting config for WDS", "name", wdsName)
 	wdsRestConfig, wdsName, err := ctrlutil.GetWDSKubeconfig(setupLog, wdsName)
@@ -144,10 +145,54 @@ func main() {
 	setupLog.Info("Got config for ITS", "name", itsName)
 	itsRestConfig = itsClientLimits.LimitConfig(itsRestConfig)
 
+	// Start controllers based on leader election setting
+	if enableLeaderElection {
+		setupLog.Info("Starting with leader election enabled")
+		startControllersWithLeaderElection(ctx, setupLog, wdsRestConfig, itsRestConfig, wdsName, itsName, allowedGroupsSet, ctlrsToStart, wdsClientMetrics, itsClientMetrics)
+	} else {
+		setupLog.Info("Starting without leader election")
+		startControllersDirectly(ctx, setupLog, wdsRestConfig, itsRestConfig, wdsName, itsName, allowedGroupsSet, ctlrsToStart, wdsClientMetrics, itsClientMetrics)
+	}
+
+	select {}
+}
+
+// startControllersWithLeaderElection starts controllers only after becoming the leader
+func startControllersWithLeaderElection(ctx context.Context, setupLog logr.Logger, wdsRestConfig, itsRestConfig *rest.Config, wdsName, itsName string, allowedGroupsSet sets.Set[string], ctlrsToStart sets.Set[string], wdsClientMetrics, itsClientMetrics ksmetrics.ClientMetrics) {
+	// Create a manager with leader election enabled
+	mgr, err := ctrl.NewManager(wdsRestConfig, ctrl.Options{
+		Scheme:                     scheme,
+		LeaderElection:             true,
+		LeaderElectionID:           "kubestellar-controller-manager",
+		LeaderElectionNamespace:    wdsName + "-system", // Use the WDS name to determine namespace
+		LeaderElectionResourceLock: "leases",
+		LeaderElectionReleaseOnCancel: true,
+	})
+	if err != nil {
+		setupLog.Error(err, "unable to create manager")
+		os.Exit(1)
+	}
+
+	// Start controllers in a goroutine that only runs when we become leader
+	go func() {
+		<-mgr.Elected()
+		setupLog.Info("Became leader, starting controllers")
+		startControllersDirectly(ctx, setupLog, wdsRestConfig, itsRestConfig, wdsName, itsName, allowedGroupsSet, ctlrsToStart, wdsClientMetrics, itsClientMetrics)
+	}()
+
+	// Start the manager (this will handle leader election)
+	if err := mgr.Start(ctx); err != nil {
+		setupLog.Error(err, "unable to start manager")
+		os.Exit(1)
+	}
+}
+
+// startControllersDirectly starts controllers immediately without leader election
+func startControllersDirectly(ctx context.Context, setupLog logr.Logger, wdsRestConfig, itsRestConfig *rest.Config, wdsName, itsName string, allowedGroupsSet sets.Set[string], ctlrsToStart sets.Set[string], wdsClientMetrics, itsClientMetrics ksmetrics.ClientMetrics) {
 	workloadEventRelay := &workloadEventRelay{}
 
 	// create the binding controller
-	bindingController, err := binding.NewController(logger, wdsClientMetrics, itsClientMetrics, wdsRestConfig, itsRestConfig, wdsName, allowedGroupsSet, workloadEventRelay)
+	bindingController, err := binding.NewController(setupLog, wdsClientMetrics, itsClientMetrics, wdsRestConfig, itsRestConfig, wdsName, allowedGroupsSet, workloadEventRelay)
 	if err != nil {
 		setupLog.Error(err, "unable to create binding controller")
 		os.Exit(1)
@@ -184,7 +229,7 @@ func main() {
 			time.Sleep(15 * time.Second)
 		}
 		setupLog.Info("Creating controller", "name", status.ControllerName)
-		statusController, err = status.NewController(logger, wdsClientMetrics, itsClientMetrics, wdsRestConfig, itsRestConfig, wdsName,
+		statusController, err = status.NewController(setupLog, wdsClientMetrics, itsClientMetrics, wdsRestConfig, itsRestConfig, wdsName,
 			bindingController.GetBindingPolicyResolver())
 		if err != nil {
 			setupLog.Error(err, "unable to create status controller")
@@ -212,8 +257,6 @@ func main() {
 			os.Exit(1)
 		}
 	}
-
-	select {}
 }
 
 // workloadEventRelay implements binding.WorkloadEventHandler and relays the notifications


MikeSpreitzer (commented):
The use of controller-runtime here is vestigial, should probably be completely removed. Use https://github.com/kubernetes/client-go/blob/v0.30.14/tools/leaderelection/leaderelection.go instead.

antedotee:
image
Now the logs are working as it was intended to be. 
In case of any improvement needed, do tell me.

MikeSpreitzer:
Re https://github.com/kubestellar/kubestellar/pull/2985#issuecomment-3007109606 : I find text more consumable than a picture of text.

cmd/controller-manager/main.go
```diff
@@ -59,13 +70,24 @@ const (
...
 	//+kubebuilder:scaffold:scheme
 }
 
+// getUniqueIdentity returns a unique identifier for leader election
+func getUniqueIdentity() string {
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "unknown"
+	}
+	return fmt.Sprintf("%s-%d", hostname, os.Getpid())
```
MikeSpreitzer: The adequacy of this depends on the user of this command. The user could create two Pods with the same hostname and same image, and the first process in each container always gets PID 1. I suggest adding more unique information. `time.Now().UnixNano()` should be pretty good.

cmd/controller-manager/main.go
```diff
@@ -144,10 +164,107 @@ func main() {
...
+	namespace := wdsName + "-system"
+	_, err = wdsClient.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			setupLog.Info("Creating namespace for leader election", "namespace", namespace)
+			_, err = wdsClient.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: namespace,
+				},
+			}, metav1.CreateOptions{})
```
MikeSpreitzer: It would be better to include a FieldManager identifying this client.

cmd/controller-manager/main.go
```diff
@@ -144,10 +164,107 @@ func main() {
...
+	leaderelection.RunOrDie(ctx, lec)
+}
+
+// startControllersDirectly starts controllers immediately without leader election
+func startControllersDirectly(ctx context.Context, setupLog logr.Logger, wdsRestConfig, itsRestConfig *rest.Config, wdsName, itsName string, allowedGroupsSet sets.Set[string], ctlrsToStart sets.Set[string], wdsClientMetrics, itsClientMetrics ksmetrics.ClientMetrics) {
 	workloadEventRelay := &workloadEventRelay{}
 
 	// create the binding controller
-	bindingController, err := binding.NewController(logger, wdsClientMetrics, itsClientMetrics, wdsRestConfig, itsRestConfig, wdsName, allowedGroupsSet, workloadEventRelay)
+	bindingController, err := binding.NewController(setupLog, wdsClientMetrics, itsClientMetrics, wdsRestConfig, itsRestConfig, wdsName, allowedGroupsSet, workloadEventRelay)
```
MikeSpreitzer: Why change which logger the controller gets? Same question for the other change that is like this one.

cmd/controller-manager/main.go
```diff
@@ -144,10 +164,107 @@ func main() {
...
+		Client: wdsClient.CoordinationV1(),
+		LockConfig: resourcelock.ResourceLockConfig{
+			Identity: getUniqueIdentity(),
+		},
+	}
+
+	// Configure leader election
+	lec := leaderelection.LeaderElectionConfig{
+		Lock:            lock,
+		ReleaseOnCancel: true,
```
MikeSpreitzer: What makes you think that this constraint is met: https://github.com/kubernetes/client-go/blob/v0.30.14/tools/leaderelection/leaderelection.go#L154-L156 ?

cmd/controller-manager/main.go
```diff
@@ -144,10 +164,107 @@ func main() {
...
+func startControllersWithLeaderElection(ctx context.Context, setupLog logr.Logger, wdsRestConfig, itsRestConfig *rest.Config, wdsName, itsName string, allowedGroupsSet sets.Set[string], ctlrsToStart sets.Set[string], wdsClientMetrics, itsClientMetrics ksmetrics.ClientMetrics) {
+	// Create WDS client for leader election
+	wdsClient, err := kubernetes.NewForConfig(wdsRestConfig)
+	if err != nil {
+		setupLog.Error(err, "unable to create WDS client for leader election")
+		os.Exit(1)
+	}
+
+	// Ensure the namespace exists for leader election
+	namespace := wdsName + "-system"
```
MikeSpreitzer: Since this namespace is within the WDS, there is no need to include the WDS name in the namespace name. I would go with something like "kubestellar-system".

MikeSpreitzer:
The end-to-end tests are still failing.

pkg/binding/binding-controller.go
MikeSpreitzer, line 1: The changes to this file look unrelated to the topic of this PR.
They also look like a bad idea. KubeStellar does NOT require that every non-KubeStellar object in a WDS be referenced by a Binding. That would be a very bad idea. Kubernetes API machinery does not allow an atomic transaction on more than one object. The Kubernetes style is to not impose ordering requirements on clients, but rather to continually react to the objects that are present.

cmd/controller-manager/main.go
```diff
@@ -134,20 +159,131 @@ func main() {
 	setupLog.Info("Got config for WDS", "name", wdsName)
 	wdsRestConfig = wdsClientLimits.LimitConfig(wdsRestConfig)
 
-	// get the config for ITS
+	// get the config for ITS with retry for resiliency
 	setupLog.Info("Getting config for ITS")
-	itsRestConfig, itsName, err := ctrlutil.GetITSKubeconfig(setupLog, itsName)
```
MikeSpreitzer: Note that there is already a retry loop in `ctrlutil.GetITSKubeconfig`. I think that it would be better to extend the retrying in that func than to write a retry loop over every call site of that func.

Also, this change is separate from the main topic of this PR, so maybe should go in a separate PR?

cmd/controller-manager/main.go
```diff
@@ -134,20 +159,131 @@ func main() {
...
+	_, err = wdsClient.CoreV1().Namespaces().Get(ctx, lockNamespace, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			setupLog.Info("Creating namespace for leader election", "namespace", lockNamespace)
+			_, err = wdsClient.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: lockNamespace,
+				},
+			}, metav1.CreateOptions{FieldManager: fieldManagerName})
+			if err != nil {
```
MikeSpreitzer: Consider what happens if I create a Deployment with 3 replicas doing this. It is not unlikely that they will concurrently try this Create. One will succeed, the other two fail with an "already exists" error (which you can identify with https://github.com/kubernetes/apimachinery/blob/v0.30.14/pkg/api/errors/errors.go#L538-L542) --- which is not a fatal problem, just a roundabout success.

MikeSpreitzer:
Ping @antedotee

cmd/controller-manager/main.go
antedotee: Yes,you are right😅. Does it mean that there has to be a shutdown logic that needs to be implemented?

cmd/controller-manager/main.go
```diff
@@ -57,15 +68,31 @@ var (
 const (
 	// number of workers to run the reconciliation loop
 	workers = 4
+	// getUniqueIdentity returns a unique identifier for leader election
+	lockNamespace    = "kubestellar-system"
+	lockName         = "kubestellar-controller-manager"
+	fieldManagerName = "kubestellar-controller-manager"
 )
 
+var readyFlag atomic.Bool
```
MikeSpreitzer, line 80: In my opinion, it is worth including a comment on exactly what state this boolean identifies.

cmd/controller-manager/main.go
```diff
@@ -144,6 +169,107 @@ func main() {
...
+			// Controller-manager is ready if:
+			// 1. It has become leader (readyFlag is true)
+			// 2. OR it's running without leader election (readyFlag is true)
+			// 3. AND it has successfully connected to WDS at least once
+			if readyFlag.Load() {
+				w.WriteHeader(http.StatusOK)
+				w.Write([]byte("ok"))
+			} else {
+				w.WriteHeader(http.StatusServiceUnavailable)
+				w.Write([]byte("not ready"))
```
MikeSpreitzer: Since readiness is defined above to be the conjunction of two things, it would be better if the response indicated both factors. This could be added in a follow-on PR.

cmd/controller-manager/main.go
```diff
@@ -144,6 +169,107 @@ func main() {
...
+			_, err = wdsClient.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: lockNamespace,
+				},
+			}, metav1.CreateOptions{FieldManager: fieldManagerName})
+			if err != nil && !errors.IsAlreadyExists(err) {
+				setupLog.Error(err, "unable to create namespace for leader election")
+				os.Exit(1)
+			}
+			// AlreadyExists is treated as success - another replica created it
```
MikeSpreitzer: FYI, to be really picky, this code does not actually know _what_ created the namespace. The expected common case is another replica of the same server, but other possibilities are, well, possible.

cmd/controller-manager/main.go
MikeSpreitzer: I do not have a lot of experience with this library. What the documentation says to me is that the client needs to EITHER (a) let this field hold `false` OR (b) have the behavior documented for `true`. In my opinion, since I see no evidence that the behavior constraint is met, I think that the simplest solution is to leave this field `false`. If and when that is known to cause a problem, then consider the other approach.

cmd/controller-manager/main.go
```diff
@@ -144,6 +169,107 @@ func main() {
...
+		},
+	}
+
+	// Configure leader election
+	lec := leaderelection.LeaderElectionConfig{
+		Lock:            lock,
+		ReleaseOnCancel: true,
+		LeaseDuration:   15 * time.Second,
+		RenewDeadline:   10 * time.Second,
+		RetryPeriod:     2 * time.Second,
```
MikeSpreitzer: This is pretty frequent. Do we really need the lease duration to be only 15 seconds?

cmd/controller-manager/main.go
```diff
@@ -144,6 +169,107 @@ func main() {
...
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(ctx context.Context) {
+				setupLog.Info("Acquired leadership, starting controllers and workers")
+				startControllersDirectly(ctx, setupLog, wdsRestConfig, itsRestConfig, wdsName, itsName, allowedGroupsSet, ctlrsToStart, wdsClientMetrics, itsClientMetrics)
+				readyFlag.Store(true)
```
MikeSpreitzer: Why repeat this statement that is also in `startControllersDirectly` ?

cmd/controller-manager/main.go
```diff
@@ -144,6 +169,107 @@ func main() {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name: lockNamespace,
+				},
+			}, metav1.CreateOptions{FieldManager: fieldManagerName})
+			if err != nil && !errors.IsAlreadyExists(err) {
+				setupLog.Error(err, "unable to create namespace for leader election")
+				os.Exit(1)
+			}
+			// AlreadyExists is treated as success - another replica created it
+		} else {
```
MikeSpreitzer, line 294: A controller should be tolerant of transient errors (i.e., if one occurs then retry after waiting a while).
See https://github.com/kubestellar/kubeflex/blob/c53037cb055baae841dedf7d9ad1927e1dcc05c1/pkg/util/errors.go#L8-L16 for an example characterization of them.

clubanderson:

MikeSpreitzer:
Ping @antedotee

antedotee:
Working on it, if I am unable to do it today, I will close this PR.

cmd/controller-manager/main.go
```diff
@@ -144,6 +213,144 @@ func main() {
...
+			if readyFlag.Load() {
+				w.WriteHeader(http.StatusOK)
+				w.Write([]byte("ok"))
+			} else {
+				w.WriteHeader(http.StatusServiceUnavailable)
+				w.Write([]byte("not ready - waiting for leadership or controller startup"))
+			}
+		})
+		http.ListenAndServe(":8081", nil)
+	}()
```
MikeSpreitzer, line 232: Please review the contents and use of `processOpts`.

cmd/controller-manager/main.go
```diff
@@ -144,6 +213,144 @@ func main() {
...
+			time.Sleep(time.Duration(i+1) * time.Second)
+		} else if i < 2 {
+			setupLog.Info("Failed to create WDS client, retrying", "attempt", i+1, "error", err)
+			time.Sleep(time.Duration(i+1) * time.Second)
+		}
+	}
+	if err != nil {
+		setupLog.Error(err, "unable to create WDS client for leader election after retries")
+		os.Exit(1)
+	}
```
MikeSpreitzer, line 267: `kubernetes.NewForConfig` does not do any network I/O and will not raise transient errors. There is no need for a retry loop here.

cmd/controller-manager/main.go
```diff
@@ -144,6 +213,144 @@ func main() {
...
+
+	// Check/create namespace for leader election with retry for transient errors
+	var namespaceErr error
+	for i := 0; i < 3; i++ {
```
MikeSpreitzer, line 271: Why retry a limited number of times?
Why not use one of the Kubernetes utility functions that retries with exponential backoff (see https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait)?

cmd/controller-manager/main.go
```diff
@@ -144,6 +213,144 @@ func main() {
...
+	lec := leaderelection.LeaderElectionConfig{
+		Lock:            lock,
+		ReleaseOnCancel: false, // Set to false as the shutdown behavior constraint is not met
+		LeaseDuration:   60 * time.Second,
+		RenewDeadline:   40 * time.Second,
+		RetryPeriod:     5 * time.Second,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(ctx context.Context) {
+				setupLog.Info("Acquired leadership, starting controllers and workers")
+				startControllersDirectly(ctx, setupLog, wdsRestConfig, itsRestConfig, wdsName, itsName, allowedGroupsSet, ctlrsToStart, wdsClientMetrics, itsClientMetrics)
```
MikeSpreitzer, line 335: This will lead to a duplicate call on `kubernetes.NewForConfig(wdsRestConfig)` (inside binding.NewController called from startControllersDirectly). It would be marginally better to have only one call on `kubernetes.NewForConfig` (and this will help prepare for unit testing of these controllers). Can be addressed in follow-on.

cmd/controller-manager/main.go
```diff
@@ -144,6 +213,144 @@ func main() {
...
+		}
+		if errors.IsNotFound(namespaceErr) {
+			setupLog.Info("Creating namespace for leader election", "namespace", lockNamespace)
+			_, createErr := wdsClient.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: lockNamespace,
+				},
+			}, metav1.CreateOptions{FieldManager: fieldManagerName})
+			if createErr == nil || errors.IsAlreadyExists(createErr) {
+				// AlreadyExists is treated as success - another replica or previous run created it
```
MikeSpreitzer, line 284: This code does not know what created the Namespace. it could be something completely unrelated, or a related but unexpected thing. It does not matter what created the Namespace. Let's not try to characterize it.

cmd/controller-manager/main.go
```diff
@@ -144,6 +213,144 @@ func main() {
...
+	// These durations follow the standard Kubernetes leader election patterns:
+	// - LeaseDuration: 60s provides more stability and reduces API server load
+	// - RenewDeadline: 40s allows for timely renewal (2/3 of lease duration)
+	// - RetryPeriod: 5s reduces retry frequency for better performance
+	lec := leaderelection.LeaderElectionConfig{
+		Lock:            lock,
+		ReleaseOnCancel: false, // Set to false as the shutdown behavior constraint is not met
+		LeaseDuration:   60 * time.Second,
+		RenewDeadline:   40 * time.Second,
+		RetryPeriod:     5 * time.Second,
```
MikeSpreitzer, line 331: FYI, I would make this a bit longer. 8 seconds gets two tries from RenewDeadline to LeaseDuration even allowing 1 second for the apiserver round trip (the SLO) and 1 sec of other delays.

MikeSpreitzer:
Ping @antedotee

MikeSpreitzer:
@antedotee, do you intend to continue working on this?

vedansh-5:
@antedotee do you wish to continue working on this?

antedotee:
@vedansh-5 Yes, you can pick this up. Thanks and regards.

PR #2986, opened 2025-06-20
🐛Fix race condition where PRs merge before all tests complete

### LINKED ISSUE #1872: bug: PRs are merged before tests are completed.

antedotee:
## Summary
Fix bug where PRs were merging before GitHub Actions tests completed, causing potential instability in the main branch.

## Problem
Prow's Tide automation was only waiting for Prow presubmit jobs to pass before merging PRs, ignoring GitHub Actions workflows. This created a race condition where PRs could merge while e2e tests, integration tests, and other GitHub Actions were still running.

## Solution
Updated Prow Tide configuration to require ALL test contexts (both Prow presubmits and GitHub Actions) to pass before allowing merges.

## Changes
Modified `.prow.yaml` to include all 7 required test contexts:

**Prow presubmits:**
- `pull-kubestellar-verify`
- `pull-kubestellar-lint` 
- `pull-kubestellar-test-unit`

**GitHub Actions:**
- `Test end to end / Tests in ginkgo`
- `Test end to end / Test in bash`
- `Run integration tests / Test`
- `PR Title Verifier / verify`

- PRs now cannot merge until every single test passes
- Eliminates race conditions between Prow and GitHub Actions test systems

Fixes #1872

### DIFF
--- a/.github/workflows/pr-test-e2e.yml
+++ b/.github/workflows/pr-test-e2e.yml
@@ -6,45 +6,45 @@ on:
   workflow_dispatch:
     inputs:
       testFlags:
-        description: 'Command line flags for run-test.sh'
+        description: "Command line flags for run-test.sh"
         required: false
-        default: ''
+        default: ""
         type: choice
         options:
-        - ''
-        - '--released'
+          - ""
+          - "--released"
   pull_request:
     branches:
       - main
     paths-ignore:
-      - 'docs/**'
-      - '**/*.md'
+      - "docs/**"
+      - "**/*.md"
   push:
     branches:
       - main
     tags:
-      - 'v*'
+      - "v*"
     paths-ignore:
-      - 'docs/**'
-      - '**/*.md'
+      - "docs/**"
+      - "**/*.md"
 
 jobs:
   ginkgo-test:
     name: Tests in ginkgo
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
 
-      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 
+      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5
         with:
           go-version: 1.24
           cache: true
 
       - name: Install kubectl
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
/ok-to-test

Has this been tested?

.prow.yaml
```diff
@@ -47,8 +47,35 @@ presubmits:
...
+    - repos:
+        - kubestellar/kubestellar
+      labels:
+        - lgtm
+        - approved
+      missingLabels:
+        - do-not-merge
+        - do-not-merge/hold
+        - do-not-merge/work-in-progress
+        - needs-rebase
```
MikeSpreitzer: Do we need this here? We already have https://github.com/kubestellar/infra/blob/main/prow/config.yaml#L102-L117 .

.prow.yaml
```diff
@@ -47,8 +47,35 @@ presubmits:
...
+    required-contexts:
+      # Prow presubmits
+      - pull-kubestellar-verify
+      - pull-kubestellar-lint
+      - pull-kubestellar-test-unit
+      # GitHub Actions workflows
+      - "Test end to end / Tests in ginkgo"
+      - "Test end to end / Test in bash"
+      - "Run integration tests / Test"
+      - "PR Title Verifier / verify"
```
MikeSpreitzer: Is this merged into https://github.com/kubestellar/infra/blob/main/prow/config.yaml#L93-L129 ? I see nothing about this question in https://docs.prow.k8s.io/docs/inrepoconfig/ .

Does this belong in https://github.com/kubestellar/infra/blob/main/prow/config.yaml instead? Considerations include the question above and whether we want the same required contexts for each of the repos involved.

antedotee:
Thanks for the advice, I considered it and read the code thoroughly and implemented the revised solution now.

.github/workflows/pr-test-e2e.yml
```diff
@@ -289,3 +289,22 @@ jobs:
           sleep 10
           curl http://localhost:8080/metrics
           kill %
+
+  # Consolidation job that ensures ALL tests pass before allowing merge
+  all-tests-complete:
+    name: All Tests Complete
```
MikeSpreitzer, line 295: "all" is not wrong, if you have a narrow enough view, but maybe "both" is better? Or "e2e"?

MikeSpreitzer:
Does the latest rev solve the problem?
If not, could we solve it by adding to https://github.com/kubestellar/infra/blob/main/prow/config.yaml#L137 ?
Remember also that required status checks can be configured in the GitHub web UI (as shown below).
image

PR #2993, opened 2025-06-22
📖 improve ITS, WDS, WEC and registration pages

rishi-jat:
Fixes #2973

PR #2997, opened 2025-06-22
📖 wec-registation doc added

Rupam-It:
## Related issue(s)

Fixes #

### DIFF
--- a/docs/content/direct/wec-registration.md
+++ b/docs/content/direct/wec-registration.md
@@ -1,5 +1,330 @@
 # Registering a Workload Execution Cluster
 
-This document will tell users what how to register a WEC in an ITS.
+- [Overview](#overview)
+- [Prerequisites](#prerequisites)
+- [Registration Process](#registration-process)
+  - [Step 1: Get the Registration Token](#step-1-get-the-registration-token)
+  - [Step 2: Join the WEC to the ITS](#step-2-join-the-wec-to-the-its)
+  - [Step 3: Accept the Registration](#step-3-accept-the-registration)
+  - [Step 4: Verify Registration](#step-4-verify-registration)
+- [Post-Registration Configuration](#post-registration-configuration)
+  - [Labeling WECs](#labeling-wecs)
+  - [Customization Properties](#customization-properties)
+- [Different Deployment Scenarios](#different-deployment-scenarios)
+  - [Kind Clusters](#kind-clusters)
+  - [K3d Clusters](#k3d-clusters)
+  - [OpenShift Clusters](#openshift-clusters)
+  - [Production Clusters](#production-clusters)
+- [Troubleshooting](#troubleshooting)
+- [Managing Registered WECs](#managing-registered-wecs)
+- [Next Steps](#next-steps)
 
-TODO: write this.
+## Overview
+
+This document provides detailed instructions on how to register a Workload Execution Cluster (WEC) with an Inventory and Transport Space (ITS) in KubeStellar. 
+
+WEC registration is the process of connecting a Kubernetes cluster to KubeStellar so that it can receive and run workloads distributed from Workload Description Spaces (WDSes). The registration process involves installing the OCM (Open Cluster Management) Agent on the WEC and establishing a secure communication channel with the ITS.
+
+## Prerequisites
+
+Before registering a WEC, ensure you have:
+
+1. **A running Kubernetes cluster** that will serve as the WEC
+2. **Network connectivity** between the WEC and the ITS
+3. **`kubectl` access** to both the WEC and ITS
+4. **`clusteradm` CLI tool** installed ([installation guide](https://open-cluster-management.io/docs/getting-started/installation/start-the-control-plane/))
+5. **An existing ITS** with OCM cluster manager running
+6. **Sufficient permissions** to create resources in both clusters
+
+To verify your ITS is ready for WEC registration:
+
+```shell
+# Check if the ITS is accessible
+kubectl --context <its-context> get ns
+
+# Verify OCM cluster manager is running
+kubectl --context <its-context> get pods -n open-cluster-management-hub
+
+# Check for the customization-properties namespace
+kubectl --context <its-context> get ns customization-properties
+```
+
+## Registration Process
+
+### Step 1: Get the Registration Token
+
+The first step is to obtain a registration token from the ITS. This token is used to authenticate the WEC during the registration process.
+
+```shell
+# Get the registration token from the ITS
+clusteradm --context <its-context> get token
+```
+
+This command will output a `clusteradm join` command that you'll use in the next step. The output will look like:
+
+```shell
+clusteradm join --hub-token <token> --hub-apiserver <api-server-url> --cluster-name <cluster_name>
+```
+
+### Step 2: Join the WEC to the ITS
+
+Execute the join command on the WEC to initiate the registration process:
+
+```shell
+# Basic join command
+clusteradm join --hub-token <token> --hub-apiserver <api-server-url> --cluster-name <your-wec-name> --context <wec-context>
+```
+
+**Important flags:**
+- `--cluster-name`: Choose a unique name for your WEC
+- `--context`: Specify the kubectl context for your WEC
+- `--singleton`: Use this flag if the WEC is a single-node cluster
+- `--force-internal-endpoint-lookup`: Required for Kind clusters and other local development setups
+
+**Example for Kind clusters:**
+```shell
+clusteradm join --hub-token <token> --hub-apiserver <api-server-url> --cluster-name cluster1 --context cluster1 --singleton --force-internal-endpoint-lookup
+```
+
+**Example for production clusters:**
+```shell
+clusteradm join --hub-token <token> --hub-apiserver <api-server-url> --cluster-name prod-cluster-east --context prod-cluster-east
+```
+
+### Step 3: Accept the Registration
+
+After the join command completes, you need to approve the registration request from the ITS side:
+
+```shell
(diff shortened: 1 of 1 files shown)


docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,330 @@
...
+- [Troubleshooting](#troubleshooting)
+- [Managing Registered WECs](#managing-registered-wecs)
+- [Next Steps](#next-steps)
 
-TODO: write this.
+## Overview
+
+This document provides detailed instructions on how to register a Workload Execution Cluster (WEC) with an Inventory and Transport Space (ITS) in KubeStellar. 
+
+WEC registration is the process of connecting a Kubernetes cluster to KubeStellar so that it can receive and run workloads distributed from Workload Description Spaces (WDSes). The registration process involves installing the OCM (Open Cluster Management) Agent on the WEC and establishing a secure communication channel with the ITS.
```
MikeSpreitzer, line 26: The first thing to say is that registering a WEC in an ITS is no more and no less than the process from Open Cluster Management. This document should refer the reader to https://open-cluster-management.io/docs/getting-started/installation/register-a-cluster/ and remind them of the mapping between OCM terminology and KubeStellar terminology.

MikeSpreitzer:
A PR that modifies the website should include a preview of the modified website. See https://docs.kubestellar.io/unreleased-development/contribution-guidelines/operations/document-management/#automatically-render-the-website

MikeSpreitzer:
@Rupam-It: can you get together with the author of #2993 and come up with a consolidated proposal?

Rupam-It:
@rishi-jat please take a look in this PR !

Rupam-It:
> A PR that modifies the website should include a preview of the modified website. See https://docs.kubestellar.io/unreleased-development/contribution-guidelines/operations/document-management/#automatically-render-the-website

https://rupam-it.github.io/kubestellar/wec-registation/direct/wec-registration/
here is the link .

rishi-jat:
Hi @Rupam-It, great job documenting the WEC registration flow this is detailed and very helpful! 

Just a few suggestions based on @MikeSpreitzer’s feedback:
	•	Consider adding a direct reference to the official [OCM registration guide](https://open-cluster-management.io/docs/getting-started/installation/register-a-cluster/) instead of duplicating the entire process. This will help avoid drift from upstream changes.
	•	A small terminology mapping table (e.g., OCM’s “hub” → KubeStellar’s “ITS”, “managed cluster” → “WEC”) could make it easier for readers new to both projects.
	•	Minor typo in the introduction:
“what how to register” → maybe simplify to “how to register”

The preview site looks good: [📄 Preview](https://rupam-it.github.io/kubestellar/wec-registation/direct/wec-registration/)

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,330 @@
...
+3. **`kubectl` access** to both the WEC and ITS
+4. **`clusteradm` CLI tool** installed ([installation guide](https://open-cluster-management.io/docs/getting-started/installation/start-the-control-plane/))
+5. **An existing ITS** with OCM cluster manager running
+6. **Sufficient permissions** to create resources in both clusters
+
+To verify your ITS is ready for WEC registration:
+
+```shell
+# Check if the ITS is accessible
+kubectl --context <its-context> get ns
```
MikeSpreitzer, line 43: FYI: this is not wrong, but on a typical OpenShift cluster returns a painfully large result. You always get a short result from `kubectl get ServiceAccount default`. Or perhaps a little more informative result from `kubectl version`.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,330 @@
...
+A successfully registered WEC will show:
+- Status: `Available` and `Joined`
+- OCM agent pods running in the `open-cluster-management-agent` namespace
+- A corresponding mailbox namespace in the ITS
+
+## Post-Registration Configuration
+
+### Labeling WECs
+
+After registration, you should label your WECs to make them selectable by BindingPolicies. Labels help categorize WECs based on their characteristics:
```
MikeSpreitzer, line 135: I would include a remark here saying that the labeling displayed below is an example, not the one true set of labels to apply.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,330 @@
...
+6. **Sufficient permissions** to create resources in both clusters
+
+To verify your ITS is ready for WEC registration:
+
+```shell
+# Check if the ITS is accessible
+kubectl --context <its-context> get ns
+
+# Verify OCM cluster manager is running
+kubectl --context <its-context> get pods -n open-cluster-management-hub
```
MikeSpreitzer, line 46: Some readers might wonder if the result the get is good or not. It would help to give a clue here.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,330 @@
...
+## Different Deployment Scenarios
+
+### Kind Clusters
+
+For local development with Kind clusters:
+
+```shell
+# Create Kind cluster
+kind create cluster --name cluster1
+kubectl config rename-context kind-cluster1 cluster1
```
MikeSpreitzer, line 178: You might point out that this renaming is not required. It is just something that we typically do for conceptual convenience.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,330 @@
...
+```shell
+clusteradm join --hub-token <token> --hub-apiserver <api-server-url> --cluster-name prod-cluster-east --context prod-cluster-east
+```
+
+### Step 3: Accept the Registration
+
+After the join command completes, you need to approve the registration request from the ITS side:
+
+```shell
+# Check for pending certificate signing requests
```
MikeSpreitzer, line 99: I would break these two commands out into separate displays and explain more about the first, as in Step 2 of https://docs.kubestellar.io/unreleased-development/direct/get-started/#create-and-register-two-workload-execution-clusters

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,330 @@
...
+```shell
+# Create Kind cluster
+kind create cluster --name cluster1
+kubectl config rename-context kind-cluster1 cluster1
+
+# Register with ITS (note the --force-internal-endpoint-lookup flag)
+clusteradm --context its1 get token | grep '^clusteradm join' | \
+  sed "s/<cluster_name>/cluster1/" | \
+  awk '{print $0 " --context cluster1 --singleton --force-internal-endpoint-lookup"}' | sh
+
```
MikeSpreitzer, line 184: Don't forget the wait for the CSR to appear.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,330 @@
...
+# Accept registration
+clusteradm --context its1 accept --clusters cluster1
+```
+
+### K3d Clusters
+
+For K3d clusters with load balancer:
```
MikeSpreitzer, line 191: This is exactly the same as the `kind` case, except for the first couplet of commands. I would revise to remove the duplication.

Not a blocker.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,330 @@
...
+# Register OpenShift cluster
+clusteradm --context its1 get token | grep '^clusteradm join' | \
+  sed "s/<cluster_name>/openshift-cluster/" | \
+  awk '{print $0 " --context openshift-cluster"}' | sh
+
+# Accept registration
+clusteradm --context its1 accept --clusters openshift-cluster
+```
+
+### Production Clusters
```
MikeSpreitzer, line 221: The relevant point here is that these clusters are from cloud providers, not that they are "production".

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,330 @@
...
+  sed "s/<cluster_name>/cluster1/" | \
+  awk '{print $0 " --context cluster1 --singleton --force-internal-endpoint-lookup"}' | sh
+
+# Accept registration
+clusteradm --context its1 accept --clusters cluster1
+```
+
+### OpenShift Clusters
+
+For OpenShift clusters, omit the `--force-internal-endpoint-lookup` flag:
```
MikeSpreitzer, line 209: This is a little incongruous with the previous section (for k3d). Currently, that section repeats all three clumps of commands, even though there is a difference only in the first. Here there is a repetition of just two clumps of commands. And again, there is a difference in only one. I would revise to remove duplication.

Not a blocker.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,330 @@
...
+```
+
+## Troubleshooting
+
+### Common Issues and Solutions
+
+**1. CSR not appearing**
+```shell
+# Check if the join command completed successfully
+kubectl --context <wec-context> get pods -n open-cluster-management-agent
```
MikeSpreitzer, line 245: The reader will be wondering what to look for in the output.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,330 @@
...
+### Deregistering a WEC
+
+To remove a WEC from KubeStellar:
+
+```shell
+# Delete the ManagedCluster resource
+kubectl --context <its-context> delete managedcluster <wec-name>
+
+# Clean up the WEC (optional)
+kubectl --context <wec-context> delete namespace open-cluster-management-agent
```
MikeSpreitzer, line 329: What about `open-cluster-management-agent` and `open-cluster-management-agent-addon` ?

MikeSpreitzer:
I think the author intends PR #3048 as a replacement for this one.

Issue #2595, opened 2024-11-18
feature: increase our percentage/grade on https://clomonitor.io/projects/cncf/kubestellar


clubanderson:
### Feature Description

CNCF is publicizing the projects that achieve high marks on https://clomonitor.io/projects/cncf/kubestellar

### Proposed Solution

KubeStellar should evaluate the metrics that are used on https://clomonitor.io/projects/cncf/kubestellar and try to fulfill them. I have added roadmap and trademark links to satisfy some of the criteria

### Want to contribute?

- [ ] I would like to work on this issue.

MikeSpreitzer:
The following PRs have contributed:

- #2593 
- #2594 
- #2601

antedotee:
<img width="611" alt="Image" src="https://github.com/user-attachments/assets/368aa1b7-3c48-41cc-9c59-28990f4aad30" />

Solved this vulnerability and scored went from 68 to 77 in security 😅 🎉, more to go

clubanderson:
Wow. That's incredible. Thank you so much @antedotee 

What other items can be addressed?

antedotee:
I am still looking to solve other vulnerabilities, but, once all are solved, I mean there should be a way or tests to ensure that the changes doesn't affect CLOMonitor guards. 

In my devops projects, I used SAST tools and DAST tools in my CI pipeline to prevent this from happening. I think we can implement some sort of solution here also (I am not sure, just a suggestion).

AkashKumar7902:
Current CLOMonitor data shows KubeStellar at 97 overall, with security still affected by signed releases and workflow token-permission findings.

PR #3765 now includes an additional token-permissions hardening pass: required write scopes are job-scoped on the reusable workflow callers, unused caller permissions were removed where possible, and a local OpenSSF Scorecard Token-Permissions run reports `10 / 10`.

Signed releases still look like a separate release-process item because they require release artifact provenance/signing changes rather than repository workflow permission edits.

PR #3005, opened 2025-06-24
✨ restricted permissions for GITHUB_TOKEN to minimum permissions to prevent security risks

antedotee:
### Problem
GITHUB_TOKEN was not set to minimum permissions which was causing security risks
image

### Solution
Restricted permissions in the following files for GITHUB_TOKEN to read only:
image


antedotee:
This PR is related to the issue #2595 .

PR #3020, opened 2025-07-01
✨Dependency policy fix for CLOMonitor

antedotee:
### Problem
The dependency policy check was not passing in CLOMonitor earlier which could have caused some security vulnerabilities.

### Solution
After making the required changes by reading the documentation, I ran the CLOMonitor tests against my local branch and it passed the dependency policy check.
CleanShot 2025-07-01 at 23 34 27@2x

### DIFF
--- a/SECURITY-INSIGHTS.yml
+++ b/SECURITY-INSIGHTS.yml
@@ -52,6 +52,6 @@ dependencies:
   dependencies-lists:
   - https://github.com/kubestellar/kubestellar/blob/main/SECURITY.md
   env-dependencies-policy:
-    policy-url: https://github.com/kubestellar/kubestellar/blob/main/SECURITY.md
+    policy-url: https://github.com/kubestellar/kubestellar/blob/main/SECURITY.md#dependencies-policy
     comment: |
       KubeStellar uses Dependabot for automated dependency updates and follows the policy described in SECURITY.md.
--- a/SECURITY-INSIGHTS.yml
+++ b/SECURITY-INSIGHTS.yml
@@ -51,4 +51,7 @@ dependencies:
   third-party-packages: true
   dependencies-lists:
   - https://github.com/kubestellar/kubestellar/blob/main/SECURITY.md
-  env-dependencies-policy: https://github.com/kubestellar/kubestellar/blob/main/SECURITY.md
+  env-dependencies-policy:
+    policy-url: https://github.com/kubestellar/kubestellar/blob/main/SECURITY.md
+    comment: |
+      KubeStellar uses Dependabot for automated dependency updates and follows the policy described in SECURITY.md.
--- a/SECURITY-INSIGHTS.yml
+++ b/SECURITY-INSIGHTS.yml
@@ -51,3 +51,4 @@ dependencies:
   third-party-packages: true
   dependencies-lists:
   - https://github.com/kubestellar/kubestellar/blob/main/SECURITY.md
+  env-dependencies-policy: https://github.com/kubestellar/kubestellar/blob/main/SECURITY.md#dependencies-policy
--- a/SECURITY-INSIGHTS.yml
+++ b/SECURITY-INSIGHTS.yml
@@ -51,4 +51,4 @@ dependencies:
   third-party-packages: true
   dependencies-lists:
   - https://github.com/kubestellar/kubestellar/blob/main/SECURITY.md
-  env-dependencies-policy: https://github.com/kubestellar/kubestellar/blob/main/SECURITY.md#dependencies-policy
+  env-dependencies-policy: https://github.com/kubestellar/kubestellar/blob/main/SECURITY.md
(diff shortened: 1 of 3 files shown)


antedotee:
This is related to #2595 .

antedotee:
If not needed, we can delete the dependency section which I have added in security.md file.

SECURITY.md
```diff
@@ -3,6 +3,15 @@
 
 Join the [kubestellar-security-announce](https://groups.google.com/u/1/g/kubestellar-security-announce) group for emails about security and major API announcements.
 
+## Dependencies Policy
+
+KubeStellar manages its dependencies with the following policy:
+
+- **Automated Updates:** We use [Dependabot](https://github.com/dependabot) to automatically check for and propose updates to dependencies in Go modules, Python requirements, Dockerfiles, and Helm charts. Dependabot PRs are reviewed and merged by maintainers.
```
MikeSpreitzer: Not exactly. The Dependabot PRs are a prompt, but we do not accept them. First, we wait a while, to give problems a chance to surface. E.g., #3042 is in that state right now. Then we do our own PR that is consistent with https://github.com/kubestellar/kubestellar/blob/11a92742824ec3aff6a36d0d94030a26a792e0bc/CONTRIBUTING.md#github-action-reference-discipline ; #3015 is an example of that.

antedotee: I added this dependency policy just for better reference. We can skip this section altogether, just adding comment in security insights file will also pass the dependencies policy check.
Should I add this section in security.md or not?

SECURITY.md
```diff
@@ -3,6 +3,15 @@
...
+- **Review Process:** All dependency update pull requests are subject to the same review process as other code changes. Maintainers verify that updates do not introduce breaking changes or known vulnerabilities before merging.
```
MikeSpreitzer: How is the vulnerability check done before merging?

There is a CI workflow (https://github.com/kubestellar/kubestellar/blob/11a92742824ec3aff6a36d0d94030a26a792e0bc/.github/workflows/verify-action-hashes.yaml) that checks that each GitHub Action reference uses the approved commit of that Action.

SECURITY-INSIGHTS.yml
```diff
@@ -51,3 +51,7 @@ dependencies:
   third-party-packages: true
   dependencies-lists:
   - https://github.com/kubestellar/kubestellar/blob/main/SECURITY.md
+  env-dependencies-policy:
+    policy-url: https://github.com/kubestellar/kubestellar/blob/main/SECURITY.md#dependencies-policy
+    comment: |
+      KubeStellar uses Dependabot for automated dependency updates and follows the policy described in SECURITY.md.
```
MikeSpreitzer: s/for automated dependency updates/for detecting dependency updates/

SECURITY.md
MikeSpreitzer: Well, Dependabot covers more than just the dependencies in GitHub Actions workflows. But for all dependencies, I think that it is dangerous to switch to the latest version as soon as it is available. Is there a general way to check for vulnerabilities in a new release of a dependency? If so, that would be worth documenting somewhere. Since we already have a section on reviewing (https://github.com/kubestellar/kubestellar/blob/11a92742824ec3aff6a36d0d94030a26a792e0bc/CONTRIBUTING.md#review-and-approval-process), that would be the logical place.

Let's not introduce redundant section(s), instead upgrade (as needed) existing sections and reference those.

MikeSpreitzer:
Thank you for working on this topic. Unfortunately, there have been concurrent changes to the website while you did this work. Please look at the current content (e.g., https://docs.kubestellar.io/unreleased-development/contribution-guidelines/contributing-inc/) and revise this PR to build on the current content. You could do that either by (a) rebasing your branch onto the latest edition of `main` or (b) starting over from `main` and adding/changing what you want to add/change.

MikeSpreitzer:
Since this PR modifies the website, it needs a preview of the modified website.

CONTRIBUTING.md
```diff
@@ -154,6 +166,20 @@ Prior to making a new release, there needs to be testing that the
 current Helm chart works with the executable behavior that will
 appear in the new release.  
 
+## GitHub Action Reference Discipline
```
MikeSpreitzer, line 169: Why is this PR inserting material that already exists in this file (https://github.com/kubestellar/kubestellar/blob/87ed2495649c87eab669fe0d413fb15f5e2da41e/CONTRIBUTING.md?plain=1#L112-L123)? Something is going wrong here.

SECURITY.md
```diff
@@ -3,6 +3,17 @@
 
 Join the [kubestellar-security-announce](https://groups.google.com/u/1/g/kubestellar-security-announce) group for emails about security and major API announcements.
 
+## Dependencies Policy
```
MikeSpreitzer, line 6: There is a lot of redundancy between this material and the material that this PR adds to `CONTRIBUTING.md`. It is bad to have two copies of the same largish chunk of information, that increases the maintenance costs.

Issue #3000, opened 2025-06-23
doc: issues with website page about WDS


MikeSpreitzer:
### Describe the issue

I noticed multiple issues with the WDS website page.

The text before the "What is a WDS?" section heading ("A Workload Description Space (WDS) is ...") looks like content that belongs in that section.

There are additional kinds of control objects that can appear in a WDS.

Note that the word "description" was chosen because it encompasses both the definition of the desired state and reports/summaries of the reported state.

There is currently only one supported way to create a WDS, that is with the core Helm chart.

Note the following rendering equivalence failure.

<img width="1625" alt="Image" src="https://github.com/user-attachments/assets/e7558b71-e65d-4624-816a-02f3ee094322" />

The kubestellar-controller-manager and transport-controller do NOT run "in" the WDS. They work _on_ the WDS, but run _in_ the KubeFlex hosting cluster.

The section "Accessing the WDS" shows two commands that do different things and should be explained. The first creates or updates the kubeconfig context named "wds1", and make it the current context. The second command only makes the context named "wds1" current.

This page and the architecture page should reference each other.

### Want to contribute?

- [ ] I would like to work on this issue.

MikeSpreitzer:
@rishi-jat
@Rupam-It 
@francostellari 
@KPRoche

MikeSpreitzer:
The cardinalities should be explicitly and clearly stated.
- There may be multiple ITSes.
- There may be multiple WDSes.
- There is a many-to-one relation from WDS to ITS.

RudraCodesForU:
@MikeSpreitzer, Hi, I’d like to work on this issue. Please let me know if anyone is already working on it or if there are any specific guidelines you’d like me to follow. Thanks!

Tob-iee:
Hi @MikeSpreitzer, I have a quick question to help me clarify some details. 

1. Are you saying the ' kflex create "<wds-name>" -p kubestellar` can’t be used to create a WDS, because a WDS object gets created when I run it?
2. Are the KubeStellar controller manager and transport controller created upon the creation of a WDS in the KubeFlex hosting cluster? Or are they part of the controllers that are included in the control plane of the Hosting cluster when it was initially created? Or does the KubeStellar controller manager created in the Hosting cluster as a result of a post-create hook run that deploys it to the hosting cluster during the creation of the WDS, and vice versa for the transport controller with the ITS

MikeSpreitzer:
[...]

PR #3002, opened 2025-06-23
📖 docs: Improve the doc content for the WDS page #3000

Tob-iee:
## Related issue(s)
#3000

Fixes #
- Moved introductory content ("A Workload Description Space (WDS) is ...") to the appropriate section, under "What is a WDS?"
- Updated documentation to reflect that currently, the only supported way to create a WDS is with the core Helm chart.
- Fixed rendering Indentation for the WDS creation section
- Clarified that the kubestellar-controller-manager and transport-controller run in the KubeFlex hosting cluster, not in the WDS.
- Updated the "Accessing the WDS" section to explain the distinction between the two commands for creating/updating the kubeconfig context.
- Added a link on the page to reference the architecture page.

PR #3023, opened 2025-07-01
📖 Improve the doc content for the WDS page #3000

Tob-iee:
## Related issue(s)

3000

Fixes #

- Moved introductory content ("A Workload Description Space (WDS) is ...") to the appropriate section, under "What is a WDS?"
- Updated documentation to reflect that currently, the only supported way to create a WDS is with the core Helm chart.
- Fixed rendering Indentation for the WDS creation section
- Clarified that the kubestellar-controller-manager and transport-controller run in the KubeFlex hosting cluster, not in the WDS.
- Updated the "Accessing the WDS" section to explain the distinction between the two commands for creating/updating the kubeconfig context.
- Added a link on the page to reference the architecture page.

### DIFF
--- a/docs/content/direct/wds.md
+++ b/docs/content/direct/wds.md
@@ -14,19 +14,17 @@
 
 ## What is a WDS?
 
-A Workload Description Space (WDS) is a space in the [KubeStellar architecture](docs/content/direct/user-guide-intro.md) that serves as the primary interface for users to define and manage workloads for multi-cluster deployment. The WDS constitue of a Kubernetes API server with storage that:
+A Workload Description Space (WDS) is a space in the [KubeStellar architecture](user-guide-intro.md) that serves as the primary interface for users to define and manage workloads for multi-cluster deployment. The WDS constitue of a Kubernetes API server with storage that:
 
 - Stores the definitions of workloads in their native Kubernetes format
-- Hosts the control objects (`BindingPolicy` and `Binding`) that define how workloads are distributed
+- Hosts the control objects (`BindingPolicy`, `Binding`, `Status Collector`, `CombinedStatus` and `CustomTransform`) that define how workloads are distributed
 - Maintains status information about deployed workloads
 - Acts as the main user interface to the KubeStellar system
-
-> A space is an abstraction (or a light weight version) of the typical Kubernetes control plane that includes just the Kubernetes API server and a subset of relevant controllers. There are different types of spaces and your choice depend on your use-case requirements.
 
 
 ## Creating a WDS
 
-Currently, we only support the use of our Kubestellar core Helm chart as the only way to  create a WDS. This is because the core Helm chart also automatically creates a kubestellar-controller-manager and transport-controller, which are contorllers that your WDS requires to function properly.
+Currently, we support the use of our Kubestellar core Helm chart as the only way to  create a WDS. This is because the core Helm chart also automatically creates a kubestellar-controller-manager and transport-controller, which are contorllers that your WDS requires to function properly.
 
 ### Use the KubeStellar Core Helm Chart to create your WDS
 
@@ -39,8 +37,8 @@ helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart 
 
 You can customize your WDS by specifying:  
 
-- `name`: A unique name for the WDS
-- `type`:  
+- `<wds-name>`: A unique name for the WDS
+- `<space-type>`:  
 
     - `k8s` (default): Creates a basic Kubernetes API Server with a subset of kube controllers
     - `host`: Uses the KubeFlex hosting cluster itself. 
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
Looking at https://github.com/Tob-iee/kubestellar/actions, I see that the "Generate and push" job successfully updated your `gh-pages` but the "pages-build-deployment" job did not run. Same mystery as in https://github.com/kubestellar/kubestellar/pull/3002#issuecomment-3017991204 , maybe.

I wonder if making any change, trivial or substantial, would get the second job to run too.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,87 @@
...
-- [Working with a WDS](#working-with-a-wds)
-  - [Accessing the WDS](#accessing-the-wds)
 
-A Workload Description Space (WDS) is a core component of the KubeStellar architecture that serves as the primary interface for users to define and manage workloads for multi-cluster deployment.
+
 
 ## What is a WDS?
 
-A WDS is a space (a Kubernetes-like API server with storage) that:
+A Workload Description Space (WDS) is a space in the [KubeStellar architecture](docs/content/direct/user-guide-intro.md) that serves as the primary interface for users to define and manage workloads for multi-cluster deployment. The WDS constitue of a Kubernetes API server with storage that:
```
MikeSpreitzer: image

docs/content/direct/wds.md
```diff
@@ -2,68 +2,87 @@
...
 
 - Stores the definitions of workloads in their native Kubernetes format
 - Hosts the control objects (`BindingPolicy` and `Binding`) that define how workloads are distributed
```
MikeSpreitzer: Quoting #3000:

> There are additional kinds of control objects that can appear in a WDS.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,87 @@
...
 - Maintains status information about deployed workloads
 - Acts as the main user interface to the KubeStellar system
 
+> A space is an abstraction (or a light weight version) of the typical Kubernetes control plane that includes just the Kubernetes API server and a subset of relevant controllers. There are different types of spaces and your choice depend on your use-case requirements.
```
MikeSpreitzer: Is this a quote from something? If so, what? If not, do not set off with `> `.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,87 @@
...
+
+
 ## Creating a WDS
 
-A WDS can be created in several ways:
+Currently, we only support the use of our Kubestellar core Helm chart as the only way to  create a WDS. This is because the core Helm chart also automatically creates a kubestellar-controller-manager and transport-controller, which are contorllers that your WDS requires to function properly.
```
MikeSpreitzer: Not great grammar: double use of "only" here.

Not a blocker.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,87 @@
...
 
-### Using the KubeStellar Core Helm Chart
+### Use the KubeStellar Core Helm Chart to create your WDS
 
 The recommended approach is to use the KubeStellar Core Chart:
```
MikeSpreitzer, line 31: I think that it is worthwhile to be very clear here that there are two different scenarios. The displayed command is not good for either one of them. One scenario is the normal usage of the core Helm chart to (1) install KubeFlex in the KubeFlex hosting cluster, (2) create some ITSes, and (3) create some WDSes. The other scenario happens after the first, and is using the core Helm chart to _add_ some WDS(es) to an already-existing system.

The existing text has this defect, which does not need to be fixed in this PR.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,87 @@
...
 
 ```shell
 helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart \
-  --set-json='WDSes=[{"name":"wds1", "type":"k8s"}]'
+  --set-json='WDSes=[{"name":"<wds-name>", "type":"<space-type>"}]'
```
MikeSpreitzer, line 35: The variable names in angle brackets here should be the same as the variable names listed below.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,87 @@
...
-- `type`: 
-  - `k8s` (default): Creates a basic Kubernetes API Server with a subset of kube controllers
-  - `host`: Uses the KubeFlex hosting cluster itself
+- `type`:  
+
+    - `k8s` (default): Creates a basic Kubernetes API Server with a subset of kube controllers
+    - `host`: Uses the KubeFlex hosting cluster itself. 
+    
 - `APIGroups`: A comma-separated list of API Groups to include
 - `ITSName`: The name of the ITS to be used by this WDS (required if multiple ITSes exist)
```
MikeSpreitzer, line 47: These do not appear in the exhibited command. Probably because of https://github.com/kubestellar/kubestellar/pull/3023/files#r2178389286 .

The existing text has this defect, which does not need to be fixed in this PR.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,87 @@
...
 
-### Using the KubeFlex CLI
+The type of space you choose determines the type of controllers that are included in your your WDS. However, when creating a WDS, the recommended space types are "K8s" (the default space) and "host". If you the decide to use your hosting cluster's control plane (which is typically a KubeFlex cluster for Kubstellar) by specifying `type: host`, it presents you some benefits such as:
```
MikeSpreitzer: "However" is not the right connector here, the sentence that it introduces is not in tension with the previous remark. I would simply remove that word.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,87 @@
...
 
-You can also create a WDS using the KubeFlex CLI:
+- Prevents you from creating a separate control plane
```
MikeSpreitzer: "Prevents you" is a bit of a misdirection. The user does not directly create ControlPlane objects, that is done indirectly through the chart that this text is talking about. I think a more on-point description would be something like one of the following.

> Avoids creating a separate space.

> Saves the cost of creating an additional space.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,87 @@
...
+- Simplifies the architecture by reusing the hosting cluster
+- Makes the WDS directly accessible through the hosting cluster's API server
 
-```shell
-kflex create wds1 -p kubestellar
-```
+> You can create multiple WDSes of the same space types and even a mix of different space types, where you can have the default and host WDSes in your cluster. A multiple WDS architecure can be valueable in cases where you want to manage workload runs across different sets of users or groups. This each users or groups can use thesame ITS for workload execution but with their different WDS
```
MikeSpreitzer: s/This/Thus/

s/each users or groups/each user or group/

s/thesame/the same/

s/for workload execution/for WEC inventory/

Is this paragraph a quote? If so, from what? If not, then do not set off with `> `.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,87 @@
...
 
-This command creates a WDS and runs a post-create hook that deploys the KubeStellar controller manager and transport controller.
+### Accessing the WDS
 
-## KubeFlex Hosting Cluster as WDS
+After creating your WDS, you will need to access it to do things like submit placements or workload definitions. There are two ways that you can do this, either by:
```
MikeSpreitzer: KubeStellar does not use OCM `Placement`.

MikeSpreitzer: The two displayed commands do different things; they are not simply alternatives to each other.

Both commands leave the current kubeconfig context set to the context for accessing the WDS.

The first command also establishes that kubeconfig context, by creating it if it does not already exist and otherwise overwriting it with the right contents. The second command relies on this context already existing with the right contents.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,87 @@
...
+```
+
+Both methods provide flexibility depending on your preferred way of working with the system.
+
+## Working with a WDS
+
+Once your WDS is created, you can:
+
+1. **Create workload objects** in their native Kubernetes format
+2. **Define BindingPolicy objects** to specify which workloads should be deployed to which WECs
```
MikeSpreitzer, line 82: There are also other kinds of control objects.

MikeSpreitzer:
Please fix the reference to Issue #3000 in the opening comment.

docs/content/direct/wds.md
Tob-iee: I just wanted to highlight and remind the user what a "space" is, but I think I will just take the entire thing out since it's already explained in the architecture page.

Tob-iee:
> Please fix the reference to Issue #3000 in the opening comment.

@MikeSpreitzer I have addressed the comments you mentioned. Kindly review these new changes when you can get to it.

MikeSpreitzer:
@Tob-iee: there is a difference between

```
Fixes #3000
```

and

```
3000

Fixes #
```

GitHub recognizes the first as asserting that this PR resolves Issue 3000, but does not recognize the second.

If you intend this relationship then you will need to edit the opening comment of this PR (https://github.com/kubestellar/kubestellar/pull/3023#issue-3193482971).

MikeSpreitzer:
This PR overlaps with #2993. I have asked the author of that PR (@rishi-jat) to factor it into several smaller PRs. Please collaborate with him on updates to the WDS page.

PR #2935, opened 2025-05-22
make changes with lookup so that no duplicate pch installed

Rupam-It:
## Related issue(s)

Fixes #

Previous Implementation: 
```
helm install ks-core ./core-chart \
  --namespace kubeflex-system \
  --set "kubeflex-operator.hostContainer=kubeflex-control-plane" \
  --set "kubeflex-operator.externalPort=9443" \
  --set-json='ITSes=[{"name":"its1"}]' \
  --set-json='WDSes=[{"name":"wds1","ITSName":"its1"}]'
```
output: 
```
NAME: ks-core
LAST DEPLOYED: Thu May 22 22:06:49 2025
NAMESPACE: kubeflex-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
```
but 
```
helm install ks-core-2 ./core-chart --namespace kubeflex-system \
  --set "kubeflex-operator.install=false" \
  --set "kubeflex-operator.hostContainer=kubeflex-control-plane" \
  --set "kubeflex-operator.externalPort=9443" \
  --set-json='ITSes=[{"name":"its2"}]' \
  --set-json='WDSes=[{"name":"wds2","ITSName":"its2"}]'
```
Error: INSTALLATION FAILED: Unable to continue with install: PostCreateHook "its-with-clusteradm" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "ks-core-2": current value is "ks-core"
you got this error !

```
ks-core-2 ./core-chart --namespace kubeflex-system \
  --set "InstallPCHs=false" \
  --set "kubeflex-operator.install=false" \
  --set "kubeflex-operator.hostContainer=kubeflex-control-plane" \
  --set "kubeflex-operator.externalPort=9443" \
  --set-json='ITSes=[{"name":"its2"}]' \
  --set-json='WDSes=[{"name":"wds2","ITSName":"its2"}]'
```
as we set manually  ` --set "InstallPCHs=false" \` you can see this is working fine! 
```
NAME: ks-core-2
LAST DEPLOYED: Thu May 22 22:12:38 2025
NAMESPACE: kubeflex-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
```
current Implementation: 

```
helm install ks-core ./core-chart \
  --namespace kubeflex-system \
  --set "kubeflex-operator.hostContainer=kubeflex-control-plane" \
  --set "kubeflex-operator.externalPort=9443" \
[...]

PR #3028, opened 2025-07-02
✨implement release-specific PostCreateHook naming for multi-instance support

### LINKED ISSUE #3021: feature: Update the CP definition of PCHs in core-chart template
### Feature Description

Update the way PCHs are used by `core-chart` to use the new multi-hook definition introduced by https://github.com/kubestellar/kubestellar/issues/2481

### Proposed Solution

Update the way the PCHs and their variables are referenced by `its.yaml` and `wds.yam' in
https://github.com/kubestellar/kubestellar/tree/main/core-chart/templates/controlplanes

### Want to contribute?

### Additional Context

_No response_

antedotee:
## Summary
Updated PostCreateHook (PCH) implementation to use release-specific multi-hook definitions, enabling multiple core-chart instantiations without conflicts.

## Problem
PostCreateHook resources were cluster-scoped with fixed names (`wds`, `its-with-clusteradm`, `its-without-clusteradm`), preventing multiple Helm releases from having different hook definitions. This created a "footgun" where multiple instantiations of the core-chart could disagree on PCH definitions, requiring users to disable PCHs or update the initial chart before creating new WDSes/ITSes with secondary charts.

## Solution
Modified PCH naming strategy to incorporate Helm release names, making each hook definition unique per release while maintaining all existing functionality and template logic.

## Changes
Updated 5 template files to implement release-specific PCH naming:

**PostCreateHook Definitions (3 files):**
- `core-chart/templates/postcreatehooks/wds.yaml` - `wds` → `{{ .Release.Name }}-wds`
- `core-chart/templates/postcreatehooks/its-with-clusteradm.yaml` - `its-with-clusteradm` → `{{ .Release.Name }}-its-with-clusteradm`
- `core-chart/templates/postcreatehooks/its-without-clusteradm.yaml` - `its-without-clusteradm` → `{{ .Release.Name }}-its-without-clusteradm`

**ControlPlane References (2 files):**
- `core-chart/templates/controlplanes/its.yaml` - Updated `postCreateHook` to use `{{ (eq $cp.install_clusteradm false) | ternary (printf "%s-its-without-clusteradm" $.Release.Name) (printf "%s-its-with-clusteradm" $.Release.Name) }}`
- `core-chart/templates/controlplanes/wds.yaml` - Updated `postCreateHook` to use `{{ printf "%s-wds" $.Release.Name }}`

**Key improvements:**
- Each Helm release now creates uniquely named PCHs (e.g., `my-release-wds`, `my-release-its-with-clusteradm`)
- ControlPlane resources correctly reference their release-specific hooks
- All existing template content, labels, and metadata remain unchanged
- Maintains backward compatibility while enabling new multi-instance capability
- Follows clean code principles with meaningful, descriptive naming patterns

**Results:**
- Eliminates cluster-scoped PCH conflicts between multiple chart instantiations
- Enables independent deployment of core-chart with different PCH configurations
- Provides clear separation of concerns between different Helm releases
- Improves deployment flexibility and reduces operational complexity

Fixes #3021

### DIFF
--- a/core-chart/templates/controlplanes/its.yaml
+++ b/core-chart/templates/controlplanes/its.yaml
@@ -7,7 +7,7 @@ metadata:
 spec:
   backend: shared
   type: {{ $cp.type | default "vcluster" }}
-  postCreateHook: {{ (eq $cp.install_clusteradm false) | ternary "its-without-clusteradm" "its-with-clusteradm" }}
+  postCreateHook: {{ (eq $cp.install_clusteradm false) | ternary (printf "%s-its-without-clusteradm" $.Release.Name) (printf "%s-its-with-clusteradm" $.Release.Name) }}
   postCreateHookVars:
     ITSSecretName: {{ or (eq $cp.type "host") (eq $cp.type "external") | ternary "admin-kubeconfig" "vc-vcluster" }}
     ITSkubeconfig: {{ or (eq $cp.type "host") (eq $cp.type "external") | ternary "kubeconfig-incluster" "config-incluster" }}
--- a/core-chart/templates/controlplanes/wds.yaml
+++ b/core-chart/templates/controlplanes/wds.yaml
@@ -7,7 +7,7 @@ metadata:
 spec:
   backend: shared
   type: {{ $cp.type | default "k8s" }}
-  postCreateHook: wds
+  postCreateHook: {{ printf "%s-wds" $.Release.Name }}
   postCreateHookVars:
     ITSName: "{{ $cp.ITSName | default "" }}"
     APIGroups: "{{ $cp.APIGroups | default "" }}"
--- a/core-chart/templates/postcreatehooks/its-with-clusteradm.yaml
+++ b/core-chart/templates/postcreatehooks/its-with-clusteradm.yaml
@@ -2,7 +2,7 @@
 apiVersion: tenancy.kflex.kubestellar.org/v1alpha1
 kind: PostCreateHook
 metadata:
-  name: its-with-clusteradm
+  name: {{ .Release.Name }}-its-with-clusteradm
   labels:
     kflex.kubestellar.io/cptype: its
 spec:
--- a/core-chart/templates/postcreatehooks/its-without-clusteradm.yaml
+++ b/core-chart/templates/postcreatehooks/its-without-clusteradm.yaml
@@ -2,7 +2,7 @@
 apiVersion: tenancy.kflex.kubestellar.org/v1alpha1
 kind: PostCreateHook
 metadata:
-  name: its-without-clusteradm
+  name: {{ .Release.Name }}-its-without-clusteradm
   labels:
     kflex.kubestellar.io/cptype: its
 spec:
--- a/core-chart/templates/postcreatehooks/wds.yaml
+++ b/core-chart/templates/postcreatehooks/wds.yaml
@@ -2,7 +2,7 @@
 apiVersion: tenancy.kflex.kubestellar.org/v1alpha1
 kind: PostCreateHook
 metadata:
-  name: wds
+  name: {{ .Release.Name }}-wds
   labels:
     kflex.kubestellar.io/cptype: wds
 spec:


francostellari:
@antedotee I'm not sure that prefixing the `.ReleaseName` to PCH is necessarily a good architecture decision.

It impacts how new WDSes are added .... see WDS2 discussion here: https://github.com/kubestellar/kubestellar/blob/main/docs/content/direct/core-chart.md

Keeping a name independent of the release allows to reuse the same PCH consistently.
On the other hand having multiple copies of PCHs in the cluster allows multiple Helm charts to deal with them independently.

@MikeSpreitzer what do you think about this change?

MikeSpreitzer:
If I recall correctly, we have a problem with some per-space configuration choices leaking into these PCHes?

francostellari:
> If I recall correctly, we have a problem with some per-space configuration choices leaking into these PCHes?

@MikeSpreitzer can you clarify what the issue is and how this would solve it?

MikeSpreitzer:
The problem is references to `{{ .Values... }}` in the PCH templates. Just one has this problem, the template for the WDS PCH.

francostellari:
@MikeSpreitzer, I still don't have a clear understanding.

The WDS PCH has 2 uses of {{ .Values...}}:
1. set the container tags
2. set the debug/verbosity levels for the controllers

This PR seems to change the the name of the hook to make it unique to the chart name used during installations

Do you have a scenario where you install different WDSes with different charts and also need to change those parameters between the charts?

francostellari:
BTW, this PR does not seem to be fixing the issue that it claims to fix https://github.com/kubestellar/kubestellar/issues/3021

antedotee:
I am thinking about passing all configuration through postCreateHookVars, which will make PCH generic and reusable compared earlier to .Values which was creating tight coupling. Am I thinking in the right direction?
(Sorry for replying after so much time, my laptop broke down)

MikeSpreitzer:
> The WDS PCH has 2 uses of {{ .Values...}}: ...

I see more than two.

```console
mspreitz@mjs13 kubestellar % grep -w Values core-chart/templates/postcreatehooks/wds.yaml 
{{- if .Values.InstallPCHs }}
            image: quay.io/kubestellar/kubectl:{{.Values.KUBECTL_VERSION}}
            image: quay.io/kubestellar/kubectl:{{.Values.KUBECTL_VERSION}}
            image: ghcr.io/kubestellar/kubestellar/ocm-transport-controller:{{ .Values.TRANSPORT_VERSION | default .Values.KUBESTELLAR_VERSION }}
            - --metrics-bind-address={{.Values.transport_controller.metrics_bind_addr}}
            - --pprof-bind-address={{.Values.transport_controller.pprof_bind_addr}}
            - --transport-qps={{.Values.transport_controller.transport_qps}}
            - --transport-burst={{.Values.transport_controller.transport_burst}}
            - --wds-qps={{.Values.transport_controller.wds_qps}}
            - --wds-burst={{.Values.transport_controller.wds_burst}}
            - -v={{.Values.verbosity.transport | default .Values.verbosity.default | default 4 }}
            - --max-num-wrapped={{.Values.transport_controller.max_num_wrapped}}
            - --max-size-wrapped={{.Values.transport_controller.max_size_wrapped}}
                - -v={{.Values.verbosity.kubestellar | default .Values.verbosity.default | default 2 }}
              image: ghcr.io/kubestellar/kubestellar/controller-manager:{{.Values.KUBESTELLAR_VERSION}}
```

MikeSpreitzer:
> It impacts how new WDSes are added .... see WDS2 discussion here: https://github.com/kubestellar/kubestellar/blob/main/docs/content/direct/core-chart.md

@francostellari, I am not sure that I can identify the intended discussion within that doc. Can you please say what is the difficulty? Or are you saying only that this PR would change how a user makes a second instance of the core Helm chart? Obviously this PR would simplify that, removing the need for having an `InstallPCHs` value; did you mean to refer to some additional change?

MikeSpreitzer:
I am of a divided mind regarding the main idea in this PR. On the one hand, it allows the chart to become simpler (not having the PCHes be _optional_). On the other hand, it makes each chart instance heavier weight. That would be a problem if the user wanted to scale up to a large number of WDSes or minimize the latency of their addition or removal. However, in such a case perhaps the inherent costs of using Helm would be too much and a non-Helm-based solution developed instead.

francostellari:
> Obviously this PR would simplify that, removing the need for having an `InstallPCHs` value; did you mean to refer to some additional change?

We already have a solution for that with the use of the `lookup` function: https://github.com/kubestellar/kubestellar/pull/2935

francostellari:
> I am of a divided mind regarding the main idea in this PR. On the one hand, it allows the chart to become simpler (not having the PCHes be _optional_). On the other hand, it makes each chart instance heavier weight. That would be a problem if the user wanted to scale up to a large number of WDSes or minimize the latency of their addition or removal. However, in such a case perhaps the inherent costs of using Helm would be too much and a non-Helm-based solution developed instead.

In my opinion if the goal is to simplify use by not requiring the suer to specify the `InstallPCH` values when adding charts... I think the aforementioned approach with `lookup` function is better and does not add more resources to the cluster.

In my opinion, we need to to justify having multiple PCHs which could create more confusion.

francostellari:
Personally, I don't like to have multiple PCHs potentially identical but with different names.
I prefer the approach proposed here here https://github.com/kubestellar/kubestellar/pull/2935#issuecomment-3090702045
@Rupam-It can you make progress on it?

clubanderson:
/ok-to-test

clubanderson:
please fix conflicts

Issue #3018, opened 2025-07-01
feature:  Detect Duplicate Object Identifiers Across WDSes


PranjaliBhardwaj:
### Feature Description

Currently, if two objects on different WDSes (Workload Distribution Systems) have the same group, version, kind, name, and namespace, this conflict is not detected or reported. This can lead to undefined or problematic behavior.

### Proposed Solution

Implement a check in the controller to detect when two objects on different WDSes have the same group, version, kind, name, and namespace.
Emit a warning or error event when such a conflict is detected.
 Identify Where to Add the Check
Implement the Duplicate Detection

### Want to contribute?

- [x] I would like to work on this issue.

PranjaliBhardwaj:
i have created pull request for same. @rishi-jat

rishi-jat:
Thanks @PranjaliBhardwaj, since the PR is already open, I’ll unassign myself.

MikeSpreitzer:
We deliberately did not attempt something like this earlier because of scalability concerns.

Also because it is not necessarily an error condition (e.g., if there is no content conflict). Consider the use case of layered software, with different layers in different WDSes. For context-creating objects such as CRDs and Namespaces, they have to be in both layers (the one providing a servers and the one consuming a service).

rishi-jat:
Hi @PranjaliBhardwaj, I noticed there hasn’t been any recent activity on this issue or its linked PR. If you’re no longer actively working on it, would it be okay if I picked it up?

clubanderson:
@PranjaliBhardwaj we can give you 24hrs to reply before releasing this back to the queue for re-assignment. Please respond in your PR. Thank you

PranjaliBhardwaj:
@clubanderson created pr!

MikeSpreitzer:
> if two objects on different WDSes (Workload Distribution Systems) have the same group, version, kind, name, and namespace, this conflict is not detected or reported.

The conflict happens regardless of API version, right?

MikeSpreitzer:
> This can lead to undefined or problematic behavior.

[...]

PR #3019, opened 2025-07-01
✨  Detect and Warn on Duplicate Object Identifiers Across WDSes

PranjaliBhardwaj:
This feature will:
Add a check in the controller to detect such duplicates.
Emit a warning or error event when a conflict is found.
issue targeting: #3018 

[ ] Duplicate object identifiers across WDSes are detected at sync time.
[ ] A clear warning or error is emitted (event/log).
[ ] Tests are added to cover this scenario.

Issue #3017, opened 2025-07-01
bug: Permissions botches in website workflows


MikeSpreitzer:
### Describe the bug

Attempts to use https://github.com/kubestellar/kubestellar/blob/main/.github/workflows/docs-gen-and-push.yml now fail.

An example is https://github.com/MikeSpreitzer/kcp-edge-mc/actions/runs/15992299097/workflow , which shows this complaint:

> The workflow is not valid. .github/workflows/docs-gen-and-push.yml (Line: 61, Col: 3): Error calling workflow 'MikeSpreitzer/kcp-edge-mc/.github/workflows/broken-links-crawler.yml@044a2b19645591b974ec3aedf5c9a8d43eb69f43'. The nested job 'broken-links-crawler' is requesting 'contents: write', but is only allowed 'contents: read'.

### Output from KubeStellar-Snapshot.sh

N/A

### Steps To Reproduce

Try to invoke the website rendering job on PR #3010 or anything later.

### Expected Behavior

Successful website rendering

### Want to contribute?

- [ ] I would like to work on this issue.

MikeSpreitzer:
@antedotee

MikeSpreitzer:
@KPRoche

MikeSpreitzer:
In #3016 I made docs-gen-and-push.yml give contents:write permission to broken-links-crawler.yml.

That got me to the next failure, in docs-gen-and-push.yml. https://github.com/MikeSpreitzer/kcp-edge-mc/actions/runs/15992569571/job/45108612558 shows the following.

```
error: failed to push branch gh-pages to origin: "remote: Permission to MikeSpreitzer/kcp-edge-mc.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/MikeSpreitzer/kcp-edge-mc/': The requested URL returned error: 403"
make: *** [Makefile:294: deploy-docs] Error 1
Error: Process completed with exit code 2.
```

PR #3016, opened 2025-07-01
✨  Various improvements mainly wrt GitHub Actions workflows

MikeSpreitzer:
## Summary
This PR makes a few related changes centered around the checking of the versions of the GitHub Actions referenced by workflows.

- Add an operation to the `hack/gha-reversemap.sh` script to do a full check of the intended discipline.
- Generalized that script to also process workflows whose filename ends with `.yaml` (since those are allowed too).
- Updated `.github/workflows/verify-action-hashes.yaml` to use the new check and to run on relevant PR events too.
- Fixed a violation uncovered by the new check.
- Updated the contributing guidlines to describe the intended discipline.
- Updated some outdated information in the same document.
- Added a list of good practices for PRs, addressing problems that have been seeing.
- Fixed two botches in workflow permissions introduced in #3010 .

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-better-verify-actions/contribution-guidelines/contributing-inc/#pull-requests

## Related issue(s)

Fixes #3017

PR #3030, opened 2025-07-02
✨ Detect and Warn on Duplicate Object Identifiers Across WDSes #3018

PranjaliBhardwaj:
Track all UIDs seen in the current call to getWrapeesFromWDS.
If a UID is seen more than once (across all objects in the binding's workload), log a warning with the duplicate UID and the object references.
Include the WDS name in the log for clarity.
#3018


MikeSpreitzer:
We deliberately did not attempt something like this earlier because of scalability concerns.

Also because it is not necessarily an error condition (e.g., if there is no content conflict). Consider the use case of layered software, with different layers in different WDSes. For context-creating objects such as CRDs and Namespaces, they have to be in both layers (the one providing a servers and the one consuming a service).

MikeSpreitzer:
Also, rebase onto the latest `main` to pick up fixes to #3017 .

MikeSpreitzer:
Please collaborate on the existing PR (#3019).

MikeSpreitzer:
Oh, wait, sorry, I did not notice that this and #3019 are from the same author.

.github/workflows/docs-gen-and-push.yml
```diff
@@ -25,6 +25,8 @@ jobs:
   debug:
     name: print relevant info
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
MikeSpreitzer, line 29: It looks like there is some confusion wrt recent changes, this is undoing #3016. I recommend rebasing onto the latest `main` or starting over from the latest `main`.

MikeSpreitzer:
https://github.com/kubestellar/kubestellar/blob/17d0b129e0092dceddfe7159fa376527492d785f/CONTRIBUTING.md?plain=1#L73

MikeSpreitzer (changes requested):
Please revise this to not include any merges, and not undo any recent changes.

MikeSpreitzer:
Also, note that we expect that every PR conforms to the discipline done by `make imports`.

.github/workflows/verify-action-hashes.yaml
```diff
@@ -30,4 +30,5 @@ jobs:
 
       - name: Run test
         run: |
+
```
MikeSpreitzer, line 33: Why add a blank line here?

pkg/transport/generic/generic_transport_controller.go
```diff
@@ -1318,4 +1334,4 @@ func customTransformToDomain(obj any) ([]string, error) {
 
 func customTransformDomainKey(apiGroup, resource string) string {
 	return apiGroup + "/" + resource
-}
+}
```
MikeSpreitzer, line 1337: Why remove the final linebreak?

pkg/transport/generic/generic_transport_controller.go
```diff
@@ -733,13 +733,29 @@ func (c *genericTransportController) getWrapeesFromWDS(ctx context.Context, bind
 	groupResources := sets.New[metav1.GroupResource]()
 	wrapees := make([]WrapeeWithUID, 0)
 	kindToResource := map[schema.GroupKind]string{}
+
+	// Map from UID to list of object references (for duplicate detection)
+	uidToRefs := make(map[string][]string)
```
MikeSpreitzer, line 738: This func (`getWrapeesFromWDS`) is only applied to one WDS at a time. There will be no duplications within one WDS (https://github.com/kubestellar/kubestellar/blob/17d0b129e0092dceddfe7159fa376527492d785f/api/control/v1alpha1/types.go#L290 , https://github.com/kubestellar/kubestellar/blob/17d0b129e0092dceddfe7159fa376527492d785f/api/control/v1alpha1/types.go#L299) and even if there were, they would not be the 'droids that #3018 is looking for.

clubanderson:
@PranjaliBhardwaj we can give you 24hrs to reply before releasing this back to the queue for re-assignment. Please respond in your PR. Thank you

PranjaliBhardwaj:
ok correcting this pr, my sem exams are finally over

PR #3040, opened 2025-07-02
🐛 add shellcheck to setup-shell.sh

AritraDey-Dev:
## Summary
 add missing shebang to setup-shell.sh

### DIFF
--- a/test/e2e/common/setup-shell.sh
+++ b/test/e2e/common/setup-shell.sh
@@ -1,4 +1,3 @@
-# shellcheck shell=bash
 # Copyright 2024 The KubeStellar Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
The shebang is not needed, because this file is invoked by the `source` command. Adding a shebang would be misleading.

AritraDey-Dev:
> The shebang is not needed, because this file is invoked by the `source` command. Adding a shebang would be misleading.

Then adding` # shellcheck shell=bash` at the top of the file is worthwhile, as it informs linters and editors of the intended shell syntax.

MikeSpreitzer:
@AritraDey-Dev: you can modify your branch to make that change.

MikeSpreitzer:
https://github.com/kubestellar/kubestellar/blob/main/hack/boilerplate/boilerplate.sh.txt is the thing

MikeSpreitzer:
@AritraDey-Dev: ping

MikeSpreitzer:
Ping @AritraDey-Dev

MikeSpreitzer (approved):
/lgtm
/approve

PR #3006, opened 2025-06-24
✨feat: add docs preview enforcement workflow and PR template 

rishi-jat:
This PR adds a GitHub Actions workflow and a supporting PR template to enforce documentation PR best practices in KubeStellar.

### 📋 What's Included

- **Workflow**: Automatically checks PRs that modify documentation files.
  - If docs are changed, the PR must include a `Preview:` line with a valid GitHub Pages URL.
  - Warns if the branch name doesn't start with `doc-`, as automatic previews rely on that.
  - Fails the check if no preview link is provided.

- **PR Template**: Adds a clear instruction block to guide contributors to include the correct preview format.

### 📎 Motivation

Ensures that all documentation PRs are previewable, consistent, and easier to review. Reduces back-and-forth and improves contributor experience.

### 🔗 Related Issue

Fixes #3001

Issue #2649, opened 2024-12-05
bug: Getting Started get csr fails


seb-codes:
### Describe the bug

From Getting Started 
Repeatedly issue the command:
    - kubectl --context its1 get csr
until you see that the certificate signing requests (CSR) for both cluster1 and cluster2 exist. Note that the CSRs condition is supposed to be Pending until you approve them in step 4.

This command `kubectl --context its1 get csr` returns "No resources found"

### Steps To Reproduce

fresh install and verified pre-req check for kubestellar

In Getting Started guide, completed up to 

Create and register two workload execution cluster(s) portion

ran 
- : set flags to "" if you have installed KubeStellar on an OpenShift cluster
flags="--force-internal-endpoint-lookup"
clusters=(cluster1 cluster2);
for cluster in "${clusters[@]}"; do
   kind create cluster --name ${cluster}
   kubectl config rename-context kind-${cluster} ${cluster}
   clusteradm --context its1 get token | grep '^clusteradm join' | sed "s/<cluster_name>/${cluster}/" | awk '{print $0 " --context '${cluster}' --singleton '${flags}'"}' | sh
done

ran 
- kubectl --context its1 get csr

### Expected Behavior

the certificate signing requests (CSR) for both cluster1 and cluster2 exist. Note that the CSRs condition is supposed to be Pending

### Want to contribute?

- [ ] I would like to work on this issue.

clubanderson:
@seb-codes can you try with our latest demo env script?

bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/heads/main/scripts/create-kubestellar-demo-env.sh) --platform kind

MikeSpreitzer:
@clubanderson: @seb-codes is talking about the step-by-step instructions, not the create-kubestellar-demo-env script.

@seb-codes: I do not understand what you think the bug is. The instructions say to repeat the command until the CSRs exist. Clearly that implies that they might not exist at first.

PR #3051, opened 2025-07-07
📖 revise WEC registration based on feedback from #2997 

rishi-jat:
## Summary  
This PR enhances the WEC registration documentation by incorporating all feedback provided in the review of PR #2997.

## What’s Improved  
- ✅ Mapped OCM terminology with references to official documentation  
- ✅ Corrected typo in the introduction (“what how to register” → “how to register”)  
- ✅ Improved explanation of verification commands  
- ✅ Reduced duplication across Kind, K3d, and OpenShift sections  
- ✅ Added clear cleanup instructions for both agent namespaces  
- ✅ Clarified labeling examples as illustrative only  
- ✅ Included detailed troubleshooting and management sections  
- ✅ Refined structure and overall readability for better comprehension

## Live Preview  
Preview → https://rishi-jat.github.io/kubestellar/doc-wec-registration/direct/wec-registration/
## Related  
Addresses feedback from PR #2997 by @MikeSpreitzer

### DIFF
--- a/docs/content/direct/wec-registration.md
+++ b/docs/content/direct/wec-registration.md
@@ -6,10 +6,14 @@ This document explains how to register a Workload Execution Cluster (WEC) with a
 
 Registering a WEC in an ITS is the same process as registering a managed cluster with an OCM hub cluster. KubeStellar uses Open Cluster Management (OCM) for cluster registration and management.
 
-**Terminology Mapping:**
-- **OCM Hub** → **KubeStellar ITS** (Inventory and Transport Space)
-- **OCM Managed Cluster** → **KubeStellar WEC** (Workload Execution Cluster)
-- **OCM Agent** → **OCM Agent** (same component, installed on WEC)
+
+### Terminology Mapping
+
+| OCM Term             | KubeStellar Equivalent                        |
+|----------------------|-----------------------------------------------|
+| **OCM Hub**          | **KubeStellar ITS** (Inventory and Transport Space) |
+| **OCM Managed Cluster** | **KubeStellar WEC** (Workload Execution Cluster) |
+| **OCM Agent**        | **OCM Agent** (same component, installed on WEC) |
 
 For the complete OCM registration process, refer to the [official Open Cluster Management documentation](https://open-cluster-management.io/docs/getting-started/installation/register-a-cluster/).
 
@@ -162,30 +166,8 @@ These properties can be used for rule-based transformations when workloads are d
 
 ### Local Development Clusters (Kind/K3d)
 
-For local development with Kind or K3d clusters:
-
-```shell
-# Create Kind cluster (context renaming is optional but convenient)
-kind create cluster --name cluster1
-kubectl config rename-context kind-cluster1 cluster1
-
-# Register with ITS (note the --force-internal-endpoint-lookup flag)
-clusteradm --context its1 get token | grep '^clusteradm join' | \
-  sed "s/<cluster_name>/cluster1/" | \
-  awk '{print $0 " --context cluster1 --singleton --force-internal-endpoint-lookup"}' | sh
-
-# Wait for CSR to appear, then accept registration
-clusteradm --context its1 accept --clusters cluster1
-```
-
-For K3d clusters, the process is identical except for the cluster creation:
-```shell
-# Create K3d cluster with port mapping
-k3d cluster create -p "9443:443@loadbalancer" cluster1
-kubectl config rename-context k3d-cluster1 cluster1
-
-# Then follow the same registration steps as above
-```
+For instructions on creating and registering local development clusters (Kind or K3d), refer to the  
+[Create and Register Two Workload Execution Clusters guide](https://docs.kubestellar.io/unreleased-development/direct/get-started/#create-and-register-two-workload-execution-clusters).
 
 ### OpenShift Clusters
 
--- a/docs/content/direct/wec-registration.md
+++ b/docs/content/direct/wec-registration.md
@@ -6,7 +6,7 @@ This document explains how to register a Workload Execution Cluster (WEC) with a
 
 Registering a WEC with an ITS follows the same process as registering a managed cluster with an OCM hub. KubeStellar uses Open Cluster Management (OCM) for cluster registration and management.
 
-For the complete OCM registration process, refer to the [official Open Cluster Management documentation](https://open-cluster-management.io/docs/getting-started/installation/register-a-cluster/). The instructions below provide a complete registration process that is more focused on KubeStellar-specific considerations.
+The instructions below provide a comprehensive registration process focused on KubeStellar-specific considerations. For additional OCM registration details, refer to the [official Open Cluster Management documentation](https://open-cluster-management.io/docs/getting-started/installation/register-a-cluster/).
 
 ### Terminology Mapping
 
@@ -20,12 +20,14 @@ For the complete OCM registration process, refer to the [official Open Cluster M
 
 Before registering a WEC, ensure you have:
 
-1. **A running Kubernetes cluster** that will serve as the WEC
-2. **Network connectivity** from the WEC to the KubeFlex hosting cluster (HTTPS connections must be possible)
-3. **`kubectl` access** to both the WEC and ITS
-4. **`clusteradm` CLI tool** installed on the machine where you'll run the registration commands ([installation guide](https://open-cluster-management.io/docs/getting-started/installation/start-the-control-plane/))
-5. **An existing ITS** with OCM cluster manager running
+1. **An existing ITS** with OCM cluster manager running
+2. **A running Kubernetes cluster** that will serve as the WEC
+3. **Network connectivity** from the WEC to the ITS (HTTPS connections must be possible)
+4. **`kubectl` access** to both the WEC and ITS
+5. **`clusteradm` CLI tool** installed on the machine where you will run the registration commands ([installation guide](https://open-cluster-management.io/docs/getting-started/installation/start-the-control-plane/))
 6. **Sufficient permissions** to create resources in both clusters (typically cluster-admin or equivalent)
+
+For a complete understanding of how WEC registration fits into the overall KubeStellar architecture, see [The Full Story](user-guide-intro.md#the-full-story).
 
 To verify your ITS is ready for WEC registration:
 
@@ -51,11 +53,13 @@ Obtain a registration token from the ITS:
 clusteradm --context <its-context> get token
 ```
 
-This command outputs a `clusteradm join` command that you'll use in the next step.
+This command outputs a `clusteradm join` command that you will use in the next step. The token and apiserver URL from this command will be used in the join command.
 
 ### Step 2: Join the WEC to the ITS
 
-Execute the join command on the WEC to initiate the registration process:
+**Important:** Before executing the join command, determine if you need additional flags based on your environment. The basic command below may not work for all cluster types.
+
+Execute the join command on the WEC to initiate the registration process. The exact command depends on your environment:
 
 ```shell
 # Basic join command (additional flags may be required based on your environment)
@@ -66,7 +70,7 @@ clusteradm join --hub-token <token> --hub-apiserver <api-server-url> --cluster-n
 - `--cluster-name`: Choose a unique name for your WEC
 - `--context`: Specify the kubectl context for your WEC
 - `--singleton`: Use this flag if the WEC is a single-node cluster
-- `--force-internal-endpoint-lookup`: Required for Kind clusters and other local development setups
+- `--force-internal-endpoint-lookup`: Required for Kind clusters and other clusters with internal-only API server endpoints
 
 **Example for Kind clusters:**
 ```shell
@@ -87,9 +91,12 @@ After the join command completes, wait for a Certificate Signing Request (CSR) t
 kubectl --context <its-context> get csr
 ```
 
-You should see a CSR with your WEC name and status `Pending`. This CSR is created by the OCM registration agent running on your WEC.
+The first few attempts might not show the CSR with your WEC name and status `Pending`. Continue checking until the CSR appears. This CSR is created by the OCM registration agent running on your WEC.
 
-**Wait for CSR to appear:**
+#### Loop that waits for Certificate Signing Request
+
+The following bash loop automates the waiting described just above:
+
 ```shell
 # Wait for CSR to be created (this may take a few moments)
 while [ -z "$(kubectl --context <its-context> get csr | grep <your-wec-name>)" ]; do
@@ -157,7 +164,7 @@ kubectl --context <its-context> label managedcluster <wec-name> team=platform bu
 
 ### Customization Properties
 
-You can define additional properties for each WEC using ConfigMaps in the `customization-properties` namespace:
+You can define additional properties for a WEC using ConfigMaps in the `customization-properties` namespace:
 
 ```shell
 # Create customization properties
--- a/docs/content/direct/wec-registration.md
+++ b/docs/content/direct/wec-registration.md
@@ -4,29 +4,28 @@ This document explains how to register a Workload Execution Cluster (WEC) with a
 
 ## Overview
 
-Registering a WEC in an ITS is the same process as registering a managed cluster with an OCM hub cluster. KubeStellar uses Open Cluster Management (OCM) for cluster registration and management.
+Registering a WEC with an ITS follows the same process as registering a managed cluster with an OCM hub. KubeStellar uses Open Cluster Management (OCM) for cluster registration and management.
 
+For the complete OCM registration process, refer to the [official Open Cluster Management documentation](https://open-cluster-management.io/docs/getting-started/installation/register-a-cluster/). The instructions below provide a complete registration process that is more focused on KubeStellar-specific considerations.
 
 ### Terminology Mapping
 
 | OCM Term             | KubeStellar Equivalent                        |
 |----------------------|-----------------------------------------------|
-| **OCM Hub**          | **KubeStellar ITS** (Inventory and Transport Space) |
-| **OCM Managed Cluster** | **KubeStellar WEC** (Workload Execution Cluster) |
-| **OCM Agent**        | **OCM Agent** (same component, installed on WEC) |
-
-For the complete OCM registration process, refer to the [official Open Cluster Management documentation](https://open-cluster-management.io/docs/getting-started/installation/register-a-cluster/).
+| **OCM Hub**          | **ITS** (Inventory and Transport Space) |
+| **OCM Managed Cluster** | **WEC** (Workload Execution Cluster) |
+| **Klusterlet**       | **Klusterlet** |
 
 ## Prerequisites
 
 Before registering a WEC, ensure you have:
 
 1. **A running Kubernetes cluster** that will serve as the WEC
-2. **Network connectivity** between the WEC and the ITS
+2. **Network connectivity** from the WEC to the KubeFlex hosting cluster (HTTPS connections must be possible)
 3. **`kubectl` access** to both the WEC and ITS
-4. **`clusteradm` CLI tool** installed ([installation guide](https://open-cluster-management.io/docs/getting-started/installation/start-the-control-plane/))
+4. **`clusteradm` CLI tool** installed on the machine where you'll run the registration commands ([installation guide](https://open-cluster-management.io/docs/getting-started/installation/start-the-control-plane/))
 5. **An existing ITS** with OCM cluster manager running
-6. **Sufficient permissions** to create resources in both clusters
+6. **Sufficient permissions** to create resources in both clusters (typically cluster-admin or equivalent)
 
 To verify your ITS is ready for WEC registration:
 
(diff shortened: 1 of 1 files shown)


rishi-jat:
Hey @MikeSpreitzer ! 👋
I’ve addressed all the feedback from #2997 and here’s a live preview of the updated WEC registration docs:
📄 Preview → https://rishi-jat.github.io/kubestellar/wec-preview/direct/wec-registration/
Let me know if anything else needs refinement!

MikeSpreitzer:
Note that this PR would fail the test introduced in #3006, because this PR does not have the preview reference _in the opening comment_ (the first comment in the PR, the one at the top). @rishi-jat , did you know that you can edit that?

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,325 @@
...
 
-TODO: write this.
+## Overview
+
+Registering a WEC in an ITS is the same process as registering a managed cluster with an OCM hub cluster. KubeStellar uses Open Cluster Management (OCM) for cluster registration and management.
+
+**Terminology Mapping:**
+- **OCM Hub** → **KubeStellar ITS** (Inventory and Transport Space)
+- **OCM Managed Cluster** → **KubeStellar WEC** (Workload Execution Cluster)
+- **OCM Agent** → **OCM Agent** (same component, installed on WEC)
```
MikeSpreitzer: This did not render as you intended.

image

Using a proper section heading would probably fix it.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,325 @@
...
+kubectl --context <its-context> get ServiceAccount default
+
+# Verify OCM cluster manager is running (look for pods in Running state)
+kubectl --context <its-context> get pods -n open-cluster-management-hub
+
+# Check for the customization-properties namespace
+kubectl --context <its-context> get ns customization-properties
+```
+
+## Registration Process
```
MikeSpreitzer, line 45: This is the same information as is in https://docs.kubestellar.io/unreleased-development/direct/get-started/#create-and-register-two-workload-execution-clusters . Rather than maintain two copies, we should maintain one copy and refer to it from elsewhere as needed.

MikeSpreitzer:
@Rupam-It: Since you worked on #2997, perhaps you should look at this.

MikeSpreitzer:
@rishi-jat: somehow the navigation tree of the `wec-preview` version of your fork of the website is all wrong. Compare. Here is the real website:
image
image

Here is your preview:
image
image

MikeSpreitzer:
Perhaps you would benefit from the `gh-pages` branch setup introduced in #3050 section "Enabling GitHub pages for a fork that already exists". Or perhaps you are somehow trying to run `mike` on your machine and push to GitHub. Don't. Just name your branch "doc-something" and let GitHub do the work.

rishi-jat:
Hi @MikeSpreitzer,
I’ve addressed all the suggested changes and updated the PR description with the correct preview link:
👉 https://rishi-jat.github.io/kubestellar/doc-wec-registration/direct/wec-registration/
Please let me know if there’s anything else you’d like to see updated!

docs/content/direct/wec-registration.md
MikeSpreitzer: I still see the information duplicated.

In addition to the copy introduced here, there is also the reference to the OCM documentation. The relationship between that reference and the copy that appears later is not explicitly stated. I think that it would be better if it were explicitly stated. The reference currently reads as follows.

> For the complete OCM registration process, refer to the [official Open Cluster Management documentation](https://open-cluster-management.io/docs/getting-started/installation/register-a-cluster/).

The implication is that the rest of this document is not going to give the complete OCM registration process --- part(s) will be missing. I think that's not exactly the true relationship. The copy shown later in this document _is_ a complete registration. The difference is that the official documentation is more comprehensive.

MikeSpreitzer:
I see that the gross problems with the navigation tree have been resolved.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,307 @@
...
+Registering a WEC in an ITS is the same process as registering a managed cluster with an OCM hub cluster. KubeStellar uses Open Cluster Management (OCM) for cluster registration and management.
+
+
+### Terminology Mapping
+
+| OCM Term             | KubeStellar Equivalent                        |
+|----------------------|-----------------------------------------------|
+| **OCM Hub**          | **KubeStellar ITS** (Inventory and Transport Space) |
+| **OCM Managed Cluster** | **KubeStellar WEC** (Workload Execution Cluster) |
+| **OCM Agent**        | **OCM Agent** (same component, installed on WEC) |
```
MikeSpreitzer: Actually, I think the OCM terms are "klusterlet" and "klusterlet agent".

```console
mspreitz@mjs13 kubestellar % kubectl --context cluster1 get pods -A | egrep NAME\|open-cluster
NAMESPACE                             NAME                                             READY   STATUS    RESTARTS   AGE
open-cluster-management-agent-addon   status-agent-5899d86d8b-sbbjf                    1/1     Running   0          173m
open-cluster-management-agent         klusterlet-agent-66b88c676d-nc2ff                1/1     Running   0          8h
open-cluster-management               klusterlet-7d8bd449cc-62qmr                      1/1     Running   0          8h
open-cluster-management               klusterlet-7d8bd449cc-b775c                      1/1     Running   0          8h
open-cluster-management               klusterlet-7d8bd449cc-nmpnq                      1/1     Running   0          8h
```

I tried searching the OCM documentation for "OCM agent", and came up empty. OTOH, OpenShift has something by that name, but it is an OpenShift-specific thing. To be fair, we may have some people and documentation using the term "OCM agent" when they mean klusterlet or klusterlet agent; they may be thinking that it is a short-hand for "OCM klusterlet agent". @pdettori, do you have an opinion on this?

Maybe you don't have to say anything about cases where KubeStellar and OCM use the same term?

I think that including the text "OCM" and "KubeStellar" in the rows below those headings is redundant.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,307 @@
...
+
+For the complete OCM registration process, refer to the [official Open Cluster Management documentation](https://open-cluster-management.io/docs/getting-started/installation/register-a-cluster/).
+
+## Prerequisites
+
+Before registering a WEC, ensure you have:
+
+1. **A running Kubernetes cluster** that will serve as the WEC
+2. **Network connectivity** between the WEC and the ITS
```
MikeSpreitzer: I think that it is true and worthwhile to say that what is required is that HTTPS connections can be opened from the WEC to the KubeFlex hosting cluster. It is important and deliberate choice that no more than this is required.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,307 @@
...
+3. **`kubectl` access** to both the WEC and ITS
+4. **`clusteradm` CLI tool** installed ([installation guide](https://open-cluster-management.io/docs/getting-started/installation/start-the-control-plane/))
```
MikeSpreitzer: There are several containers and (virtual or real) machines in sight, so it would be helpful to say _where_ this tool needs to be installed.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,307 @@
...
+5. **An existing ITS** with OCM cluster manager running
+6. **Sufficient permissions** to create resources in both clusters
```
MikeSpreitzer: IMO it would be helpful to list more specifically the privileges required. Unfortunately, I myself do not know the answer nor where it is documented by OCM. So, not a blocker.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,307 @@
...
+
+To verify your ITS is ready for WEC registration:
+
+```shell
+# Check if the ITS is accessible (this returns a short result)
+kubectl --context <its-context> get ServiceAccount default
+
+# Verify OCM cluster manager is running (look for pods in Running state)
```
MikeSpreitzer: Did you mean "look for absence of Pending or failing Pods"?

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,307 @@
...
+This command outputs a `clusteradm join` command that you'll use in the next step.
+
+### Step 2: Join the WEC to the ITS
+
+Execute the join command on the WEC to initiate the registration process:
+
+```shell
+# Basic join command
+clusteradm join --hub-token <token> --hub-apiserver <api-server-url> --cluster-name <your-wec-name> --context <wec-context>
+```
```
MikeSpreitzer: This does not hint that there might be additional flags required; this could lead an eager reader to make a mistake.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,307 @@
...
+
+**Example for cloud provider clusters:**
+```shell
+clusteradm join --hub-token <token> --hub-apiserver <api-server-url> --cluster-name prod-cluster-east --context prod-cluster-east
+```
+
+### Step 3: Wait for Certificate Signing Request
+
+After the join command completes, wait for a Certificate Signing Request (CSR) to appear on the ITS:
```
MikeSpreitzer, line 87: This text assumes more intelligence on the reader's part than the text in #3053 (hi #2649). Do we want to make different assumptions in different copies of this information? Do we want to include the `bash` wait loop in both copies? Are we getting tired of two copies yet?

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,307 @@
...
+# Register the cluster
+clusteradm join --hub-token <token> --hub-apiserver <its-api-server> \
+  --cluster-name <cloud-cluster-name> \
+  --context <cloud-cluster-context>
+
+# Accept from ITS
+clusteradm --context <its-context> accept --clusters <cloud-cluster-name>
+```
+
+## Troubleshooting
```
MikeSpreitzer: There is already a troubleshooting branch in the website. We should not have two separate bodies of troubleshooting information that show no recognition of each other.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,307 @@
...
+# Delete the ManagedCluster resource
+kubectl --context <its-context> delete managedcluster <wec-name>
+
+# Clean up the WEC (optional)
+kubectl --context <wec-context> delete namespace open-cluster-management-agent
+kubectl --context <wec-context> delete namespace open-cluster-management-agent-addon
+```
+
+## Next Steps
+
```
MikeSpreitzer, line 269: Shouldn't there be a pointer to the general big picture (https://rishi-jat.github.io/kubestellar/doc-wec-registration/direct/user-guide-intro/#the-full-story) somewhere in here?

rishi-jat:
I’ve addressed all the feedback and updated the PR accordingly. Please have a look and let me know if anything else is needed.

rishi-jat:
Hi @MikeSpreitzer — I’ve incorporated all your feedback and cleaned up the navigation and structure as well.
Please have a look when you get a chance and let me know if any further changes are needed.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,272 @@
...
+After the join command completes, wait for a Certificate Signing Request (CSR) to appear on the ITS:
+
+```shell
+# Check for pending certificate signing requests
+kubectl --context <its-context> get csr
+```
+
+You should see a CSR with your WEC name and status `Pending`. This CSR is created by the OCM registration agent running on your WEC.
+
+**Wait for CSR to appear:**
```
MikeSpreitzer: Stop making pseudo-headers with bold text. Use real section header syntax. Perhaps something like the following.

```
#### Loop that waits for Certificate Signing Request
```

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,272 @@
...
+### Step 3: Wait for Certificate Signing Request
+
+After the join command completes, wait for a Certificate Signing Request (CSR) to appear on the ITS:
+
+```shell
+# Check for pending certificate signing requests
+kubectl --context <its-context> get csr
+```
+
+You should see a CSR with your WEC name and status `Pending`. This CSR is created by the OCM registration agent running on your WEC.
```
MikeSpreitzer: No, "You should see ..." is not necessarily true. The first few `kubectl --context <its-context> get csr` might NOT show the CSR with your WEC name and status `Pending`. What you mean is to repeat this command until the expected CSR appears.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,272 @@
...
+
+**Wait for CSR to appear:**
+```shell
+# Wait for CSR to be created (this may take a few moments)
+while [ -z "$(kubectl --context <its-context> get csr | grep <your-wec-name>)" ]; do
+    echo "Waiting for CSR to appear..."
+    sleep 5
+done
+```
```
MikeSpreitzer, line 106: This is more or less duplicative with the earlier material. The reader should be explicitly told that. Perhaps introduce this loop with text like the following?

> The following `bash` loop automates the waiting described just above.

docs/content/direct/wec-registration.md
MikeSpreitzer: The current revision still starts by telling the reader to do something without warning about `--force-internal-endpoint-lookup`.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,272 @@
...
+```shell
+# Basic join command (additional flags may be required based on your environment)
+clusteradm join --hub-token <token> --hub-apiserver <api-server-url> --cluster-name <your-wec-name> --context <wec-context>
+```
+
+**Important flags:**
+- `--cluster-name`: Choose a unique name for your WEC
+- `--context`: Specify the kubectl context for your WEC
+- `--singleton`: Use this flag if the WEC is a single-node cluster
+- `--force-internal-endpoint-lookup`: Required for Kind clusters and other local development setups
```
MikeSpreitzer: This does not render as you intended. Make a practice of proof-reading your rendering.

image

docs/content/direct/wec-registration.md
MikeSpreitzer: The latest text still does not clarify that only one direction of connection-opening is required.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,272 @@
...
 
-This document will tell users what how to register a WEC in an ITS.
+This document explains how to register a Workload Execution Cluster (WEC) with an Inventory and Transport Space (ITS) in KubeStellar.
 
-TODO: write this.
+## Overview
+
+Registering a WEC with an ITS follows the same process as registering a managed cluster with an OCM hub. KubeStellar uses Open Cluster Management (OCM) for cluster registration and management.
+
+For the complete OCM registration process, refer to the [official Open Cluster Management documentation](https://open-cluster-management.io/docs/getting-started/installation/register-a-cluster/). The instructions below provide a complete registration process that is more focused on KubeStellar-specific considerations.
```
MikeSpreitzer: The problem here is that it leads off with "For the complete OCM registration process, refer to ...". That in itself implies that what follows is not a complete registration process. That is not the distinction that you want to make. "comprehensive documentation" means something different from "complete process". What follows _is_ a complete registration process. Do not start by implying otherwise. Start by saying what you mean. Even more importantly, do not start by talking about what is NOT coming. Start by talking about what _is_ coming. Finish with a reference to additional information.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,272 @@
...
+| **OCM Hub**          | **ITS** (Inventory and Transport Space) |
+| **OCM Managed Cluster** | **WEC** (Workload Execution Cluster) |
+| **Klusterlet**       | **Klusterlet** |
+
+## Prerequisites
+
+Before registering a WEC, ensure you have:
+
+1. **A running Kubernetes cluster** that will serve as the WEC
+2. **Network connectivity** from the WEC to the KubeFlex hosting cluster (HTTPS connections must be possible)
```
MikeSpreitzer: So that implies that the KubeFlex hosting cluster is also a pre-req, right? Actually, the missing pre-req is only that the Inventory and Transport Space exists --- remember that it might not be hosted by KubeFlex. See https://rishi-jat.github.io/kubestellar/doc-wec-registration/direct/user-guide-intro/#the-full-story . In fact, this might be the better place to refer the reader there.

Oh, I see that the ITS is listed later (currently number 5). I think that it makes sense to list the existence of the ITS before listing the ability to open connections into it.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,272 @@
...
+### Step 1: Get the Registration Token
+
+Obtain a registration token from the ITS:
+
+```shell
+# Get the registration token from the ITS
+clusteradm --context <its-context> get token
+```
+
+This command outputs a `clusteradm join` command that you'll use in the next step.
```
MikeSpreitzer: Pro tip regarding writing for an international audience: contractions are a bit of a difficulty for non-native English speakers. Prefer "you will" over "you'll".

MikeSpreitzer: Somewhere (here or in the next step) it should be explained _how_ the commend printed here will be used in the next step. If I understand correctly, it is this: the token and apiserver URL used in the later command are taken from the command printed here.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,272 @@
...
+
+### Step 2: Join the WEC to the ITS
+
+Execute the join command on the WEC to initiate the registration process:
```
MikeSpreitzer: This section still starts by telling the reader unconditionally to issue a particular command; unfortunately, that is not necessarily the right command. The reader is being told to do something without first being given a clue that she may also need the `--singleton` or `--force-internal-endpoint-lookup` flags.

docs/content/direct/wec-registration.md
MikeSpreitzer: FYI, "other local development setups" is vague.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,272 @@
...
+# Environment labels
+kubectl --context <its-context> label managedcluster <wec-name> environment=production tier=critical
+
+# Custom organizational labels
+kubectl --context <its-context> label managedcluster <wec-name> team=platform business-unit=engineering
+```
+
+### Customization Properties
+
+You can define additional properties for each WEC using ConfigMaps in the `customization-properties` namespace:
```
MikeSpreitzer: It is not an all-or-nothing option. s/each/a/

MikeSpreitzer (commented):

rishi-jat:
@MikeSpreitzer All suggested changes are done have a look.

docs/content/direct/wec-registration.md
MikeSpreitzer: This is still not rendering as you intend. Compare the website preview

image

with the direct view at GitHub:

image

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,279 @@
...
+clusteradm --context <its-context> get token
+```
+
+This command outputs a `clusteradm join` command that you will use in the next step. The token and apiserver URL from this command will be used in the join command.
+
+### Step 2: Join the WEC to the ITS
+
+**Important:** Before executing the join command, determine if you need additional flags based on your environment. The basic command below may not work for all cluster types.
+
+Execute the join command on the WEC to initiate the registration process. The exact command depends on your environment:
```
MikeSpreitzer, line 62: s/:/./

The colon implies that what follows is an elaboration on the remark before the colon, but that is not what actually follows.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,279 @@
...
 
-This document will tell users what how to register a WEC in an ITS.
+This document explains how to register a Workload Execution Cluster (WEC) with an Inventory and Transport Space (ITS) in KubeStellar.
 
-TODO: write this.
+## Overview
+
+Registering a WEC with an ITS follows the same process as registering a managed cluster with an OCM hub. KubeStellar uses Open Cluster Management (OCM) for cluster registration and management.
+
+The instructions below provide a comprehensive registration process focused on KubeStellar-specific considerations. For additional OCM registration details, refer to the [official Open Cluster Management documentation](https://open-cluster-management.io/docs/getting-started/installation/register-a-cluster/).
```
MikeSpreitzer, line 9: The word "comprehensive" here is not helping this sentence. I would simply remove it.

docs/content/direct/wec-registration.md
```diff
@@ -1,5 +1,279 @@
...
+clusteradm join --hub-token <token> --hub-apiserver <api-server-url> --cluster-name <your-wec-name> --context <wec-context>
+```
+
+**Important flags:**
+- `--cluster-name`: Choose a unique name for your WEC
+- `--context`: Specify the kubectl context for your WEC
+- `--singleton`: Use this flag if the WEC is a single-node cluster
+- `--force-internal-endpoint-lookup`: Required for Kind clusters and other clusters with internal-only API server endpoints
+
+**Example for Kind clusters:**
```
MikeSpreitzer, line 75: Shouldn't this be a real section heading too?

MikeSpreitzer (approved):
This is close enough to merge, with further improvements done in follow-on PR(s).

/lgtm
/approve
Thanks!

PR #3054, opened 2025-07-08
📖  Move CONTRIBUTING content into contributing-inc

MikeSpreitzer:
## Summary
This PR moves the content of the root file `CONTRIBUTING.md` into the website source file `docs/content/contribution-guidelines/contributing-inc.md`. The latter file used to mainly consist of include statements bringing in contents of the former, with ugly hacks to work around problems with relative file references. This PR simplifies things. Because the main content is in the regular website source directories, no ugly hacks are needed for relative references to other website source files.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-move-contributing/contribution-guidelines/contributing-inc/

## Related issue(s)

Fixes #


clubanderson:
ok, @MikeSpreitzer this is concerning to me.

not having a CONTRIBUTING.MD in the root - we will lose compliance with CNCF. 

You will have to leave that there and just make a dupe of it

I said this a few weeks ago when this topic came up. Yes, you should disconnect, but No you can't remove the file from root.

CONTRIBUTING.md
```diff
@@ -1,169 +1 @@
...
-<!--start-sixth-include-->
-
-
-## Certificate of Origin
-
-By contributing to this project you agree to the Developer Certificate of
-Origin (DCO). This document was created by the Linux Kernel community and is a
-simple statement that you, as a contributor, have the legal right to make the
-contribution. See the [DCO]({{ config.repo_url }}/blob/{{ config.ks_branch }}/DCO)</a> file for details.
+See [here](docs/content/contribution-guidelines/contributing-inc.md)
```
clubanderson, line 1: need to keep this file and its contents in place.

francostellari:
Instead of a copy, can't it be a file link?

MikeSpreitzer:
@clubanderson: There is still a CONTRIBUTING.md file in the root directory. I only moved the content, leaving a pointer. So if you read and think about what /CONTRIBUTING.md says, it is right.

clubanderson:
> @clubanderson: There is still a CONTRIBUTING.md file in the root directory. I only moved the content, leaving a pointer. So if you read and think about what /CONTRIBUTING.md says, it is right.

no good. You need the content in the file. This will be confusing to people that do not use the doc site to get started and if the website goes down, we have nothing. You need the full content in the root files. This is CNCF standard. I don't know of other projects with this type of redirect.

and now you know why we went with inclusions in the first place.

MikeSpreitzer:
> if the website goes down, we have nothing

No! The CONTRIBUTING.md file has a link to the full content that works.

image

Clicking on that links gets

image

That goes on, with the full content. No website involved.

clubanderson:
I dont agree with this @MikeSpreitzer - you need files in root that are specific to GitHub and not our doc site. Deleting all the content and replacing it with a link is not the right way to do this.

MikeSpreitzer:
@francostellari: we used to use symlinks instead of inclusion. We decided that we prefer source inclusion. I do not remember all the reasons why. One reason is that the inclusion technique that we developed copes with the relative reference problem (in an ugly way, but at least it works); using symlinks makes it impossible for the symlinked file to have relative references that work from both locations.

clubanderson (approved):

Issue #3075, opened 2025-07-11
Replace Bash Loops with kubectl wait in Scripts


Rupam-It:
Multiple scripts in the KubeStellar project use inefficient bash loops with sleep statements to wait for Kubernetes resources.

**Affected Scripts:**
scripts/create-kind-cluster-with-SSL-passthrough.sh
scripts/create-kubestellar-demo-env.sh
scripts/kflex-create-kubeconfig.sh
scripts/kubestellar-snapshot.sh
scripts/create-ocm-clusters.sh
scripts/create-k3s-cluster-with-SSL-passthrough.sh

**Proposed Solution**
k wait --for=create <obj>
k wat for condition ready on <obj>

Rupam-It:
@francostellari 
cc @MikeSpreitzer

francostellari:
good idea

greninja517:
@Rupam-It 
Are you still working on it ? Or, is this issue open to work ?

Rupam-It:
> [@Rupam-It](https://github.com/Rupam-It) Are you still working on it ? Or, is this issue open to work ?

as you can see above I have already change multiple file and quite check the scripts folder !
you can double check that if you find any other bash script simplification then go ahead ( I think there will be complex script present ) !
also check hack and other folders!

greninja517:
> > [@Rupam-It](https://github.com/Rupam-It) Are you still working on it ? Or, is this issue open to work ?
> 
> as you can see above I have already change multiple file and quite check the scripts folder ! you can double check that if you find any other bash script simplification then go ahead ( I think there will be complex script present ) ! also check hack and other folders!

I thought this issue is inactive. I didn't know that you were working on it.

clubanderson:
Closing as implemented. The scripts now use `kubectl wait` instead of bash loops. For example, `create-kubestellar-demo-env.sh` uses kubectl wait for controlplane readiness, job completion, and node readiness throughout the script.

PR #3081, opened 2025-07-13
🌱in kubestellar-demo.sh file while loop replace with kubectl wait...

Rupam-It:
## Related issue(s) #3075 

Fixes #

### DIFF
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -105,8 +105,11 @@ context_clean_up() {
 
 checking_cluster() {
     found=false
+
     while true; do
+
         output=$(kubectl --context its1 get csr)
+
         while IFS= read -r line; do
 
             if echo "$line" | grep -q $1; then
@@ -115,14 +118,17 @@ checking_cluster() {
                 found=true
                 break
             fi
+
         done <<< "$output"
 
         if [ "$found" = true ]; then
             break
+
         else
             echo "CSR for $1 not found. Trying again..."
             sleep 20
         fi
+
     done
 }
 
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -104,32 +104,37 @@ context_clean_up() {
 }
 
 checking_cluster() {
-    found=false
-
-    while true; do
-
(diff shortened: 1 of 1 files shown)


scripts/create-kubestellar-demo-env.sh
```diff
@@ -104,32 +104,37 @@ context_clean_up() {
...
+    echo "Waiting for CSR for $cluster_name..."
+    
+    # Wait for CSR to be created with cluster label
+    # OCM creates CSRs with specific labels for cluster identification
+    echo "Waiting for CSR creation for cluster $cluster_name..."
+    kubectl --context its1 wait --for=create csr \
+      -l cluster.open-cluster-management.io/clustername=$cluster_name \
+      --timeout=300s || {
+        # Fallback: wait for any CSR containing the cluster name
+        echo "Specific label not found, waiting for CSR with cluster name pattern..."
```
francostellari: @Rupam-It can you clarify when/why you need a fallback on the label?
I'm not very familiar with this script, so I'm trying to understand.

Rupam-It: OCM creates CSR first → Then applies label later
Gap between creation and labeling (few seconds)
kubectl wait with label fails if run too early

scripts/create-kubestellar-demo-env.sh
```diff
@@ -169,6 +169,10 @@ for cluster in "${clusters[@]}"; do
     }; then
         echo -e "\033[33m✔\033[0m Cluster $cluster was successfully created"
         kubectl config rename-context "${k8s_platform}-${cluster}" "${cluster}" >/dev/null 2>&1
+        
+        # Wait for cluster to be ready before proceeding
+        echo "Waiting for cluster $cluster nodes to be ready..."
+        kubectl --context "${cluster}" wait --for=condition=ready nodes --all --timeout=300s
```
MikeSpreitzer, line 175: Why is this wait here? The following command does not depend on the WEC cluster nodes being "ready".

MikeSpreitzer:
This one is waiting on response by @Rupam-It

Rupam-It:
> This one is waiting on response by @Rupam-It

you are right this explicit wait does not need here ! we can close this PR!

PR #3082, opened 2025-07-13
🌱using kubectl wait for cp to be ready in import-cp-context.sh file

Rupam-It:
## Related issue(s) #3075 

Fixes #

### DIFF
--- a/scripts/import-cp-contexts.sh
+++ b/scripts/import-cp-contexts.sh
@@ -51,11 +51,12 @@ get_kubeconfig() {
 
     echov "Getting the kubeconfig of KubeFlex Control Plane \"$cp_name\" of type \"$cp_type\" from context \"$context\":"
 
-    # wait for CP ready
-    while [[ $(KUBECONFIG="$in_KUBECONFIG" kubectl --context $context get controlplane $cp_name -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do
-        echov "* Waiting for KubeFlex Control Plane \"$cp_name\" in context \"$context\" to be Ready..."
-        sleep 5
-    done
+    # Wait for Control Plane to be ready using kubectl wait
+    echov "* Waiting for KubeFlex Control Plane \"$cp_name\" in context \"$context\" to be Ready..."
+    KUBECONFIG="$in_KUBECONFIG" kubectl --context "$context" wait \
+        --for=condition=Ready \
+        controlplane "$cp_name" \
+        --timeout=300s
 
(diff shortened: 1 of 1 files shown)


scripts/import-cp-contexts.sh
```diff
@@ -51,11 +51,12 @@ get_kubeconfig() {
 
     echov "Getting the kubeconfig of KubeFlex Control Plane \"$cp_name\" of type \"$cp_type\" from context \"$context\":"
 
-    # wait for CP ready
-    while [[ $(KUBECONFIG="$in_KUBECONFIG" kubectl --context $context get controlplane $cp_name -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do
-        echov "* Waiting for KubeFlex Control Plane \"$cp_name\" in context \"$context\" to be Ready..."
-        sleep 5
-    done
+    # Wait for Control Plane to be ready using kubectl wait
```
francostellari, line 54: Don't you need to wait `--for=create` first beforehand?

Rupam-It: Actually get_kubeconfig() function is called at [here](https://github.com/Rupam-It/kubestellar/blob/3075_imp-cp-ctx/scripts/import-cp-contexts.sh#L210)!

and cp already created , we only make sure that it is ready!

scripts/import-cp-contexts.sh
```diff
@@ -51,11 +51,12 @@ get_kubeconfig() {
...
+    echov "* Waiting for KubeFlex Control Plane \"$cp_name\" in context \"$context\" to be Ready..."
+    KUBECONFIG="$in_KUBECONFIG" kubectl --context "$context" wait \
+        --for=condition=Ready \
+        controlplane "$cp_name" \
+        --timeout=300s
```
MikeSpreitzer, line 59: We learned a while ago to use very generous timeouts. When someone is installing KubeStellar on their laptop at a crowded conference or a poorly-served area, it can take a _long_ time for images to get pulled from their registries.

MikeSpreitzer (approved):
I will merge this as progress, but the timeout should be increased in a follow-on PR.

/lgtm
/approve
Thanks!

scripts/import-cp-contexts.sh
francostellari: @MikeSpreitzer what value do you recommend?
@Rupam-It can you make a small PR to fix the time out in this script or other similar scripts that you are working on?

Rupam-It: We can make comment to explicitly tell user to increase the time!

PR #3084, opened 2025-07-13
🌱refine logic to make sure k3s service start and nginx ingress controler pod to be created 

Rupam-It:
## Related issue(s) #3075 

Fixes #


scripts/create-k3s-cluster-with-SSL-passthrough.sh
```diff
@@ -81,6 +81,29 @@ fi
 mkdir -p ~/.kube
 export KUBECONFIG=~/.kube/config
 sudo kubectl --kubeconfig=/etc/rancher/k3s/k3s.yaml config view --raw > "$KUBECONFIG"
+
+# Wait for k3s service to be ready and kubectl to work
+echo "Waiting for k3s service to be ready..."
+timeout=300
+elapsed=0
+while [ $elapsed -lt $timeout ]; do
+    if sudo systemctl is-active --quiet k3s && kubectl get nodes >/dev/null 2>&1; then
```
francostellari: I'm concerned that some system may not have `systemctl`

Rupam-It: Actually the problem was , in the old script k3s instantly download and try to access the command before starting the service itself that's why I'm add this script!

as you raise valid concern then we have to see alternative!

Rupam-It: hii @francostellari 
```
if [ -S /run/k3s/containerd/containerd.sock ] && kubectl get nodes >/dev/null 2>&1; then
        echo " K3s socket exists and kubectl is working! (${elapsed}s)"
```
   can we use this logic here ?

scripts/create-k3s-cluster-with-SSL-passthrough.sh
```diff
@@ -81,6 +81,29 @@ fi
 mkdir -p ~/.kube
 export KUBECONFIG=~/.kube/config
 sudo kubectl --kubeconfig=/etc/rancher/k3s/k3s.yaml config view --raw > "$KUBECONFIG"
+
+# Wait for k3s service to be ready and kubectl to work
+echo "Waiting for k3s service to be ready via socket file..."
+timeout=300
+elapsed=0
+while [ $elapsed -lt $timeout ]; do
+    if [ -S /run/k3s/containerd/containerd.sock ] && kubectl get nodes >/dev/null 2>&1; then
```
MikeSpreitzer: Is `[ -S /run/k3s/containerd/containerd.sock ] &&` necessary here? Can `kubectl get nodes` succeed when that socket does not exist?

scripts/create-k3s-cluster-with-SSL-passthrough.sh
```diff
@@ -81,6 +81,29 @@ fi
...
+    elapsed=$((elapsed + 5))
+done
+
+if [ $elapsed -ge $timeout ]; then
+    echo "❌ Timed out waiting for K3s to be ready."
+    exit 1
+fi
+
+echo "Waiting for k3s cluster nodes to be ready..."
+kubectl wait --for=condition=ready nodes --all --timeout=300s
```
MikeSpreitzer: Why is this wait here? The following commands do not depend on the cluster's Node being "ready".

scripts/create-k3s-cluster-with-SSL-passthrough.sh
```diff
@@ -81,6 +81,29 @@ fi
 mkdir -p ~/.kube
 export KUBECONFIG=~/.kube/config
 sudo kubectl --kubeconfig=/etc/rancher/k3s/k3s.yaml config view --raw > "$KUBECONFIG"
+
+# Wait for k3s service to be ready and kubectl to work
+echo "Waiting for k3s service to be ready via socket file..."
+timeout=300
+elapsed=0
+while [ $elapsed -lt $timeout ]; do
```
MikeSpreitzer, line 88: What goes wrong if this wait loop is not here? Has this failure been observed?

Rupam-It: yes, for me k3s was not install,  this script try to install it and try to run command before the k3s actually starting ! that why added a check to wait for start !

scripts/create-k3s-cluster-with-SSL-passthrough.sh
```diff
@@ -81,6 +81,26 @@ fi
 mkdir -p ~/.kube
 export KUBECONFIG=~/.kube/config
 sudo kubectl --kubeconfig=/etc/rancher/k3s/k3s.yaml config view --raw > "$KUBECONFIG"
+
+# Wait for k3s service to be ready and kubectl to work
+echo "Waiting for k3s service to be ready via socket file..."
```
MikeSpreitzer: FYI, This remark is not very on-point.

Rupam-It:
> The new `echo` could be improved. /lgtm

`Waiting for k3s service to be ready ` will this echo command will be good fit?

MikeSpreitzer:
@Rupam-It : yes, the comment right before the `echo` has good text.

Rupam-It:
> @Rupam-It : yes, the comment right before the `echo` has good text.

Thanks @MikeSpreitzer !
I think it is okk now!

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

Issue #2626, opened 2024-11-21
feature: start tackling security section of https://www.bestpractices.dev/en/projects/8266#all


clubanderson:
### Feature Description

OpenSSF scans our repo for best practices related to security of our source. We currently have 45% coverage for all OpenSSF criteria. It would be good to get this number north of 80% to show that we are taking security seriously.

### Proposed Solution

read the content on https://www.bestpractices.dev/en/projects/8266#all and come up with a plan to implement parts of it to satisfy criteria

### Want to contribute?

- [ ] I would like to work on this issue.

pratik-mahalle:
### Improve OpenSSF Criteria Compliance for KubeStellar Repository  

**Description:**  
This PR addresses the need to enhance our compliance with OpenSSF Best Practices Badge criteria. Currently, we have 45% coverage, and this effort is aimed at increasing it to over 80%. By implementing the necessary changes, we demonstrate our commitment to maintaining a high standard of security and best practices for our source code.  

**Changes Introduced:**  
1. **Documentation Improvements:**  
   - Added detailed documentation for contributing guidelines, including requirements for acceptable contributions.  
   - Updated release notes to include human-readable summaries of major changes and address publicly known vulnerabilities.  

2. **Bug Reporting and Issue Tracking:**  
   - Enhanced the visibility of the bug-reporting process and provided a public archive for reports and responses.  
   - Improved responsiveness to bug reports and enhancement requests.  

3. **Security Enhancements:**  
   - Published a clear process for reporting vulnerabilities, including guidelines for private reporting.  
   - Integrated a policy to address vulnerabilities within 14 days.  
   - Ensured cryptographic practices meet NIST requirements and avoid weak algorithms.  

4. **Build and Testing Improvements:**  
   - Implemented static analysis tools for detecting vulnerabilities in the source code.  
   - Strengthened the automated test suite to cover more code branches and input fields.  
   - Updated CI pipelines to align with FLOSS testing practices.  

5. **Version Control and Release Management:**  
   - Added semantic versioning for releases and tagged all releases in the repository.  
   - Enhanced release notes to include critical updates and upgrade impacts.  

**Remaining Tasks:**  
- [ ] Integrate additional static analysis tools and document their usage.  
- [ ] Conduct a review of existing vulnerabilities and ensure all are patched.  
- [ ] Finalize policies for adding tests for new features and document them.  

**How to Test:**  
1. Run the CI pipelines to validate the build and test suite updates.  
[...]

PR #3085, opened 2025-07-14
✨ SAST scan implementation

antedotee:
This workflow automates security checks for every push, PR, and on a weekly schedule. It runs four main jobs:

### 1. **CodeQL SAST Analysis**
- Runs GitHub’s CodeQL to scan Go code for security vulnerabilities.
- Uploads results to GitHub Security tab.

### 2. **OSV Dependency Scan**
- Generates a Software Bill of Materials (SBOM) from `go.mod`.
- Uses Google’s OSV-Scanner to check for known vulnerabilities in dependencies.
- Uploads results as SARIF (for GitHub’s code scanning UI).
- Fails the job if critical/high vulnerabilities are found.

### 3. **Trivy Filesystem Scan**
- Uses Trivy to scan the entire repo for vulnerabilities in files and binaries.
- Uploads results as both artifact and SARIF.
- Fails the job if critical/high vulnerabilities are found.

### 4. **Trivy Config & Helm Chart Scan**
- Scans Dockerfiles, docker-compose, Kubernetes YAMLs, and Helm charts for misconfigurations and vulnerabilities.
- Only runs if CodeQL and OSV jobs pass.
- Uploads results as artifacts.
- Fails the job if critical/high issues are found.

### Other Details
- All jobs are non-blocking for PRs unless critical/high issues are found.
- Results are uploaded as artifacts for auditability.
- The workflow is robust: it checks if scan results exist before uploading, and validates SARIF files.

## Effect of These Changes

- **Automates security scanning** for code and dependencies on every push/PR.
- **Catches vulnerabilities early** (in code, dependencies, Docker/K8s/Helm configs).
- **Prevents merging PRs with critical/high vulnerabilities** (unless explicitly overridden).
- **Makes results visible in GitHub’s Security tab** and as downloadable artifacts.
- **Reduces false positives and workflow failures** by validating scan outputs before upload.

Relates to #2626

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -81,10 +81,25 @@ jobs:
       - name: Install OSV Scanner
         run: |
           go install github.com/google/osv-scanner/cmd/osv-scanner@latest
+      - name: Get current date for cache key
+        id: date
+        run: echo "date=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT
       - name: Cache OSV database
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cache/osv-scanner
+            ~/.osv-scanner
+          key: osv-db-${{ steps.date.outputs.date }}
+          restore-keys: |
+            osv-db-
+      - name: Download OSV database
         run: |
-          mkdir -p ~/.cache/osv
-          echo "OSV scanner will update database automatically during scan"
+          echo "Downloading OSV vulnerability database..."
+          osv-scanner --update-databases || {
+            echo "Database download failed, continuing with existing cache"
+            echo "OSV scanner will attempt to update during scan if needed"
+          }
         continue-on-error: true
       - name: Run OSV Scanner
         run: |


antedotee:
Currently the OSV vulnerability scan is failing due to some dependency issue in go.mod. I am trying to fix it in the meantime. Also, instead of running all the jobs parallely, I have made some jobs to be only run if the previous ones are executed successfully to save time. I would be happy to implement and discuss the implications and improvements suggested.
Thanks and regards.

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,341 @@
...
+                echo "Generated modules.json as fallback"
+              fi
+            }
+          else
+            echo "No go.mod found, skipping SBOM generation"
+          fi
+      - name: Install OSV Scanner
+        run: |
+          go install github.com/google/osv-scanner/cmd/osv-scanner@latest
+      - name: Cache OSV database
```
MikeSpreitzer, line 87: I see no database fetching here.

MikeSpreitzer:
@antedotee: rebasing onto the latest `main` will fix the yq version check failure in the "Verify Action Hashes" workflow.

.github/workflows/security-scan.yml
MikeSpreitzer, line 1: Since we have Prow, and our capacity to run GitHub Actions workflows is limited, why not make this a Prow job?

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,356 @@
...
+    branches:
+      - "main"
+      - "release-*"
+  pull_request:
+    branches:
+      - "main"
+      - "release-*"
+  schedule:
+    - cron: '0 12 * * 1'
+
```
MikeSpreitzer, line 15: Wouldn't it be good to also make possible to explicitly invoke this workflow from the GitHub web UI? Could be done in a follow-on PR.

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,356 @@
...
+        run: echo "date=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT
+      - name: Cache OSV database
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cache/osv-scanner
+            ~/.osv-scanner
+          key: osv-db-${{ steps.date.outputs.date }}
+          restore-keys: |
+            osv-db-
```
MikeSpreitzer, line 95: Why is this here? In what scenario does it accomplish something? Does it accomplish something that we really want?

`${{ steps.date.outputs.date }}` will never be the empty string, right?

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,356 @@
...
+
+  trivy:
+    name: "Trivy Vulnerability Scan"
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@76071ef0d7ec797419534a183b498b4d6366cf37
```
MikeSpreitzer, line 185: There is a later version released a day after that one.

MikeSpreitzer:
The new workflow is currently failing.

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,356 @@
...
+          go install github.com/google/osv-scanner/cmd/osv-scanner@latest
+      - name: Get current date for cache key
+        id: date
+        run: echo "date=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT
+      - name: Cache OSV database
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cache/osv-scanner
+            ~/.osv-scanner
```
MikeSpreitzer, line 92: This cache is empty at the start of every run of this job (https://docs.github.com/en/actions/how-tos/write-workflows/choose-where-workflows-run/choose-the-runner-for-a-job#choosing-github-hosted-runners), right? If so then what is the point of using a caching mechanism here?

clubanderson:

PR #3089, opened 2025-07-16
🌱  Bump anchore/sbom-action/download-syft to 0.20.2

MikeSpreitzer:
## Summary
This is the principled version of #3042 

## Related issue(s)

Fixes #

PR #3090, opened 2025-07-16
🌱  List more files that should not trigger e2e/integration tests

MikeSpreitzer:
## Summary
This PR expands the list of files that should not trigger the E2E and integration test workflows. This is good because it wastes less of our human time and github workflow run capacity.

## Related issue(s)

Fixes #


MikeSpreitzer:
Ugh, this will not pass the broken yq version check that is fixed in #3089 . So I am adding this change into that PR.

PR #3092, opened 2025-07-16
📖 improve WDS page based on feedback from #3023

### LINKED ISSUE #3000: doc: issues with website page about WDS

Tob-iee:
## Summary
Before making this PR, I liaised with @rishi-jat. We can guarantee that this PR solely enhances the WDS documentation while simultaneously preventing any overlaps or duplication of efforts in other PRs, as mentioned in #3023 and #2993. 

## Related

Fixes #3000 and addresses feedback from PR #3023 by @MikeSpreitzer

## Live Preview

Preview → https://tob-iee.github.io/kubestellar/doc-fix-wds/direct/wds/

### DIFF
--- a/docs/content/direct/wds.md
+++ b/docs/content/direct/wds.md
@@ -2,68 +2,85 @@
 
 - [What is a WDS?](#what-is-a-wds)
 - [Creating a WDS](#creating-a-wds)
+
   - [Using the KubeFlex CLI](#using-the-kubeflex-cli)
-- [KubeFlex Hosting Cluster as WDS](#kubeflex-hosting-cluster-as-wds)
-- [WDS vs. ControlPlane Registration](#wds-vs-controlplane-registration)
-- [Controllers Running in a WDS](#controllers-running-in-a-wds)
-- [Working with a WDS](#working-with-a-wds)
   - [Accessing the WDS](#accessing-the-wds)
 
-A Workload Description Space (WDS) is a core component of the KubeStellar architecture that serves as the primary interface for users to define and manage workloads for multi-cluster deployment.
+- [Working with a WDS](#working-with-a-wds)
+- [WDS vs. ControlPlane Registration](#wds-vs-controlplane-registration)
+- [Controllers Running in a WDS](#controllers-running-in-a-wds)
+
+
 
 ## What is a WDS?
 
-A WDS is a space (a Kubernetes-like API server with storage) that:
+A Workload Description Space (WDS) is a space in the [KubeStellar architecture](architecture.md) that serves as the primary interface for users to define and manage workloads for multi-cluster deployment. The WDS consists of a Kubernetes API server with storage that:
 
 - Stores the definitions of workloads in their native Kubernetes format
-- Hosts the control objects (`BindingPolicy` and `Binding`) that define how workloads are distributed
+- Stores the control objects (`BindingPolicy`, `Binding`, `Status Collector`, `CombinedStatus` and `CustomTransform`) that define how workloads are distributed
 - Maintains status information about deployed workloads
 - Acts as the main user interface to the KubeStellar system
 
+
 ## Creating a WDS
 
-A WDS can be created in several ways:
+Currently, we support the use of our Kubestellar core Helm chart as the only way to  create a WDS. This is because the core Helm chart also automatically creates a kubestellar-controller-manager and transport-controller, which are contorllers that your WDS requires to function properly.
 
-### Using the KubeStellar Core Helm Chart
+### Use the KubeStellar Core Helm Chart to create your WDS
 
 The recommended approach is to use the KubeStellar Core Chart:
 
 ```shell
 helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart \
-  --set-json='WDSes=[{"name":"wds1", "type":"k8s"}]'
+  --set-json='WDSes=[{"name":"<wds-name>", "type":"<space-type>"}]'
 ```
 
-You can customize your WDS by specifying:
-- `name`: A unique name for the WDS
-- `type`: 
-  - `k8s` (default): Creates a basic Kubernetes API Server with a subset of kube controllers
-  - `host`: Uses the KubeFlex hosting cluster itself
-- `APIGroups`: A comma-separated list of API Groups to include
-- `ITSName`: The name of the ITS to be used by this WDS (required if multiple ITSes exist)
+Specify a `name` for your WDS, and you can customize it further with optional parameters:  
 
-### Using the KubeFlex CLI
+- `name` (required): A unique name for the WDS (`<wds-name>`)
+- `type` (optional, default: `k8s`):  (`<space-type>`)
 
-You can also create a WDS using the KubeFlex CLI:
+    - `k8s`: Creates a basic Kubernetes API Server with a subset of kube controllers
+    - `host`: Uses the KubeFlex hosting cluster itself. 
+    
+- `APIGroups` (optional): A comma-separated list of API Groups to include
+- `ITSName` (optional): The name of the ITS to be used by this WDS (required if multiple ITSes exist)
 
-```shell
-kflex create wds1 -p kubestellar
-```
+The type of space you choose determines the type of controllers that are included in your WDS. When creating a WDS, the recommended space types are "K8s" (the default space) and "host". If you the decide to use your hosting cluster's control plane (which is typically a KubeFlex cluster for Kubstellar) by specifying `type:host`, it presents you some benefits such as:
 
-This command creates a WDS and runs a post-create hook that deploys the KubeStellar controller manager and transport controller.
-
-## KubeFlex Hosting Cluster as WDS
-
-The KubeFlex hosting cluster can be configured to act as a WDS by specifying `type: host` when creating the WDS:
-
-```shell
-helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart \
-  --set-json='WDSes=[{"name":"wds1", "type":"host"}]'
-```
-
-This approach:
-- Avoids creating a separate control plane
+- Saves the cost of creating an additional space.
 - Simplifies the architecture by reusing the hosting cluster
 - Makes the WDS directly accessible through the hosting cluster's API server
+
+> You can create multiple WDSes of the same space types and even a mix of different space types, where you can have the default and host WDSes in your cluster. A multiple WDS architecure can be valueable in cases where you want to manage workload runs across different sets of users or groups. Thus each user or group can use the same ITS for WEC inventory but with their different WDS
+
+### Creating a kubeconfig context for accessing the WDS
+
+After creating your WDS, you will need to access it to submit workload objects. To do this you have to create a context for your WDS in kubeconfig context and set it context as your current cluster context:
+
+- **Setup WDS context with KubeFlex CLI**: KubeFlex CLI enables you to create the context for your WDS in the kubeconfig context. Then automatically switch to it by overwiting the current kubeconfig context as your WDS context.
+
+```shell
+# create a the WDS context and make it the current context
+kflex ctx --overwrite-existing-context <wds-name>
+```
+
+- **Access WDS context with Kubernetes API**: In a case where the context for your WDS already exists in the kubeconfig context, then, you can just switch to it as your current context with the Kubernetes API. 
+
+```shell
+# Switch to the WDS context
+kubectl config use-context <wds-name>
+```
+
+Both methods provide flexibility depending on your preferred way of working with the system.
+
+## Working with a WDS
+
+Once your WDS is created, you can:
+
+1. **Create workload objects** in their native Kubernetes format
+2. **Define BindingPolicy objects** to specify which workloads should be deployed to which WECs
+3. **Monitor StatusCollector and CombinedStatus objects** for a status update on your deployed workloads
 
 ## WDS vs. ControlPlane Registration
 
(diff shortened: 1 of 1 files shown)


docs/content/direct/wds.md
```diff
@@ -2,68 +2,85 @@
...
-- [Working with a WDS](#working-with-a-wds)
-  - [Accessing the WDS](#accessing-the-wds)
 
-A Workload Description Space (WDS) is a core component of the KubeStellar architecture that serves as the primary interface for users to define and manage workloads for multi-cluster deployment.
+
 
 ## What is a WDS?
 
-A WDS is a space (a Kubernetes-like API server with storage) that:
+A Workload Description Space (WDS) is a space in the [KubeStellar architecture](architecture.md) that serves as the primary interface for users to define and manage workloads for multi-cluster deployment. The WDS constitue of a Kubernetes API server with storage that:
```
MikeSpreitzer: If I understand your intent correctly, you want to use the word "consists" rather than "constitute".

docs/content/direct/wds.md
```diff
@@ -2,68 +2,85 @@
...
+- Simplifies the architecture by reusing the hosting cluster
+- Makes the WDS directly accessible through the hosting cluster's API server
 
-```shell
-kflex create wds1 -p kubestellar
-```
+> You can create multiple WDSes of the same space types and even a mix of different space types, where you can have the default and host WDSes in your cluster. A multiple WDS architecure can be valueable in cases where you want to manage workload runs across different sets of users or groups. Thus each user or group can use the same ITS for WEC inventory but with their different WDS
 
-This command creates a WDS and runs a post-create hook that deploys the KubeStellar controller manager and transport controller.
+### Accessing the WDS
```
MikeSpreitzer: A more accurate title for this section would be "Creating a kubeconfig context for accessing the WDS". Nothing in this section actually _does_ the access; this section is all about _preparing to access_. The actual access is the stuff done in the "Working with a WDS" section.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,85 @@
...
+
 
 ## What is a WDS?
 
-A WDS is a space (a Kubernetes-like API server with storage) that:
+A Workload Description Space (WDS) is a space in the [KubeStellar architecture](architecture.md) that serves as the primary interface for users to define and manage workloads for multi-cluster deployment. The WDS constitue of a Kubernetes API server with storage that:
 
 - Stores the definitions of workloads in their native Kubernetes format
-- Hosts the control objects (`BindingPolicy` and `Binding`) that define how workloads are distributed
+- Hosts the control objects (`BindingPolicy`, `Binding`, `Status Collector`, `CombinedStatus` and `CustomTransform`) that define how workloads are distributed
```
MikeSpreitzer: I think it would be better to use the word "stores" here instead of "hosts". That makes the similarity to the previous item clearer. Also, "hosts" suggests that the WDS in some sense "runs" or "implements" these objects --- but it does not actually do that.

I would also make the following item start with "stores".

docs/content/direct/wds.md
```diff
@@ -2,68 +2,85 @@
...
 ```
 
-You can customize your WDS by specifying:
-- `name`: A unique name for the WDS
-- `type`: 
-  - `k8s` (default): Creates a basic Kubernetes API Server with a subset of kube controllers
-  - `host`: Uses the KubeFlex hosting cluster itself
+You can customize your WDS by specifying:  
+
+- `<wds-name>`: A unique name for the WDS
```
MikeSpreitzer: The name was never optional.

Tob-iee: I’m sorry, I’m not quite clear on your comment here. Could you kindly explain better

MikeSpreitzer: The WDS name item is in a list introduced with the words "You can customize your WDS by specifying". The words "you can" suggests that the list items are optional. My objection is that the WDS name is not optional.

Tob-iee: Okay. I have improved the sentence clarity

docs/content/direct/wds.md
```diff
@@ -2,68 +2,85 @@
...
-- `name`: A unique name for the WDS
-- `type`: 
-  - `k8s` (default): Creates a basic Kubernetes API Server with a subset of kube controllers
-  - `host`: Uses the KubeFlex hosting cluster itself
-- `APIGroups`: A comma-separated list of API Groups to include
-- `ITSName`: The name of the ITS to be used by this WDS (required if multiple ITSes exist)
+Specify a `name` for your WDS, and you can customize it further with optional parameters:  
 
-### Using the KubeFlex CLI
+- `name` (required): A unique name for the WDS (`<wds-name>`)
```
MikeSpreitzer, line 40: You no longer need "name" in this list of parameters, you already described it just above.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,85 @@
...
-kflex create wds1 -p kubestellar
-```
+The type of space you choose determines the type of controllers that are included in your WDS. When creating a WDS, the recommended space types are "K8s" (the default space) and "host". If you the decide to use your hosting cluster's control plane (which is typically a KubeFlex cluster for Kubstellar) by specifying `type:host`, it presents you some benefits such as:
 
-This command creates a WDS and runs a post-create hook that deploys the KubeStellar controller manager and transport controller.
+- Saves the cost of creating an additional space.
+- Simplifies the architecture by reusing the hosting cluster
+- Makes the WDS directly accessible through the hosting cluster's API server
+
+> You can create multiple WDSes of the same space types and even a mix of different space types, where you can have the default and host WDSes in your cluster. A multiple WDS architecure can be valueable in cases where you want to manage workload runs across different sets of users or groups. Thus each user or group can use the same ITS for WEC inventory but with their different WDS
```
MikeSpreitzer, line 55: In Markdown, `> ` introduces a _block quote_. The text here is not being quoted from somewhere else (at least, it is not evidently so). Since it is not a quote, do not introduce it with `> `. If you want to bring extra attention to it, use an appropriate technique. For example, you might introduce it with `**Note:** `.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,85 @@
...
+
+### Creating a kubeconfig context for accessing the WDS
 
-## KubeFlex Hosting Cluster as WDS
+After creating your WDS, you will need to access it to submit workload objects. To do this you have to create a context for your WDS in kubeconfig context and set it context as your current cluster context:
```
MikeSpreitzer, line 59: The current text has a problem with a misplaced modifier. The text "in kubeconfig" appears right after "your WDS" --- so the normal English interpretation would be that "in kubeconfig" is modifying "your WDS". But that is not what you mean.

MikeSpreitzer, line 59: Why does the word "context" appear in "in kubeconfig context"? It seems confused to me.

MikeSpreitzer, line 59: s/set it context/set this context/
s/current cluster context/current kubeconfig context/

docs/content/direct/wds.md
```diff
@@ -2,68 +2,85 @@
...
 
-The KubeFlex hosting cluster can be configured to act as a WDS by specifying `type: host` when creating the WDS:
+- **Setup WDS context with KubeFlex CLI**: KubeFlex CLI enables you to create the context for your WDS in the kubeconfig context. Then automatically switch to it by overwiting the current kubeconfig context as your WDS context.
 
 ```shell
```
MikeSpreitzer, line 63: This is not indented properly. This problem was not introduced by this PR, does not have to be fixed by this PR.

image

docs/content/direct/wds.md
MikeSpreitzer: One possible way to fix this problem would be write something like "create a kubeconfig context for your WDS and set it..."

MikeSpreitzer, line 59: A colon is something that appears within a sentence. But that is not what is happening here: the stuff after the colon is not the rest of one sentence.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,85 @@
...
+
+> You can create multiple WDSes of the same space types and even a mix of different space types, where you can have the default and host WDSes in your cluster. A multiple WDS architecure can be valueable in cases where you want to manage workload runs across different sets of users or groups. Thus each user or group can use the same ITS for WEC inventory but with their different WDS
+
+### Creating a kubeconfig context for accessing the WDS
 
-## KubeFlex Hosting Cluster as WDS
+After creating your WDS, you will need to access it to submit workload objects. To do this you have to create a context for your WDS in kubeconfig context and set it context as your current cluster context:
 
-The KubeFlex hosting cluster can be configured to act as a WDS by specifying `type: host` when creating the WDS:
+- **Setup WDS context with KubeFlex CLI**: KubeFlex CLI enables you to create the context for your WDS in the kubeconfig context. Then automatically switch to it by overwiting the current kubeconfig context as your WDS context.
```
MikeSpreitzer, line 61: These two sentences here are confusing. The first one is a declaration in passive voice. The second is an imperative. With the second sentence being an imperative, the reader will think that they are being told to do two things: first one implied by the passive voice statement and then follow the imperative. But this is not what you mean. What needs to be said here is both an explanation of the relevant KubeFlex CLI functionality and an explanation of when/why to use it.

For example, you might write something like the following.

> The following command will (1) ensure that your kubeconfig file has a context that (a) has the same name as the WDS and (b) has the right contents for accessing the WDS and then (2) make that context your current one. You only need to to this once, after creating the WDS. **BEWARE:** This command must only be used either (a) while the current kubeconfig context is for accessing the KubeFlex hosting cluster or (b) after the KubeFlex CLI has added its [hosting cluster extension](https://github.com/kubestellar/kubeflex/blob/main/docs/users.md#hosting-context) into your kubeconfig file.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,85 @@
...
-  --set-json='WDSes=[{"name":"wds1", "type":"host"}]'
+# create a the WDS context and make it the current context
+kflex ctx --overwrite-existing-context <wds-name>
 ```
 
-This approach:
-- Avoids creating a separate control plane
-- Simplifies the architecture by reusing the hosting cluster
-- Makes the WDS directly accessible through the hosting cluster's API server
+- **Access WDS context with Kubernetes API**: In a case where the context for your WDS already exists in the kubeconfig context, then, you can just switch to it as your current context with the Kubernetes API.
```
MikeSpreitzer, line 68: There is no meaning to one context existing "in the kubeconfig context". I think that you mean to speak of one context existing in the (user's) kubeconfig file.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,85 @@
...
+
+Both methods provide flexibility depending on your preferred way of working with the system.
+
+## Working with a WDS
+
+Once your WDS is created, you can:
+
+1. **Create workload objects** in their native Kubernetes format
+2. **Define BindingPolicy objects** to specify which workloads should be deployed to which WECs
+3. **Monitor StatusCollector and CombinedStatus objects** for a status update on your deployed workloads
```
MikeSpreitzer, line 83: This is a subset of what can be done. I recommend not duplicating other documentation here. Instead, refer to it.

@KPRoche, website syntax erorr: The "Usage" node in the navigation tree does not have a suitable overview page.

@Tob-iee: for now, I recommend a reference like the following.

> the "User Guide > Usage" section of this website

Yes, I know that this will not be great when `wds.md` is viewed directly at GitHub. Rather than spend energy on that problem I would spend the energy on fixing the website syntax error.

docs/content/direct/wds.md
```diff
@@ -74,32 +91,14 @@ It's important to distinguish between:
 
 When using the Core Helm Chart or KubeFlex CLI with appropriate parameters, both steps happen automatically.
 
-## Controllers Running in a WDS
+## Controllers that work with WDS
 
-When a space is configured as a WDS, the following controllers are deployed:
+For the WDS to execute its tasks after it is configured, it will need to interact with the following controllers that live in the Hosting cluster:
```
MikeSpreitzer, line 96: The WDS does not execute tasks. The WDS is just a space. The controllers mentioned here access that space in the course of doing their jobs.

docs/content/direct/wds.md
```diff
@@ -2,68 +2,85 @@
...
-A WDS can be created in several ways:
+Currently, we support the use of our Kubestellar core Helm chart as the only way to  create a WDS. This is because the core Helm chart also automatically creates a kubestellar-controller-manager and transport-controller, which are contorllers that your WDS requires to function properly.
 
-### Using the KubeStellar Core Helm Chart
+### Use the KubeStellar Core Helm Chart to create your WDS
 
 The recommended approach is to use the KubeStellar Core Chart:
 
 ```shell
 helm upgrade --install ks-core oci://ghcr.io/kubestellar/kubestellar/core-chart \
```
MikeSpreitzer, line 34: This problem is not introduced in this PR and I will not insist that it be fixed in this PR, but: the writing here completely fails to give the reader the right clues. This is not referring the reader to the normal usage in the setup instructions. OK, maybe "normal" is too normative. But this text ONLY talks about using the core chart JUST for creating ONE WDS, which is really omitting critical context. It also shows an incorrect command for using the core chart to ONLY ADD ONE WDS.

PR #3093, opened 2025-07-17
📖 Fix typos in Packaging and Getting Started documentation

isagar678:
Preview: https://isagar678.github.io/kubestellar/fix-typos
## Summary
Packaging Page (packaging.md)
Original: "By our development practices and doing doing any manual hacks"
Corrected: "By our development practices and not doing any manual hacks"

Getting Started Page (get-started.md)
Original: Command block with ": set flags to..."
Corrected: # : set flags to... (formatted as a comment for clarity)


docs/content/direct/get-started.md
```diff
@@ -158,7 +158,7 @@ Note that `kind` does not support three or more concurrent clusters unless you r
 1. Execute the following commands to create two kind clusters, named `cluster1` and `cluster2`, and register them with the OCM hub. These clusters will serve as workload clusters. If you have previously executed these commands, you might already have contexts named `cluster1` and `cluster2`. If so, you can remove these contexts using the commands `kubectl config delete-context cluster1` and `kubectl config delete-context cluster2`.
 
     ```shell
-    : set flags to "" if you have installed KubeStellar on an OpenShift cluster
+    # set flags to "" if you have installed KubeStellar on an OpenShift cluster
```
MikeSpreitzer: What is wrong with using a colon here?

isagar678: The colon (:) at the start is a valid but confusing shell command. In shell scripting, : is a built-in command that does nothing and always succeeds (a "no-op"). It is sometimes used as a placeholder. However, using : here makes the line look like an executable shell command, but it actually does nothing and just displays the comment as part of the command history if run.
Using : in this way is not standard for comments. Readers may not recognize it as a comment, leading to confusion.

MikeSpreitzer: Yes, the colon was deliberately chosen --- to add a little emphasis to the comment (it appears an extra time).

MikeSpreitzer:
Being a doc PR, this needs a preview.

PR #3105, opened 2025-07-21
🌱migrate complex bash script to init container in argocd setup file

### LINKED ISSUE #3068: Migrate ArgoCD Setup Job from Bash Script to Init Containers
The current ArgoCD setup job in `core-chart/templates/argocd/setup.yml `uses a single container with a bash script that contains multiple retry loops to wait for dependencies. This approach has several drawbacks.

**Proposed Solution**
Migrate to a init container pattern that leverages Kubernetes' native dependency management
```
initContainers:
        # Wait for ArgoCD infrastructure
        - name: wait-argocd-pod-create
        - name: wait-argocd-pod-ready  
        - name: wait-argocd-secret-create
```

@francostellari 
cc @MikeSpreitzer

Rupam-It:
## Related issue(s)

Fixes #3068

### DIFF
--- a/core-chart/templates/argocd/setup.yaml
+++ b/core-chart/templates/argocd/setup.yaml
@@ -11,18 +11,24 @@ data:
     #!/usr/bin/env bash
     set -e
 
+    # Add the timestamped_echo function for consistent logging
+    timestamped_echo() {
+      echo "[$(date +"%H:%M:%S %Z")] $1"
+    }
+
     NS={{ .Release.Namespace }}
-    echo "[$(date +"%H:%M:%S %Z")] Getting ArgoCD pod and credentials..."
+    timestamped_echo "Getting ArgoCD pod and credentials..."
     ARGOCD_POD=$(kubectl -n $NS get pods -l app.kubernetes.io/name=argocd-server -o 'jsonpath={.items[0].metadata.name}')
     ARGOCD_PASSWD=$(kubectl -n $NS get secret argocd-initial-admin-secret -o jsonpath="{.data.password}")
 
-    echo "[$(date +"%H:%M:%S %Z")] Logging into Argo CD..."
+    timestamped_echo "Logging into Argo CD..."
     kubectl -n $NS exec $ARGOCD_POD -- argocd login "{{ .Release.Name }}-argocd-server.$NS" --username admin --password $(echo $ARGOCD_PASSWD | base64 -d) --insecure
 
-    echo "[$(date +"%H:%M:%S %Z")] Adding WDS clusters..."
+    timestamped_echo "Adding WDS clusters..."
     {{- range $cp := .Values.WDSes }}
       cp_name={{ $cp.name }}
-      echo "[$(date +"%H:%M:%S %Z")] Processing WDS cluster: $cp_name"
+      timestamped_echo "==> Looking for Control Plane \"$cp_name\"..."
+      timestamped_echo "==> Control Plane \"$cp_name\" is ready."
       
       # Get kubeconfig for the control plane
       key=$(kubectl get controlplane $cp_name -o=jsonpath='{.status.secretRef.inClusterKey}')
@@ -30,18 +36,29 @@ data:
       secret_namespace=$(kubectl get controlplane $cp_name -o=jsonpath='{.status.secretRef.namespace}')
       kubeconfig="/tmp/${cp_name}.kubeconfig"
       
+      timestamped_echo "==> Populating \"$kubeconfig\" from key \"$key\" in secret \"$secret_name\" in namespace \"$secret_namespace\" for control plane \"$cp_name\""
       kubectl get secret $secret_name -n $secret_namespace -o=jsonpath="{.data.$key}" | base64 -d > "$kubeconfig"
+      
+      timestamped_echo "==> Copying kubeconfig \"$kubeconfig\" to Argo CD pod \"$ARGOCD_POD\"..."
       kubectl -n $NS cp "$kubeconfig" ${ARGOCD_POD}:/tmp
       rm "$kubeconfig"
       
-      echo "[$(date +"%H:%M:%S %Z")] Registering Control Plane \"$cp_name\"..."
+      timestamped_echo "==> Registering Control Plane \"$cp_name\" as an Argo CD cluster..."
       kubectl -n $NS exec $ARGOCD_POD -- argocd cluster add ${cp_name} -y --kubeconfig "$kubeconfig"
       kubectl -n $NS exec $ARGOCD_POD -- rm "$kubeconfig"
     {{- end }}
 
-    echo "[$(date +"%H:%M:%S %Z")] Creating applications..."
+    timestamped_echo "Creating applications..."
     {{- range $app := .Values.argocd.applications }}
-      echo "[$(date +"%H:%M:%S %Z")] Creating application \"{{ $app.name }}\""
+      timestamped_echo "==> Creating application \"{{ $app.name }}\":"
+      timestamped_echo "    - project=\"{{ $app.project | default "default" }}\""
+      timestamped_echo "    - repoURL=\"{{ $app.repoURL }}\""
+      timestamped_echo "    - targetRevision=\"{{ $app.targetRevision | default "HEAD" }}\""
+      timestamped_echo "    - path=\"{{ $app.path | default "/" }}\""
+      timestamped_echo "    - destinationWDS=\"{{ $app.destinationWDS }}\""
+      timestamped_echo "    - destinationNamespace=\"{{ $app.destinationNamespace | default "default" }}\""
+      timestamped_echo "    - syncPolicy=\"{{ $app.syncPolicy | default "manual" }}\""
+
       kubectl -n $NS exec $ARGOCD_POD -- argocd app create "{{ $app.name }}" \
         --project "{{ $app.project | default "default" }}" \
         --repo "{{ $app.repoURL }}" \
@@ -53,7 +70,7 @@ data:
         --sync-policy "{{ $app.syncPolicy | default "manual" }}" || true
     {{- end }}
 
-    echo "[$(date +"%H:%M:%S %Z")] Done."
+    timestamped_echo "Done."
 ---
 apiVersion: batch/v1
 kind: Job


core-chart/templates/argocd/setup.yaml
```diff
@@ -4,117 +4,52 @@ kind: ConfigMap
 metadata:
   name: "{{ .Release.Name }}-argocd-kubestellar-setup"
 data:
-  # This script is used by the "{{ .Release.Name }}-argocd-kubestellar-setup" job to
-  # 1. register KubeStellar WDSes as Argo CD target clusters
-  # 2. create any Argo CD application listed in KubeStelar values
+  # Simplified script - most waiting logic moved to initContainers
```
francostellari: I think this new comment is not very helpful in the long term.
I suggest to keep a version of the old comment explaining what the script does.
You can add an explanation of what is done expected by the initContainers

core-chart/templates/argocd/setup.yaml
```diff
@@ -4,117 +4,52 @@ kind: ConfigMap
...
-  # 1. register KubeStellar WDSes as Argo CD target clusters
-  # 2. create any Argo CD application listed in KubeStelar values
+  # Simplified script - most waiting logic moved to initContainers
   argocd-setup.sh: |
     #!/usr/bin/env bash
     set -e
 
-    WAIT_BEFORE_RETRY=5
     NS={{ .Release.Namespace }}
+    ARGOCD_POD=$(kubectl -n $NS get pods -l app.kubernetes.io/name=argocd-server -o 'jsonpath={.items[0].metadata.name}')
```
francostellari, line 21: I think it is a good idea to keep some echo messages for debugging purposes in case something does not work

core-chart/templates/argocd/setup.yaml
```diff
@@ -4,117 +4,52 @@ kind: ConfigMap
...
+    kubectl -n $NS cp "$kubeconfig" ${ARGOCD_POD}:/tmp
+    rm "$kubeconfig"
+    
+    echo "[$(date +"%Y-%m-%d %H:%M:%S")] Registering Control Plane \"$cp_name\"..."
+    kubectl -n $NS exec $ARGOCD_POD -- argocd cluster add ${cp_name} -y --kubeconfig "$kubeconfig"
+    kubectl -n $NS exec $ARGOCD_POD -- rm "$kubeconfig"
     {{- end }}
 
-    timestamped_echo "Creating applications..."
+    echo "[$(date +"%Y-%m-%d %H:%M:%S")] Creating applications..."
```
francostellari: My suggestion would be to keep the info logging for debugging purposes.

core-chart/templates/argocd/setup.yaml
```diff
@@ -126,10 +61,8 @@ metadata:
     app.kubernetes.io/version: {{ .Chart.AppVersion }}
     helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
   annotations:
-    # This is what defines this resource as a hook. Without this line, the
-    # job is considered part of the release.
     "helm.sh/hook": post-install, post-upgrade
-    "helm.sh/hook-weight": "5" # IMPORTANT: set this weight higher than KubeFlex post-install hook, otherwise postgresql will not install
+    "helm.sh/hook-weight": "5"
```
francostellari: Please consider keeping this comment because it may help explain how the value was chosen and resolve issues in the future

core-chart/templates/argocd/setup.yaml
```diff
@@ -139,13 +72,104 @@ spec:
...
       securityContext:
         runAsNonRoot: true
       serviceAccountName: "{{ .Release.Name }}-argocd-kubestellar-setup"
+      
+      # INIT CONTAINERS - Handle all the waiting logic
+      initContainers:
+      
+      # INIT CONTAINER 1: Wait for ArgoCD Pod Creation
+      - name: wait-argocd-pod-create
+        image: quay.io/kubestellar/kubectl:{{ .Values.KUBECTL_VERSION | default "1.30.12" }}
```
francostellari: For all init containers, I would prefer not to add the default version because it's one more item to maintain in the future.

francostellari:
@Rupam-It since we do not test for the argo-cd in the workflow. Please consider adding a test or showing a local test.

Rupam-It:
> @Rupam-It since we do not test for the argo-cd in the workflow. Please consider adding a test or showing a local test.

Means argocd is working or not?( for me it's  working)!
And I'm raising an issue for this.

Also, bash test is failing ! In local it works fine for me...

Rupam-It:
Thanks @francostellari , for all comments I'm working on it...

francostellari:
Thank you!
/lgtm

core-chart/templates/argocd/setup.yaml
```diff
@@ -4,117 +4,57 @@ kind: ConfigMap
...
+  # 1. Register KubeStellar WDSes as Argo CD target clusters
+  # 2. Create any Argo CD applications listed in KubeStellar values
+  # Note: Waiting logic for dependencies moved to initContainers for better reliability
   argocd-setup.sh: |
     #!/usr/bin/env bash
     set -e
 
-    WAIT_BEFORE_RETRY=5
     NS={{ .Release.Namespace }}
+    echo "[$(date +"%Y-%m-%d %H:%M:%S")] Getting ArgoCD pod and credentials..."
```
MikeSpreitzer: I think that the time zone is much more interesting than the date (which will be clear from the context).

core-chart/templates/argocd/setup.yaml
```diff
@@ -139,13 +77,104 @@ spec:
...
+        - -n
+        - {{ .Release.Namespace }}
+        - --timeout=300s
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - ALL
+      
+      # INIT CONTAINER 4-N: Wait for Each Control Plane (Dynamic Loop)
```
MikeSpreitzer: I see a loop but it is not dynamic. The range of the loop is fixed in the "values" of the chart instance.

core-chart/templates/argocd/setup.yaml
```diff
@@ -156,9 +185,10 @@ spec:
...
           mountPath: /mnt/script
           readOnly: true
+      
       volumes:
       - name: script
         configMap:
           name: "{{ .Release.Name }}-argocd-kubestellar-setup"
           defaultMode: 0744
-{{- end }}
+{{- end }}
```
MikeSpreitzer: Please do not remove the newline at the end of the file.
image

core-chart/templates/argocd/setup.yaml
```diff
@@ -4,117 +4,57 @@ kind: ConfigMap
...
-
-      timestamped_echo "==> Copying kubeconfig \"$kubeconfig\" to Argo CD pod \"$ARGOCD_POD\"..."
-      kubectl -n $NS cp "$kubeconfig" ${ARGOCD_POD}:/tmp
-      rm "$kubeconfig"
-
-      timestamped_echo "==> Registering Control Plane \"$cp_name\" as an Argo CD cluster..."
-      kubectl -n $NS exec $ARGOCD_POD -- argocd cluster add ${cp_name} -y --kubeconfig "$kubeconfig"
-      kubectl -n $NS exec $ARGOCD_POD -- rm "$kubeconfig"
-
+    cp_name={{ $cp.name }}
```
MikeSpreitzer: Shouldn't this be indented 6 spaces instead of 4?

core-chart/templates/argocd/setup.yaml
MikeSpreitzer: The latest rev of this PR has a lot of debug logging, right?

core-chart/templates/argocd/setup.yaml
```diff
@@ -4,117 +4,57 @@ kind: ConfigMap
...
     {{- range $app := .Values.argocd.applications }}
-
-      timestamped_echo "==> Creating application \"{{ $app.name }}\":"
-      timestamped_echo "    - project=\"{{ $app.project | default "default" }}\""
-      timestamped_echo "    - repoURL=\"{{ $app.repoURL }}\""
-      timestamped_echo "    - targetRevision=\"{{ $app.targetRevision | default "HEAD" }}\""
-      timestamped_echo "    - path=\"{{ $app.path | default "/" }}\""
-      timestamped_echo "    - destinationWDS=\"{{ $app.destinationWDS }}\""
-      timestamped_echo "    - destinationNamespace=\"{{ $app.destinationNamespace | default "default" }}\""
-      timestamped_echo "    - syncPolicy=\"{{ $app.syncPolicy | default "manual" }}\""
```
MikeSpreitzer: However, there is a big loss in debug logging here.

core-chart/templates/argocd/setup.yaml
MikeSpreitzer: I see that the latest rev of this PR lacks those defaultings.

MikeSpreitzer (commented):
This looks pretty good to me. Just a few outstanding comments.

core-chart/templates/argocd/setup.yaml
francostellari: @MikeSpreitzer this is consistent with the existing chart, which is also not dynamic.
I think we can move to a dynamic loading in a separated PR where we use Argo CRDs to register the clusters.
I would try to keep this PR to a simplification of the existing bash script into smaller initCOntainers, most of which do not need bash.

core-chart/templates/argocd/setup.yaml
francostellari: I agree, I made the same comment

core-chart/templates/argocd/setup.yaml
Rupam-It: Hii @MikeSpreitzer in our previous code we use 
```
timestamped_echo() {
      echo "[$(date +"%Y-%m-%d %H:%M:%S")] $1"
    }
```
if you want then we can use : `echo "[$(date +"%H:%M:%S %Z")]`
log with global time!

core-chart/templates/argocd/setup.yaml
MikeSpreitzer: To be clear, I was not suggesting a change to making something dynamic here. I am objecting to the word "dynamic" appearing in the source comment (the line that my comment is attached to).

core-chart/templates/argocd/setup.yaml
```diff
@@ -4,105 +4,46 @@ kind: ConfigMap
 metadata:
   name: "{{ .Release.Name }}-argocd-kubestellar-setup"
 data:
-  # This script is used by the "{{ .Release.Name }}-argocd-kubestellar-setup" job to
-  # 1. register KubeStellar WDSes as Argo CD target clusters
-  # 2. create any Argo CD application listed in KubeStelar values
+  # This script is used by the "{{ .Release.Name }}-argocd-kubestellar-setup" job to:
+  # 1. Register KubeStellar WDSes as Argo CD target clusters
+  # 2. Create any Argo CD applications listed in KubeStellar values
+  # Note: Waiting logic for dependencies moved to initContainers for better reliability
```
MikeSpreitzer: FYI, IMO, this sort of comment belongs in the PR, not the file that everybody sees afterward. Most readers are not concerned with diff wrt earlier version.

core-chart/templates/argocd/setup.yaml
MikeSpreitzer: @Rupam-It: yes, I think that would be an improvement.

francostellari:
> For me, the only blocker is the removal of the debug logging at the point where the ArgoCD Application is created.

@Rupam-It, to be clear, @MikeSpreitzer and I would like to get back the debug output at line 97 to 103 that was removed.

core-chart/templates/argocd/setup.yaml
```diff
@@ -11,37 +11,54 @@ data:
...
+    timestamped_echo "Logging into Argo CD..."
     kubectl -n $NS exec $ARGOCD_POD -- argocd login "{{ .Release.Name }}-argocd-server.$NS" --username admin --password $(echo $ARGOCD_PASSWD | base64 -d) --insecure
 
-    echo "[$(date +"%H:%M:%S %Z")] Adding WDS clusters..."
+    timestamped_echo "Adding WDS clusters..."
     {{- range $cp := .Values.WDSes }}
       cp_name={{ $cp.name }}
-      echo "[$(date +"%H:%M:%S %Z")] Processing WDS cluster: $cp_name"
+      timestamped_echo "==> Looking for Control Plane \"$cp_name\"..."
+      timestamped_echo "==> Control Plane \"$cp_name\" is ready."
```
MikeSpreitzer, line 31: What guarantees that the ControlPlane is ready at this time?

PR #3116, opened 2025-07-22
🐛 Add 'v' prefix to Helm version for consistent comparison of wantver and getver in check prerequisites script

greninja517:
## Summary
The [is_installed](https://github.com/kubestellar/kubestellar/blob/main/scripts/check_pre_req.sh#L29) function compares structured versions using lexical comparison, and expects both `gotver` and `wantver` to be similarly formatted. And, Helm returns version prefixed without `v` e.g. (`3.18.1`), but the `wantver` is prefixed  with 'v' e.g. (`v3`), causing incorrect comparisons and script run failure.

So, Added `'v'` prefix to the output of `gotver` by wrapping it inside [is_installed_helm()](https://github.com/kubestellar/kubestellar/blob/main/scripts/check_pre_req.sh#L122) function, ensuring both `gotver` and `wantver` are formatted as `vX.Y.Z`.
## Related issue(s)

Fixes #

### DIFF
--- a/scripts/check_pre_req.sh
+++ b/scripts/check_pre_req.sh
@@ -124,7 +124,7 @@ is_installed_helm() {
     is_installed 'Helm' \
         'helm' \
         'helm version' \
-        'helm version --template={{.Version}}' \
+        'echo "v$(helm version --template={{.Version}})"' \
         'https://helm.sh/docs/intro/install/' \
         'v3'
     if log=$(helm show chart oci://ghcr.io/kubestellar/kubestellar/core-chart 2>&1)


MikeSpreitzer:
> Helm returns version prefixed without v e.g. (3.18.1)

Huh? That is not what I see:

```console
mspreitz@mjs13 kubestellar % helm version --template={{.Version}}; echo
v3.18.3
mspreitz@mjs13 kubestellar % 
```

Also, look at the bash-based E2E test of this PR. https://github.com/kubestellar/kubestellar/actions/runs/16450732808/job/46494993514?pr=3116#step:7:55 reports "✔ Helm (vv3.18.4)".

greninja517:
Yeah, I re-checked the script. There was some misconfiguration in my local workstation HELM installation. Thanks !

PR #3119, opened 2025-07-23
✨feature: Detecting duplicate object identifier detection across WDSes

PranjaliBhardwaj:
Why this?
1. Prevents Subtle and Hard-to-Debug Errors
2. Surfaces User/Operator Misconfiguration
Aligns with Project Roadmap and Known Limitations
4. Improves Operator Experience

Issue solved: #3018 
It lists all WDSes, aggregates all workload object identifiers, and logs a warning if any identifier appears in more than one WDS.
Previous pull request got polluted. This is working well.

### DIFF
--- a/cmd/controller-manager/duplicate_detector.go
+++ b/cmd/controller-manager/duplicate_detector.go
@@ -1,0 +1,150 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	ctrlutil "github.com/kubestellar/kubestellar/pkg/ctrlutil"
+	kfv1alpha1 "github.com/kubestellar/kubeflex/api/v1alpha1"
+	kslclient "github.com/kubestellar/kubestellar/pkg/client"
+)
+
+type ObjectIdentifier struct {
+	Group     string
+	Version   string
+	Kind      string
+	Name      string
+	Namespace string
+}
+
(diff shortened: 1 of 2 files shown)


cmd/controller-manager/duplicate_detector.go
```diff
@@ -0,0 +1,150 @@
...
+	Group     string
+	Version   string
+	Kind      string
+	Name      string
+	Namespace string
+}
+
+// DetectDuplicateObjectIdentifiersAcrossWDSes checks for duplicate object identifiers across all WDSes and logs warnings if found.
+func DetectDuplicateObjectIdentifiersAcrossWDSes(logger logr.Logger) error {
+	ctx := context.Background()
```
MikeSpreitzer, line 32: The usual pattern is to pass the `ctx context.Context` as a parameter and then use `klog.FromContext` to extract the logger from `ctx`.

cmd/controller-manager/main.go
```diff
@@ -115,6 +115,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Duplicate detection across WDSes
```
MikeSpreitzer, line 118: This is doing the check once, at controller startup. Is that really what you intended?

cmd/controller-manager/duplicate_detector.go
```diff
@@ -0,0 +1,150 @@
...
+	idToWDS := map[ObjectIdentifier][]string{}
+
+	for _, wds := range wdsList.Items {
+		wdsName := wds.Name
+		restConfig, _, err := ctrlutil.GetWDSKubeconfig(logger, wdsName)
+		if err != nil {
+			logger.Error(err, "Failed to get kubeconfig for WDS", "wds", wdsName)
+			continue
+		}
+		dynClient, err := dynamic.NewForConfig(restConfig)
```
MikeSpreitzer, line 51: Please learn from the code in the Binding controller that also enumerates all the workload objects in a WDS.

Rather than have two copies of this code, it would be much better to refactor to a DRY solution.

MikeSpreitzer:
Please provide evidence of successful testing.

Also consider: a truly supported feature will have test(s) for it in CI.

MikeSpreitzer:
Ping @PranjaliBhardwaj

Issue #2717, opened 2025-01-11
bug: secret vc-vcluster lacks config-incluster


MikeSpreitzer:
### Describe the bug

While investigating #2579, #2700, #2701, #2702, #2715, and #2716, I did some more testing. This was on the same IBM Cloud VM, and commit 45fc718da51dd5b193840584e3c2db0e271d3892 . This time, after many successful iterations, the script hung waiting for the its-with-cluteradm Job to be completed. But that Job will never complete, it failed all allowed retries. It failed because the Secret named "vc-vcluster" in the "its1-system" namespace in the KubeFlex hosting cluster does _NOT_ have a data entry named "config-incluster" --- but that job's pods are defined to use it as their kubeconfig. In particular, they have an environment variable definition for KUBECONFIG that is the pathname where that data entry would appear (in a mounted volume based on that Secret).

I extracted the log of the KubeFlex controller-manager with the following command.

```shell
kubectl  --context kind-kubeflex logs -n kubeflex-system kubeflex-controller-manager-7db8894656-f6nsk > kfcm.log
```

I have attached that log, along with the full typescript.

[kfcm.log](https://github.com/user-attachments/files/18386294/kfcm.log)

[fail8.log](https://github.com/user-attachments/files/18386299/fail8.log)

### Output from KubeStellar-Snapshot.sh

```console
mspreitz@mjs-dev7a:~$ bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/heads/main/scripts/kubestellar-snapshot.sh) -V -Y -L
KubeStellar Snapshot v0.2.0{COLOR_NONE}

Script run on 2025-01-11_15:06:05
Checking script dependencies:
✔ kubectl version v1.29.10 at /usr/local/bin/kubectl
✔ helm version v3.16.3 at /usr/sbin/helm
✔ jq version jq-1.6 at /usr/bin/jq
Using kubeconfig(s): /home/mspreitz/.kube/config
Validating contexts(s): 
✔ cluster1 
✔ cluster2 
✔ its1 *
✔ kind-kubeflex 
✔ wds1 

KubeStellar:
- Helm chart ks-core (v0.26.0-alpha.3) in namespace default in context kind-kubeflex
  - Secret=sh.helm.release.v1.ks-core.v1 in namespace default

KubeFlex:
- kubeflex-system namespace in context kind-kubeflex
[...]

PR #3121, opened 2025-07-24
✨ feat: split ITS PCHs into status and clusteradmin (#3024)

### LINKED ISSUE #3024: feature: Replace the current ITS PCHs with/without clusteradmin
### Feature Description

Replace the current ITS PCHs with/without clusteradmin with a PCH for just status and one for just cluster admin

### Proposed Solution

Leverage the new KubeFlex https://github.com/kubestellar/kubeflex/pull/384

1. Replace the current ITS PCHs with/without clusteradmin with a PCH for just status and one for just cluster admin
2. clean the _helpt.tpl
3. update the `its.yaml` for the CP definition to use the appropriate combination of the new PCHs accordingly to the `install_clusteradm` value

### Want to contribute?

### Additional Context

_No response_

rishi-jat:
fixes #3024


rishi-jat:
@MikeSpreitzer

core-chart/templates/controlplanes/its.yaml
```diff
@@ -7,8 +7,14 @@ metadata:
...
   backend: shared
   type: {{ $cp.type | default "vcluster" }}
-  postCreateHook: {{ (eq $cp.install_clusteradm false) | ternary "its-without-clusteradm" "its-with-clusteradm" }}
-  postCreateHookVars:
+  postCreateHooks:
+    {{- if $cp.install_clusteradm }}
+    - install-status-addon
+    - its-hub-init
+    {{- else }}
+    - install-status-addon
```
MikeSpreitzer: It would be simpler to have the common hooks listed outside the `if` construct.

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -0,0 +1,69 @@
...
+                - name: kubeconfig
+                  mountPath: "/etc/kube"
+                  readOnly: true
+          volumes:
+            - name: kubeconfig
+              secret:
+                secretName: "{{"{{.ITSSecretName}}"}}"
+          restartPolicy: Never
+      backoffLimit: 1
+{{- end }}
```
MikeSpreitzer: Is there any way that I could get you to have a newline at the end of your files, so that GitHub does not display the red circle-dash alert?
image

scripts/kubestellar-snapshot.sh
```diff
@@ -420,9 +420,15 @@ for i in "${!cps[@]}" ; do # for all control planes in context ${context}
...
-    cp_kubeconfig_content=$(get_kubeconfig "${helm_context}" "${cp_name[cp_n]}" "${cp_type[cp_n]}")
-    echo -e "- ${COLOR_INFO}${cp_name[cp_n]}${COLOR_NONE}: type=${COLOR_INFO}${cp_type[cp_n]}${COLOR_NONE}, pch=${COLOR_INFO}${cp_pch[cp_n]}${COLOR_NONE}, context=${COLOR_INFO}${cp_context[cp_n]}${COLOR_NONE}, namespace=${COLOR_INFO}${cp_name[cp_n]}-system${COLOR_NONE}"
+    cp_pchs=($(kubectl --context $helm_context get controlplane $name -o jsonpath='{.spec.postCreateHooks[*]}'))
+    cp_globalvars=$(kubectl --context $helm_context get controlplane $name -o jsonpath='{.spec.globalVars}')
+    # For backward compatibility, also try postCreateHook if postCreateHooks is empty
+    if [[ -z "${cp_pchs[*]}" ]]; then
+      cp_pchs=($(kubectl --context $helm_context get controlplane $name -o jsonpath='{.spec.postCreateHook}'))
+    fi
+    # Print control plane info
+    pch_list="${cp_pchs[*]}"
```
MikeSpreitzer: Why introduce `pch_list`? Why not just use `cp_pchs` instead in the next command?

core-chart/templates/postcreatehooks/its-clusteradmin.yaml
MikeSpreitzer, line 1: Why is this file here?

core-chart/templates/postcreatehooks/its-status.yaml
MikeSpreitzer, line 1: Why is this file here?

MikeSpreitzer:
@rishi-jat: You should be testing your contributions.

francostellari:
@rishi-jat focus on passing the test locally then we can work on cleaning up the PR.

Your error https://github.com/kubestellar/kubestellar/actions/runs/16496875610/job/46644656573?pr=3121#step:7:228
and  https://github.com/kubestellar/kubestellar/actions/runs/16496875610/job/46644656573?pr=3121#step:7:229

Are probably related to the extra files in the postcreatehooks folder pointed out by @MikeSpreitzer

francostellari:
@rishi-jat your failure https://github.com/kubestellar/kubestellar/actions/runs/16496875610/job/46644656573?pr=3121#step:7:227 is due to the wrong syntax in listing the hooks.

See the WDS CP definition as an example: https://github.com/kubestellar/kubestellar/blob/main/core-chart/templates/controlplanes/wds.yaml

rishi-jat:
@MikeSpreitzer @francostellari
I’ve properly fixed the issue now and tested everything locally to ensure it works as expected.

rishi-jat:
Please have a look and let me know if any further improvements are needed.

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -1,25 +1,25 @@
...
+        spec:
+          containers:
+{{ include "pch.its.statusaddon" . | indent 10 }}
+          volumes:
+          - name: kubeconfig
+            secret:
+              secretName: "{{"{{.ITSSecretName}}"}}"
+          restartPolicy: Never
+      backoffLimit: 1
+{{- end }}
```
MikeSpreitzer: You do not need a space at the end of this line, as core-chart/templates/postcreatehooks/its-hub-init.yaml demonstrates. A space (character code 0x20) is different from a newline (character code 0x0a).

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -1,25 +1,25 @@
...
+  templates:
+  - apiVersion: batch/v1
+    kind: Job
+    metadata:
+      name: "{{"{{.HookName}}"}}"
+    spec:
+      template:
+        spec:
+          containers:
+{{ include "pch.its.statusaddon" . | indent 10 }}
```
MikeSpreitzer: It would be better to move that content inline. Could be done in a follow-on PR.

docs/content/direct/get-started.md
```diff
@@ -141,8 +141,10 @@ kflex ctx --overwrite-existing-context its1
 The Helm chart above has a Job that initializes the ITS as an OCM "hub" cluster. Helm does not have a way to wait for that initialization to finish. So you have to do the wait yourself. The following commands will do that.
 
 ```shell
-kubectl --context kind-kubeflex wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.its-with-clusteradm}=true' --timeout 90s
-kubectl --context kind-kubeflex wait -n its1-system job.batch/its-with-clusteradm --for condition=Complete --timeout 150s
+kubectl --context kind-kubeflex wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.install-status-addon}=true' --timeout 90s
+kubectl --context kind-kubeflex wait -n its1-system job.batch/install-status-addon --for condition=Complete --timeout 150s
+kubectl --context kind-kubeflex wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.its-hub-init}=true' --timeout 90s
+kubectl --context kind-kubeflex wait -n its1-system job.batch/its-hub-init --for condition=Complete --timeout 150s
```
MikeSpreitzer, line 147: @francostellari , @Rupam-It: it is now sufficient to wait for the `ControlPlane` to be "ready", because that encompases both PCHs having their `Job` objects created and finished running, right?

The same comment applies in `scripts/create-kubestellar-demo-env.sh` and `test/e2e/common/setup-kubestellar.sh`.

docs/content/direct/get-started.md
```diff
@@ -141,8 +141,10 @@ kflex ctx --overwrite-existing-context its1
 The Helm chart above has a Job that initializes the ITS as an OCM "hub" cluster. Helm does not have a way to wait for that initialization to finish. So you have to do the wait yourself. The following commands will do that.
 
 ```shell
-kubectl --context kind-kubeflex wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.its-with-clusteradm}=true' --timeout 90s
-kubectl --context kind-kubeflex wait -n its1-system job.batch/its-with-clusteradm --for condition=Complete --timeout 150s
```
MikeSpreitzer, line 145: These timeouts should have been changed long ago to be much longer. I will not insist on fixing this problem in this PR.

docs/content/direct/get-started.md
francostellari: @MikeSpreitzer , not yet.
I have opened issue #3145

francostellari:
@rishi-jat please delete the `_helpers.tpl` file and move the 2 sections of code inside the 2 PCHs.
Thank you.

rishi-jat:
@MikeSpreitzer @francostellari I've completed the inlining of both PCH templates and removed `_helpers.tpl` as requested. Please take another look.

MikeSpreitzer:
This PR introduces a synchronization problem. Before the change in this PR, we have two sequenced initContainers, one doing `clusteradm init --wait` and the next installing the status add-on. Now we have two independent Jobs doing those. The status add-on install can fail because the OCM hub initialization is not done yet. You can see this in both test suites in https://github.com/kubestellar/kubestellar/actions/runs/16602462585 . The first run of the install-status-addon Pod fails because a needed CRD is not "established" yet.

MikeSpreitzer:
The synchronization problem could be fixed by adding to the Job. that installs the status add-on an initContainer that does `kubectl wait` for the CRD for `ManagedClusterSetBinding` to be "established".

MikeSpreitzer:
No, I was wrong, the current ITS init Job does not use two initContainers, it uses two plain containers. That would explain some flakiness.

rishi-jat:
@MikeSpreitzer, I’ve added the synchronization logic as discussed. Could you help guide me on why the checks are still failing? I’d appreciate any pointers.

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -14,6 +14,16 @@ spec:
     spec:
       template:
         spec:
+          initContainers:
+          - name: wait-for-its-hub-init
+            image: quay.io/kubestellar/kubectl:{{.Values.KUBECTL_VERSION | default "v1.28.0"}}
```
francostellari: remove the default version

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -14,6 +14,16 @@ spec:
...
+            command:
+            - /bin/sh
+            - -c
+            - |
+              echo "Waiting for its-hub-init job to complete..."
+              kubectl wait --for=condition=complete --timeout=600s job/its-hub-init -n {{"{{.ControlPlaneName}}"}}-system
```
francostellari: If your intention is to wait for a CRD to exist then please add a specific wait --for=create for the CRD

francostellari:
> This PR introduces a synchronization problem. Before the change in this PR, we have two sequenced initContainers, one doing `clusteradm init --wait` and the next installing the status add-on. Now we have two independent Jobs doing those. The status add-on install can fail because the OCM hub initialization is not done yet. You can see this in both test suites in https://github.com/kubestellar/kubestellar/actions/runs/16602462585 . The first run of the install-status-addon Pod fails because a needed CRD is not "established" yet.

@MikeSpreitzer I'm unclear about your statement. I don't believe the PCH used before this PR used initContainers to syn/wait for the OCM installation to complete. It consisted of 2 containers in the same job, I would think they were also running concurrently.

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -0,0 +1,107 @@
...
+              - sh
+              - -c
+              - |
+                echo "Waiting for OCM CRDs to be established..."
+                # Wait for core OCM CRDs that clusteradm creates
+                kubectl wait --for=condition=Established crd/managedclusters.cluster.open-cluster-management.io --timeout=300s
+                kubectl wait --for=condition=Established crd/managedclustersets.cluster.open-cluster-management.io --timeout=300s
+                kubectl wait --for=condition=Established crd/managedclustersetbindings.cluster.open-cluster-management.io --timeout=300s
+                kubectl wait --for=condition=Established crd/placements.cluster.open-cluster-management.io --timeout=300s
+                kubectl wait --for=condition=Established crd/placementdecisions.cluster.open-cluster-management.io --timeout=300s
```
MikeSpreitzer: Do we really need to wait for all of these?
Are all of these used while installing the status add-on?

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -0,0 +1,107 @@
...
+  templates:
+  - apiVersion: batch/v1
+    kind: Job
+    metadata:
+      name: "{{"{{.HookName}}"}}"
+    spec:
+      template:
+        spec:
+          initContainers:
+          - name: wait-for-its-hub-init
```
MikeSpreitzer: What does this initContainer add? Is not the waiting for the CRD establishments enough?

Remember that using the `its-hub-init` PCH is optional.

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -0,0 +1,107 @@
...
+                echo "All required OCM CRDs are established"
+            env:
+            - name: KUBECONFIG
+              value: "{{"/etc/kube/{{.ITSkubeconfig}}"}}"
+            volumeMounts:
+            - name: kubeconfig
+              mountPath: "/etc/kube"
+              readOnly: true
+          containers:
+          - name: "{{"{{.HookName}}-statusaddon"}}"
```
MikeSpreitzer: FYI, "-statusaddon" is redundant with the hook name.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -198,3 +202,5 @@ then
     echo "Failed to see two clusters."
     exit 1
 fi
+
+
```
MikeSpreitzer: No need to add blank lines here.

MikeSpreitzer:
@francostellari: yes, I wrote something wrong. In https://github.com/kubestellar/kubestellar/pull/3121#issuecomment-3134518927 I made a correction.

To restate, there were and remain two synchronization problems. I want this PR to solve both of them, I am tired of the flakes.

- Before using it, ensure that the in-cluster data member exists in the Secret that holds the kubeconfigs for accessing the ITS.
- Ensure that the OCM hub initialization has finished before starting to install the OCM Status Add-On.

core-chart/templates/postcreatehooks/its-hub-init.yaml
```diff
@@ -0,0 +1,50 @@
...
+        spec:
+          initContainers:
+          - name: wait-for-config-incluster
+            image: quay.io/kubestellar/kubectl:{{.Values.KUBECTL_VERSION}}
+            command:
+              - sh
+              - -c
+              - |
+                echo "Waiting for config-incluster in Secret {{.ITSSecretName}}..."
+                until kubectl get secret "{{.ITSSecretName}}" -n {{.ControlPlaneName}}-system -o jsonpath="{.data.config-incluster}" | grep -q .; do
```
MikeSpreitzer: This `kubectl` command runs with `$KUBECONFIG` referring to the thing it is intended to wait for: the in-cluster kubeconfig data member of the Secret holding kubeconfigs for accessing the ITS. This is bad because it is circular, worse because it is looking in the wrong place. The Secret holding the kubeconfigs for accessing the ITS is in the KubeFlex hosting cluster.

MikeSpreitzer: There is good news and bad news for you in https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod:

> When a volume contains data from a Secret, and that Secret is updated, Kubernetes tracks this and updates the data in the volume, using an eventually-consistent approach.

What this initContainer needs to accomplish is to ensure that when the regular container tries to use `/etc/kube/{{ .ITSkubeconfig }}`, that file exists. Because the kubelet's update of the volume is only eventually consistent with the Secret, waiting for the data member to exist in the Secret is not enough.

The good news is that it is enough to wait for the file to appear.

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -0,0 +1,107 @@
...
+            command:
+              - sh
+              - -c
+              - |
+                echo "Waiting for its-hub-init job to complete..."
+                kubectl wait --for=condition=complete --timeout=600s job/its-hub-init -n {{"{{.ControlPlaneName}}"}}-system
+                echo "its-hub-init job completed successfully"
+            env:
+            - name: KUBECONFIG
+              value: "{{"/etc/kube/{{.ITSkubeconfig}}"}}"
```
MikeSpreitzer: This initContainer and the other assume that the file `/etc/kube/{{.ITSkubeconfig}}` already exists, which is one of the synchronization gaps that this PR needs to fix.

rishi-jat:
I’ve implemented all your review suggestions for PR :
	•	Fixed the synchronization gap by replacing the circular kubectl get secret logic with a proper check for           kubeconfig file existence.
	•	Reduced CRD waiting to only the essential ones: placements and placementdecisions.
	•	Added an initContainer to wait for the kubeconfig file before proceeding.
	•	Removed the redundant -statusaddon suffix from the container name.
	•	Cleaned up unnecessary whitespace.

rishi-jat:
@MikeSpreitzer Testing confirms that the changes work as expected. The initContainers now execute in the correct order and wait for the appropriate resources. The remaining clusteradm hanging issue appears to be a pre-existing infrastructure problem and not related to the changes in this PR.

MikeSpreitzer:
> Reduced CRD waiting to only the essential ones: placements and placementdecisions

Those are not the two I saw in the log I looked at recently. There it was placement and (something)managed(something).

See https://github.com/kubestellar/kubestellar/issues/2201#issuecomment-3134414596

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -0,0 +1,123 @@
...
+              - -c
+              - |
+                echo "Waiting for OCM CRDs to be established..."
+                # Wait for essential OCM CRDs needed by the Status Add-on
+                # First check if clusteradm has created any OCM CRDs
+                echo "Checking if OCM CRDs exist..."
+                max_attempts=60
+                attempt=0
+                while [ $attempt -lt $max_attempts ]; do
+                  if kubectl get crd placements.cluster.open-cluster-management.io >/dev/null 2>&1; then
```
MikeSpreitzer, line 48: Is it enough to only wait for one of them to be created? Are we guaranteed that this one is created last?

core-chart/templates/postcreatehooks/its-hub-init.yaml
```diff
@@ -0,0 +1,56 @@
...
+  templates:
+  - apiVersion: batch/v1
+    kind: Job
+    metadata:
+      name: "{{"{{.HookName}}"}}"
+    spec:
+      template:
+        spec:
+          initContainers:
+
```
MikeSpreitzer, line 18: We do not need a blank line here.

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -0,0 +1,123 @@
...
+  - apiVersion: batch/v1
+    kind: Job
+    metadata:
+      name: "{{"{{.HookName}}"}}"
+    spec:
+      template:
+        spec:
+          initContainers:
+          - name: wait-for-kubeconfig
+            image: quay.io/kubestellar/kubectl:{{.Values.KUBECTL_VERSION}}
```
MikeSpreitzer, line 19: Could we use `busybox` instead of our kubectl image? We want to work towards removing `bash` from our kubectl image.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -197,4 +201,4 @@ if ! expect-cmd-output 'kubectl --context its1 get managedclusters -l location-g
 then
     echo "Failed to see two clusters."
     exit 1
-fi
+fi
```
MikeSpreitzer, line 204: Please do not delete the linebreak at the end of the file. That deletion makes GitHub display an alarming character here:

image

MikeSpreitzer (approved):
More work can be done in follow-on PRs. This is good enough to merge. It looks like it will help with #2201 (#2717).

/lgtm
/approve
Thanks!

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -0,0 +1,123 @@
...
+                fi
+            env:
+            - name: KUBECONFIG
+              value: "{{"/etc/kube/{{.ITSkubeconfig}}"}}"
+            volumeMounts:
+            - name: kubeconfig
+              mountPath: "/etc/kube"
+              readOnly: true
+          containers:
+          - name: "{{"{{.HookName}}"}}"
```
MikeSpreitzer, line 72: This container exists within the context of a Pod that is specific to this PCH. In my opinion, the name of this hook does not need to be repeated here.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -127,8 +127,10 @@ else
 popd
 
 : Waiting for OCM hub to be ready...
-kubectl wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.its-with-clusteradm}=true' --timeout 400s
-kubectl wait -n its1-system job.batch/its-with-clusteradm --for condition=Complete --timeout 400s
+kubectl wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.install-status-addon}=true' --timeout 400s
+kubectl wait -n its1-system job.batch/install-status-addon --for condition=Complete --timeout 400s
+kubectl wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.its-hub-init}=true' --timeout 400s
+kubectl wait -n its1-system job.batch/its-hub-init --for condition=Complete --timeout 400s
```
MikeSpreitzer, line 133: OOPS! I should not have approved this change. It is not right when the latest _release_ of the Helm chart is being used --- that has not changed yet.

docs/content/direct/get-started.md
MikeSpreitzer, line 147: OOPS! I should not have approved this change. It is not right because the latest release of the Helm chart is being used --- that has not changed yet.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -228,8 +228,10 @@ kflex ctx --overwrite-existing-context wds2
 kflex ctx --overwrite-existing-context its1
 
 echo -e "\nWaiting for OCM cluster manager to be ready..."
-kubectl --context $k8s_platform-kubeflex wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.its-with-clusteradm}=true' --timeout 24h
-kubectl --context $k8s_platform-kubeflex wait -n its1-system job.batch/its-with-clusteradm --for condition=Complete --timeout 24h
+kubectl --context $k8s_platform-kubeflex wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.install-status-addon}=true' --timeout 24h
+kubectl --context $k8s_platform-kubeflex wait -n its1-system job.batch/install-status-addon --for condition=Complete --timeout 24h
+kubectl --context $k8s_platform-kubeflex wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.its-hub-init}=true' --timeout 24h
+kubectl --context $k8s_platform-kubeflex wait -n its1-system job.batch/its-hub-init --for condition=Complete --timeout 24h
```
MikeSpreitzer, line 234: OOPS! I should not have approved this change. It is not right because the latest release of the Helm chart is being used --- that has not changed yet.

test/e2e/common/setup-kubestellar.sh
```diff
@@ -163,8 +165,10 @@ function add_wec() {
 
 "${SRC_DIR}/../../../scripts/check_pre_req.sh" --assert --verbose ocm
 
-kubectl --context $HOSTING_CONTEXT wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.its-with-clusteradm}=true' --timeout 200s
-kubectl --context $HOSTING_CONTEXT wait -n its1-system job.batch/its-with-clusteradm --for condition=Complete --timeout 400s
+kubectl --context $HOSTING_CONTEXT wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.install-status-addon}=true' --timeout 200s
+kubectl --context $HOSTING_CONTEXT wait -n its1-system job.batch/install-status-addon --for condition=Complete --timeout 400s
+kubectl --context $HOSTING_CONTEXT wait controlplane.tenancy.kflex.kubestellar.org/its1 --for 'jsonpath={.status.postCreateHooks.its-hub-init}=true' --timeout 200s
+kubectl --context $HOSTING_CONTEXT wait -n its1-system job.batch/its-hub-init --for condition=Complete --timeout 400s
```
MikeSpreitzer, line 171: BTW, this is a (functionally equivalent but textually different) copy of an earlier block of code.
I do not see a good reason for this duplication.
I introduced the duplication in #2546, perhaps unaware that I was making a duplication.

core-chart/templates/postcreatehooks/install-status-addon.yaml
rishi-jat: Good point. I’ll update the script to wait for both placements and placementdecisions CRDs to ensure all required resources are ready.

core-chart/templates/postcreatehooks/install-status-addon.yaml
rishi-jat: Makes sense. I’ll switch to busybox for this initContainer since we don’t need kubectl or bash here.

rishi-jat:
@MikeSpreitzer Thanks for the feedback. I’ll address all of these switching to busybox, waiting for the correct CRDs, and cleaning up formatting  in a follow-up PR soon.

core-chart/templates/postcreatehooks/its-hub-init.yaml
```diff
@@ -0,0 +1,56 @@
...
+          - name: wait-for-config-incluster
+            image: quay.io/kubestellar/kubectl:{{.Values.KUBECTL_VERSION}}
+            command:
+              - sh
+              - -c
+              - |
+                echo "Waiting for kubeconfig file /etc/kube/{{.ITSkubeconfig}} to exist and be non-empty..."
```
MikeSpreitzer, line 25: Something is wrong here. I am testing `main` after merging this PR, using `test/e2e/run-test.sh`. When I `kubectl get -o yaml` this Pod, I find the following (which leads to premature success, and so the flake remains).

```yaml
  - command:
    - sh
    - -c
    - |
      echo "Waiting for kubeconfig file /etc/kube/ to exist and be non-empty..."
      until [ -s /etc/kube/ ]; do
        echo "Kubeconfig file not ready yet, waiting..."
        sleep 2
      done
      echo "Kubeconfig file is ready"
```

Oddly, I also find the following in the regular container.

```yaml
  containers:
  - args:
    - init
    - -v=2
    - --wait
    - --timeout=300
    env:
    - name: KUBECONFIG
      value: /etc/kube/config-incluster
```

MikeSpreitzer: That is due to a quoting problem. To debug it, I changed the source to the following.

```yaml
            command:
              - sh
              - -c
              - |
                echo "Form 1: /etc/kube/{{.ITSkubeconfig}}"
                echo Form 2:  /etc/kube/{{.ITSkubeconfig}}
                echo Form 3: "/etc/kube/{{.ITSkubeconfig}}"
                echo Form 4:  {{"/etc/kube/{{.ITSkubeconfig}}"}}
                echo Form 5: "{{"/etc/kube/{{.ITSkubeconfig}}"}}"
                echo "Waiting for kubeconfig file '$KUBECONFIG' to exist and be non-empty..."
                until [ -s "$KUBECONFIG" ]; do
                  echo "Kubeconfig file not ready yet, waiting..."
                  sleep 2
                done
                echo "Kubeconfig file is ready"
            env:
            - name: KUBECONFIG
              value: "{{"/etc/kube/{{.ITSkubeconfig}}"}}"
```

After the Pod was constructed and run, a `kubectl get -o yaml` of it showed me the following.

```yaml
  - command:
    - sh
    - -c
    - |
      echo "Form 1: /etc/kube/"
      echo Form 2:  /etc/kube/
      echo Form 3: "/etc/kube/"
      echo Form 4:  /etc/kube/config-incluster
      echo Form 5: "/etc/kube/config-incluster"
      echo "Waiting for kubeconfig file '$KUBECONFIG' to exist and be non-empty..."
      until [ -s "$KUBECONFIG" ]; do
        echo "Kubeconfig file not ready yet, waiting..."
        sleep 2
      done
      echo "Kubeconfig file is ready"
    env:
    - name: KUBECONFIG
      value: /etc/kube/config-incluster
```

And the log:

```console
mspreitz@mjs13 ~ % kubectl --context kind-kubeflex logs -n its1-system its-hub-init-ljwbx -c wait-for-config-incluster
Form 1: /etc/kube/
Form 2: /etc/kube/
Form 3: /etc/kube/
Form 4: /etc/kube/config-incluster
Form 5: /etc/kube/config-incluster
Waiting for kubeconfig file '/etc/kube/config-incluster' to exist and be non-empty...
Kubeconfig file is ready
```

MikeSpreitzer:
I have updated PR #3163 to include the fix to the quoting problem noted above.

PR #3126, opened 2025-07-25
🐛 bug: remove variable double definition in WDS CP (fixes #3124)

### LINKED ISSUE #3124: bug: Remove unnecessary double definition of `ITSName` variables in WDS CP
### Describe the bug

The line https://github.com/kubestellar/kubestellar/blob/6baafdad09f9dc8685eda1564f2e3ff8a956246a/core-chart/templates/controlplanes/wds.yaml#L14 is unnecessary and should be removed

The variable is not different than the globalVar definition

_No response_

### Steps To Reproduce

see above

### Expected Behavior

There is not need to define twice the same variable with the same value.
It could lead to issues down the road.

### Want to contribute?

### Additional Context

_No response_

flushthemoney:
## Summary

Removed the double definition of ITSName in the wds.yaml file, it now uses the one defined in the globalVars.

## Related issue(s)

Fixes #3124

### DIFF
--- a/core-chart/templates/controlplanes/wds.yaml
+++ b/core-chart/templates/controlplanes/wds.yaml
@@ -11,7 +11,6 @@ spec:
     - hookName: kubestellar-controller
       vars:
         APIGroups: "{{ .APIGroups | default "" }}"
-        ITSName: "{{ .ITSName | default "" }}"
     - hookName: transport-controller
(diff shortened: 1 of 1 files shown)


MikeSpreitzer (commented):
/lgtm

PR #3135, opened 2025-07-28
🌱 Refactor: Improve the wait logic for CSR approval of WECs in demo-env file

greninja517:
## Summary
This PR refactors the CSR approval function to be more efficient by switching from an inefficient polling loop with `sleep` to an event-driven `kubectl --watch`. Additionally, `timeout` has been added to prevent script from indefinite hangout.

## Related issue(s) #3075 

Fixes #

### DIFF
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -104,33 +104,28 @@ context_clean_up() {
 }
 
 checking_cluster() {
-    found=false
+    local timeout_duration="100s"
 
-    while true; do
+    while IFS= read -r line; do
+        # Check for the cluster name and "Pending" status
+        if echo "$line" | grep -q "$1" && echo "$line" | grep -q "Pending"; then
+            echo "Pending CSR for $1 has been found, approving..."
+            clusteradm --context its1 accept --clusters "$1"
 
-        output=$(kubectl --context its1 get csr)
-
-        while IFS= read -r line; do
-
-            if echo "$line" | grep -q $1; then
-                echo "$1 has been found, approving CSR"
-                clusteradm --context its1 accept --clusters "$1"
-                found=true
-                break
+            if [ $? -eq 0 ]; then
+                echo -e "\033[33m✔\033[0m CSR Approved for $1."
+                return 0
+            else
+                echo -e "\033[0;31mX\033[0m Failed to approve CSR for $1."
+                return 1
             fi
-
-        done <<< "$output"
-
-        if [ "$found" = true ]; then
-            break
-
-        else
-            echo "CSR for $1 not found. Trying again..."
-            sleep 20
         fi
+    done < <(timeout ${timeout_duration} kubectl --context its1 get csr --watch) 
 
-    done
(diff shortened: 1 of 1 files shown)


scripts/create-kubestellar-demo-env.sh
```diff
@@ -104,34 +104,29 @@ context_clean_up() {
...
-        done <<< "$output"
-
-        if [ "$found" = true ]; then
-            break
-
-        else
-            echo "CSR for $1 not found. Trying again..."
-            sleep 20
         fi
+    done < <(timeout ${timeout_duration} kubectl --context its1 get csr --watch)
```
MikeSpreitzer:
```
bash-3.2$ which timeout
bash-3.2$ 
```

MikeSpreitzer:
Not loving the addition of a timeout. See https://cloud-native.slack.com/archives/C097094RZ3M/p1753727175879999

scripts/create-kubestellar-demo-env.sh
greninja517: Okay, I will remove the timeout part.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -104,32 +104,21 @@ context_clean_up() {
...
-        while IFS= read -r line; do
-
-            if echo "$line" | grep -q $1; then
-                echo "$1 has been found, approving CSR"
-                clusteradm --context its1 accept --clusters "$1"
-                found=true
-                break
+    while IFS= read -r line; do
+        # Check for the cluster name and "Pending" status
+        if echo "$line" | grep -q "$1" && echo "$line" | grep -q "Pending"; then
```
MikeSpreitzer, line 109: What if `$1` is `Pending`? (This is no worse than the problems with `grep -q "$1"`)

MikeSpreitzer, line 109: Text searching in a table is a poor way to search that table. What if `$1` is `NAME`? What if `$1` is a substring of another cluster's name? What if `$1` is `ending`?

This PR did not introduce this problem, and I will not insist that it be fixed in this PR

MikeSpreitzer:
The opening comment is no longer accurate (wrt timeout).

MikeSpreitzer:
To be clear, this PR is replacing one bash loop with another; it is not reducing the population of bash loops.

MikeSpreitzer:
Successfully tested by me on my Mac.

[test.txt](https://github.com/user-attachments/files/21476054/test.txt)

MikeSpreitzer (approved):
This is good enough to merge.
The opening comment could use some updating.
/lgtm
/approve
Thanks!

scripts/create-kubestellar-demo-env.sh
greninja517: Even if $1 is `NAME` , the same line won't contain `Pending`, so the logic holds true for header line. But, we have to assume that user won't keep the cluster name ambiguous and meaningless like `Pending`, `Name`, etc.. 
But, for the substring issue, we can use `grep -q -w $1` to exactly match $1 only. I think this is good approach.

PR #3148, opened 2025-07-29
✨ enable waitForPostCreateHooks for ITSes

### LINKED ISSUE #3145: feature: Enable `waitForPostCreateHooks`for KubeStellar ITSes and WDSes
### Feature Description

Add `waitForPostCreateHooks=true` to ITSes and WDSes created by KubeStellar core-chart
Leverage PR https://github.com/kubestellar/kubeflex/pull/447

### Proposed Solution

Add `waitForPostCreateHooks=true` to ITSes and WDSes created by KubeStellar core-chart

### Want to contribute?

### Additional Context

_No response_

AritraDey-Dev:
## Summary
Enable `waitForPostCreateHooks`: true for ITSes in the core Helm chart.
## Related issue(s)

Fixes #3145

### DIFF
--- a/core-chart/templates/postcreatehooks/transport-controller.yaml
+++ b/core-chart/templates/postcreatehooks/transport-controller.yaml
@@ -46,6 +46,7 @@ spec:
         # input parameters
         cp_name="${1%-system}" # cp name or cp namespace
         guess_its_name="$2" # true: try guessing the name of the ITS CP
+        wait_for_cp_ready="$3"
 
         # check if the CP name is valid or needs to be guessed
         while [ "$cp_name" == "" ] ; do
@@ -68,11 +69,13 @@ spec:
           fi
         done
 
-        # wait for the CP to exists and be ready
-        while [[ $(kubectl get controlplane "$cp_name" -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do
-          >&2 echo "Waiting for \"$cp_name\" control plane to exist and be ready..."
-          sleep 10
-        done
+        if [ "$wait_for_cp_ready" == "true" ]; then
+          # wait for the CP to exist and be ready
+          while [[ $(kubectl get controlplane "$cp_name" -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do
+            >&2 echo "Waiting for \"$cp_name\" control plane to exist and be ready..."
+            sleep 10
+          done
+        fi
 
         # determine the secret name and namespace
         key=$(kubectl get controlplane $cp_name -o=jsonpath='{.status.secretRef.inClusterKey}')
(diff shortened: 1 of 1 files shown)


MikeSpreitzer (commented):
/lgtm
/approve

I expect this will enable simplification of the various chunks of waiting code. I do not remember whether some has already been merged as if this change were already in place.

@Rupam-It 
@rishi-jat

AritraDey-Dev:
Any thoughts why CI is failing?

MikeSpreitzer:
This seems to be getting bitten hard by the flakes noted in #2201 (#3098).

MikeSpreitzer:
With the help of #3163, I have figured out why the tests of this PR always fail. This PR completes a cycle of dependency.

At https://github.com/kubestellar/kubeflex/blob/v0.9.0/pkg/reconcilers/vcluster/reconciler.go#L159-L160 we see that the `config-incluster` data item is not put in the Secret until after the ControlPlane is marked "available".

At https://github.com/kubestellar/kubeflex/blob/v0.9.0/internal/controller/controlplane_controller.go#L244-L245 we see that if `hcp.Spec.WaitForPostCreateHooks != nil && *hcp.Spec.WaitForPostCreateHooks` then the ControlPlane is not marked as "available" until after the PCHs have finished.

MikeSpreitzer:
This PR cannot succeed until after https://github.com/kubestellar/kubeflex/issues/490 has been addressed, a new KubeFlex release made, and KubeStellar updated to use that new KubeFlex release.

/hold

MikeSpreitzer:
https://github.com/kubestellar/kubeflex/issues/490 has been addressed and #3309 updates the KubeFlex dependency to pick that up. @AritraDey-Dev, please rebase this PR onto `main` so we can try it!

francostellari:
I looked into this issue and I think understand what is happening.

When `waitForPostCreateHooks: true` then the `Ready` state of the `wds1` CP will not be set until all the PCHs have completed.

Since the initContainer of the PCH is waiting for the `Ready` state to be set before getting the kubeconfig.... it never happens because they are waiting for each other.

The initContainer does not require the `Ready` state but just a valid kubeconfig.
We need to go back and check when that is guaranteed.

@Rupam-It @pdettori can you check?

MikeSpreitzer:
Regarding ordering of kubeconfig and PostCreateHook:

https://github.com/kubestellar/kubeflex/blob/v0.9.1/internal/controller/controlplane_controller.go#L170 comes before https://github.com/kubestellar/kubeflex/blob/v0.9.1/internal/controller/controlplane_controller.go#L205 . The latter applies the PostCreateHooks. As for the former, it includes setting up the in-cluster kubeconfig. See https://github.com/kubestellar/kubeflex/blob/v0.9.1/pkg/reconcilers/vcluster/reconciler.go#L146 .

MikeSpreitzer:
@francostellari, let me check my understanding of your comment https://github.com/kubestellar/kubestellar/pull/3148#issuecomment-3361280104 . The code that waits for a ControlPlane to be ready is the `get-kubeconfig.sh` script (data item) in the ConfigMap named transport-controller-config in the PostCreateHook named transport-controller, which is one of the two used by the WDS ControlPlane. Have I got that right?

If I have that right, then we have a waiting cycle regarding the WDS ControlPlane but not the ITS ControlPlane. The WDS waits on itself and the ITS; the ITS waits on neither. It remains necessary for the WDS PCH to wait on the ITS, because nothing else does that synchronization.

MikeSpreitzer:
See https://github.com/AritraDey-Dev/kubestellar/pull/1

Issue #3146, opened 2025-07-29
feature: Modify transport-controller to accept UTS/WDS control plane name instead of kubeconfig


francostellari:
### Feature Description

Currently the transport-controller expects to receive the kubeconfig of the ITS and WDS as files mounted in  a config map... which is insecure and require a "complex" bash init script

I propose to modify the current behavior to accept control planes names in alternative to kubeconfigs, similarly to kubestellar-controller

### Proposed Solution

Change the logic and Go code for the WDS part:
- if `wds-kunbeconfig` parameter is specified  then use it (current behavior that we want to avoid using in the chart)
- if `wds-kunbeconfig` parameter is not specified and a new `wds-name` parameter is specified, then retrieve the kubeconfig from cluster using the secret of the specified `wds-name` (like ks-controller, no more need of mounting config maps)
- fail otherwise

Change the logic and Go code for the ITS part:
- if `transport-kubeconfig` parameter is specified  then use it (current behavior that we want to avoid using in the chart)
- if `transport-kubeconfig` parameter is not specified and a new `transport-name` parameter is specified, then retrieve the kubeconfig from cluster using the secret of the specified `transport-name` (like ks-controller does for WDSes, no more need of mounting config maps)
- if NONE of the above is specified AND 1! ITS exists in the cluster, then use that as `transport-name`
- fail otherwise

### Want to contribute?

- [ ] I would like to work on this issue.

shivansh-source:
hello @Rupam-It can i work on this issue . i feel i can provide a solution for this.

Rupam-It:
> hello [@Rupam-It](https://github.com/Rupam-It) can i work on this issue . i feel i can provide a solution for this.

Hii thanks @shivansh-source for your interest! this is not an `good first issue` !
If you want to help any suggestions will be greatful! 
else I'm requesting to look into other issues!

shivansh-source:
@Rupam-It      we can add new CLI flags to accept the control plane names as input instead of kubeconfig file path . then implement kubeconfig retrieval (helper)by creating a function which fetch the corresponding secret if the name is provided and if no name is provided but file path is given load kubeconfig from that file . return a error if non of them is provided.

francostellari:
[...]

PR #3152, opened 2025-07-30
✨ Enhance Transport Controller WDS Kubeconfig Resolution

Rupam-It:
## Related issue(s) #3146 (part1 wds)

Fixes #

BEFORE (Old Transport Controller):

```
Could only connect to WDS using complex client options (--wds-qps, --wds-burst, etc.)
No direct kubeconfig file support
No cluster discovery like KubeStellar controller
Had to use init containers + volume mounts (messy setup)
```

AFTER (Enhanced Transport Controller):

```
Can use direct kubeconfig files: --wds-kubeconfig=/path/to/file 🎯
Can do cluster discovery: --wds-name=my-wds (same as KubeStellar controller) 🎯
Smart priority: File path beats cluster discovery
No volume mounting needed - cleaner setup
```

### DIFF
--- a/pkg/transport/generic/cmd/kubeconfig-resolver.go
+++ b/pkg/transport/generic/cmd/kubeconfig-resolver.go
@@ -1,0 +1,60 @@
+/*
+Copyright 2025 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"fmt"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+
+	"github.com/kubestellar/kubestellar/pkg/ctrlutil"
+)
+
+// resolveWDSKubeconfig resolves WDS kubeconfig using priority order
+// Priority 1: Direct kubeconfig file path (--wds-kubeconfig)
+// Priority 2: Cluster-based discovery using WDS name (--wds-name) - same as kubestellar controller
+func resolveWDSKubeconfig(options *TransportOptions, logger klog.Logger) (*rest.Config, string, error) {
+	// Priority 1: Direct kubeconfig file path (highest priority)
+	if options.WdsKubeconfigPath != "" {
+		logger.Info("Using WDS kubeconfig from file", "path", options.WdsKubeconfigPath)
+		config, err := loadKubeconfigFromPath(options.WdsKubeconfigPath)
+		if err != nil {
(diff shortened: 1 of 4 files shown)


Rupam-It:
Hii @MikeSpreitzer(cc @francostellari ) please take a look to this PR !
the test case passed in local but it failed in github CI environment !
I'm miss any anything ?

francostellari:
@Rupam-It while you are waiting for @MikeSpreitzer to review the go code, can you show a mock scenario of how this would impact/simplify the helm chart?.
Thanks

pkg/transport/generic/cmd/kubeconfig-resolver.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// loadKubeconfigFromPath loads kubeconfig from file path
+func loadKubeconfigFromPath(kubeconfigPath string) (*rest.Config, error) {
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build config from kubeconfig file: %w", err)
+	}
+	return config, nil
+}
```
MikeSpreitzer, line 85: Why is this here?

pkg/transport/generic/cmd/options.go
```diff
@@ -33,6 +33,7 @@ type TransportOptions struct {
 	MaxSizeWrapped         int
 	MaxNumWrapped          int
 	WdsName                string
+	TransportName          string
```
MikeSpreitzer, line 36: This is not the best choice of name. I think that "ITSName" would be better. It is parallel to "WdsName", and its meaning is clearer. Remember that KubeStellar also has the concept of pluggable transport, but this field is _not_ about that.

pkg/transport/generic/cmd/kubeconfig-resolver.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/kubestellar/kubestellar/pkg/ctrlutil"
+)
+
+// resolveWDSKubeconfig resolves WDS kubeconfig using priority order
+// Priority 1: WdsClientOptions (handles --wds-kubeconfig flag)
+// Priority 2: Cluster-based discovery using WDS name (--wds-name)
+func resolveWDSKubeconfig(options *TransportOptions, logger klog.Logger) (*rest.Config, string, error) {
+	// Priority 1: Try WdsClientOptions first (this handles --wds-kubeconfig)
+	if wdsConfig, err := options.WdsClientOptions.ToRESTConfig(); err == nil {
```
MikeSpreitzer, line 34: Did you know that when run by a Pod in the KubeFlex hosting cluster, with `options.WdsClientOptions` having the empty string for the `loadingRules.ExplicitPath`, `ToRESTConfig()` will succeed and return a client config for accessing the KubeFlex hosting cluster?

MikeSpreitzer:
I see a test failure. I see no testing of the new functionality.

MikeSpreitzer:
https://github.com/kubestellar/kubestellar/issues/3146#issuecomment-3272016862

francostellari:
@Rupam-It I think this is solved by the PR #3428 by mounting the WDS kubeconfig as a secret.
Do we still need this?

A different story is mounting the ITS secret which is being partially addressed by PR #3442. It's still problematic accessing a secret in a different ns

francostellari:
@Rupam-It can we close this PR?
I think this should solve the original issue: #3442

PR #3168, opened 2025-08-01
🐛fix(pch): finalize ITS PCH sync, CRD wait, and busybox migration (follow-up to #3121)

rishi-jat:
### Summary

Finalizes the ITS PostCreateHook updates introduced in PR #3121 by addressing remaining feedback. Improves synchronization reliability and CRD readiness handling.

### Related issue(s)

Follow-up to #3121


core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -29,7 +29,7 @@ spec:
                 echo "Kubeconfig file is ready"
             env:
             - name: KUBECONFIG
-              value: "{{"/etc/kube/{{.ITSkubeconfig}}"}}"
+              value: /etc/kube/config-incluster
```
MikeSpreitzer: What problem does this change solve?
Why is it correct to use a constant filename here instead of the one passed in the `ITSkubeconfig` variable?

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -41,38 +41,30 @@ spec:
...
-                # Wait for essential OCM CRDs needed by the Status Add-on
-                # First check if clusteradm has created any OCM CRDs
-                echo "Checking if OCM CRDs exist..."
+                echo "Waiting for essential OCM CRDs to become available..."
                 max_attempts=60
                 attempt=0
                 while [ $attempt -lt $max_attempts ]; do
-                  if kubectl get crd placements.cluster.open-cluster-management.io >/dev/null 2>&1; then
-                    echo "OCM CRDs found, waiting for them to be established..."
-                    kubectl wait --for=condition=Established crd/placements.cluster.open-cluster-management.io --timeout=300s
```
MikeSpreitzer, line 53: The wait for `condition=Established` is critical. It is not enough for the CRD object to exist. A CRD object is an instruction to the kube-apiservers to add a "resource" (kind of API object). They do this asynchronously after the CRD object is created. The condition "Established" signals completion of this critical asynchronous work.

core-chart/templates/postcreatehooks/install-status-addon.yaml
rishi-jat: Thanks for pointing that out, Mike. I made the change because I was seeing inconsistent behavior  some containers were using the hardcoded /etc/kube/config-incluster while others used the templated version. That mix seemed to be causing flakiness, so I thought hardcoding the path would simplify things and make it more reliable.

But after digging deeper, I now realize the filename actually depends on the ControlPlane type like host and external using kubeconfig-incluster. So hardcoding it would break those cases.

I’ll revert this and switch back to the templated path. Thanks again!

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -41,29 +41,27 @@ spec:
...
+                  local crd_name="$1"
+                  attempt=0
+                  while [ $attempt -lt $max_attempts ]; do
+                    if kubectl get crd "$crd_name" >/dev/null 2>&1; then
+                      echo "Found $crd_name, waiting for it to be established..."
+                      kubectl wait --for=condition=Established "crd/$crd_name" --timeout=300s && return 0
+                    fi
+                    attempt=$((attempt + 1))
+                    sleep 5
+                  done
```
MikeSpreitzer: Rather than this loop, wouldn't it be simpler to just do a sequence of two `kubectl wait` commands, first `--for=created` and then `--for=condition=Established` ?

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -41,29 +41,27 @@ spec:
               - sh
               - -c
               - |
-                echo "Waiting for OCM CRDs to be established..."
-                # Wait for essential OCM CRDs needed by the Status Add-on
-                # First check if clusteradm has created any OCM CRDs
-                echo "Checking if OCM CRDs exist..."
+                echo "Waiting for essential OCM CRDs to become available..."
```
MikeSpreitzer: "available" is not precise enough. The point here is waiting for them to become _established_.

core-chart/templates/postcreatehooks/install-status-addon.yaml
rishi-jat: Got it, I’ll replace the loop with two kubectl wait commands as suggested. Thanks!

core-chart/templates/postcreatehooks/install-status-addon.yaml
rishi-jat: Thanks for the clarification! I’ll update the message to say “established” instead of “available” for accuracy.

rishi-jat:
@MikeSpreitzer, changes made as suggested PTAL when you get a chance. Thanks!

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -41,29 +41,10 @@ spec:
...
-                    attempt=$((attempt+1))
-                  fi
-                done
-                if [ $attempt -eq $max_attempts ]; then
-                  echo "ERROR: OCM CRDs not found after $max_attempts attempts"
-                  exit 1
-                fi
+                echo "Waiting for essential OCM CRDs to become established..."
+                kubectl wait --for=condition=Established crd/placements.cluster.open-cluster-management.io --timeout=300s
+                kubectl wait --for=condition=Established crd/managedclustersetbindings.cluster.open-cluster-management.io --timeout=300s
```
MikeSpreitzer, line 46: This could be further simplified to eschew the use of `bash`, instead replace this one initContainer with a sequence of two that just use `kubectl`.

PR #3175, opened 2025-08-04
🌱 Refactor: Enhance K3d cluster creation wait to use kubectl wait instead of sleep for more reliability

greninja517:
## Summary
This PR enhances the `Kubeflex cluster creation with K3d` wait logic by replacing `sleep` with `kubectl wait`, improving reliability and efficiency of script. 

## Related issue(s) #3075

Fixes #

### DIFF
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -170,7 +170,7 @@ if [ "$k8s_platform" == "kind" ]; then
     curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/scripts/create-kind-cluster-with-SSL-passthrough.sh | bash -s -- --name kubeflex --nosetcontext
 else
     k3d cluster create -p "9443:443@loadbalancer" --k3s-arg "--disable=traefik@server:*" kubeflex
-    sleep 15
+    kubectl wait --for=condition=Ready node --all --timeout=600s #Ensure both API server and nodes are ready
     helm install ingress-nginx ingress-nginx --set "controller.extraArgs.enable-ssl-passthrough=" --repo https://kubernetes.github.io/ingress-nginx --version 4.12.1 --namespace ingress-nginx --create-namespace --timeout 24h
 fi
 echo -e "\033[33m✔\033[0m Completed KubeFlex cluster with SSL Passthrough"


scripts/create-kubestellar-demo-env.sh
```diff
@@ -170,7 +170,8 @@ if [ "$k8s_platform" == "kind" ]; then
     curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/v${kubestellar_version}/scripts/create-kind-cluster-with-SSL-passthrough.sh | bash -s -- --name kubeflex --nosetcontext
 else
     k3d cluster create -p "9443:443@loadbalancer" --k3s-arg "--disable=traefik@server:*" kubeflex
-    sleep 15
+    kubectl config use-context k3d-kubeflex
```
MikeSpreitzer: As far as I know, the `k3d cluster create` command is guaranteed to have done this. If not, then the `helm install` command below has been wrong.

scripts/create-kubestellar-demo-env.sh
```diff
@@ -170,7 +170,8 @@ if [ "$k8s_platform" == "kind" ]; then
...
+    kubectl wait --for=condition=Ready node --all --timeout=600s #Ensure both API server and nodes are ready
```
MikeSpreitzer, line 173: Why do you think that node readiness is required before instantiating the Helm chart?

scripts/create-kubestellar-demo-env.sh
greninja517: `k3d create` returns after initializing the cluster components. It doesn't wait for API server readiness and nodes readiness.

scripts/create-kubestellar-demo-env.sh
greninja517: Because it ensures both API server and node readiness before helm starts to install workloads on the cluster. We can even wait for all components readiness in `kube-system` namespace for more robustness.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: But it _does_ synchronously change the current kubeconfig context.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: I doubt that node readiness is required in order for the Helm chart to be instantiated. Helm chart instantiation requires interacting with an apiserver, but does not require any service by a node.

OTOH, `sleep 15` is even less connected to anything that matters. So I am OK with this change.

scripts/create-kubestellar-demo-env.sh
greninja517: Yes, but still `kubectl config use-context ...` is idempotent. So, it is safe to run and also it will provide better clarity and readability. But, we can remove it if not necessary.

scripts/create-kubestellar-demo-env.sh
greninja517: Yes, node readiness is not required for HELM chart installation, but needed when the actual workloads are being created. Checking node readiness provides both api server and node check, so I think it is better to include instead of sleep.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: I would prefer to remove it. Having unnecessary commands increases maintenance burden for no gain. And it even causes a little bit of pain: newbies reading the script might get the idea that there is a positive need for the command in question.

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: Overall, the Kubernetes approach to things is eventual consistency, which does not require strict sequencing. Creating API objects in the apiservers does not require nodes to be ready to run containers. We should be adding only waits that are necessary.

greninja517: Okay, I understand. So, should I close this PR ? I don't think this PR does any gain.

MikeSpreitzer: I like the change that we are discussing in this comment thread. The current code does `sleep 15`, which is worse than waiting on something specific.

scripts/create-kubestellar-demo-env.sh
greninja517: ok

scripts/create-kubestellar-demo-env.sh
greninja517: Am I supposed to make any other changes ?

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

scripts/create-kubestellar-demo-env.sh
MikeSpreitzer: No, it's good now.

PR #3203, opened 2025-08-08
✨ Add prowjob to test the kubestellar-demo-env script

### LINKED ISSUE #3176: feature: test scripts/create-kubestellar-demo-env.sh in CI
### Feature Description

CI should test whether `scripts/create-kubestellar-demo-env.sh` succeeds and produces a system that correctly executes a simple test.

### Proposed Solution

Create a Prow job that tests whether `scripts/create-kubestellar-demo-env.sh` succeeds and produces a system that correctly executes a simple test (perhaps test/e2e bash without its own setup).

### Want to contribute?

### Additional Context

#3163 

Unless something changes, making CI test the demo environment creation script would make it impossible to create the next release following regular operations (see #3164). #3177

greninja517:
## Summary
This PR creates a new `prowjob` that automatically tests the `/scripts/create-kubestellar-demo-env.sh` script for both platform `kind|k3d` on every pull request that change this `/scripts/create-kubestellar-demo-env.sh` script. This will ensure that any changes made to `demo-env` script won't break it.
## Related issue(s)

Fixes #3176

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -52,3 +52,37 @@ presubmits:
             requests:
               memory: 4Gi
               cpu: 2
+
+  - name: pull-kubestellar-test-demo-env-kind
+    always_run: false
+    optional: true
+    run_if_changed: "^scripts/create-kubestellar-demo-env\\.sh$"
+    decorate: true
+    clone_uri: "https://github.com/kubestellar/kubestellar"
+    spec:
+      containers:
+        - image: ghcr.io/kubestellar/infra/build:1.24.2-1
+          command:
+            - ./test/e2e/ci/test-demo-env.sh
+            - kind
+          resources:
+            requests:
+              memory: 8Gi
+              cpu: 4
+
+  - name: pull-kubestellar-test-demo-env-k3d
+    always_run: false
+    optional: true
+    run_if_changed: "^scripts/create-kubestellar-demo-env\\.sh$"
+    decorate: true
+    clone_uri: "https://github.com/kubestellar/kubestellar"
+    spec:
+      containers:
+        - image: ghcr.io/kubestellar/infra/build:1.24.2-1
+          command:
+            - ./test/e2e/ci/test-demo-env.sh
+            - k3d
+          resources:
+            requests:
+              memory: 8Gi
+              cpu: 4
--- a/Makefile
+++ b/Makefile
@@ -352,5 +352,4 @@ build: bin-dir require-jq require-go require-git verify-go-versions  ## Build th
 bin-dir:
 	mkdir -p bin
 
-
 include Makefile.venv
--- a/test/e2e/ci/test-demo-env.sh
+++ b/test/e2e/ci/test-demo-env.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+# Copyright 2024 The KubeStellar Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e # exit on error
+set -x # for debugging
+
+platform=${1:-kind}
+
+# Validate platform
+if [[ "$platform" != "kind" && "$platform" != "k3d" ]]; then
+    echo "ERROR: Unsupported platform '$platform'. Use 'kind' or 'k3d'"
+    exit 1
+fi
+
+SRC_DIR="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)"
+scripts_dir="${SRC_DIR}/../../../scripts"
+bash_dir="${SRC_DIR}/../bash"
+common_srcs="${SRC_DIR}/../common"
+
+source "${common_srcs}/setup-shell.sh"
+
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -352,5 +352,8 @@ build: bin-dir require-jq require-go require-git verify-go-versions  ## Build th
 bin-dir:
 	mkdir -p bin
 
+.PHONY: test-demo-env
+test-demo-env: ## Test demo environment creation script
+	./test/ci/test-demo-env.sh
```
MikeSpreitzer: What is the point of this non-make-ish entry in the Makefile? The Prow job can invoke the script directly. A phony Makefile entry that depends on nothing and is depended on by nothing is just a script written in an exceptionally arcane and painful scripting language.

test/ci/test-demo-env.sh
MikeSpreitzer, line 1: Why isn't this in the e2e directory? Can contributors not use it?

test/ci/test-demo-env.sh
```diff
@@ -0,0 +1,75 @@
...
+    
+    # Cleanup for k3d
+    if [ "$platform" == "k3d" ]; then
+        echo "Cleaning up k3d environment..."
+        k3d cluster delete kubeflex || true
+        k3d cluster delete cluster1 || true
+        k3d cluster delete cluster2 || true
+        kubectl config delete-context cluster1 || true
+        kubectl config delete-context cluster2 || true
+    fi
```
MikeSpreitzer: Why duplicate the cleanup that is already in the demo env creation script?

test/ci/test-demo-env.sh
```diff
@@ -0,0 +1,75 @@
...
+scripts_dir="${SRC_DIR}/../../scripts"
+bash_dir="${SRC_DIR}/../e2e/bash"
+common_srcs="${SRC_DIR}/../e2e/common"
+
+source "${common_srcs}/setup-shell.sh"
+
+# Test demo environment setup script in both platforms
+platforms=("kind" "k3d")
+
+for platform in "${platforms[@]}"; do
```
MikeSpreitzer: If the platform were a parameter of this script, with two CI jobs invoking this script for the two platforms, and the job did not end with cleanup, then the job could end with useful debug dumping --- like the existing E2E testing workflows.

test/ci/test-demo-env.sh
MikeSpreitzer: And why erase the evidence that a contributor might need to debug something?

Makefile
greninja517: It was just for consistency like as other prow jobs also run `make <target>` command. I will update prowjob to directly call the script.

test/ci/test-demo-env.sh
greninja517: Of course, contributors can use it. But, the script is more concerned towards CI validation rather than being a E2E test itself. So, I have kept in separate directory. We can put a `readme` in `test/ci` for 
clarification.

MikeSpreitzer: I think that it is unnecessarily complex to introduce a new category for this test. Now that we have this test, it is a fine way for contributors to test the demo env creation script. Let's just have this test under the test/e2e/ umbrella.

test/ci/test-demo-env.sh
greninja517: Done.

.prow.yaml
```diff
@@ -52,3 +52,35 @@ presubmits:
             requests:
               memory: 4Gi
               cpu: 2
+
+  - name: pull-kubestellar-test-demo-env-kind
+    always_run: false
+    run_if_changed: "^scripts/create-kubestellar-demo-env\\.sh$"
```
MikeSpreitzer, line 59: Consider the release process. Particularly https://github.com/kubestellar/kubestellar/blob/34a7145fccaa91b295a6c622702848d2dc3592a3/docs/content/direct/release.md?plain=1#L55

If this test is required to pass, then this will create a problem for the release process (https://github.com/kubestellar/kubestellar/blob/34a7145fccaa91b295a6c622702848d2dc3592a3/docs/content/direct/release.md?plain=1#L61): the PR that makes the change will not pass this test, because the release does not exist yet.

MikeSpreitzer: See the "Optional" field in https://pkg.go.dev/sigs.k8s.io/prow/pkg/config#Presubmit and discussion in https://docs.prow.k8s.io/docs/jobs/#requiring-job-statuses .

greninja517: Okay, I will add `optional` field in the prow job. But, the pre-release tag `0.29.0-alpha.1` is available, and we have also updated the version in `demo-env` script. So, `demo-env` shouldn't fail, right ? 
Am I missing anything ?

MikeSpreitzer: The demo-env script works now. The problem is what happens after this PR merges and then another PR comes along to update the self-references in preparation for the next release, as prescribed in https://github.com/kubestellar/kubestellar/blob/34a7145fccaa91b295a6c622702848d2dc3592a3/docs/content/direct/release.md#making-a-new-kubestellar-release . To be utterly clear: where https://github.com/kubestellar/kubestellar/blob/34a7145fccaa91b295a6c622702848d2dc3592a3/docs/content/direct/release.md?plain=1#L61 says "Make a new Git commit with those changes and get it into the right branch in the shared repo", this is normally done with a PR. A PR that gets merged. Once that PR merges, the demo-env script is broken, because it now refers to the upcomging release (which does not exist yet).

greninja517: Revised the changes.

MikeSpreitzer:
This looks good; it would be better if it were squashed into a single commit.

MikeSpreitzer:
/lgtm

Holding off on `/approve` for a while, giving the author a chance to squash.

MikeSpreitzer:
Do we have any evidence that this new Prow job ever worked? It is failing now (e.g., for #3363).

PR #3205, opened 2025-08-08
🐛 Getting started icon rendering at two places, one was unnecessary

### LINKED ISSUE #3200: bug:  Getting started text is not correctly positioned
### Describe the bug

Image

Getting started text is rendering at a wrong place(at top left in the image).

_No response_

### Steps To Reproduce

Click the getting started section on https://docs.kubestellar.io/release-0.28.0/direct/get-started/

### Expected Behavior

On clicking getting started, it should be rendered similar to what clicking on Guide overview is getting rendered.

Image

### Want to contribute?

### Additional Context

_No response_

antedotee:
### Summary
de-duplicate “Getting Started” in nav by using directory index

Preview:https://antedotee.github.io/kubestellar/doc-getting-started-rendering-fix/direct/get-started/

Currently it looks like this:
image

Compared to previously it looks like this:
image

Fixes #3200


docs/mkdocs.yml
```diff
@@ -35,7 +35,7 @@ nav:
         - KubeStellar Galaxy: direct/galaxy-intro.md
       - Release-notes: direct/release-notes.md
       - Roadmap: direct/roadmap.md
-  - Getting Started: direct/get-started.md
+  - Getting Started: direct/get-started/
```
MikeSpreitzer, line 38: How is this change a "remove duplicate mount"? What sort of "mount" are we talking about? I see no removal here, I see a change in the format of a reference.

antedotee: Thanks for calling that out — “mount” was a poor word choice on my part. I meant a duplicate nav entry/label. With navigation.indexes enabled, pointing to direct/get-started/ uses the folder’s index page as the section landing, while pointing to direct/get-started.md creates a second entry. That was causing “Getting Started” to appear twice in the UI. This change switches to the directory form to de-duplicate the label; no content was removed, just the reference format. I have updated the PR description to reflect that more clearly.

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #3241, opened 2025-08-15
✨ Add .gitattributes to normalize line endings and ignore generated files in diffs

### LINKED ISSUE #3211: bug: Windows Line Ending (CRLF) vs Repo Line Ending (LF) Causing Large Commit Changes
### Describe the bug

I’ve encountered a problem that I believe many Windows-based developers will face when contributing to this repo. The repository enforces LF (line feed) endings, which is standard for Linux and macOS. However, on Windows machines, the default line ending is CRLF (carriage return + line feed).

When developers on Windows clone the repo and make changes, their editors and Git automatically convert the line endings from LF to CRLF. This causes Git to detect hundreds of changes, even if no real content changes were made, cluttering commit diffs with line-ending modifications.

This issue impacts developer productivity and makes code reviews more difficult.

_No response_

### Steps To Reproduce

any window user would face this issue 

### Expected Behavior

it shouldnt be showing so many commits . i had 1983 commits even after fetching from the main

### Want to contribute?

### Additional Context

_No response_

shivansh-source:
This pull request adds a .gitattributes file to the repository to:

Enforce consistent LF line endings across all platforms by setting * text=auto eol=lf.

Mark generated files (e.g., zz_generated.*.go and files inside /pkg/client/) with linguist-generated=true and -diff to minimize diff noise both locally and in GitHub pull requests. 

fixes #3211

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+* text=auto eol=lf
 # Ignore generated code in diffs.
-**/zz_generated.*.go linguist-generated=true
-/pkg/client/** linguist-generated=true
+**/zz_generated.*.go linguist-generated=true -diff
+/pkg/client/** linguist-generated=true -diff
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
-* text=auto eol=lf
+text=auto 
 # Ignore generated code in diffs.
 **/zz_generated.*.go linguist-generated=true -diff
 /pkg/client/** linguist-generated=true -diff
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
 text=auto 
 # Ignore generated code in diffs.
 **/zz_generated.*.go linguist-generated=true -diff
-/pkg/client/** linguist-generated=true -diff
+/pkg/generated/** linguist-generated=true -diff


.gitattributes
```diff
@@ -1,3 +1,4 @@
+* text=auto eol=lf
```
MikeSpreitzer, line 1: Do we really want to do this for all files in the repo? PNG files are not text. SVG files are XML files which are text, but that is not the most important type to give them.

shivansh-source:
@MikeSpreitzer is this good enough ?? its showing 1 pending

.gitattributes
MikeSpreitzer: Do we really want `eol=lf`? If I understand correctly (I looked at https://git-scm.com/docs/gitattributes), merely setting the `text` attribute on a file is enough to have its contents in the repo normalized to using LF as the line ending. Adding `eol=lf` means that Windows users get LF line endings in their working tree, while not adding that means that Windows uses get CRLF line endings in their working tree.

MikeSpreitzer: Do we want to also mark any files as binary?

shivansh-source: ohhh yeah ohk ohk got it . Let me check and change this i think  `*text=auto` is enough

shivansh-source: binary would be a good option for non text files  (images, archives, executables, etc.) but not good for pure text file or code files i will make the changes

shivansh-source:
@MikeSpreitzer i did all the necessary changes can u have a look at it

.gitattributes
```diff
@@ -1,3 +1,4 @@
+text=auto 
 # Ignore generated code in diffs.
-**/zz_generated.*.go linguist-generated=true
-/pkg/client/** linguist-generated=true
+**/zz_generated.*.go linguist-generated=true -diff
+/pkg/client/** linguist-generated=true -diff
```
MikeSpreitzer, line 4: I am not a fan of making Git hide changes to these files. I occasionally do things like `git status` for the specific purpose of understanding what files got changed.

.gitattributes
```diff
@@ -1,3 +1,4 @@
+text=auto 
 # Ignore generated code in diffs.
-**/zz_generated.*.go linguist-generated=true
-/pkg/client/** linguist-generated=true
+**/zz_generated.*.go linguist-generated=true -diff
```
MikeSpreitzer, line 3: I am not a fan of making Git hide changes to these files. I occasionally do things like `git status` for the specific purpose of understanding what files got changed.

.gitattributes
MikeSpreitzer: Actually, it was wrong to put the `linguist-generated=true` attribute on the one file here. It is NOT generated. Please do not add `-diff` here. If anything, delete this line.

OTOH, `/pkg/generated` DOES exclusively contain generated files. Perhaps the thing to do here is fix the pattern at the start of the line. Still, I am not a fan of `-diff`.

shivansh-source: actually i added `-diff` this because it reduces noise in pull requests and reviews.

shivansh-source: so i think we are fine without `-diff`

.gitattributes
MikeSpreitzer: I have done an experiment, and it shows that the change here does NOT make Git hide the fact that there is a difference; it changes how Git reports the difference.

```console
mspreitz@mjs13 kubestellar % git status
On branch pr3241-merge
	modified:   api/control/v1alpha1/zz_generated.deepcopy.go
	modified:   pkg/generated/clientset/versioned/clientset.go
```

```console
mspreitz@mjs13 kubestellar % git diff
diff --git a/api/control/v1alpha1/zz_generated.deepcopy.go b/api/control/v1alpha1/zz_generated.deepcopy.go
index acff1b9f2..f2782e1bd 100644
Binary files a/api/control/v1alpha1/zz_generated.deepcopy.go and b/api/control/v1alpha1/zz_generated.deepcopy.go differ
diff --git a/pkg/generated/clientset/versioned/clientset.go b/pkg/generated/clientset/versioned/clientset.go
index a3238ff82..831154975 100644
--- a/pkg/generated/clientset/versioned/clientset.go
+++ b/pkg/generated/clientset/versioned/clientset.go
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 // Code generated by client-gen. DO NOT EDIT.
+// Edit for the fun of it.
 
 package versioned
```

MikeSpreitzer: So I can accept this change.

shivansh-source: ohk so we are fine with `-diff`
want me to make any changes ?

.gitattributes
MikeSpreitzer: I read the OP and saw the part about "minimize diff noise both locally and in GitHub pull requests". I have long experience with that noise. It has not bothered me very much, I am able to close the twisties in GitHub's display of diffs pretty quickly and easily, although it has gotten tedious in the few cases where there are a lot of them. But it is indeed a bit of a pain.

After experimentally verifying that adding `-diff` does NOT make Git hide the fact that the file has a change, I am OK with adding `-diff`.

MikeSpreitzer: The remaining problem with this line is that the pattern identifies the wrong directory.

shivansh-source: > The remaining problem with this line is that the pattern identifies the wrong directory.

can u explain more so i can work on it

shivansh-source: > OTOH, `/pkg/generated` DOES exclusively contain generated files. Perhaps the thing to do here is fix the pattern at the start of the line. Still, I am not a fan of `-diff`.

i think this is what u are talking about i will have a look

MikeSpreitzer: yes, that's it.

shivansh-source:
@MikeSpreitzer did the change u wanted

Issue #3262, opened 2025-08-18
Add GitHub Actions workflow for static code analysis (golangci-lint) for OpenSSF Badge compliance


katara-Jayprakash:
OpenSSF Badge and related security scanners expect to find evidence of static code analysis (with tools like golangci-lint) directly in GitHub Actions workflows. Currently, our repository does not have such a workflow, which may affect our ability to qualify for compliance and security badges.

PR #3263, opened 2025-08-18
✨ feat: add static analysis workflow for OpenSSF

katara-Jayprakash:
The OpenSSF badge gives KubeStellar a low score because it expects static analysis to be set up with GitHub Actions and a .golangci.yml file. However, KubeStellar actually uses Prow CI with a special build container to run golangci-lint, including staticcheck, gosec, and errcheck, on every pull request. This isn’t recognized by the automated checks that only look for GitHub Actions.

Similar issues have come up in other projects, like Kubernetes, which solved this by adopting a hybrid CI approach. I took inspiration from their blogs and methods.

Related issue(s)
#3262
Fixes: OpenSSF badge not recognizing prow-based static analysis, so adding a GitHub Actions workflow.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -4,14 +4,6 @@ on:
   schedule:
     - cron: "0 2 * * 1"
 
-  push:
-    branches: [main]
-    paths:
-      - "**.go"
-      - "go.mod"
-      - "go.sum"
-      - ".golangci.yaml"
-
   workflow_dispatch:
 
 permissions:
@@ -41,7 +33,7 @@ jobs:
 
       - name: Install golangci-lint
         run: |
-          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.59.1
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.4.0
 
       - name: Run golangci-lint (Fast Mode)
         run: |


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,49 @@
+name: Static Analysis
+
+on:
```
MikeSpreitzer, line 3: We also need to be able to trigger this manually in the GitHub interface. Add `workflow_dispatch` to the triggering conditions.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,49 @@
...
+  pull_request:
+    paths:
+      - "**.go"
+      - "go.mod"
+      - "go.sum"
+      - ".golangci.yaml"
+  push:
+    branches: [main, master]
```
MikeSpreitzer: We do not use `master`.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,49 @@
...
+  pull-requests: read
+
+jobs:
+  lint:
+    name: Lint Go Code
+    runs-on: ubuntu-latest
+    continue-on-error: true
+
+    steps:
+      - uses: actions/checkout@v4
```
MikeSpreitzer: See https://docs.kubestellar.io/unreleased-development/contribution-guidelines/contributing-inc/#github-action-reference-discipline . Look at other workflows for examples.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: "1.21"
```
MikeSpreitzer: This should be at least as high as is in `go.mod`.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,49 @@
...
+          cache: true
+
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@v8
+        with:
+          version: v1.59.1
```
MikeSpreitzer: Why this version?

.github/workflows/golangci-lint.yml
katara-Jayprakash: ok gonna add this in next time

.github/workflows/golangci-lint.yml
katara-Jayprakash: gonna removed it next time

.github/workflows/golangci-lint.yml
katara-Jayprakash: update it , 
can check this out

katara-Jayprakash:
@MikeSpreitzer  let me know if i have to change something.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,59 @@
...
+          fetch-depth: 0
+
+      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      - name: Install golangci-lint
+        run: |
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.59.1
```
MikeSpreitzer: Why such an old version?

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,59 @@
...
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      - name: Install golangci-lint
```
MikeSpreitzer, line 34: Why did you stop using `golangci/golangci-lint-action` ?

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,59 @@
...
+  schedule:
+    - cron: "0 2 * * 1"
+
+  push:
+    branches: [main]
+    paths:
+      - "**.go"
+      - "go.mod"
+      - "go.sum"
+      - ".golangci.yaml"
```
MikeSpreitzer: Please remove this section, which causes triggering at the wrong times. Remember, we decided to ONLY trigger this workflow weekly or when requested. See https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#push for the meaning of triggering on "push".

.github/workflows/golangci-lint.yml
MikeSpreitzer: You can see all the releases of that repo at https://github.com/golangci/golangci-lint/releases . Why did you not specify a more recent one?

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

MikeSpreitzer:
After merging, I tested this. It failed. See https://github.com/kubestellar/kubestellar/actions/runs/17077053114/job/48421097517

PR #3276, opened 2025-08-27
🌱 Refactor `ValidateNumDeploymentReplicas` in ginkgo e2e test

waltforme:
## Summary
Looks like the function body of `ValidateNumDeploymentReplicas` is not doing what the function name suggests, thus somehow not easy to follow.

It confusingly said "if there is not exactly one Deployment, then assert the replica of the concerned Deployment to be zero".

This PR tries to make this cleaner.


MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

Issue #3280, opened 2025-08-28
doc: Need to flag all uses of em-dashes in documentation headings for replacement


KPRoche:
### Describe the issue

# The Problem

Issue #3275 notes a problem with the way Github and mkdocs render the internal (anchor) links created by headings in the markdown source.

If a heading in the source includes an em-dash, to indicate a secondary part of the heading name, for example: `## Scenario 0 - look around` Github turns the dash between "0" and "look" into a single hyphen in the anchor link, but mkdocs turns it into 3 hyphens in the link.

This can be fixed by using a colon instead of an em-dash to set off the second half of the heading, for instance `##Scenario 0: look around`
This will result in identical anchor links inside the rendered pages on both Github and the mkdocs-generated webpages.

# What is needed

This problem is especially apparent in the example-scenarios.md file in the repository, but there are an unknown number of other pages that may use the em-dash as a separator instead of a colon in headings _in the markdown source_.

We need help finding all those instances! If you find such a file, please comment in this issue identifying the file with the probem.

### Want to contribute?

- [ ] I would like to work on this issue.

sourabh94:
I ran a small pattern match for all files with lines starting with ## and containing  -  , I could find 2 more files : 

/oldocs/examples.md :- has lines like ## Scenario 1 - multi-cluster workload deployment with kubectl
/docs/content/Community/partners/argocd.md :- has lines like ### ArgoCD Scale Experiment - KubeStellar Community Demo Day

Do we need changes in these files ?

I have never contributed before, but I would like to work on this if changes are required.

shiavm006:
i would like to solve thsi pls assign it 
/assign

clubanderson:
Read contributors guide. You assign yourself to issues

Raghul-M:
I would like to work on it 
/assign @raghul-m

KPRoche:
> I ran a small pattern match for all files with lines starting with ## and containing - , I could find 2 more files :
> 
> /oldocs/examples.md :- has lines like ## Scenario 1 - multi-cluster workload deployment with kubectl /docs/content/Community/partners/argocd.md :- has lines like ### ArgoCD Scale Experiment - KubeStellar Community Demo Day
> 
> Do we need changes in these files ?
> 
> I have never contributed before, but I would like to work on this if changes are required.

In general, we don't need to worry about files that are not in the /docs subfolder tree; they are not processed by mkdocs into the website.

[...]

PR #3282, opened 2025-08-28
📖  Change headings to use colons and  headline case in example-scenarios.md

KPRoche:
## Summary
This replaces instances of em-dashes in headings in the file example-scenarios.md with colons, so that the anchor links generated by mkdocs will match those generated by GitHub.
It also changes all the headings to use "Headline Case"

Preview: https://kproche.github.io/kubestellar/doc-trycolons/direct/example-scenarios/

## Related issue(s)
#3275  #3280 
Fixes #

PR #3290, opened 2025-09-04
docs: Fix heading anchors by replacing hyphens with colons

shiavm006:
Fixes inconsistent anchor link generation between GitHub and mkdocs by replacing hyphens with colons in scenario headings.

Fixes # #3280


MikeSpreitzer:
There is already an active PR on this topic, #3282

PR #3315, opened 2025-09-11
🐛 Add safeguard for kubeflex-system ns creation

### LINKED ISSUE #3314: bug: KubeStellar core-chart may try to create the `kubeflex-system` ns even when it already exists
### Describe the bug

KubeStellar core-chart proceed to create the `ubeflex-system` ns when installing kubeflex, even when the ns already exists.

_No response_

### Steps To Reproduce

1. create a kind cluster
2. k create ns kbeflex-system
3. helm install core-chart

### Expected Behavior

The core chart should only ensure that the ns exists, therefore it should not try to create it exists already

### Want to contribute?

### Additional Context

_No response_

francostellari:
## Summary

Update KubeStellar core-chart to check for the existence of `kubeflex-system` namespace before it tries to create it when installing KubeFlex.

## Related issue(s)

Fixes #3314

### DIFF
--- a/core-chart/templates/kubeflex/namespaces.yaml
+++ b/core-chart/templates/kubeflex/namespaces.yaml
@@ -1,4 +1,4 @@
-{{- if index .Values "kubeflex-operator" "install" }}
+{{- if and (index .Values "kubeflex-operator" "install") (not (lookup "v1" "Namespace" "" "kubeflex-system")) }}
 apiVersion: v1
 kind: Namespace
 metadata:


francostellari:
@Rupam-It , can you please take a look when you have time? Thx

Rupam-It:
> @Rupam-It , can you please take a look when you have time? Thx

Hii @francostellari 

```
 helm install ks-core ./core-chart/
NAME: ks-core
LAST DEPLOYED: Thu Sep 11 10:24:39 2025
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
For your convenience you will probably want to add contexts to your
kubeconfig named after the non-host-type control planes (WDSes and
ITSes) that you just created (a host-type control plane is just an
alias for the KubeFlex hosting cluster). You can do that with the
following `kflex` commands; each creates a context and makes it the
current one. See
https://github.com/kubestellar/kubestellar/blob/0.29.0-alpha.1/docs/content/direct/core-chart.md#kubeconfig-files-and-contexts-for-control-planes
for a way to do this without using `kflex`.
Start by setting your current kubeconfig context to the one you used
when installing this chart.

kubectl config use-context $the_one_where_you_installed_this_chart
kflex ctx --set-current-for-hosting # make sure the KubeFlex CLI's hidden state is right for what the Helm chart just did


Finally, you can use `kflex ctx` to switch back to the kubeconfig
context for your KubeFlex hosting cluster.
```

This is working fine for me !

PR #3324, opened 2025-09-17
📖 ArgoCD and kubectl research: Report high level working mechanism

greninja517:
## Summary
This PR provides the intial findings on:
a. How `kubectl get ...` command determines what fields to show for both Native Resources and CRDs ?
b. What fields does ArgoCD need on different resources to perform health checks  successfully ? This will help us to identify what fields to propagate during the upsync process from WECs to WDS workloads.

If Needed, I will attach ArgoCD is working fine on `singleton` mode. It should work fine but still if needed for confirmation.

## Related issue(s)

Fixes #

### DIFF
--- a/argo_and_k-get_report.md
+++ b/argo_and_k-get_report.md
@@ -1,0 +1,143 @@
+# How Kubectl decides what fields to display while running `kubectl get ...` command ?
+
+Whenever we run `kubectl get ...` command, Kubernetes API server returns a `metav1.Table` object and kubectl does formatting stuff to display neatly on the console. 
+
+### For Native Kuberenetes Resources
+The `AddHandlers function` at https://github.com/kubernetes/kubernetes/blob/master/pkg/printers/internalversion/printers.go#L95 defines column headers for all native resource types. Column headers means the fields that we see for an object in the console while running `kubectl get ...`. 
+
+Example: For pods, headers are defined as:
+```shell
+podColumnDefinitions := []metav1.TableColumnDefinition{
+		{Name: "Name", Type: "string", Format: "name", Description: metav1.ObjectMeta{}.SwaggerDoc()["name"]},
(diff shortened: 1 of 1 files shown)


greninja517:
/cc @francostellari @Rupam-It 
@rishi-jat @gaurab-khanal

francostellari:
@greninja517 thank you for the detailed report.

The fact that Argo requires fields in `spec:` may be problematic?
Those fields will already be present in the the workload definition in the WDS? or do they need to be updated from the WECs.

Additionally if we defined a deployment workload in the WDS to be 1 replica but then bind it to 2 WECs, we were thinking to report a status of 1*2... would that upset ArgoCD because the `spec: replica: 1` and `status: ... : 2`?

greninja517:
> @greninja517 thank you for the detailed report.
> 
> The fact that Argo requires fields in `spec:` may be problematic? Those fields will already be present in the the workload definition in the WDS? or do they need to be updated from the WECs.
> 
> Additionally if we defined a deployment workload in the WDS to be 1 replica but then bind it to 2 WECs, we were thinking to report a status of 1*2... would that upset ArgoCD because the `spec: replica: 1` and `status: ... : 2`?

All the `.spec` fields are the desired state defined in WDS by user. So, they don't need to be updated from WECs.

If we look at the health check implementation for `Deployment` here at: https://github.com/argoproj/gitops-engine/blob/master/pkg/health/health_deployment.go#L44, ArgoCD will show `Healthy` status because `.status.updatedReplicas` will be more than `.spec.Replicas` even after we aggregate the status from 2 WECs. Other conditions also won't satisfy as we will be aggregating all those status fields. So, it will work.

But, I feel like meaning is somehow broken when we see `Spec.Replicas` is less than `Status.Replicas` after aggregation. I don't know, but from user's perspective, it looks kind of weird in the object manifest yaml.

PR #3330, opened 2025-09-24
✨feat(#2336): add create-only support for workload downsync

### LINKED ISSUE #2336: feature: user can mark a workload object as "create-only"

rishi-jat:
## Summary

Adds support for "create-only" mode in Binding and BindingPolicy objects, ensuring that workload objects are created if missing but not updated once present. Deletion behavior remains unaffected.

Fixes #2336

### DIFF
--- a/pkg/binding/binding.go
+++ b/pkg/binding/binding.go
@@ -140,14 +140,14 @@ func (c *Controller) updateOrCreateBinding(ctx context.Context, bdg *v1alpha1.Bi
 
 	// Set CreateOnly annotation if any DownsyncModulation in the spec has CreateOnly=true
 	effectiveCreateOnly := false
-	for _, downsync := range bdg.Spec.ClusterScopeDownsyncs {
+	for _, downsync := range bdg.Spec.Workload.ClusterScope {
 		if downsync.CreateOnly {
 			effectiveCreateOnly = true
 			break
 		}
 	}
 	if !effectiveCreateOnly {
-		for _, downsync := range bdg.Spec.NamespaceScopeDownsyncs {
+		for _, downsync := range bdg.Spec.Workload.NamespaceScope {
 			if downsync.CreateOnly {
 				effectiveCreateOnly = true
 				break


rishi-jat:
@MikeSpreitzer please review.Thanks!

MikeSpreitzer:
https://github.com/kubestellar/kubestellar/issues/2336#issuecomment-3384653725

PR #3381, opened 2025-10-07
📖fixed the icon and text of github repo present in docs

### LINKED ISSUE #3380: doc: Change the top view from github text and git icon to github icon and kubestellar/kubestellar text
### Describe the issue

Image

Currently, this is set to default and ideally be customized for better readability.

### Want to contribute?

antedotee:
Fixes #3380

Preview: https://antedotee.github.io/kubestellar/doc-github-icon-fix/

I am pasted the screenshot below:
image
Earlier:
image


antedotee:
@MikeSpreitzer PTAL.

MikeSpreitzer (commented):
This looks like an improvement, but an unathorized use of a GitHub logo: https://github.com/logos

/cc @KPRoche

antedotee:
This logo is supported out of the box by default by mkdocs material. What should I use now as using an external image would require it to be managed by us?😅 @MikeSpreitzer

MikeSpreitzer:
> out of the box by default by mkdocs material

If I understand correctly, what are referring to is its presence in https://fontawesome.com/search?q=github&o=r .

As far as I can tell, GitHub does not have a problem with people downloading and using either of their approved logos. They effectively require that, by offering the approved logos only as ZIP archives. So, in my opinion, go ahead and do that.

antedotee:
@MikeSpreitzer I have implemented the suggestions. Can you have a look?

PR #3383, opened 2025-10-08
✨ Add Helm version validation in pre-req check script

sidharthvijayakumar:
scripts was failing when installation was done with the below error.

```
Starting the process to install KubeStellar core: kind-kubeflex...
Error: unknown flag: --set-json
```
This was occurring in my local due to helm version which was below 3.10.0 where --set-json flag was not a valid flag. I have added a new variable which we could control and provide user more information why the scripts are failing.

Once my change is added if the helm version is lower it would fail much earlier and tell the user to update his helm version
```
X Helm version v3.8.0 is less than required v3.10.0
```

### DIFF
--- a/scripts/check_pre_req.sh
+++ b/scripts/check_pre_req.sh
@@ -38,6 +38,16 @@ is_installed() {
     wantver="$6"
     addlver="$7"
     exclver="$8"
+    #Checks if helm version installed is compatiable 
+    local INSTALLED_VERSION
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
You do not need to modify `is_installed`. It is already capable of doing the check needed for Helm, given the right parameter values.

clubanderson:
Closing as implemented. Helm version validation exists in `scripts/check_pre_req.sh` via the `is_installed_helm()` function, which checks for Helm v3+ and also verifies that helm can fetch public charts from ghcr.io.

PR #3385, opened 2025-10-08
✨added scroll to top feature

### LINKED ISSUE #3384: doc: Add scroll to top button
### Describe the issue

It would be better to add Scroll to Top button in the docs so that when the user have scrolled a long way down, the user can click on that to go to the top.

### Want to contribute?

antedotee:
Now, as you scroll through the content and try to go up, a scroll to top button appears which takes you directly to the top.
Preview -> https://antedotee.github.io/kubestellar/doc-scroll-to-top-feature/
Fixes #3384

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -143,6 +143,7 @@ theme:
     # - navigation.expand
     - navigation.path
     - navigation.footer
+    - navigation.top
     # - navigation.tracking
     # Enable a copy button in code blocks
     - content.code.copy


antedotee:
@MikeSpreitzer @clubanderson PTAL.

MikeSpreitzer:
I tested the preview. I started at the top of a long page, and scrolled down. As long as I scrolled down or held still, the "Back to top" button _did not_ appear. Scrolling back toward the top made the "Back to top" button appear.

antedotee:
@MikeSpreitzer Yes, scroll to top button only appears once you start scrolling up. But, I believe it is a good feature to add to the docs site.
To make it appear on a long page by default once we have gone past a certain length, we will have to use custom js to override the default behaviour. But, still if you want to have it, I can do that.

antedotee:
@MikeSpreitzer any further thoughts on this PR?

MikeSpreitzer:
Looks OK to me. But I am not a Javascript expert. I would like @KPRoche to make the decision on `/lgtm`.

KPRoche:
Ah, I see what @MikeSpreitzer means: 

The back to top only appears when you reverse your scroll and begin to scroll up.

Only at that point does the back-to-top button appear. Is that working as intended, or is it supposed to appear as soon as you are more than one screen down in the page?

antedotee:
@KPRoche It is working as intended.

KPRoche:
@MikeSpreitzer The back-to-top feature is built into the Materials theme we are using, and is documented here: https://albrittonanalytics.com/features/navigation/#navigation-footer

PR #3409, opened 2025-10-10
🌱 Simplify wait-for-managedclustersetbindings-crd-created initContainer

francostellari:
## Summary

Simplify `wait-for-managedclustersetbindings-crd-created` initContainer to only use kubectl without bash

## Related issue(s)

Fixes #


MikeSpreitzer:
I merged this without Prow, because of #3410

PR #3419, opened 2025-10-14
🌱 Remove redundant code

ngvanthanggit:
## Summary

When reading the code inside the `pkg/binding/` folder, I found a redundant function that is unused and also does not match the file's content.

After that, I tried to find other redundant functions inside the project as well, but I could find only this one, and I removed it.

## Related issue(s)

No issue

### DIFF
--- a/pkg/binding/selectors.go
+++ b/pkg/binding/selectors.go
@@ -188,10 +188,3 @@ func (c *Controller) testObject(ctx context.Context, bindingName string, objIden
 
 	return matched, mod
 }
-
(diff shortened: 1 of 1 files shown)


ngvanthanggit:
@MikeSpreitzer I found this random unused function inside a file in the pkg/binding/ folder, so I removed it.
You can take a look!

MikeSpreitzer:
The change look OK but look at the failed check.

MikeSpreitzer (approved):
/lgtm
/approve
Thanks!

PR #3420, opened 2025-10-14
🌱 Add logger inside client.go

ngvanthanggit:
## Summary

The current implementation of GetClient() directly writes to stderr.
Update the function to:

- [x] Use klog structured logging.
- [x] Keep behavior consistent with other controllers (e.g., BindingController).

## Related issue(s)

No issue

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -17,10 +17,10 @@ limitations under the License.
 package client
 
 import (
-	"fmt"
 	"os"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


Rupam-It (commented):
/Lgtm

pkg/client/client.go
```diff
@@ -30,20 +30,22 @@ import (
 )
 
 func GetClient() *client.Client {
+	logger := klog.Background().WithName("client")
```
MikeSpreitzer, line 33: It would be even better to make `GetClient` take a `ctx context.Context` argument, and get the logger from there (using `klog.FromContext`).

PR #3424, opened 2025-10-17
✨Cleanup install-status-addon PCH

francostellari:
## Summary

Cleanup `install-status-addon` PCH:
- switch from `command:` to `args:`
- remove spaces
- replace `"{{"` and `"}}"` with `'{{"` and `"}}'` to reduce misinterpretation and code syntax highlight

## Related issue(s)

Fixes #

### DIFF
--- a/core-chart/templates/postcreatehooks/install-status-addon.yaml
+++ b/core-chart/templates/postcreatehooks/install-status-addon.yaml
@@ -10,7 +10,7 @@ spec:
   - apiVersion: batch/v1
     kind: Job
     metadata:
-      name: "{{"{{.HookName}}"}}"
+      name: '{{"{{.HookName}}"}}'
     spec:
       template:
         spec:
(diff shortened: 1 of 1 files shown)


core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -30,27 +30,23 @@ spec:
             - name: kubeconfig
               mountPath: "/etc/kube"
               readOnly: true
-
```
MikeSpreitzer, line 33: FYI, I find the vertical spacing helpful.

core-chart/templates/postcreatehooks/install-status-addon.yaml
```diff
@@ -30,27 +30,23 @@ spec:
...
           - name: wait-for-placements-crd
             image: quay.io/kubestellar/kubectl:{{.Values.KUBECTL_VERSION}}
-            command:
-              - kubectl
+            args:
```
MikeSpreitzer, line 35: In my opinion, this is not an improvement. It adds a burden to the reader, to check what is the entrypoint of the container image.

core-chart/templates/postcreatehooks/install-status-addon.yaml
francostellari: We have more complicated yamls and we did not use the spacing. 
I changed it for consistency

core-chart/templates/postcreatehooks/install-status-addon.yaml
francostellari: In other PCHs including using helm containers and starting controllers we already do not use args instead of command. 
Having the args coming just after the name of the container makes it pretty readable.

PR #3428, opened 2025-10-18
✨ Make transport-controller mount WDS secret directly

francostellari:
## Summary

- Improve security by having the `transport-controller` mount the WDS kubeconfig directly from the secret
- Remove one bash initContainer
- fix quotation of `{{"{{ ... }}"}}`

I will work on a separate PR to do the more complicated mounting of the ITS kubeconfig from a secret instead of configmap

## Related issue(s)

Fixes #

### DIFF
--- a/core-chart/templates/postcreatehooks/transport-controller.yaml
+++ b/core-chart/templates/postcreatehooks/transport-controller.yaml
@@ -119,7 +119,7 @@ spec:
             - --transport-kubeconfig=/mnt/shared/transport-kubeconfig
             - --transport-qps={{.Values.transport_controller.transport_qps}}
             - --transport-burst={{.Values.transport_controller.transport_burst}}
-            - --wds-kubeconfig=/etc/kube/wds-kubeconfig
+            - --wds-kubeconfig=/etc/kube/wds/{{"{{.WDSkubeconfig}}"}}
             - --wds-name={{"{{.ControlPlaneName}}"}}
             - --wds-qps={{.Values.transport_controller.wds_qps}}
             - --wds-burst={{.Values.transport_controller.wds_burst}}
@@ -131,8 +131,7 @@ spec:
               mountPath: /mnt/shared
               readOnly: true
             - name: wds-kubeconfig-volume
-              mountPath: /etc/kube/wds-kubeconfig
-              subPath: '{{"{{.WDSkubeconfig}}"}}'
+              mountPath: /etc/kube/wds
               readOnly: true
           volumes:
           - name: shared-volume


core-chart/templates/controlplanes/wds.yaml
```diff
@@ -11,9 +11,12 @@ spec:
   postCreateHooks:
     - hookName: kubestellar-controller
       vars:
-        APIGroups: "{{ .APIGroups | default "" }}"
+        APIGroups: '{{ .APIGroups | default "" }}'
     - hookName: transport-controller
+      vars:
+        WDSSecretName: admin-kubeconfig
+        WDSkubeconfig: kubeconfig-incluster
```
MikeSpreitzer, line 18: Isn't this specific to the choice of "type" for this ControlPlane?

francostellari: @MikeSpreitzer that is generally true but not in the case of WDS, which can only be of type `k8s|host`.
Both produce a secret named `admin-kubeconfig` in the namespace of the WDS and have an `inClusterKey` named `kubeconfig-incluster`
I decide to bring out the variables in case we need to change the values in the future but it's not strictly necessary.

Here is a supporting example created with the command:

```shell
    --set-json ITSes='[{"name":"its1"}]' \
    --set-json WDSes='[{"name":"wds1"},{"name":"wds0","type":"host"}]'
```

You can see that wds1 and wds0 target the same secret name/key:

```
$ k get cp wds1 -o yaml
....
  secretRef:
    inClusterKey: kubeconfig-incluster
    key: kubeconfig
    name: admin-kubeconfig
    namespace: wds1-system.
```

```
$ k get cp wds0 -o yaml
....
  secretRef:
    inClusterKey: kubeconfig-incluster
    key: kubeconfig
    name: admin-kubeconfig
    namespace: wds0-system
```

MikeSpreitzer: #3442 says that `host` and `external` are treated differently.

```
    {{- $ITSSecretName = or (eq (index $ITSes 0).type "host") (eq (index $ITSes 0).type "external") | ternary "admin-kubeconfig" "vc-vcluster" }}
...
    {{- $ITSkubeconfig = or (eq (index $ITSes 0).type "host") (eq (index $ITSes 0).type "external") | ternary "kubeconfig-incluster" "config-incluster" }}
```

francostellari: The ITS secret yes, but not the WDS secret
```
        WDSSecretName: admin-kubeconfig
        WDSkubeconfig: kubeconfig-incluster
```

MikeSpreitzer: The behavior of `kflex create` depends on `--type` but not what use KubeStellar has for the ControlPlane, right?

francostellari: I think so, as kflex does not not know what the PCHs are doing

francostellari: For WDSes we only allow/support 2 types k8s|host and kflex generates the same secret name and in-cluster key for both, as shown above

MikeSpreitzer: I went to the source and surveyed what they do.

https://github.com/kubestellar/kubeflex/blob/v0.9.2/pkg/reconcilers/external/reconciler.go#L61

https://github.com/kubestellar/kubeflex/blob/v0.9.2/pkg/reconcilers/host/reconciler.go#L76

https://github.com/kubestellar/kubeflex/blob/v0.9.2/pkg/reconcilers/k8s/reconciler.go#L120

The above three are the regular ones. They all use

```
	AdminConfSecret                      = "admin-kubeconfig"
	KubeconfigSecretKeyDefault           = "kubeconfig"
	KubeconfigSecretKeyInCluster         = "kubeconfig-incluster"
```


The case of ocm is being deprecated, if I understand correctly.

https://github.com/kubestellar/kubeflex/blob/v0.9.2/pkg/reconcilers/vcluster/reconciler.go#L153-L154

uses

```
	VClusterKubeConfigSecret             = "vc-vcluster"
	KubeconfigSecretKeyVCluster          = "config"
	KubeconfigSecretKeyVClusterInCluster = "config-incluster"
```

https://github.com/kubestellar/kubeflex/blob/v0.9.2/pkg/reconcilers/k3s/reconciler.go#L118
https://github.com/kubestellar/kubeflex/blob/v0.9.2/pkg/reconcilers/k3s/configdata.go#L64-L69

uses

```
	KubeconfigSecretName         = "k3s-config"
	KubeconfigSecretKey          = "config"
	KubeconfigSecretKeyInCluster = "config-incluster"
```

francostellari: Ok but the bottom line is that fixing the secret name and key is correct for this PR . Not so for the ITS, which in fact uses the CP type to determine those parameters

PR #3435, opened 2025-10-27
✨ Improve resource handling in kubectl-rbac-flatten

MikeSpreitzer:
## Summary
This PR improves the handling of "resources" in `kubectl-rbac-flatten`.

- Fold API group and resource together.

- Resolve rules against actual list of resources.

- Reject RoleBindings of cluster-scoped resources (because they don't work).

## Related issue(s)

Fixes #

### DIFF
--- a/cmd/kubectl-rbac-flatten/README.md
+++ b/cmd/kubectl-rbac-flatten/README.md
@@ -9,6 +9,67 @@ The kubectl-rbac-flatten command takes all the standard arguments for
 a Kubernetes command-line tool, and so can be used as a [kubectl
 plugin](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/).
 
+## Data Model
+
+The semantics of the collection of RBAC objects in a cluster is a
+collection of _privilege grants_, each of which has the following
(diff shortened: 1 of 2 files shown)


cmd/kubectl-rbac-flatten/README.md
```diff
@@ -9,6 +9,67 @@ The kubectl-rbac-flatten command takes all the standard arguments for
...
+- **Role-ism**: the Role or ClusterRole object that defines the verb
+  and object.
+
+- **Binding**: the RoleBinding or ClusterRoleBinding ojbect that
+  associates the role-ism with the subject.
+
+## Query semantics
+
+This command can be directed to filter the reported grants to include
+only those that involve a given set of subjects, set of verbs, and/or
```
MikeSpreitzer, line 70: The restriction by subject and verb are not implemented yet. The only implemented restriction so far is by "resource", which is the most interesting (as far as I know).

MikeSpreitzer:
FYI, the dependency on square/go-jose comes from Kubernetes, which has this dependency until release 1.33.

PR #3423, opened 2025-10-16
📖docs: add Multi-WEC Aggregated Status (user guide) and design proposal

rishi-jat:
## Summary

This PR adds the Multi-WEC Aggregated Status user documentation and the
corresponding design proposal.

### Included

- New user-facing page:
  `docs/content/direct/multi-wec-aggregated-status.md`
  (added to `docs/mkdocs.yml`)

- New design proposal:
  `docs/content/direct/multi-wec-aggregated-status-proposal.md`
  (not added to mkdocs)

- User documentation describes configuration, behavior, and aggregation
  semantics.

- Proposal document covers API changes, aggregation rules, and validation
  plan.

### Notes

Documentation-only changes. No code changes.

Preview at: https://rishi-jat.github.io/kubestellar/direct/multi-wec-aggregated-status/
Related: #2809

PR #3441, opened 2025-10-31
✨ Add wantMultiReportedState field to BindingPolicy and Binding API

### LINKED ISSUE #3440: feature: Extend API types to add **wantMultiReportedState** field for Binding and BindingPolicy
### Feature Description

The goal of this issue is to reach towards our implementation for https://github.com/kubestellar/kubestellar/pull/3423.

As per above mentioned PR, we need to extend the Binding and BindingPolicy. 

### Proposed Solution

Implement the following:
1. Add **wantMultiReportedState** field to **DownsyncModulation** in API types.
2. Update the related CRDs.
3. Update **DownsyncModulation** struct and its methods.  

### Want to contribute?

### Additional Context

_No response_

gaurab-khanal:
## Summary

This PR extends BindingPolicy and Binding to add **wantMultiReportedState** field. As per the requirement from https://github.com/kubestellar/kubestellar/pull/3423 PR. 

### Changes:
- Add **wantMultiReportedState** field to **DownsyncModulation**
- Generate updated CRDs. 
- Update **DownsyncModulation** struct and its methods from **bindingpolicy-resolver.go** .

## Related issue(s)

Fixes #3440

### DIFF
--- a/api/control/v1alpha1/types.go
+++ b/api/control/v1alpha1/types.go
@@ -167,15 +167,15 @@ type DownsyncModulation struct {
 	WantSingletonReportedState bool `json:"wantSingletonReportedState,omitempty"`
 
 	// WantMultiWECReportedState, in short, indicates an expectation that the matching
-	// objects are distributed to multiple WECs and requests that .status from each
-	// WEC.
+	// objects are distributed to multiple WECs, requests that .status from each
+	// WEC and propagates aggregated .status to object in WDS.
 	// If there's no matching WEC then will update empty status to WDS.
-	// When this above condition is true, the instruction below does not apply.
-	// If object is of known KIND(includes built-in aggregation logic ) then we perform per KIND aggregation of .status from each
-	// WEC.
-	// Else we perform generic aggregation of .status from each WEC.
+	//
+	// If object is of a few particular kinds that this feature handles specially,
+	// the aggregation is done with awareness of, and consideration for, the semantics of their .status sections;
+	// for the rest, the aggregation is done by simple general-purpose rules.
 	// The basis of the aggregation logic is explained in the docs.
-	// After aggregation, we propagate the aggregated .status to the object in the WDS.
+	// NOTE: This API isn't yet implemented.
 	// +optional
 	WantMultiWECReportedState bool `json:"wantMultiWECReportedState,omitempty"`
 }
--- a/pkg/crd/files/crds.yaml
+++ b/pkg/crd/files/crds.yaml
@@ -297,15 +297,16 @@ spec:
                     wantMultiWECReportedState:
                       description: |-
                         WantMultiWECReportedState, in short, indicates an expectation that the matching
-                        objects are distributed to multiple WECs and requests that .status from each
-                        WEC.
+                        objects are distributed to multiple WECs, requests that .status from each
+                        WEC and propagates aggregated .status to object in WDS.
                         If there's no matching WEC then will update empty status to WDS.
-                        When this above condition is true, the instruction below does not apply.
-                        If object is of known KIND(includes built-in aggregation logic ) then we perform per KIND aggregation of .status from each
-                        WEC.
-                        Else we perform generic aggregation of .status from each WEC.
+
+
+                        If object is of a few particular kinds that this feature handles specially,
+                        the aggregation is done with awareness of, and consideration for, the semantics of their .status sections;
+                        for the rest, the aggregation is done by simple general-purpose rules.
                         The basis of the aggregation logic is explained in the docs.
-                        After aggregation, we propagate the aggregated .status to the object in the WDS.
+                        NOTE: This API isn't yet implemented.
                       type: boolean
                     wantSingletonReportedState:
                       description: |-
@@ -493,15 +494,16 @@ spec:
                         wantMultiWECReportedState:
                           description: |-
                             WantMultiWECReportedState, in short, indicates an expectation that the matching
-                            objects are distributed to multiple WECs and requests that .status from each
-                            WEC.
+                            objects are distributed to multiple WECs, requests that .status from each
+                            WEC and propagates aggregated .status to object in WDS.
                             If there's no matching WEC then will update empty status to WDS.
-                            When this above condition is true, the instruction below does not apply.
-                            If object is of known KIND(includes built-in aggregation logic ) then we perform per KIND aggregation of .status from each
-                            WEC.
-                            Else we perform generic aggregation of .status from each WEC.
+
+
+                            If object is of a few particular kinds that this feature handles specially,
+                            the aggregation is done with awareness of, and consideration for, the semantics of their .status sections;
+                            for the rest, the aggregation is done by simple general-purpose rules.
                             The basis of the aggregation logic is explained in the docs.
-                            After aggregation, we propagate the aggregated .status to the object in the WDS.
+                            NOTE: This API isn't yet implemented.
                           type: boolean
                         wantSingletonReportedState:
                           description: |-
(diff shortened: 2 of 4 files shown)


gaurab-khanal:
@francostellari @Rupam-It @MikeSpreitzer @rishi-jat 
Please review.

francostellari:
It looks good to me, @Rupam-It  do you have any concern?

api/control/v1alpha1/types.go
```diff
@@ -165,6 +165,17 @@ type DownsyncModulation struct {
...
+	// WantMultiReportedState, in short, indicates an expectation that the matching
+	// objects are distributed to multiple WECs and requests that .status from each
+	// WEC.
+	// If object is of known KIND(includes built-in aggregation logic ) then we perform per KIND aggregation of .status from each
+	// WEC.
+	// Else we perform generic aggregation of .status from each WEC.
+	// The basis of the aggregation logic is explained in the docs.
+	// After aggregation, we propagate the aggregated .status to the object in the WDS.
+	// +optional
+	WantMultiReportedState bool `json:"wantMultiReportedState,omitempty"`
```
Rupam-It: If we can make  the variable name `WantMultiWecReportedState` like this so that it can justify the function by it's name!

pkg/crd/files/crds.yaml
```diff
@@ -561,6 +583,17 @@ spec:
...
                           type: string
+                        wantMultiReportedState:
+                          description: |-
+                            WantMultiReportedState, in short, indicates an expectation that the matching
+                            objects are distributed to multiple WECs and requests that .status from each
+                            WEC.
+                            If object is of known KIND(includes built-in aggregation logic ) then we perform per KIND aggregation of .status from each
+                            WEC.
+                            Else we perform generic aggregation of .status from each WEC.
+                            The basis of the aggregation logic is explained in the docs.
```
Rupam-It, line 603: you can specify the exact file in docs that so that it helpful user to check the generic and other aggregation!

gaurab-khanal: The doc file isnt merged yet. I will update it once it is merged.

api/control/v1alpha1/types.go
gaurab-khanal: sure will update it

gaurab-khanal:
@Rupam-It  Please check made required changes.

api/control/v1alpha1/types.go
```diff
@@ -165,6 +165,17 @@ type DownsyncModulation struct {
 	// in the WDS that was propagated there from a WEC by KubeStellar.
 	// +optional
 	WantSingletonReportedState bool `json:"wantSingletonReportedState,omitempty"`
+
+	// WantMultiWECReportedState, in short, indicates an expectation that the matching
```
MikeSpreitzer, line 169: So I can not use it if I am not sure whether there will be 1 or more WECs?

gaurab-khanal: @MikeSpreitzer  You can still use it. If there's no WEC then we get empty status in WDS.

MikeSpreitzer:
I would like to make a KubeStellar release consolidating the work done so far before starting on a new API.

gaurab-khanal:
> I would like to make a KubeStellar release consolidating the work done so far before starting on a new API.

@MikeSpreitzer  Sure. Please speed up the process for the release.

api/control/v1alpha1/types.go
MikeSpreitzer: Let me clarify my question. The comment is suggesting that a user should not set WantSingletonReportedState=true if the user is unsure whether the number of WEC might be 1. I doubt that is what you mean.

api/control/v1alpha1/types.go
```diff
@@ -165,6 +165,17 @@ type DownsyncModulation struct {
...
+	// objects are distributed to multiple WECs and requests that .status from each
+	// WEC.
+	// If object is of known KIND(includes built-in aggregation logic ) then we perform per KIND aggregation of .status from each
```
MikeSpreitzer: (A) If I understand correctly, what is known or unknown is not the "kind" but rather the particular field.

(B) "perform ... aggregation of .status from each WEC" is true whether the field is known or not, so these words do not belong in the known-vs-unknown breakdown.

gaurab-khanal: what I mean to say is, known KIND here represents the builtin health check available in Argocd.
If argocd supports the KIND then it is known KIND and we perform status aggregation based on per KIND aggregator to ensure more accuracy.

MikeSpreitzer: Regarding (A): oh, you're right. I misread it. The existing words are OK in this regard.

(B) still remains. Not a critical problem.

gaurab-khanal: I couldn't fully get your query B. Please explain more.

api/control/v1alpha1/types.go
```diff
@@ -165,6 +165,19 @@ type DownsyncModulation struct {
 	// in the WDS that was propagated there from a WEC by KubeStellar.
 	// +optional
 	WantSingletonReportedState bool `json:"wantSingletonReportedState,omitempty"`
+
+	// WantMultiWECReportedState, in short, indicates an expectation that the matching
+	// objects are distributed to multiple WECs and requests that .status from each
+	// WEC.
+	// If there's no matching WEC then will update empty status to WDS.
+	// When this above condition is true, the instruction below does not apply.
+	// If object is of known KIND(includes built-in aggregation logic ) then we perform per KIND aggregation of .status from each
```
MikeSpreitzer: The object kind does not have aggregation built in. The built-in aggregation is built in _here_, as described in the rest of the sentence. So the parenthetic remark "(includes built-in aggregation logic )" is misleading. Perhaps a clearer way to say what you mean might be something like the following.

> For objects of a few particular kinds that this feature handles specially, the aggregation is done with awareness of, and consideration for, the semantics of their `.status` sections; for the rest, the aggregation is done by simple general-purpose rules.

MikeSpreitzer:
BTW, the only blocker here is that this API is not implemented yet --- so the comments in the API must clearly say that. The other work on comments can proceed in follow-on PRs.

api/control/v1alpha1/types.go
MikeSpreitzer: (B) See how the words "perform ... aggregation of .status from each WEC" appear in both of the two cases here (known and generic)? My point here is that these words are not something that differs between the two cases, so should appear outside of the case-by-case statement. Say once that this feature is about aggregating status from multiple WECs, and in the case-by-case analysis you can simply speak of "the aggregation".

MikeSpreitzer:
Also, in the PR that adds the user-facing documentation, that documentation must say that the feature is not yet implemented. The PR that finishes the implementation can also update the user-facing documentation in that regard.

api/control/v1alpha1/types.go
gaurab-khanal: yes, for WantSingletonReportedState=true if user is unsure the number of WEC == 1 then WantSingletonReportedState=true doesn't work. But for this current feature WantMultiWECReportedState=true
it works even if user is unaware about number of WEC.

MikeSpreitzer: Sorry, I made a typo in my previous comment! I meant to say that the current comment implies that the user should not set WantSingletonReportedState=true if they are unsure whether the number of WECs is 1.

gaurab-khanal: WantSingletonReportedState=true only works when number of WEC == 1. Otherwise it won't work.

api/control/v1alpha1/types.go
```diff
@@ -165,6 +165,19 @@ type DownsyncModulation struct {
 	// in the WDS that was propagated there from a WEC by KubeStellar.
 	// +optional
 	WantSingletonReportedState bool `json:"wantSingletonReportedState,omitempty"`
+
+	// WantMultiWECReportedState, in short, indicates an expectation that the matching
+	// objects are distributed to multiple WECs, requests that .status from each
+	// WEC and propagates aggregated .status to object in WDS.
+	// If there's no matching WEC then will update empty status to WDS.
```
MikeSpreitzer, line 172: Since both WantSingletonReportedState and WantMultiWECReportedState affect the `.status` section of the object in the WDS, this API has to document the meaning of all four combinations of these two Boolean values.

Please review all of the comment on WantSingletonReportedState.

gaurab-khanal: @MikeSpreitzer  Regarding the behaviour of it, Please check https://github.com/kubestellar/kubestellar/pull/3444 PR. I have mentioned what happens for all cases.

PR #3442, opened 2025-10-31
✨ Mount ITS kubeconfig as a secret in transport-controller

### LINKED ISSUE #3146: feature: Modify transport-controller to accept UTS/WDS control plane name instead of kubeconfig

francostellari:
## Summary

- add Helm templating code to the WDS control plane creation to determine the ITS name and secret info when not specified
- add initContainers to wait for the ITS creating and readiness to the transport-controller PCH
- add initContainer to copy the ITS secret from the ITS namespace to the WDS namespace
- mount the ITS kubeconfig from the local ITS secret

## Related issue(s)

Fixes #3146


core-chart/templates/controlplanes/wds.yaml
```diff
@@ -1,4 +1,40 @@
...
+  {{- if and $ITSes (gt (len $ITSes) 1) }}
+    {{- fail "Too many ITSes found in the current chart!" }}
+  {{- else if and $ITSes (eq (len $ITSes) 1) }}
+    {{- $ITSName = (index $ITSes 0).name }}
+    {{- $ITSSecretName = or (eq (index $ITSes 0).type "host") (eq (index $ITSes 0).type "external") | ternary "admin-kubeconfig" "vc-vcluster" }}
+    {{- $ITSSecretNamespace = printf "%s-system" $ITSName }}
+    {{- $ITSkubeconfig = or (eq (index $ITSes 0).type "host") (eq (index $ITSes 0).type "external") | ternary "kubeconfig-incluster" "config-incluster" }}
+  {{- else }}
+    {{- $allCPs := lookup "tenancy.kflex.kubestellar.org/v1alpha1" "ControlPlane" "" "" }}
+    {{- $ITSes := dict }}
```
MikeSpreitzer: Since the entry with key "0" is the only one ever used, could we just use a variable to hold that value instead of a dict with additional but unused entries?

core-chart/templates/controlplanes/wds.yaml
MikeSpreitzer, line 1: We set out to slay the dragon of complex scripting and timing in bash scripts in containers and propose now to solve it by timing-sensitive complex scripting in the Go template language. I think, @francostellari, that you had a better idea early on: put this complexity in Go code. What if there were a `kflex write-cp-access-kubeconfig` command that included (a) picking the only ITS ControlPlane if requested, (b) waiting for the ControlPlane to exist and be ready, and (c) writing the kubecofig for in cluster or extra-cluster access to the ControlPlane to a file of a given name?

core-chart/templates/postcreatehooks/transport-controller.yaml
```diff
@@ -100,32 +110,59 @@ spec:
...
-            volumeMounts:
-            - name: config-volume
-              mountPath: /mnt/config
-            - name: shared-volume
-              mountPath: /mnt/shared
-          - name: setup-its-kubeconfig
+            command:
+              - sh
+              - -c
+              - 'kubectl get secret {{"{{.ITSSecretName}}"}} --namespace {{"{{.ITSSecretNamespace}}"}} -o yaml | sed "/ownerReferences:/,/^- /d;/namespace:/d;/creationTimestamp:/d;/uid:/d;/resourceVersion:/d;s/\(name: \).*/\1its-kubeconfig/" | kubectl apply -f -'
```
MikeSpreitzer: If I understand correctly, the task here is to copy a Secret from one namespace to another; have I got that right?

It might be simpler and safer to pick what to keep than what to discard. Can this script use `jq` or `yq`?

In the spirit of simplifying and owning the container images, it might be better to make our own `copy-secret` command.

francostellari: yes, it's copying the script
I agree it would be better to use jq/yq
However, I believe we removed jq/yq from the kubectl container to reduce the security impact.
Noe of the other init containers use it... so I was hesitant to reintroduce when another solution was possible,

core-chart/templates/controlplanes/wds.yaml
francostellari: I think in the long term, the solution of having the transport plugin access the its secret directly without a copy would be better. I think in the meanwhile this approach has some benefits
1. not mount a kubeconfig in a ConfigMap
2. remove the need for a ConfigMap script
In my view, I don't think we have a timing issue since the templating either look for an ITS being installed by the same chart or for an ITS already in the cluster, installed by some previous chart.

core-chart/templates/controlplanes/wds.yaml
francostellari: @MikeSpreitzer thank you for pointing it out, I think it was a leftover from different approach that I attempted
Changed in new commit.

core-chart/templates/controlplanes/wds.yaml
```diff
@@ -1,4 +1,40 @@
 {{- range .Values.WDSes }}
+{{- $ITSName := .ITSName }}
+{{- $ITSSecretName := "" }}
+{{- $ITSSecretNamespace := "" }}
+{{- $ITSkubeconfig := "" }}
```
MikeSpreitzer: This variable name leaves a little to be desired: it does not make it clear that it is talking specifically about the _key_ in the Secret. Maybe use "ITSSecretKey" instead?

Same comment applies to the name of the corresponding variable in the PCH.

core-chart/templates/postcreatehooks/transport-controller.yaml
```diff
@@ -100,23 +54,37 @@ spec:
...
-            command: [ "bin/sh", "-c", "sh /mnt/config/get-kubeconfig.sh '{{"{{.ITSName}}"}}' true true | base64 -d > /mnt/shared/transport-kubeconfig" ]
-            volumeMounts:
-            - name: config-volume
-              mountPath: /mnt/config
-            - name: shared-volume
-              mountPath: /mnt/shared
+            command:
+              - sh
+              - -c
+              - 'kubectl get secret {{"{{.ITSSecretName}}"}} --namespace {{"{{.ITSSecretNamespace}}"}} -o yaml | sed "/ownerReferences:/,/^- /d;/namespace:/d;/creationTimestamp:/d;/uid:/d;/resourceVersion:/d;s/\(name: \).*/\1its-kubeconfig/" | kubectl apply -f -'
```
MikeSpreitzer: Why put this information in a Secret, rather than directly in this Pod's filesystem?

core-chart/templates/controlplanes/wds.yaml
francostellari: I can change it but
1. it also currently matches the corresponding `WDSkubeconfig`
2. eventually this is the filename of the ITS kubeconfig once mounted

core-chart/templates/postcreatehooks/transport-controller.yaml
francostellari: @MikeSpreitzer can you help me understand how to do it?
Naively, the TC container is not running yet at this point

core-chart/templates/controlplanes/wds.yaml
MikeSpreitzer: 1. Two wrongs don't make a right.
2. OK, but see that's a pun --- the key becomes the filename. At this point in the code, it is a key, not yet a filename. And the variable name doesn't say that it's a filename either. "kubeconfig" could just as well refer to the data structure, which may be in a file or in memory.

core-chart/templates/postcreatehooks/transport-controller.yaml
MikeSpreitzer: The volumes of a Pod share the lifetime of the Pod, not any particular container. We could use a volume of type "emptyDir", with https://github.com/kubernetes/api/blob/v0.30.14/core/v1/types.go#L872 set to "Memory" to keep the confidential material out of long-term storage.

francostellari: I switched to use the emptyDir volume

core-chart/templates/postcreatehooks/transport-controller.yaml
```diff
@@ -14,13 +14,10 @@ spec:
...
       resources: ["secrets"]
-      verbs: ["get", "list", "watch"]
+      verbs: ["get"]
     - apiGroups: ["tenancy.kflex.kubestellar.org"]
       resources: ["controlplanes"]
-      verbs: ["get", "list", "watch"]
-    - apiGroups: ["tenancy.kflex.kubestellar.org"]
-      resources: ["controlplanes/status"]
-      verbs: ["get", "patch", "update"]
+      verbs: ["get", "watch"]
```
MikeSpreitzer, line 20: What does a `watch` but not a `list`? A k8s.io/client-go informer does both.

core-chart/templates/controlplanes/wds.yaml
MikeSpreitzer: This PR no longer adds timing-sensitive complex scripting in the Go template language.

PR #3449, opened 2025-11-03
🐛 Fix version check when only bounded range is desired

MikeSpreitzer:
## Summary
This PR fixes a bug in the pre-req version checking script. It was wrongly accepting any version that starts with "v0.11".

## Related issue(s)

Fixes #


MikeSpreitzer:
/cc @waltforme 
/cc @pdettori 
/cc @francostellari 
@katara-Jayprakash

clubanderson (approved):

PR #3458, opened 2025-11-06
🐛 Remove the bogus "always_run: false" from .prow.yaml

MikeSpreitzer:
## Summary
The documentation says this field is mutually exclusive with run_if_changed.

## Related issue(s)

Fixes #


MikeSpreitzer:
This was accomplished in #3463

PR #3463, opened 2025-11-06
✨ Update self-references for release 0.29.0

MikeSpreitzer:
## Summary
This PR is the prep for the next regular release.

Also remove the bogus "always_run: false" from
`.prow.yaml`.

Preview at https://mikespreitzer.github.io/kcp-edge-mc/doc-prep4-0290/

## Related issue(s)

Fixes #

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -54,7 +54,6 @@ presubmits:
               cpu: 2
 
   - name: pull-kubestellar-test-demo-env-kind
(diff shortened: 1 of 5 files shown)


MikeSpreitzer:
The failing tests of kubestellar-test-demo-env are expected. Those Prow jobs say

```yaml
    optional: true
```

so those failures should not block CI.

PR #3464, opened 2025-11-06
✨Upgrade Argo CD dependency to 9.1.0 in core-chart

### LINKED ISSUE #3462: Upgrade Argo CD version in ks-core Helm chart
### Description
Update the Argo CD version in core-chart/Chart.yaml to the latest compatible release and test the chart to ensure all ITS/WDS deployments work correctly.

rishi-jat:
## Summary
Updated Argo CD dependency in core-chart/Chart.yaml from 8.6.0 to 9.1.0 and updated Helm dependencies. 
Tested chart deployment on Kind cluster using sample ITS/WDS apps. Verified Argo CD installs and apps sync correctly.

Fixes #3462

### DIFF
--- a/core-chart/Chart.yaml
+++ b/core-chart/Chart.yaml
@@ -33,7 +33,7 @@ dependencies:
   repository: oci://ghcr.io/kubestellar/kubeflex/chart
   condition: kubeflex-operator.install
 - name: argo-cd
-  version: 8.3.5
+  version: 9.1.0
   repository: oci://ghcr.io/argoproj/argo-helm
   condition: argocd.install
   alias: argocd


francostellari:
@rishi-jat thank you for upgrading Argo, I think it's a good idea in preparation for the WEC work.
Can you confirm that you tested the Helm install and the new version works fine with KubeStellar WDSes?
Perhaps you can share the pod list and/or a snapshot of the UI with an example app.

MikeSpreitzer:
Yes, we need evidence that this works.

I see a note of caution at https://github.com/argoproj/argo-helm?tab=readme-ov-file#for-users-unable-to-upgrade:

image

MikeSpreitzer:
I see a breaking change noted in https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd#900

rishi-jat:
will do testing soon and share the Screen Short. Thanks!

MikeSpreitzer:
I prefer getting text reported as text rather than pictures of text (which typically have too little information because they capture only one screen's worth of text).

Also, in public work we should keep accessibility in mind. Every picture should have alt-text. For a picture of text, the alt-text would be the text. So there's little point in the picture part.

francostellari:
> I see a breaking change noted in https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd#900

@MikeSpreitzer reading about the impacted parameter, I don't think it affects our usage.

francostellari:
I did a quick test with ArgoCD Helm chart 9.1.0 and it seems to work as expected
image

PR #3468, opened 2025-11-07
✨ Add utility for fetching the kubeconfig for a ControlPlane

MikeSpreitzer:
## Summary
This PR adds a simple utility for fetching the kubeconfig for a KubeFlex ControlPlane, including waiting for the ControlPlane to exist and be ready.

## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ TRANSPORT_CMD_NAME ?= ocm-transport-controller
 IMAGE_TAG ?= $(shell git rev-parse --short HEAD)
 CONTROLLER_MANAGER_IMAGE ?= ${DOCKER_REGISTRY}/${CONTROLLER_MANAGER_CMD_NAME}:${IMAGE_TAG}
 TRANSPORT_IMAGE ?= ${DOCKER_REGISTRY}/${TRANSPORT_CMD_NAME}:${IMAGE_TAG}
+GET_KFCFG_IMAGE ?= ${DOCKER_REGISTRY}/kflex-get-kubeconfig:${IMAGE_TAG}
 INSTALL_KUBEFLEX ?= true
 
 KUBESTELLAR_CONTROLLER_MANAGER_VERBOSITY ?= 2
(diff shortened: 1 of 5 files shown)


MikeSpreitzer:
/cc @francostellari 
@Rupam-It 
/cc @pdettori

Makefile
```diff
@@ -19,6 +19,7 @@ TRANSPORT_CMD_NAME ?= ocm-transport-controller
 IMAGE_TAG ?= $(shell git rev-parse --short HEAD)
 CONTROLLER_MANAGER_IMAGE ?= ${DOCKER_REGISTRY}/${CONTROLLER_MANAGER_CMD_NAME}:${IMAGE_TAG}
 TRANSPORT_IMAGE ?= ${DOCKER_REGISTRY}/${TRANSPORT_CMD_NAME}:${IMAGE_TAG}
+GET_KFCFG_IMAGE ?= ${DOCKER_REGISTRY}/kflex-get-kubeconfig:${IMAGE_TAG}
```
francostellari, line 22: Should we match the KS and TC behavior of using a variable for the name?

cmd/kflex-get-kubeconfig/README.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+## Command line flags
+
+### Specific to this utility
+
+```console
+      --control-plane-label-selector string   label selector that identifies exactly one ControlPlane
+      --control-plane-name string             name of ControlPlane to read; mutually exclusive with --control-plane-label-selector
+      --in-cluster                            whether to extract the kubeconfig for use in the kubeflex hosting cluster (default true)
```
francostellari, line 32: does this take a T/F second argument?
`--in-cluster false`?

Otherwise, should it be default = false instead of true?

MikeSpreitzer: This uses the very common `pflag` library for Go command lines. I chose `true` as the default because I expect that to be the more common case. A user wishing the value of `false` can write `--in-cluster=false`.

Makefile
MikeSpreitzer: I was thinking of removing unneeded complexity. I am unaware of a use case for varying this component. Not a hypothetical invented for the purpose of making a point use case, I mean an actual case. In the procedures envisaged. Which are parallel to the procedures used for the other images, with which we have some experience. I have never seen a need to specify a non-default value for TRANSPORT_CMD_NAME or CONTROLLER_MANAGER_CMD_NAME.

PR #3470, opened 2025-11-08
✨  Add a query mechanism for wantMultiWECReportedState flag

### LINKED ISSUE #3469: Add query mechanism for WantMultiWECReportedState
Last PR https://github.com/kubestellar/kubestellar/pull/3441 adds WantMultiWECReportedState to the BindingPolicy and Binding API. 
Now implement a query mechanism to get the value of **WantMultiWECReportedState** similar to how **WantSingletonReportedState** is implemented.

gaurab-khanal:
## Summary
This PR builds the query mechanism to read the value of  **wantMultiWECReportedState** flag. Now according to this flag value(true or false), will implement the next feature for status collection from all WEC and its aggregation. 

## Related issue(s)

Fixes #3469

### DIFF
--- a/pkg/binding/bindingpolicy-resolver.go
+++ b/pkg/binding/bindingpolicy-resolver.go
@@ -108,7 +108,7 @@ type BindingPolicyResolver interface {
 	// the resolutions regarding singleton reported state and multiwec reported state requests for a given workload object.
 	// First and third is the `bool` indicating whether any BindingPolicy requests singleton or multiwec reported state return respectively
 	// for the given object.
-	// If those are true then the second and fourth are the set of qualified WECs bound to that object respectively,
+	// When first or third value is true then the second or fourth are the set of qualified WECs bound to that object respectively,
 	// otherwise the second or fourth value is empty.
 	// Returns: (wantSingleton, qualifiedWECsSingleton, wantMultiWEC, qualifiedWECsMulti)
 	GetReportedStateRequestForObject(util.ObjectIdentifier) (bool, sets.Set[string], bool, sets.Set[string])


gaurab-khanal:
\cc @MikeSpreitzer @francostellari @Rupam-It @rishi-jat

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -271,6 +271,20 @@ func (resolution *bindingPolicyResolution) getSingletonReportedStateRequestForOb
 	return false, false, nil
 }
 
+// getMultiWECReportedStateRequestForObject returns what this resolution requests regarding multi-wec reported state return.
```
MikeSpreitzer, line 274: Rather than add a parallel function, it would be simpler and more efficient to modify (and rename, of course) the existing `getSingletonReportedStateRequestForObject` function to return both request bits.

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -271,6 +271,20 @@ func (resolution *bindingPolicyResolution) getSingletonReportedStateRequestForOb
...
+// The first returned bool reports whether this resolution matches the given workload object ID;
```
MikeSpreitzer, line 275: > whether this resolution matches the given workload object ID

That is not accurate. This is copied from the comment on `getSingletonReportedStateRequestForObject`, and it is inaccurate there too. The first returned `bool` indicates whether the resolution has any data for that workload object. In other words, whether the BindingPolicy matched that workload object.

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -117,6 +117,14 @@ type BindingPolicyResolver interface {
 	// If the resolution doesn't exist then returns `nil`.
 	GetSingletonReportedStateRequestsForBinding(bindingPolicyKey string) []SingletonReportedStateReturnStatus
 
+	// GetMultiWECReportedStateRequestForObject returns the combined effects of all
+	// the resolutions regarding multi-wec reported state return for a given workload object.
+	// The returned `bool` indicates whether multi-wec reported state return is requested.
+	// When that is `true`, the returned `int` is the number of WECs that the resolutions
+	// collectively associate with the workload object;
+	// otherwise the returned `int` is unspecified.
+	GetMultiWECReportedStateRequestForObject(util.ObjectIdentifier) (bool, int)
```
MikeSpreitzer: It would be simpler and more efficient to extend (and rename, of course) the existing `GetSingletonReportedStateRequestsForBinding` method than to introduce a second parallel method.

pkg/status/singletonstatus.go
```diff
@@ -64,24 +64,41 @@ func (c *Controller) updateWorkStatusToObject(ctx context.Context, workStatusON
...
+	isMultiWECRequested, nWECsMulti := c.bindingPolicyResolver.GetMultiWECReportedStateRequestForObject(wObjID)
 
-	if (isMultiWECRequested || isSingletonRequested) && nWECs == 1 {
-		logger.V(4).Info("Either singleton or multiWEC status is requested and nWEC == 1", "object: ", wObjID, "nWEC", nWECs)
+	// nWECs is the count of WEC matched by clusterselectors in binding policy when wantSingletonReportedState = true
+	// nWECsMulti is the count of WEC matched by clusterselectors in binding policy when wantMultiWECReportedState = true
+	// when both wantSingletonReportedState and wantMultiWECReportedState is set to true then nWECs and nWECsMulti is equal
+
+	logger.V(4).Info("Workload object (singleton status requested)", "object", wObjID, "isSingletonRequested", isSingletonRequested, "nWECs", nWECs,
+		"Workload object (multiWEC status requested)", "object", wObjID, "isMultiWECRequested", isMultiWECRequested, "nWECsMulti", nWECsMulti)
```
MikeSpreitzer: Clearly nobody looked at the result of actually executing this statement.

There is no reason to duplicate the `, "object", wObjID`.

Look at the definition of the `Info` method on a Logger. See how it takes one message followed by a series of key-value pairs? `"Workload object (multiWEC status requested)"` here is a second message, which does not belong in the series of key-value pairs. Instead, generalize the one and only message at the start of this call.

pkg/status/singletonstatus.go
```diff
@@ -64,24 +64,41 @@ func (c *Controller) updateWorkStatusToObject(ctx context.Context, workStatusON
...
-	if isMultiWECRequested && nWECs > 1 {
-		logger.V(4).Info("multiWEC status is requested and nWEC != 1", "object: ", wObjID, "nWEC", nWECs)
-		return c.handleMultiWEC(ctx, wObjID, nWECs)
+	if isMultiWECRequested && nWECsMulti > 1 {
+		logger.V(4).Info("multiWEC status is requested and nWEC != 1", "object: ", wObjID, "nWEC", nWECsMulti)
+		return c.handleMultiWEC(ctx, wObjID, nWECsMulti)
 	}
 
-	logger.V(4).Info("None of the above condition for singleton and multiwec is true for workload object", wObjID)
+	logger.V(4).Info("None of the condition mentioned in doc for execution of handleSingleton and handleMultiWEC function is matched.", wObjID)
```
MikeSpreitzer: An explicit reference to documentation is little better than an implicit reference to code. State the condition here. Or just say something like "neither singleton nor multi-WEC reported state return applies".

gaurab-khanal:
@MikeSpreitzer  Please check.

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -104,15 +104,16 @@ type BindingPolicyResolver interface {
...
 
-	// GetSingletonReportedStateRequestForObject returns the combined effects of all
-	// the resolutions regarding singleton reported state return for a given workload object.
-	// The returned `bool` indicates whether singleton reported state return is requested.
+	// GetReportedStateRequestForObject returns the combined effects of all
+	// the resolutions regarding singleton reported state and multiwec reported state return for a given workload object.
+	// The returned `bool` indicates whether singleton or multiwec reported state return is requested.
 	// When that is `true`, the returned `sets.Set[string]` is the qualified WECs that the resolutions
 	// collectively associate with the workload object;
 	// otherwise the returned `sets.Set[string]` is empty.
```
MikeSpreitzer: This commentary is not consistent with the method's signature.

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -355,40 +356,74 @@ func (resolver *bindingPolicyResolver) ResolutionExists(bindingPolicyKey string)
...
 
-// GetSingletonReportedStateRequestForObject returns two things.
-// First is the `bool` indicating whether any BindingPolicy requests singleton reported state return
+// GetReportedStateRequestForObject returns four things.
+// First and third is the `bool` indicating whether any BindingPolicy requests singleton or multiwec reported state return respectively
 // for the given object.
-// If that is true then the second is the set of qualified WECs bound to that object,
-// otherwise the second value is empty.
-func (resolver *bindingPolicyResolver) GetSingletonReportedStateRequestForObject(objId util.ObjectIdentifier) (bool, sets.Set[string]) {
+// If that is true then the second and fourth is the set of qualified WECs bound to that object respectively,
```
MikeSpreitzer: Grammar problem here. "that" is singular, but the previous sentence talked about two things. "respectively" is clued in.

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -104,15 +104,16 @@ type BindingPolicyResolver interface {
...
 	// collectively associate with the workload object;
 	// otherwise the returned `sets.Set[string]` is empty.
-	GetSingletonReportedStateRequestForObject(util.ObjectIdentifier) (bool, sets.Set[string])
+	// Returns: (wantSingleton, qualifiedWECsSingleton, wantMultiWEC, qualifiedWECsMulti)
+	GetReportedStateRequestForObject(util.ObjectIdentifier) (bool, sets.Set[string], bool, sets.Set[string])
 
-	// GetSingletonReportedStateRequestsForBinding calls GetSingletonReportedStateRequestForObject
+	// GetSingletonReportedStateRequestsForBinding calls GetReportedStateRequestForObject
 	// for each of workload objects in the resolution if the resolution exists.
 	// If the resolution doesn't exist then returns `nil`.
```
MikeSpreitzer, line 120: This is no longer a good characterization of what this method returns.

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -402,8 +437,8 @@ func (resolver *bindingPolicyResolver) GetSingletonReportedStateRequestsForBindi
 	objIds := resolution.getWorkloadReferences()
 	ans := make([]SingletonReportedStateReturnStatus, len(objIds))
 	for idx, objId := range objIds {
-		want, qualifiedWECs := resolver.GetSingletonReportedStateRequestForObject(objId)
-		ans[idx] = SingletonReportedStateReturnStatus{objId, want, qualifiedWECs.Len()}
+		want, qualifiedWECsSingleton, _, _ := resolver.GetReportedStateRequestForObject(objId)
```
MikeSpreitzer, line 442: Let's not compute a `sets.Set` and then discard it. Can be optimized in a follow-on PR.

pkg/status/multiwec-status.go
```diff
@@ -25,10 +25,10 @@ import (
 	"github.com/kubestellar/kubestellar/pkg/util"
 )
 
-func (c *Controller) handleMultiWEC(ctx context.Context, wObjID util.ObjectIdentifier, qualifiedWECs sets.Set[string]) error {
+func (c *Controller) handleMultiWEC(ctx context.Context, wObjID util.ObjectIdentifier, qualifiedWECsMulti sets.Set[string]) error {
```
MikeSpreitzer: Since it's already in the context of "multi", I do not think it is necessary to include the word "multi" in this parameter name. Not a critical problem.

pkg/status/singletonstatus.go
```diff
@@ -63,25 +63,36 @@ func (c *Controller) updateWorkStatusToObject(ctx context.Context, workStatusON
...
-	var isMultiWECRequested bool
-	logger.V(4).Info("Workload object (multiWEC status requested)", "wObjID", wObjID, "isMultiWECRequested", isMultiWECRequested, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
+	logger.V(4).Info("Workload object reported state request", "object", wObjID, "isSingletonRequested", isSingletonRequested, "qualifiedWECsSingleton", util.K8sSet4Log(qualifiedWECsSingleton),
+		"isMultiWECRequested", isMultiWECRequested, "qualifiedWECsMulti", util.K8sSet4Log(qualifiedWECsMulti))
 
-	if (isMultiWECRequested || isSingletonRequested) && qualifiedWECs.Len() == 1 {
-		logger.V(4).Info("Either singleton or multiWEC status is requested and qualifiedWECs.Len() == 1 for object", "wObjID", wObjID, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
-		return c.handleSingleton(ctx, wObjID, qualifiedWECs)
+	if isMultiWECRequested && isSingletonRequested {
+		logger.V(4).Info("Both singleton and multiWEC reported state return are requested for the same object", wObjID)
```
MikeSpreitzer: This information already appeared in the previous log statement. Don't junk up the log with repeating it here. This comment also applies to the other redundant log statments below.

pkg/status/singletonstatus.go
```diff
@@ -63,25 +63,36 @@ func (c *Controller) updateWorkStatusToObject(ctx context.Context, workStatusON
...
-	// just declaring these variables, we will get its actual value once above function is implemented.
-	var isMultiWECRequested bool
-	logger.V(4).Info("Workload object (multiWEC status requested)", "wObjID", wObjID, "isMultiWECRequested", isMultiWECRequested, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
+	logger.V(4).Info("Workload object reported state request", "object", wObjID, "isSingletonRequested", isSingletonRequested, "qualifiedWECsSingleton", util.K8sSet4Log(qualifiedWECsSingleton),
+		"isMultiWECRequested", isMultiWECRequested, "qualifiedWECsMulti", util.K8sSet4Log(qualifiedWECsMulti))
 
-	if (isMultiWECRequested || isSingletonRequested) && qualifiedWECs.Len() == 1 {
-		logger.V(4).Info("Either singleton or multiWEC status is requested and qualifiedWECs.Len() == 1 for object", "wObjID", wObjID, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
-		return c.handleSingleton(ctx, wObjID, qualifiedWECs)
+	if isMultiWECRequested && isSingletonRequested {
```
MikeSpreitzer, line 71: The chain of `if` statements here seems more complicated than necessary. However, it is also true that the API does not yet clearly state what should happen in each case. You might want something like the following.

if `isSingletonRequested && qualifiedWECsSingleton.Len() == 1 && isMultiWECRequested && qualifiedWECsMulti.Len() > 1` then you have an interesting question to answer!; otherwise

if `isSingletonRequested && qualifiedWECsSingleton.Len() == 1 || isMultiWECRequested && qualifiedWECsMulti.Len() == 1` then `c.handleSingleton`; otherwise

if `isMultiWECRequested && qualifiedWECsMulti.Len() > 0` then `c.handleMultiWEC`

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -104,16 +104,19 @@ type BindingPolicyResolver interface {
...
-	// collectively associate with the workload object;
-	// otherwise the returned `sets.Set[string]` is empty.
-	GetSingletonReportedStateRequestForObject(util.ObjectIdentifier) (bool, sets.Set[string])
-
-	// GetSingletonReportedStateRequestsForBinding calls GetSingletonReportedStateRequestForObject
+	// GetReportedStateRequestForObject returns the combined effects of all
+	// the resolutions regarding singleton reported state and multiwec reported state requests for a given workload object.
+	// First and third is the `bool` indicating whether any BindingPolicy requests singleton or multiwec reported state return respectively
+	// for the given object.
+	// If those are true then the second and fourth are the set of qualified WECs bound to that object respectively,
```
MikeSpreitzer: "those are true" means that both of them are true. That is not what you mean to say here.

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -104,16 +104,19 @@ type BindingPolicyResolver interface {
...
-	// the resolutions regarding singleton reported state return for a given workload object.
-	// The returned `bool` indicates whether singleton reported state return is requested.
-	// When that is `true`, the returned `sets.Set[string]` is the qualified WECs that the resolutions
-	// collectively associate with the workload object;
-	// otherwise the returned `sets.Set[string]` is empty.
-	GetSingletonReportedStateRequestForObject(util.ObjectIdentifier) (bool, sets.Set[string])
-
-	// GetSingletonReportedStateRequestsForBinding calls GetSingletonReportedStateRequestForObject
+	// GetReportedStateRequestForObject returns the combined effects of all
+	// the resolutions regarding singleton reported state and multiwec reported state requests for a given workload object.
```
MikeSpreitzer, line 108: s/multiwec/multi-WEC/

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -355,40 +358,74 @@ func (resolver *bindingPolicyResolver) ResolutionExists(bindingPolicyKey string)
...
+
+	// Query multiwec
+	var multiWECRequested bool
 	for _, resolution := range resolver.bindingPolicyToResolution {
-		matches, thisRequest, thisDests := resolution.getSingletonReportedStateRequestForObject(objId)
-		if !matches {
-			continue
+		matches, thisRequest, _ := resolution.getMultiWECReportedStateRequestForObject(objId)
+		if matches && thisRequest {
+			multiWECRequested = true
```
MikeSpreitzer, line 387: Why create a second loop for this, when it could have been done in the first loop?

`resolution.getMultiWECReportedStateRequestForObject` works a mutex. Don't do that more than necessary.

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -355,40 +358,74 @@ func (resolver *bindingPolicyResolver) ResolutionExists(bindingPolicyKey string)
...
+			break
 		}
-		requested = requested || thisRequest
-		if thisRequest {
-			matchingWECs = matchingWECs.Union(thisDests)
+	}
+
+	if !singletonRequested && !multiWECRequested {
+		return false, sets.New[string](), false, sets.New[string]()
+	}
```
MikeSpreitzer, line 394: What would go wrong if you deleted this `if` statement?

gaurab-khanal: It is an early exit to avoid unnecessary loops and re-compute the same thing when no reported state is requested.

pkg/binding/bindingpolicy-resolver.go
gaurab-khanal: what do you mean here?

pkg/binding/bindingpolicy-resolver.go
gaurab-khanal: the 1st loop is only to check if the request for it exists or not.

gaurab-khanal:
@MikeSpreitzer  Please check

pkg/binding/bindingpolicy-resolver.go
MikeSpreitzer: I mean replace the text "multiwec" with "mult-WEC".

pkg/binding/bindingpolicy-resolver.go
MikeSpreitzer: And the second loop is only to determine `multiWECRequested`.

pkg/binding/bindingpolicy-resolver.go
MikeSpreitzer: The `if singletonRequested` and `if multiWECRequested` below also do this.

pkg/status/singletonstatus.go
```diff
@@ -63,25 +63,34 @@ func (c *Controller) updateWorkStatusToObject(ctx context.Context, workStatusON
...
-	logger.V(4).Info("Workload object (multiWEC status requested)", "wObjID", wObjID, "isMultiWECRequested", isMultiWECRequested, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
+	logger.V(4).Info("Workload object reported state request", "object", wObjID, "isSingletonRequested", isSingletonRequested, "qualifiedWECsSingleton", util.K8sSet4Log(qualifiedWECsSingleton),
+		"isMultiWECRequested", isMultiWECRequested, "qualifiedWECsMulti", util.K8sSet4Log(qualifiedWECsMulti))
 
-	if (isMultiWECRequested || isSingletonRequested) && qualifiedWECs.Len() == 1 {
-		logger.V(4).Info("Either singleton or multiWEC status is requested and qualifiedWECs.Len() == 1 for object", "wObjID", wObjID, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
+	if isMultiWECRequested && isSingletonRequested {
+		// if both are requested for same object then we can combine all qualified clusters(qualifiedWECsSingleton and qualifiedWECsMulti)
+		// then call handleMultiWEC function
+		// TODO: Implement combine all qualified WECs.
```
MikeSpreitzer, line 74: FYI, it will be easy: `sets.Set` has a union operator.

PR #3477, opened 2025-11-12
🐛 Test whether singleton status return respects WEC qualification

MikeSpreitzer:
## Summary
This PR adds an end-to-end test of whether the implementation of singleton status return pays attention to whether a WEC is "qualified", as the API says it should (https://github.com/kubestellar/kubestellar/blob/95a24541f3be524563658fc9536318b927be9bfc/api/control/v1alpha1/types.go#L160-L162).

The test added here will fail, because what the implementation does is wrong in this regard. I have now documented this bug as #3478

## Related issue(s)

Fixes #


MikeSpreitzer:
@rishi-jat 
@gaurab-khanal 
@francostellari

MikeSpreitzer:
Closing this because it is being included in #3481 .

PR #3481, opened 2025-11-13
🐛 Get only qualified WEC for singleton status

### LINKED ISSUE #3478: bug: singleton reported state return does not respect limitation to "qualified" WECs
### Describe the bug

The API is very precise about which WECs to count when asking whether the count is 1. See https://github.com/kubestellar/kubestellar/blob/95a24541f3be524563658fc9536318b927be9bfc/api/control/v1alpha1/types.go#L160-L162 and https://github.com/kubestellar/kubestellar/blob/95a24541f3be524563658fc9536318b927be9bfc/api/control/v1alpha1/types.go#L149-L152

The implementation currently takes no such care. It simply counts all the WECs related to the workload object by any binding. PR #3477 demonstrates this bug, adding an end-to-end test of the semantics defined in the API, and the new test fails.

_No response_

### Steps To Reproduce

See the new E2E test case in #3477

### Expected Behavior

The new test case shows the expected behavior, as does the API.

### Want to contribute?

### Additional Context

_No response_

gaurab-khanal:
## Summary
This PR helps to get only qualified WEC for singleton status. 
Introduced new method **GetQualifiedWECForSingletonReportedStateRequest** which checks for both match and thisRequest. If only both are true it returns the set of all WEC name. 
Here 
https://github.com/gaurab-khanal/kubestellar/blob/fix-singleton/pkg/status/singletonstatus.go#L117
the new check is done to ensure only qualifiedWECName object is extracted. 

## Related issue(s)

Fixes #3478

## Local test logs:
```

Ran 27 of 27 Specs in 2048.976 seconds
SUCCESS! -- 27 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

Ginkgo ran 1 suite in 34m19.225521833s
Test Suite Passed
```

### DIFF
--- a/pkg/status/singletonstatus.go
+++ b/pkg/status/singletonstatus.go
@@ -64,24 +64,24 @@ func (c *Controller) updateWorkStatusToObject(ctx context.Context, workStatusON 
 func (c *Controller) syncWorkloadObject(ctx context.Context, wObjID util.ObjectIdentifier) error {
 	logger := klog.FromContext(ctx)
 	isSingletonRequested, qualifiedWECs := c.bindingPolicyResolver.GetSingletonReportedStateRequestForObject(wObjID)
-	logger.V(4).Info("Workload object (singleton status requested)", wObjID, "isSingletonRequested", isSingletonRequested, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
+	logger.V(4).Info("Workload object (singleton status requested)", "wObjID", wObjID, "isSingletonRequested", isSingletonRequested, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
 
 	// TODO: GetMultiWECReportedStateRequestForObject is yet to be implemented.
 	// just declaring these variables, we will get its actual value once above function is implemented.
 	var isMultiWECRequested bool
-	logger.V(4).Info("Workload object (multiWEC status requested)", wObjID, "isMultiWECRequested", isMultiWECRequested, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
+	logger.V(4).Info("Workload object (multiWEC status requested)", "wObjID", wObjID, "isMultiWECRequested", isMultiWECRequested, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
 
 	if (isMultiWECRequested || isSingletonRequested) && qualifiedWECs.Len() == 1 {
-		logger.V(4).Info("Either singleton or multiWEC status is requested and qualifiedWECs.Len() == 1 for object", wObjID, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
+		logger.V(4).Info("Either singleton or multiWEC status is requested and qualifiedWECs.Len() == 1 for object", "wObjID", wObjID, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
 		return c.handleSingleton(ctx, wObjID, qualifiedWECs)
 	}
 
 	if isMultiWECRequested && qualifiedWECs.Len() > 1 {
-		logger.V(4).Info("multiWEC status is requested and qualifiedWECs.Len() != 1 for obect ", wObjID, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
+		logger.V(4).Info("multiWEC status is requested and qualifiedWECs.Len() != 1 for object", "wObjID", wObjID, "qualifiedWECs", util.K8sSet4Log(qualifiedWECs))
 		return c.handleMultiWEC(ctx, wObjID, qualifiedWECs)
 	}
 
-	logger.V(4).Info("None of the above condition for singleton and multiwec is true for workload object", wObjID)
+	logger.V(4).Info("None of the above condition for singleton and multiwec is true for workload object", "wObjID", wObjID)
 	if err := c.updateObjectStatus(ctx, wObjID, nil, c.listers); err != nil {
 		return err
 	}


gaurab-khanal:
\cc @MikeSpreitzer @francostellari @Rupam-It @rishi-jat

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -104,6 +104,9 @@ type BindingPolicyResolver interface {
 	// given bindingpolicy key.
 	ResolutionExists(bindingPolicyKey string) bool
 
+	// GetQualifiedWECForSingletonReportedStateRequest returns the set of WECs that are qualified for singleton reported state return for the given workload object.
```
MikeSpreitzer: Rather than add this new method, it would be simpler and more efficient to modify the `GetSingletonReportedStateRequestForObject` method to return the set of qualified WECs.

pkg/status/singletonstatus.go
```diff
@@ -91,17 +91,35 @@ func (c *Controller) syncWorkloadObject(ctx context.Context, wObjID util.ObjectI
...
+	// check for qualified WECs
+	qualifiedWEC := c.bindingPolicyResolver.GetQualifiedWECForSingletonReportedStateRequest(wObjID)
+	if qualifiedWEC.Len() != 1 {
+		if err := c.updateObjectStatus(ctx, wObjID, nil, c.listers); err != nil {
+			return err
+		}
+		logger.V(4).Info("Cleaned singleton status for workload object - wrong number of qualified WECs",
+			wObjID, "qualifiedWEC", qualifiedWEC)
+		return nil
+	}
```
MikeSpreitzer: This num!=1 situations is already handled elsewhere. Just change the signature of `handleSingleton` to add the set of qualified WECs as an argument (the caller will have that, after you modify `GetSingletonReportedStateRequestForObject` as I suggested above).

gaurab-khanal: Made the changes please check @MikeSpreitzer

pkg/binding/bindingpolicy-resolver.go
```diff
@@ -110,7 +110,8 @@ type BindingPolicyResolver interface {
 	// When that is `true`, the returned `int` is the number of WECs that the resolutions
 	// collectively associate with the workload object;
 	// otherwise the returned `int` is unspecified.
-	GetSingletonReportedStateRequestForObject(util.ObjectIdentifier) (bool, int)
+	// sets.Set[string] is represents the sets of qualified WEC
+	GetSingletonReportedStateRequestForObject(util.ObjectIdentifier) (bool, int, sets.Set[string])
```
MikeSpreitzer: No need to return the size of the set, now that the set itself is being returned.

pkg/status/singletonstatus.go
```diff
@@ -63,7 +63,7 @@ func (c *Controller) updateWorkStatusToObject(ctx context.Context, workStatusON
 
 func (c *Controller) syncWorkloadObject(ctx context.Context, wObjID util.ObjectIdentifier) error {
 	logger := klog.FromContext(ctx)
-	isSingletonRequested, nWECs := c.bindingPolicyResolver.GetSingletonReportedStateRequestForObject(wObjID)
+	isSingletonRequested, nWECs, qualifiedWEC := c.bindingPolicyResolver.GetSingletonReportedStateRequestForObject(wObjID)
```
MikeSpreitzer: The new variable name here should be plural rather than singular, because it refers to a set of a general size.

MikeSpreitzer:
You want to extend the E2E test suite to test this fix, right? See https://github.com/gaurab-khanal/kubestellar/pull/3

gaurab-khanal:
> You want to extend the E2E test suite to test this fix, right? See [gaurab-khanal#3](https://github.com/gaurab-khanal/kubestellar/pull/3)

I already have the updated test from your PR. But haven't add it to the stagging and committed it. Will merge it.

MikeSpreitzer:
This looks strange:

image

You don't normally merge a branch into itself. Also, the changes in the last commit are the same as the changes in the previous commit (merging https://github.com/gaurab-khanal/kubestellar/pull/3). In short, I do not think you want commit https://github.com/kubestellar/kubestellar/pull/3481/commits/43da3fc06fbefcbd7ce6594090ff052a7a0eeaec

pkg/status/multiwec-status.go
```diff
@@ -19,15 +19,16 @@ package status
...
 
 	"github.com/kubestellar/kubestellar/pkg/util"
 )
 
-func (c *Controller) handleMultiWEC(ctx context.Context, wObjID util.ObjectIdentifier, nWECs int) error {
+func (c *Controller) handleMultiWEC(ctx context.Context, wObjID util.ObjectIdentifier, qualifiedWECs sets.Set[string]) error {
 	// TODO: Implement multiwec handling logic
 	logger := klog.FromContext(ctx)
-	logger.V(4).Info("Implement multiwec handling logic", "object", wObjID, "nWECs", nWECs)
+	logger.V(4).Info("Implement multiwec handling logic", "object", wObjID, "nWECs", qualifiedWECs)
```
MikeSpreitzer: "nWECs" is no longer the right key here.

pkg/status/singletonstatus.go
```diff
@@ -63,22 +63,22 @@ func (c *Controller) updateWorkStatusToObject(ctx context.Context, workStatusON
 
 func (c *Controller) syncWorkloadObject(ctx context.Context, wObjID util.ObjectIdentifier) error {
 	logger := klog.FromContext(ctx)
-	isSingletonRequested, nWECs := c.bindingPolicyResolver.GetSingletonReportedStateRequestForObject(wObjID)
-	logger.V(4).Info("Workload object (singleton status requested)", "object", wObjID, "isSingletonRequested", isSingletonRequested, "nWECs", nWECs)
+	isSingletonRequested, qualifiedWECs := c.bindingPolicyResolver.GetSingletonReportedStateRequestForObject(wObjID)
+	logger.V(4).Info("Workload object (singleton status requested)", "object", wObjID, "isSingletonRequested", isSingletonRequested, "nWECs", qualifiedWECs)
```
MikeSpreitzer: "nWECs" does not go together with the _set_. I would make it `"qalifiedWECs", qualifiedWECs` or `"nWECs", qualifiedWECs.Len()`.

This comment applies to all such mismatches.

gaurab-khanal:
> This looks strange:
> 
> image
> You don't normally merge a branch into itself. Also, the changes in the last commit are the same as the changes in the previous commit (merging [gaurab-khanal#3](https://github.com/gaurab-khanal/kubestellar/pull/3)). In short, I do not think you want commit [43da3fc](https://github.com/kubestellar/kubestellar/commit/43da3fc06fbefcbd7ce6594090ff052a7a0eeaec)

This is what I did:
1. Merge your PR to my branch in github.
2. My local git wasn't sync with github branch.
3. Pull the changes from github branch to local git branch which leads to creation of the mentioned commit [43da3fc].

gaurab-khanal:
@MikeSpreitzer  Please re-run the tests cases. the above command isn't working. I don't think the test failed due to this code changes.

pkg/status/singletonstatus.go
```diff
@@ -63,22 +63,22 @@ func (c *Controller) updateWorkStatusToObject(ctx context.Context, workStatusON
...
 	logger := klog.FromContext(ctx)
-	isSingletonRequested, nWECs := c.bindingPolicyResolver.GetSingletonReportedStateRequestForObject(wObjID)
-	logger.V(4).Info("Workload object (singleton status requested)", "object", wObjID, "isSingletonRequested", isSingletonRequested, "nWECs", nWECs)
+	isSingletonRequested, qualifiedWECs := c.bindingPolicyResolver.GetSingletonReportedStateRequestForObject(wObjID)
+	logger.V(4).Info("Workload object (singleton status requested)", wObjID, "isSingletonRequested", isSingletonRequested, "qualifiedWECs", qualifiedWECs)
 
 	// TODO: GetMultiWECReportedStateRequestForObject is yet to be implemented.
 	// just declaring these variables, we will get its actual value once above function is implemented.
 	var isMultiWECRequested bool
-	logger.V(4).Info("Workload object (multiWEC status requested)", "object", wObjID, "isMultiWECRequested", isMultiWECRequested, "nWECsMulti", nWECs)
```
MikeSpreitzer, line 72: The deletion of `"object", ` here was unintended, right?

pkg/status/singletonstatus.go
```diff
@@ -63,22 +63,22 @@ func (c *Controller) updateWorkStatusToObject(ctx context.Context, workStatusON
...
+	logger.V(4).Info("Workload object (multiWEC status requested)", wObjID, "isMultiWECRequested", isMultiWECRequested, "qualifiedWECs", qualifiedWECs)
 
-	if (isMultiWECRequested || isSingletonRequested) && nWECs == 1 {
-		logger.V(4).Info("Either singleton or multiWEC status is requested and nWEC == 1", "object: ", wObjID, "nWEC", nWECs)
-		return c.handleSingleton(ctx, wObjID)
+	if (isMultiWECRequested || isSingletonRequested) && qualifiedWECs.Len() == 1 {
+		logger.V(4).Info("Either singleton or multiWEC status is requested and qualifiedWECs.Len() == 1 for object", wObjID, "qualifiedWECs", qualifiedWECs)
```
MikeSpreitzer: (A) You need to learn about how the Info method works. It takes one message followed by a series of key-value pairs. You don't get to omit keys.

(B) "nWEC == 1" was a fine message fragment to describe the situation at hand. You could change it to "numWECs == 1" if you want to be a little more verbose (rather than a lot more verbose).

MikeSpreitzer:
FYI, the bash-based E2E test is _not_ failing the functional test; it is failing in the attempt to scrape Prometheus metrics from the kubestellar-controller-manager after the functional test is over.

MikeSpreitzer:
Both E2E test suites are crashing the kubestellar-controller-manager at github.com/kubestellar/kubestellar/pkg/status/singletonstatus.go:67, which reads:

```go
	logger.V(4).Info("Workload object (singleton status requested)", wObjID, "isSingletonRequested", isSingletonRequested, "qualifiedWECs", qualifiedWECs)
```

The problem with log statements like that is that the logger can only log data that more or less looks like what `json.Unmarshal` into a `*any` would produce. In particular, it can not log general maps. Use https://github.com/kubestellar/kubestellar/blob/v0.29.0/pkg/util/log-wrappers.go#L59 when logging a `sets.Set`, and https://github.com/kubestellar/kubestellar/blob/v0.29.0/pkg/util/log-wrappers.go#L34 when logging a general map.

gaurab-khanal:
@MikeSpreitzer Test are working fine now.

Rupam-It (commented):
/lgtm

MikeSpreitzer:
> Pull the changes from github branch to local git branch which leads to creation of the mentioned commit [https://github.com/kubestellar/kubestellar/commit/43da3fc06fbefcbd7ce6594090ff052a7a0eeaec].

This pull created a merge commit because your local copy of the branch had a commit that was not in github when you did the merge into github.

image

Commit f4382ee was not in your fork when you merged https://github.com/gaurab-khanal/kubestellar/pull/3

This is not a critical problem.

MikeSpreitzer (approved):
/approve
Thanks!

PR #3492, opened 2025-11-16
🐛fixed the example scenario bug for openshift

### LINKED ISSUE #3465: bug: Example scenario 1 does not work when the WEC is OpenShift (OCP)
### Describe the bug

The nginx image (public.ecr.aws/nginx/nginx@sha256:707d6194e372337c708ca4a8da8bd87085b5626f60a3e2c52bc164e64ed8682b, which is what the tag `latest` resolves to at this time) says

```yaml
            "User": "",
```

which is apparently too strong a statement for ordinary workload in OpenShift. It assigns the annotation `openshift.io/scc: restricted-v2` to the nginx Pod, and also assigns non-root uid and gid.

The Pod is in CrashLoopBackOff; following is the log from a recent run.

```
bash-3.2$ kubectl --context "$wec1_context" logs -n nginx nginx-deployment-5487657745-9gbd2 -p
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: can not modify /etc/nginx/conf.d/default.conf (read-only file system?)
/docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2025/11/07 05:55:58 [warn] 1#1: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
2025/11/07 05:55:58 [emerg] 1#1: mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
nginx: [emerg] mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
```

_No response_

### Steps To Reproduce

Outlined above.

### Expected Behavior

No sign of failure

### Want to contribute?

### Additional Context

_No response_

antedotee:
I discovered that the nginx image we were using in KubeStellar tests had a configuration issue in OpenShift environments - it was designed to run as root, but OpenShift's security policies force containers to run as non-root users with restricted permissions, causing the nginx container to crash with permission denied errors when trying to create cache directories and modify config files. To fix this, I added automatic OpenShift detection in our test utilities and modified the deployment creation function to apply appropriate security contexts when running in OpenShift, specifically setting AllowPrivilegeEscalation to false, RunAsNonRoot to true, and dropping all capabilities while using the RuntimeDefault seccomp profile. This ensures nginx deployments work correctly in OpenShift's restricted security environment while maintaining compatibility with other Kubernetes clusters.

Fixes #3465


antedotee:
@MikeSpreitzer Please take a look.

Makefile
MikeSpreitzer, line 1: The changes made here to this file are unrelated to the alleged mission of this PR.

antedotee: Sorry, I will look into it and resolve it as soon as possible.

test/e2e/ginkgo/multiple_cluster_deployment_test.go
```diff
@@ -956,5 +956,71 @@ var _ = ginkgo.Describe("end to end testing", func() {
 				return nil
 			})
 		})
+
+		ginkgo.Context("OpenShift compatibility", func() {
+			ginkgo.It("creates nginx deployment compatible with OpenShift security constraints", func(ctx context.Context) {
```
MikeSpreitzer: There is nothing specific to OpenShift here. This test case is almost a subset of what is tested by `ginkgo.It("propagates deployment to the WECs while applying CustomTransform"`. This test case only adds checking of `deployment.Status.AvailableReplicas` in the WECs --- which is beside the point of this PR.

MikeSpreitzer:
This PR does not touch example scenario 1.

shivansh-source:
@MikeSpreitzer @antedotee  can u re run the tests need to check the logs after #3669 got merged

test/util/util.go
```diff
@@ -227,6 +249,17 @@ func DeleteDeployment(ctx context.Context, wds *kubernetes.Clientset, ns string,
...
 	ginkgo.GinkgoHelper()
+
+	// Check if we're in OpenShift and adjust container configuration accordingly
+	isOpenShift := IsOpenShiftCluster(ctx, wds)
+
+	var securityContext *corev1.SecurityContext
+	if isOpenShift {
+		// For OpenShift, use minimal security context that works with restricted-v2 SCC
+		// Let OpenShift's SCC handle user assignment and privilege management
+		securityContext = &corev1.SecurityContext{}
```
MikeSpreitzer: This statement has no effect. A `nil` value for `securityContext` has the same meaning as the value assigned here.

MikeSpreitzer:
@shivansh-source: For some reason I can not ask for the failed workflow job to be retried. And even if I could, it I fear that the retry would test the same merge commit as before. I think that it would be best for @antedotee to rebase onto `main` and force-push, that would trigger fresh tests of the right stuff.

MikeSpreitzer:
Ping @antedotee

shivansh-source:
@MikeSpreitzer ginkpo test pass now . after @antedotee did the rebase

clubanderson:

/lifecycle stale

clubanderson:

/lifecycle rotten

clubanderson:

/close

PR #3496, opened 2025-11-18
🐛 clusteradm 0.11 SA name compatibility Fixes #3437

### LINKED ISSUE #3437: clusteradm 0.11.0 is incompatible with 0.10.0 and earlier
For an example of the problem: install clusteradm 0.11.0 and hack `scripts/check_pre_req.sh` to accept that (e.g., by changing the upper bound from 0.11.0 to 0.12.0). Then run `test/e2e/run-test.sh`. It will fail like this:

```console
...
+ eval '(($(kubectl --context its1 get csr 2>/dev/null | grep -c Pending) >= 2))'
+++ kubectl --context its1 get csr
+++ grep -c Pending
++ (( 0 >= 2 ))
+ (( 37 > 36 ))
+ echo 'Failed to (($(kubectl --context its1 get csr 2>/dev/null | grep -c Pending) >= 2)).'
Failed to (($(kubectl --context its1 get csr 2>/dev/null | grep -c Pending) >= 2)).
+ exit 1
```

Looking in `test/e2e/common/setup-kubestellar.sh`, used by the E2E tests to setup, you will find the command

```shell
clusteradm --context its1 get token
```

And if you try it yourself, it will fail like this:

```console
mspreitz@mjs13 kubestellar % clusteradm --context its1 get token
Error: failed to get token from sa open-cluster-management/agent-registration-bootstrap: serviceaccounts "agent-registration-bootstrap" not found
```

And if you look, you will find that the expected ServiceAccount does indeed not exist, but a similarly-named one _does_.

```console
mspreitz@mjs13 kubestellar % kubectl --context its1 get sa -A | egrep NAME\|boot
NAMESPACE                     NAME                                 SECRETS   AGE
open-cluster-management       cluster-bootstrap                    0         15m
```

vedpawar2254:
## Summary

Fixed clusteradm 0.11 compatibility by updating the bootstrap ServiceAccount name detection.
Ensures scripts and E2E tests work across versions.

## Related issue(s)

Fixes #3437

Let me know if there are any changes needed, haven't worked on this stack before so might have made some mistakes.
Thanks


SakshamDutt:
@vedpawar2254 Thanks for your contribution!

MikeSpreitzer:
This does not come close to addressing the problem.

vedpawar2254:
@MikeSpreitzer can you give help me with it. What do I need to change.
Thanks

test/scale-infra/deploy_ks_core.yaml
```diff
@@ -102,41 +102,38 @@
...
+          sa_name="cluster-bootstrap"
+        elif [ "$(printf '%s\n' 0.11.0 "$clusteradm_version" | sort -V | head -n1)" = "0.11.0" ]; then
+          sa_name="cluster-bootstrap"
+        else
+          sa_name="agent-registration-bootstrap"
+        fi
+
+        echo "Using ServiceAccount: $sa_name"
+
+        clusteradm --context kscore get token --serviceaccount "$sa_name" | grep '^clusteradm join' || true
```
shivansh-source, line 124: @vedpawar2254 cna u explain me a little bit . what u wanted to do in this pr . We can work around ur pr to merge this .

shivansh-source:
can u explain me what u did overall because i was doing this checks i found no issues thst the checks fail . also i came across a bug in clusteradm i wanted to confirm that too .

shivansh-source:
also proof of sucessful testing too .

MikeSpreitzer:
This PR is all wrong. PLEASE see https://github.com/kubestellar/kubestellar/issues/3437#issuecomment-3627663976 (Ugh, I don't know why that link does not actually go to the right comment. Read them all, particularly the later ones; the one I want you to notice is the one that says the answer is #3546).

I expect that there will need to be no need to discover or adapt to a difference in ServiceAccount name, because both client-side and server-side versions of clusteradm will advance and be compatible with each other.

PR #3503, opened 2025-11-19
🐛Unify reported-state loops, remove redundant check, fix multi-WEC comments

rishi-jat:
## Summary
Follow-up for PR #3470: merge singleton/multi-WEC logic into one loop, remove
redundant early-exit, fix “multi-WEC” comments, and improve fallback log
message.

### DIFF
--- a/pkg/binding/bindingpolicy-resolver.go
+++ b/pkg/binding/bindingpolicy-resolver.go
@@ -375,19 +375,15 @@ func (resolver *bindingPolicyResolver) GetReportedStateRequestForObject(objId ut
 	// Single loop to check both singleton and multi-WEC requests and collect WECs
 	for _, resolution := range resolver.bindingPolicyToResolution {
 		singMatches, singRequest, singDests := resolution.getSingletonReportedStateRequestForObject(objId)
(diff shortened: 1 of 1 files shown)


pkg/binding/bindingpolicy-resolver.go
```diff
@@ -359,68 +359,34 @@ func (resolver *bindingPolicyResolver) ResolutionExists(bindingPolicyKey string)
...
-			singletonRequested = singletonRequested || thisRequest
-			if thisRequest {
-				singletonWECs = singletonWECs.Union(thisDests)
+	multiWECs := sets.New[string]()
+
+	// Single loop to check both singleton and multi-WEC requests and collect WECs
+	for _, resolution := range resolver.bindingPolicyToResolution {
+		singMatches, singRequest, singDests := resolution.getSingletonReportedStateRequestForObject(objId)
+		if singMatches {
+			if singRequest {
```
MikeSpreitzer: This could be simplified to `if singMatches && singRequest {`

rishi-jat: Done.

Issue #3363, opened 2025-10-03
bug: create-kubestellar-demo-env.sh fails with "ctr: content digest not found" on WSL


shivansh-gohem:
### Describe the bug

When running the `create-kubestellar-demo-env.sh` script on a fresh Ubuntu WSL 2 environment, it consistently fails when trying to load the `vcluster` image into the `kind` cluster.

### Output from KubeStellar-Snapshot.sh

I did not run the snapshot script. However, the full terminal log from the failed `create-kubestellar-demo-env.sh` run is provided below in the "Additional Context" section. It includes all the prerequisite version checks.

### Steps To Reproduce

1. Set up an environment with Windows, Docker Desktop, and Ubuntu on WSL 2.
2. In the WSL 2 terminal, run the command: `bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/main/scripts/create-kubestellar-demo-env.sh)`

### Expected Behavior

The script should complete successfully and set up the demo environment without errors.

### Want to contribute?

- [ ] I would like to work on this issue.

### Additional Context

The script should complete successfully and set up the demo environment without errors.

#####THESE ARE THE LOGS#####
 time bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/main/scripts/create-kubestellar-demo-env.sh)
Selected Kubernetes platform: kind
Checking container runtime...
Container runtime is running.
KubeStellar Version: 0.29.0-alpha.1
Checking that pre-req softwares are installed...
Checking pre-requisites for KubeStellar:
✔ KubeFlex (Kubeflex version: v0.9.1.a6b8746 2025-09-08T01:47:31Z)
  version (unstructured): Kubeflex version: v0.9.1.a6b8746 2025-09-08T01:47:31Z
     path: /usr/local/bin/kflex
✔ OCM CLI (:v0.10.1-0-g18a991c)
  version (unstructured): client                version :v0.10.1-0-g18a991c
     path: /usr/local/bin/clusteradm
✔ Helm (v3.19.0)
  version (unstructured): version.BuildInfo{Version:"v3.19.0", GitCommit:"3d8990f0836691f0229297773f3524598f46bda6", GitTreeState:"clean", GoVersion:"go1.24.7"}
     path: /usr/local/bin/helm
✔ helm can fetch public charts
✔ kubectl (v1.32.2)
  version (unstructured): Client Version: v1.32.2
     path: /usr/local/bin/kubectl
[...]

Issue #2797, opened 2025-02-13
bug: Command Output: ctr: content digest sha256:9106d83981393d7489c50e7cfa87a9d30b78444420c22931094256e55f9d0e5f: not found


gaurab-khanal:
### Describe the bug

I was trying to install kubestellar version 0.26.0-rc.1 using kind script 
```
bash <(curl -s https://raw.githubusercontent.com/kubestellar/kubestellar/refs/tags/v0.26.0-rc.1/scripts/create-kubestellar-demo-env.sh) --platform kind
```

Error I got : 
```
Selected Kubernetes platform: kind
Checking container runtime...
Container runtime is running.
KubeStellar Version: 0.26.0-rc.1
Checking that pre-req softwares are installed...
Checking pre-requisites for KubeStellar:
✔ KubeFlex (Kubeflex version: v0.8.0.2fc9be0 2025-01-28T23:57:34Z)
  version (unstructured): Kubeflex version: v0.8.0.2fc9be0 2025-01-28T23:57:34Z
     path: /usr/local/bin/kflex
✔ OCM CLI (:v0.10.1-0-g18a991c)
  version (unstructured): client		version	:v0.10.1-0-g18a991c
     path: /usr/local/bin/clusteradm
✔ Helm (v3.17.0)
  version (unstructured): version.BuildInfo{Version:"v3.17.0", GitCommit:"301108edc7ac2a8ba79e4ebf5701b0b6ce6a31e4", GitTreeState:"clean", GoVersion:"go1.23.4"}
     path: /usr/local/bin/helm
✔ kubectl (v1.32.1)
  version (unstructured): Client Version: v1.32.1
     path: /usr/local/bin/kubectl
✔ Docker (Docker version 27.5.1, build 9f9e405)
  version (unstructured): Docker version 27.5.1, build 9f9e405
     path: /usr/local/bin/docker
✔ Kind (kind v0.26.0 go1.23.4 linux/amd64)
  version (unstructured): kind v0.26.0 go1.23.4 linux/amd64
     path: /usr/local/bin/kind

Starting environment clean up...
Starting cluster clean up...
✔ Cluster space clean up has been completed

Starting context clean up...
✔ Context space clean up completed

Creating two kind clusters to serve as example WECs
✔ Cluster cluster1 was successfully created
✔ Cluster cluster2 was successfully created
Creating KubeFlex cluster with SSL Passthrough
[...]

PR #3506, opened 2025-11-20
🐛 Use direct ctr pull to fix kind load digest error (#3363)

shivansh-gohem:
This PR fixes the persistent "ctr: content digest not found" error when running `create-kubestellar-demo-env.sh` on WSL 2 with Docker Desktop, which was preventing users from setting up the demo environment (#3363 and #2797).

The fix involves:
1. Detecting the host architecture (AMD64/ARM64).
2. Bypassing the faulty `kind load docker-image` command.
3. Instead, using `docker exec ... ctr images pull --platform` to pull the image directly inside the Kind node's Containerd runtime.

This ensures the node receives a clean, single-architecture manifest, which resolves

### DIFF
--- a/scripts/create-kubestellar-demo-env.sh
+++ b/scripts/create-kubestellar-demo-env.sh
@@ -179,7 +179,6 @@ kubectl config use-context $k8s_platform-kubeflex
 
 echo -e "\nPulling container images local..."
 images=("ghcr.io/loft-sh/vcluster:0.16.4"
-        "rancher/k3s:v1.27.2-k3s1"
         "quay.io/open-cluster-management/registration-operator:v0.15.2"
         "quay.io/kubestellar/postgresql:16.0.0-debian-11-r13")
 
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
This has the disadvantage of not using the caching in the host's store of Docker images.

I wonder how the solution in this PR compares to simply removing the image preloading? What is the difference in time to run `scripts/create-kubestellar-demo-env.sh` ?

shivansh-gohem:
Hi @MikeSpreitzer,

The fix is finalized and pushed. I ran the script again, and it completed successfully, meaning the docker exec ctr pull logic handles the multi-arch digest issue and the network instability.

The PR is ready for final testing/approval. Thanks for the guidance!

shivansh-gohem:
@MikeSpreitzer Both the kind and k3d demo environment tests failed with the same infrastructure error:

Error: The script cannot continue because Docker or Podman is not running.

This happens immediately at the start of the script (container runtime check), so my changes haven't even run yet. It seems the CI runners are having trouble starting the Docker daemon today.

Could you please re-trigger the tests? The script logic itself is verified locally.

shivansh-gohem:
@MikeSpreitzer The demo-env checks are consistently failing with the infrastructure error:

Error: The script cannot continue because Docker or Podman is not running.

This happens immediately at the start of the script on the CI runner, before my changes even run. It seems the CI environment is broken.

Since I have verified the fix locally (it runs successfully and solves the original bug), could you please review the code . I cannot resolve this CI infrastructure error from my end.

PR #3515, opened 2025-11-22
✨ Aggregation status for Daemonset and Replicaset

### LINKED ISSUE #3514: Add aggregation for remaining Kinds.
# Aggregate the status for the following:

## apps
- StatefulSet
- ReplicaSet
- DaemonSet

## extensions
- Ingress

## apiregistration.k8s.io
- APIService

## networking.k8s.io
- Ingress

## core
- Service
- PersistentVolumeClaim
- Pod

## batch
- Job

## autoscaling
- HorizontalPodAutoscaler

gaurab-khanal:
## Summary
This PR aggregates status for Daemonset and Replicaset. 

Screenshot 2025-11-22 at 7 12 16 PM

## Related issue(s)

Partially-Fixes #3514

### DIFF
--- a/pkg/status/multiwec-status.go
+++ b/pkg/status/multiwec-status.go
@@ -80,7 +80,7 @@ func (c *Controller) handleMultiWEC(ctx context.Context, wObjID util.ObjectIdent
 	}
 
 	if len(statuses) == 1 {
-		if err := c.updateObjectStatus(ctx, wObjID, statuses[0], c.listers, true); err != nil {
+		if err := c.updateObjectStatus(ctx, wObjID, statuses[0], c.listers, false); err != nil {
 			return err
 		}
 		return nil


gaurab-khanal:
@Rupam-It @francostellari @MikeSpreitzer @clubanderson  Please check

Rupam-It (commented):
/LGTM

pkg/status/aggregation/daemonset.go
```diff
@@ -0,0 +1,39 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package aggregation
+
+func AggregateDaemonSetStatus(statuses []map[string]interface{}) (map[string]interface{}, error) {
+	if len(statuses) == 0 {
```
MikeSpreitzer: The caller ensures that `len(statuses)` is not zero. And, BTW, the caller handles the case of `len(statuses) == 0` differently. This case should be handled in only one place, and correctly there.

pkg/status/aggregation/daemonset.go
```diff
@@ -0,0 +1,39 @@
...
+		return nil, nil
+	}
+
+	if len(statuses) == 1 {
+		return statuses[0], nil
+	}
```
MikeSpreitzer: This code is going to be repeated for every kind of object, right? Wouldn't it be better for the caller to handle this case once and for all?

pkg/status/aggregation/replicaset.go
```diff
@@ -0,0 +1,68 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package aggregation
+
+func AggregateReplicaSetStatus(statuses []map[string]interface{}) (map[string]interface{}, error) {
```
MikeSpreitzer: It's OK to write `any` instead of `interface{}` now (https://go.dev/ref/spec#Basic_interfaces).

pkg/status/aggregation/daemonset.go
```diff
@@ -0,0 +1,39 @@
...
+	aggregatedStatus := make(map[string]interface{})
+
+	aggregatedStatus["currentNumberScheduled"] = GetMin(statuses, "currentNumberScheduled")
+	aggregatedStatus["numberMisscheduled"] = GetMax(statuses, "numberMisscheduled")
+	aggregatedStatus["desiredNumberScheduled"] = GetMin(statuses, "desiredNumberScheduled")
+	aggregatedStatus["numberReady"] = GetMin(statuses, "numberReady")
+	aggregatedStatus["observedGeneration"] = GetMin(statuses, "observedGeneration")
+	aggregatedStatus["updatedNumberScheduled"] = GetMin(statuses, "updatedNumberScheduled")
+	aggregatedStatus["numberAvailable"] = GetMin(statuses, "numberAvailable")
+
```
MikeSpreitzer, line 30: What about conditions?

pkg/status/aggregation/replicaset.go
```diff
@@ -0,0 +1,68 @@
...
+	aggregatedStatus["readyReplicas"] = GetMin(statuses, "readyReplicas")
+	aggregatedStatus["availableReplicas"] = GetMin(statuses, "availableReplicas")
+	aggregatedStatus["observedGeneration"] = GetMin(statuses, "observedGeneration")
+
+	aggregatedStatus["conditions"] = aggregateReplicaSetConditions(statuses)
+
+	return aggregatedStatus, nil
+}
+
+func aggregateReplicaSetConditions(statuses []map[string]interface{}) []interface{} {
```
MikeSpreitzer: The behavior of this function does not match the documentation in https://docs.kubestellar.io/unreleased-development/direct/multi-wec-aggregated-status/

pkg/status/aggregation/daemonset.go
gaurab-khanal: Argo don't need condition for determining health for daemonset. https://github.com/argoproj/argo-cd/blob/master/gitops-engine/pkg/health/health_daemonset.go

pkg/status/aggregation/replicaset.go
gaurab-khanal: Screenshot 2025-11-26 at 5 23 19 PM

We use the field that are required by argo for health evaluation. 
In above function it is checking if any of the WEC cluster has condition of type ReplicaFailure which is the only thing argo looks in condition for replicaset. 
https://github.com/argoproj/argo-cd/blob/master/gitops-engine/pkg/health/health_replicaset.go#L31

MikeSpreitzer:
Do not merge from `main` into your PR's branch.

https://github.com/kubestellar/kubestellar/blob/24cfe95988cbda629b3f14570be349c8884578a9/CONTRIBUTING.md?plain=1#L93

image

pkg/status/aggregation/daemonset.go
MikeSpreitzer: This is inconsistent with https://github.com/gaurab-khanal/kubestellar/blob/8ffacf8f10dddacae7100e73ba092b43ab7600b0/docs/content/direct/multi-wec-aggregated-status.md?plain=1#L42 and the later remarks in that file about Conditions.

The problems in that doc can be fixed in other PRs.

pkg/status/multiwec-status.go
```diff
@@ -79,15 +79,26 @@ func (c *Controller) handleMultiWEC(ctx context.Context, wObjID util.ObjectIdent
 		return nil
 	}
 
+	if len(statuses) == 1 {
+		if err := c.updateObjectStatus(ctx, wObjID, statuses[0], c.listers, true); err != nil {
```
MikeSpreitzer: Think carefully about what is the most useful meaning for the two labels that identify the source of the status in a workload object in a WDS. Perhaps the most useful meaning is to distinguish between "this is exactly the status from one WEC" and "this is an approximation of the status from multiple WECs".

pkg/status/aggregation/replicaset.go
```diff
@@ -0,0 +1,61 @@
...
+	aggregatedStatus["observedGeneration"] = GetMin(statuses, "observedGeneration")
+
+	aggregatedStatus["conditions"] = aggregateReplicaSetConditions(statuses)
+
+	return aggregatedStatus, nil
+}
+
+func aggregateReplicaSetConditions(statuses []map[string]any) []any {
+	// Focus on type ReplicaFailure condition which indicates issues with replica creation
+	// If any of the statuses contains it we return its condition as aggregated condition
```
MikeSpreitzer, line 36: This is not mentioned in https://github.com/gaurab-khanal/kubestellar/blob/8ffacf8f10dddacae7100e73ba092b43ab7600b0/docs/content/direct/multi-wec-aggregated-status.md

Problems in that doc can be fixed in other PRs.

pkg/status/multiwec-status.go
gaurab-khanal: Okay. I will update the label to use singleton cause this is not approximation but the exactly status from one WEC.

pkg/status/aggregation/replicaset.go
gaurab-khanal: will raise issue to fix this.

pkg/status/aggregation/replicaset.go
MikeSpreitzer: You seem to be saying that this implementation is correct and the documentation is wrong. Then fix the documentation.

pkg/status/aggregation/replicaset.go
MikeSpreitzer: That is #3535, right?

MikeSpreitzer (approved):
/lgtm
/approve

pkg/status/aggregation/replicaset.go
gaurab-khanal: Yes

PR #3517, opened 2025-11-22
🐛fix: handle wantMultiWECReportedState flag updates Notify status cont…

rishi-jat:
## Summary

 Problem

When updating an existing BindingPolicy to change the `wantMultiWECReportedState` flag, the status wasn't being updated in the WDS workload objects. The flag change was silently ignored.

**Steps to reproduce:**
1. Create a BindingPolicy without `wantMultiWECReportedState`
2. Deploy a workload → WDS status is empty 
3. Update the same BindingPolicy to set `wantMultiWECReportedState: true`
4. Expected: WDS shows aggregated status
5. Actual: WDS status remains empty 

## Root Cause

The `ensureObjectData()` function in `bindingpolicy-resolution.go` only checked for `WantSingletonReportedState` changes. When `WantMultiWECReportedState` changed, the status controller was never notified to re-sync.

## Changes

- Added check for `WantMultiWECReportedState` changes in `ensureObjectData()`
- Added check for `WantMultiWECReportedState` in `removeObjectIdentifier()`
- Now both singleton and multi-WEC flag changes trigger proper notifications

## Testing
Follow Example Scenario 1 from the docs:
1. Deploy without flags → verify empty status
2. Update BindingPolicy to add `wantMultiWECReportedState: true`
3. Verify aggregated status appears in WDS

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/binding/bindingpolicy-resolution-broker.go
+++ b/pkg/binding/bindingpolicy-resolution-broker.go
@@ -43,9 +43,9 @@ type ResolutionBroker interface {
 	// The callbacks are called the same goroutine as this call.
 	NotifyBindingPolicyCallbacks(bindingPolicyKey string)
 
-	// NotifySingletonRequestCallbacks calls all the registered callbacks for the
+	// NotifyReportedStateRequestCallbacks calls all the registered callbacks for the
 	// given policy and workload object, in the same goroutine.
-	NotifySingletonRequestCallbacks(bindingPolicyKey string, objId util.ObjectIdentifier)
+	NotifyReportedStateRequestCallbacks(bindingPolicyKey string, objId util.ObjectIdentifier)
 }
 
 type ResolutionCallbacks struct {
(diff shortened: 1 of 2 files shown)


rishi-jat:
[test-on-fix-branch.txt](https://github.com/user-attachments/files/23691911/test-on-fix-branch.txt)
[test-on-main-branch.txt](https://github.com/user-attachments/files/23691912/test-on-main-branch.txt)

pkg/binding/bindingpolicy-resolution.go
```diff
@@ -123,7 +123,13 @@ func (resolution *bindingPolicyResolution) ensureObjectData(objIdentifier util.O
...
-		if objData == nil && modulation.WantSingletonReportedState || objData != nil && objData.Modulation.WantSingletonReportedState != modulation.WantSingletonReportedState {
+		// Notify when singleton or multi-WEC status flags change
+		singletonChanged := objData == nil && modulation.WantSingletonReportedState ||
+			objData != nil && objData.Modulation.WantSingletonReportedState != modulation.WantSingletonReportedState
+		multiWECChanged := objData == nil && modulation.WantMultiWECReportedState ||
+			objData != nil && objData.Modulation.WantMultiWECReportedState != modulation.WantMultiWECReportedState
+
+		if singletonChanged || multiWECChanged {
 			klog.InfoS("Noting addition/change of object to resolution", "resolution", fmt.Sprintf("%p", resolution), "objId", objIdentifier)
 			resolution.singletonRequestChangeConsumer(objIdentifier)
```
gaurab-khanal: you are using the singletonRequestChangeConsumer for multiwec and it is working perfectly but the name of function clearly doesn't define it would work for multiwec case as well. So it would be better if you can change the function name. 
Just like how we did for https://github.com/kubestellar/kubestellar/pull/3470/files#diff-f169185bb3380044a3351e0de0900a520b150a8d31d374698738dc8c0193234aR114

gaurab-khanal (changes requested):
This is working. But it would be better if you can change the naming of function to something meaningful that describes it handles both singleton and multiwec flag changes. 
/cc @MikeSpreitzer

pkg/binding/bindingpolicy-resolution.go
rishi-jat: make sence, will do

rishi-jat: Done.

pkg/binding/bindingpolicy-resolution-broker.go
```diff
@@ -148,13 +148,13 @@ func (broker *resolutionBroker) NotifyBindingPolicyCallbacks(bindingPolicyKey st
...
+func (broker *resolutionBroker) NotifyReportedStateRequestCallbacks(bindingPolicyKey string, objId util.ObjectIdentifier) {
 	broker.Lock()
 	defer broker.Unlock()
-	klog.InfoS("Relaying singleton request callback", "broker", fmt.Sprintf("%p=%v", broker, broker), "binding", bindingPolicyKey, "objId", objId, "numCallbacks", len(broker.callbackses))
+	klog.InfoS("Relaying reported state request callback", "broker", fmt.Sprintf("%p=%v", broker, broker), "binding", bindingPolicyKey, "objId", objId, "numCallbacks", len(broker.callbackses))
 
 	for idx, callbacks := range broker.callbackses {
-		klog.InfoS("Relaying singleton request callback", "binding", bindingPolicyKey, "objId", objId, "idx", idx, "callbacks", callbacks)
+		klog.InfoS("Relaying reported state request callback", "binding", bindingPolicyKey, "objId", objId, "idx", idx, "callbacks", callbacks)
 		callbacks.SingletonReportedStateRequestChanged(bindingPolicyKey, objId)
```
MikeSpreitzer, line 158: The name of this method should change too.
Can be done in a follow-on PR.

rishi-jat:
@MikeSpreitzer The PR is approved by you but still not merged ??

PR #3519, opened 2025-11-23
🐛 Add status clearing for multiwec

### LINKED ISSUE #3518: bug: handle status clearing for multiwec
### Describe the bug

current implementation in **singletonstatus.go** only handles singletonstatus clearing when the singleton flag is set to false from true. It doesn't handles the status clearing for multiwec when flag is changed from true to false. 

_No response_

### Steps To Reproduce

1. Create a binding policy with **wantMultiWECReportedState: true**
2. Deploy the app
3. See the aggregated status in WDS.
4. Update  **wantMultiWECReportedState:** to false
5. You can still see aggregated status in WDS 

### Expected Behavior

Change in **wantMultiWECReportedState** to false from true should clear the status in WDS. 

### Additional Context

_No response_

gaurab-khanal:
## Summary
This PR handles status clearing for multiwec when the **wantMultiWECReportedState** is set to false from true. 

## Related issue(s)

Fixes #3518

### DIFF
--- a/pkg/status/workstatus.go
+++ b/pkg/status/workstatus.go
@@ -171,14 +171,14 @@ func (c *Controller) updateObjectStatus(ctx context.Context, objectIdentifier ut
 		logger.V(5).Info("Updated status of workload object", "objectIdentifier", objectIdentifier)
 	}
 
-	if haveSingleton && !wantReturn {
+	if haveSingleton && !wantSingleton {
 		err = c.handleStatusReturnLabel(ctx, unstrObj, objectIdentifier.GVR(), false, util.BindingPolicyLabelSingletonStatusKey)
 		if err != nil {
 			return err
 		}
 	}
 
-	if haveMultiWEC && !wantReturn {
+	if haveMultiWEC && !wantMultiWEC {
 		err = c.handleStatusReturnLabel(ctx, unstrObj, objectIdentifier.GVR(), false, util.BindingPolicyLabelMultiWECStatusKey)
 		if err != nil {
 			return err


pkg/status/singletonstatus.go
```diff
@@ -93,10 +93,16 @@ func (c *Controller) syncWorkloadObject(ctx context.Context, wObjID util.ObjectI
...
+	// handles status clearing for singleton
 	if err := c.updateObjectStatus(ctx, wObjID, nil, c.listers, false); err != nil {
 		return err
 	}
 
+	// handles status clearing for multiwec
+	if err := c.updateObjectStatus(ctx, wObjID, nil, c.listers, true); err != nil {
+		return err
+	}
+
```
MikeSpreitzer: No. `c.updateObjectStatus` (already called above) should ensure that _both_ labels are set correctly. When `status == nil`, that means both labels should be absent.

gaurab-khanal: Please check. I have made requested chages.

pkg/status/workstatus.go
```diff
@@ -162,10 +170,21 @@ func (c *Controller) updateObjectStatus(ctx context.Context, objectIdentifier ut
 		}
 		logger.V(5).Info("Updated status of workload object", "objectIdentifier", objectIdentifier)
 	}
-	if haveReturn && !wantReturn {
-		// Need to remove the label alleging that the workload object has returned reported state
-		return c.handleStatusReturnLabel(ctx, unstrObj, objectIdentifier.GVR(), false, labelKey)
+
+	if haveSingleton && !wantReturn {
```
MikeSpreitzer: What if `haveSingleton && !wantSingleton` (because wantMultiWEC)?

gaurab-khanal: I couldn't understand what you are asking for. Please describe it.

MikeSpreitzer: If `haveSingleton && !wantSingleton` then the label saying that the status came from a single WEC needs to be removed --- but I do not see the current version of this PR doing that.

Issue #3539, opened 2025-12-03
bug: demo env test jobs always fail


MikeSpreitzer:
### Describe the bug

The Prow presubmits named "pull-kubestellar-test-demo-env-k3d" and "pull-kubestellar-test-demo-env-kind" (defined in `.prow.yaml`, introduced in #3203) always fail. See https://prow2.kubestellar.io/job-history/s3/prow-logs/pr-logs/directory/pull-kubestellar-test-demo-env-k3d and https://prow2.kubestellar.io/job-history/s3/prow-logs/pr-logs/directory/pull-kubestellar-test-demo-env-kind .

### Output from KubeStellar-Snapshot.sh

N/A

### Steps To Reproduce

See above.

### Expected Behavior

No sign of failure.

### Additional Context

I suspect that there are special provisions that need to be made in order to use docker inside a Prow job.

This Issue is preventing successful CI testing of #3506 .

MikeSpreitzer:
I got a few clues in a Kubernets Slack conversation, see https://kubernetes.slack.com/archives/C09QZ4DQB/p1764743055675749 .

I added a comment to a `kind` Issue, see https://github.com/kubernetes-sigs/kind/issues/303#issuecomment-3607593972

MikeSpreitzer:
The Slack discussion at https://kubernetes.slack.com/archives/C09QZ4DQB/p1764743055675749 continued. I am discouraged because
- getting `kind` to be usable inside a Prow job is tricky;
- we have less Prow expertise and available manpower than the Kubernetes project;
- the security controls over Prow jobs are more lax than over GitHub Actions workflows.

For these reasons, I currently favor moving the tests of the demo env creation scripts from Prow to GitHub Actions workflows (even though I think that this is a backwards step in general).

MikeSpreitzer:
Closing this Issue because of the change to run these tests via GitHub Actions workflows instead of Prow jobs.

PR #3541, opened 2025-12-04
🐛 Test demo env creation on GitHub instead of PROW

MikeSpreitzer:
## Summary
This PR fixes bugs in `test/e2e/ci/test-demo-env.sh`, stops testing it with Prow, and starts testing it with a new GitHub Actions workflow.

This is intended to address #3539 

You can see it tested in https://github.com/MikeSpreitzer/kcp-edge-mc/actions/workflows/test-demo-env-creation-script.yml .

Something is still going wrong, I do not yet know what.

## Related issue(s)

Fixes #

### DIFF
--- a/test/e2e/ci/test-demo-env.sh
+++ b/test/e2e/ci/test-demo-env.sh
@@ -40,6 +40,24 @@ if ! "${scripts_dir}/create-kubestellar-demo-env.sh" --platform $platform; then
     exit 1
 fi
 
+# Wait for all controllers to come up
+for ctx in ${platform}-kubeflex its1; do
+    kubectl --context $ctx get deploy -A --no-headers | while read ns name rest; do
+        kubectl --context ${ctx} wait -n $ns --for condition=Available deploy/$name --timeout 200s
+    done
+done
+
+# Wait for Pod status to be shadowed from its1 to hosting cluster
+sleep 30
+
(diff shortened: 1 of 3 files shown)


MikeSpreitzer:
Oh! It is being run here too.

MikeSpreitzer:
This finally seems to be working.

/cc @francostellari 
/cc @clubanderson 
@katara-Jayprakash

test/e2e/ci/test-demo-env.sh
```diff
@@ -40,11 +40,35 @@ if ! "${scripts_dir}/create-kubestellar-demo-env.sh" --platform $platform; then
...
+# Wait for Pod status to be shadowed from its1 to hosting cluster
+sleep 30
+
+# Check that there are no Pods in trouble
+listing=$(date; kubectl --context ${platform}-kubeflex get pods -A | grep -vw Running | grep -vw Completed)
+if ! wc -l <<<"$listing" | grep -qw 2; then
+    echo "Some KubeFlex hosting cluster Pods are in trouble" >&2
+    echo "$listing" >&2
+    exit 1
+fi
```
MikeSpreitzer, line 59: Later experience shows that this test is too strict. For example, it fails in the case of a Job that gets tried twice and succeeds on the second try; the failed first try Pod still exists and appears in `listing`.

PR #3549, opened 2025-12-05
🐛 Fix handling of absence from reversemap

MikeSpreitzer:
## Summary
This PR fixes the way `hack/gha-reversemap.sh apply-reversemap` handles the case of a workflow referencing an Action that is not listed in the reversemap. Previously such an absence caused no complaint but a replacement of the Action's version with "null". The source code appeared to intend this situation to cause a fatal error, but the detection was botched. This PR makes this case provoke a warning message and no change in the version in the reference.

Remember that for checking of complete coverage by the reversemap there is `gha-reversemap.sh verify-mapusage`.

This PR also fixes `gha-reversemap.sh verify-mapusage` to conform with its help message, checking the given workflow files if any otherwise the usual suspects.

## Related issue(s)

Fixes #

### DIFF
--- a/hack/gha-reversemap.sh
+++ b/hack/gha-reversemap.sh
@@ -82,6 +82,11 @@ _loginfo() {
     echo -e "$(date -Iseconds);INFO;$1"
 }
 
+# Log a warning
+_logwarn() {
+    echo -e "$(date -Iseconds);WARNING;$1"
+}
+
 # Exit with an error
 _exit_with_error() {
     echo -e "$(date -Iseconds);ERROR;$2" >&2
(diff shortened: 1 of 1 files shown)


hack/gha-reversemap.sh
```diff
@@ -208,13 +213,11 @@ _update_action_version_infile() {
     fi
 }
 
-# Get commit sha of an action ref stored in the reverse map
+# Get commit sha of an action ref stored in the reverse map.
+# Returns `null` if not found.
 _get_sha_from_reversemap() {
     action_ref=$1
     query=$(yq ".\"$action_ref\".sha" $REVERSEMAP_FILE)
-    if [[ $? -ne 0 ]]; then
```
MikeSpreitzer, line 215: `yq` exits with status 0 even if the action_ref is not found.

PR #3584, opened 2025-12-29
✨ Qualify excluded resources using GroupResource

### LINKED ISSUE #3583: feature: Qualify excluded resources by API group to avoid CRD name collisions
### Feature Description

Excluded resources are currently matched by name only, which can conflict with user CRDs that reuse the same name under a different API group.

### Proposed Solution

Qualify exclusions using group (or GroupResource) instead of name-only matching.

### Additional Context

_No response_

xonas1101:
## Summary
This change qualifies excluded resources using GroupResource instead of
name-only matching, preventing accidental exclusion of CRDs that reuse
common resource names.

## Related issue(s)
Fixes #3583

### DIFF
--- a/pkg/binding/binding-controller.go
+++ b/pkg/binding/binding-controller.go
@@ -99,6 +99,7 @@ var excludedGroups = map[string]bool{
 // Qualified by API group to avoid collisions with user CRDs having the same resource name.
 var excludedGroupResources = map[schema.GroupResource]bool{
 	{Group: "", Resource: "events"}:                             true,
+	{Group: "events.k8s.io", Resource: "events"}:                true,
 	{Group: "", Resource: "nodes"}:                              true,
 	{Group: "storage.k8s.io", Resource: "csistoragecapacities"}: true,
 	{Group: "storage.k8s.io", Resource: "csinodes"}:             true,
@@ -423,7 +424,6 @@ func (c *Controller) run(ctx context.Context, workers int, cListers chan interfa
 				Group:    gv.Group,
 				Resource: resource.Name,
 			}
-
 			if isExcludedGroupResource(gr) {
 				continue
 			}


pkg/binding/binding-controller.go
```diff
@@ -416,9 +421,14 @@ func (c *Controller) run(ctx context.Context, workers int, cListers chan interfa
...
-				continue
-			}
+			gr := schema.GroupResource{
+		Group:    gv.Group,
+		Resource: resource.Name,
+	  }
+
+	  if isExcludedGroupResource(gr) {
+		  continue
+	  }
```
shivansh-source: indenting are not consistent look code above and below see how indenting is happening do it i the same way .

pkg/binding/crd.go
```diff
@@ -118,9 +118,15 @@ func (c *Controller) includedToWatch(r APIResource) bool {
...
-		return false
-	}
+	gr := schema.GroupResource{
+	Group:    r.groupVersion.Group,
+	Resource: r.resource.Name,
+ }
+
+ if isExcludedGroupResource(gr) {
+	 return false
+ }
```
shivansh-source: same indenting problem .

pkg/binding/binding-controller.go
```diff
@@ -95,18 +95,23 @@ var excludedGroups = map[string]bool{
...
-	"nodes":                true,
-	"csistoragecapacities": true,
-	"csinodes":             true,
-	"endpoints":            true,
-	"workstatuses":         true,
+// Resource GroupResources to exclude for watchers as they should not be delivered to other clusters.
+// Qualified by API group to avoid collisions with user CRDs having the same resource name.
+var excludedGroupResources = map[schema.GroupResource]bool{
+	{Group: "", Resource: "events"}:                             true,
+	{Group: "", Resource: "nodes"}:                              true,
```
shivansh-source, line 103: i have doubt about this actually . in kubes the core api is represented as an empty string which is fine tbh .

pkg/binding/binding-controller.go
xonas1101: @shivansh-source  sure, I will get to it.

pkg/binding/binding-controller.go
xonas1101: I might be missing something, but here’s how I’m thinking about it.

What we have done is to qualify the resources that don't need to be propagated based on their API groups rather than their names. 

For instance, if a user builds a CRD named "Pod" or "Event", it wont be excluded, which is the target behavior we want.

shivansh-source: actually i saw other repo using "v1" or "core" but idk lets see what mike says .

xonas1101: Hi, I looked into this a bit more.

In Kubernetes, the core API group is actually represented by an empty string ("").
"v1" is the version, not the group, and "core" is more of a naming convention as I found out, but internally the API machinery still treats the core group as "".

So with:

{Group: "", Resource: "events"}, we’re specifically targeting only the built-in core resources, and we won’t accidentally exclude a CRD with the same name in another API group.

If Mike prefers a different convention for readability, I’m totally fine changing it but I feel like this should be correct, cant be too sure though.

xonas1101:
Hi @clubanderson @dumb0002 all checks are passing now. Would appreciate you taking a look when you have time. Thanks!

xonas1101:
Hi @clubanderson, quick follow up on this, is it ready to be merged, or do we need some tweaks still?

pkg/binding/binding-controller.go
```diff
@@ -95,16 +95,19 @@ var excludedGroups = map[string]bool{
...
-	"events":               true,
-	"nodes":                true,
-	"csistoragecapacities": true,
-	"csinodes":             true,
-	"endpoints":            true,
-	"workstatuses":         true,
+// Resource GroupResources to exclude for watchers as they should not be delivered to other clusters.
+// Qualified by API group to avoid collisions with user CRDs having the same resource name.
+var excludedGroupResources = map[schema.GroupResource]bool{
+	{Group: "", Resource: "events"}:                             true,
```
MikeSpreitzer, line 101: There are actually two to list here.

```console
mspreitz@mjs13 kubestellar % kubectl api-resources | grep event
events                              ev           v1                                            true         Event
events                              ev           events.k8s.io/v1                              true         Event
```

pkg/binding/binding-controller.go
MikeSpreitzer: @xonas1101: you got it exactly right.

pkg/binding/binding-controller.go
```diff
@@ -416,7 +419,12 @@ func (c *Controller) run(ctx context.Context, workers int, cListers chan interfa
 		}
 		logger.V(1).Info("Working on APIResourceList", "groupVersion", list.GroupVersion, "numResources", len(list.APIResources))
 		for _, resource := range list.APIResources {
-			if _, excluded := excludedResourceNames[resource.Name]; excluded {
+			gr := schema.GroupResource{
+				Group:    gv.Group,
+				Resource: resource.Name,
+			}
+
```
MikeSpreitzer: In my opinion there is no great pause or distinction to be made here, no need for a blank line.

pkg/binding/binding-controller.go
```diff
@@ -95,16 +95,19 @@ var excludedGroups = map[string]bool{
...
-var excludedResourceNames = map[string]bool{
-	"events":               true,
-	"nodes":                true,
-	"csistoragecapacities": true,
-	"csinodes":             true,
-	"endpoints":            true,
-	"workstatuses":         true,
+// Resource GroupResources to exclude for watchers as they should not be delivered to other clusters.
+// Qualified by API group to avoid collisions with user CRDs having the same resource name.
+var excludedGroupResources = map[schema.GroupResource]bool{
```
MikeSpreitzer, line 100: I don't recall a reason to not simply use a `Set` here (see k8s.io/apimachinery/pkg/util/sets).

pkg/binding/binding-controller.go
xonas1101: sure, I will make the changes in a bit.

pkg/binding/binding-controller.go
xonas1101: that must have slipped out, I will add it.

xonas1101:
Thanks @MikeSpreitzer, I have started the testing locally after making the changes just to be extra sure, and i will push the changes in a bit.

PR #3618, opened 2026-01-12
📖 docs: Add implementation note for condition aggregation

### LINKED ISSUE #3535: doc: Update multi-wec-aggregated-status doc
### Describe the issue

For known KIND that are specially handled by Kubestellar and present as Argocd builtin health check, in the code implementation we are only focusing on the such conditions type which are necessary for health evaluation by Argocd. 

See this as well: https://github.com/kubestellar/kubestellar/pull/3515#discussion_r2578480703

## Requested changes: 
In [Condition Aggregation](https://github.com/gaurab-khanal/kubestellar/blob/8ffacf8f10dddacae7100e73ba092b43ab7600b0/docs/content/direct/multi-wec-aggregated-status.md#condition-aggregation) add a note mentioning explictly about the implementation approach.

AresPhoenix345:
## Summary

Adds a clarifying note in the Condition Aggregation section explaining that for workload kinds with Argo CD built-in health checks (Deployment, StatefulSet, ReplicaSet, DaemonSet), only condition types necessary for Argo CD health evaluation are aggregated.

This makes the selective aggregation approach explicit in the documentation.

## Related issue(s)

Fixes #3535

### DIFF
--- a/docs/content/direct/multi-wec-aggregated-status.md
+++ b/docs/content/direct/multi-wec-aggregated-status.md
@@ -100,6 +100,8 @@ These rules avoid making assumptions about workload semantics while still provid
 
 Conditions use uniform rules across all workload kinds.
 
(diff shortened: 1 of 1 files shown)


gaurab-khanal (approved):
LGTM

PR #3634, opened 2026-01-13
🐛fix: prevent snapshot crash when open-cluster-management namespace is missing

### LINKED ISSUE #2724: bug: snapshot script crashes on missing namespace
### Describe the bug

`scripts/kubestellar-snapshot.sh` crashes when an ITS does not have a namespace named "open-cluster-management". See https://github.com/kubestellar/kubestellar/issues/2717#issuecomment-2600623673 for an example.

I think that there are other things whose absence will cause the current edition of this script to crash.

See above.

### Steps To Reproduce

See above.

### Expected Behavior

No sign of failure

### Want to contribute?

### Additional Context

_No response_

shivansh-gohem:
### Description
This PR fixes a bug where `scripts/kubestellar-snapshot.sh` would crash if the `open-cluster-management` namespace was not present.

The script runs with `set -e`, so the previous command (`kubectl get deploy -n open-cluster-management ...`) caused an immediate exit when it returned a non-zero status code on clusters without OCM.

### Changes
- Added a conditional check (`kubectl get ns open-cluster-management`) before attempting to query resources inside that namespace.
- If the namespace is missing, the script now safely reports "Open-cluster-manager: not found" and continues execution.

### Related Issue
Fixes #2724

### DIFF
--- a/scripts/kubestellar-snapshot.sh
+++ b/scripts/kubestellar-snapshot.sh
@@ -461,7 +461,12 @@ for i in "${!cps[@]}" ; do # for all control planes in context ${context}
         echo -n -e "  - Status addon controller: pod=${COLOR_INFO}$status_pod${COLOR_NONE}, ns=${COLOR_INFO}$status_ns${COLOR_NONE}, version=${COLOR_INFO}$status_version${COLOR_NONE}, status="
         echostatus "$status_status"
         if [ -n "${cp_kubeconfig[cp_n]}" ]; then
-            ocm_version=$(KUBECONFIG=${cp_kubeconfig[cp_n]} kubectl get deploy -n open-cluster-management cluster-manager -o jsonpath={.spec.template.spec.containers[0].image} | cut -d: -f2)
+            ocm_version=""
+            # Fix for Issue #2724: Check if namespace exists to prevent crash
+            if KUBECONFIG=${cp_kubeconfig[cp_n]} kubectl get ns open-cluster-management > /dev/null 2>&1; then
+                ocm_version=$(KUBECONFIG=${cp_kubeconfig[cp_n]} kubectl get deploy -n open-cluster-management cluster-manager -o jsonpath={.spec.template.spec.containers[0].image} 2>/dev/null | cut -d: -f2 || true)
+            fi
+
             if [ -n "$ocm_version" ]; then
                 echo -e "  - Open-cluster-manager: version=${COLOR_INFO}$ocm_version${COLOR_NONE}"
             else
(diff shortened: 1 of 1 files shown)


shivansh-gohem:
@clubanderson could you please review this too !! thank you

clubanderson:
/ok-to-test

this is for @MikeSpreitzer

scripts/kubestellar-snapshot.sh
```diff
@@ -461,7 +461,12 @@ for i in "${!cps[@]}" ; do # for all control planes in context ${context}
         echo -n -e "  - Status addon controller: pod=${COLOR_INFO}$status_pod${COLOR_NONE}, ns=${COLOR_INFO}$status_ns${COLOR_NONE}, version=${COLOR_INFO}$status_version${COLOR_NONE}, status="
         echostatus "$status_status"
         if [ -n "${cp_kubeconfig[cp_n]}" ]; then
-            ocm_version=$(KUBECONFIG=${cp_kubeconfig[cp_n]} kubectl get deploy -n open-cluster-management cluster-manager -o jsonpath={.spec.template.spec.containers[0].image} | cut -d: -f2)
+            ocm_version=""
+            # Fix for Issue #2724: Check if namespace exists to prevent crash
+            if KUBECONFIG=${cp_kubeconfig[cp_n]} kubectl get ns open-cluster-management > /dev/null 2>&1; then
+                ocm_version=$(KUBECONFIG=${cp_kubeconfig[cp_n]} kubectl get deploy -n open-cluster-management cluster-manager -o jsonpath={.spec.template.spec.containers[0].image} 2>/dev/null | cut -d: -f2 || true)
```
MikeSpreitzer: Let's omit the `2>/dev/null`, as in the original; it is better if something is printed when there is an unexpected error.

scripts/kubestellar-snapshot.sh
```diff
@@ -699,4 +704,4 @@ for h in "${!cp_pch[@]}" ; do
             done
         done
     fi
-done
+done
```
MikeSpreitzer: Please do not delete the linebreak at the end of this file.

Image

scripts/kubestellar-snapshot.sh
shivansh-gohem: Done! I removed 2>/dev/null so errors are visible. Ready for re-review

scripts/kubestellar-snapshot.sh
shivansh-gohem: resolved!!

shivansh-gohem:
hello @MikeSpreitzer could you please have a look at this pr!!

clubanderson:

/lifecycle stale

clubanderson:

/lifecycle rotten

clubanderson:

/close

PR #3661, opened 2026-01-15
🐛 Fix docs-release workflow JSON payload

clubanderson:
## Summary
Fix the docs-release workflow to properly pass JSON payload to gh api.

The `-f client_payload="{...}"` syntax was passing the payload as a string instead of a JSON object, causing HTTP 422 errors:
```
For 'properties/client_payload', "{...}" is not an object.
```

## Changes
- Use heredoc with `--input -` to pass proper JSON body

## Testing
Verified fix works in kubectl-claude v0.4.3 release automation.

🤖 Generated with Claude Code


clubanderson:
Closing - the docs-release.yml workflow was removed from main, so this fix is no longer applicable.

PR #3713, opened 2026-02-12
✨ util: add String() method to ObjectIdentifier for readable logs

AritraDey-Dev:
`ObjectIdentifier` is used in log messages across the binding and status controllers. Without a String() method, klog prints it as a big nested JSON blob that is hard to read:
```
  objectIdentifier={"GVK":{"Group":"apps","Version":"v1","Kind":"Deployment"},"Resource":"deployments","ObjectName":{"Namespace":"default","Name":"nginx"}}
```

With this change, the same log line becomes:
```
  objectIdentifier="apps/v1/deployments default/nginx"
```

For cluster-scoped objects (no namespace), it prints just the name:
```
  objectIdentifier="rbac.authorization.k8s.io/v1/clusterroles admin"
```

### DIFF
--- a/pkg/util/object-identifier_test.go
+++ b/pkg/util/object-identifier_test.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2024 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/cache"
+)
+
+func TestObjectIdentifierString(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/util/object-identifier.go
```diff
@@ -36,6 +38,18 @@ type ObjectIdentifier struct {
...
 }
 
+// String returns a human-readable representation of the ObjectIdentifier.
+// For namespaced objects it returns "group/version/resource namespace/name",
+// for cluster-scoped objects it returns "group/version/resource name".
+// Core resources (empty group) render as "v1/pods" rather than "/v1/pods".
+func (identifier ObjectIdentifier) String() string {
+	gvr := identifier.GVK.GroupVersion().String() + "/" + identifier.Resource
+	if identifier.ObjectName.Namespace != "" {
+		return fmt.Sprintf("%s %s/%s", gvr, identifier.ObjectName.Namespace, identifier.ObjectName.Name)
```
MikeSpreitzer: I think this would be even better if the returned value did not have a space in it; a space in such a value makes parsing log messages a bit tricky. Perhaps something like `fmt.Sprintf("%s(%s/%s)", gvr, identifier.ObjectName.Namespace, identifier.ObjectName.Name)` (analogous to a Go-language type-cast)?

AritraDey-Dev: How about using parentheses like this` apps/v1/deployments(default/nginx)` ?

AritraDey-Dev: @MikeSpreitzer any suggestions?

MikeSpreitzer: You suggested the same change that I did.

PR #3714, opened 2026-02-12
🌱 util: add missing test cases for `IsAPIGroupAllowed`

AritraDey-Dev:
The existing tests only covered positive cases where a group is allowed. This adds two more cases to cover the full behavior:

- a group that is not in the allowed set (should return false)
- an empty set, which means no restrictions (should return true, same as nil)

These new cases make it easier to understand the function's semantics and catch regressions if someone changes the len==0 check.


MikeSpreitzer (commented):
/lgtm
/approve
Thanks!

PR #3718, opened 2026-02-13
🐛 Improve troubled Pod detection in demo env test

### LINKED ISSUE #3711: The Pod health survey in test/e2e/ci/test-demo-env.sh may be too strict
The Pod health survey introduced in #3541 may be too strict: sometimes it makes the script fail when the only failure is a Job that gets tried twice and succeeds on the second try; the failed first try Pod still exists and appears in the listing of troubled Pods.

The fact that the Job fails even once may be a signal of a problem. This is worth debugging. Look at https://github.com/kubestellar/kubestellar/actions/workflows/test-demo-env-creation-script.yml for examples (but you have to look carefully, because a retry hides the error in that summary). Here is one: https://github.com/kubestellar/kubestellar/actions/runs/21892252651/job/63200434712

xonas1101:
## Summary
Replace grep-based Pod status checks with structured logic and proper Job status inspection. Prevents false failures from legitimate Job retries and improves test reliability.

## Related issue(s)

Fixes #3711

### DIFF
--- a/test/e2e/ci/test-demo-env.sh
+++ b/test/e2e/ci/test-demo-env.sh
@@ -51,10 +51,49 @@ done
 sleep 30
 
 # Check that there are no Pods in trouble
-listing=$(date; kubectl --context ${platform}-kubeflex get pods -A | grep -vw Running | grep -vw Completed)
-if ! wc -l <<<"$listing" | grep -qw 2; then
+echo "Checking for troubled Pods..."
+
+troubled=0
+
+while read ns name ready status rest; do
+    # Skip healthy Pods
+    if [[ "$status" == "Running" || "$status" == "Completed" ]]; then
+        continue
+    fi
+
+    # Check if Pod is owned by a Job
+    owner_kind=$(kubectl --context ${platform}-kubeflex get pod -n "$ns" "$name" \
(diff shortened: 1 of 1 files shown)


xonas1101:
/cc @MikeSpreitzer 
I really don't know if I did this right. I ran some tests locally tho, and they seemed to be working just fine. Would appreciate a review.

test/e2e/ci/test-demo-env.sh
```diff
@@ -51,10 +51,49 @@ done
...
+        # Get Job status in one call (handles race if Job deleted)
+        job_info=$(kubectl --context ${platform}-kubeflex get job -n "$ns" "$owner_name" \
+            -o jsonpath='{.status.conditions[?(@.type=="Complete")].status},{.status.conditions[?(@.type=="Failed")].status},{.status.succeeded},{.status.active}' 2>/dev/null || echo "")
+
+        if [[ -z "$job_info" ]]; then
+            echo "Ignoring Pod $ns/$name (Job $owner_name not found, likely cleaned up)"
+            continue
+        fi
+
+        IFS=',' read -r complete_status _ succeeded active <<< "$job_info"
```
MikeSpreitzer, line 80: Why is the bespoke output above including a value that is discarded here?

PR #3735, opened 2026-03-07
🐛 fix: correct ManagedCluster resource name typo in metrics GVR

Sanchit2662:
## Summary

I noticed that the ManagedCluster metrics client was using the wrong resource name - `manageclusters` instead of `managedclusters` (missing the 'd'). This was causing all Prometheus metrics for ManagedCluster operations to carry the incorrect label manageclusters.v1.cluster.open-cluster-management.io rather than the canonical managedclusters.v1.cluster.open-cluster-management.io.

The **fix** is a single character: adding the missing 'd' to the resource string in the GVR construction. The actual API client was already correct - this only affected the metrics wrapping label, so there's zero functional risk, just correct telemetry.
I verified the correct spelling against the OCM API definitions and other references in the codebase (like the integration tests). With this change, dashboards and alerts filtering by the proper resource name will correctly pick up these metrics.


Sanchit2662:
Hey @MikeSpreitzer , whenever you have a moment, would appreciate a quick look - thanks!

MikeSpreitzer (commented):
/lgtm
/aprove
Thanks!

PR #3758, opened 2026-04-10
🐛 binding.go bad SR capacity and nil spec handling

tushar743-ui:
🐛 Fix off-by-one in badSR cap, nil-spec race condition, and minor code quality improvements

## Summary

A few small but meaningful fixes in the binding sync path.

The `badSR` slice was silently collecting one extra element before the cap kicked in. Moved the guard before the append and introduced a named constant `maxBadSRExamples` to make the intent clear.

A transient nil Binding spec during a TOCTOU window was bubbling up as a hard error, which wasn't consistent with how the rest of the function handles similar races. It now logs quietly and returns nil instead, matching the BindingPolicy-gone behavior nearby.

Struct fields in `objectWithNumWECs` are now named explicitly, so they're safe from future reordering. The operator-facing error message swaps `%#v` for `%v` to show something readable instead of a raw Go struct dump. 

## Related issue(s)
Fixes #

### DIFF
--- a/pkg/binding/binding.go
+++ b/pkg/binding/binding.go
@@ -29,6 +29,10 @@ import (
 	"github.com/kubestellar/kubestellar/pkg/util"
 )
 
+// maxBadSRExamples is the maximum number of singleton-reported-state violation
+// examples included in the BindingPolicy status error message.
+const maxBadSRExamples = 3
+
 // syncBinding syncs a binding object with what is resolved by the bindingpolicy resolver.
 func (c *Controller) syncBinding(ctx context.Context, bindingName string) error {
 	logger := klog.FromContext(ctx)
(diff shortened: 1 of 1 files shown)


pkg/binding/binding.go
```diff
@@ -92,16 +99,16 @@ func (c *Controller) syncBinding(ctx context.Context, bindingName string) error
...
 				break
 			}
+			badSR = append(badSR, objectWithNumWECs{ObjectID: srStatus.ObjectId, NumWECs: srStatus.NumWECs})
 		}
 	}
 	if len(badSR) > 0 {
 		badSRBytes, err := json.Marshal(badSR)
 		if err != nil {
-			policyErrors = append(policyErrors, fmt.Sprintf("Failed to json.Marshal some example blighted objects (%#v): %s", badSR, err))
+			policyErrors = append(policyErrors, fmt.Sprintf("Failed to json.Marshal some example blighted objects (%v): %s", badSR, err))
```
MikeSpreitzer, line 111: What was wrong with `%#v` ?

`objectWithNumWECs` was born with named fields.

 I find the formatting of complex data structures by `%#v` to be clearer than by `%v`. In this case, reporting a failure to marshal, it is particularly important for the error message to be precise.

Failures to marshal should be very unusual. Do you have examples of this being logged?

pkg/binding/binding.go
```diff
@@ -71,8 +75,11 @@ func (c *Controller) syncBinding(ctx context.Context, bindingName string) error
 
 	// generate binding spec from resolver
 	generatedBindingSpec := c.bindingPolicyResolver.GenerateBinding(bindingPolicyIdentifier)
-	if generatedBindingSpec == nil { // resolution does not exist, abort syncing
-		return fmt.Errorf("syncing Binding was stopped because it has no counterpart resolution")
+	if generatedBindingSpec == nil {
+		// Resolution was concurrently removed between ResolutionExists and GenerateBinding.
+		// This is a normal transient race during deletion; treat it like the Policy-gone case above.
```
MikeSpreitzer, line 80: Look at the handling above and you will see that something (`NotifyBindingPolicyCallbacks`) is missing here.

There is not a bug here. Returning an error will cause a retry, which will go through the one copy of the proper code. This is a rare circumstance, and duplicating code is bad.

MikeSpreitzer:
> The badSR slice was silently collecting one extra element before the cap kicked in.

Who says it was extra? The limit is 4.

clubanderson:

/lifecycle stale

PR #3759, opened 2026-04-11
🐛 fix(transport): clean up ctNameToSpec entries when no bindings remain for a GroupResource

tushar743-ui:
## Summary

In `setBindingGroupResources`, when a GroupResource is removed from a Binding's dependency set and no other Binding references it anymore, the `grToTransformData` entry was correctly deleted, but the corresponding `ctNameToSpec` entries for that GroupResource's CustomTransform names were left behind, causing unbounded memory growth over time as CustomTransforms cycle through GroupResources.

This fix mirrors the cleanup already done in `invalidateCacheEntryLocked`, which correctly deletes `ctNameToSpec` entries alongside the `grToTransformData` entry when a GroupResource is invalidated.

**Change:** In the `grTransformData.bindingsThatCare.Len() == 0` block inside `setBindingGroupResources`, added cleanup of `ctNameToSpec` entries for the removed GroupResource's associated CustomTransform names.

## Related issue(s)
Fixes #


clubanderson:

/lifecycle stale

PR #3765, opened 2026-04-25
🌱 Restrict workflow token permissions

AkashKumar7902:
## Summary

- Add explicit least-privilege `GITHUB_TOKEN` permissions to the docs preview workflow.
- Move remaining required write scopes in reusable-workflow callers from workflow-wide permissions to job-scoped permissions.
- Reduce unused write scopes where the called workflow only reads PR or repository metadata, including removing the unused Scorecard `id-token: write` caller permission.
- Leave action references unchanged.

Part of #2626. Also improves the CLOMonitor token-permissions signal tracked by #2595.

## Validation

- `git show --check --oneline --decorate HEAD`
- `git diff --check HEAD~1..HEAD`
- `hack/gha-reversemap.sh verify-mapusage`
- `yq e '.' .github/workflows/ai-fix.yml .github/workflows/copilot-automation.yml .github/workflows/copilot-dco.yml .github/workflows/image-scanning.yml .github/workflows/pr-verifier.yml .github/workflows/scorecard.yml >/dev/null`
- `go run github.com/ossf/scorecard/v5@latest --checks=Token-Permissions --local=. --show-details` reports `10 / 10` for Token-Permissions

### DIFF
--- a/.github/workflows/copilot-automation.yml
+++ b/.github/workflows/copilot-automation.yml
@@ -11,10 +11,7 @@ on:
     types: [opened, synchronize, ready_for_review]
 
 permissions:
-  contents: write
-  pull-requests: write
-  issues: write
-  statuses: write
+  contents: read
 
 concurrency:
   group: copilot-automation-${{ github.event.pull_request.number || github.event.inputs.pr_number || github.sha }}
@@ -22,6 +19,11 @@ concurrency:
 
 jobs:
   copilot-automation:
+    permissions:
+      contents: read
+      pull-requests: read
+      issues: write
+      statuses: write
     uses: kubestellar/infra/.github/workflows/reusable-copilot-automation.yml@main
     with:
       pr_number: ${{ github.event.inputs.pr_number || '' }}
--- a/.github/workflows/copilot-dco.yml
+++ b/.github/workflows/copilot-dco.yml
@@ -6,10 +6,12 @@ on:
 
 permissions:
   contents: read
-  issues: write
-  pull-requests: read
-  statuses: write
 
 jobs:
   copilot-dco:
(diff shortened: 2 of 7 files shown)


.github/workflows/copilot-automation.yml
```diff
@@ -11,17 +11,19 @@ on:
...
 
 concurrency:
   group: copilot-automation-${{ github.event.pull_request.number || github.event.inputs.pr_number || github.sha }}
   cancel-in-progress: true
 
 jobs:
   copilot-automation:
+    permissions:
+      contents: read
+      pull-requests: read
```
MikeSpreitzer, line 24: Are these necessary, considering the workflow-level declaration?
(Same question for other workflows.)

.github/workflows/copilot-dco.yml
```diff
@@ -4,6 +4,14 @@ on:
...
     types: [opened, synchronize, reopened]
 
+permissions:
+  contents: read
+
 jobs:
   copilot-dco:
+    permissions:
+      contents: read
+      issues: write
```
MikeSpreitzer, line 14: Why was this workflow working before this permission bump? Same question for the statuses permission.

MikeSpreitzer: Because this repo's default is very permissive.

.github/workflows/copilot-automation.yml
MikeSpreitzer: I think the `pull-requests: read` may be needed but the `contents: read` makes no change, so is not needed.

MikeSpreitzer (commented):
/lgtm
/approve

PR #3768, opened 2026-04-30
✨ Bump go version in core.Dockerfile

MikeSpreitzer:
## Summary
Build with Go 1.26 instead of 1.24.
Needed because some dependencies now require 1.25.

## Related issue(s)

Fixes #


MikeSpreitzer:
See #3769 instead

PR #3805, opened 2026-06-07
🐛 fix(status): use DynType for inventory CEL variable

Norway-02:
## Summary
Fixes `CheckExpression` in the CEL evaluator to use `DynType` instead of `StringType` for the inventory variable, which allows non-string values to be properly evaluated.
Also updates the tests in `cel-evaluator_test.go` to explicitly assert that the result is a boolean type, providing more informative error messages if the evaluation returns an unexpected type.

## Related issue(s)
N/A


Norway-02 (commented):
Addressed review comments: added explicit boolean type assertions to `result.Value()` checks in tests to ensure clear failure messages.

Norway-02 (commented):

PR #3811, opened 2026-06-08
Fix/status deepcopy informer cache corruption

Norway-02:
#### Problem
In `pkg/status`, `updateObjectStatus` retrieved the workload object directly from the lister (backed by the informer cache) and mutated `unstrObj.Object["status"]` in-place before calling `UpdateStatus`. If `UpdateStatus` failed for any reason — network error, conflict, server unavailability — the informer cache was left permanently holding the incorrect status value.

On the next reconciliation, the controller would read this corrupted cached state, find the status already equal to the desired value, and silently skip the API update entirely — leaving the workload object in WDS with stale status until a full informer resync.

#### Fix
**DeepCopy before mutation (`workstatus.go`)**
Added `unstrObj = unstrObj.DeepCopy()` immediately before the status mutation in the `else` block, so the original cached object is never touched. If `UpdateStatus` fails, the cache remains intact and the next reconciliation correctly retries the update.

This mirrors the existing pattern in `handleStatusReturnLabel` in the same file, which already applied `DeepCopy` before mutating labels for exactly this reason. 

*(Note: The unrelated CEL inventory type mismatch fixes previously in this PR have been extracted into their own PR to keep this PR focused purely on the cache corruption issue.)*

#### Tests added (`workstatus_test.go`)
Test | Purpose
-- | --
TestUpdateObjectStatus_CacheNotMutatedOnFailure | Regression: proves informer cache is not mutated when UpdateStatus fails
TestUpdateObjectStatus_UpdatesStatusWhenDifferent | Happy path: status is correctly written to the API server when it differs
TestUpdateObjectStatus_NoUpdateWhenStatusMatches | Idempotency: no status subresource call is made when status already matches
TestUpdateObjectStatus_NilStatusNoLabel | No-op: nil status with no existing label results in no API call

### DIFF
--- a/pkg/status/workstatus.go
+++ b/pkg/status/workstatus.go
@@ -138,13 +138,13 @@ func (c *Controller) updateObjectStatus(ctx context.Context, objectIdentifier ut
 	}
 
 	if wantSingleton && !haveSingleton {
-		err = c.handleStatusReturnLabel(ctx, unstrObj, objectIdentifier.GVR(), true, util.BindingPolicyLabelSingletonStatusKey)
+		err = c.handleStatusReturnLabel(ctx, unstrObj, objectIdentifier.GVR(), true, util.BindingPolicyLabelSingletonStatusKey, false)
 		if err != nil {
 			return err
 		}
 	}
 	if wantMultiWEC && !haveMultiWEC {
-		err = c.handleStatusReturnLabel(ctx, unstrObj, objectIdentifier.GVR(), true, util.BindingPolicyLabelMultiWECStatusKey)
+		err = c.handleStatusReturnLabel(ctx, unstrObj, objectIdentifier.GVR(), true, util.BindingPolicyLabelMultiWECStatusKey, false)
 		if err != nil {
 			return err
 		}
@@ -155,6 +155,8 @@ func (c *Controller) updateObjectStatus(ctx context.Context, objectIdentifier ut
 	if apiequality.Semantic.DeepEqual(unstrObj.Object["status"], status) {
 		logger.V(5).Info("Workload object found to already have intended status", "objectIdentifier", objectIdentifier)
 	} else {
+		// DeepCopy before mutating to avoid corrupting the informer cache.
+		unstrObj = unstrObj.DeepCopy()
 		// set the status and update the object
 		unstrObj.Object["status"] = status
 
@@ -172,14 +174,14 @@ func (c *Controller) updateObjectStatus(ctx context.Context, objectIdentifier ut
 	}
 
 	if haveSingleton && !wantSingleton {
-		err = c.handleStatusReturnLabel(ctx, unstrObj, objectIdentifier.GVR(), false, util.BindingPolicyLabelSingletonStatusKey)
+		err = c.handleStatusReturnLabel(ctx, unstrObj, objectIdentifier.GVR(), false, util.BindingPolicyLabelSingletonStatusKey, true)
 		if err != nil {
 			return err
 		}
 	}
 
 	if haveMultiWEC && !wantMultiWEC {
-		err = c.handleStatusReturnLabel(ctx, unstrObj, objectIdentifier.GVR(), false, util.BindingPolicyLabelMultiWECStatusKey)
+		err = c.handleStatusReturnLabel(ctx, unstrObj, objectIdentifier.GVR(), false, util.BindingPolicyLabelMultiWECStatusKey, true)
 		if err != nil {
 			return err
 		}
@@ -189,7 +191,7 @@ func (c *Controller) updateObjectStatus(ctx context.Context, objectIdentifier ut
 }
 
 func (c *Controller) handleStatusReturnLabel(ctx context.Context, unstructuredObj *unstructured.Unstructured,
-	objGVR schema.GroupVersionResource, wantLabel bool, labelKey string) error {
+	objGVR schema.GroupVersionResource, wantLabel bool, labelKey string, alreadyCopied bool) error {
 
 	labels := unstructuredObj.GetLabels() // gets a copy of the labels
 	_, foundLabel := labels[labelKey]
@@ -206,7 +208,9 @@ func (c *Controller) handleStatusReturnLabel(ctx context.Context, unstructuredOb
 		message = fmt.Sprintf("Removed %s label from workload object", labelKey)
 		delete(labels, labelKey)
 	}
-	unstructuredObj = unstructuredObj.DeepCopy() // avoid mutating the original object
+	if !alreadyCopied {
+		unstructuredObj = unstructuredObj.DeepCopy() // avoid mutating the original object
+	}
 	unstructuredObj.SetLabels(labels)
 	namespace := unstructuredObj.GetNamespace()
 	rscIfc := util.DynamicForResource(c.wdsDynClient, objGVR, namespace)
--- a/pkg/status/workstatus_test.go
+++ b/pkg/status/workstatus_test.go
@@ -1,0 +1,234 @@
+/*
+Copyright 2024 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package status
+
+import (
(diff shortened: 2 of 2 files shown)


pkg/status/cel-evaluator.go
```diff
@@ -52,11 +52,11 @@ func newCELEvaluator() (*celEvaluator, error) {
 	env, err := cel.NewEnv(
 		cel.Variable(sourceObjectKey, cel.MapType(cel.StringType, cel.DynType)),
 		cel.Variable(returnedKey, cel.MapType(cel.StringType, cel.DynType)),
-		cel.Variable(inventoryKey, cel.MapType(cel.StringType, cel.StringType)),
+		cel.Variable(inventoryKey, cel.MapType(cel.StringType, cel.DynType)),
```
MikeSpreitzer: This is also a good bug fix, should be mentioned in the PR description.

pkg/status/workstatus.go
```diff
@@ -155,6 +155,10 @@ func (c *Controller) updateObjectStatus(ctx context.Context, objectIdentifier ut
 	if apiequality.Semantic.DeepEqual(unstrObj.Object["status"], status) {
 		logger.V(5).Info("Workload object found to already have intended status", "objectIdentifier", objectIdentifier)
 	} else {
+		// DeepCopy before mutating to avoid corrupting the informer cache.
+		// If UpdateStatus fails the original cached object must remain unchanged.
+		// See handleStatusReturnLabel in this file which does the same for labels.
```
MikeSpreitzer: The first sentence here is enough; this is a totally standard concern, this is fixing an oversight.

pkg/status/workstatus.go
```diff
@@ -155,6 +155,10 @@ func (c *Controller) updateObjectStatus(ctx context.Context, objectIdentifier ut
...
+		unstrObj = unstrObj.DeepCopy()
```
MikeSpreitzer, line 159: This does indeed fix the bug, but in some cases causes the object to be deep copied twice: here and in a call below to `c.handleStatusReturnLabel`. It would be better to add a parameter to that method telling it whether the deep copy has already been done.

pkg/status/workstatus.go
MikeSpreitzer, line 164: FYI, this may also be a place that can side-effect `unstrObj`. What I know is a few releases old, but at least then the _typed_ client stubs would side-effect the `ObjectMeta` of the object being marshaled.

pkg/status/workstatus_test.go
```diff
@@ -0,0 +1,234 @@
...
+var testGVR = schema.GroupVersionResource{
+	Group:    "apps",
+	Version:  "v1",
+	Resource: "deployments",
+}
+
+var testGVK = schema.GroupVersionKind{
+	Group:   "apps",
+	Version: "v1",
+	Kind:    "Deployment",
```
MikeSpreitzer: FYI: it would be DRYer to use the `WithResource` and `WithKind` methods of https://github.com/kubernetes/api/blob/v0.32.13/apps/v1/register.go#L29

pkg/status/workstatus_test.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+func newTestUnstructured(namespace, name string, status map[string]interface{}) *unstructured.Unstructured {
+	obj := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "apps/v1",
```
MikeSpreitzer: FYI, it would be DRYer to use `apps.SchemeGroupVersion.String()`.

pkg/status/workstatus_test.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	actualStatus, _, _ := unstructured.NestedMap(cachedObj.Object, "status")
+	if fmt.Sprintf("%v", actualStatus) != fmt.Sprintf("%v", originalStatus) {
+		t.Errorf(
+			"BUG: informer cache was mutated after UpdateStatus failure.\ngot:  %v\nwant: %v",
+			actualStatus,
+			originalStatus,
+		)
+	}
```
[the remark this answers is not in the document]
MikeSpreitzer: I am not too concerned about the possibility of `NestedMap` failing on arguments that are fixed in this test.

The point about map iteration order is valid. I think that it would be better to compare using `reflect` because the whole point of `apiequality` is to make more things equal but the point of this test is to be as stringent as possible.

pkg/status/workstatus_test.go
```diff
@@ -0,0 +1,234 @@
...
+		"replicas": int64(5),
+	}
+
+	err := c.updateObjectStatus(ctx, objectIdentifier, newStatus, listers, false)
+	if err == nil {
+		t.Fatal("expected an error when UpdateStatus fails, got nil")
+	}
+
+	actualStatus, _, _ := unstructured.NestedMap(cachedObj.Object, "status")
+	if fmt.Sprintf("%v", actualStatus) != fmt.Sprintf("%v", originalStatus) {
```
MikeSpreitzer: If I understand correctly, this test does not actually test that there was no side-effecting of `cachedObj`. Since `originalStatus` is embedded in `cachedObj`, a side-effect to `cachedObj` might also side-effect `originalStatus`. Better to make a copy of `originalStatus` before invoking `c.updateObjectStatus`.

pkg/status/workstatus_test.go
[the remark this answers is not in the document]
Norway-02: Thanks for the clarification — switched to reflect.DeepEqual to keep the assertion as strict as possible, since apiequality would mask differences we actually want to catch here.

pkg/status/workstatus_test.go
```diff
@@ -0,0 +1,234 @@
...
+		"replicas": int64(1),
+	})
+
+	fakeDynClient := dynamicfake.NewSimpleDynamicClient(newFakeScheme(), cachedObj)
+
+	updateCalled := false
+	fakeDynClient.PrependReactor("update", "deployments", func(action k8stesting.Action) (bool, runtime.Object, error) {
+		updateCalled = true
+		return false, nil, nil
+	})
```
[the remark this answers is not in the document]
Norway-02: Done. Updated the reactor to assert action.(k8stesting.UpdateAction).GetSubresource() == "status" before setting updateCalled = true, so the test only passes when the update is specifically on the status subresource.

pkg/status/cel-evaluator.go
Norway-02: Done. Added a dedicated section in the PR description for the inventoryKey cel.StringType → cel.DynType fix.

pkg/status/workstatus.go
Norway-02: Done. Trimmed the comment to just the first sentence — the remaining two lines were redundant given the context.

pkg/status/workstatus.go
Norway-02: Done. Added an alreadyCopied bool parameter to handleStatusReturnLabel. Call sites in the label-removal path (lines 177, 184) pass true since DeepCopy was already done earlier in updateObjectStatus. Call sites in the label-addition path (lines 141, 147) pass false since no prior copy exists there.

pkg/status/workstatus.go
Norway-02: Thanks for the heads-up. Since unstrObj is already a deep copy at this point (from the fix above), any side-effects from UpdateStatus on ObjectMeta would only affect the local copy and not the informer cache. Happy to add another DeepCopy before UpdateStatus if you think it's necessary.

pkg/status/workstatus_test.go
Norway-02: Done! Used appsv1.SchemeGroupVersion.WithResource/WithKind as suggested. Tests pass.

pkg/status/workstatus_test.go
Norway-02: Done! Replaced hardcoded "apps/v1" with appsv1.SchemeGroupVersion.String(). Tests pass.

pkg/status/workstatus_test.go
Norway-02: Done! Added a shallow copy of originalStatus before passing it to newTestUnstructured, so mutations to cachedObj can't side-effect the reference value used in the comparison. Tests pass.

Norway-02 (commented):
@MikeSpreitzer Thanks for all the detailed review comments! I have addressed all of your feedback:
- Replaced the CEL changes with a dedicated PR so this PR only focuses on the cache corruption fix.
- Trimmed the verbose comment as suggested.
- Updated the DeepCopy flow to pass `alreadyCopied` to avoid double-copying.
- Switched the test assertions to strictly use `reflect.DeepEqual`.
- Updated test references to be DRYer using `appsv1.SchemeGroupVersion`.
- Cloned `originalStatus` in the test setup so mutations don't side-effect the reference value.
- Tightened the status subresource reactor check.

Ready for another look!

PR #3830, opened 2026-06-09
refactor(util): prevent deadlock risks and add tests for ConcurrentMap

Norway-02:
PR Title: 🐛 refactor(util): prevent deadlock risks and add tests for ConcurrentMap

## Summary
This PR fixes a severe concurrency design defect in `pkg/util/concurrent-map.go` and adds critical missing unit test coverage for the package. 

Previously, the `Iterator` method executed the user-supplied `yield` callback while actively holding the package's internal read lock (`mm.RLock()`). If the callback invoked any write operation on the map (like `Set`) or encountered a blocking state, it triggered a deterministic deadlock. 

The updated routine isolates memory reading by copying keys and values safely into a short-lived local snapshot before yielding context control to the caller. 

Additionally, this PR includes:
- `TestConcurrentMap_Basic`: Verifies standard operations under parallel read/write load.
- `TestConcurrentMap_Iterator_SafeMutation`: Explicitly captures the deadlocking pattern and passes successfully with the new early-unlock snapshot layout.

Locally verified via `go fmt`, `go vet`, `go test -race`, and `golangci-lint run`.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/util/concurrent-map.go
+++ b/pkg/util/concurrent-map.go
@@ -82,17 +82,21 @@ func (mm *rwMutexMap[K, V]) Get(key K) (V, bool) {
 	return value, ok
 }
 
-// Iterator iterates over the map and calls the given function for each
-// key/value pair sequentially.
+// Iterator iterates over a snapshot of the map and calls the given function
+// for each key/value pair sequentially. This prevents deadlocks if the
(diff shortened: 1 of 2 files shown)


Norway-02:

PR #3831, opened 2026-06-11
🐛 fix(status): handle AlreadyExists on create and fix alreadyCopied bug

Norway-02:
## Summary
Resolves three issues related to the status controller:
1. Fixes a bug in `workstatus.go` where `alreadyCopied` was incorrectly passed as true when status matched, causing `SetLabels` to mutate the informer-cached object. It now properly tracks whether a DeepCopy occurred.
2. In `combinedstatus.go`, returns a retryable error when `AlreadyExists` occurs during creation, rather than swallowing the error by returning nil. This ensures the update intent is not dropped during create/update races.
3. Updates `workstatus_test.go` to use the comma-ok idiom for type assertion on `UpdateAction`, avoiding panics on unexpected actions.

## Related issue(s)
N/A

### DIFF
--- a/pkg/status/cel-evaluator.go
+++ b/pkg/status/cel-evaluator.go
@@ -52,11 +52,11 @@ func newCELEvaluator() (*celEvaluator, error) {
 	env, err := cel.NewEnv(
 		cel.Variable(sourceObjectKey, cel.MapType(cel.StringType, cel.DynType)),
 		cel.Variable(returnedKey, cel.MapType(cel.StringType, cel.DynType)),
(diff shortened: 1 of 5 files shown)


Norway-02 (commented):
Addressed review comments:
- Tracked `copied` explicitly to avoid mutating the informer cache when setting labels.
- Returned a retryable error on `AlreadyExists` during CombinedStatus creation to ensure the object converges.
- Switched to the comma-ok idiom for type assertion in the test to avoid panics.

Norway-02 (commented):

Issue #3832, opened 2026-06-11
bug: tests on k3d have started failing


MikeSpreitzer:
### Describe the bug

https://github.com/kubestellar/kubestellar/actions/workflows/test-demo-env-creation-script.yml shows that the .github/workflows/test-demo-env-creation-script.yml started failing its k3d job last week around the time of the 5.9.0 release of k3d. My own local testing confirms that `test/e2e/ci/test-demo-env.sh k3d` succeeds if I have release 5.8.3 of k3d and fails if I have release 5.9.0

### Steps To Reproduce

See above

### Expected Behavior

CI passes `main`

MikeSpreitzer:
Using k3d release 5.8.3, if I do `k3d cluster create` without `--image` then `kubectl version` reports "Server Version: v1.31.5+k3s1".

https://github.com/k3d-io/k3d/discussions/474 explains that the Kubernetes version is selectable from the available image builds at https://hub.docker.com/r/rancher/k3s/tags .

Govind07-max:
i would like to work on this issue.

MikeSpreitzer:
This has been resolved by #3833

PR #3833, opened 2026-06-11
Specify k3s image for k3d

MikeSpreitzer:
## Summary

This PR updates the way that k3d clusters are created to pin the Kube server version to 1.32, because the latest has been preventing vcluster from working.

This PR also lengthens the wait for the special customization namespace to appear, because it would fail otherwise.

This PR also applies image flattening for loading into k3d, because the newer k3d or kube needs it.

## Related issue(s)

This is related to Issue #3832


waltforme (commented):
/lgtm

PR #3834, opened 2026-06-12
🐛 fix(status): preserve existing status on transient WorkStatus errors

Norway-02:
## Summary
Fixes the logic in `multiwec-status` to only preserve the existing status if *all* WorkStatus reads or extractions failed. Previously, the condition `fetchErrors > 0` would preserve stale status even if some reads were successful.
Also renames `fetchErrors` to `workStatusReadErrors` for clarity and updates the corresponding log message.

## Related issue(s)
N/A


Norway-02 (commented):
Addressed review comments:
- Changed condition to `workStatusReadErrors == len(wsObjects)` to correctly detect when all fetches fail.
- Aligned log message wording with logic.
- Renamed `fetchErrors` to `workStatusReadErrors`.

Norway-02 (commented):

PR #3835, opened 2026-06-13
fix: resolve deadlock in Iterator when yield calls Get

Norway-02:
🐛 fix: resolve deadlock in Iterator when yield calls Get

SUMMARY
Iterator held RLock. yield call Get. Get try RLock again.
RWMutex not reentrant. Program hang forever. No panic. No error.

Fix: copy map under lock. Release lock. Walk copy.
yield run outside lock. Safe now.

Test added. Prove deadlock before. Pass after.
RELATED ISSUE
Fixes #


Norway-02:
@waltforme

PR #3837, opened 2026-06-13
cleanup: Remove unused type alias workStatusON

srajang1805:
Remove the unused `workStatusON` type alias from `pkg/status/status-controller.go`.

The alias was defined as:

type workStatusON cache.ObjectName

but it is never referenced anywhere in the codebase. The package already uses `cache.ObjectName` directly in function signatures and related logic, making the alias redundant.

Type: Cleanup


clubanderson:

/lifecycle stale

PR #3839, opened 2026-06-14
🐛 fix(util): trim spaces in API groups and resolve map deadlock risk

Norway-02:
## Summary
1. Fixes an issue where API groups containing spaces after the comma were not parsed correctly by trimming spaces during `ParseAPIGroupsString`.
2. Resolves a panic-induced deadlock risk in `concurrent-map.go`'s Iterator. By scoping the snapshot creation and using `defer mm.RUnlock()`, the lock is guaranteed to be released even if a runtime panic occurs before yielding.

## Related issue(s)
N/A


Norway-02 (commented):
Addressed review comments: made the Iterator snapshot exception-safe using an inner function and `defer mm.RUnlock()`.

Norway-02 (commented):

PR #3849, opened 2026-06-23
🐛 bug: fix objectIsQueried infinite loop on embedded-then-standalone matches

### LINKED ISSUE #3848: bug: objectIsQueried infinite loop on embedded-then-standalone field names
### Describe the bug

`objectIsQueried` in `pkg/status/combinedstatus-resolution.go` can enter an
infinite loop (a CPU spin, not a deadlock) when a queried field name first
appears as an embedded prefix of a longer token in the query string and then
also appears as a standalone whole-word token later in the same string.

The status-controller goroutine evaluating that StatusCollector query hangs
permanently. There is no panic or error in normal operation — the goroutine
simply stops making progress, silently stalling CombinedStatus updates for the
affected workload.

Minimal reproducer: objectIsQueried(&query, "foo") with query = "foobar foo".

N/A — this is a pure-logic bug in objectIsQueried, independent of any running
cluster or environment. Reproducible with a Go unit test alone; no snapshot
applicable.

### Steps To Reproduce

1. Add a unit test (or call directly) that invokes:
       q := "foobar foo"
       objectIsQueried(&q, "foo")
2. Run it:  go test ./pkg/status/... -run TestObjectIsQueried -timeout 15s
3. The call never returns. With a timeout, the test panics:

panic: test timed out after 15s
    running tests:
        TestObjectIsQueried/foobar_foo/foo (15s)

goroutine 27 [runnable]:
strings.Index(...)
github.com/kubestellar/kubestellar/pkg/status.objectIsQueried(...)
    pkg/status/combinedstatus-resolution.go:1231

The goroutine state is [runnable] (a CPU spin), stuck at line 1231.

### Expected Behavior

objectIsQueried(&"foobar foo", "foo") should return true — the standalone
"foo" token is a whole word — and terminate immediately. It should never hang.

### Additional Context

Root cause: objectIsQueried slices the query before searching —
strings.Index((*query)[idx:], obj) — which returns an offset RELATIVE to that
slice. The code then uses that relative value as an ABSOLUTE index into the
original string when calling isWholeWord, and re-slices from it next iteration.
After the first embedded (non-whole-word) match, the cursor oscillates between
wrong positions and never advances, looping forever.

Fix: track base (absolute window start) and rel (relative hit) separately,
compute abs = base + rel, and advance base = abs + len(obj).

I have a fix and a regression test ready and will open a PR.

/assign

VignanNallani:
## Bug

`objectIsQueried` in `pkg/status/combinedstatus-resolution.go` enters an
infinite loop whenever a queried field name first appears as an embedded
prefix of a longer token and then appears as a standalone whole-word token
later in the same query string (e.g. query `"foobar foo"`, field `"foo"`).

The function slices the query string before searching:

```go
idx = strings.Index((*query)[idx:], obj)   // returns RELATIVE offset
isWholeWord(query, idx, len(obj))           // but treats it as ABSOLUTE
idx += len(obj)
```

`strings.Index` returns the offset within the **sub-slice** starting at `idx`,
not within the original string. When the search window starts at absolute
position 3 and `strings.Index` returns 4 (relative), the code sets `idx = 4`
instead of `3 + 4 = 7`. The `isWholeWord` check then inspects the wrong
character, rejects the match, advances to 7, where `strings.Index` returns 0
(relative), sending `idx` back to 0, and the cycle repeats indefinitely.

**Impact:** any StatusCollector query where a field name is a prefix of another
identifier causes the status controller goroutine for that collector to hang
permanently, silently stalling CombinedStatus updates for the affected workload.

## Fix

Track `base` (absolute window start) and `rel` (relative result from
`strings.Index`) separately; compute the absolute position as `abs = base + rel`:

```go
rel := strings.Index((*query)[base:], obj)
if rel == -1 {
    return false
}
abs := base + rel
if isWholeWord(query, abs, len(obj)) {
    return true
}
base = abs + len(obj)
```

**Changed file:** `pkg/status/combinedstatus-resolution.go` — `objectIsQueried` function only.

## Regression test

`pkg/status/objectisqueried_test.go` — new file, two test functions:

- `TestObjectIsQueried` — 8 table-driven cases including the key regression
  case `"foobar foo" / "foo"` that hangs forever on the buggy code.
- `TestIsWholeWord` — 6 cases covering boundary conditions for the helper.

## How to run

No cluster, controller, or Kubernetes infrastructure needed:

```bash
go test ./pkg/status/... -run 'TestObjectIsQueried|TestIsWholeWord' -v -timeout 30s
```

Expected result: 14 subtests, all `--- PASS`, completing in under 5 seconds.

To verify the bug is real, revert the `objectIsQueried` change and re-run with
`-timeout 15s` — the test panics with `test timed out` in the
`foobar_foo/foo` subtest.

## Notes

- No other production code was modified.
- If maintainers prefer, the regression test can instead be added to an
  existing test file in `pkg/status`.

Fixes #3848

### DIFF
--- a/pkg/status/combinedstatus-resolution.go
+++ b/pkg/status/combinedstatus-resolution.go
@@ -1225,19 +1225,23 @@ func sortedStringSlice(s []string) []string {
 // That means, it must not be surrounded by alphanumeric characters.
 // query is assumed to be non-nil.
 func objectIsQueried(query *string, obj string) bool {
-	idx := 0
+	base := 0
 
 	for {
-		idx = strings.Index((*query)[idx:], obj) // slices share the same storage, therefore efficient
-		if idx == -1 {
+		// strings.Index returns a relative offset within (*query)[base:], so we
+		// must add base to get the absolute position before passing it to
+		// isWholeWord (which operates on the original string).
+		rel := strings.Index((*query)[base:], obj)
+		if rel == -1 {
 			return false
 		}
 
-		if isWholeWord(query, idx, len(obj)) {
+		abs := base + rel
+		if isWholeWord(query, abs, len(obj)) {
 			return true
 		}
 
-		idx += len(obj)
+		base = abs + len(obj)
 	}
 }
 
--- a/pkg/status/objectisqueried_test.go
+++ b/pkg/status/objectisqueried_test.go
@@ -1,0 +1,89 @@
+/*
+Copyright 2024 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Regression test for the objectIsQueried infinite loop (issue #3848).
+// Before the fix, inputs where the first occurrence of obj is an embedded
+// prefix of a longer word caused an infinite loop: strings.Index returns a
+// relative offset within the sliced substring, but the original code used it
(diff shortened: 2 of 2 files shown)


clubanderson:

/lifecycle stale

PR #3859, opened 2026-06-26
fix(util): prevent nil panic and handle cluster-scoped resources in G…

Norway-02:
🐛 fix(util): prevent nil panic and handle cluster-scoped resources in GetWorkStatusSourceRef

## Summary
Two bugs fixed in `GetWorkStatusSourceRef` in `pkg/util/unstructured.go`:

1. Unsafe type assertions on `spec` and `sourceRef` caused nil pointer panic
   when either field was absent. Fixed using two-value assertion form before nil check.

2. `namespace` field was required, breaking all cluster-scoped WorkStatus objects.
   Namespace is legitimately empty for cluster-scoped resources, so strict error
   check removed and empty string used as default.

## Related issue(s)
/kind bug

### DIFF
--- a/pkg/util/unstructured.go
+++ b/pkg/util/unstructured.go
@@ -104,12 +104,22 @@ func GetWorkStatusSourceRef(workStatus runtime.Object) (*SourceRef, error) {
 	}
 
 	spec, ok := obj.Object["spec"].(map[string]interface{})
+<<<<<<< HEAD
 	if !ok || spec == nil {
 		return nil, fmt.Errorf("could not find spec in object")
 	}
 	sourceRef, ok := spec["sourceRef"].(map[string]interface{})
 	if !ok || sourceRef == nil {
 		return nil, fmt.Errorf("could not find sourceRef in object spec")
+=======
+	if !ok {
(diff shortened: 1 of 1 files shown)


Norway-02:
@waltforme

pkg/util/unstructured.go
```diff
@@ -137,9 +137,13 @@ func GetWorkStatusSourceRef(workStatus runtime.Object) (*SourceRef, error) {
...
 	}
 
-	namespace, ok := sourceRef["namespace"].(string)
-	if !ok {
-		return nil, fmt.Errorf("could not find namespace in sourceRef or it's not a string")
+	namespace := ""
+	if nsObj, found := sourceRef["namespace"]; found && nsObj != nil {
+		ns, ok := nsObj.(string)
+		if !ok {
+			return nil, fmt.Errorf("could not find namespace in sourceRef or it's not a string")
```
waltforme: Since a missing namespace is now valid for cluster-scoped resources, this message is stale.

waltforme:
Looks good, thank you!

PR #3861, opened 2026-06-27
fix: use WorkStatus constants in CheckWorkStatusPresence

Norway-02:
🐛 fix: use WorkStatus constants in CheckWorkStatusPresence

## Summary
`CheckWorkStatusPresence` hardcodes GVR strings `"control.kubestellar.io"`,
`"v1alpha1"`, `"workstatuses"` instead of using constants `WorkStatusGroup`,
`WorkStatusVersion`, `WorkStatusResource` already defined in the same file.

If any constant changes, `CheckWorkStatusPresence` silently diverges and
returns `false` with no error logged, breaking WorkStatus feature detection.

Fix replaces hardcoded strings with existing constants.

## Related issue(s)
N/A

### DIFF
--- a/pkg/util/unstructured.go
+++ b/pkg/util/unstructured.go
@@ -206,9 +206,9 @@ func CheckWorkStatusPresence(config *rest.Config) bool {
 	}
 
 	gvr := schema.GroupVersionResource{
-		Group:    "control.kubestellar.io",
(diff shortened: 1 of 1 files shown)


Norway-02:
@waltforme

waltforme (approved):

PR #3868, opened 2026-07-02
fix(controller-manager): resolve exponential backoff and graceful shu…

Norway-02:
🐛 fix(controller-manager): resolve exponential backoff and graceful shutdown bugs

## Summary
This PR addresses two structural bugs in the controller-manager initialization (`cmd/controller-manager/main.go`), as well as a few typos in the flags and error handling:

**1. Status Controller Exponential Backoff Bug**
The loop waiting for the `WorkStatus` add-on attempts to log its status using a power-of-2 bitwise check `(i & (i - 1)) == 0`. However, the loop counter `i` was being incremented both in the `for` statement and inside the loop block. This caused `i` to only evaluate as odd numbers (1, 3, 5, 7...), meaning the exponential backoff log would never trigger after the first iteration. Removed the redundant `i++` to restore the intended behavior.

**2. Graceful Shutdown Bug**
The `main` function previously ended with a blocking `select {}`, ignoring the initialized OS signal context. Replaced this with `<-ctx.Done()` to allow the controller manager to shut down cleanly upon receiving termination signals.

*(Note: Also fixed a few minor spelling errors caught by golangci-lint: `emtpy` -> `empty`, `specifed` -> `specified`, `unkown` -> `unknown`)*

## Related issue(s)

Fixes #

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -80,7 +80,7 @@ func main() {
 	pflag.StringVar(&itsName, "its-name", "", "name of the Inventory and Transport Space to connect to (empty string means to use the only one)")
 	pflag.StringVar(&wdsName, "wds-name", "", "name of the workload description space to connect to")
 	pflag.StringVar(&allowedGroupsString, "api-groups", "", "list of allowed api groups, comma separated. Empty string means all API groups are allowed")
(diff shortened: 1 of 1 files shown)


Norway-02:
@waltforme

waltforme (approved):

PR #3879, opened 2026-07-04
🐛 fix(kubectl-rbac-flatten): prevent non-resource rules from expanding to all resources

Norway-02:
## Summary
Fixes `kubectl-rbac-flatten` to correctly handle RBAC rules that only contain `nonResourceURLs` (such as the default `system:discovery` ClusterRole). Previously, these rules caused the tool to default to wildcard `*` resources due to empty `APIGroups` and `Resources`, falsely reporting cluster-wide access.
This adds an explicit check for `len(rule.NonResourceURLs) > 0` to map the non-resource paths and safely bypass the wildcard resource expansion.

## Related issue(s)
N/A

### DIFF
--- a/cmd/kubectl-rbac-flatten/README.md
+++ b/cmd/kubectl-rbac-flatten/README.md
@@ -34,7 +34,8 @@ object, role-ism, and binding --- with the exception of wildcards.
 
     A non-resource URL path is just the path part of a URL; the schema
     and authority of the authorized URL are implicitly those of the
-    apiserver.
+    apiserver. Note that rules defining only `NonResourceURLs` (like `system:discovery`)
+    apply exclusively to those paths and will not be expanded against API resources.
 
     An API object is identified by a "resource", "object name"
     (misnamed "ResourceName" in the RBAC API), and possibly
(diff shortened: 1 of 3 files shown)


Norway-02 (commented):

magic-peach:
## Review: fix(kubectl-rbac-flatten): prevent non-resource rules from expanding to all resources

**Problem Analysis:** ✅ Well understood
- Rules with only `NonResourceURLs` (like `system:discovery`) were incorrectly defaulting to wildcard `*` resources
- This is a real bug that could cause security confusion

**Solution:** ✅ Appropriate
- Adding `len(rule.NonResourceURLs) > 0` check is the right approach
- 12 lines of focused changes

**Suggestions:**
1. Add a test case that verifies the fix with a `system:discovery`-like rule
2. Consider documenting this edge case in the tool's README

**Questions:**
1. Are there other tools in the kubestellar ecosystem that might have the same issue?

**Overall:** Approve. This is a correct and important fix for the RBAC flattening tool.

Norway-02:
Thanks @magic-peach! I've added a test case (in `main_test.go`) simulating a `system:discovery` rule to verify it doesn't expand into API resources, and I've also documented this edge case in the README. Force-pushed the updates!

Norway-02:

Norway-02:

PR #3883, opened 2026-07-08
🐛 fix(binding): prevent double-close panic on informer stopper channels

Norway-02:
## Summary
Prevents a "close of closed channel" panic on informer stopper channels. 
Previously, a `defer close(stopper)` was used inside a loop within `run()`, but deferred calls execute when the function returns, not at the end of the loop. If `syncCRD` already closed a stopper for a removed CRD, the deferred close would fire again on shutdown, causing a panic. 
This fix removes the defer and explicitly safely closes any remaining stoppers during shutdown.

## Related issue(s)
N/A

### DIFF
--- a/pkg/binding/binding-controller.go
+++ b/pkg/binding/binding-controller.go
@@ -464,9 +464,13 @@ func (c *Controller) run(ctx context.Context, workers int, cListers chan interfa
 				// run the informer
 				// we need to be able to stop informers for APIs (CRDs) that are removed
 				// after startup, therefore we use a stopper channel for each informer
-				// instead than informerFactory.Start(ctx.Done())
+				// instead than informerFactory.Start(ctx.Done()).
+				// Do NOT use defer close(stopper) here: defer runs when run() returns, not
+				// at the end of the loop iteration. If syncCRD already closed a stopper
+				// for a removed CRD, the deferred close would fire again on shutdown and
+				// panic with "close of closed channel". Stoppers are closed either by
+				// syncCRD (for individual CRD removal) or by the shutdown goroutine below.
 				stopper := make(chan struct{})
-				defer close(stopper)
 				c.stoppers.Set(gvr, stopper)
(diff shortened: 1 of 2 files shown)


Norway-02 (commented):

magic-peach:
## Review: fix(binding): prevent double-close panic on informer stopper channels

**Problem Analysis:** ✅ Correct
- The issue is well-understood: `defer close(stopper)` inside a loop runs when the function returns, not at the end of each iteration
- If `syncCRD` closes a stopper for a removed CRD, the deferred close would panic on shutdown

**Solution:** ✅ Appropriate
- Removing the `defer` and using explicit shutdown is the right approach
- `safeClose` with `recover()` is a defensive pattern for handling concurrent closes

**Concerns:**
1. The `safeClose` function silently swallows all errors, including unexpected ones. Consider logging at a debug level:
   ```go
   func safeClose(ch chan struct{}) {
       defer func() {
           if r := recover(); r != nil {
               // Optionally log: "channel already closed"
           }
       }()
       close(ch)
   }
   ```

2. There's an indentation issue in `crd.go:100` - the `} else {` has extra indentation

**Suggestions:**
1. Add a comment explaining why `safeClose` is needed (concurrent close from `syncCRD` and shutdown)
2. Consider adding a test that verifies no panic occurs when closing an already-closed channel

Overall: **Approve with minor suggestions**. The fix is correct and prevents a real panic.

Norway-02:
Thanks for the detailed review, @magic-peach! I've addressed the indentation issue, added documentation for safeClose, and added a log inside the recover block.

Norway-02:
Thanks @magic-peach! Great suggestion on the debug log inside `safeClose`. I've added a `klog.V(4).Infof` to explicitly log when a double-close is recovered from, rather than silently swallowing it. Force-pushed the update!

Norway-02:

Norway-02:

PR #3898, opened 2026-07-12
🐛 Fix panic on groupBy expression returning a list or map

### LINKED ISSUE #3897: bug: groupBy expression returning a list or map panics the status controller (hash of unhashable type)
### Describe the bug

A StatusCollector whose `groupBy` expression evaluates to a composite value (a list or a map) crashes the status controller with `panic: hash of unhashable type`.

In `handleAggregationReadLocked` the evaluated `groupBy` value is used directly as a key in a `map[any]int`:

```go
// pkg/status/combinedstatus-resolution.go:800
ValueToNumber := map[any]int{}
...
// pkg/status/combinedstatus-resolution.go:830
uid, exists := ValueToNumber[groupByValue.Value()]
```

`groupByValue.Value()` is the native Go value of the CEL evaluation. When the expression returns a list or a map, that value is a `[]interface{}` / `map[string]interface{}`, which is not comparable, so indexing the map panics.

Validation does not prevent this. `validateStatusCollector` runs each `groupBy` expression through `CheckExpression`, which only parses and type-checks the expression against a CEL environment where `obj` is `map[string]DynType`. An expression such as `obj.spec.containers` type-checks successfully, so it passes admission and only fails later at evaluation time.

The panic happens inside a workqueue worker (`syncCombinedStatus` → `CompareCombinedStatus` → `compareCombinedStatus` → `generateCombinedStatus` → `handleAggregationReadLocked`). Because the offending StatusCollector stays in the cache, the item is reprocessed and the controller keeps crashing.

### Steps To Reproduce

1. Have a workload whose objects contain a list or map field (e.g. a Deployment/Pod with `spec.containers`).
2. Create a StatusCollector that groups by an expression returning that composite value, for example:

```yaml
kind: StatusCollector
metadata:
  name: group-by-list
spec:
  combinedFields:
    - name: count
      type: COUNT
  groupBy:
    - name: containers
      def: "obj.spec.containers"   # evaluates to a list
```

3. Reference the StatusCollector from a BindingPolicy so it is applied to reported state.
4. The status controller evaluates the groupBy for a returned WorkStatus and panics with `hash of unhashable type: []interface {}`.

This can be reproduced in isolation with the package's own evaluator: evaluate a list-returning expression via `celEvaluator`, then use the result of `.Value()` as a key into a `map[any]int`. `CheckExpression` accepts the expression, `.Value()` returns `[]interface {}`, and the map index panics with `hash of unhashable type: []interface {}`.

### Expected Behavior

A `groupBy` expression that returns a composite value should not crash the controller. It should either be rejected at validation time or handled at evaluation time, rather than panicking in a workqueue worker and crash-looping.

### Additional Context

Affected code: `pkg/status/combinedstatus-resolution.go:830` (map key), reached from `handleAggregationReadLocked`. Validation entry point: `validateStatusCollector` in `pkg/status/statuscollector.go` via `CheckExpression` in `pkg/status/cel-evaluator.go`.

santhil-cyber:
## Summary

A StatusCollector whose `groupBy` expression evaluates to a composite value (a list or a map) crashes the status controller with `panic: hash of unhashable type`.

In `handleAggregationReadLocked` the evaluated groupBy value was used directly as a key in a `map[any]int`:

```go
ValueToNumber := map[any]int{}
...
uid, exists := ValueToNumber[groupByValue.Value()]
```

`groupByValue.Value()` is the native Go value of the CEL evaluation. For a list or map that is a `[]interface{}` / `map[string]interface{}`, which is not comparable, so indexing the map panics. Validation doesn't catch it either: groupBy expressions are checked against a CEL environment where `obj` is `map[string]DynType`, so something like `obj.spec.containers` type-checks fine and only fails later at evaluation time. Because that happens inside a workqueue worker and the offending StatusCollector stays in the cache, the controller crash-loops.

The fix reduces each groupBy value to a stable string key before the unique-number mapping. Scalars are keyed by their type and value; composite values are keyed by their JSON encoding, which mirrors how `refValToValue` already renders composite values downstream. Equal values still land in the same group and distinct values stay apart, so grouping behaviour is unchanged for the scalar cases that worked before.

## Testing

- Added `aggregation_groupby_test.go`: one test drives `handleAggregationReadLocked` with a list-valued groupBy across three WECs (two sharing a value, one differing) and asserts no panic plus the expected two groups; another exercises the key helper directly (equal composites collide, different composites and scalars don't). Reverting the fix makes the first test panic with `hash of unhashable type: []interface {}`, so it genuinely guards the regression.
- `go build ./...`, `go vet ./pkg/status/...`, and `go test ./pkg/status/...` all pass.

## Related issue(s)

Fixes #3897

### DIFF
--- a/pkg/status/aggregation_groupby_test.go
+++ b/pkg/status/aggregation_groupby_test.go
@@ -1,0 +1,112 @@
+/*
+Copyright 2024 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package status
+
+import (
+	"testing"
+
+	celtypes "github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+
+	"github.com/kubestellar/kubestellar/api/control/v1alpha1"
+)
+
+func nativeToRefVal(native any) ref.Val {
+	return celtypes.DefaultTypeAdapter.NativeToValue(native)
+}
+
+// scDataGroupingByList builds a statusCollectorData that groups by a single
+// expression whose evaluated value is a list (a composite, non-comparable
+// value). Each WEC's groupBy value is provided by wecToList.
+func scDataGroupingByList(wecToList map[string][]any) *statusCollectorData {
+	wecToData := map[string]*workStatusData{}
+	for wec, list := range wecToList {
+		wecToData[wec] = &workStatusData{
+			groupByEval:        rowFragment{"g": nativeToRefVal(list)},
+			combinedFieldsEval: rowFragment{"count": nil},
+		}
+	}
+	return &statusCollectorData{
+		collectorSpec: &v1alpha1.StatusCollectorSpec{
+			GroupBy: []v1alpha1.NamedExpression{{Name: "g", Def: "obj.spec.list"}},
+			CombinedFields: []v1alpha1.NamedAggregator{
+				{Name: "count", Type: v1alpha1.AggregatorTypeCount},
+			},
+		},
+		WECToData: wecToData,
+	}
+}
+
+// TestHandleAggregationGroupByComposite verifies that a groupBy value that is a
+// composite (list) type does not panic and that equal values are grouped
+// together while distinct values are grouped apart. Before the fix, the raw
+// composite value was used as a map key and this panicked with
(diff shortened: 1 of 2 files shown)


magic-peach:
## Review: Fix panic on groupBy expression returning a list or map

**Problem Analysis:** ✅ Well understood
- The root cause is clear: `map[any]int` panics on unhashable types like `[]interface{}` and `map[string]interface{}`
- The validation gap (type-checks OK but fails at runtime) is a good catch

**Solution:** ✅ Appropriate
- Reducing composite values to stable string keys via JSON encoding is a clean approach
- Maintains grouping semantics for scalar cases that worked before
- The JSON encoding approach is consistent with how `refValToValue` already handles composite values

**Test Quality:** ✅ Good
- Tests cover both list-valued groupBy and the key helper directly
- Regression test that panics without the fix is valuable

**Suggestions:**
1. Consider adding a comment explaining why JSON encoding was chosen over other serialization approaches (consistency with existing code)
2. The `keyForGroupByValue` function could use a brief doc comment explaining its purpose

Overall: **Strong approval**. This fixes a real crash bug with proper testing and a clean solution. 🎉

PR #3900, opened 2026-07-12
🐛 Fix recursive read-lock deadlock in compareCombinedStatus

### LINKED ISSUE #3899: bug: recursive read-lock in compareCombinedStatus can deadlock the status controller
### Describe the bug

`compareCombinedStatus` in `pkg/status/combinedstatus-resolution.go` takes a read lock on the resolution and then, while still holding it, calls `generateCombinedStatus`, which takes the same read lock again:

```go
func (c *combinedStatusResolution) compareCombinedStatus(...) *v1alpha1.CombinedStatus {
	c.RLock()
	defer c.RUnlock()

	localCombinedStatus := c.generateCombinedStatus(bindingName, sourceObjectIdentifier)
	...
}

func (c *combinedStatusResolution) generateCombinedStatus(...) *v1alpha1.CombinedStatus {
	c.RLock()
	defer c.RUnlock()
	...
}
```

Go's `sync.RWMutex` is not reentrant, and recursive read-locking is explicitly unsafe. From the [docs](https://pkg.go.dev/sync#RWMutex): "If a goroutine holds a RWMutex for reading and another goroutine might call Lock, no goroutine should expect to be able to acquire a read lock until the initial read lock is released." So if a writer calls `Lock()` on the same resolution in the window between the two `RLock()` calls, the writer blocks waiting for the first reader to release, and the second `RLock()` blocks waiting for the writer — a classic deadlock.

The same resolution takes a write lock in several places (`setCollectorsData`, `removeCollector`, the update/note paths around lines 155, 191, 238, 264, 366), and the status controller and binding controller share the resolver, so a concurrent write on the same resolution is possible.

### Steps To Reproduce

This is hard to trigger deterministically because it depends on a writer landing in the gap between the two read locks, so I haven't reproduced it live. The path is:

1. `syncCombinedStatus` → `combinedStatusResolver.CompareCombinedStatus` → `resolution.compareCombinedStatus` (first `RLock`) → `generateCombinedStatus` (second `RLock`).
2. Concurrently, something mutates the same resolution (e.g. a status collector update) and calls `Lock()` on it in between.

Under that interleaving the controller goroutine parks permanently and combined-status updates for that object stop.

### Expected Behavior

Generating the combined status while already holding the read lock should not risk a deadlock. Splitting the shared work into an unlocked helper that both `generateCombinedStatus` and `compareCombinedStatus` call (each acquiring the lock once at the top) would avoid the recursive acquisition.

### Additional Context

Relevant code: `pkg/status/combinedstatus-resolution.go`, `compareCombinedStatus` and `generateCombinedStatus`. This is separate from #3864, which is about lock ordering in `pkg/binding/bindingpolicy-resolution.go`.

santhil-cyber:
## Summary

`compareCombinedStatus` takes the resolution's read lock and, while holding it, calls `generateCombinedStatus`, which takes the same read lock again. `sync.RWMutex` is not reentrant — recursive read-locking is documented as unsafe. If a writer calls `Lock()` in the gap between the two `RLock()` calls, the writer blocks behind the first reader and the second `RLock()` blocks behind the writer, so the status controller goroutine parks permanently.

The fix pulls the body of `generateCombinedStatus` into an unlocked helper, `generateCombinedStatusReadLocked`, that assumes the caller already holds the lock. This matches the existing `handleSelectReadLocked` / `handleAggregationReadLocked` convention in the same file. `generateCombinedStatus` stays a thin wrapper that acquires the read lock once, and `compareCombinedStatus` calls the helper directly since it already holds the lock. No behavior change on the single-lock paths.

The regression test runs `compareCombinedStatus` in a loop while another goroutine contends for the same resolution's write lock. It deadlocks (times out) on the old code and passes on the fixed code; verified both ways locally, and `go test -race ./pkg/status/` is clean.

## Related issue(s)

Fixes #3899

### DIFF
--- a/pkg/status/combinedstatus-resolution.go
+++ b/pkg/status/combinedstatus-resolution.go
@@ -273,12 +273,22 @@ func (c *combinedStatusResolution) noteStatusCollectorAbsence(statusCollectorNam
 }
 
 // generateCombinedStatus calculates the combinedstatus from the statuscollector
-// data in the combinedstatus resolution.
+// data in the combinedstatus resolution. It acquires the read lock and delegates
+// to generateCombinedStatusReadLocked.
 func (c *combinedStatusResolution) generateCombinedStatus(bindingName string,
 	workloadObjectIdentifier util.ObjectIdentifier) *v1alpha1.CombinedStatus {
 	c.RLock()
 	defer c.RUnlock()
 
+	return c.generateCombinedStatusReadLocked(bindingName, workloadObjectIdentifier)
+}
+
+// generateCombinedStatusReadLocked calculates the combinedstatus from the
+// statuscollector data in the combinedstatus resolution. The caller must hold
+// at least the read lock, since sync.RWMutex is not reentrant and recursively
+// acquiring the read lock can deadlock if a writer is waiting.
+func (c *combinedStatusResolution) generateCombinedStatusReadLocked(bindingName string,
+	workloadObjectIdentifier util.ObjectIdentifier) *v1alpha1.CombinedStatus {
 	combinedStatus := &v1alpha1.CombinedStatus{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      c.Name,
@@ -318,7 +328,10 @@ func (c *combinedStatusResolution) compareCombinedStatus(status *v1alpha1.Combin
 	c.RLock()
 	defer c.RUnlock()
 
-	localCombinedStatus := c.generateCombinedStatus(bindingName, sourceObjectIdentifier)
+	// Use the read-locked helper: we already hold the read lock, and
+	// sync.RWMutex is not reentrant, so calling generateCombinedStatus (which
+	// re-acquires the read lock) here can deadlock if a writer is waiting.
+	localCombinedStatus := c.generateCombinedStatusReadLocked(bindingName, sourceObjectIdentifier)
 
 	// check labels
 	if !validateCombinedStatusLabels(status, bindingName, sourceObjectIdentifier) {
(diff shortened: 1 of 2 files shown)


santhil-cyber:
Thanks for the review. Both addressed in 199c6dc:

- Writer no longer spins tight — added a yield between Lock/Unlock cycles. Used `runtime.Gosched()` rather than `time.Sleep(0)`, since a zero duration returns immediately without yielding the goroutine.
- `stop` is now closed before `t.Fatal` on the timeout path so the writer doesn't keep running into later tests.

Re-checked that the test still reproduces the deadlock (fails on timeout with the recursive lock restored) and passes on the fix, race detector clean.

magic-peach:
## Review: Fix recursive read-lock deadlock in compareCombinedStatus

**Code Quality:** ✅ Excellent

The fix is clean and follows the existing pattern in the codebase (`handleSelectReadLocked` / `handleAggregationReadLocked`):
- Creating `generateCombinedStatusReadLocked` as an unlocked helper is the right approach
- The caller-inherits-lock pattern is well-documented and consistent with the rest of the file
- No behavior change on single-lock paths

**Test Quality:** ✅ Solid regression test
- The concurrent writer goroutine with `runtime.Gosched()` creates realistic contention
- 15-second timeout is generous enough to avoid flakiness
- Properly cleans up goroutines on both success and failure paths

**Suggestions:**
1. The comment on `generateCombinedStatusReadLocked` says "The caller must hold at least the read lock" - consider also adding a `//go:nosplit` or a runtime check in debug mode, though this is optional
2. Consider adding a brief comment in the test explaining why 2000 iterations is sufficient (to prevent flakiness concerns)

Overall: **Strong approval**. This is a real concurrency bug fix with proper testing. 🎉

PR #3915, opened 2026-07-17
🐛 fix(crd): fix timeout context bug during CRD initialization

Norway-02:
## Summary
This PR fixes a distinct logic bug during the initialization of KubeStellar Control CRDs (`pkg/crd/crds.go`):

**Timeout Context Bug**: The `ApplyCRDs` function creates a `ctxLimited` context with a 1-minute timeout, but the `Create`, `Get`, and `Update` API calls mistakenly used the parent `ctx`. This completely circumvented the intended timeout, potentially causing the controller manager to hang indefinitely if the API server was unresponsive.

## Related issue(s)
N/A - bug fixes

### DIFF
--- a/pkg/crd/crds.go
+++ b/pkg/crd/crds.go
@@ -81,13 +81,13 @@ func ApplyCRDs(ctx context.Context, controllerName string,
 		if err != nil {
 			return fmt.Errorf("unable to convert from Unstructured to CRD, name=%s: %w", crdU.GetName(), err)
 		}
-		_, err = clientsetExt.Create(ctx, desiredCRD, metav1.CreateOptions{FieldValidation: metav1.FieldValidationStrict, FieldManager: controllerName})
+		_, err = clientsetExt.Create(ctxLimited, desiredCRD, metav1.CreateOptions{FieldValidation: metav1.FieldValidationStrict, FieldManager: controllerName})
 		if err == nil {
 			logger.Info("Created CRD", "name", desiredCRD.Name)
 		} else if !apierrors.IsAlreadyExists(err) {
 			return fmt.Errorf("unable to create CRD named %s: %w", crdU.GetName(), err)
 		} else {
-			existingCRD, err := clientsetExt.Get(ctx, desiredCRD.Name, metav1.GetOptions{})
+			existingCRD, err := clientsetExt.Get(ctxLimited, desiredCRD.Name, metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


Norway-02:
Hi @waltforme and @MikeSpreitzer, this PR fixes a context timeout issue and a file descriptor leak in the CRD initialization logic. Could you please take a look when you have a moment? Correctness and reviewability are my top priorities here. Thank you!

waltforme:
I wonder whether '2. File Descriptor Leak' is a true bug, because files returned by `embed.FS.Open` are backed by in-memory data, not OS file descriptors.

Norway-02:
Hi @waltforme, you are absolutely right! Since `embed.FS` is backed entirely by in-memory data, `Open` does not consume OS file descriptors, and `Close()` is essentially a no-op. I've removed that change completely from this PR so it can strictly focus on the `ApplyCRDs` context timeout bug (which is the actual high-impact issue here). Thanks for catching that!

magic-peach:
Good catch. Using `ctxLimited` instead of `ctx` ensures the CRD operations respect the timeout. Without this, a slow or hung CRD create/update could block indefinitely since the parent context might not have a deadline.

Norway-02:

magic-peach:
**Code Review: LGTM** ✅

Using `ctxLimited` instead of the parent `ctx` for CRD Create/Get/Update operations ensures that the timeout set for CRD initialization is respected. Previously, using the unbounded `ctx` meant these API calls could hang indefinitely if the API server was slow, bypassing the timeout that `ctxLimited` was specifically created for.

The changes are minimal and target exactly the three API calls that should be bounded. No concerns.

Norway-02:

magic-peach (commented):
The fix looks correct: `Create`/`Get`/`Update` now honor the same `ctxLimited` deadline that `waitForCRDAccepted` already used, closing the gap where a stalled API server could hang startup indefinitely.

One thing worth being aware of (not introduced by this PR, but visible now): `ctxLimited` is a single 1-minute budget created once for the whole function, shared across all 5 CRDs in the loop *and* their `waitForCRDAccepted` calls. If an early CRD is slow, later CRDs inherit a mostly-expired deadline. That's arguably the intent of "apply in under a minute total", but if the goal is a per-CRD timeout, the `context.WithTimeout` would need to move inside the loop. No change required here either way.

PR #3916, opened 2026-07-17
🐛 fix(status): prevent status-controller from silently swallowing startup errors

Norway-02:
## Summary
This PR fixes a bug in `pkg/status/status-controller.go` where initialization errors could be silently swallowed.

In `Controller.Start`, the main controller loop (`c.run`) is executed inside a goroutine, and any error is sent to a buffered `errChan`. The parent function waits up to 3 seconds for an error during startup. If `c.run` fails *after* 3 seconds, the error is sent to the channel but is never handled or logged, resulting in a silent failure. This PR adds a check inside the goroutine to log any error returned by `c.run` before sending it to the channel, ensuring delayed initialization failures are captured in the logs.

## Related issue(s)
N/A - bug fix

### DIFF
--- a/pkg/status/status-controller.go
+++ b/pkg/status/status-controller.go
@@ -202,7 +202,11 @@ func (c *Controller) Start(parentCtx context.Context, workers int, cListers chan
 	ctx := klog.NewContext(parentCtx, logger)
 	errChan := make(chan error, 1)
 	go func() {
-		errChan <- c.run(ctx, workers, cListers)
+		err := c.run(ctx, workers, cListers)
+		if err != nil {
+			logger.Error(err, "status-controller failed")
+		}
+		errChan <- err
 	}()
 
 	// check for errors at startup, after all started we let it continue


Norway-02:
Hi @waltforme and @MikeSpreitzer, this PR fixes a silent startup failure bug in the status-controller initialization logic. I wanted to make sure these silent failures are captured in the logs! Could you please take a look when you get a chance? Thank you!

magic-peach:
Good improvement. Silently swallowing the error from `c.run()` made it impossible to diagnose status-controller startup failures. Logging it before sending to the channel ensures operators see the error in logs regardless of whether anyone reads from errChan.

Norway-02:

magic-peach:
**Code Review: LGTM** ✅

Simple and correct. Currently when `c.run()` returns an error, the error is sent to `errChan` but never logged. This makes debugging startup/runtime failures difficult since the controller silently exits. The fix adds `logger.Error(err, "status-controller failed")` before sending to the channel.

No concerns — the error is still sent to `errChan` so the caller can handle it as before. This is purely additive logging.

Norway-02:

Issue #3914, opened 2026-07-16
Replace panic() with proper error handling in HTTP server startup (pkg/controller/run.go)  Description:


karthik120710:
The Start() function in run.go uses panic() when HTTP servers fail to start (lines 51-74). This is problematic for production environments.

Current Behavior
Three HTTP servers are started in goroutines:

Health probe server 
Prometheus metrics server 
pprof debug server 
When any of these servers fail to start (e.g., port already in use, permission denied), the code calls panic(err), which causes the entire process to crash immediately.

PR #3917, opened 2026-07-17
Replace panic() with proper error handling in HTTP server startup

karthik120710:
This PR replaces panic() calls with proper error handling in the HTTP server startup code in run.go. The previous implementation would crash the entire process when HTTP servers failed to start, which is not suitable for production environments.

## Problem
The Start() function used panic() when HTTP servers (health probe, metrics, pprof) failed to start. This caused immediate process crashes without:

1.Graceful shutdown
2.Error propagation to callers
3.Opportunity for error recovery
4.Proper cleanup of resources

## Changes
Modified run.go:
1.Changed Start() signature to return error instead of void
2.Replaced http.ListenAndServe() with http.Server instances for proper lifecycle management
3.Added error channel to communicate goroutine errors to the caller
4.Implemented graceful shutdown on context cancellation with 5-second timeout
5.Replaced all panic(err) calls with proper error propagation
6.Added informative logging for server startup and shutdown
Updated main.go:
1/Modified the call to ksctlr.Start() to handle the returned error
2.Added error logging and graceful exit if HTTP servers fail to start

#3914

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -115,10 +115,12 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := ksctlr.Start(ctx, processOpts); err != nil {
-		setupLog.Error(err, "error starting HTTP servers")
-		os.Exit(1)
-	}
+	go func() {
+		if err := ksctlr.Start(ctx, processOpts); err != nil {
+			setupLog.Error(err, "error running HTTP servers")
+			os.Exit(1)
+		}
+	}()
 
 	spacesClientMetrics := ksmetrics.NewMultiSpaceClientMetrics()
 	ksmetrics.MustRegister(legacyregistry.Register, spacesClientMetrics)
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -115,7 +115,10 @@ func main() {
 		os.Exit(1)
 	}
 
-	ksctlr.Start(ctx, processOpts)
+	if err := ksctlr.Start(ctx, processOpts); err != nil {
+		setupLog.Error(err, "error starting HTTP servers")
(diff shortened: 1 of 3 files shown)


magic-peach:
**Code Review: Good direction, some concerns**

The overall goal of replacing `panic()` with proper error handling and graceful shutdown is correct. However, there are a few concerns worth addressing:

**Concerns:**

1. **Duplicate shutdown paths**: There are now two goroutines that both handle `ctx.Done()` — one in the "handle graceful shutdown" block and one in the final `select`. This means when the context is cancelled, both goroutines race to shut down the same servers. Consider consolidating into a single shutdown path.

2. **`Start` now runs servers concurrently**: The original code called `ListenAndServe` synchronously in goroutines. The new code runs them in goroutines too, but `Start()` itself returns after the first error or context cancellation. Since `cmd/controller-manager/main.go` now wraps `Start` in a goroutine and calls `os.Exit(1)` on error, the main function continues with `spacesClientMetrics` setup even while `Start` may be failing. This changes the startup ordering semantics.

3. **5-second shutdown timeout**: The hardcoded `5*time.Second` timeout for graceful shutdown may be too short for in-flight requests. Consider making it configurable or using the parent context's deadline if one is set.

4. **The `errChan` is buffered to 3 but there are 3 servers**: This is correct, but if all 3 servers fail simultaneously, only the first error is returned. Consider using a loop to drain remaining errors for logging.

**Positive:**
- Replacing bare `panic()` in library code (`pkg/controller/run.go`) with returned errors is the right pattern
- Graceful shutdown with `server.Shutdown` is correct
- Error wrapping with `fmt.Errorf` provides good context

Overall this is a solid improvement. The concerns above are about edge cases rather than correctness issues.

karthik120710:

PR #3919, opened 2026-07-18
🐛 fix(transport): remove fmt.Sprintf anti-pattern from structured logger

Norway-02:
## Summary
This PR fixes a structured logging anti-pattern in the generic transport controller (`pkg/transport/generic/generic_transport_controller.go`).

Currently, an error log on unwrapping failure uses `fmt.Sprintf` directly inside the log message:
`logger.Error(err, fmt.Sprintf("Failed to unwrap %#v", currentWrappedObject))`

This dynamic string defeats the purpose of structured logging (such as `logr` or `klog`), making log messages impossible to index or query efficiently in downstream aggregators. This PR fixes the call by passing the object as a structured key-value pair instead:
`logger.Error(err, "Failed to unwrap", "object", currentWrappedObject)`

## Related issue(s)
N/A - bug fix

### DIFF
--- a/pkg/transport/generic/generic_transport_controller.go
+++ b/pkg/transport/generic/generic_transport_controller.go
@@ -1083,7 +1083,7 @@ func (c *genericTransportController) propagateWrappedObjectToClusters(ctx contex
 			} else {
 				gloss, err := c.transport.UnwrapObjects(currentWrappedObject, kindToResource)
 				if err != nil {
-					logger.Error(err, fmt.Sprintf("Failed to unwrap %#v", currentWrappedObject))
+					logger.Error(err, "Failed to unwrap", "object", currentWrappedObject)
 				}
 				desiredGeneration := task.ObjU.GetAnnotations()[originOwnerGenerationAnnotation]
 				actualGeneration := currentWrappedObject.GetAnnotations()[originOwnerGenerationAnnotation]


Norway-02:
Hi @waltforme, this PR fixes a structured logging anti-pattern in the transport controller where `fmt.Sprintf` was being passed directly as the log message, preventing log indexers from parsing the event. Could you please take a look? Thank you!

magic-peach:
Nice improvement. The structured logger form is better for log aggregation and avoids the overhead of `fmt.Sprintf` on every unwrap failure. The key-value pair format is consistent with the rest of the codebase.

Norway-02:

magic-peach:
**Code Review: LGTM** ✅

Good fix. Using `fmt.Sprintf("Failed to unwrap %#v", currentWrappedObject)` in a logger call defeats structured logging — the entire object is serialized into the format string rather than being passed as a structured key-value pair. The fix correctly uses `logger.Error(err, "Failed to unwrap", "object", currentWrappedObject)` which:

1. Preserves structured logging (the object can be inspected in log aggregation tools)
2. Avoids unnecessary string interpolation when the log level might be suppressed
3. Follows the `klog`/`logr` convention of alternating key-value pairs

No concerns.

Norway-02:

Norway-02:

PR #3920, opened 2026-07-19
🐛 fix(status): handle json.Unmarshal errors in TypeObject equality

Norway-02:
## Summary

This PR adds error checking for `json.Unmarshal` calls within the `valueEqual` function in `combinedstatus-resolution.go`. Previously, unhandled unmarshal errors could lead to inaccurate evaluation of `TypeObject` equality by mistakenly comparing zero values. 

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/status/combinedstatus-resolution.go
+++ b/pkg/status/combinedstatus-resolution.go
@@ -1154,8 +1154,12 @@ func valueEqual(a, b *v1alpha1.Value) bool {
 		return *a.Bool == *b.Bool
 	case v1alpha1.TypeObject:
 		var v1, v2 interface{}
-		json.Unmarshal([]byte(a.Object.Raw), &v1)
-		json.Unmarshal([]byte(b.Object.Raw), &v2)
+		if err := json.Unmarshal([]byte(a.Object.Raw), &v1); err != nil {
+			return false
+		}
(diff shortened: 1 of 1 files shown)


Norway-02:
Hi @waltforme, this PR adds proper error checking for `json.Unmarshal` calls when evaluating `TypeObject` equality in status resolution. This prevents silent fallbacks to comparing zero-values if the decoding fails. Could you please take a look when you have a chance? Thank you!

magic-peach:
Good catch. Ignoring `json.Unmarshal` errors could lead to silent comparison failures where `reflect.DeepEqual` compares a nil against a populated value. Returning `false` on unmarshal error is the correct behavior here since unequal serialization implies unequal values.

Norway-02:

magic-peach:
**Code Review: LGTM** ✅

Good fix. The original code silently ignored `json.Unmarshal` errors, which could lead to incorrect equality comparisons (comparing partially parsed or zero-value data). Returning `false` on unmarshal failure is the safe default — two values with unparseable JSON should not be considered equal.

**Note**: PR #3921 includes this same change along with the `errors.New` fix in `cmd/controller-manager/main.go`. These two PRs have overlapping changes. Consider coordinating to avoid merge conflicts — one of them will need to be rebased after the other merges.

Norway-02:

Norway-02:

PR #3921, opened 2026-07-19
🐛 fix(cmd): replace formatless fmt.Errorf with errors.New

Norway-02:
## Summary

This PR replaces a formatless `fmt.Errorf` call with `errors.New` in the controller-manager setup logs. Using `fmt.Errorf` without formatting verbs incurs unnecessary parsing overhead and is flagged as an anti-pattern by Go linters (`go vet` / `staticcheck`). 

## Related issue(s)

Fixes #

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -20,8 +20,8 @@ package main
 // to ensure that exec-entrypoint and run can make use of them.
 
 import (
+	"errors"
 	"flag"
-	"fmt"
 	"os"
 	"strings"
 	"time"
@@ -111,7 +111,7 @@ func main() {
 		strings.ToLower(binding.ControllerName),
 		strings.ToLower(status.ControllerName),
 	).IsSuperset(ctlrsToStart) {
-		setupLog.Error(fmt.Errorf("unknown controller specified"), "'controllers' flag has incorrect value")
+		setupLog.Error(errors.New("unknown controller specified"), "'controllers' flag has incorrect value")
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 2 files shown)


Norway-02:
Hi @waltforme, this PR replaces a formatless `fmt.Errorf` with `errors.New` in the controller-manager setup logs, fixing a minor anti-pattern. Could you please take a look when you have a moment? Thank you!

magic-peach:
**Code Review: Good changes, one minor suggestion**

Two independent fixes in this PR:

**1. `errors.New` vs `fmt.Errorf` (cmd/controller-manager/main.go)** ✅

Correct — when no format verbs are used, `errors.New` is preferred over `fmt.Errorf` because it avoids the overhead of the format parser and makes intent clearer. Clean change.

**2. `json.Unmarshal` error handling (pkg/status/combinedstatus-resolution.go)** ✅

Good catch. Silently ignoring unmarshal errors in `valueEqual` could lead to incorrect equality comparisons (e.g., returning `false` when both values are equally broken). Now it returns `false` explicitly when unmarshal fails, which is the safe default — two values with unparseable JSON should not be considered equal.

**Suggestion**: The `case v1alpha1.TypeObject:` block now has two error checks that both `return false`. Consider wrapping them to make the logic clearer:

```go
case v1alpha1.TypeObject:
    var v1, v2 interface{}
    if err := json.Unmarshal([]byte(a.Object.Raw), &v1); err != nil ||
       err := json.Unmarshal([]byte(b.Object.Raw), &v2); err != nil {
        return false
    }
    return reflect.DeepEqual(v1, v2)
```

This is a stylistic suggestion — the current code is correct as-is.

Norway-02:

Norway-02:

PR #3924, opened 2026-07-20
🐛 fix(binding): prevent panic from unsafe type assertions in binding-controller.go

Norway-02:
## Summary

This PR prevents potential nil pointer and interface conversion panics by replacing unsafe type assertions (e.g., `obj.(mrObject)` and `obj.(metav1.Object)`) with their safe, two-value equivalents (the comma-ok idiom) in `pkg/binding/binding-controller.go`. If an unexpected or nil object is encountered, it gracefully handles it instead of crashing the controller manager.

## Related issue(s)

Fixes #

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -1,0 +1,1 @@
+defaultBaseImage: gcr.io/distroless/static:nonroot
--- a/pkg/binding/binding-controller.go
+++ b/pkg/binding/binding-controller.go
@@ -655,8 +655,11 @@ func verbsSupportInformers(verbs []string) bool {
 // At this time it is very simple, more complex processing might be required
 // here.
 func (c *Controller) handleObject(oldObj, obj any, resource string, eventType WorkloadEventType, wasDeletedFinalStateUnknown bool) {
-	objMR := obj.(mrObject)
-	oldObjMR := oldObj.(mrObject)
+	objMR, ok := obj.(mrObject)
+	if !ok {
+		return
+	}
+	oldObjMR, _ := oldObj.(mrObject)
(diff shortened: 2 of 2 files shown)


Norway-02:
Hi @waltforme, this PR prevents potential panics in `binding-controller.go` by using the safe comma-ok idiom for type assertions, similar to recent fixes in other packages. Could you please take a look when you have a moment? Thank you!

magic-peach:
Good defensive changes. The silent return on obj assertion failure is appropriate here since handleObject is called from a watch handler and a panic would crash the controller. The oldObj using `_, _` is fine since it can legitimately be nil on add events.

Norway-02:

magic-peach:
**Code Review: Good, with one concern**

The PR replaces unsafe type assertions with safe comma-ok patterns in the binding controller. Two observations:

**Good:**
- `isBeingDeleted` returning `false` when the type assertion fails is the safe default.
- `handleObject` returning early when `obj.(mrObject)` fails prevents a crash.

**Concern:**
- `oldObjMR, _ := oldObj.(mrObject)` discards the ok value. If `oldObj` doesn't implement `mrObject`, `oldObjMR` will be nil, which could cause a nil pointer dereference later in the function. For example, if any code after this line uses `oldObjMR` without nil checks, it will panic.

Could you verify that `oldObjMR` is nil-safe at all usage points? If not, consider adding a nil check or returning early when `oldObj` is not an `mrObject` (similar to how `obj` is handled). For example:

```go
oldObjMR, ok := oldObj.(mrObject)
if !ok {
    return
}
```

The rest of the changes are clean.

Norway-02:

Norway-02:

PR #3925, opened 2026-07-20
🐛 fix(binding): prevent panic from unsafe type assertions in selectors.go

Norway-02:
## Summary

This PR replaces an unsafe type assertion (`obj.(mrObject)`) with the safe comma-ok idiom in `pkg/binding/selectors.go`. This change prevents potential interface conversion panics during object reconciliation if the informer yields an unexpected type, aligning the error handling pattern with other recently merged fixes.

## Related issue(s)

Fixes #

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -1,0 +1,1 @@
+defaultBaseImage: gcr.io/distroless/static:nonroot
--- a/pkg/binding/selectors.go
+++ b/pkg/binding/selectors.go
@@ -50,7 +50,10 @@ func (c *Controller) updateResolutions(ctx context.Context, objIdentifier util.O
 		return fmt.Errorf("failed to get runtime.Object from object identifier (%v): %w", objIdentifier, err)
 	}
 
-	objMR := obj.(mrObject)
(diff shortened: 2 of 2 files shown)


Norway-02:
Hi @waltforme, this PR prevents potential interface conversion panics in `selectors.go` by switching to safe type assertions. Could you please take a look when you have a moment? Thank you!

magic-peach:
Clean fix. The unsafe assertion would panic if an object from the informer cache ever failed to implement mrObject. The comma-ok pattern with an error return is the standard approach here.

Norway-02:

magic-peach:
**Code Review: LGTM** ✅

This is a clean, minimal fix. The unsafe type assertion `obj.(mrObject)` is replaced with the safe comma-ok idiom, returning an error instead of panicking if the object doesn't implement `mrObject`. This is the correct pattern for handling interface conversions where the type may vary.

The error message is clear and includes the identifier for debugging. No concerns with this change.

Norway-02:

Norway-02:

PR #3929, opened 2026-07-21
🐛 fix(util): prevent panic from unsafe metav1.Object assertion

Norway-02:
## Summary

This PR hardens the object identifier utility against panics caused by unconditionally casting arbitrary runtime objects to `metav1.Object`.

**Changes:**
- Update `RefToRuntimeObj` in `pkg/util/gvk-obj-ref.go` to use the safe comma-ok idiom before accessing `GetName()` and `GetNamespace()`, preventing a crash if a non-standard runtime object is processed.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/util/gvk-obj-ref.go
+++ b/pkg/util/gvk-obj-ref.go
@@ -35,12 +35,16 @@ func (ref GKObjRef) String() string {
 // RefToRuntimeObj creates a GVKObjRef to a runtime.Object.
 func RefToRuntimeObj(obj runtime.Object) GKObjRef {
 	gvk := obj.GetObjectKind().GroupVersionKind()
-	mObj := obj.(metav1.Object)
+	var namespace, name string
+	if mObj, ok := obj.(metav1.Object); ok {
+		namespace = mObj.GetNamespace()
+		name = mObj.GetName()
+	}
 	ans := GKObjRef{
 		GK: gvk.GroupKind(),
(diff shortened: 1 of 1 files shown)


Norway-02:
Hi @waltforme @MikeSpreitzer , this is another highly focused `size/XS` bug fix for a panic-inducing type assertion. Could you please take a look?

magic-peach:
Good approach. Defaulting to empty strings when the object doesn't implement metav1.Object is reasonable since the function's primary purpose is building a GVK reference, and the klog.ObjectRef is secondary. The zero values are safe for logging.

Norway-02:

magic-peach:
**Code Review: LGTM** ✅

Clean fix for the unsafe `obj.(metav1.Object)` assertion. The change gracefully handles objects that don't implement `metav1.Object` by falling back to empty namespace/name strings instead of panicking.

**One suggestion**: The default empty strings could mask bugs if the caller expects valid namespace/name. Consider logging a debug message when the type assertion fails, similar to PR #3937:

```go
if mObj, ok := obj.(metav1.Object); ok {
    namespace = mObj.GetNamespace()
    name = mObj.GetName()
} else {
    // Log at a low level for debugging
}
```

This is a minor suggestion — the current code is correct and prevents a crash.

Norway-02:

PR #3930, opened 2026-07-21
🐛 fix(ci): quote tag name in goreleaser chart step

waltforme:
## Summary

Pass `github.ref_name` through an env var and strip the leading `v` with shell parameter expansion, instead of interpolating it directly into the `run:` block. Safer handling of the tag value; no behavior change for valid `vX.Y.Z` tags.


MikeSpreitzer (commented):
/lgtm
/approve
Thanks!

PR #3936, opened 2026-07-22
🐛 fix(status): replace formatless fmt.Errorf with errors.New in statuscollector

Norway-02:
## Summary

Two calls to `fmt.Errorf` in `validateStatusCollector` use static strings with no format verbs. The idiomatic Go replacement is `errors.New`, which avoids the unnecessary overhead of `fmt.Errorf` parsing.

**Changes:**
- Replace `fmt.Errorf("...")` with `errors.New("...")` for two static error strings in `pkg/status/statuscollector.go`
- Alias the `k8s.io/apimachinery/pkg/api/errors` import to `apierrors` to avoid collision with the newly-added stdlib `errors` package

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/status/statuscollector.go
+++ b/pkg/status/statuscollector.go
@@ -18,9 +18,10 @@ package status
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 
@@ -35,7 +36,7 @@ func (c *Controller) syncStatusCollector(ctx context.Context, ref string) error 
 	statusCollector, err := c.statusCollectorLister.Get(ref)
 	if err != nil {
 		// The resource no longer exist, which means it has been deleted.
-		if !errors.IsNotFound(err) {
+		if !apierrors.IsNotFound(err) {
 			return err
 		}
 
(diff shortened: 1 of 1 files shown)


Norway-02:
Hi @waltforme @MikeSpreitzer, this is a clean lint fix — two `fmt.Errorf` calls with no format verbs replaced by `errors.New`. Could you please review?

magic-peach:
Clean change. Using `errors.New` for static strings is the right call, and the `apierrors` alias for the k8s package makes the import distinction clear.

Norway-02:

magic-peach:
**Code Review: LGTM** ✅

Two good improvements:

1. **Import rename**: Renaming `k8s.io/apimachinery/pkg/api/errors` to `apierrors` prevents shadowing the standard library `errors` package. This is important because the file now uses `errors.New` from the standard library. Without the rename, `errors.IsNotFound` would have been ambiguous. Good catch.

2. **`errors.New` vs `fmt.Errorf`**: The two `fmt.Errorf` calls with no format verbs are correctly replaced with `errors.New`. This avoids unnecessary format string parsing and makes the intent clearer.

No concerns — both changes are minimal and correct.

Norway-02:

pkg/status/statuscollector.go
```diff
@@ -18,9 +18,10 @@ package status
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
MikeSpreitzer, line 24: +1

MikeSpreitzer (commented):
This is an improvement.
FYI, there are higher priorities.
One that sounds simpler than it is: move up to a higher version of Kubernetes.

/lgtm
/approve

PR #3937, opened 2026-07-22
🐛 fix(status): prevent panic from unsafe runtime.Object assertion in resolver

Norway-02:
## Summary

The `evaluateWorkStatusesPerBindingReadLocked` function unconditionally casts `objs[0]` to `runtime.Object` without checking. This is a defensive hardening to use the safe comma-ok idiom instead.

**Changes:**
- Replace the unsafe `objs[0].(runtime.Object)` with a safe comma-ok assertion in `pkg/status/combinedstatus-resolver.go`
- On failure, log the error via `utilruntime.HandleError` and `continue` to the next iteration, matching the existing error-handling pattern in the same block

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/status/combinedstatus-resolver.go
+++ b/pkg/status/combinedstatus-resolver.go
@@ -24,7 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	machtypes "k8s.io/apimachinery/pkg/types"
-	runtime2 "k8s.io/apimachinery/pkg/util/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
@@ -485,7 +485,7 @@ func (c *combinedStatusResolver) evaluateWorkStatusesPerBindingReadLocked(ctx co
 
 			objs, err := workStatusIndexer.ByIndex(workStatusIdentificationIndexKey, indexKey) // one obj expected
 			if err != nil {
-				runtime2.HandleError(fmt.Errorf("failed to get workstatus with indexKey %s: %w", indexKey, err))
+				utilruntime.HandleErrorWithContext(ctx, err, "failed to get workstatus", "indexKey", indexKey)
 				continue
 			}
 			var workStat *workStatus
@@ -508,10 +508,28 @@ func (c *combinedStatusResolver) evaluateWorkStatusesPerBindingReadLocked(ctx co
 					logger.V(3).Info("Found more than one WorkStatus object, using the first", "binding", bindingName,
 						"workloadObjIdentifier", workloadObjIdentifier, "destination", destination)
 				}
-				workStat, err = runtimeObjectToWorkStatus(objs[0].(runtime.Object))
-				if err != nil {
-					runtime2.HandleError(fmt.Errorf("failed to convert runtime.Object to workStatus: %w", err))
-					continue
+				runtimeObj, ok := objs[0].(runtime.Object)
+				if !ok {
+					utilruntime.HandleErrorWithContext(ctx, nil, "value does not implement runtime.Object", "type", fmt.Sprintf("%T", objs[0]))
+					workStat = &workStatus{
+						workStatusRef: workStatusRef{
+							Name:                   "",
+							WECName:                destination,
+							SourceObjectIdentifier: workloadObjIdentifier,
+						},
+					}
+				} else {
+					workStat, err = runtimeObjectToWorkStatus(runtimeObj)
+					if err != nil {
+						utilruntime.HandleErrorWithContext(ctx, err, "failed to convert runtime.Object to workStatus")
+						workStat = &workStatus{
+							workStatusRef: workStatusRef{
+								Name:                   "",
+								WECName:                destination,
+								SourceObjectIdentifier: workloadObjIdentifier,
+							},
+						}
+					}
 				}
 			}
 


Norway-02:
Hi @waltforme @MikeSpreitzer, this is a focused panic-prevention fix — replacing an unsafe type assertion with the safe comma-ok idiom. Could you please review?

pkg/status/combinedstatus-resolver.go
```diff
@@ -508,7 +508,12 @@ func (c *combinedStatusResolver) evaluateWorkStatusesPerBindingReadLocked(ctx co
 					logger.V(3).Info("Found more than one WorkStatus object, using the first", "binding", bindingName,
 						"workloadObjIdentifier", workloadObjIdentifier, "destination", destination)
 				}
-				workStat, err = runtimeObjectToWorkStatus(objs[0].(runtime.Object))
+				runtimeObj, ok := objs[0].(runtime.Object)
+				if !ok {
+					runtime2.HandleError(fmt.Errorf("object %T does not implement runtime.Object", objs[0]))
+					continue
+				}
```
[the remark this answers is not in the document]
MikeSpreitzer: The cache will not do that. See https://github.com/kubernetes/client-go/blob/v0.32.13/tools/cache/index.go#L26-L28

pkg/status/combinedstatus-resolver.go
```diff
@@ -508,7 +508,12 @@ func (c *combinedStatusResolver) evaluateWorkStatusesPerBindingReadLocked(ctx co
 					logger.V(3).Info("Found more than one WorkStatus object, using the first", "binding", bindingName,
 						"workloadObjIdentifier", workloadObjIdentifier, "destination", destination)
 				}
-				workStat, err = runtimeObjectToWorkStatus(objs[0].(runtime.Object))
+				runtimeObj, ok := objs[0].(runtime.Object)
+				if !ok {
+					runtime2.HandleError(fmt.Errorf("object %T does not implement runtime.Object", objs[0]))
```
MikeSpreitzer: `object %T` is inconsistent: `%T` prints a type, not an object.

Norway-02:
Thanks for the feedback @MikeSpreitzer — good catch on the inconsistent error message, I've updated it from `"object %T"` to `"value of type %T"` since `%T` prints a type. You're also right that the Indexer only holds actual objects per the Store contract, so tombstones won't appear here — the safe assertion is purely defensive. Force-pushed with both fixes.

magic-peach:
Looks good! The comma-ok assertion prevents the panic cleanly. One minor thought: the `continue` on line 513 skips the entire workload object for this destination. If this happens frequently, it might be worth logging at a higher verbosity level so operators can detect data loss. But the core fix is correct and important.

Norway-02:
Thanks @magic-peach! That's a great point about data loss visibility. I've added a `logger.V(3).Info` log right before the `continue` so operators can detect when a workload object is skipped due to the type assertion failure. Force-pushed the update!

magic-peach:
**Code Review: LGTM** ✅

Clean fix for the unsafe type assertion `objs[0].(runtime.Object)`. The safe comma-ok idiom is used correctly with:
- `runtime2.HandleError` for error reporting (consistent with the existing pattern in the codebase)
- A V(3) info log for debugging context
- `continue` to skip the problematic object instead of panicking

This prevents a crash when an informer yields a value that doesn't implement `runtime.Object`. No concerns.

pkg/status/combinedstatus-resolver.go
```diff
@@ -508,7 +508,13 @@ func (c *combinedStatusResolver) evaluateWorkStatusesPerBindingReadLocked(ctx co
 					logger.V(3).Info("Found more than one WorkStatus object, using the first", "binding", bindingName,
 						"workloadObjIdentifier", workloadObjIdentifier, "destination", destination)
 				}
-				workStat, err = runtimeObjectToWorkStatus(objs[0].(runtime.Object))
+				runtimeObj, ok := objs[0].(runtime.Object)
+				if !ok {
+					runtime2.HandleError(fmt.Errorf("value of type %T does not implement runtime.Object", objs[0]))
+					logger.V(3).Info("Skipping workload object for destination due to type assertion failure", "binding", bindingName, "destination", destination, "workloadObjIdentifier", workloadObjIdentifier)
```
MikeSpreitzer: Logging _in addition to_ `runtime2.HandleError` is overkill. Particularly for code that will never be reached.

pkg/status/combinedstatus-resolver.go
```diff
@@ -508,7 +508,13 @@ func (c *combinedStatusResolver) evaluateWorkStatusesPerBindingReadLocked(ctx co
 					logger.V(3).Info("Found more than one WorkStatus object, using the first", "binding", bindingName,
 						"workloadObjIdentifier", workloadObjIdentifier, "destination", destination)
 				}
-				workStat, err = runtimeObjectToWorkStatus(objs[0].(runtime.Object))
+				runtimeObj, ok := objs[0].(runtime.Object)
+				if !ok {
+					runtime2.HandleError(fmt.Errorf("value of type %T does not implement runtime.Object", objs[0]))
```
MikeSpreitzer: If we're going to pretend this code could be reached, then it should use `HandleErrorWithContext`.

Norway-02:
Thank you for the feedback. I have updated the PR to use `HandleErrorWithContext` and removed the redundant logging as requested. Also, thank you for merging the two PRs recently.

pkg/status/combinedstatus-resolver.go
MikeSpreitzer: BTW, "2" is not a very informative qualifier. I have used "utilruntime" in other files to distinguish this "runtime" from others.

Norway-02: Updated. The import qualifier has been changed to `utilruntime`.

pkg/status/combinedstatus-resolver.go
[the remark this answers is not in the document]
MikeSpreitzer: But I agree with the thrust of Copilot's comment here. If we are going to pretend that `objs[0]` might not implement `runtime.Object` then the better response is the same behavior that happens `if len(objs) == 0`. And the same critique applies to the pre-existing `continue` just below for the case of failure in producing `workStat`.

pkg/status/combinedstatus-resolver.go
```diff
@@ -508,9 +508,14 @@ func (c *combinedStatusResolver) evaluateWorkStatusesPerBindingReadLocked(ctx co
...
-				workStat, err = runtimeObjectToWorkStatus(objs[0].(runtime.Object))
+				runtimeObj, ok := objs[0].(runtime.Object)
+				if !ok {
+					utilruntime.HandleErrorWithContext(ctx, nil, "value does not implement runtime.Object", "type", fmt.Sprintf("%T", objs[0]))
+					continue
+				}
+				workStat, err = runtimeObjectToWorkStatus(runtimeObj)
 				if err != nil {
-					runtime2.HandleError(fmt.Errorf("failed to convert runtime.Object to workStatus: %w", err))
+					utilruntime.HandleError(fmt.Errorf("failed to convert runtime.Object to workStatus: %w", err))
```
MikeSpreitzer: _HandleErrorWithContext_ here too.

pkg/status/combinedstatus-resolver.go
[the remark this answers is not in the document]
Norway-02: I have updated the logic. Now, if the type assertion fails or the conversion to workStatus fails, the code will log the error with `HandleErrorWithContext` and properly fall back to constructing the blank `workStatus`, matching the behavior of `len(objs) == 0` instead of skipping.

pkg/status/combinedstatus-resolver.go
Norway-02: Done.

Norway-02:
Hi @MikeSpreitzer, I've updated the PR with the latest fixes for `HandleErrorWithContext` and the `workStatus` fallback logic per your feedback. Let me know if you need any further changes!

MikeSpreitzer (commented):
/lgtm
/approve

PR #3948, opened 2026-07-25
🌱 chore: remove unused code detected by staticcheck (U1000/SA4006)

Norway-02:
## Summary

Remove unused types, functions, fields, and variables detected by `staticcheck` (U1000, SA4006). This is a non-functional cleanup PR that reduces dead code and improves maintainability.

| File | What was removed | Rule |
|------|-----------------|------|
| `pkg/binding/bindingpolicy-resolution.go` | Unused method `getDestinationsList` | U1000 |
| `pkg/binding/bindingpolicy.go` | Unused method `requeueSelectedWorkloadObjects` | U1000 |
| `pkg/status/status-controller.go` | Unused type alias `workStatusON` | U1000 |
| `pkg/status/binding.go` | Use `var` declaration for zero-value struct instead of composite literal | style |
| `test/performance/.../controller.go` | Unused struct field `bindingCreated` | U1000 |

No functional behavior changes. All removed symbols had zero references in the codebase.

## Related issue(s)

N/A — proactive cleanup of dead code detected by `staticcheck`.

cc @MikeSpreitzer @waltforme

### DIFF
--- a/pkg/status/binding.go
+++ b/pkg/status/binding.go
@@ -69,7 +69,7 @@ func (c *Controller) syncBinding(ctx context.Context, key string) error {
 // missingSCs, a slice of the missing StatusCollector object name(s), must be sorted.
 func (c *Controller) createOrUpdateStatusCollectorAvailableCondition(ctx context.Context, bdg *v1alpha1.Binding, missingSCs []string) error {
 	// compose tentative condition where LastTransitionTime is TBD
-	conditionTentative := v1alpha1.BindingPolicyCondition{}
+	var conditionTentative v1alpha1.BindingPolicyCondition
 	if len(missingSCs) != 0 {
 		conditionTentative = v1alpha1.BindingPolicyCondition{
 			Type:    v1alpha1.TypeStatusCollectorsAvailable,
--- a/pkg/binding/bindingpolicy-resolution.go
+++ b/pkg/binding/bindingpolicy-resolution.go
@@ -239,15 +239,6 @@ func (resolution *bindingPolicyResolution) matchesBindingSpec(bindingSpec *v1alp
 	return true
 }
 
-// getDestinationsList returns a sorted list of v1alpha1.Destination in the
-// resolution.
-func (resolution *bindingPolicyResolution) getDestinationsList() []v1alpha1.Destination {
-	resolution.RLock()
-	defer resolution.RUnlock()
-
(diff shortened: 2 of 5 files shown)


Norway-02:

Norway-02:

pkg/status/binding.go
```diff
@@ -69,7 +69,7 @@ func (c *Controller) syncBinding(ctx context.Context, key string) error {
 // missingSCs, a slice of the missing StatusCollector object name(s), must be sorted.
 func (c *Controller) createOrUpdateStatusCollectorAvailableCondition(ctx context.Context, bdg *v1alpha1.Binding, missingSCs []string) error {
 	// compose tentative condition where LastTransitionTime is TBD
-	conditionTentative := v1alpha1.BindingPolicyCondition{}
+	var conditionTentative v1alpha1.BindingPolicyCondition
```
MikeSpreitzer, line 72: Why is this change needed?

Norway-02: The original code used `conditionTentative := v1alpha1.BindingPolicyCondition{}` which explicitly assigned an empty struct to the variable. However, because `conditionTentative` is immediately overwritten in both branches of the `if len(missingSCs) != 0` block below, that initial assigned value is never used. This triggers the staticcheck rule `SA4006: A value assigned to a variable is never read before being overwritten`. Changing it to a `var` declaration avoids this redundant assignment and fixes the linter warning.

MikeSpreitzer: Please let the person who started a comment thread decide if and when it is resolved to their satisfaction.

IMHO this instance of SA4006 is a silly complaint. The Go language promises that the two statements mean exactly the same thing.

I have opened an Issue about this, https://github.com/dominikh/go-tools/issues/1729

Norway-02: Thanks for the context on thread resolution — noted, I won't do that again.

Point taken on SA4006 — if the team considers this a false positive / linter noise, I'm happy to revert this specific change and instead add a staticcheck suppression or config exclusion for it. Let me know your preference and I'll update the PR.

MikeSpreitzer: That would be overkill. The problem here is upstream in staticcheck, and there is now an Issue there for it. Simplest here is to do the proposed rewrite.

MikeSpreitzer (commented):
/lgtm
/approve

Issue #3950, opened 2026-07-25
Add Unit Test Coverage for ConcurrentMap and Transport Components


karthik120710:
## Description: 
Critical backend utilities concurrent-map.go and transport.go currently lack test coverage. This task requires writing comprehensive unit tests for:

concurrent-map.go: Test the thread-safe ConcurrentMap implementation including Set, Get, Remove, Len, and Iterator operations, with appropriate concurrency and edge case coverage
transport.go: Test the Transport interface helper types (Wrapee, NewWrapee, GetObject, GetID) and potentially concrete implementations

PR #3856, opened 2026-06-25
🌱 test: add concurrentmap tests

Norway-02:
🌱 test: add unit tests for ConcurrentMap

## Summary

Add unit tests for `ConcurrentMap` in `pkg/util/concurrent-map.go`.

Covers:
- Set/Get, including value update on existing key
- Remove, including remove on non-existent key
- Len, after insert, update, and remove
- Iterator, full pass and early stop on error
- Concurrent Set/Get/Len/Iterator/Remove from 50 goroutines
- Generic struct key/value types

Verified with:

    go test -v -race ./pkg/util/...

All tests pass, no races detected.

## Related issue(s)

Fixes #

PR #3951, opened 2026-07-25
Add Unit Test Coverage for ConcurrentMap and Transport Components

karthik120710:
## Description: 
This PR adds comprehensive unit test coverage for critical backend utilities that previously lacked tests:

pkg/util/concurrent-map.go: Added 9 test functions covering Set, Get, Remove, Len, Iterator operations, including concurrent access testing with 100 goroutines to verify thread-safety, and edge cases like overwrites and empty maps.
pkg/transport/transport.go: Added 7 test functions covering Wrapee helper types (NewWrapee, GetObject, GetID) for various resource types including namespaced resources, core group resources, and cluster-scoped resources.

#3950

### DIFF
--- a/pkg/transport/transport_test.go
+++ b/pkg/transport/transport_test.go
@@ -1,0 +1,210 @@
+/*
+Copyright 2024 The KubeStellar Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package transport
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func TestNewWrapee(t *testing.T) {
+	obj := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "apps/v1",
+			"kind":       "Deployment",
+			"metadata": map[string]interface{}{
+				"name":      "test-deployment",
+				"namespace": "default",
+			},
+		},
+	}
+
+	wr := NewWrapee(obj, true)
+
+	if wr.Object != obj {
+		t.Errorf("expected object to be set correctly")
+	}
+
+	if !wr.CreateOnly {
+		t.Errorf("expected CreateOnly to be true")
+	}
+}
+
+func TestNewWrapee_CreateOnlyFalse(t *testing.T) {
+	obj := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "v1",
+			"kind":       "ConfigMap",
+			"metadata": map[string]interface{}{
+				"name":      "test-configmap",
+				"namespace": "default",
+			},
+		},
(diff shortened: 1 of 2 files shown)


karthik120710:

karthik120710:
Hi @MikeSpreitzer and @waltforme, could you please take a look at this PR , its ready to merge.

pkg/util/concurrent-map_test.go
```diff
@@ -0,0 +1,234 @@
+/*
+Copyright 2024 The KubeStellar Authors.
```
MikeSpreitzer, line 2: This file was created in 2026, right?

pkg/util/concurrent-map_test.go
```diff
@@ -0,0 +1,234 @@
...
+			return &testError{msg: expectedErr}
+		}
+		return nil
+	})
+
+	if err == nil {
+		t.Errorf("expected error from iterator")
+	}
+
+	if err.Error() != expectedErr {
```
MikeSpreitzer, line 146: This fails to test whether `err` is `==` the `error` value returned by the func argument to `Iterator`.

pkg/util/concurrent-map_test.go
MikeSpreitzer, line 1: I see a lot of individual tests here that could be chained together in one, much shorter, Test function.

pkg/transport/transport_test.go
```diff
@@ -0,0 +1,210 @@
...
+package transport
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func TestNewWrapee(t *testing.T) {
```
MikeSpreitzer, line 26: Why does this one lack a more specific name?

pkg/transport/transport_test.go
```diff
@@ -0,0 +1,210 @@
...
+			"metadata": map[string]interface{}{
+				"name": "test-config",
+			},
+		},
+	}
+
+	wr1 := Wrapee{Object: obj, CreateOnly: true}
+	if !wr1.CreateOnly {
+		t.Errorf("expected CreateOnly to be true")
+	}
```
MikeSpreitzer, line 204: Isn't this covered by `TestNewWrapee`?

pkg/transport/transport_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+	wr2 := Wrapee{Object: obj, CreateOnly: false}
+	if wr2.CreateOnly {
+		t.Errorf("expected CreateOnly to be false")
+	}
```
MikeSpreitzer, line 209: Isn't this covered by `TestNewWrapee_CreateOnlyFalse` ?

pkg/transport/transport_test.go
MikeSpreitzer, line 1: This is much more source code than is needed. One TestXXX function can test multiple conditions, as is already exhibited. One TestXXX function can cover a collection of test cases, varying the api group, the create-only value, and the namespace.

pkg/util/concurrent-map_test.go
```diff
@@ -0,0 +1,234 @@
...
+		go func(i int) {
+			defer wg.Done()
+			cm.Remove(i)
+		}(i)
+	}
+
+	wg.Wait()
+
+	// Verify final state
+	if cm.Len() != 50 {
```
MikeSpreitzer, line 185: The condition tested here is not guaranteed to be true. There is no guarantee that the `cm.Remove` calls all happen after the corresponding `cm.Set`.

MikeSpreitzer:
Why did you add this when #3856 already exists?

pkg/transport/transport_test.go
MikeSpreitzer, line 1: Why did you add this when #3874 already exists?

PR #3957, opened 2026-07-27
Add proper error handling in CombinedStatus Resolver

### LINKED ISSUE #3886: bug: Add proper error handling in CombinedStatus Resolver
### Describe the bug

In the status package, there are explicit TODO comments calling out missing error handling. Errors occurring in these execution paths are silently ignored (swallowed), which can cause hard-to-debug issues in production where combined status fails to update without logging any warning or error to the controller logs.

N/A

### Steps To Reproduce

1. Inspect `pkg/status/combinedstatus-resolver.go` around line 330 and note the comment: `// TODO: handle errors`
2. Inspect `pkg/status/combinedstatus-resolution.go` around line 1009 and note the comment: `// TODO: handle errors`
3. Observe that errors in `getCombinedContentMap` and `getCombinedFieldSubject` are either swallowed or not surfaced to the main controller sync reconciliation loop.

### Expected Behavior

1. Errors from `getCombinedContentMap` (fetching source object metadata & spec) should be propagated up to `NoteWorkStatus`, `NoteStatusCollector`, and `NoteBindingResolution`.
2. The controllers (`workstatus`, `statuscollector`, `binding`) should receive these errors and retry the sync.
3. Errors or invalid types in `getCombinedFieldSubject` should be logged using the `klog.ErrorS` framework and reflected in the evaluation status.

### Additional Context

Locations of identified missing error handling:
- `pkg/status/combinedstatus-resolver.go:330`
- `pkg/status/combinedstatus-resolution.go:1009`

Rucha0901:
Fixes #3886

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,20 @@
+# GitHub CODEOWNERS File
+# Automatically assign reviewers based on which files are modified in a PR.
+# https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-codeowners
+
+# Default owners for all files in the repository
+*                   @MikeSpreitzer @KPRoche @pdettori @clubanderson @dumb0002 @waltforme
+
+# Component-specific code owners
+/pkg/binding/       @MikeSpreitzer @KPRoche @pdettori @clubanderson @dumb0002 @waltforme
+/pkg/status/        @MikeSpreitzer @KPRoche @pdettori @clubanderson @dumb0002 @waltforme
+/pkg/transport/     @MikeSpreitzer @KPRoche @pdettori @clubanderson @dumb0002 @waltforme
+/monitoring/        @MikeSpreitzer @KPRoche @pdettori @clubanderson @dumb0002 @waltforme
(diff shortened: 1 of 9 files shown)


magic-peach:
Thanks for the PR! The error propagation through the resolver is a real improvement — previously `getCombinedContentMap` logged with `runtime.HandleError` and silently generated CombinedStatuses from partial content, and the returned errors are now wrapped (`%w`) so the controller can requeue with backoff.

Two things to fix before this is mergeable:

**1. Stray files at the repo root (blocking)**
`issue_description.md` and `pr_description.md` are planning artifacts that should not be committed — they appear to have been added accidentally. Please remove both from the PR. Also, the `CODEOWNERS` file is an unrelated change to the error-handling fix; it would be cleaner as a separate PR so it can be reviewed on its own merits.

**2. Minor: repeated klog noise in `getCombinedFieldSubject`**
With the new `klog.ErrorS` calls, a transient unparseable subject will now be logged at `Error` level on every evaluation cycle for every affected WEC row. Since this function is called per-row per-aggregation pass (see the loop in `calculateCombinedFieldAggregation`), that could be noisy on a busy cluster. Consider logging once (the error is already surfaced in the aggregate's error string and reported to the user via the CombinedStatus errors), or downgrading the log level.

PR #3966, opened 2026-07-29
Support Wildcards [*] and Array Traversal in Custom JSONPath Parser

### LINKED ISSUE #3965: Support Wildcards [*] and Array Traversal in Custom JSONPath Parser
## Description
The custom JSONPath parser in `lexer.go` only supports a restricted subset of JSONPath: `.name` and `["name"]`. If a user attempts to use wildcard expressions (such as `[*]`) to filter or remove fields from array elements, the lexer fails with a syntax error, preventing array-level transformations in custom transforms.

## Proposed Changes
- Extend the `jsonpath` package (`lexer.go`) to support parsing wildcard symbols (`*` and `[*]`).
- Define a segment Kind structure (`SegmentKindField` and `SegmentKindWildcard`) to represent array wildcard operations.
- Update `QueryValue` in `eval.go` to recursively traverse `[]any` slices when encountering a wildcard segment.
- Implement `ArrayItemNode` structure to encapsulate referencing array elements.
- Add unit tests in `lexer_test.go` and `eval_test.go` targeting wildcard selections.
- Add transport integration tests using `HorizontalPodAutoscaler` `spec.metrics` array wildcard transformation.

## Goal
Support array wildcard `[*]`/`.*` traversal and item field removal in CustomTransformSpec.

Rucha0901:
## Summary
This PR extends the `jsonpath` package parser and evaluator to support wildcard segments `.*` and `[*]`. It introduces recursive array/slice traversal and defines `ArrayItemNode` to represent slice elements, enabling `Remove()` on fields inside items of slices.

## Related issue(s)
Fixes #3965

## Changes
- Modified `lexer.go` to parse dot-wildcard (`.*`) and bracket-wildcard (`[*]`), returning `Query` as a slice of `Segment`s.
- Modified `eval.go` to implement `ArrayItemNode` and a recursive `QueryValue` helper that traverses slice elements when encountering a wildcard.
- Updated `lexer_test.go` and `eval_test.go` with unit tests for wildcard selections and removal.
- Added integration test inside `generic_transport_controller_test.go` to verify array-level field stripping.

## Testing
- Ran unit tests `go test ./pkg/jsonpath/...` (passed).
- Ran integration tests `go test ./pkg/transport/generic/...` (passed).

### DIFF
--- a/pkg/jsonpath/eval.go
+++ b/pkg/jsonpath/eval.go
@@ -75,20 +75,61 @@ func (fn FieldNode) Remove() {
 	delete(fn.Object, fn.Key)
 }
 
+// ArrayItemNode is an element of a JSON array
+type ArrayItemNode struct {
+	Slice []any
+	Index int
+}
+
+var _ Node = ArrayItemNode{}
+
+func (an ArrayItemNode) Get() (JSONValue, bool) {
+	if an.Index < 0 || an.Index >= len(an.Slice) {
+		return nil, false
+	}
+	return an.Slice[an.Index], true
+}
+
+func (an ArrayItemNode) Remove() {
+	if an.Index >= 0 && an.Index < len(an.Slice) {
+		an.Slice[an.Index] = nil
+	}
+}
+
 // QueryValue applies `query` to `node`, invoking `yield` on each
 // of the nodes that the query produces, in a context where the document
 // root is `root`.
 func QueryValue(query Query, node Node, yield func(Node)) {
-	for _, fieldName := range query {
-		objA, ok := node.Get()
-		if !ok {
-			return
-		}
-		objM, ok := objA.(map[string]any)
-		if !ok {
-			return
-		}
-		node = FieldNode{objM, fieldName}
+	queryValueHelper(query, node, yield)
+}
+
+func queryValueHelper(query Query, node Node, yield func(Node)) {
+	if len(query) == 0 {
+		yield(node)
+		return
 	}
-	yield(node)
+
+	segment := query[0]
+	val, ok := node.Get()
+	if !ok {
+		return
+	}
+
+	if segment.Kind == SegmentKindField {
+		objM, ok := val.(map[string]any)
+		if !ok {
+			return
+		}
+		nextNode := FieldNode{Object: objM, Key: segment.Name}
+		queryValueHelper(query[1:], nextNode, yield)
+	} else if segment.Kind == SegmentKindWildcard {
+		sliceVal, ok := val.([]any)
+		if !ok {
+			return
+		}
+		for i := range sliceVal {
+			nextNode := ArrayItemNode{Slice: sliceVal, Index: i}
+			queryValueHelper(query[1:], nextNode, yield)
+		}
+	}
 }
--- a/pkg/jsonpath/eval_test.go
+++ b/pkg/jsonpath/eval_test.go
@@ -44,6 +44,63 @@ func TestEval(t *testing.T) {
 	if !jsonEqualities.DeepEqual(expected, actual) {
 		t.Errorf("Expected %#v, got %#v", expected, actual)
 	}
+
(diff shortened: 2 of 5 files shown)


pkg/jsonpath/eval.go
```diff
@@ -75,20 +75,61 @@ func (fn FieldNode) Remove() {
...
+func (an ArrayItemNode) Get() (JSONValue, bool) {
+	if an.Index < 0 || an.Index >= len(an.Slice) {
+		return nil, false
+	}
+	return an.Slice[an.Index], true
+}
+
+func (an ArrayItemNode) Remove() {
+	if an.Index >= 0 && an.Index < len(an.Slice) {
+		an.Slice[an.Index] = nil
```
magic-peach, line 95: `ArrayItemNode.Remove()` sets the array slot to `nil` rather than removing the element from the slice. This leaves `[null, ...]` holes in the document instead of deleting the node, which conflicts with the `Node.Remove` contract ("Remove deletes the node from the JSON document"). 

This is reachable by an end user: a `CustomTransform` with `Remove: ["$.spec.metrics[*]"]` would yield `ArrayItemNode`s (the wildcard is the terminal segment, so `queryValueHelper` hits the `len(query) == 0` base case and yields them directly), and each call would set `metrics[i] = null` rather than shrink the array. The resulting HPA object would contain `metrics: [null, null]` instead of `metrics: []`.

I realize shrinking the slice in place during traversal would be unsafe (the `for i := range sliceVal` range evaluates the length once), so `nil` may be a deliberate choice — but if so, it should be documented on `ArrayItemNode.Remove` and/or the PR description, since a user would reasonably expect `Remove` on a wildcard match to actually remove the matched elements. At minimum, there is no test covering `ArrayItemNode.Remove()` (the current Remove test in `eval_test.go` only removes a `FieldNode` via the wildcard).

pkg/jsonpath/eval.go
```diff
@@ -75,20 +75,61 @@ func (fn FieldNode) Remove() {
...
+	if segment.Kind == SegmentKindField {
+		objM, ok := val.(map[string]any)
 		if !ok {
 			return
 		}
-		objM, ok := objA.(map[string]any)
+		nextNode := FieldNode{Object: objM, Key: segment.Name}
+		queryValueHelper(query[1:], nextNode, yield)
+	} else if segment.Kind == SegmentKindWildcard {
+		sliceVal, ok := val.([]any)
```
magic-peach, line 126: Scope note: the wildcard branch only matches `[]any` values. Per RFC 9535 a wildcard segment should also enumerate all members of an object, so a query like `$.spec.*.name` (or `$["spec"].*`) on an object silently matches nothing — `queryValueHelper` returns without ever yielding. The PR is deliberately scoped to array traversal, which is fine, but this asymmetry is worth (a) documenting on the `Query`/`ParseQuery` doc comment and (b) a tiny lexer/eval unit test asserting the current (no-match) behavior, so the limitation is explicit rather than a silent surprise.

magic-peach (commented):
Nice, well-scoped change with a clean `Segment`/`SegmentKind` refactor and tests. Two inline comments above on `ArrayItemNode.Remove()` semantics (the `nil`-hole issue) and the array-only wildcard scope. A couple of smaller nits:

- `ParseQuery`'s doc comment in `lexer.go` still says "This parser only accepts segments of the form `.name` or `[string]`" — now stale given `.*` and `[*]` support. Worth updating along with the `Query` doc comment.
- `queryValueHelper`'s `if / else if` chain silently ignores any unknown `SegmentKind`; a `switch` with an explicit default would make future segment kinds fail loudly instead of producing no matches.
- Consider a couple of lexer error-case tests for `$[*` (unclosed bracket-wildcard) and `$[x]` (non-quote, non-wildcard content), since those are new branches in the parser.

PR #3972, opened 2026-08-09
🐛 fix(binding): prevent recursive read-lock deadlock in GetSingletonReportedStateRequestsForBinding

magic-peach:

`bindingPolicyResolver.GetSingletonReportedStateRequestsForBinding` acquires the resolver's read lock and then **re-acquires** it via two nested calls:

- `getResolution(bindingPolicyKey)` — which calls `resolver.RLock()` again
- `GetReportedStateRequestForObject(objId)` — which calls `resolver.RLock()` again

Go's `sync.RWMutex` is **not reentrant**. As soon as a writer (e.g. `createResolution`, `DeleteResolution`, or `NoteBindingPolicy` → `createResolution`) queues behind the outer read lock, any nested `RLock()` blocks forever. The queued writer then waits for the read lock to be released — deadlock. This function is called from `pkg/binding/binding.go:90` on **every binding sync**, so the controller can hang indefinitely.

This is the same bug class as the previously-fixed `compareCombinedStatus` recursive read-lock deadlock (#3945 / #3900), just in a different location.

**The fix:**

Acquire the resolver's read lock **once** and use unlocked helpers for the inner accesses:
- `getResolutionLocked` — the unlocked map lookup (the existing public/private `getResolution` now delegates to it)
- `getReportedStateRequestForObjectLocked` — the unlocked body of `GetReportedStateRequestForObject`

**Tests added:**

- `TestGetResolutionLockedWhileWriterQueued` — deterministic: holds the read lock, queues a writer, and proves the locked helpers return without re-locking (this is exactly the scenario that deadlocked before).
- `TestGetSingletonReportedStateRequestsForBinding` — verifies correct results (per-object singleton request flags and qualified WEC counts), exercising the new helpers.
- `TestGetSingletonReportedStateRequestsForBindingNoResolution` — nil resolution returns nil.
- `TestGetSingletonReportedStateRequestsForBindingConcurrent` — concurrent readers + a writer; **verified it hangs on the old code** (recursive lock) and passes with the fix, run under `-race`.

Verified: reverting to the nested-locking implementation makes `TestGetSingletonReportedStateRequestsForBindingConcurrent` deadlock and fail; the fix makes it pass (`-race` clean).

**Which issue(s) this PR fixes:**

Related to the same deadlock class as #3899/#3945.

None.

### DIFF
--- a/pkg/binding/bindingpolicy-resolver.go
+++ b/pkg/binding/bindingpolicy-resolver.go
@@ -363,6 +363,13 @@ func (resolver *bindingPolicyResolver) GetReportedStateRequestForObject(objId ut
 	resolver.RWMutex.RLock()
 	defer resolver.RWMutex.RUnlock()
 
+	return resolver.getReportedStateRequestForObjectLocked(objId)
+}
+
+// getReportedStateRequestForObjectLocked is the unlocked implementation of
+// GetReportedStateRequestForObject. The caller must hold the resolver's
+// read lock.
+func (resolver *bindingPolicyResolver) getReportedStateRequestForObjectLocked(objId util.ObjectIdentifier) (bool, sets.Set[string], bool, sets.Set[string]) {
 	var singletonRequested bool
 	var multiWECRequested bool
 	singletonWECs := sets.New[string]()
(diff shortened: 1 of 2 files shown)


vjymisal0 (commented):
Very thorough fix for the recursive read-lock deadlock in \pkg/binding/bindingpolicy-resolver.go\.

- Resolves the non-reentrant \sync.RWMutex\ deadlock when writers queue behind the outer \RLock()\ by delegating to unlocked helpers (\getResolutionLocked\ and \getReportedStateRequestForObjectLocked\).
- The concurrency test \TestGetSingletonReportedStateRequestsForBindingConcurrent\ and deterministic writer queue test provide solid regression protection under \-race\.

LGTM!

PR #3973, opened 2026-08-09
🐛 fix(transport): stop removeFinalizer from mutating informer-cache Binding objects

magic-peach:

`removeFinalizer` in `pkg/transport/generic/generic_transport_controller.go` compacts the finalizers slice **in place** and only creates a deep copy afterwards:

```go
finalizersList := binding.GetFinalizers()  // backing array of the cache object
...
finalizersList[index] = finalizersList[i]  // ← mutates the cache object
...
updatedBinding := binding.DeepCopy()       // ← too late
```

`GetFinalizers()` returns the underlying slice of the object handed in, and that object comes straight from the informer's `BindingLister`. The file's own `addFinalizer` documents the invariant: *"objects returned from a BindingLister must be treated as read-only"* — and correctly deep-copies **first**. `removeFinalizer` violates the same invariant:

- Removing `"B"` from `["A","B","C"]` leaves the shared cache object holding `["A","C","C"]` — silent corruption of the informer cache.
- Multiple workers reading the same cached `Binding` concurrently can race on the shared backing array.

**The fix:**

Check for the finalizer's presence **without mutating** (using `slices.Contains`), then `DeepCopy()` first and compact the copy's own slice — mirroring `addFinalizer`. The returned result is identical to before; only the (buggy) in-place mutation of the input is gone.

**Tests added:**

`TestRemoveFinalizer` with subtests:
- removes a middle finalizer without mutating the input object
- removes the last remaining finalizer
- finalizer-not-present returns the same object, unchanged

**Verified** the test fails against the old implementation (it reports `original object was mutated: [keep-1 keep-2 keep-3 keep-3]`) and passes with the fix, clean under `-race`.

None.

### DIFF
--- a/pkg/transport/generic/generic_transport_controller.go
+++ b/pkg/transport/generic/generic_transport_controller.go
@@ -1224,25 +1224,24 @@ func addFinalizer(binding *v1alpha1.Binding, finalizer string) (*v1alpha1.Bindin
 // removeFinalizer accepts Binding object and removes the provided finalizer if present.
 // It returns the updated (or not) Binding and an indication of whether it updated the object's list of finalizers.
 func removeFinalizer(binding *v1alpha1.Binding, finalizer string) (*v1alpha1.Binding, bool) {
-	finalizersList := binding.GetFinalizers()
-	length := len(finalizersList)
-
-	index := 0
-	for i := 0; i < length; i++ {
-		if finalizersList[i] == finalizer {
-			continue
-		}
-		finalizersList[index] = finalizersList[i]
-		index++
-	}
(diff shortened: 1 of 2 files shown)


vjymisal0 (commented):
Crucial bug fix for informer cache integrity in \pkg/transport/generic/generic_transport_controller.go\.

- Eliminates in-place mutation of the shared \BindingLister\ cache slice before \DeepCopy()\, protecting against silent cache corruption and worker race conditions.
- Uses \slices.Contains\ for clean read-only presence checking.
- The unit test in \TestRemoveFinalizer\ explicitly verifies input object immutability.

LGTM!

PR #3974, opened 2026-08-09
🐛 fix(status): correct combined-field AVG/MIN/MAX aggregation over missing subjects

magic-peach:

`calculateCombinedFieldAggregation` in `pkg/status/combinedstatus-resolution.go` mishandles rows whose combined-field subject is missing or non-numeric (`getCombinedFieldSubject` returns `nil` for those):

1. **AVG divides by the wrong denominator.** It divides the sum by `len(rows)` instead of the number of valid subjects, so missing values skew the average down (e.g. three rows with two valid values `5`,`5` produced `10/3 ≈ 3.33` instead of `5`).
2. **MIN / MAX report `+Inf` / `-Inf`** when no row has a valid subject, because the `math.Inf` sentinels are formatted and returned as the aggregate with no error.
3. **AVG with rows but no valid subject returned `0`** (with an empty error) instead of an error.

**The fix:**

- Count only rows with a valid subject for the AVG denominator.
- When MIN/MAX/AVG have nothing to aggregate, report a `no values to ...` error (mirroring the existing `no values to average` handling) instead of emitting a meaningless numeric sentinel. A more specific per-WEC parse error, when present, is preserved.

**Tests added:**

`TestCalculateCombinedFieldAggregation` covers all five aggregator types (COUNT, SUM, AVG, MIN, MAX), including missing subjects, string subjects parsed as floats, and the all-missing cases for AVG/MIN/MAX.

**Verified** the new tests fail against the old implementation (the AVG denominator, all-missing AVG, MIN `+Inf`, and MAX `-Inf` cases) and pass with the fix, clean under `-race`.

None.

### DIFF
--- a/pkg/status/combinedstatus-resolution.go
+++ b/pkg/status/combinedstatus-resolution.go
@@ -938,28 +938,33 @@ func calculateCombinedFieldAggregation(combinedFieldNamedAgg v1alpha1.NamedAggre
 		}
 		numStr = strconv.FormatFloat(sum, 'g', -1, 64)
 	case v1alpha1.AggregatorTypeAvg:
-		var avg float64 = math.NaN()
-		if count := len(rows); count > 0 {
-			sum := 0.0
-			for dest, row := range rows {
-				subject, err1 := getCombinedFieldSubject(combinedFieldNamedAgg, row)
-				if err1 != "" {
-					if errStr == "" {
-						errStr = fmt.Sprintf("for WEC %s, %s", dest.ClusterId, err1)
-					}
+		avg := math.NaN()
+		sum := 0.0
+		count := 0
+		for dest, row := range rows {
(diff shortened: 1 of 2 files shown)


vjymisal0 (commented):
Excellent fix for \calculateCombinedFieldAggregation\ edge cases in \pkg/status/combinedstatus-resolution.go\:

- Correctly uses the count of valid subjects rather than total row count for the \AVG\ denominator, preventing missing subjects from skewing averages downward.
- Prevents \math.Inf\ sentinels from leaking into \MIN\ and \MAX\ when no rows contain a valid subject, returning clean error messages instead.
- Unit tests in \combinedstatus-resolution_test.go\ comprehensively cover all 5 aggregation types and nil-subject edge cases.

LGTM!

PR #3975, opened 2026-08-09
fix: remove stale empty key sets from reverse index map

magic-peach:
## Summary

`IndexedMapToComparable` keeps the reverse index map (`Value -> Set[Key]`) in sync with the forward map, but `Put` and `Delete` only remove the affected key from the old value's key set. When that set becomes empty, the empty-set entry stays in the reverse map forever.

### Impact
- **Memory leak**: one stale entry per reassignment/deletion, unbounded over the lifetime of a long-running controller. In the status controller this is `workStatusToObject` (`cache.ObjectName -> ObjectIdentifier`), whose values churn as WorkStatus objects are reassigned to workload objects.
- **Incorrect inverse iteration**: `ReadInverse().Iterate2` reports values that no longer have any keys in the forward map, which can mislead inverse-map consumers.

## Fix
After removing a key from the old value's key set in both `Put` and `Delete`, drop the reverse entry entirely if the set is now empty. The `val == oldVal` re-put path remains correct: deleting the emptied set triggers the `hadVal == false` branch, which recreates and re-inserts the set.

## Testing
- Added `TestIndexedMapToComparableNoStaleReverseEntries` with 4 subtests (delete, reassign, shared-value reassign, re-put same value). Verified it **fails on the old code** (stale empty sets reported) and passes with the fix.
- `go test ./pkg/abstract/...` and `go vet ./pkg/abstract/...` both clean.

### DIFF
--- a/pkg/abstract/map-to-comparable.go
+++ b/pkg/abstract/map-to-comparable.go
@@ -61,6 +61,9 @@ func (imc *IndexedMapToComparable[Key, Val]) Put(key Key, val Val) (Val, bool) {
 	if had {
 		oldKeys, _ := imc.reverse.Get(oldVal)
 		oldKeys.Delete(key)
+		if oldKeys.Len() == 0 {
+			imc.reverse.Delete(oldVal)
+		}
 	}
 	valKeys, hadVal := imc.reverse.Get(val)
 	if hadVal {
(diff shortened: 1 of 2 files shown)


vjymisal0 (commented):
Clean fix for the reverse index map memory leak in \IndexedMapToComparable\.

- Dropping the reverse entry when \oldKeys.Len() == 0\ in both \Put\ and \Delete\ ensures inverse iteration stays strictly consistent with the forward map.
- The 4 subtests in \TestIndexedMapToComparableNoStaleReverseEntries\ provide solid regression coverage for reassignment, deletion, shared-value, and re-put paths.

LGTM!

